PR #4, opened 2026-06-15
Release the mach host port in memoryUsage()

Cashie1597:
`SystemInfo.memoryUsage()` calls `mach_host_self()` inline but never releases the returned send right. `mach_host_self()` hands the caller a port reference it must deallocate with `mach_port_deallocate`, so every call leaks one mach port.
`memoryUsage()` runs every ~2s while the menu panel is open (via `SystemMonitor.refresh()`), plus in `--selftest`, so the leak accumulates during normal use.
Capture the host port, use it, and `mach_port_deallocate` it after `host_statistics64` — the same pattern already applied to the twin leak in `SystemMonitor.readCPUUsage()`. This was the only remaining unpatched `mach_host_self()` call site.
- `swift build` (debug + release): clean, no new warnings.
🧙 Built with [WOZCODE](https://wozcode.com)

### DIFF
--- a/Sources/Vorssaint/Services/SystemInfo.swift
+++ b/Sources/Vorssaint/Services/SystemInfo.swift
@@ -35,11 +35,13 @@ enum SystemInfo {
     static func memoryUsage() -> (used: UInt64, total: UInt64)? {
         var stats = vm_statistics64()
         var count = mach_msg_type_number_t(MemoryLayout<vm_statistics64>.stride / MemoryLayout<integer_t>.stride)
+        let host = mach_host_self()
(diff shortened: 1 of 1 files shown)


vorssaint:
Thank you for catching this, and for the clear write-up! 🙏

This is fixed in v2.10.0, which just shipped. I went with a `defer`-based `mach_port_deallocate` so the host port is released on every exit path, and applied the same fix to the matching `mach_host_self()` call in the CPU sampling (`SystemMonitor.readCPUUsage`), so both call sites are now covered.

Since the fix is already on `main` and released, I'm closing this PR. The catch was spot on and much appreciated. Thanks for contributing!

PR #26, opened 2026-06-16
fix: resize popover in real time as sections collapse and expand

mr-steveryan:
Collapsing or expanding a panel section via its chevron didn't resize
the popover to match. The window stayed at its previous height, leaving
empty space after a collapse or showing a scrollbar instead of growing
after an expand. Toggling a feature in Settings (e.g. Volume Mixer)
would temporarily "fix" the height but then overwrite it again with a
stale reading — because both paths shared the same fragile mechanism.

**Issue**
Issue
**Root cause**

`OverlayScrollView.updateNSView` is the only place `contentHeight` gets
measured, and it only runs when `MenuPanelView.body` re-evaluates.
Collapsing a section flips `@State` inside a child `PanelSection`, which
re-renders inside the `NSHostingView`'s own SwiftUI graph and never
re-triggers the parent's body — so `updateNSView` never fired and the
height was never re-read. The single synchronous `fittingSize` read that
did fire (e.g. on a feature toggle) raced the 0.18 s collapse animation
and landed on a stale value.

**Fix**

Fix

Subclass `NSHostingView` as `HeightReportingHostingView` and override
`layout()` to fire a callback after every AppKit layout pass. Wire that
callback to update `measuredHeight`, guarded by a 0.5 pt threshold to
break any measure → resize → measure loop. `updateNSView` re-binds the
callback each pass so it always captures the current binding.

This means the popover height tracks the real content height continuously
— including through the frames of the collapse animation — rather than
from a single reading taken whenever SwiftUI happened to re-evaluate the
parent view.

The existing `.frame(width: 332, height: min(contentHeight, maxHeight))`
cap is unchanged, so the overlay scroller still appears only when content
genuinely exceeds the usable screen height.[](url)

### DIFF
--- a/Sources/Vorssaint/UI/MenuPanel/MenuPanelView.swift
+++ b/Sources/Vorssaint/UI/MenuPanel/MenuPanelView.swift
@@ -135,8 +135,9 @@ struct MenuPanelView: View {
 /// A vertical scroll container that always uses an overlay scroller, so it never
 /// reserves a legacy gutter on the right (which, when the system is set to always
 /// show scroll bars, would push the fixed-width panel content off-center). The
-/// content is pinned to the full width; its natural height is reported back so the
-/// popover can size itself and cap to the screen.
+/// content is pinned to the full width and reports its natural height back after
+/// every layout pass, so the popover sizes itself to fit and only scrolls once the
+/// content is taller than the screen.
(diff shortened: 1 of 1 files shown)


vorssaint:
Thank you for the contribution! I tested the behavior locally in the Developer build and it feels great. The popover now follows collapse and expand changes much more naturally, so I am going to merge this.

PR #55, opened 2026-06-18
Add universal binary support to release builds

jdenoy:
## What

Adds a `--universal` flag to `build.sh` that produces a universal (fat) binary supporting both Apple Silicon (arm64) and Intel (x86_64) Macs. The release workflow now passes this flag, so every tagged release ships a single DMG that runs natively on both architectures.

## How

When `--universal` is passed, `build.sh` compiles the target twice using Swift's `-target` flag (once for `arm64-apple-macosx14.0`, once for `x86_64-apple-macosx14.0`), then merges the two binaries with `lipo -create`.

Without the flag the script behaves exactly as before (native arm64 only), so local dev builds are unaffected.

## Changes

- `build.sh`: added `--universal` flag and conditional lipo merge step
- `.github/workflows/release.yml`: passes `--universal` to the build step; updated release notes footer to mention Apple Silicon and Intel


vorssaint:
thanks for reporting this, and sorry. i am not planning to add intel support right now. it is unstable because permissions work differently, and i do not have an intel machine to test it properly. as a solo developer, supporting it would create more bugs and support requests that i could not properly verify, so it does not make sense for the project at the moment. sorry again!

PR #71, opened 2026-06-20
feat: raise per-app volume boost limit from 200% to 500%

alhaouem:
Raise the per-app volume mixer's boost ceiling from 200% to 500%.

**Before:** per-app volume sliders capped at 200%.
**After:** cap raised to 500%, giving quiet sources more headroom.

- Centralized on `AppVolumeMixer.maxVolume` (2.0 → 5.0), which drives the slider range, gain clamping, custom-slider math, and the accessibility increment.
- `Defaults.sanitizedAppVolume` keeps its own literal `5` (the constant isn't importable in the standalone test build) with a sync comment.
- The realtime gain path already hard-clips to `[-1, 1]` when boosting, so no distortion/overflow changes were needed.

## Testing

- `./build.sh` builds clean, no warnings.
- `./build.sh --test` passes (438 checks), including updated clamp assertions (`6 → 5`, plus an in-range `3 → 3` case).
- Verified live: boosting and attenuating are audible end-to-end.

## Notes

- No new user-facing strings, so no translation changes are required.

### DIFF
--- a/Sources/Vorssaint/Core/Defaults.swift
+++ b/Sources/Vorssaint/Core/Defaults.swift
@@ -27,7 +27,7 @@ enum DefaultsKey {
     static let previewSize = "previewSize"                // app switcher + dock preview thumbnail size
     static let autoCheckUpdates = "autoCheckUpdates"
     static let releaseNotesOnUpdate = "releaseNotesOnUpdate" // show What's New after an update
-    static let appVolumes = "appVolumes"                  // [bundle id: 0...2]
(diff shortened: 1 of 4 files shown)


Arachne1111:
Could you please add a universal audio-output switch menu above the app-specific audio-outputs to quickly adjust the entire system to one system-output. Setting the system-output should reset all app-specific audio-outputs to what hass been set as the universal one.

vorssaint:
not planned sorry, in my test, higher volumes than 200% don't "boost" it actually, just clip the audio
but thx for the pr!!

PR #86, opened 2026-06-22
Fix mixer silence with virtual output devices (BoomAudio, SoundSource)

AlexDesign420:
## Summary

- The volume mixer produces complete silence when the system's default output is a virtual audio device (BoomAudio, SoundSource, BlackHole, etc.)
- Root cause: virtual devices lack a hardware clock, so the aggregate device never drives its IO proc — the process tap mutes the app but no scaled audio reaches the speakers
- Fix detects virtual output devices via `kAudioDevicePropertyTransportType` and adds the real hardware output (built-in or USB) as the aggregate's clock source
- Also syncs the aggregate's sample rate to the target device and zeroes output buffers before writing to prevent partial-buffer drops on multi-channel setups

## Test plan

- [ ] Set a virtual device (BoomAudio, SoundSource, BlackHole) as default output
- [ ] Open the mixer, adjust any app's volume to 99% or 101%
- [ ] Verify audio continues playing (previously: complete silence)
- [ ] Test with built-in speakers as default output (should still work as before)
- [ ] Test volume boost (>100%) and mute (0%) functionality

### DIFF
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -652,12 +652,17 @@ private final class TapGainEngine: GainEngine {
 
         let description = CATapDescription(stereoMixdownOfProcesses: objects)
         description.muteBehavior = .mutedWhenTapped
-        description.isPrivate = true
         guard AudioHardwareCreateProcessTap(description, &tapID) == noErr, tapID != 0 else {
             return nil
         }
 
-        let aggregate: [String: Any] = [
+        // Virtual output devices (BoomAudio, SoundSource, etc.) lack a
+        // hardware clock, so an aggregate backed only by them never drives
+        // its IO proc. Detect virtual devices and add the real hardware
+        // output as clock source.
+        let clockUID: String? = Self.hardwareClockUID(for: outputDeviceUID)
+
+        var aggregate: [String: Any] = [
             kAudioAggregateDeviceNameKey: "Vorssaint Mixer",
             kAudioAggregateDeviceUIDKey: UUID().uuidString,
             kAudioAggregateDeviceIsPrivateKey: true,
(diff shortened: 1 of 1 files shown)


vorssaint:
Thanks for the PR. I tested this locally in the Developer build.
The patch builds, the test suite passes, selftest passes, and I was able to exercise the mixer with the available local virtual output device without crashes or permission errors. The general idea makes sense, a virtual output may need a real hardware clock source for the aggregate device.
I do not think this is ready to merge as-is yet.
Main blockers:
Please keep description.isPrivate = true on the process tap. Removing it is unrelated to the clock fix and makes the tap visible outside the creating process, which is not the intended security/privacy boundary for this mixer.

The hardware clock selection is too narrow and somewhat arbitrary. It currently chooses the first Built-in or USB output device. That misses common real outputs like HDMI, DisplayPort, Bluetooth, AirPlay and Thunderbolt, and it does not prove that the selected device is the actual physical destination behind the virtual device.

This still needs validation with the real target scenario: BoomAudio, SoundSource or BlackHole set as the system default output, then changing an app volume away from 100%. I could only test with the local virtual device available on my machine, which is not equivalent to those routing setups.

Please also keep the existing clipping comment or otherwise avoid unrelated cleanup in this patch. The diff should stay focused on the virtual-output clock fix.

If you update the PR to keep the tap private, make the clock selection safer, and confirm the fix on one of the real virtual-output setups listed above, I can review it again.

TYSM for contributing!

AlexDesign420:
Thanks for the detailed review! I've pushed an update (701d4ac) addressing each point:

**Tap stays private** — Restored `description.isPrivate = true` on the process tap. Removing it was an accidental, unrelated change and has been reverted.

**Safer, wider clock selection** — The clock source is no longer limited to the first Built-in or USB device. It now accepts any device that is *not* virtual or aggregate and has output channels (HDMI, DisplayPort, Bluetooth, AirPlay, Thunderbolt, FireWire, PCI, etc.), and it **prefers the current system default output** when that is real hardware, falling back to any other hardware output. This removes the arbitrary built-in/USB-only assumption.

**Clock is a clock, not a second output** — The hardware clock device is now set only via `kAudioAggregateDeviceClockDeviceKey`; it is no longer added to the sub-device list, so audio is never duplicated onto a second device.

**Focused diff** — Kept the existing clipping comment and avoided unrelated cleanup. Additionally synced the aggregate's buffer frame size (in addition to sample rate) to the target device.

**Confirmed on a real virtual-output setup** — Tested with BoomAudio (`GDAudioDevice`, transport type Virtual) set as the system default output. Before: adjusting any app's volume away from 100% produced complete silence. After: audio plays at every volume level — 0% = silence, partial = attenuated, >100% = boosted. Built-in/wired outputs still behave exactly as before (untouched passthrough at 100%).

Ready for another look — thanks again!

AlexDesign420:
@vorssaint

vorssaint:
Thanks for the update. This is much closer.

Before I can merge it, please remove third party app names from source comments. Generic wording like virtual output devices is enough in the code.

I also need to test this on my machine with a real virtual output setup before merging, since this path touches Core Audio routing. The build, unit tests, and selftest pass on my side, but I do not want to merge this until the real audio path has been checked.

vorssaint:
Already fixed!

PR #107, opened 2026-06-25
Improve German wording for the lid-closed feature

FinkeFlo:
Small German translation polish for the Keep Awake / lid-closed feature.

Two things read awkwardly in German:

- „geschlossener Deckel" isn't really idiomatic — „zugeklappt" (and „geschlossenes Display" where a noun is needed) sounds much more natural.
- „Fragt **einmal** nach dem Administratorpasswort" is ambiguous (it can read as „just ask" rather than „once"). „**einmalig**" makes the intended „once" meaning clear, matching the English string.

All changes are confined to `Strings+German.swift` (8 strings, no key or behavior changes). `./build.sh` finishes without warnings and `--selftest` reports `SELFTEST OK`.


vorssaint (approved):
Thanks for the PR! Merged

PR #117, opened 2026-06-28
Gate memory purge controls behind manual confirmation

Cashie1597:
Summary

* Makes standard and auto modes non-privileged simulation only.
* Keeps deep and max manual-only and confirmation-gated.
* Adds receipts for every purge attempt.
* Adds safety tests proving the auto and launch paths cannot reach the privileged purge flow.

Validation

* swift test passed (4 tests).

Safety

* Standard and auto modes never invoke a purge binary.
* Deep runs /usr/bin/purge only after explicit user confirmation.
* Max runs /usr/sbin/purge via AdminShell only after explicit user confirmation.
* Every purge attempt generates a receipt.
* No purge occurs on launch.
* Auto mode is disabled by default.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,6 +11,11 @@ let package = Package(
         .executableTarget(
             name: "Vorssaint",
             path: "Sources/Vorssaint"
+        ),
+        .testTarget(
+            name: "VorssaintTests",
+            dependencies: ["Vorssaint"],
+            path: "Tests/VorssaintTests"
         )
     ]
 )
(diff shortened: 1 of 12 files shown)


vorssaint:
Thanks for the contribution. I’m going to close this PR without merging.

This change is not a safe fit for Vorssaint as-is. It appears to turn the app into a different product named MemoryKill, changes the app identity and menu bar behavior, replaces the main panel, repurposes the existing global hotkey away from Keep Awake, changes several core defaults, and adds destructive or high-risk actions such as killing processes, clearing broad user cache/log folders, flushing DNS, and running purge commands.

It is also a draft PR, currently conflicting, based on an old branch, and it fails the repository’s current test path with default-behavior regressions. For this project, any memory-related feature would need to be redesigned as a small, optional Vorssaint feature inside the existing Monitor/panel structure, off by default, localized, non-destructive by default, and validated through the current build/test flow.

Because of that, this PR will not be merged. Thanks again for taking the time to explore the idea.

PR #125, opened 2026-06-29
Fix mixer crash on duplicate app ids

lintran-cn:
## Summary
- Coalesce mixer app entries that share the same stable app id before publishing them.
- Merge duplicate audio object lists and playing state so multi-process apps remain a single mixer row.
- Avoid Dictionary(uniqueKeysWithValues:) in engine reconciliation so duplicate ids cannot trigger a Swift fatal error.

## Why
Vorssaint can discover multiple audio process groups for the same bundle id, for example com.tencent.WeWorkMac. The previous reconciliation path assumed ids were unique and crashed with Duplicate values for key.

## Tests
- ./build.sh --test
- swift build


vorssaint:
Thanks for the fix. Merged.

PR #135, opened 2026-06-30
add x86_64 CPU support

MoliZhaoCn:
At present, only 19 MacBook Pro Intel Core i7 9750H CPUs have been tested. I don't have an additional MacBook for testing, which is theoretically feasible on other versions of CPUs.
And I don't have a MacBook with Apple chips for testing, so I can't verify the current availability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         run: ./Tools/ci-setup-signing.sh
 
       - name: Build
-        run: ./build.sh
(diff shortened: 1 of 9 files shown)


vorssaint:
hi, i don't have plans to add intel support rn

i'm the only maintainer, and i don't own an intel mac. adding support would mean that if issues come up in the app on intel macs, i wouldn't be able to fix them, or even test fixes/new features, which makes it not viable for now. but i'm working on getting an intel mac so i can test properly and add support in the future

PR #149, opened 2026-07-01
feat(keyboard): add keyboard cleaning mode

Games55k:
## Summary

This PR adds a toggleable Keyboard Cleaning feature for temporarily locking keyboard input while cleaning MacBook keys.

The feature is available from Quick Controls and has its own Settings page. It shows the current cleaning status, can be hidden from the panel like other controls, and guides the user to grant the required Input Monitoring and Accessibility permissions before activation.

Screenshot 2026-07-02 at 01-33-19
Screenshot 2026-07-02 at 01-32-52

While Keyboard Cleaning is active, keyboard debounce is suspended so the cleaning lock owns keyboard handling. When cleaning mode is turned off, keyboard debounce is restored from the user's preferences.

The PR also adds the new user-facing strings across the supported localizations.


vorssaint:
Thanks for the contribution. I am going to close this without merging.

This removes the existing safety exits from Keyboard Cleaning, including the visible overlay, the keyboard unlock gesture, and the automatic timeout. A keyboard lock needs those recovery paths before it can ship.

A future version would need to keep the current escape paths while improving the panel and Settings placement.

Games55k:
Thx，it is a mistake.

PR #178, opened 2026-07-06
Fix repeated Shift+Tab back navigation in the switcher

victorbuchalla:
## Behavior before

With the switcher open, holding the shortcut modifier plus Shift and pressing Tab repeatedly only moved the selection backward once. Every Tab after the first was ignored until Shift was released and pressed again, so walking backward through several windows required re-pressing Shift for each step. Starting a session directly with e.g. ⌘⇧Tab had the same problem.

## Behavior after

Holding the modifiers and pressing Tab repeatedly moves the selection backward one step per press, matching the system switcher. Pressing Shift alone still navigates backward immediately, and a combined shift+tab gesture still moves a single step.

## Cause and fix

`handleKeyDown` swallowed Tab whenever `shiftBackNavigationHeld` was set, i.e. for as long as Shift stayed down. That guard exists so the Tab belonging to the same shift+tab gesture does not double-step after the Shift press already navigated via `flagsChanged`, but it also discarded every subsequent Tab.

The permanent flag is replaced with a short one-shot window (`shiftNavigationTabSwallowDeadline`, 300 ms) armed when a Shift press navigates backward: only the Tab of that gesture is swallowed, and later Tabs with Shift held navigate normally.

## Checks

- `./build.sh` finishes without warnings
- `./build/Vorssaint --selftest` passes (SELFTEST OK)
- Verified manually: forward cycling, repeated backward cycling with Shift held, shift-only back navigation, and sessions started reversed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
+++ b/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
@@ -5,6 +5,7 @@ import AppKit
 import ApplicationServices
 import Combine
 import CoreGraphics
+import QuartzCore
 import SwiftUI
 
 private struct SwitcherSourceContext {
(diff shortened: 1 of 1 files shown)


vorssaint:
Hello, @victorbuchalla 

First of all, thanks for the contribution, but this behavior is intended. You don't need to press Shift + Tab to move backwards in the App Switcher, just need to press Shift (without the Tab)

Shift + Tab is actually a longer shortcut and I personally don't think it is necessary (especially when you can press fewer buttons to do the same thing)

PR #179, opened 2026-07-07
fix: monitor memory option placement

Games55k:
## Summary

This PR fixes the placement of the memory menu bar option in Settings > Monitor.

Previously, the memory-specific "pressure dot" option appeared after the full menu bar metric order list, while the network-specific "upload above download" option appeared directly under the Network row. That made the memory option feel detached from the Memory row it controls.

After this change, the memory pressure dot option is rendered directly under the Memory row, matching the existing Network option layout.

This also includes a small i18n cleanup for the `Dock Preview` title across supported localizations.

## Behavior

Before:
  - Enabling the Memory menu bar metric showed "Pressure dot" at the bottom of the menu bar metrics section.
  - Network's upload/download order option appeared correctly under Network.

Screenshot 2026-07-07 at 10-24-22

After:
  - Enabling the Memory menu bar metric shows "Pressure dot" directly under Memory.
  - Network behavior is unchanged.

Screenshot 2026-07-07 at 10-21-34

 ## Testing

  - `swift build`
  - `./build.sh --dev`

### DIFF
--- a/Sources/Vorssaint/UI/Settings/MonitorSettings.swift
+++ b/Sources/Vorssaint/UI/Settings/MonitorSettings.swift
@@ -11,10 +11,8 @@ import UniformTypeIdentifiers
 struct MonitorSettings: View {
     @ObservedObject private var l10n = L10n.shared
 
-    @AppStorage(DefaultsKey.menuBarMemory) private var menuBarMemory = false
     @AppStorage(DefaultsKey.menuBarCombineTemperatures) private var combineTemperatures = true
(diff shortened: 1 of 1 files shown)


vorssaint:
Thanks for the fix. I merged the memory option placement change and it works well.

I removed the second commit before merging. Dock Preview is the feature name and it stays the same in every language on purpose. The rename was also partial since the intro screen still said Dock Preview. The Switcher wording fix for Simplified Chinese was a good catch though and it will land with the next release.

Games55k:
It's ok!

vorssaint:
Thanks for the contribution @Games55k!

PR #182, opened 2026-07-07
Add dock click window cycling (like Command-Tilde)

CN120:
## Motivation

macOS's Command-Tilde shortcut cycles through an app's windows, but there's no way to do this from the Dock. This adds an option so clicking an active app's Dock icon rotates through its windows instead of (or in addition to) minimizing them -- useful for quickly switching between windows without reaching for the keyboard.

## Approach

The existing `DockClickService` already intercepts clicks on Dock icons via a CGEvent tap and resolves which app was clicked using Accessibility. This PR adds a new `cycleWindows` action alongside the existing `minimize`/`restore` actions:

- When the feature is enabled and the frontmost app has 2+ unminimized windows, a Dock click raises the next window using `kAXRaiseAction` and sets it as the focused window
- The cycle finds the currently focused window in the AX window list and advances to the next one (wrapping around)
- A separate `dockClickCycleWindows` preference controls the feature independently from the minimize behavior
- When both are enabled, cycling takes priority (since it only fires with multiple windows); single-window apps still minimize normally

## Changes

- `DockClickSupport.swift`: New `.cycleWindows` action case; added `minimizeEnabled`/`cycleWindowsEnabled` parameters to `action()` so the two features compose cleanly
- `DockClickService.swift`: Reads the new preference, handles `.cycleWindows` by calling a new `cycleWindows(pid:windows:)` method
- `Defaults.swift`: New `dockClickCycleWindows` key with `false` default
- `SettingsView.swift`: Toggle in the Dock section
- `Localization.swift` + all language files: UI strings (English placeholders for non-English locales)


vorssaint:
Thanks for this. I tested it locally with multiple windows and it works well. Nice idea to make it compose with the minimize option. Merging now and it will ship with the next update along with a small tweak so the rotation always visits every window.

PR #190, opened 2026-07-08
Add dock click cycle windows toggle to menu panel

CN120:
## Motivation

The dock click window cycling feature (from #182) only had a toggle in the Settings view. Other utilities in this app also appear in the menu bar panel for quick access. This adds the cycle windows toggle there too, so users can enable/disable it without opening Settings.

## Approach

Follows the exact same pattern as the existing `dockClick` (minimize) panel item:

- Adds a new `dockClickCycle` case to `ControlPanelItem`
- Categorized under `.windows` alongside dock preview, dock click minimize, etc.
- Gets its own visibility preference (`panelControlDockClickCycle`) so users can hide it from the panel if they want
- Uses `"dock.rectangle"` SF Symbol to differentiate from the minimize toggle's `"dock.arrow.down.rectangle"`
- Requests Accessibility permission on enable, same as the minimize toggle


vorssaint:

PR #317, opened 2026-07-20
Screenshot: optional automatic copy to the clipboard

kingstyles:
## Behavior before

Every capture has to be routed by hand. The floating preview appears and the image only reaches the clipboard once **Copy** is clicked. With `screenshotOpenEditorDirectly` on there is no preview at all, so the fastest path to annotating is also the slowest path to pasting — the two preferences work against each other.

For the common case of "grab this, paste it into a chat", that is one deliberate click on top of every single capture.

## Behavior after

A new **Copy to the clipboard automatically** toggle in Settings → Screenshot, **off by default**, so nothing changes for anyone who does not turn it on.

When it is on, the capture is placed on the pasteboard as soon as the selection finishes, before the preview or the editor is shown. `Cmd`+`V` works immediately. The preview still appears with all of its buttons; the editor still opens when the direct-edit preference is on. The two preferences are fully orthogonal:

| Auto-copy | Open editor directly | Result |
|---|---|---|
| off | off | Unchanged: preview with Copy / Save / Delete / Edit |
| on | off | On the clipboard immediately, preview still appears |
| on | on | On the clipboard immediately, editor opens for annotation |
| off | on | Unchanged: straight to the editor |

Saving to a file is untouched and stays an explicit choice from the preview or the editor.

## Implementation

Four files, 53 added lines, nothing removed:

- `Core/Defaults.swift` — `screenshotCopyToClipboard`, registered `false`.
- `Services/QuickTools/ScreenshotService.swift` — `route(_:)` calls a new private `autoCopy(_:)` before deciding between preview and editor, so the copy also covers the direct-to-editor path. It reuses the existing `flatten(_:)` pipeline, so the 1x downscale preference applies exactly as it does to the preview's Copy button, and the existing `ScreenshotEditorController.copyImage(_:)`, so the pasteboard gets the same PNG + TIFF pair as every other copy in the app.
- `UI/Settings/ScreenshotSettings.swift` — the toggle and its caption, in the output section above the save folder.
- `Core/ScreenshotStrings.swift` — `autoCopyToggle` and `autoCopyCaption` in all 13 languages.

No new dependencies, no new types, no changes to existing call sites.

### Two decisions worth a second opinion

**No HUD on success.** `copyDirect(_:)` shows the "Screenshot copied" HUD because the preview closes right after it, leaving nothing on screen. With auto-copy a surface is always about to appear, so a HUD would land on top of the preview and repeat what it already says. A failure still beeps, since an empty clipboard would otherwise only be discovered at the paste. Happy to add the HUD if you would rather have the confirmation.

**Annotations do not re-copy.** If auto-copy puts the raw capture on the clipboard and the capture is then annotated in the editor, the clipboard still holds the un-annotated version until Copy is pressed in the editor. That is a real trap for anyone who edits, but fixing it means refreshing the pasteboard on editor close, which changes editor behavior and reads as a separate topic. Left out to keep this to one change; glad to open a follow-up if you want it.

## Localization

`autoCopyToggle` and `autoCopyCaption` are filled in for all 13 languages in `ScreenshotFeatureStrings`: en, pt-BR, tr, ru, es, de, fr, it, ja, ko, zh-Hans, zh-TW, zh-HK. Terminology follows what the file and `Core/Localizations/` already use, notably 剪贴板 for zh-Hans and 剪貼板 / 拷貝 for zh-TW and zh-HK rather than 剪貼簿 / 複製, and zh-HK mirrors zh-TW here the same way the neighboring `openEditor` strings do.

These are machine-assisted translations reviewed against the surrounding entries for tone and consistency, not native review. Please correct anything that reads wrong — I would rather the strings be right than mine.

## Testing

- `./build.sh` completes with no warnings, before and after the change.
- Built against the Command Line Tools, Swift 6.3.3, on macOS 26.5.2, Apple M4.

I have not exercised an end-to-end capture on the built bundle: it is ad-hoc signed and would need its own Screen Recording grant alongside the installed copy, which I did not want to do on this machine. The change is small and reuses existing, already-tested copy and flatten paths, but the actual capture → clipboard round trip is unverified by me and worth a manual check before merging.

## Notes

No existing issue covers this — I searched the tracker for screenshot and clipboard first and only found closed clipboard-history bugs. Happy to open a feature request to discuss it there instead if you would rather talk it through before looking at code.

Contributed under GPL-3.0-or-later per `CONTRIBUTING.md`.

### DIFF
--- a/Sources/Vorssaint/Core/Defaults.swift
+++ b/Sources/Vorssaint/Core/Defaults.swift
@@ -312,6 +312,7 @@ enum DefaultsKey {
     static let screenshotBackdropStyle = "screenshotBackdropStyle"
     static let screenshotBackdropPresets = "screenshotBackdropPresets"
     static let screenshotOpenEditorDirectly = "screenshotOpenEditorDirectly"
+    static let screenshotCopyToClipboard = "screenshotCopyToClipboard"
     static let panelUtilityScreenshot = "panelUtilityScreenshot"
 
     // Window Layout — snapping, global shortcuts and optional pointer gestures.
@@ -787,6 +788,7 @@ enum Defaults {
         DefaultsKey.screenshotBackdropStyle: "",
         DefaultsKey.screenshotBackdropPresets: "[]",
         DefaultsKey.screenshotOpenEditorDirectly: false,
+        DefaultsKey.screenshotCopyToClipboard: false,
         DefaultsKey.panelUtilityScreenshot: true,
         DefaultsKey.windowLayoutShortcutsEnabled: false,
         DefaultsKey.windowGestureEnabled: false,
--- a/Sources/Vorssaint/Core/ScreenshotStrings.swift
+++ b/Sources/Vorssaint/Core/ScreenshotStrings.swift
@@ -76,6 +76,8 @@ struct ScreenshotFeatureStrings {
     let backdropCustomLabel: String
     let openEditorToggle: String
     let openEditorCaption: String
+    let autoCopyToggle: String
+    let autoCopyCaption: String
     let hintLoupe: String
 }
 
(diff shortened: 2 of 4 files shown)


403Denied:
I did something similar in #280 but took a different approach, a default action setting is added to Ask each time, save, save & copy, copy, or edit; if something other than ask each time is selected, the popup gets moved to the corner of the screen, appears for a shorter time, and disables buttons whose actions have already been taken by the default action.

PR #336, opened 2026-07-21
Speed up App Switcher presentation

medeiroshudson:
## Summary

- keep an immutable App Switcher window catalog warm in the background and serve shortcut sessions from its prepared snapshot
- warm the panel and SwiftUI layout before installing the event tap, anchor the 100 ms appearance threshold to the physical key event, and reuse snapshot WindowServer IDs for previews
- move preview warming and close verification Accessibility enumeration to serial worker queues, with generation and per-window token checks that discard stale results
- add presentation signposts, deterministic regression coverage, and a changelog entry

## Root cause

Opening an App Switcher session synchronously enumerated WindowServer and Accessibility state on the main thread before scheduling the visual delay. Accessibility calls can wait on every open application, so that discovery time was added ahead of the intended 100 ms threshold. The first panel layout, preview warming, and post-close verification also left avoidable work on the main thread.

## Impact

The switcher can now present from a ready snapshot without waiting for applications to answer Accessibility queries. Quick shortcut flicks still commit without showing UI, while normal presentation, preview warming, close verification, simple mode, and existing selection behavior are preserved. Background refreshes are serialized, debounced, and guarded against stale lifecycle results.

## Verification

- `./build.sh --test` (`TESTS OK`, 3256 checks)
- `./build.sh` (release bundle assembled successfully)
- `./build/Vorssaint --selftest` (`SELFTEST OK`)
- manually verified App Switcher presentation and close behavior on macOS


vorssaint:
Thanks for the work on this and for investigating the presentation latency.

We ended up addressing this on main with a dedicated tap thread and bounded background queries, so the switcher no longer waits on slow applications. Keeping permanent observers across all system windows introduces too much idle background work for our architecture.

I am closing this PR since the underlying latency issue is already resolved on main. Thanks again for contributing.

PR #360, opened 2026-07-24
feat(menu-bar-organizer): add hardened menu bar organizer MVP

ruvelro:
## Summary

- add an opt-in menu bar organizer with visible, hidden and always-hidden sections
- arrange items from Settings with drag and drop and undo
- reveal hidden items in the menu bar or a secondary bar for notch and overflow cases
- integrate the MVP with the Feature Hub, permissions, settings backup/defaults and app lifecycle
- require Accessibility only for automatic interaction; Screen Recording is not used
- support macOS 14 through macOS 26 and fail closed with an explicit compatibility notice on macOS 27 and later

## MVP scope

This PR intentionally limits the first release to:

- visible, hidden and always-hidden sections
- divider controls and native manual Command-drag fallback
- Settings arrangement with undo
- secondary-bar presentation for notch and overflow
- conflict detection for other running menu bar managers

Presets, automation, groups, spacers, styling, search, hotkeys and additional reveal triggers are deferred until the cross-version substrate has been proven.

## macOS 26 hardening

- resolve Control Center-hosted items through source-PID/AX correlation
- keep unresolved or host-only identities provisional rather than persisting them as unrelated Control Center items
- send synthetic drag and click events to the process that owns the window under the pointer; on macOS 26 that is commonly Control Center rather than the logical source app
- serialize moves, gate them while a relevant menu is open, use a fresh snapshot for the bounded retry, and verify the same window ID reached the requested section and relative position
- filter WindowServer menu-bar furniture and degrade safely when private WindowServer helpers or permissions are unavailable

## macOS 27 compatibility

macOS 27 replaces the per-item WindowServer model used by this MVP with a MenuBarAgent-hosted model. The maintainer confirmed that enabling the current implementation crashes and that the organizer does not work there.

Until a dedicated backend is implemented:

- installation, presets, launch sync and service startup are gated on macOS 27 and later
- preferences carried over from an older macOS remain inert and are preserved for a future update
- the Feature Hub shows a localized compatibility notice and does not expose the organizer Settings destination
- other Vorssaint features remain unaffected

Ice currently reports the same macOS 27 breakage. Thaw has an experimental macOS 27 backend based on Accessibility enumeration, MenuBarAgent preferred positions and a separate private visibility-restriction mechanism. That research is useful for a follow-up, but is intentionally outside this MVP.

## Attribution

- original implementation and PR: @ruvelro
- MVP scope trim, identity work, movement hardening and documentation: @frogkingg / KK, integrated from #627 with the original commits and authorship preserved
- macOS 27 safety gate: @frogkingg / KK, integrated from [ruvelro/vorssaint-utils#1](https://github.com/ruvelro/vorssaint-utils/pull/1) with the original commit and authorship preserved
- Tahoe window-host targeting and localized compatibility UX: @ruvelro
- related feature request: #481

## Validation

- `./build.sh --test` — `TESTS OK (7739 checks)`
- `./build.sh` — production bundle assembled successfully
- `git diff --check`
- merged the current `origin/main`
- maintainer validation: LGTM on macOS 26 Tahoe

## Compatibility status

The MVP is ready for macOS 14 through macOS 26. macOS 27 support is intentionally deferred to a dedicated backend rather than attempting to reuse unsafe WindowServer assumptions.

Refs #481


frogkingg:
I would like to help harden this PR rather than start a competing implementation.

After comparing the current branch with the macOS 26 menu-bar hosting model and Thaw's current implementation, I found three areas that look worth resolving before adding more product surface:

1. `MenuBarWindowProvider` currently derives identity from the CG window owner. On macOS 26, hosted third-party status-item windows are commonly owned by Control Center, so unrelated items can collapse to `com.apple.controlcenter`/generic titles. A source-PID/AX identity pass, with unresolved items treated as provisional and non-movable, would make moves and persisted references safer.
2. The PR description notes that the Command-drag path has not yet been exercised with real Accessibility grants. The current five blind retries would benefit from fresh-frame lookup, a serialized move, menu-open gating, one bounded retry, and post-move identity/section verification.
3. The branch now covers 31 files and 4k+ added lines (presets, automation, groups, spacers, styling, search, and shortcuts). I suggest landing a smaller first slice: visible/hidden/always-hidden sections, Settings drag arrangement, and a secondary bar for notch/overflow. The remaining features can follow once the cross-version substrate is proven.

This also lines up with the user request in #481. I can prepare focused, cherry-pickable commits against `ruvelro:codex/menu-bar-organizer` for the MVP trim, macOS 14/15 vs 26+ identity adapter, move/recovery hardening, and pure regression tests. I will preserve Vorssaint's macOS 14+ target and avoid adding Screen Recording to the MVP.

Would that collaboration/scope direction work for you and the maintainer?

Update from a macOS 26.6 read-only probe: the private WindowServer menu-bar list resolved successfully, and all 19 real status-item windows on the test Mac were owned by Control Center. With Accessibility deliberately absent, every hosted item stayed provisional/non-movable and the WindowServer Menubar furniture was filtered out. Enabling the feature in a Developer build added only the app's own status item (19 -> 20), created no organizer dividers, and logged no faults, confirming the fail-closed path.

The local hardening branch now has 7050 passing checks, a successful full developer build/signature verification, and SELFTEST OK. It remains unpushed while I wait for agreement on the collaboration/scope direction; the remaining merge gate is a real permission-granted Command-drag pass.

frogkingg:
Since my [collaboration offer](https://github.com/vorssaint/vorssaint-utils/pull/360#issuecomment-5249904985) on Aug 11 has not received a response, I have opened a standalone hardened MVP PR: https://github.com/vorssaint/vorssaint-utils/pull/627

It keeps this PR's full commit history and attribution, and links #481. If you prefer to incorporate the hardening here instead, I'm happy to close mine and make the commits available for cherry-picking.

@vorssaint flagging for visibility.

ruvelro:
@frogkingg thank you for the hardening work and for offering it as a collaboration rather than a competing implementation.

I have incorporated #627 into this branch by fast-forward, so your original commits, authorship and co-author trailers are preserved unchanged:

- `e386e65` — MVP hardening
- `3872fbf` — scope and permission documentation
- `385b885` — Settings routing fix

I then merged the current `main` and added a separate follow-up commit, `14231bc`, for the macOS 26 window-host behavior: drag/click events now target the owning window process (normally Control Center for hosted items), generic Control Center AX matches remain provisional, and post-move verification follows the same window ID.

Current validation is 7699 passing checks, a successful production build and `SELFTEST OK`. The PR description now reflects the agreed MVP and credits.

The remaining merge gate is the real Accessibility-granted Command-drag pass on macOS 26. If you can run that pass against this branch, your result would be especially useful. Once you are satisfied that the integration is represented correctly here, please close #627 in favor of #360 so discussion and review continue in one place.

frogkingg:
@vorssaint, when this PR is ready to merge, could you please use "Create a merge commit" to preserve the individual commit history and authorship (including KK's hardening commits)? Squashing would collapse them into a single commit. Thanks!

ruvelro:
CI is fully green on head `14231bc`: both **Swift 6.0.3 compatibility** and **Build & selftest** completed successfully, including 7699 tests, signed build, selftest, DMG packaging and artifact upload: https://github.com/vorssaint/vorssaint-utils/actions/runs/32118054516

Agreed on using **Create a merge commit** rather than squash when this is ready, so KK's individual commits and authorship remain visible in the repository history.

The only remaining gate is the real Accessibility-granted Command-drag result on macOS 26.

vorssaint:
Thanks for all of that, guys! LGTM on macOS 26 Tahoe 🙌

Could we just restrict this feature to macOS 26 and earlier?

I tested it on macOS 27 Golden Gate, and it crashes when trying to enable the feature. Even if we fix the crash, the current implementation doesn’t work on macOS 27 at all, so I think it’d be better to prevent it from being installed/enabled there for now. If you guys can implement this, that would be perfect. It would help me a lot!

frogkingg:
Implemented the macOS 27 restriction requested by the maintainer on top of the current `codex/menu-bar-organizer` branch: the organizer is now gated to macOS 26 and earlier (no install, no enable, and the service stays inert even if defaults were carried over from an older macOS). PR: https://github.com/ruvelro/vorssaint-utils/pull/1 — verified locally with 7702 tests, a full release build and SELFTEST OK.

ruvelro:
@vorssaint thanks for the macOS 26 and macOS 27 validation.

The macOS 27 safety gate from @frogkingg has now been merged through [ruvelro/vorssaint-utils#1](https://github.com/ruvelro/vorssaint-utils/pull/1) with KK's original commit `53c0bef` and authorship preserved. A separate follow-up commit, `90c8ea6`, adds a localized “not yet compatible with macOS 27” notice, keeps inherited configuration intact but inert, removes the unsupported Settings/Command Bar destination, and excludes the organizer from active/supported counts on macOS 27.

Validation on the combined head:

- `./build.sh --test`: `TESTS OK (7739 checks)`
- `./build.sh`: release bundle assembled successfully
- `git diff --check`
- [GitHub CI run 32182746033](https://github.com/vorssaint/vorssaint-utils/actions/runs/32182746033): **success**
  - Swift 6.0.3 compatibility: tests, build and selftest passed
  - Build & selftest: tests, signed build, selftest, DMG packaging and artifact upload passed

I also checked the current state of similar tools. [Ice #954](https://github.com/jordanbaird/Ice/issues/954) reports the same Golden Gate breakage without a linked fix. [Thaw #687](https://github.com/thaw-app/Thaw/issues/687) confirms that macOS 27 moved status items into a new MenuBarAgent-hosted model and now has an experimental backend based on per-app AX enumeration, MenuBarAgent preferred positions, and a separate private visibility-restriction mechanism.

That work is substantial enough to stay outside this MVP. The single follow-up tracker is now #698; it records the evidence, a staged implementation path, and the safety/real-device acceptance criteria. #360 remains the canonical macOS 14–26 implementation. Please continue to use **Create a merge commit** when merging #360 so KK's individual commits remain visible.

ruvelro:
Updated the branch to resolve the conflicts with the current `main` (165 commits had landed since the last sync). The 74 conflicted files were almost entirely older drafts of unrelated features that later landed on `main` in evolved form; they resolve to `main`'s side, while the organizer wiring (feature catalog, availability gating, settings navigation, docs and the organizer test suite) merges both sides. The merge also surfaced and fixes three leftovers the auto-merge missed: duplicated display-recovery helpers, a duplicated Super key test block, and a stale desktop-move draft that broke the production target.

Three small hardening fixes from a fresh review are included on top:

- the synthetic mouse-up is now guaranteed once a Command-drag or click has posted its mouse-down, so a thrown mid-drag step can no longer leave the target process with a phantom button-down
- conflicting menu bar managers launched mid-session are now detected and stop the organizer instead of fighting it
- the divider `windowID` guards against non-positive `windowNumber` values, removing a crash path

Validation: `./build.sh --test` → TESTS OK (8283 checks), `./build.sh` production bundle, `./build/Vorssaint --selftest` → SELFTEST OK. As agreed above, please use **Create a merge commit** when merging so KK's commits keep their authorship.

tolgaerdonmez:
When will this be in main?

PathGao:
This has gone `CONFLICTING` against `main` @ `9d9c6563`, and it is the top of the queue in #837 — a rebase here unblocks #700 and #824 too, since both carry ~378 of this branch's commits.

The good news is that none of it is real. `git merge-tree origin/main <head>` reports 8 hunks across 5 files, and **every one of them is `bluetoothSleep` (from #861) landing in the same list position as `menuBarOrganizer`**:

| File | Hunks |
|---|---|
| `Core/FeatureCatalog.swift` | 3 |
| `UI/Settings/FeatureHubSettings.swift` | 2 |
| `App/AppDelegate.swift` | 1 |
| `App/FeatureRuntime.swift` | 1 |
| `Tests/MetricsTests.swift` | 1 |

Sample, from `FeatureCatalog.swift`:

```
<<<<<<< origin/main
    case keepAwake, brightness, extraBrightness, bluetoothSleep
=======
    case keepAwake, brightness, extraBrightness
    // Menu bar
    case menuBarOrganizer
>>>>>>> this branch
```

The resolution is "keep both cases" in all 8 places. Nothing about the MVP itself is affected — no conflict touches `Services/MenuBarOrganizer/`.

Could you rebase onto current `main`? You already have the maintainer's LGTM and the macOS 26-and-earlier restriction is in (`53c0bef1`, `90c8ea65`), so the conflict is the only thing standing between this and a merge. Worth doing soon rather than later: every feature PR that lands adds another case to those same five lists.

PathGao:
Ref #481, #615

PathGao:
Two things the conflict analysis above does not cover, both outside `Services/MenuBarOrganizer/`.

**`build.sh` now defines `codesign_with_timestamp_retry` twice, and the second copy is weaker.** The merge base already had it at `build.sh:53`, where it calls `/usr/bin/codesign`. This branch adds a second definition at `build.sh:410` that calls bare `codesign`. Bash takes the last definition, so `codesign_app` (`:424`) and `codesign_fan_helper` (`:436`) sign through the branch's copy and lose the absolute path. `build.sh` is not in the conflict list, so this merges clean and nothing flags it. Delete the block at `:410-421`; `:53` already provides the function at that point in the script.

**`MenuBarWindowServerBridge` is a second window-server bridge.** `Services/Switcher/SpaceWindowBridge.swift:19-30` already resolves `CGSMainConnectionID` and caches the connection id once, with a `symbol(_:)` helper over `RTLD_DEFAULT`. `MenuBarWindowServerBridge.swift:43-53` opens its own `dlopen` handle for CoreGraphics and adds a second `symbol(_:)` at `:90-95`, then re-invokes `mainConnection()` on every call in `frame(for:)` (`:76`) and `level(for:)` (`:86`) rather than caching it. Adding the four menu-bar symbols to `SpaceWindowBridge` and reusing its cached `connection` removes the duplicate resolution path and the per-call connection lookup, which matters here because `refreshNow()` runs these once per item on the refresh timer. Also `deinit { dlclose(handle) }` at `:55-57` never runs on a `static let shared`, so it is dead either way.

Nit: `SettingsDirectory.swift:53` loses a space on the existing `.monitor` item's `keywords:` alignment — unrelated to this feature.

ruvelro:
Addressed the review feedback in 5b77d5f:

- merged the current `main` and resolved the organizer/Bluetooth Sleep conflicts without dropping either feature
- consolidated the private WindowServer symbol loading and connection cache in `SpaceWindowBridge`
- removed the duplicate, weaker `codesign_with_timestamp_retry` implementation
- fixed the Settings spacing nit

Validation is green: 8,609 local checks, release build, self-test, and both GitHub CI jobs. The branch is now mergeable. Please re-review and clear the `incomplete` label if everything looks good.

ruvelro:
Follow-up from the #824 review is now fixed at 7e40e14. FeatureRuntime.replaceAvailable writes unsupported availability to false and then continues before invoking the feature binding, so applying a preset on macOS 27 cannot instantiate MenuBarOrganizerService. I also restored the existing FeatureCatalog indentation. Local validation: 8,609 checks, release build, self-test, and diff check pass; Both GitHub CI jobs are green.

PathGao:
`replaceAvailable` is fixed and reads correctly at 7e40e14. Two sibling call sites still run the binding for an unsupported feature, so the "applying anything on macOS 27 cannot instantiate `MenuBarOrganizerService`" invariant isn't held yet.

**`AppDelegate.swift:237`** gates teardown on availability alone:

```swift
if AppFeature.menuBarOrganizer.isAvailable {
    MenuBarOrganizerService.shared.stop()
}
```

With an availability key carried over from macOS 26, `isAvailable` is `true` on macOS 27, so whenever that teardown path runs it instantiates `.shared` — the exact instantiation the gate exists to prevent. Needs `&& AppFeature.menuBarOrganizer.isSupportedOnCurrentSystem`.

**`FeatureRuntime.swift:65` and `:110`** miss the same case on the *off* path. `setAvailable`'s early return only covers `available == true`:

```swift
if available, !feature.isSupportedOnCurrentSystem { return }
```

so turning a carried-over feature off falls through to `Self.bindings[feature]?()` → `syncWithPreferences()`. `setAllAvailable(false)` is the same shape: its `where` clause guards only the `available` side. `replaceAvailable` is now the only one of the four paths that gets this right.

Root cause is that `isAvailable && isSupportedOnCurrentSystem` is spelled out at nine places in this diff and two of them are missing the second half — every new reader is another chance to forget one. Making `AppFeature.isAvailable` itself return `false` when `!isSupportedOnCurrentSystem` covers all readers including future ones, and the writers keep touching `availabilityKey` directly so carried-over prefs stay preserved. I can't see every `isAvailable` call site from the diff, so confirm none of them need the raw pref value before doing that.

Not verified: `build.sh` and `MenuBarWindowServerBridge.swift` both fall past the truncation point of the diff I can read, so I have not confirmed the duplicate `codesign_with_timestamp_retry` removal or the `SpaceWindowBridge` consolidation. That's "haven't seen it", not "looks wrong".

**Impact:** macOS 27 users who enabled the organizer on macOS 26 and then upgraded. The teardown path, or toggling the feature off in the hub, instantiates the service the compatibility gate is meant to keep dormant.
**Merge:** needs the two remaining `isSupportedOnCurrentSystem` guards first.

ruvelro:
Addressed the follow-up review in `2b46ba78`:

- the AppDelegate teardown path now checks `isSupportedOnCurrentSystem` before touching `MenuBarOrganizerService.shared`
- `setAvailable(false)` can clear a carried-over availability key without invoking an unsupported binding
- `setAllAvailable(false)` does the same for every unsupported feature

The same commit has been propagated to the stacked #824 branch. Local validation: 8,609 checks, release build, self-test, and diff check all pass. Please re-review and clear `incomplete` if this closes the macOS 27 upgrade path.

PathGao:
All three guards read correctly at `2b46ba78`.

- `AppDelegate` teardown now checks `isSupportedOnCurrentSystem` before touching `.shared`.
- `setAvailable` writes the key and returns before `bindings[feature]?()`, so the off path no longer falls through for a carried-over feature.
- `setAllAvailable` does the same per feature and still sets `changed`, so the hub refreshes.

With `mayFlip` refusing the on direction, and `replaceAvailable`, `syncAtLaunch` and `sync` already filtered, I can't find a path that reaches `MenuBarOrganizerService.shared` on macOS 27 in the part of the diff I can read.

Not verified: the diff I get is truncated at roughly 40%, so `build.sh`, `MenuBarWindowServerBridge.swift`/`SpaceWindowBridge.swift` and the Settings UI all fall past the cut. The duplicate `codesign_with_timestamp_retry` removal and the WindowServer consolidation are still "haven't seen it" on my side, not "looks wrong". Check states also come through empty for me on this head, so CI green on `2b46ba78` is your report, not something I confirmed.

**Impact:** none outstanding from this review.
**Merge:** ship it.

ref #481
ref #615

ruvelro:
Follow-up: current `main` advanced after the review fix was pushed, so I merged it at `18383b9b` and resolved the two overlapping gates. The combined runtime now requires both OS and hardware support for installs, still permits uninstalling carried-over preferences, and never invokes an OS-unsupported binding. GitHub now reports the PR mergeable. Validation on the current head: 8,709 checks, release build, self-test, and `origin/main...HEAD` diff check pass.

PathGao:
The merge at `18383b9b` re-resolves the two gates my last two reviews were about, and it falls past the truncation point of the diff I can read, so `ready to merge` is still anchored to `2b46ba78`.

Both earlier misses in this PR were half-guards on the write side — `setAvailable`'s early return covering only `available == true`, and `setAllAvailable`'s `where` clause doing the same. Folding OS support and hardware support into one condition during a conflict resolution lands in exactly that spot. Worth re-reading five paths on the merged head: `mayFlip`, `setAvailable`, `replaceAvailable`, `setAllAvailable`, and the `AppDelegate` teardown. `!available` has to stay unconditional on the off side, and `isSupportedOnCurrentSystem` has to gate the binding call, not only the key write.

That's a re-check request, not a finding — I have no view of the merged head. If those five read the way they did at `2b46ba78`, nothing changes.

**Impact:** none identified; the merge commit itself is unreviewed on my side.
**Merge:** ship it.

ref #481
ref #615

PathGao:
Closing this, with the maintainer's agreement. It is not a rejection of the work, so let me set out the reasoning.

The macOS 27 gate you added does exactly what was asked for in https://github.com/vorssaintapp/vorssaint-utils/pull/360#issuecomment-5327155215, and it works. That is also the problem: with the gate in place, merging this lands ~3,400 lines of a feature that is switched off on the current system, and the notice in thirteen languages telling people so.

The blocker is the platform, not the branch. `CGSGetProcessMenuBarWindowList` no longer exposes individual status item windows on macOS 27 — only per-app `AXExtrasMenuBar` enumeration is left — which is what #698 tracks, and which is why the feature crashed on macOS 27 Golden Gate before the gate went in.

**Keep the branch, please do not delete it.** When #698 has a backend that holds up on macOS 27, this is the front end for it, and the review rounds it has already been through still count. #824 and #1276 are closed alongside for the same reason.

Thanks to everyone who worked on it — this got a LGTM on macOS 26, which almost nothing in the queue has.

ches:
Thanks @PathGao for your continuing effort to coordinate.

Trying to follow along, I am missing a takeaway that's perhaps obvious somewhere:

What happens to work for 14-26?

I understand deferring macOS 27 until APIs are settled as GM.

Is the conclusion that gating 27, with affordances for forward-compatibility, amounts to so much code that the whole feature sits on the shelf for 14-26 until designs can confidently be made for both?

If so, I can understand that decision—this isn't a complaint, just a wish for it to be plainly stated. (Perhaps it should be noted on #481, as a top-level synchronization point for people following the feature request in general).

PathGao:
Yes. That's the decision, and you've put it more plainly than I did.

The 14–26 implementation is done and works — it isn't blocked on itself. What it can't do is ship alone. Merging it now means every release from here carries a large feature that is dark on macOS 27, plus the notice in thirteen languages explaining why, and 27 is the direction the installed base moves in. That cost is paid continuously; the benefit is capped at the versions that are on their way out.

So it waits on the branch until #698 has a macOS 27 backend, then it ships for all versions at once. Nothing is thrown away: the review rounds still count and this is the front end when the backend exists.

You're right that #481 is where people are watching. I'll write this up there.

PR #361, opened 2026-07-25
Adjust legacyBlockAttachmentNudge for macOS < 27

wzxu:
Addresses the over-correction of offset as mentioned in https://github.com/vorssaint/vorssaint-utils/issues/253#issuecomment-5013731850

Now, with the nudge becoming so small I'm not even sure whether or not there's really differences between macOS 27 and below, but I don't have macOS 27 to check, and at least on Tahoe as well as even older version like Sonoma, this change makes the icons perfectly center aligned.

### DIFF
--- a/Sources/Vorssaint/App/MenuBarRenderer.swift
+++ b/Sources/Vorssaint/App/MenuBarRenderer.swift
@@ -802,11 +802,11 @@ enum MenuBarRenderer {
     }
 
     /// macOS 26 lays out button titles dominated by tall attachments about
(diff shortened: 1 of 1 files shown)


wzxu:
### Before:
image
image

### After:
Tahoe
file-3d3c9b77335cd79b46f3efa0dca9aef3
Sonoma
sonoma

vorssaint:
Thanks for the fix. I measured your screenshots against the system highlight and the blocks land dead center now on both versions, so I merged it.

Issue #171, opened 2026-07-05
[Feature]: Multiple Feature request


dionysusv:
### Problem or Motivation

Vorssaint > Controls > Windows > Quit on Close does not have an "exclude" option like Supercharge does, or the option to customize the quit trigger like left click vs right click on red x to quit. 

Vorssaint > Utilities > Homebrew > Update does not work as all say failed to update from untrusted tap, no way to add the tap to trust list. 

Vorssaint > System > currently reads:
CPU Temp
GPU Temp
Battery Temp
CPU Usage % > click to show processes using most cpu power
GPU Usage % > click to show processes using most gpu  power
Battery Remaining %  > click to show processes using most battery power
Memory Pressure > click to show processes using most memory power
runtime

I would like to see more features from IstatsMenu ported into the System stats. Heres some screenshots of the information I would be looking for:

<img width="578" height="691" alt="Image" src="https://github.com/user-attachments/assets/427ffb56-d6e8-4e78-b665-5ca3a454cc15" />
<img width="576" height="663" alt="Image" src="https://github.com/user-attachments/assets/a6d0edcc-551d-4831-b3c4-cbafd4aa54a3" />
<img width="582" height="653" alt="Image" src="https://github.com/user-attachments/assets/42f33c9a-64fa-45b1-99c0-76b8fd544900" />
<img width="599" height="658" alt="Image" src="https://github.com/user-attachments/assets/1e56ee54-6dd4-4998-aeec-824cf5651762" />
<img width="329" height="663" alt="Image" src="https://github.com/user-attachments/assets/baabfc53-a625-45c7-be52-e8b21d004dec" />
<img width="319" height="344" alt="Image" src="https://github.com/user-attachments/assets/2ea7a575-1ec1-41f8-a7df-7d7ab5558298" />
<img width="580" height="914" alt="Image" src="https://github.com/user-attachments/assets/7ccb55ba-3a0c-4eb2-ba94-2e67810ad59e" />
<img width="605" height="877" alt="Image" src="https://github.com/user-attachments/assets/b46d3c3d-42e8-4221-bf22-15196ca7e5c9" />
<img width="549" height="753" alt="Image" src="https://github.com/user-attachments/assets/db70070f-f0bd-4957-8418-e53418523800" />

[...]

PR #365, opened 2026-07-26
System panel: hardware sensors, GPU/memory/storage detail, and an iStat-style combined dashboard (#171)

dionysusv:
Addresses #171 — expands the System monitor with iStat Menus-style hardware sensors and a combined dashboard.

### Hardware sensors (the #171 ask)
- Named temperature list (CPU P/E cores, GPU, battery, SSD, Wi-Fi, airflow) via SMC, curated + plausibility-filtered
- Per-fan RPM with iStat-style min–max % scaling (an idle Apple-silicon fan reads ~1%, not ~30%)
- Total system power (SMC `PSTR`)
- CPU E/P-cluster and GPU frequency via IOReport DVFS residency
- GPU utilization % and GPU memory % (in-use ÷ allocated)

### System dashboard
- Combined view folding Network/Disk/Power into one System tab (toggleable)
- Per-core CPU speedometer gauges with efficiency/performance split
- Memory breakdown (app/wired/compressed/free) with a segmented ring
- Storage Used/Purgeable/Free breakdown; battery + health donuts; uptime on the storage line

### Settings
- Per-reading toggles for every sensor and block; Disk/Network toggles in the System group

### Notes
- Built against current `main`; `./build.sh --test` passes (3656 checks); verified on an M4 Pro / macOS 26.
- Frequency uses IOReport via `dlopen` of `/usr/lib/libIOReport.dylib` (isolated in `FrequencySampler`, degrades gracefully).
- FPS is intentionally omitted — the accelerator's PerformanceStatistics exposes no frame counter on Apple silicon.
- Broad by design; happy to split into focused PRs (e.g. sensors-only) if you'd prefer.


PathGao:
Correcting the label on this one: it was marked `blocked`, and nothing is blocking it — the branch cannot be merged in its current form for a reason only you can clear.

```
$ git merge-base origin/main <this branch>
(nothing)
$ git merge-tree origin/main <this branch>
fatal: refusing to merge unrelated histories
```

There is no common ancestor with `main` at all, so there is no base to rebase onto. That is different from a conflict, and it is why the earlier advice in #837 — "ask for a rebase" — was wrong.

The work itself is not the problem. This is not a duplicate of anything on `main`, and the sensor / GPU / storage panel is wanted. What it needs is a fresh branch cut from current `origin/main` with the changes re-applied on top, at which point it becomes reviewable.

Re-labelling `blocked` → `incomplete`, which is the label for "not mergeable as it stands, and here is the gap".

PathGao:
Ref #171.

For context on where that request stands: #171 was closed as not planned because it bundled several unrelated asks into a single issue, not because the ideas were turned down. The closing comment there says some of them are genuinely interesting and asks for one request per issue, and notes that two of them had already shipped.

This PR implements the System panel slice of that bundle. Since the issue it points at is closed for a process reason, there is no live, reviewable request behind the PR right now. Opening a focused issue for the System panel readouts would give this work something to be reviewed against.

PathGao:
Closing this so it can come back as a fresh pull request off current `main`. This is about the shape of the branch, not the work in it.

As it stands the branch shares no common base with current `main`, so there is nothing to rebase onto and the diff cannot be read as your change. Your commits carry over on their own, though — `cherry-pick` copies by patch and does not need a shared base:

```bash
git remote add upstream https://github.com/vorssaint/vorssaint-utils.git
git fetch upstream
git log --author="$(git config user.name)" --no-merges --reverse feature/hardware-sensors   # your 21 commits
git checkout -b sensors upstream/main
git cherry-pick <those, in order>
```

**Worth checking before you spend that time:** `main` has moved in exactly this area since 07-25. `TemperatureSensorSelector.swift` now carries per-chip CPU temperature families (M1–M5), adjustable fan control shipped, Memory Used accounting was corrected (#763), swap plus compressed memory and cached files were added to the System panel, and system versus adapter power was split (#585). Several of your readouts now exist upstream in some form. Diffing against that first is cheaper than discovering it during review.

Also #171 is closed as not planned, so there is no live request behind this work right now. A focused issue for the System panel readouts would give it something to be reviewed against.

**And take up your own offer to split.** Sensors-only as the first pull request. The part that folds Network/Disk/Power into a single System tab changes navigation for every user, and that decision should be made separately from the readouts themselves.

The work itself was not the problem — verified on real hardware, `./build.sh --test` green, and you cleaned the research dumps out of the branch yourself. Send it back off current `main` and it gets a full review from a clean base.

PR #368, opened 2026-07-26
Add precise volume roller

ruvelro:
## Summary

Adds an opt-in precise volume roller control to the mixer panel. When enabled, Vorssaint listens for hardware volume up/down media-key events and re-emits them as fine system volume steps, while filtering fast duplicate pulses and noisy direction reversals.

## Details

- Adds a new `PreciseVolumeRollerService` with a focused support helper for deterministic step gating.
- Exposes the toggle in the volume mixer UI with Accessibility guidance and tap failure feedback.
- Wires the service into feature runtime, app shutdown, and uninstall teardown so the event tap does not outlive the app or permission state.
- Updates feature permission/energy metadata so Accessibility is only reported when this mixer option is enabled.
- Adds localized strings for the new control across the existing app languages.
- Adds unit coverage for default opt-in behavior, media key mapping, debounce, and direction reversal confirmation.

## Validation

- `./build.sh --test` → `TESTS OK (3666 checks)`
- `./build.sh` → `✓ Bundle ready: build/stage/Vorssaint.app`


vorssaint:
Thanks for the contribution. I tested it locally and merged it.

PR #374, opened 2026-07-27
Feature: Forward EarPods Media Key to mute MS Teams (Resolves #373)

### LINKED ISSUE #373: Forward EarPods Media Key to mute MS Teams
**The Problem:**
When using wired Apple EarPods, pressing the inline remote button does not mute MS Teams. Vorssaint already intercepts the media key (`NX_KEYTYPE_PLAY`) perfectly – which prevents Apple Music from launching – but it doesn't pass the mute intent to Teams. Other macOS tools offer this, and it’s a big pain point for meeting users.

**Proposed Solution:**
I’d like to add a small extension to the existing media key / mute logic. 
When the media key is intercepted (and consumed via `return nil`), we simply add an option to post the Teams mute shortcut (`Cmd + Shift + M`) directly to the `com.microsoft.teams` / `com.microsoft.teams2` PID. 

This would just need a new toggle in the UI (e.g., *"Forward inline remote mute to Microsoft Teams"*).

FinkeFlo:
This PR adds the ability to use the inline remote (Play/Pause button) on wired Apple EarPods to mute/unmute Microsoft Teams, addressing #373.

**Changes:**
1. Added `EarPodsTeamsMuteService` that hooks into the hardware events via `cghidEventTap` to intercept `NX_KEYTYPE_PLAY`.
2. Handled the media key by returning `nil` (blocking Apple Music from launching).
3. Added functionality to dispatch `Cmd + Shift + M` to Microsoft Teams if running.
4. Added UI toggle under Music Block settings.
5. All 13 localizations updated to include the new toggle text.


FinkeFlo:
@codex review

FinkeFlo:
@codex review

FinkeFlo:
@codex review

vorssaint:
Thanks for the contribution. I am going to close this without merging.

Vorssaint focuses on native macOS utilities and avoids hardcoded integrations for specific third-party apps. Consuming the global Play/Pause media key also interferes with normal media playback across the system.

Thanks again for opening the PR and exploring the idea.

Issue #303, opened 2026-07-19
Dynamic Media Mini Player with Customizable Layouts


mickolas9:
### Problem or Motivation

Managing media playback across various applications on macOS (like Spotify, Apple Music, or web browsers) currently requires switching back to those specific apps or using the native macOS Control Center, which feels clunky and lacks glanceable information on the desktop. There is no lightweight, native-feeling desktop utility to permanently see what is playing and quickly control it without interrupting my workflow.

### Proposed Solution

A native, lightweight media mini player that works universally with any active macOS media app. To keep it flexible for different desktop setups, it would be incredibly helpful to have a couple of basic layout variations:
- Control Bar Layout: A compact horizontal strip that permanently displays track info and basic playback controls.
- Minimal Art Layout: A clean, square layout showing only the album art. Hovering the cursor over the art would reveal the overlay controls.

### Feature Area

Other

### Mockups / References

N/A

ruvelro:
Hi! Take a look: https://github.com/vorssaint/vorssaint-utils/pull/297 

it works for you?

mickolas9:
> Hi! Take a look: [#297](https://github.com/vorssaint/vorssaint-utils/pull/297)
> 
> it works for you?

To be honest, I am not a radial-menu user.

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

Blocked on PR #380 ("Add Dock Preview media controls"), which is open and currently conflicting with main.

Important caveat rather than a straight dependency: #380 only covers a mini player shown while hovering the Dock, not the persistent desktop player with customisable layouts this issue asks for. So when #380 lands, the remaining scope here should be re-read rather than assumed covered.

#380 also has a merge-order constraint — it touches the same Dock Preview files (`DockPreviewService.swift`, `DockPreviewSupport.swift`, `DockPreviewPanelView.swift`) as #862, and #830 has already merged ahead of both, so it needs a rebase before review.

PR #380, opened 2026-07-28
feat(dock-preview): add Dock Preview media controls

ruvelro:
## Summary

- Adds an opt-in mini-player to Dock Preview for the app that owns the current MediaRemote Now Playing session.
- Shows artwork when available, track title, artist, album, and previous/play-pause/next controls.
- Keeps normal window preview cards visible; media controls supplement a real window-preview session and do not create a separate panel for windowless apps.
- Ships disabled by default for existing and new users.

## Implementation

The player reuses the shared event-driven MediaRemote bridge, so it supports any Now Playing app without AppleScript, bundle-ID allowlists, Automation permission, or a polling timer. Playback commands use the same system media-key path as the radial player. The panel intentionally omits elapsed time and progress because MediaRemote requires timestamp/rate interpolation and a repaint cadence to keep those values truthful.

## Validation

- `./build.sh --test` -> `TESTS OK (8543 checks)`
- `git diff --check` -> clean
- GitHub CI restarted for db5158d

Ref #303


PathGao:
This went `CONFLICTING` when #830 (Dock Preview card and drag rework) landed. Two files:

- `Services/DockPreview/DockPreviewService.swift` — `@Published orientation` vs this branch's `@Published mediaPlayer` at the same point, plus the `DockPreviewSupport.panelSize(...)` call that #830 gave an `orientation:` argument.
- `UI/Switcher/DockPreviewPanelView.swift` — three hunks, all `orientation` vs `mediaPlayer` in the same parameter lists and struct fields.

All add-alongside; nothing here is a design collision with #830.

Could you rebase onto current `main`? Note #862 adds `closeActionTitle` to those same two structs, so whichever lands first, the other needs a second pass.

PathGao:
Before the conflicts: this adds a third mechanism for something the app already answers twice, and that is the thing to settle first.

**"What is playing" already has an answer here.** `RadialNowPlayingService.swift:291` reads Now Playing through the private MediaRemote framework for **any** app, with parsing already factored out into `RadialNowPlayingSupport.snapshot(info:)` (`RadialMenuSupport.swift:305`). `DockMediaPlayerService.swift:57-61` instead hardcodes three bundle ids — `com.spotify.client`, `com.apple.Music`, `com.apple.iTunes` — over AppleScript. Safari, Chrome, Podcasts, TV, IINA, VLC get nothing.

**"Send a transport command" also already has one.** `RadialMenuMediaKey` (`RadialMenuSupport.swift:255-264`) posts NX media keys for play/pause/next/previous, app-agnostically. `DockMediaPlayerService.swift:190-205` sends AppleScript instead. A user with Now Playing in the radial menu and a hover on the Spotify Dock icon is driving two different code paths for the same two buttons.

Reusing those two would delete most of this PR and make it work for every player at the same time.

**A behaviour change the title does not mention.** `beginSession` now prefers the media panel over window thumbnails for Music, Spotify and iTunes. While something is playing, those three apps lose Dock Preview window previews entirely, with no modifier to get them back. The fallback to `beginWindowSession` only happens after the AppleScript snapshot returns nil — one round trip later — so hovering with nothing playing shows a placeholder media panel (title = app name, state `.stopped`) that then resizes into the window panel.

**Default is `true`.** `dockPreviewMediaControls` is registered `true`, so any existing user who already has Dock Preview enabled gets this on upgrade. Combined with `AppleScriptRunner.consentToAutomate(bundleID:)` calling `AEDeterminePermissionToAutomateTarget(..., askUserIfNeeded: true)`, merely hovering the Spotify or Music Dock icon raises a macOS Automation permission dialog for a feature they never asked for. Defaulting a new feature off is the convention here — `dockPreviewEnabled` itself is `false`.

Smaller:

- `startMediaRefreshTimer` fires an AppleScript round trip every second for as long as the panel is visible, plus an extra `scheduleMediaRefresh(after: 0.28)` per button press. Serial queue, so slow targets back up rather than overlap, but there is no coalescing or backoff. (The AppleScript is correctly kept off the main thread.)
- `loadArtwork(from:)` does `Data(contentsOf: url)` on a Spotify artwork URL — a remote https fetch with no timeout and no `URLSession`, blocking the media queue for as long as the network takes.
- `perform(_:for:)` discards the AppleScript result, so a refused or failed transport command is silent.
- `DockPreviewPanelView.swift:42-60` hardwires the pinned panel to `mediaPlayer: nil` and hides the pin button in media mode, so a mini player can never be pinned. Consistent with `DockPreviewService.togglePinned` already guarding `!windows.isEmpty`, but it means the feature never joined the pin mechanism.
- No test coverage added, unlike #862 and #844 in the same area.

On the merge itself: it conflicts with `main` in `DockPreviewService.swift` and `DockPreviewPanelView.swift` where #830 added `orientation`, and textually with #862 at `Defaults.swift:71` and `SettingsView.swift:850`.

PathGao:
Ref #303

PathGao:
Following up with a concrete list, since the two comments above are analysis and this is what would actually move the PR. In this order — the rebase is last on purpose.

### 1. Read "what is playing" from MediaRemote instead of three bundle ids

`RadialNowPlayingService.swift:291` already has `MediaRemoteNowPlayingBridge`, and the parsing is already a pure function: `RadialNowPlayingSupport.snapshot(info:isPlaying:appBundleIdentifier:appPID:)` at `RadialMenuSupport.swift:305`, returning a `RadialNowPlayingSnapshot` with title / artist / album / artwork. `playbackIsActive(remoteIsPlaying:info:)` at `:300` is the "is anything playing" test.

Calling those deletes `DockMediaPlayerService`'s AppleScript metadata path, the `supportedBundleIDs` set at `:57-61`, and `loadArtwork(from:)` — and the feature starts working for Safari, Chrome, Podcasts, TV, IINA, VLC and everything else, which the hardcoded three never will.

It also removes the `AEDeterminePermissionToAutomateTarget` prompt entirely, which is worth having gone on its own.

### 2. Send transport with the media keys, not AppleScript

`RadialMenuMediaKey` at `RadialMenuSupport.swift:253-264` already maps play/pause, previous and next to the aux key codes the physical keys post (16 / 20 / 19). That is app-agnostic and it is what the radial menu's own buttons use.

Replacing `perform(_:for:)` (`DockMediaPlayerService.swift:190-205`) with those drops the last AppleScript call, and with it the silent-failure problem — `perform` currently discards the script result, so a refused command looks identical to a successful one.

### 3. Default the preference off

`DefaultsKey.dockPreviewMediaControls` is registered `true`. `dockPreviewEnabled` itself is `false`, and that is the convention here: a new feature arrives off. As it stands, anyone who already uses Dock Preview gets this on upgrade without asking.

### 4. Don't replace window previews without a way back

`beginSession` currently prefers the media panel for the supported apps whenever something is playing, so those apps lose Dock Preview window thumbnails entirely with no modifier to get them back. And because the fallback to `beginWindowSession` only happens after the snapshot returns nil — one round trip later — hovering with nothing playing shows a placeholder panel (title = app name, state `.stopped`) that then resizes into the window panel.

Two options, either is fine: show the mini player *alongside* the window cards rather than instead of them, or keep the swap but decide it synchronously off the already-cached Now Playing state so there is no placeholder frame. What should not stay is the current shape, where a user who wants the window list for Spotify has no way to ask for it.

### 5. Drop the 1s timer

`startMediaRefreshTimer` fires a round trip every second the panel is visible. MediaRemote publishes change notifications — `registerForNotifications()` in the existing bridge — so the panel can be driven by those instead of by polling. If a timer is still wanted as a backstop, it should be much slower and coalesced.

### 6. Then rebase

Leaving this last because the conflict has grown: `DockPreviewPanelView.swift` is at 9 hunks now and `DockPreviewService.swift` at 2, all of them `orientation` / `onTogglePinned` versus `mediaPlayer` in the same parameter lists and struct fields — "keep both" every time. Rebasing before steps 1–5 means doing it twice.

One coordination note: #862 adds `closeActionTitle` to the same two structs. Whichever of you lands first, the other gets one more pass — so it is worth knowing that is coming rather than discovering it.

### Not asking you to change

The pinned panel being hardwired to `mediaPlayer: nil` is consistent with `DockPreviewService.togglePinned` already guarding `!windows.isEmpty`, so leaving mini players unpinnable is a defensible call — just say so in the description rather than leaving it looking accidental. And `AppleScriptRunner` being kept off the main thread was correct; that part matched the doc comment exactly.

PathGao:
Separate from the design points above, the localization pass has two defects that will survive whatever you decide about MediaRemote.

**Portuguese was never translated.** `Localization.swift:1310-1311` sits inside the `ptBR` table — its neighbours are "Diminua para ver mais do que está atrás do painel." — and both new values are the English strings verbatim, identical to `:2250-2251` in the `enUS` table. Portuguese users get English for this toggle.

**Four locales are stripped of their diacritics.** Every other line in those files carries them; these do not:

- `Strings+German.swift:227` — "Mediensteuerung **fuer** Musik-Apps anzeigen", two lines above "Öffnungsverzögerung"
- `Strings+French.swift` — "commandes **multimedia**", "les **details** du morceau", "miniatures de **fenetres**"
- `Strings+Spanish.swift` — "apps de **musica**", "**caratula**", "**reproduccion**"
- `Strings+Turkish.swift` — "**Muzik** uygulamalari icin medya denetimlerini **goster**"

Chinese, Japanese, Korean, Russian and Italian are fine. Fixing this is mechanical: für, multimédia / détails / fenêtres, música / carátula / reproducción, Müzik / uygulamaları / için / göster, and a real Portuguese translation.

**One code point, small but on the hot path.** `DockMediaPlayerService.swift:45` compares `artwork?.tiffRepresentation` inside `==`, so every equality check re-encodes both album covers to TIFF. It buys nothing today — `refreshMediaNow` assigns `self.mediaPlayer = snapshot` unconditionally, so the comparison never suppresses a publish — but SwiftUI may still call `==` while diffing the panel, and I could not determine from the source which paths do. Carry the artwork URL in `DockMediaPlayer` and compare that instead; it is the actual identity, and it costs a string compare.

ruvelro:
Addressed the review feedback in b1f1fd0:

- replaced the three-app AppleScript implementation with the shared MediaRemote bridge and media-key commands
- made updates event-driven (no one-second polling timer) and supports any current Now Playing app
- defaults the mini-player off and shows it alongside window previews instead of replacing them
- compares artwork data directly instead of TIFF encodings
- corrected the localized copy/diacritics and generalized it beyond the original three apps

Validation is green: 8,543 local checks, release build, self-test, and both GitHub CI jobs. The branch is mergeable. Please re-review and clear the incomplete label if everything looks good.

PathGao:
b1f1fd0 does what it says on the four design points — MediaRemote instead of the three bundle ids, media keys instead of AppleScript, default off, alongside instead of replacing. Diacritics and the Portuguese translation are correct now. Three things left.

**Progress is wired to nothing.** `snapshot(for:)` passes `position: nil, duration: nil`, and `RadialNowPlayingSnapshot` has no elapsed/duration fields to pass anyway. So `progress` is always nil: the capsule renders at width 0 and both labels read `--:--`, for every track, forever. The PR body still promises "position, duration, progress".

Wiring it properly is more than two extra keys — MediaRemote's elapsed time is a value sampled at `kMRMediaRemoteNowPlayingInfoTimestamp`, so a bar that moves needs elapsed + timestamp + rate and a repaint timer, which is the polling you just deleted. Deleting `progressView` is the cheaper answer, and it takes `position` / `duration` / `progress` out of `DockMediaPlayer` with it.

**The status label is hardcoded English and redundant.** `Text(player.state == .playing ? "Playing" : ...)` is the only user-facing string in this PR that doesn't go through `l10n.s`, in the same PR that added 13 locale entries. It also states what the play/pause icon two rows below already states. Delete it rather than adding three keys across 13 files. (`.stopped` is never constructed either — `snapshot(for:)` only produces `.playing` or `.paused`.)

**A media-only panel can't happen, but the view expects one.** `if !windows.isEmpty` guards the card run, yet `beginWindowSession` returns early when the list is empty (the "app with no real windows shows nothing" path in the hunk's context lines), so `startMediaIfEnabled` never runs for a windowless app. Hover a playing app with no previewable window and you get no panel at all. Either that branch is dead or the early return needs an exception — worth deciding which.

Smaller:

- `startMediaIfEnabled` assigns `mediaObservers = observeChanges(...)` without removing the previous tokens. If any re-point path reaches it without `tearDownVisuals` first, three observers leak per hover for the process lifetime. I can't see enough of `beginSession` to tell whether that path exists; `stopMediaObserving()` as the first line makes it moot either way.
- `DockPreviewSupport.mediaPanelSize(screenVisibleFrame:)` is added and never called. `DockMediaPlayer.hasTrack` likewise.
- `mediaPanelWidth` / `mediaPanelHeight` scale with `PreviewSizing.scale`, but the artwork (112), spacing and button sizes inside `DockMediaPlayerPanel` are fixed. At a small enough scale a 112pt cover won't fit a 174×scale panel. I can't see the range `PreviewSizing.scale` takes, so I don't know whether that's reachable.
- `Defaults.swift:805` — `switcherShowFullscreenWindows` and `switcherAppRules` got joined onto one line, unrelated to this PR.
- The Summary still says "for Spotify, Music, and iTunes" and the Notes paragraph still describes Apple Events and the Automation prompt. Both describe the implementation b1f1fd0 removed.

**Impact:** anyone who turns the toggle on gets a progress bar that never moves, `--:--` at both ends, and an English status word in all 13 locales. No crash, no data loss.
**Merge:** needs the progress row settled (wire it or delete it) and the body rewritten to match b1f1fd0 first. Keeping `incomplete` until then.

ruvelro:
Addressed the remaining review in db5158d:

- removed position, duration and progress from DockMediaPlayer and deleted the inert progress row
- removed the hardcoded/redundant playback-state label and simplified the model to isPlaying
- made the UI explicit that media supplements a real window-preview session; windowless apps do not open a separate media-only panel
- stop existing media observers before every new registration
- removed mediaPanelSize and hasTrack
- made the fixed-content media panel dimensions fixed as well, and reduced its height after removing progress
- restored the joined Defaults entries
- rewrote the body for MediaRemote, media keys, any Now Playing app, default-off behavior and the absence of progress

Local validation: 8,543 checks, release build, self-test and diff check pass. Both GitHub CI jobs are green.

PathGao:
db5158d clears the last round: the progress row and its three model fields are gone, the hardcoded state label with it, `stopMediaObserving()` runs before every registration, `mediaPanelSize` and `hasTrack` are gone, the panel dimensions are fixed, and the body describes what the code does. One copy defect left, and it is the thing b1f1fd0 was for.

**Seven locales still say "music apps."** English is `Show media controls for playing apps`. These kept the pre-generalization wording:

- `Strings+ChineseSimplified.swift:227` — 显示**音乐 App** 的媒体控制
- `Strings+ChineseTraditionalHK.swift:228` and `Strings+ChineseTraditionalTW.swift:228` — 顯示**音樂 App** 的媒體控制
- `Strings+Japanese.swift:227` — **音楽アプリ**のメディアコントロールを表示
- `Strings+Korean.swift:228` — **음악 앱**의 미디어 컨트롤 표시
- `Strings+Russian.swift:228` — для **музыкальных** приложений
- `Strings+Italian.swift:227` — per **app musicali**

The caption directly beneath each one is already generalized — 正在播放媒体的 App, 재생 중인 앱, воспроизводящее приложение, un'app in riproduzione — so inside a single locale the label and its own caption disagree. Someone reading the Japanese or Russian settings list is told this is for music apps and never turns it on for Safari or Podcasts, which is the reach the MediaRemote rewrite bought. `Strings+Turkish.swift:227` "Müzik çalan uygulamalar" is the same shape, milder: it reads "apps playing music" rather than "playing apps."

Smaller:

- `Defaults.swift:76` — the key comment still reads "show player controls for supported music apps." There is no supported set any more.
- `stopMediaObserving()` nils `currentMediaSource` but leaves `mediaPlayer` set. `startMediaIfEnabled` calls it first, so a session that starts without `tearDownVisuals` having run carries the previous app's mini player into `showPanel`'s sizing and into the panel until the next snapshot lands. I still can't see the middle of `beginSession`, so I don't know whether that path exists — `mediaPlayer = nil` inside `stopMediaObserving()` makes it moot either way.
- `snapshot(for:)` builds `NSImage(data:)` inside the `DispatchQueue.main.async` block, so album art decodes on the main thread on every track change. The completion already arrives off-main; constructing `artwork` before the hop is a two-line move. Same class as the TIFF issue, mirrored.
- `showPanel` reads `mediaPlayer` before `startMediaIfEnabled` runs, so the panel always opens window-sized and grows one MediaRemote round trip later. The shared bridge is registered for change notifications now, so it could hold the last snapshot and answer that synchronously.
- `observeChanges` latches `notificationsRegistered` and never clears it, so the first hover with the toggle on registers the process with MediaRemote for its lifetime, including after the toggle goes back off. Whether that is worth an unregister call depends on the symbol existing, which I can't check from here.
- The three transport buttons carry no accessibility label, and `.accessibilityElement(children: .combine)` plus `.accessibilityLabel(accessibilityText)` names the whole block after the track, so previous / play-pause / next arrive unnamed. If the radial menu already has localized names for those three, reuse them rather than adding keys — I can't see whether it does.

**Impact:** anyone reading the settings list in Chinese, Japanese, Korean, Russian or Italian is told the feature is for music apps and won't enable it for the browsers and podcast players b1f1fd0 added. No runtime effect.
**Merge:** needs the seven toggle labels regeneralized first. Keeping `incomplete`.

ruvelro:
Addressed the new review in `efd2f2f5`:

- generalized the Dock media-controls toggle in Simplified Chinese, both Traditional Chinese locales, Japanese, Korean, Russian and Italian; Turkish is generalized too
- updated the stale Defaults comment
- `stopMediaObserving()` now clears the previous `mediaPlayer`
- album art is decoded before hopping to the main queue
- previous, play/pause and next now reuse the localized Radial Menu labels for accessibility

I left the suggested MediaRemote snapshot cache and unregister path unchanged because those were exploratory and no verified unregister symbol/API was identified. Local validation: 8,543 checks, release build, self-test, and diff check all pass. Please re-review and clear `incomplete` if the required copy fix is now settled.

PathGao:
efd2f2f5 settles the copy: all seven toggle labels are generalized and agree with their own captions, Turkish included. The stale `Defaults.swift` comment, the `mediaPlayer` clear in `stopMediaObserving()`, the artwork decode before the main hop, and the reused radial labels on the three buttons are all in. Leaving the snapshot cache and the unregister path alone is right — neither had a verified symbol behind it.

Read the rest of the diff again; nothing at a level that holds it. Two things I did not verify:

- The three notification names and `MRMediaRemoteRegisterForNowPlayingNotifications` are string lookups into a private framework, so a green build says nothing about them. Only a real machine shows whether the panel updates on track change.
- `stopMediaObserving()` clearing `mediaPlayer` made the re-point case better, not moot. `showPanel` still reads `mediaPlayer` before `startMediaIfEnabled` runs, and `refreshMediaNow` returns early when the new snapshot and the current value are both nil, so it never resizes. If a re-point path reaches `beginWindowSession` without `tearDownVisuals` first, the panel opens 144pt too tall for a non-playing app and stays that way for the hover. Same blind spot as before: I can't see the middle of `beginSession`.

**Impact:** none I can name from the diff. Off by default, and the residual risk is the sizing case above, which needs a hover on real hardware to settle either way.
**Merge:** ship it.

ref #303

PathGao:
Since your work impact UI，did you match our new dock preview UI from v3.3.3 beta3?

PathGao:
Holding this one — the data path is the part macOS restricted.

**MediaRemote is entitlement-gated.** Apple closed third-party access to `MRMediaRemoteGetNowPlayingInfo` and the Now Playing notifications in 15.4; an unentitled process gets an empty answer. I can't test that from here, and a green build proves nothing: every symbol is a `dlsym` lookup and every call is optional-chained, so on 27 the failure is silent — `registerNotifications?(.main)` no-ops, the observers never fire, `fetch` finishes nil at the 0.75s timeout, and the panel simply never appears.

Unless you've already confirmed otherwise: run a signed build on 27, hover a playing app, and check the panel shows up and that play/pause flips its icon when pressed. If it comes back empty, this needs a different source (the mediaremote-adapter system-perl route is where other apps went) and the branch gets redone on top of that. The same question covers the radial menu's Now Playing, which already ships on this bridge — if that one works on your Mac today, say so and this closes.

Also `CONFLICTING` again after #830's card and drag rework. Rebase after the 27 answer, not before.

**Impact:** if MediaRemote is closed on 27, the toggle does nothing and reports nothing. Off by default.
**Merge:** needs the macOS 27 check answered, then the rebase. Swapping `ready to merge` for `incomplete`.

ref #303

PathGao:
Closing this, and taking the problem over on our side; you did nothing wrong here, the platform moved under the branch.

Measured today on macOS 26.6.2 with the shipped 3.3.3-beta.3: with a session playing (Control Center showed it, and Vorssaint's own play/pause toggled it), the radial menu's Now Playing card never appears. `MRMediaRemoteGetNowPlayingInfo` answers an unentitled process with nothing since macOS 15.4, the bridge times out at 0.75 s, and the action falls into the "nothing playing" branch, which shows nothing. So the bridge this PR builds on has been empty on every 15.4+ Mac since it shipped, and the Dock Preview card would be empty the same way.

The fix is a different data source, not a rebase. A PR replacing the bridge (the route DockDoor took after 15.4: `/usr/bin/perl` loading a small helper library, since perl carries the entitlement) is being prepared; once it is in, the Dock Preview mini-player gets redone on top of it with you credited for the design and the #303 work. Keep the branch; the panel layout and the toggle copy are reusable as they are.

PR #381, opened 2026-07-28
Add Dock Preview background opacity

ruvelro:
## Summary

- add a Dock Preview background opacity preference
- expose a 0-100% slider in Switcher settings
- apply the opacity only to the preview panel material and border, leaving thumbnails and controls fully readable

## Validation

- swift build


vorssaint (approved):
Thanks for this. Good idea and the strings were all there already.

I pushed some changes onto your branch. The base was behind, so the material had moved into a private view and the option needed rewiring to reach it. I also matched the slider to the other sliders on that page, gave it a floor so the panel keeps its shape, and made reduce transparency win over the stored value.

Tested it here and merging it now.

PR #400, opened 2026-07-31
Fix saved search losing its argument in the command bar

tenbux:
## Summary

- add CommandBarLinks.rankingTitle, scoring a saved-search row against the
  whole query once its name is followed by an argument, against its own name
  otherwise
- wire it into CommandBarService.searchRows for rows with source .links only
- add three CommandBarLinks.rankingTitle checks alongside the other link
  tests in Tests/MetricsTests.swift

## Root cause

Every row is scored by matching each typed word against its title
(CommandBarSearch.score). A saved search's row matches on its own name, so
typing anything after that name, like "ddg the titanic" instead of just
"ddg", introduced words the title could not match, and the row fell out of
the results right before the moment it was meant to run. The row's own
subtitle ("Type what to look for after the name") promises the opposite.

## Verification

- ./build.sh --test: TESTS OK, 5637 checks (5634 on main before this change,
  +3 for the new checks)
- ./build.sh: release build, no warnings
- ./build/Vorssaint --selftest: SELFTEST OK
- ./build.sh --dev --install: built and ran a Developer-id build alongside
  the real app, added a saved search (ddg -> https://duckduckgo.com/?q={query}),
  confirmed "ddg the titanic" now stays matched and opens DuckDuckGo with the
  right query in the default browser

### DIFF
--- a/Sources/Vorssaint/Services/CommandBar/CommandBarLinks.swift
+++ b/Sources/Vorssaint/Services/CommandBar/CommandBarLinks.swift
@@ -117,6 +117,13 @@ enum CommandBarLinks {
         return words.dropFirst(nameWords).joined(separator: " ")
     }
 
+    /// What a link's row is scored against. Its own name works until the
+    /// person types something after it — from there the row has to be scored
+    /// against everything typed, or the extra words sink it out of the list.
+    static func rankingTitle(name: String, query: String) -> String {
+        trailingArgument(query: query, name: name) != nil ? query : name
+    }
+
     /// The URL a link opens, or nil when what was saved cannot be opened. A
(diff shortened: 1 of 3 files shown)


vorssaint:
Thanks for this one. You were right about the worst part of it, the row left the list at the exact moment it was about to run, and the app itself walks people into that by prefilling the name and keeping the bar open.

Merged, with a small follow up on top. The whole query now ranks only the saved searches that actually read what you type after the name, so a plain link or folder does not take over the list when you keep typing, and the ranking used inside a category got the same treatment.

It ships in 3.2.1 and you are credited in the changelog.

Issue #392, opened 2026-07-30
[Feature]: Options to alter scratchpad background opacity.


AB-boi:
### Problem or Motivation

Currently, the translucency interferes with the legibility and is a bit distracting.

### Proposed Solution

Would be nice to have an opacity slider for it, all the way to **opaque**.

### Feature Area

Other

vorssaint:
The Scratchpad opacity control in 3.3.0 now reaches a truly solid background. Update and move the slider all the way up.

PR #430, opened 2026-08-02
Add a Scratchpad background opacity slider

hash00:
## Summary

- add a Scratchpad background control from translucent to fully opaque (#392)
- place clear endpoint labels beside the pad's existing Quick Tools options
- preserve the familiar frosted background by default and adapt the solid fill to light and dark appearances

## Validation

- `./build.sh --test` — TESTS OK (6127 checks)
- `./build.sh --dev --install` — Developer bundle installed and signed
- installed Developer selftest — SELFTEST OK
- real light and dark previews, plus Scratchpad checks at translucent, 50%, and opaque with persistence across relaunch

Related to #392.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,6 @@ pause in chosen apps, and the App Switcher, Scratchpad and radial menu gain clea
   starts on F2, off by default. Thanks to @Mito450.
 - Copied images can become PNG files with ⌘V in Finder, off by default under
   Clipboard. Thanks to @AsphaltDemon.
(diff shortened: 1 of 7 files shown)


hash00:
Good catch from the review, corrected in a506c25.

I had attributed the 0.4 floor to an accessibility-contrast target, but that rationale belongs to `HUDBackdrop`'s high-contrast plate, not to `DockPreviewSupport.backgroundOpacityRange` — whose floor is about the panel still looking like a panel once its title starts reading against the desktop. The comment now gives that reason for the pad, and notes the two share a floor so the sliders behave alike.

Comment-only change; no behaviour difference.

PR #448, opened 2026-08-02
Let a selected annotation move through the drawing order

### LINKED ISSUE #382: [Feature]: Add layer ordering ("Bring Forward" / "Send Backward") for Screenshot "Editor Tools"
### Problem or Motivation

Currently, annotations created in the Screenshot "Editor Tools" (text, arrows, shapes, etc.) strictly follow their creation order for depth/z-index.

If I write a piece of text and then decide to create a solid box behind it to highlight it, the box will inevitably cover and hide the text because it was created afterwards. To fix this, I currently have to delete the text and re-create it on top of the box, which is inconvenient and time-consuming.

### Proposed Solution

Add an option to adjust the layer depth/order of existing objects. 

When an object is selected (or via right-click / context menu), provide controls to:
* **Bring Forward** / **Bring to Front**
* **Send Backward** / **Send to Back**

### Feature Area

Other

### Mockups / References

_No response_

hash00:
## Summary

- let a selected annotation move one step through the drawing order (#382)
- Bring forward / Send backward appear in the style bar once something is selected and there is another annotation to pass
- reordering goes through the undo stack, so it steps back with everything else

Before, annotations were drawn strictly in creation order: a box drawn to highlight text always covered it, and the only fix was deleting the text and drawing it again on top.

One consequence worth flagging: counters take their number from their place in the order, so moving a counter past another renumbers both. That matches what deleting a counter already does, but say the word if you would rather the number stayed with the badge.

The reordering itself is a pure function in `ScreenshotSupport`, so the edge cases are unit tested rather than only reachable through the editor.

## Validation

- `./build.sh --test` — TESTS OK, including four new checks for the reorder
- `./build.sh` — no warnings

Closes #382


hash00:
Both review points fixed in ef45d88.

The buttons stayed live at the ends of the order, so they offered a move that quietly did nothing — and the comment claimed otherwise. Each direction now dims on its own via a `canReorder` check, with the edge cases unit tested alongside the reorder itself.

Test message reworded: the assertion covers both ends, not just the last one.

Issue #371, opened 2026-07-27
[Feature]: a button added to open activity monitor under system monitor tab


chiwaisy-cpu:
### Problem or Motivation

when viewing system tab to monitor CPU GPU and Memory usage, sometimes i would like to open activity monitor app for more information and details

### Proposed Solution

1. i would like to request a feature to be added when viewing system tab (monitor of CPU, GPU, and Memory). add a button to open activity monitor app. 
2. add swap memory usage in system tab

### Feature Area

Other

vorssaint:
3.3.0 adds a system process-inspector button to the CPU, GPU and Memory lists. Please open a separate focused request if swap usage is still important to you.

PR #449, opened 2026-08-02
Open Activity Monitor from the process list

hash00:
## Summary

- add a small button in the process card header that opens Activity Monitor (#371)
- shown on every metric detail that lists processes, next to the card title

The panel already answers "what is using this", and the next question is usually one it does not answer. This saves going to look for the app.

The label is a new string in all supported languages, using Apple's own name for the app in each so it matches what the user sees in Finder.

Scope note: #371 also asks for swap memory in the system tab. That is a sampler and a metric row rather than a button, so I left it out to keep this to one topic. Happy to do it separately if you want it.

## Validation

- `./build.sh --test` — TESTS OK
- `./build.sh` — no warnings

Refs #371


hash00:
Fixed in the latest commit.

The comment claimed macOS resolved the location while the code hardcoded an absolute path — the review was right that those disagreed, and that the path has moved between releases. It now asks for `com.apple.ActivityMonitor` by bundle id and keeps the known path only as a fallback.

PR #459, opened 2026-08-03
Add optional right-click Keep Awake toggle

yspreen:
## Before
Right-clicking a Vorssaint menu bar item always opened the context menu.

## After
A new opt-in Energy setting toggles Keep Awake directly on right-click. The existing context menu remains the default.

This gives Amphetamine users familiar one-click switching without changing the workflow for existing users. Includes all 13 supported languages.

## Tested
- `./build.sh` without warnings
- `./build/Vorssaint --selftest`
- Both right-click modes in the developer app

`./build.sh --test` is locally blocked by existing `UInt8` inference errors at `Tests/MetricsTests.swift:7148`. A clean `main` checkout reproduces the same errors, confirming they are unrelated to this change.

AI used for translations


yspreen:
@vorssaint this is the only thing keeping me from switching over as an amphetamine user. 2 LOC changed plus locales. do you think we could get this merged? or should I stick with amphetamine for the time being?

vorssaint (approved):

PR #545, opened 2026-08-07
feat: Integrar controle de ventoinha e métricas profundas na interface do usuário

alebypegasus:
## O que mudou

* **Menu Principal Integrado**: As seções de *Fan Control* (controle das ventoinhas - Auto/Frio/Médio/Máximo) e o *Monitoramento Profundo* (uso de bateria, rede, e temperatura de todos os sensores) foram movidos diretamente para dentro do painel principal e perfeitamente integrados com o `Theme.swift`, obedecendo os estilos e `ColorScheme` já estabelecidos.
* **Métricas Detalhadas via TouchID**: Criada uma *sheet* protegida para mostrar dados que requerem `sudo`, como `powermetrics` de CPU e GPU.
* **Crash Fix**: Adicionados `nil-checks` de `bundleIdentifier` para previnir exceções quando o usuário tenta executar o script via terminal sem um `Info.plist`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,14 @@ E2E/
 *.xcodeproj
 xcuserdata/
 .swiftpm/
+
+# IDEs / Editor
+.vscode/
+.idea/
+*.swp
+*~
+test.swift
+test_*
+test
+*.o
+.o
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ test_*
 test
 *.o
 .o
+*.zip
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,24 @@ Vorssaint 3.3.3 adds full manual and temperature-based Fan Control, directional 
 - The recording editor now reads composition duration directly without background
   queries. Thanks to @Bald-M.
 
+## [3.3.3] - 2026-08-22
+
+### Summary
+Vorssaint 3.3.3 introduces Universal Fan Control for Apple Silicon and Intel Macs,
+Processor Power Modes (Low Power Eco, Balanced, Maximum Turbo), Safe Memory Purge,
+expanded Developer Caches cleaning, Finder Delete and Revert shortcuts, and Deep
+Hardware and Energy metrics.
+
+### Added
+- Processor Power Modes: Low Power (Eco) to save energy and reduce heat, Balanced (Default),
+  and Maximum (Turbo) for heavy compiling and sustained workloads without throttling.
+- Memory Purge: Instant RAM cache release to free inactive memory when pressure is high.
+- Universal Fan Control: Full hardware support for Apple Silicon and Intel Macs with SMC
+  FS! bitmask and dynamic RPM management.
+- Extended Developer Cleaner: Cleans CocoaPods, SwiftPM, Gradle, and NPM build caches.
+- Finder Delete & Revert shortcuts: Move to Trash with Delete and undo with Shift+Delete.
+- Deep Metrics Dashboard: Detailed breakdown of hardware, storage, networking, and battery.
+
 ## [3.3.2] - 2026-08-20
 
 ### Summary
--- a/Resources/Info.plist
+++ b/Resources/Info.plist
@@ -43,7 +43,7 @@
 		<string>MacOSX</string>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>79</string>
+	<string>80</string>
 	<key>LSApplicationCategoryType</key>
 	<string>public.app-category.utilities</string>
 	<key>LSMinimumSystemVersion</key>
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -109,6 +109,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
             KeepAwakeManager.shared.activateOnLaunchIfNeeded()
         }
         FanControlService.recoverIfNeeded()
+        ProcessorPowerModeService.shared.syncWithPreferences()
         // One binding per feature: only available features are touched, so a
         // feature switched off in the hub never even instantiates here.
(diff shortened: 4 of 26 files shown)


PathGao:
Triage note from a collaborator sweeping the open PR queue, not a maintainer verdict. The `declined` label is about this PR's current shape — several pieces inside it are worth keeping, and this note says which, so the label isn't just a closed door.

## Why the PR as a whole can't be taken

**The fan half rewrites code that already ships, and drops its safety work.** `Services/FanControl/` is the real implementation on `main`. There, `startCooling` refuses to run when another tool already owns the fans (`FanControlHardware.swift:92`, `:95`), writes the manual-mode/target pair, sleeps, re-reads the mode, and falls back to a force-test path that applies a protective maximum target before verifying (`:117`, `:133-136`, `:149`). This PR replaces all of it with `_ = setManualMode(...)` plus `if writeTargetRPM(...) { anySuccess = true }`, and reports success if any single fan accepted a write. An SMC write returning success is not evidence the fan entered manual mode — that is precisely what the `verifyCooling` read-back exists to catch, and the stopped-fan-on-Apple-Silicon case is what the force-test branch exists to handle. `updateCooling` likewise loses its `activeTargets.count == fans.count` and mode checks.

**It claims the release.** `Tests/MetricsTests.swift` moves the support-prompt and highlights-tour pins from 3.3.2 to 3.3.3, and `Resources/Info.plist` bumps `CFBundleVersion` 79 → 80. Those assertions exist to force a deliberate decision at release time, so a PR editing them is making a maintainer's call. It is also stale now: `main` is at 3.3.3-beta.2 / build 81.

**New UI ships hardcoded Portuguese.** `ProcessorPowerModeService.swift` (`title`/`description`), `DeepMetricsView.swift` and `SystemSection.swift` bypass `Core/Localizations/Strings+*.swift` (`CONTRIBUTING.md:70-75`).

## Worth keeping — best as separate small PRs

**1. `build.sh`.** Two independent one-liners. `TARGET="$(uname -m)-apple-macosx14.0"` in place of the hardcoded `arm64-apple-macosx14.0` (`build.sh:42` on main), and probing `-Xfrontend -interface-compiler-version` before using it rather than applying it unconditionally (`build.sh:138-142`). The probe is the more valuable of the two: as `main` stands, the flag is added whenever `SDK == PINNED_SDK`, so a toolchain that doesn't accept it fails the build outright. The `export PATH="/usr/bin:/bin:/usr/sbin:/sbin:$PATH"` line added in the same file should **not** come along — it compensates for one machine's environment and changes tool resolution for everyone.

**2. `Services/Power/ProcessorPowerModeService.swift`.** Nothing on `main` touches `pmset lowpowermode` or `purge`, so this is genuinely new work rather than a second copy of something. Three things to fix first:

- `setMode` sets `success = true` after the `AdminShell.runSync` fallback without reading its return value, so a cancelled admin prompt reports as a successful mode change.
- `.maximum` is `IOPMAssertionCreateWithName(kIOPMAssertionTypePreventUserIdleSystemSleep, ...)` — the same assertion `KeepAwakeManager.swift:382` already takes. It does nothing to CPU performance; it is Keep Awake wearing a "Turbo" label. Either drop the third mode or name it for what it does.
- Widening the NOPASSWD sudoers rule (`ShellSupport.swift:174`, today restricted to `pmset disablesleep 0/1`) to cover `pmset -a lowpowermode` and `/usr/sbin/purge` is a permission-surface decision. It deserves its own discussion, separate from the feature.

**3. The admin-gated deep-metrics sheet in `DeepMetricsView.swift` — the idea, not this implementation.** As written, `.onAppear` unconditionally runs `NSAppleScript` with `do shell script ... with administrator privileges`, so simply opening the sheet prompts for the admin password every time, with a hardcoded 700x600 frame. Gating it behind an explicit user action and going through the existing `AdminShell` path instead of a fresh `NSAppleScript` would make it reviewable.

**4. `TemperatureSensorSelector.isCPUTemperatureKey` gaining the `TC` prefix.** A plausible Intel fix, but it lands unguarded by platform while the neighbouring `Tf` rule is gated on `.appleM3Family`, and nothing tests it. Small enough to be its own PR with one assertion.

`Services/Finder/FinderDeleteShortcuts.swift` is also new relative to `main` (which has cut/paste and rename, but no delete), so it isn't duplicate work either — it is just unrelated to a PR titled for fan control and deep metrics, and belongs on its own.

## One mechanical note

This PR is opened from your fork's `main`, and this repo's `main` has had its history rewritten, so rebasing this branch is more painful than it looks. Opening the items above as fresh topic branches off current `origin/main` will be less work than repairing this one.

PathGao:
Thanks for the work here. Closing this one, for three reasons that are about how it sits against current `main` rather than about the ideas in it.

**It re-implements features that already shipped.** `Sources/FanControlHelper` (a privileged LaunchDaemon), `SMCClient` and `Sudoers` are all on `main` today, and this changes `FanControlHardware.swift` and `FanControlXPC.swift` alongside a second fan-control path. Whatever this branch was started from, it has not been reconciled against what the app now does.

**It hand-rolls privilege escalation around the infrastructure that already exists**, in a SwiftUI view file:

```swift
private func runSudoAppleScript(commands: [String]) -> String {
    let script = "do shell script \"\(commands.joined(separator: " && "))\" with administrator privileges"
```

`AdminShell` exists in `ShellSupport.swift` for exactly this, with escaping. The version here interpolates the command string straight into an AppleScript literal with none. There are also several direct `Shell.run("/usr/bin/sudo", ...)` calls for `pmset`, `purge` and `powermetrics`.

**The admin prompt fires from a view's load path.** Opening the metrics panel asks for an administrator password, rather than a specific action the user chose asking for it. That is the wrong place for that dialog regardless of the plumbing behind it.

On top of that it is +946 lines across 26 files, touching AppDelegate, FeatureRuntime, FeatureCatalog, Defaults and `build.sh`, which makes all three harder to unpick in review than to restate here.

If you still want the deep metrics part, it would be much easier to review as its own branch off current `main`: no fan-control changes, everything privileged routed through `AdminShell`, and the admin prompt attached to an explicit user action.

Issue #441, opened 2026-08-02
[Bug]: Three-finger click unreliable (frequently misidentified as two-finger click)


justareported-blip:
### Description

The three-finger click gesture mapped to a middle-click is highly unreliable. When performing a three-finger click, the app frequently fails to register the third finger, misidentifying the gesture as a two-finger click instead. The failure rate is around 50% which is completely unacceptable.

Alternative app tested (MiddleClick) does not exhibit this issue on the exact same hardware/system configuration, suggesting an issue with touch detection thresholds or gesture recognition logic within the app.

### Steps to Reproduce

1. Enable the Three-Finger Click acts as Middle Click setting in the Vorssaint app settings.
2.	Place three fingers on the trackpad and perform a click.
3.	Observe the registered action (e.g., in a web browser, try opening a link in a new tab using the gesture).
4.	Repeat the click multiple times to test consistency.

Expected Behavior
Every deliberate three-finger click reliably triggers a middle-click action (e.g., opens a link in a new tab, closes a tab).

Actual Behavior
Around 50% of three-finger clicks fail to trigger the middle-click action. Instead, the system acts as if a two-finger click (secondary/right-click) was performed.

### Vorssaint Version

3.2.0

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

AB-boi:
Okay so i have occasionally had the same issue with this three finger click (it is one of my most used gestures) but I couldn't find a specific situation that caused it consistently. It happens and then sometimes it doesn't. I use it most within my browser (Safari). It could happen on a link or on tab that I am trying to close with middle click.

vorssaint:
Three-finger middle click handling was tightened in 3.3.0 so an initial secondary-click reading no longer wins. Let me know if it still misfires.

PR #551, opened 2026-08-08
Improve three-finger middle-click reliability

huseyinbal05:
## What changed

This improves how physical three-finger clicks are detected and handled.

- Listen for mouse events at the HID event tap instead of the session event tap.
- Remove the 40 ms settling requirement before accepting a three-finger click.
- Reduce the duplicate-click guard from 300 ms to 90 ms.
- Update the related tests for the new timing behavior.

## Why

I was running into two separate responsiveness problems.

The 40 ms settling requirement could cause a quick three-finger click to pass through as a normal left or right click. Removing this delay allows the click to be transformed as soon as a fresh three-finger contact is detected.

Fast repeated clicks also felt unresponsive because the 300 ms duplicate guard could suppress a valid second click. This was especially noticeable when opening several links or closing browser tabs. Reducing the guard to 90 ms still filters the immediate duplicate event it was designed for, while allowing deliberate consecutive clicks to register normally.

Using the HID event tap also handles the mouse event closer to where it enters the system, making the transformation more consistent.

The existing safety checks remain in place. Clicks are only transformed when exactly three fingers are detected from a recent touch frame. Two-finger, four-finger, stale, and conflicting system gestures continue to pass through normally.

## Testing

I built and tested the developer version on my Mac. Quick three-finger clicks and repeated clicks now register reliably, and normal left and right clicks continue to work as expected.

I also ran the focused middle-click tests and type-checked the Swift sources.

Related to #441.

### DIFF
--- a/Sources/Vorssaint/Services/MiddleClick/MiddleClickService.swift
+++ b/Sources/Vorssaint/Services/MiddleClick/MiddleClickService.swift
@@ -54,8 +54,6 @@ final class MiddleClickService: ObservableObject {
     private let stateLock = NSLock()
     private var fingerCount = 0
     private var lastFrameUptime: TimeInterval = 0
-    /// When the contact count last became exactly three.
-    private var threeFingersSince: TimeInterval?
 
     // Tap-to-middle-click (issue #161), all under `stateLock`. A candidate
     // starts when the chosen finger count lands, collects movement, and is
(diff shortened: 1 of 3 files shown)


vorssaint (changes requested):
Thanks for working on this. I cannot accept this change yet.

Removing the 40 ms settle check reopens the false middle click caused by tap to click. Shortening the 300 ms guard can also let the synthesized follow-up click through again. Moving the event tap to the HID entry point does not solve the race between the click and the third contact frame.

Please keep these safety guards and base the reliability change on a reproduced timing trace that distinguishes a physical press from synthesized clicks.

PR #589, opened 2026-08-10
Show what a main switch turns off in Settings

PathGao:
Follow-up to #493, which tied the Clipboard History settings to their switch. Two panels were left inconsistent with it, in opposite directions.

### Quit on close

With the feature off, the exception list stayed fully editable. Nothing reads it there: `autoQuitExceptions` has exactly one reader, `AutoQuitService.checkWindows`, which only runs while the service does. Adding an app while the feature is off does nothing and gives no sign of it.

The list and the *Add app…* button now follow the switch, the same way the Clipboard rows do.

### Clipboard

The opposite problem. `Show in the panel` sat in the same section as the four rows that #493 disabled, but it deliberately does *not* follow the switch, and shouldn't: with capture off the panel entry stays, says so in its caption (`MenuPanelView`, `clipboardEnabled ? caption : disabled`) and still opens the saved items. One live row in the middle of four greyed ones reads as a row that was missed.

No behavior change here, only grouping: it moves to its own section, so the group that follows the switch is the group that looks like it.

### Not changed

Text Snippets looks like the same defect and is not one. `textSnippetsEnabled` only gates automatic expansion; `SnippetLibraryService` reads `snippetLibraryEnabled` and the shared snippet list independently of it, so the library and the list still work with the main switch off. Disabling them would break something that works — the same reason #493 kept saved items usable. Left alone deliberately.

### Checks

`./build.sh` clean, `./build.sh --test` (TESTS OK, 6993 checks) and `./build/Vorssaint --selftest` (SELFTEST OK). No new strings, so no translations needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,18 @@ All notable changes to this project are documented here. The format follows
 
 ### Summary
 Vorssaint 3.3.2 prevents a Switcher shortcut crash and keeps its labels in sync.
-It also removes false App Updates listings, improves feature setup and clarifies Full Disk Access.
+It also removes false App Updates listings, improves feature setup, clarifies Full Disk Access
+and shows which options a main switch turns off.
 
 ### Changed
 - Installed Features rows open their Settings page and highlight the relevant
   controls. Thanks to @dorlugasigal.
 - Cleaning and the Uninstaller now explain how to grant Full Disk Access and
   that the app must reopen. Thanks to @PathGao.
+- Quit on close exceptions are no longer editable while the feature is off, since
+  nothing reads them there. Thanks to @PathGao.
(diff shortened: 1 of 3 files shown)


PathGao:
Rebased onto `main` (2df4a46). The only conflict was CHANGELOG.md, where #590 and #593 had landed in the shared 3.3.2 section — the two entries here are folded into your `### Changed` list and the summary, in your wording. The two source hunks are unchanged.

Re-checked after the rebase: `./build.sh` clean, `./build.sh --test` (TESTS OK, 7008 checks), `./build/Vorssaint --selftest` (SELFTEST OK). #600's `.settingsSectionAnchor(.clipboardHistory)` sits on the first section and is untouched by the section split here.

PR #600, opened 2026-08-10
feat(settings): open installed features at their settings

### LINKED ISSUE #599: [Feature]: Open installed features directly in their Settings section
### Problem or Motivation

The Features hub is useful for installing only the parts of Vorssaint a user wants, but after installing a feature the hub does not show where that feature is configured. Because Vorssaint has many features and several Settings pages contain multiple feature sections, a new user can install something and then need to search the sidebar to find its controls.

### Proposed Solution

Make each **installed** feature row in Settings > Features navigable:

- Clicking anywhere on an installed row opens the relevant Settings page.
- If the feature lives in a subsection of a shared page, scroll that section into view and briefly highlight it.
- Show a trailing chevron on rows that can navigate.
- Keep uninstalled rows non-navigable; their Install button continues to be the only action.
- Keep the existing Install/Uninstall behavior and avoid letting the nested Uninstall button also trigger navigation.

The implementation can use an exhaustive feature-to-destination map with stable, non-localized section identifiers so future features must explicitly choose their Settings destination. Features without a dedicated page would open the nearest relevant Settings control (for example General panel configuration, General music blocking, or Shortcuts).

Before implementing, I would appreciate maintainer guidance on those fallback destinations, especially panel-oriented features such as Mixer and features with little or no dedicated configuration.

### Feature Area

Other

### Mockups / References

The existing Features hub row layout can stay the same apart from a trailing navigation chevron for installed rows.

dorlugasigal:
Installed Features hub rows now open the Settings page where each feature is
configured. Features that share a page land on their exact section, which is
scrolled into view and briefly highlighted.

## **Demo**:

https://github.com/user-attachments/assets/bebf3d69-ba54-4e89-8ece-09dc4e2048ec

## Changes

- Added an exhaustive, typed **feature-to-Settings destination map** with stable
  section anchors for shared pages.
- Made installed hub rows navigable with a trailing chevron while preserving
  independent Install and Uninstall controls and accessible labels.
- Added one-shot Settings routing that handles repeated requests and prevents
  stale request consumption.
- Added reusable section focusing with a guarded layout retry, transient
  highlighting, and Reduce Motion support.
- Anchored feature controls across General, Energy, Mouse, Switcher, Clipboard,
  Cut/Paste, Quick Tools, Shortcuts, and Monitor settings.
- Kept features without a separate configuration surface free of dead-end
  navigation affordances.
- Added coverage for destination visibility, anchor validity, repeated requests,
  fallback behavior, and Cleaner tool-hint preservation.

## Related Issues

Closes #599

## Notes

- `./build.sh --test` passed with 7006 checks.
- `./build.sh` completed successfully.
- `./build/Vorssaint --selftest` reported `SELFTEST OK`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and the project uses
 [Semantic Versioning](https://semver.org/).
 
+## [3.3.2]
+
+### Summary
+Vorssaint 3.3.2 makes installed features easier to configure.
+
+### Changed
+- Installed Features rows open their Settings page and highlight the relevant
+  controls. Thanks to @dorlugasigal.
+
 ## [3.3.1] - 2026-08-09
 
 ### Summary
--- a/Resources/Info.plist
+++ b/Resources/Info.plist
@@ -37,7 +37,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.3.1</string>
+	<string>3.3.2</string>
 	<key>CFBundleSupportedPlatforms</key>
 	<array>
 		<string>MacOSX</string>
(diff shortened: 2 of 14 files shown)


vorssaint:
thanks @dorlugasigal 

merged

Issue #504, opened 2026-08-05
[Feature]: A yt-dlp GUI and able to use it in command bar(just like Raycast)


Andrewzhang00GH:
### Problem or Motivation

Raycast have a simple yt-dlp tool, I want it in vorssaint too :)

### Proposed Solution

Just a simple yt-dlp GUI that shows:
1. A place to paste url
2. Choice for quality/type/...
3. Nice looking download button
4. It is fine for me if it is either a status bar that shows download percentage, or just the actual terminal output.
5. Shows where the file is. (Better if it directly open the dir of the file.)

### Feature Area

Other

### Mockups / References

Raycast

<img width="756" height="476" alt="Image" src="https://github.com/user-attachments/assets/eb6c3409-17c8-451f-8b50-0a8a25df74f4" />

<img width="745" height="472" alt="Image" src="https://github.com/user-attachments/assets/826132c3-69ee-4c48-b734-71597040c151" />

PathGao:
Thanks for the write-up and the screenshots — the request is clear, and #661 showed it can be built well. The reasons for declining are about what shipping it would commit the project to, not about the idea being uninteresting.

**The dependency chain.** A working downloader needs yt-dlp, FFmpeg (with ffprobe) and Deno. Deno is in that list because yt-dlp needs a JavaScript runtime to solve YouTube's signature challenge. That mechanism — working around a technological protection measure — is the specific thing the 2020 RIAA notice against youtube-dl was written about under DMCA 1201. A signed, branded app whose feature drives that puts the project's distribution identity behind it in a way that you running the CLI yourself does not.

**The cookie path sharpens it.** To match what people actually expect from such a tool it has to read browser login cookies, under Full Disk Access, for age-restricted and members-only videos. That is closer to working around an access control than to saving a public video, and it is the hardest part to defend if anyone asks.

**Maintenance lands here rather than upstream.** yt-dlp breaks whenever YouTube changes its player and ships fixes on a near-daily cadence. No app release cadence tracks that. Every breakage would arrive as a Vorssaint bug whose root cause lives in another project, and the reporter has no way to tell the difference.

**Cost.** The one implementation that was attempted came to +10,520 / -185 across 62 files, touching AppDelegate, FeatureCatalog and every localization. That is a lot of surface to carry for a feature the project may have to withdraw later.

Closing as not planned. Raycast's extension runs in a different distribution and liability position than a signed Mac app, which is why the comparison does not carry over.

PR #661, opened 2026-08-16
Add Video Downloader

huseyinbal05:
### Summary

Adds **Video Downloader**, a native, local-first media downloader built directly into Vorssaint.

Paste a link from YouTube or any supported site to check available resolutions, audio streams, and subtitles, then save an Apple-ready MP4/MKV video or extract clean M4A audio. Everything runs locally on your Mac using `yt-dlp` and `ffmpeg`—no third-party cloud services, accounts, or tracking.

Ref #504

---

### What’s inside

* **Video & Audio Downloads**:
  * Download video up to max resolution (MP4-first remuxing with MKV fallback for codecs Apple containers don't support).
  * Extract clean `.m4a` audio directly from available audio streams.
* **Embeddings & Metadata**:
  * Embeds cover art/thumbnails, chapters, and metadata directly into the media container without leaving stray sidecar files behind.
  * Subtitle selector with support for embedded manual tracks and automatic captions (`mov_text` for MP4, subtitle tracks for MKV).
* **Live Progress**:
  * Real-time progress bar with live download speed (MB/s) and ETA calculations.
* **Privacy & Isolation**:
  * URLs are passed to `yt-dlp` via standard input (`--batch-file -`) so links never leak into process arguments.
  * Runs with `--ignore-config --no-config-locations --no-plugin-dirs` to keep extractor execution isolated and predictable.
  * Optional browser cookie session support for age-restricted or member-only content (read in-memory during inspection/download only; never stored).
* **Tool Setup in Settings**:
  * Built-in checker under Settings for `yt-dlp`, `ffmpeg` (with `ffprobe`), and `deno` with one-click Homebrew installation and Terminal fallback.
* **Full Localization**:
  * Translated across all 12 supported languages.

---

### Screenshots

<details>
<summary>View Screenshots</summary>

<br>

Video Downloader Main View

Stream & Quality Selector

Download Progress and Speed

Settings & Tool Setup

</details>

---

### Verification

- [x] `./build.sh --test` passes with no regressions.
- [x] `./build.sh --dev` compiles without warnings and signs bundle cleanly.
- [x] `./build/Vorssaint --selftest` reports `SELFTEST OK`.
- [x] Verified video and M4A audio downloads with embedded cover art on YouTube.
- [x] Verified Homebrew missing-tool detection and custom download folder selection.


PathGao:
Thank you for this — it is a large and carefully built feature, and what follows is about what it commits the project to, not about the quality of the work.

**The dependency chain.** Downloading needs yt-dlp, FFmpeg (with ffprobe) and Deno, installed through Homebrew. Deno is in that list because yt-dlp needs a JavaScript runtime to solve YouTube's signature challenge. That mechanism — working around a technological protection measure — is the specific thing the 2020 RIAA notice against youtube-dl was written about under DMCA 1201. A signed, branded app whose headline feature drives that puts the project's distribution identity behind it in a way that a user running the CLI themselves does not.

**The cookie path sharpens it.** The feature reads browser login cookies, under Full Disk Access, to fetch age-restricted and members-only videos. That is closer to working around an access control than to saving a public video, and it is the hardest part to defend if anyone asks.

**Maintenance lands here rather than upstream.** yt-dlp breaks whenever YouTube changes its player and ships fixes on a near-daily cadence. No app release cadence tracks that. Every breakage arrives as a Vorssaint bug whose root cause lives in another project, and the reporter has no way to tell the difference.

**Size.** +10,520 / -185 across 62 files, touching AppDelegate, FeatureCatalog and every localization. That is a lot of surface to carry for a feature the project may have to withdraw later.

Closing for those reasons. #504 stays open — this does not settle the request either way, and I would rather leave that conversation to the issue.

PR #673, opened 2026-08-17
Fix mixer percentage input focus in popover

### LINKED ISSUE #620: [Feature]: Clicking on percentage number in volume mixer should allow you to type a specific percentage
### Problem or Motivation

It just makes sense to me.

### Proposed Solution

When you click on the percentage icon or number associated with any of the apps on the volume mixer page, it'll open a simple text box next to it where you can type a specific number (e.g. 40) and press enter, and it'll set the volume of that app to the number you typed in.

This should also work for the primary volume slider.

### Feature Area

Menu bar

### Mockups / References

_No response_

KSI-cell:
## Summary

- replace the popover-sensitive SwiftUI focus request with a native text field that focuses and selects its value after joining the popover window
- support exact primary-output and per-app percentages, clamped to their supported ranges
- commit with Enter, cancel with Escape or focus loss, and keep invalid input editable for correction
- keep only one percentage editor active at a time

## Context

Direct percentage input landed on `main` in `bf6018a` while this branch was in development. In the developer build, the SwiftUI `FocusState` request could run before the menu-bar popover attached its backing field, so clicking a percentage did not reliably make the next keystroke replace the value. This change hardens that interaction while preserving the feature's compact appearance.

## Testing

- `./build.sh --test` — `TESTS OK (7392 checks)`
- `./build.sh` — release bundle built successfully
- `./build.sh --dev` — developer bundle built successfully
- manually verified direct entry for primary and per-app volume, Enter to commit, and Escape to cancel

Closes #620


vorssaint:

PR #674, opened 2026-08-17
Add Quarantine Manager feature

naveenkrdy:
## Summary

Ports the "quarantine-manager" Raycast extension idea into Vorssaint natively - inspect and clear the `com.apple.quarantine` extended attribute (the flag that triggers the Gatekeeper "downloaded from the internet" prompt) on files and apps, plus full extended-attribute view/removal for anyone who wants to go further.

- **Settings page**: scan `/Applications`, drag files/apps in, or pick them via a panel; app bundles collapse into a single row (with the real app name/icon) instead of flooding the list with every internal file; review and clear the quarantine flag from a checklist.
- **Menu-panel tile**: the same flow, in a compact form.
- **Command Bar action**: with files selected in Finder, an action clears quarantine from the current selection directly - toggleable independently from the main feature switch, in a "Show in Command Bar" option on the Settings page.
- **Detail sheet**: every extended attribute on a file, not just quarantine, with per-attribute removal. SIP-protected attributes (`com.apple.provenance` is the known case - Apple made it kernel-managed and unremovable, even as root, starting in Ventura) are detected and shown locked in the UI instead of silently failing or prompting for an admin password for something that can never succeed.

Ships as an opt-in beta feature (`AppFeature.quarantineManager`), off by default like Fan Control.

## Test plan

- [x] `./build.sh --test` passes (`TESTS OK`, 7404 checks)
- [x] Manually exercised: scanning `/Applications`, dragging/picking files, clearing a batch selection, the Command Bar action on a Finder selection, and the detail sheet's per-attribute and "remove all" flows

[See it in Action](https://pub-d1f174c205bf431e9aa4eb83be698fc8.r2.dev/snapzy/1787236863-b7d625e1-snapzy_recording_2026-08-20_20-09-37.mp4)

Snapzy_2026-08-20_20-40-16_067


PathGao:
Thanks for digging into the recursion question — I checked the code against your answer and everything you said holds. Recursive clear on `.app` bundles is `xattr -dr com.apple.quarantine`, and there is no code-signature check anywhere in the PR.

Two things your answer doesn't cover, though, and they're the ones I'd actually want scoped before merge:

**1. `removeAllAttributes` uses `xattr -cr`.** That clears *every* extended attribute on every file in the bundle, not just quarantine — `com.apple.macl` (TCC's per-file access grants), Finder tags, and the `com.apple.cs.*` attributes that carry signatures for non-Mach-O files. It also doesn't go through the `isRemovable(name)` guard; only `removeAttribute(named:)` does. That action reaches well past what "Quarantine Manager" implies, and it's the one I'd gate or narrow.

**2. Both paths escalate on failure.** `AdminShell.runSync` is `do shell script … with administrator privileges` (`ShellSupport.swift:28`), so the fallback is the sudo version — one password prompt, a batch of paths joined with `&&`. That's a step beyond "same as running the command yourself in Terminal" and worth saying out loud in the PR description. The quoting itself is fine: POSIX single-quote escaping, correct.

So I'd skip the signature check you offered — I don't think that's the guardrail that matters here. Scoping `-cr` is.

Minor, unrelated: `shellQuote` is defined twice, once in `QuarantineManagerService` and once in the view.

PathGao:
This is a genuinely good feature, and I want to be clear that the code is not the problem. The shell quoting is correct, the batch-then-retry fallback is sensible, and collapsing app bundles into one row instead of flooding the list with every internal file is a nice touch. I clear quarantine flags myself while developing, often enough that I would use this.

I am closing it anyway, because of what the capability is rather than how it is built.

`com.apple.quarantine` is the only thing that makes Gatekeeper evaluate a downloaded app before it runs. Removing it is not a cleanup step next to the other extended attributes — it is the one that turns the check off:

```swift
commands.append("/usr/bin/xattr -dr com.apple.quarantine " + recursiveAdmin.map(shellQuote).joined(separator: " "))
let ok = AdminShell.runSync(commands.joined(separator: " && "), prompt: ...)
```

As root, recursive, in batch, reachable from a `/Applications` scan, a menu-panel tile and a Finder-selection Command Bar action.

Two things follow from that, and neither is fixable by revision:

**Malicious distribution.** "Download this, then run `xattr -dr com.apple.quarantine` on it" is the standard instruction attackers give their victims, precisely because Gatekeeper would otherwise stop the payload. This feature turns that instruction into a click inside a signed, notarized app. Someone talking a user through installing something harmful gets a much easier script to read out.

**Legal and distribution exposure.** Shipping a signed app whose advertised feature is disabling a macOS security control puts the project's signing identity behind that use, in a way that a developer running the command themselves does not.

A developer who needs this already has the one-line command, and the terminal is the right place for it — it is a deliberate act each time, with no UI making it routine for people who do not know what the flag is for.

So: not a judgment on the work, and not something a rewrite fixes. Closing. If you want to keep any of the general extended-attribute inspection that has nothing to do with quarantine, that is a separate and much smaller conversation.

Note for #694, which is stacked on this: it will need rebasing onto `main`.

PathGao:
You can check Issue #977—this is a feature I plan to implement.

We receive too many requests that boiled down to running a one- or two-line terminal command. These features are often niche (even though I use many of them daily) and carry non-trivial risks—both legally and operationally. Hardcoding them as permanent UI buttons shipped to every user is simply not the right approach.

Issue #977 outlines a generalized architecture: allowing users to package their own local scripts into custom shortcut buttons. Shifting control and permission management back to the user is a far cleaner and more scalable solution.

For these reasons, I personally lean toward rejecting this PR (though the maintainer will make the final call). I believe the generalized approach mentioned above is the correct long-term fix. The vast majority of users shouldn't have access to high-risk controls they don't fully understand, which could easily be exploited. If users explicitly opt to configure and run their own scripts, both the control and the associated risks remain in their hands—just like running commands in a terminal.

PR #683, opened 2026-08-17
Add Kill Process feature

naveenkrdy:
## Summary
- Ports Raycast's Kill Process command: list, kill, force-kill, kill-all-by-name, kill-process-tree, and restart running processes.
- Reachable from a dedicated Settings page (sortable Process/CPU/Memory/PID columns) and from the Command Bar, both as per-process rows and an inline browsable category.
- Ships off by default as an opt-in beta (same posture as Fan Control and Quarantine Manager) since force-killing the wrong process can destabilize the Mac.

## Test plan
- [x] `./build.sh` — clean, no warnings
- [x] `./build/Vorssaint --selftest` — `SELFTEST OK`
- [x] `./build.sh --test` — `TESTS OK (7629 checks)`
- [x] Manually verified in Settings → Kill Process: sortable columns, kill/force-kill/kill-all/kill-tree/restart actions, list stays stable during background refresh
- [x] Manually verified in Command Bar: typing "kill process" surfaces the Settings row and the browse row; entering the browse category shows a live filterable process list; "Show in Command Bar" toggle gates both

[see it in Action](https://pub-d1f174c205bf431e9aa4eb83be698fc8.r2.dev/snapzy/1787239585-3c875ed2-snapzy_recording_2026-08-20_20-55-24.mp4)

image


vorssaint:
Thanks for the contribution. I tested it locally, added protections for critical system processes and merged it.

PR #685, opened 2026-08-17
Shelf: refresh a tile immediately when a drop merges into it

tenbux:
# Shelf: refresh a tile immediately when a drop merges into it

## Summary

Dropping a file onto an existing Shelf tile, to merge it into a pile, mutated
the underlying data correctly and immediately, but the tile itself kept
showing its old, single-item appearance until some unrelated later event
forced a redraw. The drop had visibly worked (the file was gone from wherever
it came from) but the Shelf didn't show it until you clicked something.

## How it works

Shelf tiles are AppKit views, rebuilt from scratch on every SwiftUI
`updateNSView` pass via `ShelfTilesView.rebuildTiles`. The merge itself
(`ShelfService.mergePasteboard`) mutates state synchronously, but
`updateNSView` was observed lagging behind that mutation while the AppKit
drag session triggering it was still unwinding. Two less invasive fixes were
tried first and both failed empirically: pumping the run loop synchronously,
and a `DispatchQueue.main.async` nudge. What worked: `ShelfTileView`, the
tile being dropped onto, calls `rebuildTiles` directly with the service's
own current state, bypassing SwiftUI's reconciliation for the one thing that
has to be immediate.

Calling `rebuildTiles` matters for scope reasons: it tears down every tile
in the Shelf, not just the one that changed, since that's the same
mechanism every other rebuild already uses. Doing that from inside
`performDragOperation` would mean tearing the current tile, `self`, out of
the view hierarchy while a method on it is still on the call stack. That's
harmless today (the only other AppKit callback afterward,
`concludeDragOperation`, is a no-op on the detached view) but fragile
against any future change. `performDragOperation` now only records that a
rebuild is owed; `concludeDragOperation`, the actual last callback AppKit
makes on this view for the drag, performs it, at the same point in the
drag's lifecycle without mutating the hierarchy out from under a still-
running method.

## Try it yourself

Drop a file onto an existing Shelf tile so it becomes (or grows) a pile. The
tile should immediately show pile chrome, the count badge and expand
chevron, with no click needed first.

## Before / after

| Situation | Before | After |
|---|---|---|
| Drop a file onto an existing tile | Merge succeeds, tile stays visually stale until something else redraws it | Tile updates immediately |

## Test plan

- [x] `./build.sh --test`: `TESTS OK (7629 checks)`
- [x] `./build.sh --dev`: zero errors, zero warnings
- [x] Manual: repeated merge-onto-tile drops, from outside the app and from
      dragging one Shelf tile onto another, tile updates instantly both ways

## Notes

No new dependencies, no user-facing strings, no permission changes. Pure
AppKit timing fix, no behavior change to what a merge actually does.


Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
```diff
@@ -492,11 +507,30 @@ final class ShelfTileView: NSView, NSDraggingSource {
...
+        // own updateNSView observably lags behind it while this AppKit drag
+        // session is still unwinding, leaving the merged tile visually stale
+        // until some unrelated later event forces a redraw. Rebuilding
+        // directly here, with the service's own current state, bypasses that
+        // lag for the one thing that has to be immediate: showing the drop
+        // actually worked. Deferred to here rather than done inline in
+        // performDragOperation: rebuildTiles tears down every tile,
+        // including self, and concludeDragOperation is the last AppKit
+        // callback expected on self for this drag, so self is safe to
+        // remove from the hierarchy by the time this runs.
```
vorssaint: hey, can we make only this comment a bit more concise? i think it’s a little too long and technical for what we need

the rest LGTM!

PR #694, opened 2026-08-18
feat(uninstaller): add Uninstall Application to the Command Bar

naveenkrdy:
## Summary
- Adds a per-feature "Show in Command Bar" toggle to the Uninstaller, off the existing pattern from `killProcessCommandBarEnabled`/`KillProcessView`
- A dedicated "Uninstall Application" row drills into a filterable list of installed apps, reusing the bar's existing category-browse mechanism (the same one that drives the "Emoji" row) rather than adding per-app rows to the flat search pool
- A contextual "Uninstall `<App>`" row appears when exactly one non-system app inside `/Applications` (or `~/Applications`, at any depth) is selected in Finder
- Selecting either row opens the full leftover-files checklist inline in the bar (categories, sizes, per-item toggles) and removes only what's checked, without leaving the bar
- Homebrew-managed apps get an inline confirm step and the same live-progress view (`HomebrewOperationStatusView`) the menu panel already shows, so removal never has to jump out to Settings
- The done screen now responds to Return (previously a no-op there) and always clears the field instead of restoring whatever was typed before the uninstall started
- `Services/Finder/FinderCutPaste.swift`'s `FinderBridge` changes from `private enum` to `enum`, needed so `CommandBarService`'s Finder-selection loader can call `FinderBridge.selectionURLs()` across files - feature-agnostic, not Quarantine-specific

## Test plan
- [x] `./build.sh` - clean, no warnings
- [x] `./build/Vorssaint --selftest` - `SELFTEST OK`
- [x] `./build.sh --test` - `TESTS OK`
- [x] `./build.sh --install` - installed and signed into `/Applications/Vorssaint.app`
- [x] Typed search: enable the toggle, type "Uninstall" → select "Uninstall Application" → confirm the app list appears with the field cleared → pick an app → review checklist → remove → done screen responds to Return with the field cleared
- [x] Finder selection: select one non-system `.app` in `/Applications` → confirm the contextual row appears → negative-test a non-app file, multiple selection, an app outside Applications, and a system app (none should show the row)
- [x] Homebrew-managed app: confirm the inline confirm card, live progress, and completion all happen without leaving the bar
- [x] Mid-review dismissal: opened the checklist, closed the bar with the global shortcut instead of stepping back out of it, reopened Settings → Uninstaller and the menu panel - confirmed neither shows a leftover scanned checklist. This holds only before Remove is pressed: closing the bar *during* an in-flight plain removal deliberately does not reset now (the earlier fix in this PR), so the next open of Settings › Uninstaller or the menu panel shows that removal's done screen instead of a blank state. Intentional - a result you can find beats one silently dropped - but worth knowing going in.

[See it in Action](https://pub-d1f174c205bf431e9aa4eb83be698fc8.r2.dev/snapzy/1787237602-7f7347e0-snapzy_recording_2026-08-20_20-22-25.mp4)


PathGao:
Confirmed on my side — this is resolved.

The PR is now +607/-15 across 22 files (was +2585/-35 / 36 files), three commits, and there are no `QuarantineManager/` paths left in the tree. That matches your description exactly: the ~2000 lines that disappeared are #674's, so the branch really was carrying its commit rather than just following its pattern. The Finder-selection loader is self-contained now, same shape as `killProcess`.

Nothing further from me here. Only leftover is that it's `BEHIND` main, so it'll want a rebase before merge — no conflicts.

PathGao:
**The stated blocker is stale — this does not need #674.** `main` already has the per-feature Command Bar toggle pattern from #683: `killProcessCommandBarEnabled` at `Defaults.swift:186` and `:903`, `CommandBarCatalog.swift:681`, `CommandBarService.swift:2126`, `KillProcessView.swift:13`. This PR already uses exactly that pattern, so it can be reviewed and merged on its own.

On the code, two things and one note.

**Two exit paths from `.uninstallReview` disagree.** `stepBack()` (~`CommandBarService.swift:1912`) calls `AppUninstaller.shared.reset()`. `dismiss()` (`:360-389`) sets `mode = .search` and clears `uninstallSelectionEntries` but never resets the uninstaller. Close the bar with the global shortcut mid-review and `AppUninstaller.shared` keeps a selected target and a scanned checklist, which then appears unprompted in Settings › Uninstaller and in the menu panel.

**A stale doc comment says the opposite of the code.** On `uninstallReviewCard` in `CommandBarView.swift`: "Homebrew-managed apps are not handled here: pressing Remove on one hands off to Settings, where the confirmation and live progress already live." The code renders `HomebrewOperationStatusView` inline and `confirmUninstallHomebrewRemoval` calls `AppUninstaller.shared.removeSelectedWithHomebrew()` in the bar. Left over from an earlier revision.

**The safety story is better than the summary suggests, with one asymmetry.** The row does not delete — it opens the scan checklist with every file, its size and a per-item checkbox behind a red Remove button; removal goes through `AppUninstaller.removeSelected()` and `fm.trashItem`, guarded by `removalIsStillSafe(_:allowedPaths:targetURL:)` and the `exclusiveOwnedBundleIDs` ownership check, and `confirmUninstallReview` no-ops while `phase != .results`. Wrong-app risk is low: `uninstallEntries` is deliberately kept out of the ranking pool (`:1214`) and only lives in `indexableEntries`, matching how `quitEntries` and `killProcessEntries` are already treated. The asymmetry is that once results render, a single Return removes a plain app, while a Homebrew one gets a second confirmation step.

Smaller:

- `Services/Finder/FinderCutPaste.swift` changes `private enum FinderBridge` to `enum FinderBridge`. Needed by `loadUninstallSelectionEntries` and called out in `562605d8`, but the description does not mention it.
- The only test change is the source-id list (+2/−2). `InstalledApps.isInApplicationsFolder` and the `uninstallApps` branches of `sourcePriority` / `acceptsAlias` / `acceptsPin` are all inside `build.sh`'s `--test` allowlist and got no assertions, in a file that already has four `acceptsPin`/`acceptsAlias` tests.
- The new toggle sits in `UninstallerView.emptyState`, so it disappears once an app is selected. `KillProcessView.swift:13` puts the equivalent one in the always-visible header.

PathGao:
Two updates on top of the review above.

**The rebase is no longer conflict-free, and the one conflict is not mechanical.** `Services/CommandBar/CommandBarService.swift` collides in `refresh`: `main` now ends the `.search` branch with `setCompactHome(CommandBarHome.isCollapsed(compact:query:hasCategory:isPeeking:))` at `:958-961` and gives every other mode `setCompactHome(false)` at `:962-963`, where this branch still has `case .argument, .confirm, .uninstallReview, .uninstallHomebrewConfirm, .actions, .naming, .capturingShortcut: break`.

Keeping both sides duplicates `.argument` and will not compile. The resolution is to take `main`'s shape and add the two new modes to its non-search arm:

```swift
case .argument, .confirm, .uninstallReview, .uninstallHomebrewConfirm, .actions, .naming,
     .capturingShortcut:
    setCompactHome(false)
```

Worth checking `:883` at the same time — `setCompactHome(compactMode)` is a second site of the same mechanism that landed with it.

**`isInApplicationsFolder` and `installedApplications` disagree about what an installed app is.** `InstalledApps.swift:57-68` requires the bundle's parent directory to be exactly `/Applications` or `~/Applications`. `installedApplications` at `:70-89` walks those same roots with `fm.enumerator` and only `.skipsPackageDescendants`, so it accepts nested bundles — `/Applications/Adobe Acrobat DC/Acrobat.app` and anything else a vendor installs into its own folder — and then adds whatever `applicationScanPaths` finds via Spotlight on top.

The result is that the same app is offered in the "Uninstall Application" browse list but refused when it is the app selected in Finder, with no message explaining why. Your doc comment reads as deliberate, so if it is, say which of the two definitions is the intended one and make the other match. If the intent was just "an installed app", `hasPrefix(root + "/")` against the two roots is the smaller change and lines up with the enumerator.

naveenkrdy:
Thanks - the merge conflict, the app-detection mismatch, and both correctness issues are fixed.

**Merge conflict.** Took `main`'s shape exactly as suggested - folded `.uninstallReview`/`.uninstallHomebrewConfirm` into the non-search arm's `setCompactHome(false)`. Checked `:883`'s `setCompactHome(compactMode)` too - that one's in the unconditional `isLoadingHome` early return, not the mode switch, so it didn't need the same change.

**`isInApplicationsFolder` vs `installedApplications`.** Not deliberate - went with your smaller-change read: `hasPrefix(root + "/")` against both roots, so a bundle a vendor installs into its own subfolder now counts the same way whether it's reached through the browse list or through a Finder selection.

**Two exit paths from `.uninstallReview`.** `hide()` now resets `AppUninstaller` the same way `stepBack()` does for that mode, guarded the same way too (leaves an in-flight Homebrew removal alone). Verified by hand: open the checklist, close with the global shortcut instead of stepping back out, and neither Settings nor the menu panel shows a leftover scan afterward.

**Stale doc comment.** Fixed - it now says what the code does (handled inline, not handed off).

Smaller three:
- Added the `FinderBridge` access-level change to the PR description.
- Added the missing coverage: `rankBias`/`acceptsAlias`/`acceptsPin`'s `uninstallApps` branches, and `isInApplicationsFolder` including the nested-subfolder case the fix above is for.
- Left the toggle placement as-is for this pass - moving it to an always-visible header like `KillProcessView`'s would mean restructuring `UninstallerView`'s per-phase layout, and I don't have a way to visually verify that change in this environment. Can take it on as a small follow-up if you'd rather not carry the inconsistency.

Also rebased onto current `main` - it had drifted further behind since your last check.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8643 checks).

PathGao:
All four land as described — the `refresh` arm now matches `main`'s shape, `isInApplicationsFolder` uses `hasPrefix(root + "/")`, `hide()` resets, and the doc comment says what the code does. Your `:883` reading is right, that one is in the unconditional early return and shouldn't change.

One new thing, same shape as the `hide()` bug you just fixed: **the reset guard covers a Homebrew removal in flight but not a plain one.** Both `hide()` and `stepBack()` skip the reset only when `isRemovingWithHomebrew`. `phase == .removing` is reachable for a plain app too — that's what `uninstallReviewBusy(l10n.s.uninstallerRemoving)` renders — and nothing blocks Esc or the global shortcut during it, so `AppUninstaller.reset()` can run while `removeSelected()` is still going. If the completion then writes `.done` onto a reset uninstaller, that's the same leftover screen in Settings and the menu panel, reached one keystroke earlier.

`AppUninstaller.swift` isn't in the diff, so I can't tell whether `reset()` mid-removal actually hurts — if `removeSelected()` iterates a captured copy and `reset()` is harmless there, this is nothing. The window is short either way. If you do want it closed, at both sites:

```swift
case .uninstallReview, .uninstallHomebrewConfirm:
    let uninstaller = AppUninstaller.shared
    if case .removing = uninstaller.phase { break }
    if !uninstaller.isRemovingWithHomebrew { uninstaller.reset() }
```

Pattern match rather than `phase != .removing`, since `.done` carries `[Leftover]` and the enum may not be `Equatable`.

Nothing else in the new diff. The added assertions cover the branches that were bare.

**Impact:** closing the bar during a plain removal may leave `AppUninstaller` in a state that surfaces unprompted in Settings — narrow window, and I could not verify it without `AppUninstaller.swift`.
**Merge:** ship it.

naveenkrdy:
Read `AppUninstaller.swift` to check - your specific worry doesn't actually happen, but there's a related gap worth closing anyway.

`removeSelected()` captures its own working state (`chosen`, `allowedPaths`, `targetURL`, etc.) into locals before dispatching to the background queue, so a `reset()` racing in on the main thread was never going to corrupt the removal itself. And the completion already guards `self.phase == .removing` before writing `.done` - so if `reset()` flips `phase` to `.empty` mid-removal, the completion just no-ops instead of writing `.done` onto a reset instance. No leftover-screen bug.

What actually happens instead: the removal still runs to completion (files still move to Trash), but the result gets silently dropped - freed bytes and any failures never make it into `.done`, so nobody ever sees them. Closed that too, same guard shape as the Homebrew case, in both `hide()` and `stepBack()`.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8643 checks).

PathGao:
Both checks are red as I see them right now — "Build & selftest" and "Swift 6.0.3 compatibility". That's why `ready to merge` came off. I can't read the log, but I can narrow it: it isn't a missing string or a missing symbol. All 13 `Strings` initializers got the three new fields (the 11 files under `Core/Localizations/` plus the two inline in `Localization.swift`), `bar.uninstallAppFormat` already exists at `CommandBarStrings.swift:136`, and `Self.listCeiling`, `runSelected()`, `setCategory(_:)`, `HomebrewOperationStatusView`, `HomebrewTrustCard` and `Category.allCases` are all on `main` with the shapes you call. `Phase` is `Equatable` at `AppUninstaller.swift:15`, so `phase != .removing` compiles. Your local run green and both CI jobs red is the thing to reconcile.

Your reading of `removeSelected()` is right — `:213` guards `self.phase == .removing` before writing `.done`, and everything the background work touches is captured into locals at `:161-164` first.

The new guard has a consequence worth naming, because it's the same leftover screen from the other side. Close the bar during a plain removal: `hide()` now skips the reset, the background work finishes, the completion's `phase == .removing` check passes and writes `.done(freed:failed:)` onto a live uninstaller with `target` still set. `UninstallerView.content` and `PanelUninstallerView.content` both switch on `phase`, so whichever is opened next shows a done screen for a removal the person already walked away from, and it stays until they press Done there or select another app. **The only way the preserved result becomes visible is as the leftover screen the reset exists to prevent.**

The window isn't narrow, either — if anything lands in `stubborn`, `trashViaFinder` puts up the administrator prompt and `.removing` lasts as long as it takes to type a password.

I'd leave the code alone; a result you can find in Settings beats one that's silently dropped. But then the invariant in your test plan ("closed the bar with the global shortcut … confirmed neither shows a leftover scanned checklist") only holds before Remove is pressed, and that's worth a line in the PR body so the next person doesn't file it as the bug you just fixed.

**Impact:** closing the bar during a plain removal leaves a done screen waiting in Settings › Uninstaller and the menu panel — deliberate under the new guard, but stated nowhere.
**Merge:** needs the CI failure explained or fixed first.

naveenkrdy:
Both red checks were the same root cause: the CI runner builds this PR's synthetic merge with `main`, not the raw branch, and `main` had since added a `.uninstallApps` case to `CommandBarSource` that this branch's `isHubOwned(_:)` switch didn't know about - a non-exhaustive switch, not a missing string or symbol. `acceptsAlias`/`acceptsPin` (the other two switches over `CommandBarSource` in the file) already handle it correctly, both treating it as `false` - `.uninstallApps` rows are "never in the flat search pool" per their own doc comment, so neither aliasing nor pinning applies to them. Added it to `isHubOwned` the same way, in the not-hub-owned group alongside `.apps`/`.menus`: it reflects real installed-app state on disk, not a Vorssaint feature that can be toggled off.

Merged `main` in properly (not just diagnosed against it) so this stays in sync going forward. Both checks are green on d33279f now.

Also added the line you asked for on the leftover-done-screen behavior - it's in the "Mid-review dismissal" test-plan bullet in the PR body now, scoped to "before Remove is pressed."

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8714 checks).

PathGao:
Green, and `isHubOwned` was the last one — that's all six switches over `CommandBarSource` in `CommandBarPreferences.swift` covered (`icon`, `prefix`, `rankBias`, `acceptsAlias`, `acceptsPin`, `isHubOwned`), and every other one in the tree is compiler-enforced. Your not-hub-owned answer is also inert either way: `isHubOwned` is read only by `listedPins`, and `acceptsPin` already refuses every `uninstall.` id, so no pin with that prefix can exist to be listed.

One thing from `main` the diff can't show: **a third "Uninstall &lt;App&gt;" door still jumps to Settings.** `actions(for:)` at `CommandBarService.swift:1465-1471` hangs a `uninstallApp` row action off every `app.*` entry, titled from the same `bar.uninstallAppFormat` your Finder-selection row uses, and its one caller path `openUninstaller(for:)` (`:1728-1733`) does `hide()` → `select` → open Settings. Type "Chrome", ⌘K, "Uninstall Google Chrome" → Settings window. Select that same app in Finder and open the bar → inline checklist. Identical words, two destinations, and the ⌘K one is what someone who typed the app's name reaches first. It also narrows the summary's "removal never has to jump out to Settings" to the two doors you added.

Not a blocker — that path went to Settings before this PR and nothing here made it worse. If you want it closed, `openUninstaller`'s call site has `entry` in scope, so it can set `mode = .uninstallReview(entryID: entry.id)` instead of hiding, gated on `uninstallerCommandBarEnabled` so it still falls back to Settings when the toggle is off. I haven't checked how `refreshPanelLayout` behaves entering that mode from `.actions`.

**Impact:** none new from this diff. Pre-existing: the ⌘K uninstall action on an app row leaves the bar while the identically-titled Finder-selection row stays in it.
**Merge:** ship it.

naveenkrdy:
Good catch on the third door, and agreed it's pre-existing rather than something this PR needs to fix - the ⌘K path went straight to Settings before any of this landed. Not folding it into this PR since it's a separate change to CommandBarService's action dispatch, not the uninstaller work this one scopes to, but I'll open a follow-up for the `mode = .uninstallReview(entryID:)` redirect you sketched, gated on `uninstallerCommandBarEnabled` the way you said. Thanks for flagging it rather than letting it sit as an inconsistency nobody named.

andreisuslov:
@PathGao if it's ready to merge, why don't you start with the oldest PR's instead of always taking the newest ones?

Issue #341, opened 2026-07-22
Volume Mixer leaves Firefox permanently muted after macOS wakes from sleep


ukokoshu:
### Description

  Description:

  I am experiencing an issue where Vorssaint’s Volume Mixer sometimes leaves Firefox
  completely muted after macOS wakes from sleep.

  ### Environment

  - macOS 26.5.2
  - Apple Silicon
  - Vorssaint 3.1.15
  - Firefox 152.0.6
  - Audio output: Universal Audio Volt 176 via USB
  - Sample rate: 48 kHz

  ### Steps to reproduce

  1. Enable the Volume Mixer in Vorssaint.
  2. Change the Firefox volume in Vorssaint. In my case, it was set to approximately
     92%.

  3. Play audio in Firefox.
  4. Allow the display or Mac to sleep.
  5. Wake the Mac.

  ### Actual result

  Firefox sometimes becomes completely silent after waking.

  Other audio applications continue working. Changing the Firefox tab, pausing and
  resuming playback, or checking the system output device does not restore the sound.

  The CoreAudio log shows Vorssaint creating a process tap that mutes Firefox:

  HALS_Client::AddMuter: Process org.mozilla.firefox muted by [tap UUID] for all devices

  For the affected tap, there is no matching RemoveMuter event after the wake/
  reconfiguration sequence.

  As soon as I quit Vorssaint, CoreAudio removes the same stale muter:

  HALS_Client::RemoveMuter: Process org.mozilla.firefox unmuted by [same tap UUID]

  Firefox audio returns immediately without restarting Firefox.

  The issue occurs while the USB audio interface and displays are being reconfigured
  after wake. The Vorssaint log also contains CoreAudio errors during this sequence,
  including:

  AudioObjectRemovePropertyListener: no object with given ID
  HALC_ProxyObject::SetPropertyData: got an error from the server
  HALC_ProxyIOContext::IOWorkLoop: received an out of order message

  ### Expected result

  When the Mac wakes or an audio device is reconfigured, Vorssaint should either:

  - keep the existing mixer tap working correctly; or
  - destroy every old process tap and aggregate device before creating replacements.

  Firefox must never remain muted if the corresponding Vorssaint rendering engine is no
  longer active.

  ### Workaround

  Disabling only the Volume Mixer feature fixes the problem:

  featureAvailable.mixer = false

  Vorssaint can remain running, and all its other features continue working. Re-enabling
[...]

PR #704, opened 2026-08-20
feat(audio): add one-click Core Audio recovery

audip:
## Summary
- add a localized Reset Core Audio control to the current Mixer section
- reuse Vorssaint’s serialized AdminShell prompt and run the audited absolute `/usr/bin/killall coreaudiod` command only on explicit user action
- show in-panel progress and success/failure feedback without storing credentials
- keep the pure command coverage and documentation

## Permission model
Restarting `coreaudiod` is a system-level operation, so macOS may require administrator authentication. The prompt appears only after the user clicks Reset audio; credentials are not stored.

## Relationship with #813
This PR and #813 are independent audio features. PR 704 performs an explicit daemon recovery action and owns no audio monitoring or playback state. PR 813 observes Core Audio activity and controls the selected music player through Automation. If both land, a reset may temporarily interrupt audio observations, but neither feature depends on the other and PR 813 can resume on its next evaluation.

## Merge order
Merge PR 704 before PR 821. PR 704 adds fields to the existing `Strings` contract, while PR 821 introduces the Vietnamese localization whose completeness is compiler-enforced. Vietnamese translations for fields from PR 704 and the other pending localization PRs should be added as part of PR 821 when it lands last.

## Validation
- rebased onto current `main`
- `./build.sh --test` — TESTS OK (8,390 checks)
- `./build.sh` — release bundle assembled and ad-hoc signature verified
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ The rest bends the same way: panel sections reorder and hide, the compact layout
 - **Per app output.** Send your music to the speakers and a call to your headset at the same time.
 - **Output switcher.** Cycle between chosen outputs with one shortcut, and drop the volume automatically when headphones disconnect.
 - **Microphone tools.** Pin your favorite input so the Mac stops guessing, and mute every microphone at once with a click or shortcut, whichever one an app is using.
+- **Audio recovery.** Restart macOS Core Audio with one click when sound crackles, stutters or disappears. Vorssaint asks for administrator access only when you use it and stores no credentials.
 - **Music app blocker.** Stops the Music app from bursting in when headphones connect. You can still open it yourself.
 
 ### Know what your Mac is doing
--- a/Sources/Vorssaint/Core/Localization.swift
+++ b/Sources/Vorssaint/Core/Localization.swift
@@ -648,6 +648,13 @@ struct Strings {
 
     // MARK: Panel — volume mixer
     let mixerSection: String
+    let audioRecoveryTitle: String
+    let audioRecoveryCaption: String
+    let audioRecoveryButton: String
+    let audioRecoveryResetting: String
+    let audioRecoverySuccess: String
+    let audioRecoveryFailure: String
+    let audioRecoveryAdminPrompt: String
     let mixerEmpty: String
     let mixerUnavailable: String
     let mixerPermissionBody: String
@@ -1626,6 +1633,13 @@ extension Strings {
         breakdownMeasuring: "Medindo…",
 
         mixerSection: "Mixer de volume",
+        audioRecoveryTitle: "Reiniciar o Core Audio",
+        audioRecoveryCaption: "Reinicia o serviço de áudio do macOS quando o som estala, falha ou desaparece.",
+        audioRecoveryButton: "Reiniciar áudio",
+        audioRecoveryResetting: "Reiniciando…",
+        audioRecoverySuccess: "Core Audio reiniciado.",
+        audioRecoveryFailure: "Não foi possível reiniciar o Core Audio.",
+        audioRecoveryAdminPrompt: "O Vorssaint reiniciará o Core Audio do macOS para recuperar o som. Uma senha de administrador pode ser necessária.",
         mixerEmpty: "Apps que usam áudio aparecem aqui",
         mixerUnavailable: "Disponível a partir do macOS 14.4",
         mixerPermissionBody: "Para ajustar o volume por app, permita “Gravação de Tela e Áudio do Sistema” nos Ajustes do Sistema. O áudio nunca é gravado.",
@@ -2586,6 +2600,13 @@ extension Strings {
         breakdownMeasuring: "Measuring…",
 
         mixerSection: "Volume mixer",
+        audioRecoveryTitle: "Reset Core Audio",
+        audioRecoveryCaption: "Restarts macOS’s audio service when sound is crackling, stuttering or missing.",
+        audioRecoveryButton: "Reset audio",
+        audioRecoveryResetting: "Resetting…",
+        audioRecoverySuccess: "Core Audio restarted.",
+        audioRecoveryFailure: "Could not restart Core Audio.",
+        audioRecoveryAdminPrompt: "Vorssaint will restart macOS Core Audio to recover sound. An administrator password may be required.",
         mixerEmpty: "Apps that use audio show up here",
         mixerUnavailable: "Available on macOS 14.4 and later",
         mixerPermissionBody: "To adjust per-app volume, allow “Screen & System Audio Recording” in System Settings. Audio is never recorded.",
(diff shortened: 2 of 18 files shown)


audip:
@PathGao @nvsinha can you review this again?

PathGao:
Read the diff against current `main`. The shape is right — pure support enum wired into the `--test` list, `AdminShell` reused instead of a new privileged path, all 13 string catalogs filled. Five things I would want settled first.

**1. The mixer's own taps do not survive the restart.** `AppVolumeMixer` holds process taps and aggregate engines built on HAL objects that a `coreaudiod` restart destroys. `main` already treats this class of event as needing deterministic arming rather than passive detection: the wake handler (`AppVolumeMixer.swift:188-204`) drops `engineRenderProgress`, clears `engineRecovery`, refreshes, reconciles, and schedules a recheck — precisely because the stalled-render detector from #341 is not prompt enough on its own. Killing `coreaudiod` is the same event with a sharper edge. As written, the panel says "Core Audio restarted." while per-app volume may still be wedged, which is the opposite of what the button promises. Either arm the same sequence after a successful reset, or show that per-app volume comes back on its own.

**2. `false` from `AdminShell` is not "could not restart".** `ShellSupport.swift:36-39` is explicit: `runSync` returns false when a prompt is already up, with no dialog shown at all, and "every caller already treats false as permission not granted". So a cancelled password dialog — or a click while another Vorssaint prompt is open — now renders "Could not restart Core Audio.", an error for something that never ran. `KeepAwakeManager.swift:561` is the established handling: act on true, say nothing on false. A third state, or no message on cancel.

Related: `killall` exiting 0 means the signal was delivered, not that the daemon is back, so the success line is asserting more than it knows. `launchctl kickstart -k system/com.apple.audio.coreaudiod` (label verified on macOS 26.5) restarts under launchd supervision instead of relying on respawn-after-kill.

**3. `lastResult` is never cleared.** Nothing resets it, so "Core Audio restarted." sits in the Mixer section until the next reset or a relaunch — open the panel two days later and it still reports it. The sibling message in the same view, `mixer.outputSwitchError`, is cleared on success and again when the device list changes (`AppVolumeMixer.swift:245`, `:496`, `:831`).

**4. No Features hub entry.** `FeatureCatalog.swift:6` — "Every feature the Features hub can switch off entirely" — and the sound group already carries `mixer, soundOutputSwitcher, micMute, musicBlock` (`:25`). This adds a user-facing control that escalates privileges, with no way to switch it off. I would expect `case audioRecovery` in that group plus its catalog metadata and hub strings.

**5. Placement and the doubled caption.** `audioRecoveryControl` lands as the first row of `audioDevicesSection`, so a rare recovery action sits above the output, system-sound and microphone pickers people use daily; after `microphonePicker` reads better. And `audioRecoveryCaption` serves as both the visible caption and the `.help()` tooltip, where its siblings use dedicated tooltip strings (`mixerSystemOutputTooltip`, `mixerSoundEffectsOutputTooltip`, `mixerInputTooltip`).

One process note, not an implementation point: I could not find an issue asking for this, and it would be the first control in the panel that escalates privileges on a click. Whether it belongs there is the maintainer's call — everything above is about the implementation if it does.

PathGao:
One more thing, and it is a standing point rather than something about this PR alone: we are not taking new-feature PRs that have not been run on real hardware. `./build.sh --test` and a signed bundle prove the code compiles and the pure helpers hold — they say nothing about whether the feature works.

So before the next push, run it on a real Mac and put what you saw in the PR body:

- Per-app volume across a reset. Play audio in two apps, pull one to 40%, hit Reset audio, then report whether the sliders still do anything and how long it took to come back. This is point 1 above and it is the one that decides whether the PR is correct.
- Cancel the password dialog and say what the panel shows.
- Whether sound is actually back at the moment "Core Audio restarted." appears.
- Input pinning, mic mute and the output switcher after a reset — they hold HAL state too.

A short screen recording, or screenshots plus your macOS version, is enough.

PathGao:
Three more, from re-reading against current `main` (75ce1840). None of these change my earlier five.

**1. The panel is gone before the result message can be read.** `MixerSection` never sets `PanelInteractionState.shared.keepsPopoverOpen`, and `AppDelegate.swift:726-731` installs a *global* mouse monitor that closes the popover on any `leftMouseDown`/`rightMouseDown` delivered to another application. SecurityAgent's password dialog is another application, so the first click in it — the click into the password field — dismisses the panel. `lastResult` then has nothing to render into: `audioRecoverySuccess` / `audioRecoveryFailure` are only visible if the user reopens the panel afterwards, which also makes my point 3 (never cleared) worse rather than better.

The established handling is `QuickTogglesSection.swift:26-28`, whose comment is this exact case: "Actions here activate other apps ... which would dismiss the panel mid-flow; holding it open keeps a toggle-back one click away." Hold `keepsPopoverOpen = true` from `resetAudio()` until the completion fires. I could not run this, so the claim is from the monitor's code path, not from watching it happen — but it is worth checking before anything else here, because it decides whether the in-panel feedback design is reachable at all.

**2. `isResetting` has no bound.** `AdminShell.runSync` passes `timeout: 600` to `Shell.run` (`ShellSupport.swift:71`). A password dialog left on screen keeps `isResetting` true — button disabled, "Resetting…" — for up to ten minutes, with no way to get out of it. If the panel does stay open (point 1), that is the state the user sits in.

**3. Try the passwordless path first.** You asked whether a non-privileged route exists: for the daemon itself, no — `coreaudiod` runs as `_coreaudiod` and `killall` needs root either way, so the prompt is not avoidable for what the button currently does. But the app already owns a password-free recovery for the part of the audio stack it is actually responsible for: `AppVolumeMixer.stopAll()` (`AppVolumeMixer.swift:209-220`) tears every process tap down and returns each app to untouched system output. For a Vorssaint user whose sound went wrong, a wedged mixer engine is at least as likely a cause as a wedged daemon, and that path costs no prompt and no system-wide interruption. Running it first, and escalating to the daemon restart only if audio is still wrong, would make the control both safer and more honest about what it did.

Related, on scope: `KillProcessSupport.isProtected` (`KillProcessSupport.swift:11-24`) protects only `kernel_task`, `launchd`, `WindowServer`, `loginwindow` and this app, so Kill Process can already terminate `coreaudiod` today through the same `AdminShell` prompt (`KillProcessService.swift:207`). This PR is the app's second privileged-kill path. That does not make the one-click affordance worthless, but it is worth saying in the description why the general one is not enough.

PathGao:
If you really want it，consider it again， rework it as a new pr。But I dont think this very niche needs shoud join our product。should be a local sh file

PathGao:
hey，I checked this pr again。with some miner change of command bar，plus sudo or privilege will turn this pr into a custom command shortcut in command bar script。lets figure it again。

PR #713, opened 2026-08-20
Add Selection Actions floating menu feature

naveenkrdy:
## Summary

Selecting text anywhere on the Mac now surfaces a small floating bar of quick actions next to the selection.

- **Actions**: copy/cut/paste (plain or rich), delete, search the web, open link/mail, translate, case and format transforms (uppercase/lowercase/capitalize, remove spaces, underscore, join lines, comma list, sort/reverse/shuffle - all three fall back from lines to words to letters for a single word), quotes/brackets, URL/Base64 encode-decode, calculate, convert currency, run in Terminal, send to an AI chat (ChatGPT or Claude, with a temporary-chat option), and add to Scratchpad.
- **Settings**: per-action enable/disable and drag reordering, icon or word display, overflow packing past a configurable count, exclude specific apps or websites, and a global keyboard shortcut to summon it without a fresh selection.
- **Reading a selection**: layers Accessibility (fast path for well-behaved apps) with a synthetic-copy fallback for apps that don't answer the standard selected-text query (verified against VS Code and Mail), guarded against triggering on unrelated UI (window drags, list/outline views) and against surfacing a stale selection left over from before the click.
- **Rewriting a selection**: a shared pasteboard-swap-and-synthetic-paste primitive, extracted from the existing Paste Plain feature so both share one hardened implementation; Cut and Delete use a synthetic keystroke instead, which proved more reliable across browser chrome and native fields than pasting an empty string.

Ships as an opt-in beta feature (`AppFeature.selectionActions`), off by default like Fan Control, gated on Accessibility (and Automation for Run in Terminal).

## Test plan

- [x] `./build.sh --test` passes (`TESTS OK`, 7824 checks)
- [x] Manually exercised across native apps, Safari/Chrome/Edge, VS Code, Mail, and the app's own Scratchpad: selecting and each action, the no-selection Paste bar, drag reordering, display style, overflow packing, app/website exclusions, and the global shortcut

Snapzy_2026-08-20_19-54-42_869

[See it in Action how it works](https://pub-d1f174c205bf431e9aa4eb83be698fc8.r2.dev/snapzy/1787235539-1754ea8b-snapzy_recording_2026-08-20_19-47-09.mp4)


PathGao:
Checked your network claim and it holds precisely: the only `URLSession.dataTask` in the whole diff is in `CurrencyConversionSupport.swift` (frankfurter.dev). `sendToAI`, `searchWeb` and `translate` do build a URL and hand it to `NSWorkspace.shared.open`, same as `openLink`/`openMail`.

Three things on top of that:

**1. `Run in Terminal` isn't in your list.** `selectionActionsTerminalConfirm` and `selectionActionsTerminalTarget` are both in `Defaults.swift`. Selected text one click away from executing in a shell is the highest-risk action in the set — above `convertCurrency` by a good margin. Defaulting confirm to `true` is the right call, but when we're enumerating what the 30 actions do, that one should be named.

**2. A browser tab doesn't mean the text stayed local.** `sendToAI` puts the selection into a `chatgpt.com` / `claude.ai/new` query string. That changes who makes the request, not whether the text leaves the machine — and query strings end up in browser history.

**3. My concern was the permission surface, not the network.** Continuously reading selected text out of any app (`SelectionReader`, plus the synthetic-copy fallback in `ShadowCopySelectionReader`) and `BrowserURLReader` reading the current URL — that's what "largest permission expansion in the repo" was pointing at, and it's the part still open. Whether that ships as one PR or two matters less to me than that it gets reviewed as a permission change rather than as a menu of features.

One more thing: you undersold your own case. `selectionActionsEnabled` defaults to `false`, and you already built both `SelectionActionsExcludedApps` and `SelectionActionsExcludedDomains`. Those three facts argue for keeping this as one PR more effectively than the single-network-call point does — they're worth putting in the PR description.

PathGao:
My recommendation is to close this and reopen it as four smaller PRs, and I want to be exact about which part of that is a verdict on the work: none of it. Closing is @vorssaint's call, not mine — this is the ask, with what I would do first.

## The part that should not wait for the rest

**`SyntheticPasteSupport.swift` is finished, it merges without this feature, and the repo needs it whether Selection Actions ever ships or not.** Open it as its own PR and it can go in this week.

`git grep kVK_ANSI_V` on `main` finds three independent synthetic-⌘V implementations, and they have already drifted:

| | modifier wait | event source | down/up beat | `IsSecureEventInputEnabled` guard |
|---|---|---|---|---|
| `PastePlainService.swift:251` | yes | `nil` | 0.04s | no |
| `CommandBarService.swift:1913` | yes | `.hidSystemState` | none | yes |
| `ClipboardHistoryService.swift:1013` | **no** | `.hidSystemState` | none | no |

Two of those columns come with a stated reason in your version and nowhere else: an event tied to the HID state "can re-poison the flags `waitForCleanModifiers` just waited out", and "some apps skip equivalents delivered as a zero-length tap". Whether the other two sites need those properties depends on how each is triggered, and that is worth settling — but it cannot be settled while the answer is written three times. Today, fixing one copy fixes none of the others and nothing requires them to agree.

It is about to be four. #917 adds a fresh paste path in `TextSnippetService.swift` (+56/-4) that joins none of the four things `PastePlainService` does — no `ignoreNextChange`, unconditional restore, no re-entrancy state, no wait for a clean keyboard. Land your primitive and that PR becomes a call site instead of a fourth implementation.

Two notes for when it stands alone:

- **It is a real extraction, not a re-export.** `PastePlainService` loses its private `snapshot` and `postPasteShortcut` outright and keeps only its own specialization, unregistering the ⌘V hotkey around the post. No second bridge, which is the failure mode I would have looked for first.
- **Fold in the one thing it drops.** `CommandBarService.swift:1927` refuses to post under `IsSecureEventInputEnabled()`; `SyntheticPasteSupport` has no such guard. Of the three, that is the only hardening the shared version loses.

Converting the other two call sites is a second PR, not this one's job.

## The rest

The remaining ~30 actions are the largest permission expansion in the repo, and 34 files is not a size anyone reviews as a permission change — it is a size that gets skimmed. Suggested split:

1. **`SyntheticPasteSupport` alone.** No feature, no registration, no strings. Reviewable in one sitting, useful immediately.
2. **The feature skeleton and the permission it costs.** `SelectionReader`, `ShadowCopySelectionReader`, `BrowserURLReader`, the bar and its controller, the exclusion lists, the settings page, the `AppFeature.selectionActions` registration — and an action set cut down to copy / cut / paste / delete, which is exactly what item 1 already supports. Continuously reading selected text and the frontmost URL out of every running app is the whole subject of this PR, and it should be the whole diff while someone reads it.
3. **The local transforms.** Case, spaces, underscore, join, comma list, sort/reverse/shuffle, quotes and brackets, URL and Base64, calculate. Roughly fifteen actions that are pure functions on a `String` plus a catalog entry each. Cheap once 2 is settled, and safe to review in bulk because none of them reach anything.
4. **Everything that leaves the machine or the sandbox.** `sendToAI`, `searchWeb`, `translate`, `convertCurrency`, `Run in Terminal`. Separate discussion, separate defaults, and `Run in Terminal` probably a separate decision from the other four.

I am reversing myself from four days ago, when I said one PR or two mattered less to me than whether it got reviewed as a permission change. That was wrong: at this size the second does not happen without the first.

Nothing is lost by splitting. `feature/selection-actions` stays on your fork, this PR reopens with one click if @vorssaint would rather take it whole, and its only conflict against current `main` is `Core/FeatureCatalog.swift` — a registration list, resolved by keeping both cases. I would rather see item 1 merged next week than see all four wait on a review of item 2 that has not started.

naveenkrdy:
Split into the four PRs discussed above:

1. #1032 - `refactor(quick-tools): extract synthetic paste into a shared helper`
2. #1033 - `feat(selection-actions): add floating menu with copy, cut, paste, delete`
3. #1037 - `feat(selection-actions): add local text-transform actions`
4. #1038 - `feat(selection-actions): add actions that reach outside the app`

#1033 already has a round of review from @PathGao addressed (enabled-set storage direction, an off-main-thread fix). #1037 and #1038 are both stacked on #1033 and independent of each other. All four currently carry each other's diffs since none have merged yet - will rebase each down to just its own changes as they land in order.

PathGao:
If "all four carry each other's diffs" includes #1032, that's the one to unstack now rather than as the rest land. It depends on nothing else in the stack, and its whole case for going in first is that it reviews in one sitting — carrying the other three takes exactly that away. Rebase it down to just its own changes and it can be read today. #1037 and #1038 sitting on #1033 is fine as is.

I'll pick up #1032 and #1033 in that order.

PR #715, opened 2026-08-20
fix(url-cleaner): let the Clean URL fields take a click across their row

PathGao:
The extra-parameter field on the Clean URL page could not be clicked where it looked like it should be. In a grouped Form a `TextField`'s first argument renders as a leading row label rather than a placeholder, so "ref, source" sat on the left as static words while the editable strip was a short run on the right. Clicking the words did nothing, and the field's own frame started well past the middle of the row, so most of the row was dead. It reads as a description of built-in behaviour rather than as something to fill in.

The manual clean-now field on the same page had the same shape.

## What changed

Both fields pass their hint as `prompt:` with an empty label, so each owns its whole row and takes a click anywhere across it. That is the shape the Command Bar and Screenshot pages already use for a field of their own.

With the field usable, two things were finished off in the same section:

- **A Save button.** The list was bound straight to its stored value, so every keystroke was a saved setting and `URLCleanerService` picked it up on its next 0.8 s poll — typing `source` published `s`, then `so`, on the way. The field now edits a draft; Save commits it, Return commits it too, and the button enables only when the draft parses to a different set. Small window, but "when did this apply?" now has an answer.
- **The parsed names are shown** under the field, which is where lowercasing, trimming and de-duplication become visible. What is committed is that normalised list, so the stored value, the field and the live rules always read the same.

## Alternatives considered

**Keep the label and just add a border.** That was the first attempt and it is not enough: a border makes the short strip visible without making the rest of the row live. The words on the left stay unclickable, which is the actual complaint.

**Commit on Return and focus loss, with no button.** Rejected: this page was just fixed for a control whose only affordance was invisible. Return still works; the button is what says so.

**Render the parsed names as chips.** Rejected: SwiftUI has no wrapping stack on this deployment target, so chips mean writing a `Layout` for a row that usually holds two to five names. A monospaced separated line wraps for free.

## Not changed

The stored format is still one comma-separated string, so nothing migrates and settings backups keep working. `URLCleaning` and the cleaner service are untouched. The built-in rules and the clean-now flow behave as before.

Tests: `TESTS OK (7939 checks)`, plus a full `./build.sh` for the view. Assertions pin the hint travelling as a prompt, no field spending its row on a label, and typing reaching a draft rather than the stored list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,12 @@ Vorssaint 3.3.3 adds full manual and temperature-based Fan Control, an opt-in Ki
   current and target RPM for every fan.
 
 ### Fixed
-- The Clean URL settings fields now span their whole row and draw a border. Their
-  hint used to render as a label beside a short strip of field, so clicking the
-  words did nothing. Thanks to @PathGao.
+- The Clean URL parameter list is now saved with a button instead of taking effect
+  as it is typed, and shows the names as the cleaner reads them. Half a name used
+  to apply for as long as it took to finish typing the rest, so a link copied at
+  that moment could lose a parameter that only matched the fragment. Both fields on
+  the page also span their whole row: their hint used to render as a label beside a
+  short strip of field, so clicking the words did nothing. Thanks to @PathGao.
 - Scratchpad tabs and the pin, close, new pad and pad actions buttons now respond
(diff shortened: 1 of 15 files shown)


vorssaint:

Issue #200, opened 2026-07-10
[Bug]: Click Dock Icon to Minimize


danilo-alm:
### Description

The Click Dock Icon to Minimize behavior does not work with DBeaver. I suppose it might be something related to the app opening a loading window first and then only after the app itself? I tested affinity photo, that also has a loading screen, but it works correctly.
When on windows, I remember pinning the icon on the dock and opening it would trigger a behavior where the app opened itself appeared as other separate icon. Can this be fixed or is it a limitation on how the app is done somehow?

### Steps to Reproduce

1. Toggle the mentioned option in vorssaint
2. Open dbeaver
3. Focus dbeaver window
4. Click on the dock icon. Nothing happens

### Vorssaint Version

3.1.11

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Self-Updated

PR #717, opened 2026-08-20
fix(dock-click): restore only the windows the click minimized

PathGao:
**Depends on #716.** Both branches touch the same `.restore` path; merge #716 first, then this one. The diff here shows #716's commit as its parent.

---

The switch is called "Click Dock icon to minimize" and its caption promises the pair: click to minimize, click again to restore. The restore side does not check who minimized the window. `toggleIntentWindow` is 1.5 seconds, so only a second click inside that window takes the advertised `.toggle(.restore)` path; anything later — and anything the user minimized with ⌘M, the window's own button, or the app itself — falls through to the state-derived `.restore` and gets taken over too.

Taking those over buys nothing. The batch's stacking is rebuilt from `minimizeZOrder`, captured when this feature does the minimizing. For windows it never minimized there is nothing captured, and the restore walk has to fall back to guessing at a front window — as its own comment puts it, "the batch order is a guess, and forcing focus onto a guess is worse than leaving it to activation".

## What changed

`action(...)` gains `ownsMinimize`, and `.restore` now requires it. The service passes `minimizeZOrder[pid] != nil` — that table is written when this feature minimizes an app and consumed when it restores one, so it means exactly "we put these away and have not brought them back".

Effect:

- Windows minimized by other means keep the Dock's native restore.
- The advertised toggle is unchanged: a second click inside 1.5 s never reaches this rule, and after that `minimizeZOrder` still keeps the pair working.
- The switch's name and its actual reach now agree.

## Alternatives considered

Keying on `lastAction` instead was rejected: it is only meaningful inside the 1.5 s intent window, which would drop the feature's own minimize as soon as the user waited a beat. `minimizeZOrder` survives until the restore consumes it.

## Not changed

The guess fallback in `restoreBackToFront` stays. It is still reachable for apps whose accessibility server answers with nothing (issue #200) — their `onScreenWindowIDs` comes back empty even when this feature did the minimizing. Narrowing the scope makes that path cold, not dead.

**Unverified:** the rule's previous comment claimed the Dock's native click "would activate without unminimizing — Finder would even open a brand-new window". This change routes non-owned minimizes back to that path. Native restore behaves correctly for a regular app on macOS 26; Finder specifically was not re-checked.

Tests: `TESTS OK (7828 checks)` on this branch (7826 baseline + 2). The two existing `.restore` cases now pass `ownsMinimize: true` explicitly, and two cases cover a non-owned minimize passing through.


vorssaint:

Issue #520, opened 2026-08-06
Clipboard history is written with default file permissions


ThomasWaldmann:
Hi, and thanks for Vorssaint — I was reading through the source (v3.3.1) out of interest and came away impressed by how carefully the sensitive paths are handled. One small inconsistency stood out, so I thought I'd mention it.

### What I noticed

Clipboard history is persisted with whatever the process umask happens to give (typically `0644` for the file, `0755` for the directory):

- `Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift:823` — `createDirectory` with no `posixPermissions`
- `Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift:827` — `data.write(to:options:.atomic)` for `ClipboardHistory.json`
- `Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift:1212` and `:1215` — the same for the copied-image PNGs under `ClipboardImages/`

The reason it caught my eye is that the neighbouring store for temporary screenshot links does exactly the opposite, and deliberately so:

- `Sources/Vorssaint/Services/QuickTools/ScreenshotShareService.swift:207-213` — `0o700` on the directory, `0o600` on `records.json`, with a comment explaining that the private deletion token needs a safe place to live

A deletion token for a link that expires within 24 hours is arguably less sensitive than a rolling history of everything the person has copied, so the stricter treatment seems like it belongs on the clipboard store at least as much.

### Why I think it's worth a small change

To be clear about the scope: this is hardening, not an exploitable vulnerability. `~/Library` is already `0700` on macOS, so another user account on the same Mac cannot walk in, and any process running as the same user can read the file regardless of its mode. I looked at whether this belonged in a private advisory instead and concluded it does not cross a privilege boundary, so a public issue seemed like the right venue — happy to move it if you would rather handle it differently.

What tighter modes would still buy:

- Consistency with the screenshot store, so the intent is legible to anyone reading the code later
- A slightly better story for backups, cloud-synced or shared home directories, and admin users browsing another account's files
- Group- and other-readable bits removed for anyone running with an unusual umask

### Possible fix

The same three lines the screenshot store already uses, applied in `persist()` and in `ClipboardImageStore.store(_:)`:

```swift
try FileManager.default.createDirectory(at: directory,
                                        withIntermediateDirectories: true,
                                        attributes: [.posixPermissions: 0o700])
try data.write(to: file, options: .atomic)
[...]

PR #719, opened 2026-08-20
fix(storage): store user data owner-only

PathGao:
Ref #520

Everything the app keeps in `~/Library/Application Support/<bundle id>` is the
user's own material, and five of the seven stores there created directories and
wrote files with no permission attributes, so they took the process umask:
`0755` directories and `0644` files, readable by every other account on the Mac.
Clipboard history and its copied images (`ClipboardHistoryService.swift:781`,
`:1273`), shelf payloads (`ShelfService.swift:216`) and recording takes
(`RecorderTakeStore.swift:51`).

The other two did the opposite on purpose. `ScreenshotShareService` and
`RecordingShareService` create `0700` and write `0600`, with a comment
explaining that the private deletion token needs a safe place to live. So the
same question had two answers in one codebase, and the answer that lost covers
the larger amount of private data. `RecorderTakeStore` even documents its takes
as "the same private, disposable master a screen recording gets" while creating
them world readable. Reported by @ThomasWaldmann in #520, who read the source
and found the inconsistency.

### The change

`PrivateFileStore` becomes the single place that builds paths inside the
container and the only way its contents are created: directories `0700`, files
`0600`. The six stores lose their own copy of the same five-line path builder
and the two share services lose their inline permission block.

Both modes are set again on **every** write, not once at creation, for two
separate reasons:

- `createDirectory` leaves a directory that already exists with the mode it was
  created with, so an install from any earlier version would keep `0755`
  forever. Walking from the written path up to the container corrects it on the
  next write, with no migration step and no version stamp.
- `Data.write(options: .atomic)` swaps in a freshly created file each time, so a
  mode applied once would not survive the next save.

After: the container and every store folder under it are `drwx------`, and
`ClipboardHistory.json`, the `ClipboardImages` PNGs, `ShelfFiles` payloads and
both `records.json` are `-rw-------`, on a fresh install and on an existing one
alike. Nothing above the container is touched; the walk stops there.

### Trade-offs

- **Tightening only `ClipboardHistory.json`**, the file the report names, was
  the smaller change. The neighbouring stores hold the same class of data
  through the same mistake, so that fix would leave the report half true.
- **Tightening only the container and leaving file modes alone** would block
  other accounts just as effectively, since `0700` stops traversal. Rejected
  because it contradicts the two share services, which already write `0600`,
  and would reinstate the inconsistency this removes.
- **A one-shot migration at launch** was the alternative to setting the mode on
  every write. Rejected: it needs a version stamp and an ordering guarantee
  against the first store access, and it cannot heal a container that something
  outside the app loosens later.

### Not changed

- `UpdateService`, which writes a downloaded installer and a helper script.
  That is not user data and the updater deserves its own look.
- The Scratchpad export path (`ScratchpadService.swift:329`), which writes where
  the user picks in a save panel. A file exported to Documents should get
  ordinary permissions.
- Files already on disk are not swept. Each is corrected when its store is next
  written, which is the first copy, drop or recording after updating.
- The `Bundle.main.bundleIdentifier ?? "com.vorssaint.utils"` fallback that only
  the two share services carried is gone. The container is now nil outside a
  real bundle, as it already was for the other five stores.

### Tests

`./build.sh --test` 7832 checks OK (7826 before) · `./build.sh --dev` clean ·
`--selftest` OK. MacBookAir (M5), macOS 26.5.2.

The new checks pin the path walk as a pure function, then exercise the real
`FileManager`: a root created `0755` on purpose, a store directory and a file
written into it, asserting `0700`, `0600`, the file's contents, and that the
container itself was tightened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ Vorssaint 3.3.3 adds full manual and temperature-based Fan Control, an opt-in Ki
   @PathGao.
 - App Switcher window preview warm-up no longer crashes right after startup when
   the system returns duplicate window identifiers. Thanks to @james-rose.
+- Clipboard history, its copied images, shelf files, recordings and temporary share
+  records are now stored so only your own account can read them. Folders an earlier
+  version left readable by other accounts on the Mac are corrected on the next
+  write. Thanks to @ThomasWaldmann.
 
 ## [3.3.2] - 2026-08-20
 
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
@@ -779,13 +779,7 @@ final class ClipboardHistoryService: ObservableObject {
     /// pushes back past a few megabytes, which a large history of long texts
     /// can reach. Without a resolvable home the blob stays in UserDefaults.
     private static var storeURL: URL? {
-        guard let base = FileManager.default.urls(for: .applicationSupportDirectory,
-                                                  in: .userDomainMask).first,
-              let bundleID = Bundle.main.bundleIdentifier
-        else { return nil }
-        return base
-            .appendingPathComponent(bundleID, isDirectory: true)
-            .appendingPathComponent("ClipboardHistory.json")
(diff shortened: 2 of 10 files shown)


vorssaint:

PR #729, opened 2026-08-20
Add input source (language) switching to Super Key solo actions

BenjaminD2023:
## Summary

- add a separate Super Key solo action for switching the active input source (language)
- treat a solo press shorter than 500 ms as an input-source switch and a longer press as Caps Lock
- preserve configured Super Key chords and the existing Nothing, Caps Lock, and Escape actions
- handle active IME composition by committing the selected candidate before switching
- add localized labels for every supported language and focused state-machine coverage

## Behavior before

When Super Key remapped Caps Lock to F18, macOS no longer received the native Caps Lock action used for switching input sources. A quick solo press therefore could not switch languages. Control-Space was also consumed while a Pinyin candidate session was active.

## Behavior after

With **Switch input source; hold for Caps Lock** selected:

- quick solo Caps Lock press: sends the configured macOS Control-Space input-source shortcut through the existing in-process System Events runner
- active marked IME text: commits the selected candidate first, then switches while the same app remains frontmost
- solo hold for at least 500 ms: toggles Caps Lock
- Caps Lock with another key: keeps applying the configured Super Key modifiers without running a solo action
- existing solo actions retain their prior autorepeat behavior

The input-source action requires System Events Automation permission and an enabled Control-Space input-source shortcut.

## Verification

- `./build.sh` — passed; release bundle produced
- `./build.sh --test` — `TESTS OK (7831 checks)`
- signed local app smoke test — quick presses toggled ABC/Pinyin, holds toggled Caps Lock, and chord events retained all configured modifiers
- active Pinyin composition check — selected candidate committed and the source switched to ABC
- `git diff --check` — passed


fzlzjerry (approved):
LGTM, good effort

Issue #483, opened 2026-08-04
[Bug]: Apps appearing werildy in switcher when no windows are open and preview size is set to small


Yahddyyp:
### Description

So when finder has no windows open the finder text appears to be not in its right place, this only happens when the preview scale is set to small

### Steps to Reproduce

1. open vorssisant 
2. set switcher scale to small 
3. close all finder windows and make sure to toggle the setting to make finder appear when no windows are open in the switcher
4. cmd + tab 

### Screenshots / Recordings

<img width="1010" height="620" alt="Image" src="https://github.com/user-attachments/assets/a8693e19-ab79-48ca-862b-025b0ede5dd1" />

### Vorssaint Version

3.3.0

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

Yahddyyp:
I have confirmed that this can sometimes also happen to other apps as well

vorssaint:
Windowless apps now keep their labels aligned when the App Switcher uses Small previews. This is fixed in 3.3.1.

Yahddyyp:
This issue still persistes after updating to v3.3.2

Yahddyyp:
It is still persistant in the v3.3.3-beta1

Issue #736, opened 2026-08-21
[Bug]: Issue #483 still persists after updating to v3.3.2


Yahddyyp:
### Description

Same as #483 

### Steps to Reproduce

Same as #483

### Vorssaint Version

v3.3.2

### macOS Version

macOS 26 (Tahoe)

### Installation Method

homebrew

PR #746, opened 2026-08-22
fix(app-switcher): size grid cards from their parts

PathGao:
Independent of #744 and #745 — different files, no conflicts, reviewable in any order.

## What was wrong

An App Switcher grid card sized its thumbnail by subtracting a hard-coded number from the card:

```swift
.frame(width: SwitcherGrid.cardWidth - 20,
       height: SwitcherGrid.cardHeight - 72)
```

The 20 is right: 10pt of padding on each side. The 72 is not. The chrome below the thumbnail is 10pt padding twice, 7pt of `VStack` spacing and a 29pt title block — 56pt, not 72. The card had been reserving 16pt it never used, at every preview size, because the parts are all fixed while the number that stood for them was never revisited.

The title block was short as well. Measured against the fonts it holds:

| | needs | had |
|---|---|---|
| 13pt title line | 16 | |
| 10.5pt subtitle line | 13 | |
| 2pt between them | 2 | |
| **total** | **31** | **29** |

With `alignment: .top` that clipped the bottom of the subtitle's descenders on every card that had one.

This is the same defect #720 removed from Dock Preview cards, where the comment reads "the old code hard-coded a 54pt deduction that no longer matched the parts it stood for". The same shape survived here because nothing pointed at it.

## What changed

The parts are named and the thumbnail is derived from them, so the two can no longer drift apart:

```swift
static let padding: CGFloat = 10
static let titleSpacing: CGFloat = 7
/// One 13pt line over one 10.5pt line, 2pt apart, descenders included.
static let titleHeight: CGFloat = 31
static var thumbnailHeight: CGFloat { height - padding * 2 - titleSpacing - titleHeight }
```

Net effect: the thumbnail gains 14pt of height at every preview size (16 recovered, 2 given to the title band), the card itself is unchanged, and the panel geometry around it is unchanged. Subtitles stop clipping.

#793 has since given that stand-in app icon its preview scale. Naming it here is what lets a test hold it to the thumbnail it sits in — the two only became comparable once both sides stopped being literals.

## Why the metrics moved file

They now live in `SwitcherSupport.swift` next to `SwitcherIconRowLayout`, which is where the other switcher layout math already is. `SwitcherGrid` keeps the panel geometry that needs an `NSScreen`; the card is pure arithmetic. That split is also what makes the relationship testable — the selftest target compiles a whitelist of pure-logic files, and `AppSwitcher.swift` is not on it.

## What was checked and left alone

The other two card layouts were audited for the same defect and do not have it. `SwitcherIconRowLayout`'s preview card deducts a fixed 16 and 38 that do decompose correctly (16 padding + 6 spacing + a 13pt line in 16pt of room), and its icon row already builds `rowHeight` from its parts with a comment explaining why the text must not shrink. `SwitcherIconRowLayout` also caps preview scaling at 1.15, so its cards never blow up at Large or Extra Large; `SwitcherGrid` deliberately does not, and that is unchanged here.

Tests: 8376 checks pass, 6 of them new. Baseline on `1d961c6b` is 8370.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,9 @@ Vorssaint 3.3.3-beta.1 introduces the opt-in beta release channel and in-app fee
   the window shortcut reach each window, even when showing one entry per app.
 - Windowless apps no longer repeat or misplace their name in the App Switcher's Small
   size. Thanks to @Yahddyyp.
+- App Switcher grid cards now give their thumbnail the 14 points their chrome
+  was reserving and not using, and their title band holds a subtitle without
+  clipping its descenders. Thanks to @PathGao.
 - Clicking the Dock icon of a background app to restore its minimized windows now
   brings that app forward, instead of leaving its windows behind whichever app was
   already in front. Thanks to @PathGao.
--- a/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
+++ b/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
@@ -1542,10 +1542,10 @@ struct SwitcherGrid: Equatable {
     let visibleRows: Int
     let panelSize: CGSize
 
-    // Base sizes and breathing room scale together, so making previews smaller
-    // also keeps the panel from spending that saved space on empty gaps.
-    static var cardWidth: CGFloat { 288 * PreviewSizing.scale }
-    static var cardHeight: CGFloat { 214 * PreviewSizing.scale }
+    // Breathing room scales with the cards, so making previews smaller also
+    // keeps the panel from spending that saved space on empty gaps.
+    static var cardWidth: CGFloat { SwitcherGridCard.width }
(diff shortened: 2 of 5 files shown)


PathGao:
Updating the justification for this one, since the ground moved: #793 merged and closed #483 and #736. **This is no longer a bug fix.**

What is left is the half #793 did not do:

- `SwitcherGridCard` replaces the drifting magic numbers `-20`, `-72`, `29` and the bare `80 * PreviewSizing.scale` with one place deriving the card from its parts.
- `Tests/MetricsTests.swift` gains the assertion that holds the fallback icon inside the thumbnail at every preview size. That is currently the only thing in the repo that would catch #483 coming back.

It merges clean against `main` @ `9d9c6563` (verified with `git merge-tree`), and it does not conflict with #810. Re-labelling from `blocked` to `ready to merge`; the blocker was #793, which has landed.

Merge it as a guard or close it — but if it is closed, #483's shape has nothing watching it.

PR #753, opened 2026-08-22
feat(kill-process): show listening ports

Bald-M:
PixPin_2026-08-23_01-52-21

## Summary

- Show listening TCP ports in the Kill Process table.
- Allow processes to be found by port number.
- Merge, sort, and de-duplicate ports when related processes are grouped.
- Keep the Process, CPU, Memory, PID, Port, and action columns aligned across window sizes.
- Add coverage for `lsof` parsing, port filtering, grouped ports, and column-width calculations.

## Testing

- `./build.sh` — passed
- `./build/Vorssaint --selftest` — passed
- Verified the Kill Process table and Port column in the built app
- `./build.sh --test` — the new tests passed; 5 pre-existing natural-language date parsing assertions still fail

### DIFF
--- a/Sources/Vorssaint/Services/KillProcess/KillProcessService.swift
+++ b/Sources/Vorssaint/Services/KillProcess/KillProcessService.swift
@@ -19,6 +19,8 @@ struct KillProcessEntry: Identifiable, Equatable {
     let bundleURL: URL?
     let groupedCount: Int
     let isProtected: Bool
+    /// Listening TCP ports owned by this process, sorted and de-duplicated.
+    let ports: [Int]
 
     var id: pid_t { pid }
 }
@@ -77,10 +79,8 @@ final class KillProcessService: ObservableObject {
             }
             return ascending ? !result : result
         }
-        let needle = query.trimmingCharacters(in: .whitespaces).lowercased()
-        guard !needle.isEmpty else { return sorted }
         return sorted.filter {
-            $0.name.lowercased().contains(needle) || String($0.pid) == needle
+            KillProcessSupport.matches(query: query, name: $0.name, pid: $0.pid, ports: $0.ports)
         }
     }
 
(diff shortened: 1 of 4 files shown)


Bald-M:
Maybe there's better solution. Dont merge.

Bald-M:
Closed. @PathGao will implement this in the future

PathGao:
Whether this stays open is your call — the notes below are for whoever picks the feature up, because the use case is real and I expect it to come back.

Measured on a machine with 778 processes and 16724 open descriptors visible unprivileged, 5 of which hold a listening TCP port (20-run averages):

| | wall | CPU |
|---|---|---|
| `ps -eo pid,ppid,pcpu,rss,comm` | 42 ms | 0.02 s |
| `lsof -nP -iTCP -sTCP:LISTEN -Fpn` | 59 ms | 0.04 s |

**1. The Command Bar pays for the lookup and never shows a port.** `listeningTCPPorts()` sits inside `snapshot()`, and `CommandBarService.loadKillProcessEntries` shares that cache, but `CommandBarCatalog.killProcessEntries` is unchanged here, so the bar has no Port column. Every open of the bar outside the 3s cache window now runs `lsof` before the process rows are indexed: about 42 ms to about 101 ms on a hotkey path, for a column that does not exist on that surface.

**2. With grouping on, Kill does not target the process holding the port.** `ResponsibleProcess.owner` is `responsibility_get_pid_responsible_for_pid`, so a java process started from a terminal is owned by the terminal. `mergedPorts` then puts 8080 on the terminal's row, and `kill(_:force:)` signals `entry.pid`, which is the terminal. The port survives and the terminal does not — in exactly the scenario the feature is for. A grouped row has to carry the pid that actually owns the port, or the column has to be empty while grouping is on.

**3. Ports appear only for your own processes.** Unprivileged `lsof -iTCP -sTCP:LISTEN` cannot read other users' descriptors, so a port held by a root daemon or by Docker shows as no port at all. It is the same reason the hand-typed command needs `sudo` for those, and it covers a good share of the "something is already on 8080" cases.

On shape: the cost model does not match how the feature is used. The flow is "type 8080, kill it", so run `lsof` only when the query parses as a port in 1...65535 and leave the table itself alone. That removes 1 outright, makes the empty column question disappear, and costs nothing on any path that is not a port search. What it gives up is browsing ports without knowing which one you want, which does not seem to be what you were after.

None of this is about whether the feature is wanted. It is.

Issue #747, opened 2026-08-22
App becomes unresponsive after turning an external display off and back on


ozimosko:
### Description

When an external monitor is connected, I can turn it off using the power button in the Displays section of the app. However, when I try to turn the external monitor back on using the same button, the app becomes completely unresponsive.
The app does not recover, and I have to force quit it using Activity Monitor.

### Steps to Reproduce

Connect an external monitor to the Mac.
Open Vossaint.
Go to the Displays section.
Turn off the external monitor using the power button.
Try to turn the external monitor back on.
The app becomes unresponsive.

Expected behavior
The external monitor should turn back on and the app should continue working normally.

Actual behavior
The app freezes and stops responding, requiring a force quit through Activity Monitor.

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

PathGao:
Tracking PR: #762 (open, `ready to merge`).

It moves display power changes onto the main thread at all five `configureDisplay()` call sites, plus a main-thread guard. Same root cause as #647 and, most likely, #649. Not merged yet.

PathGao:
Merged Please test on next release

googlejam:
I was just about to post this issue.

Issue #647, opened 2026-08-15
Freezing when disabling external monitor


yusrmuttaqien:
### Description

When disabling external monitor from Vorssaint:

1. It slow, took a while to turn off the display (compared to LightsOut)
2. When it's done disabling external display, it goes unrespoinsive (mouse goes beachball when hovering the menubar app)

### Steps to Reproduce

1. Open Vorssaint menubar app
2. Select display tab
3. Click off button on external display
4. The button switches to loading icon
5. Can either wait with the menubar app open or closed. It eventually disable the external display
6. By now, Vorssaint is unresponsive, showing beachball mouse when hover the menubar app
7. If you left the menubar app when display is disabled, the whole popup can't be closed until Force Quit

### Vorssaint Version

3.3.1

### macOS Version

macOS 15 (Sequoia)

### Installation Method

DMG

PathGao:
Tracking PR: #762 (open, `ready to merge`) — same root cause as #747: `configureDisplay()` runs on the background work queue while the CoreGraphics reconfiguration callback touches AppKit in place. Not merged yet.

PathGao:
Merged，Please test on next release

yusrmuttaqien:
Just updated Vorssaint, the latest build fixes the hang after toggling external monitor.

At first try it still took a while to disable the monitor. But, subsequent tries completes in instant. Monitor detection seems to be more reliable, too.

I'll close the issue now. Will reopen/create new one if the issue came back.

PR #762, opened 2026-08-22
fix(display): run display power changes on the main thread

PathGao:
Ref #747, #647

Context: #747 — turning an external display off from Displays works, turning it back on with the same button freezes the whole app with no recovery.

## What was wrong

`toggleDisplay` ran the display reconfiguration on `workQueue`, the serial queue that also carries DDC probing and topology rebuilds. A `CGBeginDisplayConfiguration` → `CGSConfigureDisplayEnabled` → `CGCompleteDisplayConfiguration` transaction is not background work: `CGDisplayConfiguration.h` states that a reconfiguration's callbacks are invoked "from within the display reconfiguration function when in the program that is driving the reconfiguration". In a Cocoa app those callbacks are AppKit's — rebuilding `NSScreen` and posting `NSApplicationDidChangeScreenParametersNotification` — so driving the transaction off-main runs main-thread AppKit work on the work queue, while the main thread is itself inside CoreGraphics (`refresh()` → `CGGetOnlineDisplayList`/`CGGetActiveDisplayList`) and cannot return until the transaction closes. Neither side finishes.

Off and on differ only in how long that window stays open. Removing a display closes in milliseconds; bringing one back keeps the transaction open while the panel actually lights up, which is long enough for the main thread to enter CoreGraphics every time — hence "off always works, on always hangs".

There was no way out of it either. `applicationWillTerminate` calls `restoreDisplaysBeforeTermination`, which did `workQueue.sync`, so once the queue was wedged Quit hung too and Activity Monitor was the only exit, exactly as reported. Every sibling caller had the same defect: `stop()`, `restoreDisplaysLeftOff()`, `restoreManagedDisplayIfHeadless()` and the termination path all reached `configureDisplay` from the work queue.

## What changed

- `configureDisplay` is main-thread only and enforces it: a call from any other thread is refused with an `os_log` error and returns `false`, which the callers already surface as `.failed` in the panel. A future regression fails visibly instead of hanging.
- All five call sites reach it on the main thread. The toggle keeps its work-queue hop for the pre-flight check and the gamma baseline, then hands the transaction to main in the new `commitDisplayToggle`. `stop()` fences on the queue, reconfigures on main, restores gamma back on the queue, preserving the previous order. `restoreDisplaysLeftOff` and `restoreManagedDisplayIfHeadless` are already called on main and no longer hop.
- `restoreDisplaysBeforeTermination` and `stop()` shared a duplicated restore loop; it is now one `restoreManagedDisplays()`, and its `workQueue.sync` window shrinks to the gamma restore alone, so a graceful Quit no longer waits behind a display transaction.

Effect: the on-toggle completes instead of hanging. The UI is blocked for as long as the display takes to come up — a bounded stall of roughly a second or two rather than a permanent freeze — and Quit works throughout.

**Keeping the transaction on the work queue with a timeout.** A `DispatchQueue.main.sync` bounce would deadlock against `restoreDisplaysBeforeTermination`'s `workQueue.sync`, and a reconfiguration already blocking cannot be cancelled. It also leaves the unsupported threading in place.

**Keeping the work queue and moving `currentTopology()` off main instead.** Smaller diff, but it widens the race window rather than closing it, and leaves AppKit's reconfiguration callbacks on a background thread.

The main thread is what comparable implementations use: [InternalDisplayOff](https://github.com/RonaldPark89/InternalDisplayOff), which drives the same private `CGSConfigureDisplayEnabled` transaction, runs it straight from the UI action on main. Vorssaint was the outlier.

## Not touched

`displayName()` calls `NSScreen.screens` from `rebuild()` on the work queue — also AppKit off-main, also worth fixing, but it runs identically on both toggle directions and on every plain hotplug, so it is not what makes this hang deterministic. Sent separately as #765. The transaction still uses `.forAppOnly`, and DDC probing, gamma handling and pacing stay on the work queue. No user-facing strings.

## Please confirm on real hardware

This was diagnosed by reading, not reproduced: I have no external display to disconnect.

1. Attach an external display, enable display control, open Displays in the panel.
2. Press the power button on the external display's row — it goes off (unchanged).
3. Press it again. **Expected:** it comes back within a second or two and the app keeps responding; the row's spinner clears and the brightness slider returns.
4. While a display is off, quit from the menu. **Expected:** it quits and the display comes back; the previous build could hang here.
5. `log show --last 5m --predicate 'subsystem == "com.vorssaint.utils" AND category == "display"'` — a line reading `refused to reconfigure display N off the main thread` means a call site was missed.

Tests: 8040 checks pass (8038 before), full `./build.sh` clean, `--selftest` OK. Which thread drives the transaction is not a proposition a pure helper can carry, so it is pinned by source: the guard must sit inside the same function as `CGBeginDisplayConfiguration(`, matched on the CoreGraphics symbol and the `func` keyword rather than on any name or line break.


bayujo:
Hit #747 today on 3.3.2 (macOS 26.6, Apple silicon) and sampled the live process before force-quitting. This patch fixes it, but the hang I captured closes through a different pair than the one described above, and the difference matters for how the fix is reviewed.

`sample` on the frozen process, trimmed to the two threads that matter:

```
Thread_2166370   DispatchQueue_1: com.apple.main-thread  (serial)
  -[NSView _layoutSubtreeWithOldSize:]  (in AppKit)
    HeightReportingHostingView.layout()  (in Vorssaint)
      NSHostingView.layout()  (in SwiftUI)
        ViewGraphRootValueUpdater.render(interval:updateDisplayList:targetTimestamp:)
          GraphHost.flushTransactions()  (in SwiftUICore)
            ViewBodyAccessor.updateBody(of:changed:)  (in SwiftUICore)
              protocol witness for View.body.getter in conformance DisplayPowerButton
                closure #1 in DisplayPowerButton.body.getter  (in Vorssaint)
                  DisplayPowerButton.label.getter  (in Vorssaint)
                    BrightnessService.canToggleDisplay(_:)  (in Vorssaint) + 220
                      _pthread_mutex_firstfit_lock_slow  (in libsystem_pthread)
                        __psynch_mutexwait  (in libsystem_kernel)      <- blocked on stateLock

Thread_2488890   DispatchQueue_153: com.vorssaint.utils.brightness  (serial)
  _dispatch_call_block_and_release  (in libdispatch)
    partial apply for closure #1 in BrightnessService.toggleDisplay(_:)  (in Vorssaint) + 1240
      specialized static BrightnessService.forgetDisplaySwitchedOff(_:)  (in Vorssaint) + 592
        -[NSNotificationCenter postNotificationName:object:userInfo:]  (in Foundation)
          _CFXNotificationPost  (in CoreFoundation)
            -[NSOperation waitUntilFinished]  (in Foundation)
              _pthread_cond_wait  (in libsystem_pthread)
                __psynch_cvwait  (in libsystem_kernel)                 <- waiting for the main queue
```

Neither thread is inside CoreGraphics. The cycle is `stateLock` against the main queue:

1. `toggleDisplay`'s work-queue closure takes `stateLock` at `BrightnessService.swift:432` and, in the `targetEnabled` branch only, calls `forgetDisplaySwitchedOff` at :438 while still holding it.
2. That writes `UserDefaults`, which posts `UserDefaults.didChangeNotification` synchronously. Two observers are registered with `queue: .main` — `StatusItemController.swift:190` and `Permissions.swift:70` — so CoreFoundation wraps delivery in an `NSOperation` and blocks the work queue until the main thread drains it.
3. The main thread is meanwhile in `DisplayPowerButton.body` → `canToggleDisplay` → `stateLock.lock()`, waiting on the lock that closure holds.

This also explains cleanly why off works and on freezes, which the CoreGraphics account only explains by duration: the turn-off path is the `else` branch and never calls `forgetDisplaySwitchedOff` under the lock. :438 is the only one of the five call sites in the file that posts inside a locked region — :302, :524 and :1068 all post after `stateLock.unlock()`, and :424 is unlocked too.

This PR fixes that deadlock as well, because moving the whole block into `commitDisplayToggle` on the main thread means the write, the notification and `canToggleDisplay` are all on the same thread and can no longer wait on each other. Worth flagging for review: it fixes it by relocating *all* of the bookkeeping, not just the reconfiguration. A narrower revision that moved only `configureDisplay` to the main thread and left the `stateLock` region on `workQueue` would satisfy the CoreGraphics requirement described here and still hang exactly as above.

As a belt-and-braces follow-up, moving the :438 call below `stateLock.unlock()` would bring that site in line with the other four and keep the UserDefaults write off the lock regardless of which thread later owns that block.

Probably the same root cause as #647 (same beachball, reported on the disable direction, 3.3.1).

PathGao:
Thanks — the sample is the useful half of this. I checked the call sites you name against `main`: `forgetDisplaySwitchedOff` has five, and :438 is the only one inside a locked region (432/447); :302, :524 and :1068 all run after `stateLock.unlock()`, and :426 never holds it. The on/off asymmetry falling out of the `else` branch rather than out of how long CoreGraphics takes is the part I had wrong.

Pushed the belt-and-braces change you suggested: the write now happens with the lock released, so the locked region no longer depends on which thread owns the block. Added a source test pinning that nothing writes that list under `stateLock`, since this fails as a hang rather than as a wrong answer. I also swept the rest of the app for the same shape — a locked region containing anything that can block on the main queue — and this was the only site.

Your review note is the right one to act on: a narrower revision that moved only `configureDisplay` and left the `stateLock` region on `workQueue` would satisfy CoreGraphics and still hang exactly as you captured. That is now blocked by the lock being clean on its own, not by the thread.

Agreed on #647 — same cycle, reported from the direction that happens to freeze.

PR #763, opened 2026-08-22
fix(memory monitoring): corrected "Memory Used" reporting

pergioa:
## Description

Replace the aggregate free/file-backed-page calculation with Activity Monitor-style accounting.

The monitor now reports app memory, wired memory, compressed memory, and the full reserved tagged-memory storage region. The tagged storage is reserved unified memory outside the ordinary free-page pool; omitting it caused a persistent undercount on current Apple-silicon macOS releases.

This was validated against two local Activity Monitor snapshots: the app and wired components matched directly, while including the 2 GiB reserved tagged-storage region brought the total within normal sampling variation.

The diagnostic command used for validation is intentionally not part of this PR.

## Swift 6.0.3 compatibility

Xcode 16.2 does not declare the tagged-storage members of `vm_statistics64`, even though newer kernels can report them. A header-only C compatibility module defines the public HOST_VM_INFO64 rev3 structure with C ABI layout and compile-time assertions for its 248-byte size, 8-byte alignment, and field offsets. The common prefix is also checked against the active SDK's `vm_statistics64`; SDKs that expose rev3 additionally verify the full revision count and tagged-storage offset.

The implementation requests the rev3 payload through that checked C type and uses the kernel-returned field count as the ABI revision discriminator. Older kernels return a shorter revision, in which case tagged storage remains zero while the legacy fields continue to work. Tagged storage is also rejected if its page count would exceed physical memory.

Verification:
- all 8,222 unit checks pass, including legacy, truncated, rev2, rev3, and invalid tagged-storage cases
- the complete development app builds and links with the compatibility module
- the optimized release build passes
- a live HOST_VM_INFO64 read returned the expected rev3 count of 62 words

## Screenshots

Before:
image

After:
image


pergioa:
Following up on the ABI concern raised by @PathGao in #837: `host_statistics64` does not expose a separate revision field; its in/out `count` is the revision-negotiation mechanism, and the kernel returns the count it populated. The current count check therefore correctly detects the tagged-storage revision.

I agree that duplicating raw byte offsets in Swift is harder to audit. I am going to replace that part with an ABI compatibility layer whose layout is checked at compile time, keep the count-based revision detection, and add coverage for truncated/legacy payloads and invalid tagged-storage values. This should preserve Swift 6.0.3 compatibility while making an ABI mismatch fail visibly instead of silently affecting Memory Used.

PathGao:
You're right about the revision detection, and I've corrected #837.

I went and checked the SDK headers: `vm_statistics64` has no version field anywhere in it, and `mach/host_info.h` defines the revisions *as* counts:

```c
#define HOST_VM_INFO64_REV3_COUNT  HOST_VM_INFO64_COUNT   // 62
#define HOST_VM_INFO64_REV2_COUNT  ...                    // 40
#define HOST_VM_INFO64_REV1_COUNT  ...                    // 38
#define HOST_VM_INFO64_REV0_COUNT  ...                    // 24
```

So the revision *is* the count for this API, and the count check is the sanctioned way to detect it. My "checks length, not version" line treated those as two separate things when they're one. Sorry for the noise.

While I was in there I also verified all six offsets with `offsetof` against the macOS 26.5 SDK, so you don't have to re-derive them:

| constant in the PR | actual field | actual offset |
|---|---|---|
| `wired = 12` | `wire_count` | 12 |
| `purgeable = 88` | `purgeable_count` | 88 |
| `compressor = 128` | `compressor_page_count` | 128 |
| `external = 136` | `external_page_count` | 136 |
| `internal = 140` | `internal_page_count` | 140 |
| `totalTagStorage = 160` | `total_tag_storage_pages` | 160 |

All six correct, and `rev3ByteCount = 248` matches both `sizeof(vm_statistics64_data_t)` and `HOST_VM_INFO64_COUNT * 4`. I briefly misread `160` as `total_uncompressed_pages_in_compressor` (that one is at 144, with `swapped_count` at 152) before spotting that you're reading the tag-storage field, which is what your variable name says.

One thing on the compile-time layout you're planning. For it to actually catch an ABI change, the assertion has to be anchored to the real struct rather than to a second copy of the same numbers, e.g. `MemoryLayout<vm_statistics64_data_t>.offset(of:)` or a small C shim asserting `offsetof`. Re-declaring 12/88/128/136/140/160 inside a nicer wrapper and static-asserting them against themselves would compile just as happily on a kernel that moved a field.

Related, and worth a comment in the code: the count check tells you *how many ints the kernel wrote*, not *which layout it used*. Those are the same thing only because Apple has always appended. If a field ever gets inserted before offset 160, the count still comes back >= 62, `hasRev3` is still true, and offset 160 silently reads something else. That's exactly the failure mode you're trying to design out, which is why the assertion needs to sit on the real struct.

Coverage for truncated and legacy payloads sounds right to me. Thanks for digging into it.

PathGao:
I went to measure the question I was going to ask you, so here's the answer instead.

I was about to ask whether Activity Monitor actually counts the tagged-storage region, because if it didn't, the whole raw-byte path could be dropped: the other five fields you read (`wire_count` 32, `compressor_page_count` 32, `internal_page_count` 35, `purgeable_count`) all sit inside `HOST_VM_INFO64_REV1_COUNT` (38), so they're available from the plain imported `vm_statistics64` on any SDK. `total_tag_storage_pages` is at int index 40, past `REV2_COUNT` (40), so it is the single field that forces the raw buffer, the hand-copied offsets and the compatibility layer.

Turns out it does count it, so that trade is justified. Two samples of Activity Monitor about 30 seconds apart on a 24 GB machine, macOS 26.5, 16 KiB pages:

| | sample 1 | sample 2 |
|---|---|---|
| Memory Used | 17.62 GB | 17.78 GB |
| App Memory | 10.75 | 10.95 |
| Wired | 3.00 | 2.97 |
| Compressed | 3.01 | 3.00 |
| **Used − (App + Wired + Compressed)** | **0.86** | **0.86** |

Every component moved between the two samples and the gap did not. Activity Monitor's own displayed breakdown does not add up to its own Memory Used, and the shortfall is stable, so it is counting a fourth bucket it doesn't show.

`total_tag_storage_pages` is exactly 49152 pages x 16384 = 805306368 bytes = 768 MiB on this machine, and it did not move across three readings either. A fixed reservation matching a fixed gap.

So dropping the field would leave the number systematically about 0.75 GB low, which is the exact class of mismatch this PR exists to fix. Not worth suggesting.

One thing I could not account for, in case it's useful. 0.86 - 0.75 leaves roughly 0.11 GB, more than display rounding can explain (four values at two decimals, so at most ~0.02). My own per-component readings also don't line up exactly with Activity Monitor's: I read App as `internal_page_count - purgeable_count` = 11.10 GB where it showed 10.75, and wired 2.72 where it showed 3.00. So its component definitions differ slightly from the obvious mapping, and the residual probably lives there rather than in a sixth bucket. If you already know where its App Memory and Wired differ, that would close the last 0.11.

Measurement code, in case you want to run it on your own machine: plain `host_statistics64(HOST_VM_INFO64)` with `HOST_VM_INFO64_COUNT`, reading through the imported struct, then compared against the numbers at the bottom of Activity Monitor's Memory tab.

PathGao:
Two follow-ups: what that field actually is, and a way to shrink the hand-copied part from six offsets to one.

## What `total_tag_storage_pages` is

It's the reservation for ARM Memory Tagging (MTE), and the size is not arbitrary. On this machine:

```
physical memory   25,769,803,776 bytes  (24 GiB)
total_tag_storage    805,306,368 bytes  (measured, 49152 pages x 16 KiB)
24 GiB / 32          805,306,368 bytes  (predicted)
```

Exact, to the byte. MTE stores a 4-bit tag per 16-byte granule, so 4/128 = 1/32 of RAM. `sysctl hw.optional.arm.FEAT_MTE` returns 1 here. Chip is an M5 (Mac17,3), macOS 26.5.

That explains the behaviour I reported earlier: it never moves because it's a boot-time reservation, which is why Activity Monitor's gap was 0.86 GB in both samples while every other component drifted. It also belongs in "used" in a way file cache doesn't. Nothing can reclaim it short of turning the feature off.

**A prediction worth testing before you invest more in this, especially if you're not on an M5.** On hardware where `FEAT_MTE` is 0, I'd expect `total_tag_storage_pages` to be 0, and therefore App + Wired + Compressed to match Activity Monitor's Memory Used exactly, with no gap to explain. If that's your machine, you cannot reproduce the discrepancy this PR fixes, and the tagged-storage term is a no-op for you. Worth a quick check:

```
sysctl hw.optional.arm.FEAT_MTE
```

I've only measured one machine, so this is a prediction, not a finding. If it holds, it's also a fair argument that the term is correct to include: it's zero where it doesn't apply and exactly right where it does.

## Shrinking the raw part to one field

Of the six fields you read, five are inside `HOST_VM_INFO64_REV1_COUNT` (38 ints): `wire_count`, `purgeable_count`, `compressor_page_count` (int 32), `external_page_count`, `internal_page_count` (int 35). Those have been in the header for years and are available from the imported struct on any SDK you support. Only `total_tag_storage_pages`, at int 40, sits past `REV2_COUNT` and forces the raw path.

So the buffer still has to be over-allocated, but only one read has to bypass the struct:

```swift
let buf = UnsafeMutableRawPointer.allocate(byteCount: 248, alignment: 8)
var count = mach_msg_type_number_t(62)
host_statistics64(host, HOST_VM_INFO64, buf.assumingMemoryBound(to: integer_t.self), &count)

let stats = buf.load(as: vm_statistics64.self)   // compiler-checked, old SDKs included
stats.wire_count
stats.compressor_page_count
stats.internal_page_count
stats.purgeable_count
stats.external_page_count

// the only hand-copied offset left
let tagStorage = hasRev3 ? buf.load(fromByteOffset: 160, as: UInt64.self) : 0
```

The `load(as:)` is safe regardless of which SDK compiled it: fields are only ever appended to this struct, so its size on an older SDK can only be smaller than the 248 bytes allocated. I have only the 26.5 SDK locally, so I could not check 15.2's exact `sizeof`, but the argument doesn't depend on the number.

That takes it from six unguarded offsets to one, and the one that's left is the only field that genuinely cannot be named on the old SDK. If you still want the compile-time layout check, it now has a single number to defend instead of six, and it can be conditional: assert `MemoryLayout` agreement when built against an SDK that has the field, skip it otherwise.

Not a disagreement with your plan, just a smaller surface for it to cover.

pergioa:
@PathGao, thank you for checking both the ABI and the Activity Monitor accounting independently.

The implementation is now in `9f7cf4c`. The compatibility layout lives in a header-only C module. Its five shared-field offsets are `_Static_assert`ed against the active SDK’s real `vm_statistics64_data_t`; when the SDK exposes rev3, the revision count and `total_tag_storage_pages` offset are also asserted directly against the real struct. This is intended to catch exactly the self-referential-wrapper failure mode you pointed out.

Swift now receives a typed snapshot, uses the kernel-returned count as the revision discriminator, rejects truncated payloads, and ignores a tagged-storage value larger than physical memory. Tests cover legacy, truncated, rev2, rev3, and invalid tagged-storage inputs. All 8,222 checks pass, as do the full development and optimized release builds.

Your measurements are also useful independent confirmation that dropping `total_tag_storage_pages` would preserve the systematic undercount this PR is fixing. I agree that the remaining ~0.11 GB is more likely related to Activity Monitor’s definitions of App Memory and Wired Memory—or sampling differences—than another missing bucket. I do not think we should alter this PR’s formula based on that residual alone; exact component parity would be better investigated separately with synchronized samples before changing the accounting again.

Thanks again for correcting #837 and for validating the offsets and tagged-storage behavior.

PathGao:
Our comments crossed by about 17 seconds, so this is a review of `9f7cf4c` plus the part of mine you would not have seen.

`9f7cf4c` is better than what I was about to suggest. I was going to propose keeping five of the six fields on the imported struct and leaving only `total_tag_storage_pages` on a raw offset, taking it from six unguarded offsets to one. Declaring the whole rev3 layout in C and letting the C compiler compute every offset takes it to zero, and the `offsetof(ours) == offsetof(vm_statistics64_data_t)` pairs anchor it to the SDK's real struct rather than to a second copy of the same numbers. That's the failure mode closed properly. Ignore my earlier suggestion.

`decode(_:returnedCount:)` being a pure function over a fabricated struct is the right seam too, for a reason that turns out to matter, below.

## Two small gaps left in the assertion set

Both are unconditional, both compile against the current SDK (I built them), and both are two lines.

**1. The prefix is anchored up to offset 140, not up to 160.** You assert `wire_count`, `purgeable_count`, `compressor_page_count`, `external_page_count` and `internal_page_count` against the real struct. Between `internal_page_count` (140) and `total_tag_storage_pages` (160) sit `total_uncompressed_pages_in_compressor` (144) and `swapped_count` (152), which are not anchored. They exist in rev2, so they are present on Xcode 16's SDK and can be checked without the `#ifdef`:

```c
_Static_assert(offsetof(vorssaint_vm_statistics64_rev3_t, total_uncompressed_pages_in_compressor) ==
                   offsetof(vm_statistics64_data_t, total_uncompressed_pages_in_compressor),
               "total_uncompressed_pages_in_compressor layout mismatch");
_Static_assert(offsetof(vorssaint_vm_statistics64_rev3_t, swapped_count) ==
                   offsetof(vm_statistics64_data_t, swapped_count),
               "swapped_count layout mismatch");
```

Without these, the only thing pinning offset 160 on an Xcode 16 build is `_Static_assert(offsetof(ours, total_tag_storage_pages) == 160)`, which checks your struct against a constant you also wrote. A size-compensating typo in those two `uint64_t`s would pass it. Unlikely, but these close it for free and on the SDK that CI actually gates on.

**2. `REV1_COUNT` and `REV2_COUNT` are derived but never checked against the SDK.** You anchor `REV3_COUNT` inside the `#ifdef`, but the other two are computed from your own struct and never compared to Apple's macros, which do exist on older SDKs:

```c
_Static_assert(VORSSAINT_HOST_VM_INFO64_REV1_COUNT == HOST_VM_INFO64_REV1_COUNT,
               "HOST_VM_INFO64 rev1 count mismatch with the active SDK");
_Static_assert(VORSSAINT_HOST_VM_INFO64_REV2_COUNT == HOST_VM_INFO64_REV2_COUNT,
               "HOST_VM_INFO64 rev2 count mismatch with the active SDK");
```

This one has teeth beyond tidiness: `rev1Count` is the truncation guard in `decode`, so it decides whether a short payload is rejected. Right now that boundary is self-derived.

## What the field actually is, and why your machine may matter

From my earlier comment, which crossed with yours. `total_tag_storage_pages` is the ARM Memory Tagging (MTE) reservation, and the size is not arbitrary:

```
physical memory   25,769,803,776 bytes  (24 GiB)
total_tag_storage    805,306,368 bytes  (measured)
24 GiB / 32          805,306,368 bytes  (predicted: 4-bit tag per 16-byte granule)
```

Exact to the byte. `hw.optional.arm.FEAT_MTE` is 1 here; the chip is an M5 (Mac17,3), macOS 26.5. It is a boot-time reservation, which is why it never moved across readings and why Activity Monitor's gap was 0.86 GB in both samples while every component drifted.

The practical consequence for you: **on hardware where `FEAT_MTE` is 0 I would expect this field to be 0**, so App + Wired + Compressed would match Activity Monitor exactly and there would be no discrepancy to reproduce. Worth checking which side you are on:

```
sysctl hw.optional.arm.FEAT_MTE
```

I have measured one machine, so that is a prediction rather than a finding. It does not block anything, precisely because `decode` takes the struct as a parameter, so the rev3 path is covered by fabricated inputs rather than by hardware you may not have. But it is worth knowing that the live path may be a no-op on your Mac and non-zero on a reviewer's.

Agreed on leaving the ~0.11 GB residual alone. It is a component-definition question, not a missing bucket, and it needs synchronized sampling to answer rather than a change to this formula.

pergioa:
@PathGao, thanks for the follow-up. I kept the full C rev3 layout from `9f7cf4c` and added the missing anchors in `48cd709`.

There is one SDK-version qualification: the installed macOS 15 and 15.4 SDK headers expose `total_uncompressed_pages_in_compressor` and `HOST_VM_INFO64_REV1_COUNT`, but they do not expose `swapped_count` or `HOST_VM_INFO64_REV2_COUNT`. The new checks therefore:

- unconditionally anchor `total_uncompressed_pages_in_compressor` and `REV1_COUNT`;
- conditionally anchor `swapped_count` and `REV2_COUNT` when the SDK defines rev2;
- retain the conditional rev3 count and tagged-storage offset checks.

I compiled the header against SDK 15, 15.4, and 26, type-checked the Swift decoder against both SDK 15 variants, ran all 8,222 unit checks, and passed the optimized release build.

Your MTE prediction also reproduces exactly on this machine: M5 Pro (Mac17,9), 64 GiB RAM, 16 KiB pages, `hw.optional.arm.FEAT_MTE = 1`, and 131,072 tag-storage pages. That is 2 GiB, exactly `64 GiB / 32`. So we now have the same 1/32 reservation result on two different M5 configurations.

Thanks for identifying the remaining assertion boundaries and for the independent MTE analysis.

vorssaint:

Issue #569, opened 2026-08-09
Auto-adapt Tab bar of Scratchpad to various scratchpad sizes


AB-boi:
### Description

The tab bar doesn't resize to use all the width it can when the scratchpad window size is altered:

<img width="736" height="139" alt="Image" src="https://github.com/user-attachments/assets/e33a84e4-06fd-4eb9-a562-c3074060ee13" />

### Steps to Reproduce

Launch Scratchpad
Resize the pad

### Vorssaint Version

3.3.1 (77)

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

AB-boi:
Hey @vorssaint this issue still exists in the latest version 🥲

vorssaint:
hello, @AB-boi!

sorry for that! if possible, could you send another screenshot/video recording showing how is it behaving in 3.3.2? after 3.3.2, i cant reproduce this anymore
tested in macos 26 tahoe and macos 27 golden gate. i did not test in sequoia, just want to confirm if its the exact same problem first

https://screenshots.vorssaint.com/s/2l_WNYaRJdUTdVMQR60_37_kmY1NhJrG

AB-boi:
> https://screenshots.vorssaint.com/s/2l_WNYaRJdUTdVMQR60_37_kmY1NhJrG

This link has expired. Could you resend?

My recording (please ignore the struggle at the beginning of the video to resize the window, for some reason screen recording was interfering with it)

https://github.com/user-attachments/assets/51017d89-51ea-4c10-9b9a-0bf8eae3b1b0

PathGao:
Tracking PR: #764 (open, `ready to merge`).

It gives the tab strip the width its row already has (the missing `.fixedSize()` on `.menuStyle(.borderlessButton)`), plus a repo-wide scan assertion that catches the same pattern at the other call sites. Not merged yet.

PathGao:
Merged，please test on next release

AB-boi:
All of it is fixed. Closing this issue.

PR #764, opened 2026-08-22
fix(scratchpad): give the tab strip the width its row has free

PathGao:
Ref #569

## What was wrong

The Scratchpad tab strip does not grow with the pad (#569). In the reporter's screenshot the pad is 736pt wide, the strip stops around 355pt, tab names are cut on both ends, and the space to the right of the pad actions button is empty.

The strip already asks for the room: `ScrollView(.horizontal)` with `.frame(maxWidth: .infinity)`. Its row-mate does not give it up. A `Menu` with `.menuStyle(.borderlessButton)` and no size of its own is greedy on macOS 15 — it takes the free width of its HStack and draws its 22pt label at the leading edge of that claim, which is exactly the empty gap in the screenshot. macOS 26 and 27 size the same menu to its label, which is why this could not be reproduced there while the reporter (Sequoia) still sees it.

## What changed

`.fixedSize()` on the pad actions menu, so it is the size of its label and the strip gets the rest of the row. One line plus its reason.

Measured on macOS 26 with an ImageRenderer probe of the same row: with and without `.fixedSize()` the plus and menu buttons land in identical positions, so this is inert where the menu already behaved and only changes the Sequoia layout.

**`.layoutPriority(1)` on the ScrollView.** The scroll view is flexible up to infinity, so serving it first lets it swallow the row and push the two buttons out when there are many tabs.

**A fixed width on the menu (`.frame(width: 22)`).** Same effect today, but it restates a number the label already owns, and a different label would silently clip.

**Raising the 120pt per-tab cap.** A different complaint: the cap is what keeps one long pad name from eating the strip, and it is not what leaves the space beside the menu empty.

## Not touched

Tab sizing (`minWidth: 46, maxWidth: 120`), the scroll-to-selected behaviour, and the plus button. The header row above the tabs is a separate view.

Tests: 8058 checks pass (8057 on this base), `./build.sh` release build clean. The check is a sweep rather than one assertion about this file: every `.menuStyle(.borderlessButton)` in `Sources/Vorssaint/UI` must carry `.fixedSize()` somewhere in its own modifier chain, so the next borderless menu written cannot repeat #569. Kill Process is the one deliberate exception — its row controls take a shared minimum width so the Kill button and the menu beside it line up down the list. Removing `.fixedSize()` from the Scratchpad menu turns it red, naming the file and line.

I cannot run macOS 15, so the Sequoia behaviour itself rests on the screenshot and on the version split in the reports; @AB-boi can confirm on a build from this branch.


AB-boi:
@PathGao what would you like me to do? I'm not a dev/coder, but I have decent tech knowledge. Just tell me if I need to test any beta or something.

PathGao:
Thanks @AB-boi — no beta needed, and no coding. There is one 10-second check on the version you already have that would settle this.

Open Scratchpad, widen it until the empty strip appears to the right of the `…` button, then **click once in that empty area** — a few centimetres to the right of the button, not on it.

- **A menu pops open (Rename / Close)** → confirmed. That button is quietly claiming the whole empty strip for itself, which is why the tabs never get the width. Taking that claim away is exactly what this PR does.
- **Nothing happens** → please say so. It does not rule the theory out, but it tells me to keep looking.

Two more things that would help either way:

1. Your exact macOS version (Apple menu → About This Mac, e.g. 15.6.1). The reason I cannot reproduce this is that macOS 26 and 27 lay that button out differently, so the exact 15.x matters.
2. A screenshot of the pad narrow and the same pad wide, so I can see how far the tab strip's right edge moves when the window grows.

A test build does exist (every CI run on this PR attaches one), but it would replace your installed Vorssaint with an ad-hoc signed copy and you would have to grant its permissions again, so I would rather not send you down that road — @vorssaint folding the one-line change into the next build is the cleaner way to confirm it on your Mac.

AB-boi:
[Recording](https://github.com/user-attachments/assets/25d75634-4c9d-4dff-8451-ecac2c856169)
All answered ❤️
NOTE: Ignore the weird cursor behaviour when resizing edges, only happens when screen recording.

macOS 15.7.9 (24G830)

@PathGao

PathGao:
Confirmed by your recording, thank you — that was the decisive frame:

The cursor is roughly 350pt to the right of the `…` button, in what looks like empty space, and the **Rename / Close scratchpad** menu opens there. So that button is not 22pt wide: on macOS 15.7.9 it is claiming everything from its glyph to the right edge of the pad, and it only draws the three dots at the left end of that claim. The tab strip beside it gets whatever is left, which is why widening the pad never gives the tabs more room.

That is exactly what this PR takes away, so your recording is the missing half of the diagnosis.

While confirming it I checked the other eight borderless menus in the app — every one of them already pins itself the way this PR now pins the Scratchpad one, so this was the single member that never joined an existing convention. I pushed a second commit that turns that convention into a test, so a future menu cannot bring the bug back.

Nothing further needed from you. @vorssaint — the layout claim is now verified on 15.7.9 rather than inferred; the recording is in the comment above.

Issue #191, opened 2026-07-09
[Bug]: Extra brightness not available on M3 Max MBP


dwax05:
### Description

The extra brightness feature is not available on my macbook pro on Sequoia 15.7.7. The display is set to Apple XDR Display in settings

### Steps to Reproduce

1. Open Vorssaint
2. be on 16 inch m3 max mbp on Sequoia 15.7.7
3. be sad

### Vorssaint Version

3.1.9

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

vorssaint:
Looking into it!

Thanks for the report!

acdi33:
Does not work for me either. Running version 3.1.9 on macOS Tahoe 26.5.2 on a MacBook Pro M4 Pro.

<img width="838" height="511" alt="Image" src="https://github.com/user-attachments/assets/9166de02-a8ef-43f2-a948-16e27c965720" />

vorssaint:
@acdi33 @dwax05 

Update 3.1.10 should fix it. Can you guys update and tell me here if it works properly now?

acdi33:
@vorssaint The toggle and controls appears, but if I turn it on it seems to work and brighten the screen for one second but then the effect flickers off back to its usual brightness. The controls indicate that it's still on though; if I turn off and on the toggle the behavior repeats (screen brightens for one second then darkens back to its usual state).

vorssaint:
fixed @acdi33 

can you update and try it out?

acdi33:
@vorssaint seems to work now on my end with the update! 🎉

PR #765, opened 2026-08-22
fix(display): resolve display names on the main thread

PathGao:
Follows #762, which fixed the display power transaction and named this one as the separate change left behind. Independent of it: different functions, no shared implementation path, either can merge first.

## What was wrong

`displayName(_:info:)` resolved a display's user-facing name through `NSScreen.screens`. Its only caller is `rebuild(generation:previousDisplays:)`, which runs on the private serial `workQueue`, so every topology rebuild — hotplug, wake, panel open, Settings appearing — read AppKit from a background thread.

`NSScreen.screens` is not a hardware query. It returns AppKit's internal array of `NSScreen` objects, which the main thread rebuilds wholesale on a display configuration change: the old objects are released and new ones take their place. AppKit does not lock it, because the contract is that only the main thread touches it.

The collision window is the same event on both sides. AppKit rebuilds the array, then posts `NSApplicationDidChangeScreenParametersNotification`; `screensChanged()` debounces that by 0.5s and calls `refresh()`, which dispatches the rebuild onto the work queue. A hotplug or a wake is not one event — the displays keep negotiating for seconds, so further parameter changes arrive while the rebuild is still running. And the rebuild is slow by design: a DDC/CI transaction per external display, with a retry loop and `usleep` pauses, so it sits on that array for exactly as long as the topology is in motion.

The shapes this surfaces as are all hard to attribute: a crash whose stack lands inside `NSScreen.screens` or `localizedName` rather than anywhere near "a display was plugged in", a Main Thread Checker report in debug, or a slider labelled with the wrong display when a half-updated array is read.

## What changed

`refresh()` snapshots the whole set on the main thread before it dispatches, and hands the map to the rebuild:

```swift
let screenNames = Self.localizedScreenNames()   // NSScreen, main thread
workQueue.async { [weak self] in
    self?.rebuild(generation: generation, previousDisplays: previousDisplays,
                  screenNames: screenNames)
}
```

`displayName(_:info:screenNames:)` looks names up in that map. Its `DisplayProductName` → `"Display"` fallback chain is untouched, and every `refresh()` call site is already on the main thread (`start()`, `screensChanged()`, `displaysWokeUp()`, `finishDisplayToggle`, the DDC write failure hop, `SettingsView`, `CommandBarCatalog.applyBrightness`).

Effect: the names on the sliders are unchanged — that is the point, this is a threading fix and should be invisible in the UI. What changes is that the work queue no longer reads an array the main thread is rewriting.

**Dropping `NSScreen` and using the `DisplayProductName` already read from the display info dictionary.** The smallest diff, and it needs no snapshot — but it changes what people see. AppKit reports "Built-in Retina Display" for a built-in panel where the EDID product name is "Color LCD", and external monitors move from their localized System Settings name to their raw EDID string. A threading fix should not rename anyone's displays.

**Reusing the answer in `ExtraBrightnessSupport.swift:48`.** That note explains why an XDR *panel* cannot be identified by its NSScreen name (AppKit calls every built-in panel "Built-in Retina Display", which was the 3.1.9 bug in #191). It is about identification, not labelling, so there was nothing to share — and it is in fact the reason the generic AppKit name is the right label here.

**Hopping to the main thread from inside `rebuild()` with `DispatchQueue.main.sync`.** Reintroduces the deadlock geometry #762 removed, once per display, in the middle of DDC work.

## Not touched

The other `NSScreen` read in the file, in `handleKeyEvent`: its tap's run loop source is on `CFRunLoopGetMain()`, so it is already main-thread, and the function key tap that does run on its own thread reaches no AppKit through `routeFunctionKey`. The snapshot is taken before the queue hop, so a display appearing in that gap is named from `DisplayProductName` for one rebuild — a real name, and the parameter change that brought it in queues another rebuild immediately. No user-facing strings.

Tests: 8058 checks pass (8057 before), full `./build.sh` clean with no warnings, `--selftest` OK. One check kills the class rather than the instance: every section below the service's `// MARK: - Rebuild (work queue)` runs on that queue, and none of them may reach `NSScreen`. Comments are stripped before the check, so a note naming the symbol it bans does not count as a call — verified both ways: a prose mention stays green, a real `NSScreen.screens` call goes red.


vorssaint:

PR #766, opened 2026-08-22
feat(snippets): add a visual date/time builder

tenbux:
## Summary

Putting a date or time into a snippet meant knowing DateFormatter's pattern syntax. This adds a popover that picks the type and the style, previews the result against the current time, and can pin a timezone searchable by identifier or by common abbreviation. The existing `{{date:...}}` syntax is unchanged and still works typed by hand.

It is built on `DateFormatter` and `TimeZone`, both already used here, and adds no dependency and nothing that runs at rest: the popover exists only while open. Permanent surface is one new UI file and the token grammar below.

**This changes what a saved snippet can contain.** A `-tz(...)` part is new syntax, so a snippet using it renders literally on any build older than this. Both format captions document it in all 13 languages. An identifier this Mac cannot resolve leaves the tag literal rather than quietly formatting in the device's own zone, so a typo stays visible the way every other unknown tag does.

Named styles (Short through Full) resolve to the pattern the Mac's region uses at the moment you build the token, and that literal is what gets saved. The popover says so, since the labels on their own suggest the opposite.

Inserting at the caret needs the caret, which SwiftUI's `TextEditor` only reports from macOS 15. The shared plain-text editor from #944 gains that as an optional binding, tracked as character offsets so an insert cannot read positions measured against text that has since changed.

## Testing

```sh
./build.sh --test              # 1 of 8663 fails, see below
```

The one failing check is `the dispatch pool starvation this check needs was actually reached`, which is not from this branch: it fails identically on a clean `main` checkout here, and passes on `main` from before #989 added it. It occupies 64 global-queue threads and asserts the pool is genuinely starved; on a machine with more cores that precondition is not met, so the assertion fails rather than the behavior under test. Everything else passes.

Checked on Apple Silicon against a release build, in English with a Gregorian regional calendar. The builder was used by hand earlier in development: typing timezone queries, inserting tokens and reopening them to edit. Since that pass the insert was rewritten to go through the text view, the timezone list was ranked and uncapped, and timezones were allowed on the Date type, and those three are covered by tests rather than by use.

Not tried on macOS 14 or 15, which matters more here than usual: the caret tracking exists because `TextEditor(text:selection:)` needs macOS 15. Also not tried with a non-Gregorian regional calendar or another app language.

## Anything else

Rebased onto current `main` now that #944 has merged, so this is a single commit against it.

### DIFF
--- a/Sources/Vorssaint/Core/SnippetStrings.swift
+++ b/Sources/Vorssaint/Core/SnippetStrings.swift
@@ -43,6 +43,31 @@ struct SnippetFeatureStrings {
     let folderPlaceholder: String
     let showInLibraryLabel: String
     let variablesFormatCaption: String
+    let editorFormatCaption: String
+    let dateTimeInsertButton: String
+    let dateTimeEditButton: String
+    let dateTimeTypeLabel: String
+    let dateTimeKindDate: String
+    let dateTimeKindTime: String
+    let dateTimeKindDateTime: String
+    let dateTimeStyleLabel: String
+    let dateTimeStyleShort: String
+    let dateTimeStyleMedium: String
+    let dateTimeStyleLong: String
+    let dateTimeStyleFull: String
+    let dateTimeStyleISO8601: String
+    let dateTimeStyleCustom: String
+    let dateTimeStyleLocaleNote: String
+    let dateTimeTimezoneLabel: String
+    let dateTimeTimezoneDeviceDefault: String
+    let dateTimeTimezoneValid: String
+    let dateTimeTimezoneInvalid: String
+    let dateTimeTimezoneClear: String
+    let dateTimeTimezoneSearchPlaceholder: String
+    let dateTimePatternLabel: String
+    let dateTimePreviewLabel: String
+    let dateTimeConfirmInsert: String
+    let dateTimeConfirmUpdate: String
 }
 
 extension FeatureStrings {
@@ -102,7 +127,32 @@ extension SnippetFeatureStrings {
         folderLabel: "폴더",
         folderPlaceholder: "업무",
         showInLibraryLabel: "빠른 메뉴에 표시",
-        variablesFormatCaption: "콜론 뒤에 형식을 적으면 원하는 모양이 됩니다. 예: {{date:yyyy-MM-dd}}"
+        variablesFormatCaption: "콜론 뒤에 형식을 적으면 원하는 모양이 됩니다. 예: {{date:yyyy-MM-dd}}. 시간대는 {{date-tz(America/New_York):yyyy-MM-dd}}처럼 -tz(...)로 지정합니다.",
+        editorFormatCaption: "콜론 뒤에 형식을 적으면 원하는 모양이 됩니다. 예: {{date:yyyy-MM-dd}}. 또는 위의 날짜/시간 버튼을 사용하세요. 시간대는 {{date-tz(America/New_York):yyyy-MM-dd}}처럼 -tz(...)로 지정합니다.",
+        dateTimeInsertButton: "날짜/시간 삽입",
+        dateTimeEditButton: "날짜/시간 편집",
+        dateTimeTypeLabel: "유형",
+        dateTimeKindDate: "날짜",
+        dateTimeKindTime: "시간",
+        dateTimeKindDateTime: "날짜시간",
+        dateTimeStyleLabel: "스타일",
+        dateTimeStyleShort: "짧게",
+        dateTimeStyleMedium: "보통",
+        dateTimeStyleLong: "길게",
+        dateTimeStyleFull: "전체",
+        dateTimeStyleISO8601: "ISO 8601",
+        dateTimeStyleCustom: "사용자 지정",
+        dateTimeStyleLocaleNote: "이름이 있는 형식은 Mac의 지역 표기를 지금 그대로 저장합니다.",
+        dateTimeTimezoneLabel: "시간대",
+        dateTimeTimezoneDeviceDefault: "기기 기본값",
+        dateTimeTimezoneValid: "유효한 시간대",
+        dateTimeTimezoneInvalid: "인식할 수 없는 시간대",
+        dateTimeTimezoneClear: "시간대 지우기",
+        dateTimeTimezoneSearchPlaceholder: "시간대 검색",
+        dateTimePatternLabel: "패턴",
+        dateTimePreviewLabel: "미리보기",
+        dateTimeConfirmInsert: "삽입",
+        dateTimeConfirmUpdate: "업데이트"
     )
 }
 
@@ -143,7 +193,32 @@ extension SnippetFeatureStrings {
         folderLabel: "Folder",
         folderPlaceholder: "Work",
         showInLibraryLabel: "Show in the quick menu",
-        variablesFormatCaption: "A format after a colon picks how they look, like {{date:yyyy-MM-dd}}."
+        variablesFormatCaption: "A format after a colon picks how they look, like {{date:yyyy-MM-dd}}. A -tz(...) part sets the timezone, like {{date-tz(America/New_York):yyyy-MM-dd}}.",
+        editorFormatCaption: "A format after a colon picks how they look, like {{date:yyyy-MM-dd}}, or use the date/time button above. A -tz(...) part sets the timezone, like {{date-tz(America/New_York):yyyy-MM-dd}}.",
+        dateTimeInsertButton: "Insert date/time",
+        dateTimeEditButton: "Edit date/time",
+        dateTimeTypeLabel: "Type",
+        dateTimeKindDate: "Date",
+        dateTimeKindTime: "Time",
+        dateTimeKindDateTime: "DateTime",
+        dateTimeStyleLabel: "Style",
+        dateTimeStyleShort: "Short",
+        dateTimeStyleMedium: "Medium",
+        dateTimeStyleLong: "Long",
+        dateTimeStyleFull: "Full",
+        dateTimeStyleISO8601: "ISO 8601",
+        dateTimeStyleCustom: "Custom",
+        dateTimeStyleLocaleNote: "A named style saves the format your Mac's region uses right now.",
+        dateTimeTimezoneLabel: "Timezone",
+        dateTimeTimezoneDeviceDefault: "Device default",
+        dateTimeTimezoneValid: "Valid timezone",
+        dateTimeTimezoneInvalid: "Unrecognized timezone",
+        dateTimeTimezoneClear: "Clear timezone",
+        dateTimeTimezoneSearchPlaceholder: "Search timezones",
+        dateTimePatternLabel: "Pattern",
+        dateTimePreviewLabel: "Preview",
+        dateTimeConfirmInsert: "Insert",
+        dateTimeConfirmUpdate: "Update"
     )
 
     static let ptBR = SnippetFeatureStrings(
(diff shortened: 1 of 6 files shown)


PathGao:
Read this at line level now. Correcting something first: an earlier note in #837 said this PR adds a second `NSTextView` bridge alongside `ScratchpadView.PlainTextEditor`. **It does the opposite** — it lifts the existing `private struct PlainTextEditor` out of `ScratchpadView.swift` into a shared `UI/PlainTextEditor.swift`, parameterizes it (`selectedRange`, `textColor`, `textContainerInset`, `onCreate`) and reuses it. `ScratchpadView.swift` loses 67 lines. Scratchpad behaviour is preserved: font `13` becomes `NSFont.systemFontSize` (13.0 on macOS), inset and `registerTextView` pass through. That is a de-duplication and it is welcome.

The token grammar is also handled correctly: one parser (`formattedPattern`), one expander, both updated together, and 152 lines of tests against `TextSnippetSupport`, which is already in `build.sh`'s allowlist. Localization is complete — 22 fields × 13 initializers.

**The scope is where this needs work.** The title says "a visual date/time builder", and the PR also adds a new persistent expansion syntax — `{{date-tz(America/New_York):yyyy-MM-dd}}` — plus timezone search and resolution. That is a format change, not a UI affordance: a snippet saved with `-tz(...)` renders literally on any build that predates this. It deserves its own PR, or at minimum its own paragraph in the description and a line in the two format captions. Neither `variablesFormatCaption` (`TextSnippetsSettings.swift:85`) nor the new `editorFormatCaption` (`:313`) mentions it, so today the syntax is reachable only through the popover and documented nowhere.

Two risks worth a look:

- **A stale `String.Index` can trap.** `SnippetEditor.replacementSelection` holds a `Range<String.Index>` in `@State`. `PlainTextEditor.updateNSView` sets `isApplyingExternalText` and replaces `textView.string` without clearing that binding, so after a programmatic text replace the stored indices belong to a string that no longer exists; `insertDateToken` then passes them to `replaceSubrange`, which traps if the new text is shorter. `insertDateToken` clears the selection before mutating, which covers the normal path — the external-replace path is the one that is not covered. I could not run this to confirm it triggers in practice.
- **The Style picker freezes a locale's pattern.** `resolvedDatePattern` for `.short`/`.medium`/`.long`/`.full` returns `DateFormatter.dateFormat` — the ICU pattern resolved for the *current* locale — and embeds that literal in the saved token. A token built as "Full date" under one language stays that language's pattern forever. `matchingDateStyle`'s own doc comment acknowledges the trade, but the picker's labels imply the opposite to the user.

Smaller: `matchingTimeZoneIdentifiers` scans `TimeZone.knownTimeZoneIdentifiers` plus `abbreviationDictionary` and does a `Set` + `sorted()` on every render, and `body` calls it twice. Popover-scale, so probably invisible. And `matchingTimeZoneIdentifiers(for: "new york") == ["America/New_York"]` asserts on the OS tz database — true today, version-coupled tomorrow.

**Merge order with #767:** let #767 go first. You conflict only in `Core/SnippetStrings.swift`, where its side is three appended fields and yours is twenty-two — landing the small one first makes your rebase a mechanical append.

Labelling `incomplete`, for the undisclosed `-tz(...)` syntax more than anything else.

PathGao:
Three more from a line-level pass over the extracted editor and the builder. My earlier points stand; these are separate.

**1. `makeNSView` has the same hazard `updateNSView` guards against.** `PlainTextEditor.swift:39` assigns `textView.delegate = context.coordinator`, and `:59` then does `textView.string = text` with no guard. `updateNSView` wraps its own identical assignment in `isApplyingExternalText` (`:74-76`), and your comment there states the reason: setting `.string` fires `textViewDidChangeSelection`, and that callback writes SwiftUI state, which is undefined behaviour from inside a view update. `makeNSView` is a view update too, and `Coordinator.textViewDidChangeSelection` (`:102-107`) writes both bindings — `text.wrappedValue` and `selectedRange?.wrappedValue`. One site of the shape is guarded and the other is not.

Either set `isApplyingExternalText` around `:59` as well, or move `:39` below `:59` so the delegate is not attached yet when the initial string lands. This is not specific to the snippet editor — Scratchpad goes through the same initializer, and on `main` its coordinator had no selection callback at all, so this is new behaviour for that screen too. I could not run it to see whether AppKit posts the notification for the first assignment on an empty view.

**2. `{{date-tz(...)}}` is a token the builder cannot build.** `DateVariableBuilder.swift:49` hides the timezone picker for `.date`, and `:82-84` clears `timeZoneIdentifier` whenever the type changes to `.date`. Meanwhile `formattedPattern` accepts `{{date-tz(...):...}}` (`TextSnippetSupport.swift:181-189`) and your own doc comment at `TextSnippetSupport.swift:135-137` uses `{{date-tz(America/New_York):yyyy-MM-dd}}` as *the* example of the new syntax. A date is exactly where a timezone is most visible to a user — it is what decides which calendar day gets written — so the restriction reads as an oversight rather than a decision.

There is a second-order effect: `.onChange(of: kind)` does not fire during `init`, so opening the builder on an existing `{{date-tz(Asia/Tokyo):yyyy-MM-dd}}` keeps the identifier in `@State` while the picker that would show it is hidden, and Update rewrites the token with a timezone the user could not see. Switching the type away and back then silently drops it. Pick one side: allow the timezone for `.date`, or make `formattedPattern` reject `date-tz` so the grammar and the builder agree.

**3. The timezone list truncates the wrong end.** `TextSnippetSupport.swift:339` does `Array(Set(...)).sorted().prefix(50)` — the cap is applied *after* an alphabetical sort of the whole match set, not after any ranking. There are roughly 140 `America/*` identifiers, so a user typing "america" gets everything up to about `America/Costa_Rica` and `America/New_York` is not in the list at all. `resolvedTimeZoneIdentifier` is unaffected (it needs count == 1), so the search field's checkmark and the visible list disagree for exactly these queries. Either rank exact and prefix matches ahead of substring matches before capping, or drop the cap and let the `ScrollView` at `DateVariableBuilder.swift:126-142` do its job.

tenbux:
Thanks, this was a good catch list. Everything is fixed; the two structural points are worth answering directly.

**The `-tz(...)` syntax.** You are right that it is a format change and not a UI affordance. It is now documented in both format captions in all 13 languages, and the description has a paragraph on it including that older builds render such a snippet literally. I did not split it into its own PR: the builder's headline case is an ISO 8601 timestamp in a fixed zone, so a builder without timezones would ship an affordance for the syntax people least need help with. Happy to split it if you would rather review them apart.

**The stale `String.Index`.** Real, and worse than a trap: an index read against a string it was not measured from is undefined behavior, so no validation makes it safe after the fact. Both the editor's selection binding and `DetectedDateToken` now carry character offsets, clamped where they are used. The insert also goes through the text view rather than the binding now, which keeps it undoable and leaves the caret after the token instead of at the end.

The style picker's locale trade is now stated in the popover for the four styles it applies to. The timezone search runs once per render instead of three times. The test that pinned the OS timezone database asserts containment instead.

From your second pass: the delegate in `makeNSView` is assigned after the initial string, so that path is guarded like `updateNSView`. The timezone picker is available for Date as well, so the syntax the captions document is one the builder can actually produce, and the silent drop on switching type is gone with it. The suggestion list is ranked and no longer capped, so a broad query like "america" reaches the city you typed rather than stopping in the C's.

Two changes you did not ask for: an identifier this Mac cannot resolve now leaves the tag literal instead of quietly formatting in the device's own zone, which matches how every other unknown tag behaves; and `CHANGELOG.md` is out of the branch per rule 8.

The shared editor extraction is now #944, so this no longer touches `ScratchpadView`. This depends on it.

PathGao:
Split it? No. The objection was that the syntax was undocumented, not that it was bundled. Both captions and the description now cover it, so the pair is reviewable as one thing. Splitting now would land a grammar whose only teaching surface sits in a different PR.

Confirmed from the diff:

- `makeNSView` assigns `.string` before `.delegate` (`PlainTextEditor.swift:59-62`), so the initial assignment cannot reach the coordinator.
- Character offsets replace `String.Index` in both `DetectedDateToken` and the editor's selection binding, clamped in `selectionRange(in:offsets:)`.
- The timezone picker is offered for `.date`, and nothing clears the identifier on a type change any more, so the silent drop is gone with it.
- `matchingTimeZoneIdentifiers` ranks exact → city exact → city prefix → identifier prefix → substring, and is uncapped; `timeZoneSearch` is read once per `body`.
- `insertDateToken` goes through `shouldChangeText`/`replaceCharacters`/`didChangeText`. Side benefit worth knowing: `textView.string == text` afterwards, so `updateNSView`'s `removeAllActions()` never fires for a builder insert and the undo stack survives.

Two things.

**The description and your comment disagree about `ScratchpadView`.** The body says "Until that lands, this diff includes it"; the comment says "this no longer touches `ScratchpadView`". The diff I can see still deletes the private `PlainTextEditor` from `ScratchpadView.swift` and adds `UI/PlainTextEditor.swift`. My snapshot may be behind your last push — but whichever is true, state it in one place only, because anyone reviewing #944 needs to know whether the two diffs overlap.

**A comment now says the opposite of its code.** `TextSnippetsSettings.swift`: "A plain reference rather than @State" sits directly above `@State private var replacementEditor = EditorHandle()`. The `@State` is what makes it work — a plain `let` would be re-initialized every time the struct is rebuilt, and the handle would be nil by the time the builder needs it. The reason given for the class is right, the reason given for the storage is inverted, and as written it reads as an invitation to drop the wrapper.

I could not check the test changes: my copy of the diff is truncated partway through `Tests/MetricsTests.swift`, so the containment assertion that replaced the pinned timezone list is unread.

Merge order is now #944, then #767, then this. #944 because you depend on it; #767 because its three appended `SnippetStrings` fields rebase into your twenty-two mechanically in that direction and not the other.

**Impact:** none on existing snippets. A snippet saved with `-tz(...)` renders literally on older builds, which the description and both captions now state.
**Merge:** hold for #944.

tenbux:
Both fixed, and thanks for the correction on the split.

The comment now says what is actually true: `@State` so the handle outlives a view rebuild, and a class so the text view can be stored into it while it is being made without publishing from inside a view update. You are right that as written it argued for deleting the thing that makes it work.

On `ScratchpadView`: the description was right and my comment was wrong, so take the description. This branch's own commit touches six files and that is not one of them, but the PR carries #944's commit underneath until it lands, so GitHub shows seven files including `ScratchpadView.swift`. The overlap is exactly #944 and nothing else, and it disappears from here the moment that merges.

Also folded in the two notes from #944 while it was still unpushed, since they touch the file this depends on.

PathGao:
#944 merged and this is rebased onto it, so the hold is off. Swapping `blocked` for `ready to merge`.

Both fixes confirmed in the diff:

- The `EditorHandle` comment now gives the right reason for each half: `@State` so the handle outlives a rebuild, class so the text view can be stored into it during `makeNSView` without publishing from a view update.
- `ScratchpadView.swift` is out of the diff, so the description and the file list agree.

Withdrawing the #767 ordering. That argument only held while you still had a rebase ahead of you. You've done it, and #767's three appended `SnippetStrings` fields drop into thirteen initializers just as mechanically after this lands as before. Either order is fine now.

Nothing new at the line level. `formattedDateKinds` is the only place that knows the tag set, `configure` is the only place that maps a pattern and an identifier onto a formatter, and the plain and `-tz` prefixes are tried in one loop — no second copy to drift out of sync. The `-tz(...)` sentence does exist twice per language across the two captions, but flat per-language literals is this file's existing shape and I would leave it.

Two things I did not check: my copy of the diff is still truncated inside `Tests/MetricsTests.swift`, partway through the "america" ranking test, so the tail of that file is unread. And "Swift 6.0.3 compatibility" has not reported a result yet.

**Impact:** none on existing snippets. A snippet saved with `-tz(...)` renders literally on older builds, which the description and both captions state.
**Merge:** ship it.

PR #767, opened 2026-08-22
feat(snippets): play a sound when a trigger expands

tenbux:
## Summary

A trigger expansion replaces what you typed with no feedback beyond the text itself, which is easy to miss mid-sentence. This adds an optional short system sound when a typed trigger expands, with a picker that previews each sound as you select it.

The sound is hung off the replacement going out, not off `postExpansion`'s return value: a transient paste reports success as soon as it reaches the pasteboard lane. On the paste path #917 added it fires from the ⌘V's own completion, so a paste that fails open to typing plays nothing and the sound follows the paste rather than leading it.

Off by default, and the controls only appear when text snippets are on. It is built on `NSSound` and the alert sounds already in `/System/Library/Sounds`, so it adds no dependency, no background work and nothing that runs while the toggle is off. Permanent surface is two defaults keys and four strings.

Scoped deliberately to trigger-based expansion. Inserting from the snippet library or the Command Bar is an explicit action that already shows its own result, so a sound there would be noise; the toggle and its caption say "typed trigger" for that reason.

The sound is resolved once when preferences change rather than per expansion. The expansion runs inside the window the event tap blocks the main thread for, and both the first `NSSound(named:)` for a name and starting playback are too slow to sit there, so playback is dispatched after that window closes.

## Testing

```sh
swift build                    # Build complete!
./build.sh --test              # TESTS OK (10712 checks)
```

Clean `origin/main` at the same commit this is based on reports `TESTS OK (10698 checks)`, so this branch adds 14.

Then installed with `./build.sh --dev --install` and used it. A multi-line expansion, which `requiresPaste` sends through `TransientPaste`, lands its text first and the sound follows it, with the beat the modifier-release loop and the 60 ms settle account for. A single-line expansion stays on the typed path and sounds essentially at once. So both routes play, and on the paste route the sound trails the text rather than leading it.

Checked on Apple Silicon against a `--dev --install` build, in English. Not tried on Intel, or in another app language. The unavailable-sound row cannot be produced without hand-editing defaults, so it is covered by tests rather than by use, and I have not measured the tap callback under load to confirm the timeout headroom, only kept the work outside that window.


PathGao:
Read this at line level now. Two things, and one of them is on the event tap's critical path.

**1. The sound resolves inside the tap's blocking window.** `expand`'s `post` closure runs under `DispatchQueue.main.sync` (`TextSnippetService.swift:344-347`) from the CGEvent tap callback, and the `NSSound(named:)` lookup now sits inside it. `play()` returns immediately, but the *first* `NSSound(named:)` for a given name loads the file from `/System/Library/Sounds` synchronously — a disk read added to a window the OS will disable the tap for overrunning.

`syncWithPreferences` already reads the name; resolving and holding the `NSSound` there instead of per expansion removes this entirely, and `stop()` + `play()` on a held instance is exactly what the code comment already describes.

**2. Only one of three expansion paths plays it.** `git grep postExpansion` finds `TextSnippetService.swift:324` (typed trigger), `SnippetLibraryService.swift:205` (quick menu) and `CommandBarExtras.swift:30` (Command Bar). The sound is added only to the first.

Whether that is a bug depends on which of the PR's own claims you take, and it makes both: the toggle label says "Play a sound when a snippet **expands**", the caption directly under it says "each time a **typed trigger** expands", and the README line this PR adds says "an optional sound on expansion". Two of the three promise all three paths. Pick one and make the label match the code.

Smaller:

- The picker's `.onChange` plays a preview *and* calls `syncWithPreferences()`, which calls `reloadSnippets()` and restarts the tap. Arrowing through the sound list tears down and rebuilds the event tap once per sound.
- `expansionSoundName` is assigned only in the `if enabled, hasWork, Permissions.shared.accessibility` branch, so the `else` branch leaves the old value. Harmless while the tap is stopped, but it is state that can be stale relative to the toggle.
- `alertSoundNames` is a hardcoded 14-name list claiming parity with System Settings. The test that each resolves via `NSSound(named:)` catches a macOS release dropping one, but nothing picks up one macOS adds. Enumerating `/System/Library/Sounds` would be the general rule.
- `SettingsDirectory.swift:160` has no keyword, so searching Settings for "sound" does not find this page.

Localization is complete, both Defaults keys are registered, and writing `NSSound` inline is right — `git grep NSSound` shows the repo has no helper to reuse, every other use is `NSSound.beep()`.

**Merge order with #766:** this one should go first. You two conflict only in `Core/SnippetStrings.swift`, where this PR's side is three fields appended to 13 initializers and #766's is twenty-two. Landing this one first turns #766's rebase into a mechanical append; the other way round pushes a 3-field patch through a file that just moved 320 lines.

tenbux:
Thanks, all four are fixed.

The sound is resolved once in a new `syncExpansionSound()` and retained, so no expansion does a lookup. Playback also moved out of the `main.sync` block entirely: it is dispatched after that window closes, since starting CoreAudio inside it was the same hazard as the disk read.

On scope, I took the narrow reading and made the copy match the code: the toggle, the caption and the README now all say a typed trigger. The library and Command Bar paths insert on an explicit action that already shows its result, so a sound there is noise.

Also from your list: the picker no longer reloads snippets or rebuilds the tap on selection, the sound now always mirrors the preference instead of going stale when the feature is off, `alertSoundNames` is read from `/System/Library/Sounds` with the old list kept only as a fallback, and the page registers both new labels as search keywords.

Two things you did not raise that I changed while in there: a stored sound this Mac does not have used to leave the picker blank and the expansion silent, so it now resolves to the default and the missing entry gets its own labelled row; and I removed the `CHANGELOG.md` entry, having belatedly read rule 8.

Agreed on merge order, this one first.

PathGao:
All four are in the diff and read right: the sound is resolved once in `syncExpansionSound()` and retained, playback is dispatched after the `main.sync` window rather than inside it, the picker no longer goes through `syncWithPreferences`, the sync is unconditional so the retained sound can't outlive the toggle, and `alertSoundNames` reads the directory with the old list as fallback. The scope copy matches the code in all thirteen languages.

One thing left, and it is the shape of the original problem rather than a new bug: "is the sound on, and which one" is now decided in three places. `syncWithPreferences` computes `enabled` (just above the diff context), `syncExpansionSound` recomputes `AppFeature.textSnippets.isAvailable && bool(textSnippetsEnabled)` for itself, and `previewExpansionSound` reads the defaults and resolves the name again one line after the same `.onChange` already called `syncExpansionSound`. The two gates are equal only by inspection; if `enabled` is anything more than that conjunction, the sound silently disagrees with the tap. `enabled` is not in the diff, so I could not check that.

Laziest version: `syncExpansionSound(enabled:)` taking the value the caller already holds, and preview playing what sync just stored, resolving only when that is nil.

Not verified: no runtime here, so the tap headroom claim stays unchecked, same as you say. `syncWithPreferences`'s callers, and whether the master toggle routes through it, are outside the diff.

**Impact:** none.
**Merge:** ship it, before #766.

tenbux:
Took the last one too. `syncExpansionSound(featureEnabled:)` now accepts the value `syncWithPreferences` has already computed and only works it out itself when called directly from the picker, so the two gates cannot answer differently. The preview plays the sound the sync just retained and resolves for itself only when nothing is retained, which is when the feature is off.

You were right that `enabled` was the thing to check: it is `AppFeature.textSnippets.isAvailable && bool(textSnippetsEnabled)` and nothing more, so the two agreed, but only by inspection. That is now structural rather than coincidental.

tenbux:
Rebased onto `cb26014`. #917 landed in between and rewrote `postExpansion`, which changed something this PR depends on, so the resolution is a behavior change rather than a mechanical one.

Expansion can now fail. `expand` and `postExpansion` return `Bool`, and a transient paste that cannot take the pasteboard fails open to normal typing, with the caller passing the trigger character through. The sound was firing unconditionally after the blocking window, so on that path it would have confirmed an expansion the user did not get, which is worse than silence because it stops them checking. It now fires only when the result is `true`:

```swift
let expanded = Thread.isMainThread ? post() : DispatchQueue.main.sync(execute: post)
if expanded, let sound = inputLock.withLock({ expansionSound }) { … }
return expanded
```

Still async and outside the window the tap callback waits on, so the latency fix is unchanged. Single-line snippets keep the typed path and behave as before; the difference is only visible on a multi-line snippet, which is the path #917 introduced.

On the earlier leftover about three places deciding whether the sound is on: that is closed in this branch. `syncWithPreferences` passes its computed value through `syncExpansionSound(featureEnabled:)`, and `previewExpansionSound` plays the retained instance rather than resolving its own, falling back only when nothing is retained.

Verified: `./build.sh` exit 0 with no warnings, `--selftest` OK. `./build.sh --test` reports one failure, `the dispatch pool starvation this check needs was actually reached`, which fails identically on a clean `main` here and passes on `main` from before #989 added it.

PathGao:
Gating on `expanded` is the right idea, and the async dispatch keeps playback clear of the tap's window. It doesn't catch the failure you're gating on, though.

`TransientPaste.paste` returns `true` as soon as it dispatches to `GeneralPasteboardAccess` — that `return true` is the last line of the function, outside the async block. Every real failure (unsnapshottable flavor, `changeCount` race, `setString` false) calls `didFail` from a later main-queue block, well after `postExpansion` already returned `true`. The only synchronous `false` is the `isPerforming` re-entry. So on the paste path — `requiresPaste` is `contains(where: \.isNewline)`, exactly what #917 added — a clipboard holding promised data fails open to typing and still gets the sound. That is the case the body says can't happen.

Same root cause on success: the sound leads the paste, which only goes out after the modifier-release loop plus 60ms.

`TransientPaste` already has the hook and `PastePlainService` already uses it. Thread a closure through `postExpansion`, call it from `didPostShortcut` on the paste branch and before `return true` on the typed branch.

Not verified: no runtime here, so which apps actually leave unreadable promised flavors on the pasteboard is a code reading, not an observation.

**Impact:** multi-line expansions that silently fail still play the confirmation sound.
**Merge:** needs the sound hung off the paste's own completion first.

tenbux:
Confirmed, and the fix is where you pointed.

`postExpansion` now takes a `didExpand` closure. On the paste branch it goes in as `didPostShortcut`, so it runs from `postPasteShortcut`'s completion after the ⌘V key-up is posted, past the modifier-release loop and the 60ms, so the sound follows the paste instead of leading it. The unsnapshottable flavor, the `changeCount` race, `setString` false and the event-creation failure all reach `didFail` instead and play nothing. On the typed branch it fires once before `return true`. `paste`'s return value gates nothing now.

It still dispatches async: the typed branch calls it while the tap callback may be blocked on `main.sync`, so starting playback there would sit inside that window.

Corrected the wording too. The comment and the description both claimed the return value separated a paste that landed from one that failed open. It never did. The new comment says what `didPostShortcut` actually proves, which is that the ⌘V went out. Whether the target app inserted anything is still not observable from here.

Same as your note: this is a code reading, not a runtime observation of which apps leave unreadable promised flavors.

tenbux:
Both points from the summary.

**The three places.** `previewExpansionSound` and `syncExpansionSound` each read `snippetSoundName` and called `resolvedSoundName` themselves; they now share one `preferredExpansionSound()`. The picker was the third, deciding a stored name was unavailable with its own `!alertSoundNames.contains(soundName)`, which is the same question the resolver answers by a different route. It now asks the resolver:

```swift
if TextSnippetSupport.resolvedSoundName(stored: soundName) != soundName {
    Text(text.soundUnavailable).tag(soundName)
}
```

That is also the property the row means rather than a proxy for it: what would play is not what is stored. The two existing assertions pin both directions of it, so I did not add a third that would only restate them.

On the "whether it is on" half I want to be sure I fixed the right thing, because I have not changed how that is derived. Tell me which places you counted and I will collapse them the same way.

**The conflict count.** Measured at the current base, `ac98c896`, both for the head your row was reading and for the rebased one:

```
$ git merge-tree --write-tree origin/main cb0a0d92   # exit 0, no conflict entries
$ git merge-tree --write-tree origin/main da819bde   # exit 0, no conflict entries
```

Zero either way, so there are no files to name. The branch is now rebased onto `ac98c896`, still one commit.

On that tree: `swift build` clean, `./build.sh --test` reports `TESTS OK (10503 checks)`, `./build.sh` no warnings, `--selftest` `SELFTEST OK`.

Sources/Vorssaint/UI/Settings/TextSnippetsSettings.swift
```diff
@@ -34,6 +37,35 @@ struct TextSnippetsSettings: View {
...
+                        .onChange(of: soundEnabled) { _, _ in
+                            TextSnippetService.shared.syncExpansionSound()
+                        }
+                    Text(text.soundCaption)
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                    if soundEnabled {
+                        Picker(text.soundPickerLabel, selection: $soundName) {
+                            ForEach(TextSnippetSupport.alertSoundNames, id: \.self) { name in
+                                Text(name).tag(name)
```
vorssaint: these are the file names, and the mac has shown these sounds under other names since big sur, so what is tink here is boop in the sound settings and ping is sonar. can the list use the names the mac shows, translated where the app has that language

tenbux: Fixed. Added AlertSoundStrings.displayName(for:language:), sourced from Apple's own AlertSounds.loctable (read directly off /System/Library/ExtensionKit/Extensions/Sound.appex/Contents/Resources/AlertSounds.loctable on this Mac, not guessed), translated for each of the thirteen languages this app supports. Also confirmed against that same table that Japanese, Korean and Chinese keep the English names there too, so the picker matches for those as well. The stored preference is still the file name.

Sources/Vorssaint/UI/Settings/TextSnippetsSettings.swift
```diff
@@ -34,6 +37,35 @@ struct TextSnippetsSettings: View {
...
+                        .onChange(of: soundEnabled) { _, _ in
+                            TextSnippetService.shared.syncExpansionSound()
+                        }
+                    Text(text.soundCaption)
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                    if soundEnabled {
+                        Picker(text.soundPickerLabel, selection: $soundName) {
+                            ForEach(TextSnippetSupport.alertSoundNames, id: \.self) { name in
+                                Text(AlertSoundStrings.displayName(for: name, language: l10n.language)).tag(name)
```
vorssaint, line 52: the list still follows the file names, so what people see now runs mezzo, breeze, pebble, jump and ends with boop. can it sort by the names it shows, in the order the app's language would put them

tenbux: Fixed. Added AlertSoundStrings.sortedNames(_:language:), which orders the file names by what each one's displayName resolves to, using a Locale built from the app's own language rather than the system's, and wired it into the picker's ForEach. The stored preference and tag are unchanged; only the display order moved.

Issue #489, opened 2026-08-05
请求新增截图 OCR 文字提取、翻译功能


SecOnline521:
### Problem or Motivation

1、针对无法复制的文字无法便捷的提取
2、针对不同的语言无法快速的翻译

### Proposed Solution

在截图功能中新增 OCR 文字识别功能和翻译功能

### Feature Area

Other

### Mockups / References

<img width="1588" height="418" alt="Image" src="https://github.com/user-attachments/assets/d284a549-c2f8-468b-a731-650af32c4449" />

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

Half of this is already implemented and the other half is blocked on an open PR.

Offline OCR exists today: `Sources/Vorssaint/Services/QuickTools/ScreenTextService.swift`, sharing the capture framing with the screenshot tool. So the text-extraction request is done.

What remains is translation, and that is PR #773 ("Add Live Translation"), open and currently mergeable. One flag for whoever reviews it: #773 introduces a Google Cloud Translation API key stored in the Keychain plus usage and cost tracking, which puts a paid third-party cloud dependency into a local macOS utility — that is a product decision ahead of a code review. If only the on-device Apple translation path lands, this issue is still satisfied.

PathGao:
Folding **#691** in here. This issue, #691 and item 2 of #738 are all built-in translation, and PR #773 ("Add Live Translation") is the implementation for all three.

What #691 adds, and what neither this issue nor #738 says anywhere: **which backend does the translating.**

- @2548085610 asks for a provider list — 「支持mac自带翻译/谷歌翻译/有道翻译，然后也可以使用三方api」: macOS's built-in translation, Google Translate, Youdao, plus a user-supplied third-party API endpoint.
- PR #773 ships two of those: Apple Translate (on-device) and Google Cloud Translation. **Youdao and a user-configurable endpoint are covered by nothing.** So if the pluggable-backend question is not settled by the time #773 lands, it should be tracked from here.
- This is also relevant to the flag already raised on #773 — a Google Cloud Translation API key in the Keychain with usage and cost tracking. A provider list where the paid cloud one is one option among several is a different product decision from a single hardwired cloud dependency.

Also carried over from #691:

- **Screenshot translation** (「截图翻译功能」) — same shape as the OCR flow in this issue's body.
- Reference project: MoePeek, https://github.com/cosZone/MoePeek
[...]

PR #773, opened 2026-08-22
feat(live-translation): add Live Translation feature

naveenkrdy:
## Summary

- Select a region of the screen and see the text inside it translated live, redrawn either directly over the original text or as a separate floating overlay, updating automatically as the underlying content changes.
- **Two recognition engines**: a native Vision document-layout engine on macOS 26+ that understands real paragraph/list/table structure, with an automatic fallback to a geometry-based line-grouping heuristic on macOS 15-25. Users on macOS 26+ get a Settings toggle to switch back to the compatibility engine if they prefer it; the toggle is hidden entirely on older macOS since there's nothing to switch.
- **Two translation providers**: Apple Translate (fully on-device, private) or Google Cloud Translation (API key stored in Keychain), with a usage/cost tracker and an optional hard cap on character usage for the Google path.
- Settings now explains that Apple Translate runs on-device and links directly to System Settings → General → Language & Region → Translation Languages so users know exactly where to download a language pair.
- Overlay controls: pause/resume, switch between in-place and window rendering, copy original/translated text, copy or save the composited image, drag to reposition, and live-resize the tracked region (content clears immediately when a resize starts rather than showing stale text).
- Performance: unchanged frames are skipped via a cheap image fingerprint so OCR only runs when the region actually changes; translated text is cached per session so repeated paragraphs are never re-sent to a translator; the recognition model is proactively warmed before a session starts and re-warmed during idle ticks so the first real capture isn't paying a cold-load cost.
- Ships as an opt-in feature (`AppFeature.liveTranslation`), gated on Screen Recording permission.

## Test plan

- [x] `./build/Vorssaint --selftest` passes
- [x] `./build.sh --test` passes
- [x] New user-facing strings added to all 13 supported language files
- [x] Manually verified: region selection and live translation start/stop, drag-to-move, live resize (including the clear-on-resize-start behavior and the resize-specific session-continuity fix), in-place and window display modes, Apple on-device provider end-to-end, Google provider end-to-end including usage tracking and the usage cap, engine toggle on macOS 26+, and the on-device language download link

Snapzy_2026-08-23_03-41-54_046

[See it in Action](https://pub-d1f174c205bf431e9aa4eb83be698fc8.r2.dev/snapzy/1787433475-98e419a7-snapzy_recording_2026-08-23_02-46-01.mp4)

Ref #489


PathGao:
You're right, and I need to correct something I wrote in the PR summary.

I described this as having a "hard character cap." It doesn't — `liveTranslationGoogleUsageCapEnabled` defaults to `false`, so the cap is opt-in like the rest of it. My error, not a gap in the PR.

The rest of your answer checks out too: `liveTranslationProvider` defaults to `"apple"`, and once the Strings files are excluded the Google-specific surface is roughly 65 added lines (`LiveTranslationKeyStore` 53, plus ~9 in `TranslationProvider` and ~23 in the service). Coupling is as shallow as you said, which also means it stays cheap to remove later if the call goes the other way.

The advanced-toggle idea sounds reasonable to me. Past that, whether a paid third-party cloud provider belongs in this app is a product question rather than a code one, so I'll leave it there.

PathGao:
Ref #489

PathGao:
Separate from the product question about the Google provider, here is what I found reading the Keychain and network code. These apply whether or not that path survives.

**1. The API key travels in the URL, and there is a log line that will print it.** `TranslationProvider.swift:63-64` puts the key in a query item; `:73-76` builds the request. Google Cloud Translation v2 accepts `X-Goog-Api-Key`, so this is `request.setValue(key, forHTTPHeaderField: "X-Goog-Api-Key")` at `:75` and no query item at all.

It matters concretely here because of the error path. `:81-84` wraps every `URLSession` failure as `TranslationProviderError.network(error)`, and `LiveTranslationService.swift:671` logs `String(describing: error)` at `privacy: .public`. A bridged `URLError` renders its `userInfo`, which carries `NSErrorFailingURLStringKey` — the full request URL, `?key=…` included. Any offline blip or DNS failure then writes the user's credential into the unified log, where `log show` hands it to anything running on that Mac. `:674` has the same shape for the untyped catch. Move the key to a header and mark both log interpolations `privacy: .private`; either change alone leaves one of the two holes open. I could not run this to capture the exact string macOS emits, but the fix costs nothing and removes the question.

**2. `LiveTranslationKeyStore` discards every OSStatus.** `SecItemUpdate` (`:35`), `SecItemAdd` (`:40`) and `SecItemDelete` (`:45`) all have their results dropped. A failed write is then indistinguishable from a successful one: the settings field keeps showing what the user typed, and the feature reports `missingAPIKey` later with no way to connect the two. `save` should return the status (or a `Bool`) and `ScreenCaptureSettings` should show the failure. Related: `kSecAttrAccessible` is set only on the add branch (`:39`), so an item that already exists keeps whatever accessibility it was first created with — set it in the update attributes too.

**3. The key is written to the Keychain on every keystroke.** `ScreenCaptureSettings.swift:349-350` calls `LiveTranslationKeyStore.save(newValue)` from `SecureField`'s `.onChange`, so a 39-character key is stored as 39 successive prefixes of itself, each one a `read()` plus a `SecItemUpdate`. Deleting back to empty runs `delete()`. Save on `.onSubmit` or when focus leaves the field instead.

**4. The usage cap does not hold the guarantee its comment claims.** `LiveTranslationService.swift:618-620` says the gate "guarantees the persisted total never actually exceeds a configured cap". `recordGoogleUsage` (`:653`) runs only on the success branch, and `catch is DeadlineExceeded` (`:663`) is reached *after* the request went out — the client stopped waiting, Google still received, processed and billed it. Every tick that hits `translateDeadline` is billed and never counted, so on a slow link the cap drifts arbitrarily far below real usage and the number shown in Settings is wrong in the direction that costs money. Record the characters when the request is dispatched and reverse the charge only for the failures that provably were not billed (`missingAPIKey` never left the process; 400/403 is a rejected request).

**5. The character count is not the unit Google bills.** `:626` uses `String.count`, which counts grapheme clusters. Google bills per code point, so any text with combining marks, emoji or regional-indicator pairs under-counts — a flag is one grapheme and two code points. `uniqueTexts.reduce(0) { $0 + $1.unicodeScalars.count }` matches what appears on the invoice, and the cap in `googleUsageWouldExceedCap` (`:732-738`) then compares like with like.

The rest of the wiring checks out: `.liveTranslation` reaches both explicit tool lists (`GlobalShortcut.swift:771-774`, `FeatureVisibilitySupport.swift:255`) rather than only the `CaseIterable` ones the compiler would have caught, ⌃⌥⌘6 does not collide with an existing default, and nothing sends the captured image anywhere — only recognized text leaves the machine, and only on the Google path.

naveenkrdy:
Thanks for the thorough pass - all five are real, fixed.

**1. Key in the URL + logging.** Moved it to an `X-Goog-Api-Key` header instead of `?key=`. Also marked both error-path log lines (`translate failed with TranslationProviderError:` / `translate failed with error:`) `.private` instead of `.public`, since an `NSError`'s description can embed the failing request even without the key living in the URL specifically - didn't want to rely on the header move alone to close this.

**2. `LiveTranslationKeyStore` discarding `OSStatus`.** `save` now returns whether the write actually succeeded. `ScreenCaptureSettings` shows an inline message when it doesn't, instead of assuming the Keychain took it.

**3. Keychain write on every keystroke.** Moved to `.onSubmit` and focus-loss (`@FocusState`), plus a last-chance save `.onDisappear` so a key typed right before closing Settings doesn't get lost by never blurring.

**4. Usage-cap guarantee.** You're right that "only counted on success" was the actual bug, not a hedge - two overlapping ticks could both pass the pre-request gate before either recorded, so the real total could overshoot the cap despite the check. Reservation now happens before the request goes out; I only reverse it for `missingAPIKey`, since that's the one failure that provably never reaches Google (thrown before any connection opens). A timeout or a real HTTP response back from Google's servers - I can't prove either wasn't billed, so those charges stand rather than risk undercounting.

**5. Character counting.** Switched both the cap-gating count and the matching diagnostic log to `unicodeScalars.count`. Agreed `String.count`'s grapheme-cluster counting doesn't match what shows up on the invoice for anything with combining marks or most emoji.

Also rebased onto current `main` and pushed - the branch should be mergeable again.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8234 checks). Didn't have a way to force an actual Keychain write failure or a real network drop to exercise those specific paths live - reasoned through both from the code rather than triggering them, flagging that same as before.

PathGao:
Four are closed. Item 2 is half-closed: `save` returns the status now, but `kSecAttrAccessible` is still only on the `SecItemAdd` branch — the update path passes `[kSecValueData: data]` alone, so an item that already exists keeps whatever accessibility it was first created with.

Nobody hits it today: this is the only writer and it always creates with `kSecAttrAccessibleAfterFirstUnlock`. It bites the day that constant changes — every existing install silently keeps the old value, with nothing to show for it. Writing it once and using it on both paths kills the branch and the second copy:

```swift
let attributes: [String: Any] = [
    kSecValueData as String: data,
    kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlock,
]
status = read() != nil
    ? SecItemUpdate(baseQuery as CFDictionary, attributes as CFDictionary)
    : SecItemAdd(baseQuery.merging(attributes) { $1 } as CFDictionary, nil)
```

GitHub still reports the branch as conflicting from where I'm looking. That may just be a stale snapshot taken before your push — worth a check.

**Impact:** none today; the accessibility attribute only diverges from the intended value if that constant is ever changed.
**Merge:** hold for the call on whether the Google provider ships at all.

ref #489

naveenkrdy:
Good catch, and you're right about item 2. Fixed - both `SecItemUpdate` and `SecItemAdd` now go through one `attributes` dict with `kSecAttrAccessible` in it, so there's no branch left that can diverge.

The `CONFLICTING` status was real, not stale - the branch had drifted behind `main` again since my last push. Rebased and pushed; should be clean now.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8730 checks).

PathGao:
Read the new `save` — both paths go through one `attributes` dict, so `kSecAttrAccessible` can't diverge any more. That closes item 2, and nothing new turned up in the change.

No code objection left from me. What's still holding this is the product call on whether the Google provider ships at all, which is @vorssaint's.

**Impact:** none.
**Merge:** hold for the call on whether the Google provider ships at all.

ref #489

naveenkrdy:
@vorssaint waiting on you . thanks!

vorssaint:
closing this one. the app stays local first and never names other companies' products, so a paid cloud translator with an api key, usage counters and a dollar estimate does not fit, and the on device path is the only one that could ship
the settings section stacks a dozen controls at the top level with wording like engine, strategy and latency, and the strategy choice is not read by anything, so that page would need to be redone from scratch
translation only works between the app's own thirteen interface languages, so a screen in any other language just sits at translating with nothing shown
the branch is more than three hundred commits behind and the screen capture settings page has been restructured since, so the settings part does not rebase mechanically
a smaller version could still come in as a new pull request, on device translation only, one tool in the capture chooser, target language following the system, at most five controls in settings and wording without product names

PR #785, opened 2026-08-23
Show real video-frame thumbnails on the shelf

tenbux:
## Summary

- video files dropped onto the shelf now show a real decoded video frame instead of macOS's generic file-type icon, matching Finder/QuickLook
- generation is async via `AVAssetImageGenerator`, grabbing a frame slightly into the video rather than at time zero, so an arbitrary video's black/title-card first frame is never what shows
- the shelf item is shelved immediately with today's fallback icon; the real frame patches in once ready through the shelf's existing "build now, patch in later" mechanism (`replaceItem`, already used to heal moved/renamed files)
- `ShelfService.Item` compares by id alone, so a same-id icon swap leaves a SwiftUI view taking `[Item]` comparing equal and its `updateNSView` is never called. The tiles now also take a revision counter that changes on every `items` mutation, which lets the content check `rebuildTiles` already does decide the redraw
- every shelf content thumbnail now decodes at one size sized for the tile's own well, so a dropped image, a dropped video frame, a pasted image and a pasted GIF are all equally sharp. They previously decoded at the 20pt small-icon default and were upscaled into a 112x84px area
- app-launch restore decodes image thumbnails off the main thread rather than inline, so rebuilding a saved shelf no longer holds launch for the sum of every image on it. An interactive drop keeps the inline decode, where it is one file's work and shows the thumbnail immediately

## Behavior before

A video file on the shelf showed the OS's generic file-type icon, resized, with no indication of its actual content.

## Behavior after

A video file shows a real decoded frame from partway into the clip. A corrupt or unsupported video keeps the generic icon with no crash. A batch drop of several videos, including inside a collapsed pile, updates independently as each finishes.

## Testing

- `swift build`: passed, no errors
- `./build.sh --test`: `TESTS OK (8498 checks)`
- `./build.sh`: passed without warnings
- `./build.sh --dev --install` plus manual testing in the running app: individual videos each patching in their own frame; a batch forming a collapsed pile whose first item is a video, where the pile tile picks up that frame at the content-thumbnail inset; a pasted image and a pasted GIF; and relaunching with videos and images already shelved so they come back through the restore path.

Thumbnail sizing was checked against the tile rather than by eye. The shelf's content well is 56x42pt, so 112x84px at 2x; the old 20pt default decoded to 40x30px and was upscaled into it, the tile size decodes to 128x96px.

Decode cost was measured before deciding not to bound the fan-out: 200 concurrent 1080p exact-frame extractions finish in 2.8s off the main thread with a 65MB peak footprint, and 200 image thumbnails in 147ms with no cooperative-pool starvation.

Checked on Apple Silicon, a single Retina display and English, against a `--dev --install` build rather than a release one. Not tried on Intel, on a non-Retina or multi-display setup, or in another app language.

### DIFF
--- a/Sources/Vorssaint/Services/ImageThumbnailer.swift
+++ b/Sources/Vorssaint/Services/ImageThumbnailer.swift
@@ -8,10 +8,25 @@ enum ImageThumbnailer {
     static let defaultPointSize: CGFloat = 20
 
     static func thumbnail(for url: URL, pointSize: CGFloat = defaultPointSize) -> NSImage? {
+        thumbnail(for: url, pointSize: pointSize, scale: backingScale)
+    }
+
+    /// The same decode, off the main thread. Decoding is synchronous work
+    /// that scales with the source file, so a caller with many files to get
+    /// through (the shelf restoring a saved list) would otherwise hold the
+    /// main thread for the sum of all of them. NSScreen is main-thread-only,
+    /// so the scale is read there and the decode runs off the main actor,
+    /// the same split VideoThumbnailer uses.
+    static func thumbnail(for url: URL, pointSize: CGFloat = defaultPointSize) async -> NSImage? {
+        let scale = await MainActor.run { backingScale }
+        return thumbnail(for: url, pointSize: pointSize, scale: scale)
+    }
+
+    private static func thumbnail(for url: URL, pointSize: CGFloat, scale: CGFloat) -> NSImage? {
         let sourceOptions = [kCGImageSourceShouldCache: false] as CFDictionary
         guard let source = CGImageSourceCreateWithURL(url as CFURL, sourceOptions) else { return nil }
 
-        let maxPixelSize = pixelSize(for: pointSize)
+        let maxPixelSize = pixelSize(for: pointSize, scale: scale)
         let options = [
             kCGImageSourceCreateThumbnailFromImageAlways: true,
             kCGImageSourceCreateThumbnailWithTransform: true,
@@ -20,7 +35,6 @@ enum ImageThumbnailer {
         ] as CFDictionary
 
         guard let cgImage = CGImageSourceCreateThumbnailAtIndex(source, 0, options) else { return nil }
-        let scale = backingScale
         let size = NSSize(width: CGFloat(cgImage.width) / scale,
                           height: CGFloat(cgImage.height) / scale)
         return NSImage(cgImage: cgImage, size: size)
@@ -74,11 +88,11 @@ enum ImageThumbnailer {
         return pixels * pixels * 4
     }
 
-    private static var backingScale: CGFloat {
+    static var backingScale: CGFloat {
         max(1, NSScreen.main?.backingScaleFactor ?? 2)
     }
 
-    private static func pixelSize(for pointSize: CGFloat) -> Int {
-        max(16, Int((pointSize * backingScale).rounded(.up)))
+    private static func pixelSize(for pointSize: CGFloat, scale: CGFloat = backingScale) -> Int {
+        max(16, Int((pointSize * scale).rounded(.up)))
     }
 }
--- a/Sources/Vorssaint/Services/Shelf/ShelfService.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfService.swift
@@ -31,17 +31,24 @@ final class ShelfService: ObservableObject {
         let title: String
         let icon: NSImage
         let isImage: Bool
+        /// True once `icon` is a real decoded frame of the item's own
+        /// content (an image thumbnail or a patched-in video frame), as
+        /// opposed to a generic fallback icon. Purely a tile-sizing signal
+        /// for ShelfTilesView: unlike `isImage`, it does not gate any
+        /// decoding or kind-classification behavior.
+        let hasContentThumbnail: Bool
         /// Finds a file payload again after a move or rename; nil for
         /// non-file payloads and for files whose bookmark could not be made.
         let bookmark: Data?
 
         init(id: UUID = UUID(), payload: Payload, title: String, icon: NSImage,
-             isImage: Bool, bookmark: Data? = nil) {
+             isImage: Bool, hasContentThumbnail: Bool = false, bookmark: Data? = nil) {
             self.id = id
             self.payload = payload
             self.title = title
             self.icon = icon
             self.isImage = isImage
+            self.hasContentThumbnail = hasContentThumbnail
             self.bookmark = bookmark
         }
 
@@ -53,7 +60,9 @@ final class ShelfService: ObservableObject {
         /// healed URL/title after a moved or renamed file's bookmark
         /// resolves - an id-only comparison would call that unchanged.
         func hasSameContent(as other: Item) -> Bool {
-            guard id == other.id, title == other.title, isImage == other.isImage else { return false }
+            guard id == other.id, title == other.title, isImage == other.isImage,
+                  hasContentThumbnail == other.hasContentThumbnail,
+                  icon === other.icon else { return false }
             switch (payload, other.payload) {
             case let (.file(lhs), .file(rhs)): return lhs == rhs
             case let (.text(lhs), .text(rhs)): return lhs == rhs
@@ -96,6 +105,7 @@ final class ShelfService: ObservableObject {
 
     @Published private(set) var items: [Item] = [] {
         didSet {
+            contentRevision &+= 1
             scheduleRefit()
             schedulePersist()
             // Emptying the shelf (removing or dragging out the last item) always
@@ -115,6 +125,14 @@ final class ShelfService: ObservableObject {
     /// into view. Not persisted: it means "just now", and a relaunch has no
     /// just now.
     @Published private var lastAddedID: UUID?
+    /// Counts every mutation of `items`, including one that swaps an item for
+    /// a same-id replacement. `Item.==` is id-only by design, so a SwiftUI
+    /// view taking `[Item]` alone compares equal after such a swap and its
+    /// `updateNSView` is never called: an async-patched video thumbnail would
+    /// sit in `items` and never reach the screen. Feeding this to the tiles
+    /// gives them a value that always differs, so the redraw is decided by
+    /// `rebuildTiles`' own content check rather than by id equality.
+    @Published private(set) var contentRevision = 0
     /// Counts adds so the tiles can tell a genuine arrival from a redraw.
     /// The resolved reveal target is not enough on its own: it changes when a
     /// pile is expanded, and it repeats when two files land in the same pile.
@@ -1196,6 +1214,7 @@ final class ShelfService: ObservableObject {
         return Item(id: item.id, payload: .file(resolved),
                     title: resolved.lastPathComponent,
                     icon: item.icon, isImage: item.isImage,
+                    hasContentThumbnail: item.hasContentThumbnail,
                     bookmark: (try? resolved.bookmarkData()) ?? bookmark)
     }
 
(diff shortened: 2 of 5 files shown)


PathGao:
The thumbnail work is fine — off-main, correct SE-0338 reasoning, undecodable video degrades to the workspace icon. The part to reconsider is `forceRebuildTilesIfLive()`, because this PR already contains the fix that makes it unnecessary.

**`hasSameContent` was the actual bug.** `ShelfService.swift:60` omitted `icon`, so when the decoded frame was patched in, the coordinator cache at `ShelfTilesView.swift:160` declared the item unchanged and `updateNSView` skipped the rebuild. This PR adds `icon === other.icon` — and with that in, `items` is `@Published`, `ShelfView` observes it via `@EnvironmentObject`, and `updateNSView` → `rebuildTiles` already runs on every mutation. The published-property path is sufficient. The Service-layer walk through `panel.contentView` hunting for the first `NSScrollView`, the context-menu counter, the deferral flag and `firstScrollView` can all come out.

It is also strictly worse than the path it bypasses: `forceRebuildTilesIfLive` calls `rebuildTiles` with `coordinator: nil`, so it skips the `unchanged` cache *and* never updates `lastRebuiltItems`. Every forced rebuild is immediately followed by a second full teardown/rebuild from `updateNSView`.

And the guard does not cover what its comment claims. It says it covers "the active drag source or an open context menu's target", but `isInternalDragActive` (`ShelfService.swift:1310`) is only true for drags that *originated* in the shelf. A Finder drag hovering or dropping on a tile runs `draggingEntered` / `performDragOperation` / `concludeDragOperation` (`ShelfTilesView.swift:627-666`) with the flag false, and `concludeDragOperation` relies on the tile still being in the hierarchy (`superview?.enclosingScrollView`). A thumbnail landing in that window tears it out.

Two things independent of the above:

- `ShelfService.swift:1462` — the new batch branch in `replaceItem` propagates the child's video frame into the pile's `icon` but never sets `hasContentThumbnail`, so `ShelfTilesView.swift:320` draws that real frame at the generic-icon inset (13/8) instead of the thumbnail inset (4/4). The sizing bug this PR fixes for leaf tiles, left standing one level up.
- `ShelfService.swift:1863` — `maxRestoreVideoThumbnails` bounds video only. The same restore pass runs `ImageThumbnailer.thumbnail(for:pointSize: 64)` synchronously on the main thread for every restored image (`fileItem`, `:1375`), and this PR raised that from 20pt to 64pt — roughly ten times the pixels, unbounded, on the launch path. The budget was built and the image path never joined it.

Also worth a look: `addFileBatch` maps every dropped URL through `fileItem`, each spawning its own `Task` and `AVAssetImageGenerator` with `requestedTimeToleranceBefore/After = .zero`, which forces an exact-frame decode from the preceding keyframe. The comment asserts "interactive drops are a handful of files"; nothing enforces it.

PathGao:
The fixes after #837 are the 08-23 20:46 force-push, and that is the version my last comment was written against — head is still `742f2db5`, nothing has changed since.

That push added the deferral: `openTileContextMenuCount`, `pendingForcedTileRebuild`, `ShelfTileView` as `NSMenuDelegate`, retried from `finishInternalDrag` and `menuDidClose`. It makes `forceRebuildTilesIfLive` safer, but it does not answer why the function is there. `hasSameContent` now compares `icon === other.icon` (`ShelfService.swift:65`), `items` is `@Published`, and `updateNSView` → `rebuildTiles` already runs on every mutation. With that fix in, the forced path is not what makes the thumbnail appear — it is a second mechanism now being hardened to sit alongside the one that works.

All four concrete items are still standing at `742f2db5`:

- `ShelfService.swift:1293` — `rebuildTiles` is called with `coordinator: nil`, so it skips the `unchanged` cache and never updates `lastRebuiltItems`. Every forced rebuild is followed by a second full teardown from `updateNSView`.
- The new guard covers internal drags only. `isInternalDragActive` is true for drags that *originated* in the shelf; a Finder drag runs `draggingEntered` / `performDragOperation` / `concludeDragOperation` with it false, and `concludeDragOperation` relies on the tile still being in the hierarchy (`superview?.enclosingScrollView`). A thumbnail landing in that window still tears it out.
- `ShelfService.swift:1247-1250` — the batch branch propagates the child's frame into the pile's `icon` but builds the `Item` without `hasContentThumbnail`, so it defaults to false and that real frame draws at the generic-icon inset (13/8) instead of 4/4.
- `ShelfService.swift:1492` — `contentThumbnailPointSize` is 64 and the restore pass calls `ImageThumbnailer.thumbnail` synchronously on the main thread for every restored image. `ImageThumbnailer.defaultPointSize` on main is 20, so that is roughly ten times the pixels, unbounded, on the launch path, outside the `maxRestoreVideoThumbnails` budget the video path got.

The branch still merges cleanly against main (`3ed44acf`), so these are the only things in the way.

PathGao:
Three more at `742f2db5`, none of them repeats of the four still standing.

**1. `openTileContextMenuCount` can stick above zero for the rest of the session, and the guard it feeds protects nothing.** `ShelfTileView` makes itself the delegate of both the context menu and the "Open With" submenu (`ShelfTilesView.swift:470`, `:499`), and `NSMenu.delegate` is a weak reference. Nothing defers the *ordinary* rebuild path: `replaceItem` assigns `items = current` before it calls `forceRebuildTilesIfLive()` (`ShelfService.swift:1256-1258`), that publish drives `updateNSView` → `rebuildTiles`, and `rebuildTiles` does `document.subviews.forEach { $0.removeFromSuperview() }` (`ShelfTilesView.swift:180`) regardless of `openTileContextMenuCount`.

So while a tile's context menu is open, a thumbnail patch already tears that tile out through the published path. Two consequences:

- The deferral does not do what its comment says. By the time `forceRebuildTilesIfLive` (`:1285`) checks the counter and defers, the menu's target view has already been removed by the path the counter does not gate.
- Once that view deallocates, the weak delegate goes nil and `menuDidClose` (`ShelfTilesView.swift:511`) is never delivered. `endTileContextMenuTracking` (`ShelfService.swift:1355`) never runs, the count stays at 1, and every later forced rebuild defers forever — `runPendingForcedTileRebuildIfSafe` has the same guard, so nothing ever clears it either. `max(0, ...)` absorbs an extra close but there is no recovery from a missing one.

I could not run this, so the deallocation timing is read from the code rather than observed. If the forced path stays, the counter needs to be reset somewhere unconditional (`rebuildTiles` itself, or a token the tile owns and releases in `deinit`). Removing `forceRebuildTilesIfLive` removes the counter, the deferral flag and this failure mode together, which is the same conclusion as before, now with a concrete way for the current shape to break.

**2. `batchItem` has the `hasContentThumbnail` omission too.** `ShelfService.swift:1603-1606` takes `children.first?.icon` and constructs the pile with `isImage: false` and the default `hasContentThumbnail: false`. `ShelfTilesView.swift:320` therefore insets 13/8 for every pile whose first child is an image and already carries a real thumbnail — at creation, not only after a video patch. It is the same defect I flagged in `replaceItem`'s batch branch, one site earlier; fixing only the patch path leaves every freshly built pile wrong. `batchItem` can derive it the same way it derives the icon: `children.first?.hasContentThumbnail ?? false`.

**3. The unbounded decode has a second entry point.** `items(from pasteboard:)` at `ShelfService.swift:1613` also does `fileURLs.map { fileItem(for: $0) }`, so a paste of a Finder multi-selection spawns the same per-URL `Task` + zero-tolerance `AVAssetImageGenerator` fan-out as `addFileBatch`. Whatever bound the drop path gets should sit inside `fileItem` (or in a small shared gate it calls) rather than at one caller, otherwise the next caller added misses it as well.

tenbux:
All nine are addressed in the force-push. `forceRebuildTilesIfLive` is gone, and with it `firstScrollView`, the `forceRedraw` parameter, `openTileContextMenuCount`, `pendingForcedTileRebuild` and the `NSMenuDelegate` tracking. The service no longer touches the view hierarchy, so items 1, 2, 3 and 7 go with it.

The reasoning isn't right though, and it's worth correcting since someone could hit it again. The published path is not sufficient on its own: removing the forced rebuild and relying on it would have broken the feature quietly. `ShelfService.Item.==` compares id only, so after `replaceItem` swaps in a same-id item with a new icon, the `items` array compares equal, the `ShelfTilesView` value compares equal, and `updateNSView` is never called. `hasSameContent` and its `icon === other.icon` check decide what happens once `rebuildTiles` runs, but nothing gets that far.

I logged `updateNSView` entry, the `unchanged` verdict and whether the coordinator was nil, then dropped one video. With the forced rebuild suppressed, the patch landed and nothing followed for the next 8 seconds. So there was no second rebuild either, because there was no `updateNSView` call to begin with.

What replaces it is a revision counter bumped in `items.didSet` and passed to the tiles, so the value differs after an in-place swap and the redraw decision falls to the content check `rebuildTiles` already does. After the change: patch at `40.672`, `updateNSView` at `40.673`, `unchanged=false`, real coordinator. One rebuild, through the normal path.

Fixed as described:

- `hasContentThumbnail` now propagates in `replaceItem`'s batch branch and in `batchItem`. You were right that fixing only the patch path leaves every freshly built pile wrong.
- Restore no longer decodes image thumbnails inline on the main thread; they go through the same async patch as video.

Two of them I measured and got a different answer:

- The point size isn't where the cost is. Through the real `ImageThumbnailer` path, 200 4000x3000 JPEGs decode in 7.96 ms each at 20pt and 7.88 ms each at 64pt. Decoding the full source dominates and the output size is nearly free, so raising it didn't make restore more expensive. The per-image cost on that path is real, and that's what the async change above is for.
- On the fan-out, 200 concurrent 1080p exact-frame extractions finish in 2.8s off the main thread with a 65MB peak, and 200 image thumbnails in 147ms without starving the cooperative pool. So instead of bounding the drop path I removed `maxRestoreVideoThumbnails` altogether. Restore order is stable, so that budget permanently denied a thumbnail to every video past the twentieth, which is the thing this PR is meant to fix. With it gone, videos and images work the same way.

On the stuck counter, your read of the deallocation looks right to me. The trigger isn't quite the one described, since the ordinary path doesn't fire for a same-id swap, but any other mutation that changes ids or count would rebuild, drop the tile and strand the count. Gone either way.

Not on your list, but I hit the same problem in `imageItem` and `gifItem`. Both set `hasContentThumbnail: true` while decoding at the 20pt default, so a pasted image or GIF got the big inset at small-icon resolution. The well is 56x42pt, or 112x84px at 2x, and 20pt only gives you 40x30px. They decode at the same size as everything else now.

PathGao:
`Item.==` being id-only is the piece I got wrong, and the log settles it — `hasSameContent` only decides anything *after* `updateNSView` runs, and it never ran. Revision counter is the right shape.

Four small things in the current diff. None of them block.

**1. `pixelSize(for:scale:)` keeps `scale: CGFloat = backingScale` (`ImageThumbnailer.swift:95`).** The whole point of the split is that `backingScale` touches `NSScreen` and has to be read on main. The only caller now passes `scale` explicitly, so the default is dead — and it is exactly the thing that puts an off-main `NSScreen` read back in the moment someone omits the argument. Delete the default.

**2. `item.isImage || item.hasContentThumbnail` (`ShelfTilesView.swift:323`) defeats the flag for the case the flag exists for.** `fileItem`'s comment says the flag "tracks the icon actually in hand, so an unreadable image (or one still decoding) is not sized as though it had a thumbnail." For an image it never does — `isImage` is true regardless, so a restored image still decoding and an image that never decodes both draw the generic fallback at the 4/4 inset. `hasContentThumbnail` is now set true on every inline decode, every paste and every patch, so `isImage` is a second predicate for the same fact. Either drop the `||` and the flag delivers what the comment claims (cost: a visible inset change when a restored image's patch lands), or the comment is overstating and should say so.

**3. `videoExtensions` (`ShelfService.swift:1422`) is a second hardcoded type table, wrong in both directions.** AVFoundation does not decode `mkv` or `webm`, so those entries can only ever spawn a `Task` and an asset load that fails; meanwhile `m2ts`, `mts`, `ogv` and any extension-less video get no thumbnail at all. `UTType(filenameExtension:)?.conforms(to: .movie)` is one line and is the same classification `NSWorkspace` already used to pick the fallback icon. I have not checked which of those container types macOS declares as conforming to `.movie`, so the doomed-`Task` half may survive the change; the coverage half does not. (`imageExtensions` predates this PR — folding both into `UTType` is a bigger swing than this needs.)

**4. `patchContentThumbnail` (`ShelfService.swift:1450`) discards the frame silently.** One 50ms sleep, one recheck, then the decoded image is dropped and nothing records it. `acceptMixedBatch` is not in the diff, so I cannot tell how long a slow sibling provider can actually hold the append — this is read off your own retry comment, not measured. The race exists only because the decode is fired from `fileItem`, a factory that runs before the item is in `items`. Triggering the patch at the insertion point removes the timer instead of tuning it.

One more to look at rather than fix: `replaceItem`'s batch branch (`:1238`) now duplicates `batchItem`'s icon+flag derivation, and the comment says as much. Two copies that already disagree on title, before a third shows up.

**Impact:** none — every one of these degrades to today's generic icon, nothing regresses against main.
**Merge:** ship it.

PR #787, opened 2026-08-23
Slow App Switcher overflow-row hover to one icon at a time

### LINKED ISSUE #786: [Bug]: App Switcher overflow row races through icons on edge hover
### Description

When the App Switcher icon row overflows, hovering the last visible icon on either side immediately center-scrolls the row. Because the newly revealed icon lands under the still-parked pointer, the 0.15s easeOut jump repeats and the list races through several apps.

**Expected:** hovering a middle icon only selects it. Parking on the last visible icon of a side, when more icons wait beyond it, should reveal the next icon one at a time, slowly enough to aim at.

**Actual:** the overflow row jumps the hovered icon to the center as soon as the pointer reaches the last visible tile, so a parked mouse walks the entire list.

### Steps to Reproduce

1. Enable the App Switcher icon-row layout (icon row or simple mode).
2. Open the switcher with more apps or windows than fit in the visible row.
3. Move the pointer onto the last visible icon on the right (or the first visible icon once the row has already scrolled).
4. Leave the pointer parked there.

The row center-scrolls immediately and keeps jumping until the end of the list.

### Screenshots / Recordings

_No response_

### Vorssaint Version

3.3.3-beta.2 (local Developer build)

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Local Developer build

BenjaminD2023:
## Summary

- stop the overflow icon row from center-scrolling on every hover
- select a middle icon without moving the row
- park on the last visible icon of a side, and only then, to reveal the next icon one at a time
- keep preview tiles, title chips, and the card grid on immediate hover-select

Fixes #786

## Behavior before

Hovering the last visible overflow icon jumped that tile to the center in 0.15s. The newly revealed icon landed under the still-parked pointer, so the row raced through several apps.

## Behavior after

The overflow row is a clipped `HStack` that slides by `iconRowFirstVisibleIndex`. Middle icons only select. The last visible icon on a side waits 0.42s, then steps one icon (0.28s easeOut) and highlights the newly revealed edge tile. Keyboard navigation and leaving that edge icon cancel the timer. Hover stays inert until the pointer moves off the session `hoverAnchor`.

## Verification

- `./build.sh --test` — `TESTS OK (8220 checks)`
- `./build.sh --dev` — compiled without warnings
- `./build/VorssaintDeveloper --selftest` — `SELFTEST OK`
- installed into `/Applications/vorssaint-local.app` and launched


BenjaminD2023:
Follow-up: the first edge-hover wait is now 0.24s and later steps repeat at 0.20s with a 0.18s slide, so a parked pointer walks one icon at a time without feeling stuck. Still one-icon-only; middle tiles do not move the row.

PR #799, opened 2026-08-23
chore(tests): sweep the preference files a run leaves behind

PathGao:
Reopens #751, rebuilt on today's `main`.

## What was wrong

A test run wrote a preference file into `~/Library/Preferences` for every defaults suite it used, and nothing removed them. They piled up unnoticed.

## What changed

`build.sh` sweeps the two test namespaces (`vorss.tests.` and `com.vorssaint.tests.`) plus the harness's own `metrics-tests.plist` after each run, and fails the run if any survive. A test reads the swept prefixes back out of `build.sh` and checks every defaults suite the tests open is named inside one of them, so the sweep and the suites cannot drift apart.

## Why the branch was rebuilt

The original branch was cut at `814b94a`. Since then `main` gained two suites named `VorssaintTests.BetaActivation` and `VorssaintTests.StableActivation`, which fall outside both swept namespaces — so the guard this PR adds went red the moment it met current `main`, and the two files were being left behind for real (there was one sitting in `~/Library/Preferences` on my machine).

They are named `com.vorssaint.tests.betaActivation` / `.stableActivation` now, inside the swept namespace. That is the only difference from the closed branch besides the rebase.

Tests: `TESTS OK (8225 checks)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,9 @@ Vorssaint 3.3.3-beta.1 introduces the opt-in beta release channel and in-app fee
   already in front. Thanks to @PathGao.
 - The recording editor now trims reliably from the beginning of a video when
   dragging the left handle. Thanks to @lmilojevicc.
+
+- A test run no longer leaves a preference file behind in `~/Library/Preferences`
+  for every defaults suite it uses.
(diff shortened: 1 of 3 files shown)


vorssaint:

Issue #775, opened 2026-08-22
Scratchpad window can be resized to a thin line


AB-boi:
### Description

Currently there is no lower limit on how small the scratch pad can be resized to, which means it's window can be completely compressed into a thin line as can be seen in [this recording](https://github.com/user-attachments/assets/25d75634-4c9d-4dff-8451-ecac2c856169) in #764 

### Steps to Reproduce

Invoke scratch pad, resize window.

### Vorssaint Version

3.3.2

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

PathGao:
Tracking PR: #803 (open, currently labelled `incomplete`).

It keeps the scratchpad from collapsing (the hosting view's first layout zeroes `contentMinSize`). Note the Settings window at `AppDelegate.swift:1178-1183` has the same shape and was not checked in that PR.

AB-boi:
This is now completely fixed.

PR #803, opened 2026-08-23
fix(capture): keep the editor window from resizing below its minimum

iltonandrew:
Refs #775.

**Scope narrowed since the last review — please re-read, this is no longer the three-window PR you approved the shape of.** Two of the three windows were fixed on `main` this morning, by the maintainer, in a different shape:

- `0135ce8 fix(scratchpad)` adds `panel.minSize = 280x220` plus a `windowWillResize` clamp, and keeps `contentMinSize`.
- `4c91d32 fix(settings)` adds `host.view` width/height Auto Layout constraints, `window.minSize`, a `windowWillResize` clamp and a validity check on the persisted size, and keeps `contentMinSize`.

Both land the minimum this PR was landing, so this PR's scratchpad and Settings commits are dropped rather than re-litigated. What is left is the one window nobody else has touched: `ScreenshotEditorController.swift` still carries a bare `window.contentMinSize = minimumSize` at `:1009` and nothing else, and no commit on `main` has touched that file since `890dd0b`.

## What was wrong

`NSHostingController` rewrites the window's size limits from what it tracks on its first layout pass, so a `contentMinSize` assigned after `NSWindow(contentViewController:)` is silently dropped. The screenshot editor could then be edge-dragged down to a **761x142** strip.

## The change

Declare the computed minimum on the hosted view with `.frame(minWidth:minHeight:)` and track exactly that with `host.sizingOptions = [.minSize]`. The hosting controller then maintains `contentMinSize` itself. Only `.minSize` is tracked — the other options let SwiftUI content drive the window's size — and the dead `contentMinSize` assignment is removed. The declared minimum still comes from `ScreenshotSupport.editorMinimumContentSize(visibleSize:)` — but under `[.minSize]` that is a floor rather than the whole story; see the note under Environment.

## Consequence worth recording

The minimum is now enforced against Accessibility resizes too, and Vorssaint resizes its own windows through AX: `WindowLayoutService.swift:277-279` targets the app's own key window when it is `.resizable` and not an `NSPanel`, which includes the screenshot editor. So a half or a quarter on a narrow display leaves the editor overhanging the target rect instead of shrinking under its minimum. `WindowLayoutSupport.swift:428-442` accepts an edge-anchored window with over 0.45 overlap and no failure feedback fires, so this reads as fine rather than broken — and stopping at the minimum is what the minimum is for. Not drag-tested; reasoning is from those two files. (Credit to @PathGao, who found this and asked for it to be written down here.)

## Environment

MacBook Pro (Mac17,9, Apple M5 Pro), macOS 26.6.2 (25G83), built-in Liquid Retina XDR, 3024x1964 native / 1512x982 points.

This matters more than usual: `ScreenshotSupport.swift:790-794` computes the editor minimum from the visible frame, so the 980x712 below is the top of the clamp on this display and a narrower screen lands somewhere else.

And a second thing decides that number. Under `[.minSize]` the hosting controller reports the **larger** of the declared frame and the content's own intrinsic minimum. The tool rail is `.fixedSize(horizontal: true, vertical: false)` (`ScreenshotEditorView.swift:51`) so it never compresses, which puts the intrinsic width at 761pt. On this display 980 > 761, so the declared value governs and the drag stops at exactly 980x712. Below a roughly 761pt-wide visible frame the relationship inverts and the effective minimum would exceed the display — the case `ScreenshotSupport.swift:792` and the 700x500 expectation at `MetricsTests.swift:13921-13924` exist to protect. No Mac ships a display that narrow. (Measured by @PathGao in review.)

## Verification

- Interactive drag test, run 2026-08-24 against a dev build on this Mac, **not re-run against today's `main`**: unmodified, a scripted edge drag collapsed the editor to **761x142**; with this change the identical drag, plus a second harder one, stopped at exactly **980x712**. The editor hunk itself is unchanged since that run, and no commit has touched `ScreenshotEditorController.swift` since `890dd0b`, so the result should still hold — but I am dating it rather than implying a fresh run.
- `./build.sh` — 0 warnings. `./build.sh --test` — 8797 checks, **1 failure**: `the dispatch pool starvation this check needs was actually reached`. That is the known machine-dependent check #1014 exists to skip: red on this 15-core M5 Pro, green on CI's smaller runners. It came up identically on two other unrelated branches gated in the same sitting, so it is environmental and not from this change — I am reporting it rather than writing `TESTS OK`. `./build/Vorssaint --selftest` — `SELFTEST OK`.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
@@ -980,8 +980,16 @@ final class ScreenshotEditorController: NSObject, NSWindowDelegate {
     }
 
     func show() {
+        let screen = NSScreen.pointerVisibleFrame
+        let minimumSize = ScreenshotSupport.editorMinimumContentSize(visibleSize: screen.size)
+        // The hosting view rewrites the window's size limits on its first
+        // layout pass, so a contentMinSize set on the window is silently
+        // lost. Declare the minimum on the hosted view and track just that:
+        // the hosting controller then maintains contentMinSize itself.
         let content = ScreenshotEditorView(model: model, controller: self)
+            .frame(minWidth: minimumSize.width, minHeight: minimumSize.height)
         let host = NSHostingController(rootView: content)
+        host.sizingOptions = [.minSize]
         let window = NSWindow(contentViewController: host)
         // One continuous surface: the canvas fills the window and the
         // controls float over it, so the editor reads as a single object.
@@ -1000,13 +1008,10 @@ final class ScreenshotEditorController: NSObject, NSWindowDelegate {
         // Chrome must equal the view's designed margins exactly (rail 64 +
         // sides, action band above, style band below), so a fresh window
         // opens with zero leftover stage around the capture.
-        let screen = NSScreen.pointerVisibleFrame
         let contentSize = ScreenshotSupport.editorContentSize(
             imagePointSize: model.pointSize,
             visibleSize: screen.size)
-        let minimumSize = ScreenshotSupport.editorMinimumContentSize(visibleSize: screen.size)
         window.setContentSize(contentSize)
-        window.contentMinSize = minimumSize
         window.center()
 
         self.window = window


PathGao:
The fix is right where it lands. One more window has the same shape.

Clearing `contentMinSize` on the hosting view is not unique to the Scratchpad window — `App/AppDelegate.swift:1178-1182` builds a window the same way, and I have not been able to confirm from reading whether its hosting view leaves a `contentMinSize` behind that would let it be dragged down to a line too.

Could you check that one and either fix it here or say why it does not apply? If it is the same bug, fixing both in one PR is a smaller diff than two rounds. That is the only thing the `incomplete` label is about; the change itself is correct.

iltonandrew:
Sure thing, will investigate it, and fix where is needed, if necessary.

Will return to you soon! 🤠

iltonandrew:
Checked it — same bug, now fixed in this PR (12741c9).

Settings is `.resizable` and sets `contentMinSize` on the window, but the hosting view rewrites the window's size limits from whatever its `sizingOptions` track on the first layout pass — and the empty set added in #683 tracks nothing, so the minimum is gone before the window is ever on screen. Dragging the bottom edge took a 900x700 Settings window to 87x33, title bar only. v3.3.2 wasn't affected because it used the default options, which propagated `SettingsView`'s own `.frame(minWidth: 772, minHeight: 528)` — which is why reading alone was inconclusive and I drag-tested both builds.

Fixed with the same shape as the scratchpad: `sizingOptions = [.minSize]`, the minimum declared on the hosted view from `SettingsWindowSupport`, and no `contentMinSize` left on the window to be silently cleared. Re-dragged: it now stops at 772x560 (772x528 content + title bar) on every settings page. Gate re-run: `./build.sh` zero warnings, selftest OK.

While in there I swept the other `NSHostingView` windows: `ClipboardHistoryService` panel isn't `.resizable` (safe), `RecorderEditorController` declares the same minimum on the hosted view (safe). `ScreenshotEditorController.swift:1007` looks like it may have the same inert-`contentMinSize` shape, but the editor only appears after a capture and I couldn't runtime-verify it here — left it out of this PR as a possible separate follow-up.

PathGao:
12741c9 closes the point from the last round: `App/AppDelegate.swift:1170-1186` no longer leaves a `contentMinSize` on the window for the hosting view to drop. I re-checked the two windows you called safe and both hold by reading: `Services/Clipboard/ClipboardHistoryService.swift:1030` has no `.resizable` in its style mask, and `UI/Recorder/RecorderEditorView.swift:51` declares `.frame(minWidth: 940, minHeight: 560)` on the hosted view. Stale sizes also recover on their own — `UI/Settings/SettingsSearchSupport.swift:78` discards a saved pair below the minimum, so anyone who already dragged Settings to 87x33 gets the default back.

Two things left.

1. `Services/QuickTools/ScreenshotEditorController.swift:1007` — please settle it in this PR rather than as a follow-up. You deferred it because you could not runtime-verify, but the reading is more decisive than it looks: it is the same shape as the recorder editor (default `sizingOptions`, `window.contentMinSize` assigned after `NSWindow(contentViewController:)`), and the recorder editor is only safe because `UI/Recorder/RecorderEditorView.swift:51` also declares the minimum on the view. `ScreenshotEditorView` has no such declaration — the only `minWidth` in the whole file is `UI/Screenshot/ScreenshotEditorView.swift:528`, on an inner control. So `editorMinimumContentSize` is stated exactly once, on the window, in the position that gets rewritten. That asymmetry between two editors written the same way is what makes it worth one drag, not what makes it safe. The window is `.resizable` (`ScreenshotEditorController.swift:987`), so the test is the one you already ran twice: open an editor after a capture and drag the bottom-left corner in. If it goes under the computed minimum, the fix is the shape you used here — move the `let minimumSize = ...` at line 1005 above the host at line 982, build the controller as `NSHostingController(rootView: content.frame(minWidth: minimumSize.width, minHeight: minimumSize.height))`, set `host.sizingOptions = [.minSize]`, and drop line 1007. If it stops where it should, say so in the PR and I will stop asking.

2. `App/AppDelegate.swift:1183-1184` is redundant. `UI/Settings/SettingsView.swift:42` already ends the body with `.frame(minWidth: 772, maxWidth: .infinity, minHeight: 528, maxHeight: .infinity)`, so `[.minSize]` picks up 772x528 with or without the wrapper you added, and 772/528 is now stated in two files. Your note above describes the minimum as declared on the hosted view from `SettingsWindowSupport`, which is only half true while `SettingsView.swift:42` hardcodes the same pair. Drop the `.frame(...)` at 1183-1184 and change `SettingsView.swift:42` to `minWidth: CGFloat(SettingsWindowSupport.minContentWidth)` / `minHeight: CGFloat(SettingsWindowSupport.minContentHeight)` — one declaration, and it is on the view where `[.minSize]` reads it. The scratchpad wrapper is different and should stay: `ScratchpadView` declares no root minimum, so `ScratchpadService.swift:390` is the only place 280x220 exists.

Neither of these is about the Settings fix itself, which I read as correct. `positionSettingsWindow` calling `layoutSubtreeIfNeeded()` at `AppDelegate.swift:1241` also forces the layout pass before the window is shown, so the minimum is installed rather than waiting for first display.

iltonandrew:
Both settled.

**ScreenshotEditorController — you were right to insist; it did not stop where it should.** Drag-tested on a live build of current main (editor opened through the clipboard-image path, which goes through the same `openEditor` → `show()` as a capture): the AX resize to 900×700 was accepted below the 980-wide computed minimum, and the edge drags then collapsed it to **761×142** — the 761 is an incidental intrinsic minimum from an inner control, the height is a strip. Same inert-contentMinSize shape, now fixed in 8525910 with the same fix as the other two windows: `minimumSize` computed before the host, declared on the hosted view via `.frame(minWidth:minHeight:)`, `sizingOptions = [.minSize]`, line 1007 dropped. Re-dragged from an enlarged 1246×850: it stops at exactly 980×712 on this display (980×680 computed content minimum + 32 pt title strip), both axes, and refuses AX resizes below the minimum too.

**772/528 stated once.** The `.frame(...)` wrapper in `openSettingsWindow` is gone and `SettingsView.swift` now declares its root frame from `SettingsWindowSupport.minContentWidth/minContentHeight` — one declaration, on the view where `[.minSize]` reads it (folded into the Settings commit, now 4fffba6). Re-dragged: Settings still clamps at 772×560.

Branch is rebased onto current main, and the gate re-ran on the result: `./build.sh` zero warnings, `--selftest` SELFTEST OK.

PathGao:
Both settled, and the sweep is complete. Reading 8525910 and 4fffba6 against current main: the three inert assignments are gone (`ScratchpadService.swift:382`, `AppDelegate.swift:1183-1184`, `ScreenshotEditorController.swift:1007`), each minimum is declared on the hosted view with `sizingOptions = [.minSize]`, and each number now exists once — `SettingsView.swift:42` reads `SettingsWindowSupport.minContentWidth/minContentHeight` from `SettingsSearchSupport.swift:69-70`, `ScratchpadView` still declares no root minimum, and `ScreenshotEditorView` declares none at its root either; its `minWidth`s are inner controls (`:533`, `:1127`). Moving `minimumSize` above the host is safe against the shrink that assigning `contentViewController` causes, since `ScreenshotSupport.swift:810-811` already clamps `editorContentSize` to at least the minimum.

There is no fourth window. The only `.resizable` style masks in the tree are `AppDelegate.swift:1182`, `ScratchpadService.swift:368`, `ScreenshotEditorController.swift:987`, `RecorderEditorController.swift:1132` and `ScreenshotPinController.swift:95`. The pin panel hosts a plain `NSView` (`ScreenshotPinController.swift:186`) and sets `minSize` at :108, so it has no hosting view to rewrite it, and the recorder is the one that holds by reading. That clears `incomplete`. The three drag results — 380×220, 772×560, 980×712 — I am taking on your word; nothing in the diff argues with them.

**Two things before this lands.** The title is what squash merge writes, and it still names only the scratchpad while the PR fixes three windows, and it does not follow the `type(scope): lowercase imperative phrase` form. Rename it to cover all three; note `screenshot` is not a scope this repo has used, `capture` is where screenshot code has landed. Second, the branch is not on current main — its merge base is a175fcee, and #905 (a8186833) merged at 00:46, minutes before your comment. No file overlap, so no conflict, but rebase once more so the gate you ran is the gate on what merges.

Also add the Mac, the macOS build and the display to the description, per the pull request rules. It matters more than usual here: `ScreenshotSupport.swift:790-794` computes the editor minimum from the visible frame, so 980×712 is the top of the clamp on a large display and a narrower one lands somewhere else. Optional while you are in there — `RecorderEditorController.swift:1147` is the last copy of the shape you removed three times. It is dead rather than wrong, since the same 940×560 comes back from `RecorderEditorView.swift:51`, so leaving it is fine.

PathGao:
Re-read the three hunks against current main. Nothing new to flag, and the sweep holds: `contentMinSize` exists in exactly four places on main — `ScratchpadService.swift:382`, `AppDelegate.swift:1329`, `ScreenshotEditorController.swift:1009`, `RecorderEditorController.swift:1147` — this PR removes the first three and the fourth is the dead one. Five `.resizable` style masks, the same five, and `ScreenshotPinController.swift:95` still hosts a plain `NSView`. No `as? NSHostingController<ScratchpadView>`-shaped cast anywhere in the tree, so wrapping the two root views in `.frame(...)` breaks no runtime lookup.

On the rebase: the pre-image of all three hunks is byte-identical on current main (`AppDelegate.swift:1315-1336`, `ScratchpadService.swift:379-390`, `ScreenshotEditorController.swift:982-1010`) — only the line offsets moved, and `ScratchpadService` did not move at all. It is mechanical, no conflicts to resolve. I compared source text, not git history, so I cannot tell you whether main moved again in the last two days.

One consequence nobody has written down. The minimums are now enforced against AX resizes, and Vorssaint resizes its own windows through AX: `WindowLayoutService.swift:277-279` deliberately targets the app's own key window when it is `.resizable` and not an `NSPanel`, which is the Settings window and the screenshot editor — not the scratchpad. So a half or a quarter on a narrow display now leaves those two overhanging the target rect instead of shrinking into it. It reads as fine rather than broken, since `WindowLayoutSupport.swift:428-442` accepts an edge-anchored window with over 0.45 overlap and no failure feedback fires. Not asking for a change — it belongs in the description next to the display note, since the editor minimum is display-dependent and this is where a reader would look for it. I have not drag-tested this; the reasoning is from the two files above.

**Impact:** none beyond the fix. Window-layout actions on Settings and the screenshot editor now stop at the minimum instead of going under it, which is what the minimum is for.
**Merge:** needs the title rename and the rebase from the last round first.

ref #775

iltonandrew:
Both of your merge conditions are done, but the reason they were easy is the thing to read first: **two thirds of this PR is now redundant and I have dropped it.** The PR is materially different from the one you reviewed, so please re-read rather than re-approving the shape.

**What changed under us.** You hedged that you could not tell whether `main` had moved. It had, this morning, in exactly this area:

- `0135ce8 fix(scratchpad): improve window dragging, resizing and tab closing` — adds `panel.minSize = 280x220` and a `windowWillResize` clamp to `ScratchpadService`.
- `4c91d32 fix(settings): enforce design minimum window size and bounds` — adds `host.view` width/height constraints, `window.minSize`, a `windowWillResize` clamp and an `isValidContentSize` check on the persisted size, in `AppDelegate`.

Both land the minimum this PR was landing, for those two windows, in a different shape — and notably both **keep** `contentMinSize` and enforce through `windowWillResize` instead of through the hosting controller. That is a design call, it is the maintainer's, and it is already merged. Re-landing my version on top would be a second fix for a bug that no longer exists.

So the scratchpad and Settings commits are dropped (`git rebase --onto origin/main 4fffba6`). The three commits were cleanly separated by file, so nothing else moved. What is left is the one window neither commit touched: `ScreenshotEditorController.swift:1009` still carries a bare `window.contentMinSize = minimumSize`, and nothing has touched that file since `890dd0b`.

**Title.** Your two constraints could not both be met before — "cover all three" and CONTRIBUTING's "reuse a scope that already shows up in `git log`", when the repo has no umbrella window scope and the three windows sit in three different ones. Narrowing dissolves it: one window, and `capture` is the scope you pointed at. Now `fix(capture): keep the editor window from resizing below its minimum`, on the commit as well as the PR.

**Rebase.** Done, and now trivial — one commit, one file, +8/-3, no conflicts.

**Description.** Rewritten for the narrowed scope, and it now carries the two things you asked for: the Mac, macOS build and display (it matters here, since `ScreenshotSupport.swift:790-794` derives the minimum from the visible frame, so 980x712 is this display's clamp and a narrower screen lands elsewhere), and your `WindowLayoutService` consequence, credited to you.

**Verification, stated honestly.** 0 warnings, `SELFTEST OK`, and `--test` reports 8797 checks with one failure — the known `dispatch pool starvation` precondition that #1014 exists to skip. It is red on this 15-core M5 Pro and green on CI, and it came up identically on two unrelated branches I gated in the same sitting. The interactive drag numbers (761x142 collapsed, 980x712 clamped) are from the 2026-08-24 run and I have **not** re-run them against today's `main`; the editor hunk has not changed since and neither has that file, so they should still hold, but they are dated rather than re-claimed.

ref #775

PathGao:
Both conditions closed, and the narrowing holds. Re-read as a new PR.

Checked against current main rather than on your word: `ScratchpadService.swift:383-390` and `AppDelegate.swift:1316-1337` are the two commits you named — both `sizingOptions = []` plus `window.minSize` plus a `windowWillResize` clamp — and neither touches `ScreenshotEditorController`. The one live bare `contentMinSize` left in the tree is `:1009`, plus the dead `RecorderEditorController.swift:1147`. Both hunks' pre-image is byte-identical to main at the lines the diff names, so the rebase is real and the dated drag numbers are dated against a file that has not moved. Title is in form. The fix's premise still holds: `ScreenshotEditorView`'s root is a plain `ZStack` with no `.frame` minimum (`:37-62`), nothing in the tree casts the editor's host to a concrete `NSHostingController<ScreenshotEditorView>`, and `ScreenshotSupport.swift:811-812` clamps the initial content size to at least the minimum, so `setContentSize` at `:1008` cannot fight the new floor.

One correction to the description, not a change request. "The minimum itself is unchanged; it still comes from `ScreenshotSupport.editorMinimumContentSize(visibleSize:)`" is half true under `[.minSize]`: the hosting controller reports the larger of the declared frame and the content's own intrinsic minimum, and this PR measured that intrinsic at 761 wide — the tool rail is `.fixedSize(horizontal: true, vertical: false)` at `ScreenshotEditorView.swift:51`, so it never compresses. On your display 980 > 761 and the declared value governs, which is why you got exactly 980x712. Below a ~761pt-wide visible frame it inverts and the effective minimum exceeds the display, which is the case `ScreenshotSupport.swift:792` and the 700x500 expectation at `MetricsTests.swift:13921-13924` exist to protect. No Mac ships a display that narrow. Leave the code alone; put the sentence in the description next to the display note.

The shape divergence is the maintainer's call, not something to block on — but yours is the only one of the three windows that does not leave an inert `contentMinSize` behind. `ScratchpadService.swift:383` is assigned before `contentViewController` at `:390`, `AppDelegate.swift:1336` after `NSWindow(contentViewController:)` at `:1325`; both are rewritten by exactly the bug this PR describes, and the real enforcement in both is `window.minSize`. That is about main, not this diff, and not yours to fix here.

Not verified: the interactive drag results, and whether main moved in the last hours — I compared source text, not git history. `Swift 6.0.3 compatibility` shows no state on my side; build and selftest are green.

**Impact:** none beyond the fix. The editor stops at its computed minimum instead of collapsing to a strip, and AX-driven window layout on it stops there too.
**Merge:** ship it, once the Swift 6.0.3 check reports.

ref #775

Issue #784, opened 2026-08-22
Repeated Cmd+Shift+Tab does not step backwards in the App Switcher


justin-chiam:
### Description

While the switcher is open, holding Cmd+Shift and pressing Tab repeatedly only navigates backwards on the first press. Every following Tab press is swallowed and does nothing. You have to release Shift and press it again for each step backwards.

**Expected:** hold Cmd+Shift and tap Tab repeatedly to keep walking backwards through the list, the same way holding Cmd and tapping Tab walks forwards.

**Actual:** only the first Cmd+Shift+Tab moves. Subsequent Tab presses with Shift do not do anything.

### Steps to Reproduce

1. Hold Cmd and press Tab to open the switcher (or open it with Cmd+Shift+Tab).
2. Keep Cmd and Shift held down.
3. Press Tab several times.
4. Selection moves back by one on the first press only; it does not move again until Shift is released and re-pressed.

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

PathGao:
Tracking PR: #804 (open, `blocked` — first-time contributor CI approval, not a conflict).

Root cause: `shiftBackNavigationHeld` carries two meanings at once, the level of "is Shift held" at `:851` and a one-shot key-swallow flag at `:610`/`:625`. Not merged yet.

MYK2K:
@vorssaint 

This is actually another issue altogether. Instead of Cmd+Shift+Tab. Only, Cmd+Shift while the switcher is open moves the switcher in reverse. The latest update fixed the Cmd+Shift+Tab behaviour as it should be.

However, if the switcher is open and you press only Cmd+Shift. It still goes in reverse, this should not happen.

Steps to reproduce -
1. Open Switcher - Cmd+Tab or Cmd+Shift+Tab.
2. Keep holding Cmd.
3. Press Shift. The selector can be seen moving in reverse.

Issue #187, opened 2026-07-08
[Bug]: Switcher for windows does not work


martzell:
### Description

The App Switcher shortcut to switch windows does nothing. I've reassigned the shortcut to ⌃⌥⌘L, but it still doesn't work.

After switching on "Show ⌘Tab with large Icons" I can switch windows in this now different App Switcher, so probably a misunderstanding of what the Windows Switcher does. I expected it to switch windows in an app. If it is intended to do nothing, when the Vorssaint App Switcher is not in grouped mode, then it is only a usability issue. You could consider to rename and regroup the options in Vorssaint Settings › Switcher to make it more clear.

But a bug with German keyboard layout remains: I cannot assign the key above the tab key – labeled:
 °
^
as shortcut. Probably because American and Asian keyboards are ANSI and have one key less than German ISO keyboards. I guess on German ISO keyboards the key next to the left shift key sends the same key code as the key above the tab key on ANSI keyboards and the key above the tab key on German ISO keyboards is the real additional key with its own key code.

Another thing I'd like to draw attention to: Apples cmd tab switcher stops at the end if you hold the tab key. Another press of tab jumps to the beginning. This is nice and helps to navigate quickly.

### Steps to Reproduce

Press the shortcut to switch windows.

### Vorssaint Version

3.1.8

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

vorssaint:
Looking into it, great feedback! Thanks!

vorssaint:
Fixed in 3.1.9. The windows shortcut now works in the plain grid too, and the extra key on ISO keyboards can now be recorded as a shortcut. Holding the switcher key also stops at the end of the list now instead of wrapping around.

PR #804, opened 2026-08-23
Let repeated Shift+Tab keep stepping backwards in the switcher

iltonandrew:
Ref #784.

## What was wrong

Before: with the switcher open and Cmd+Shift held, only the first Tab press stepped backwards; every following Tab press was swallowed until Shift was released and pressed again.

The keyDown handler bailed out of both app-navigation branches whenever Shift was held **and** `shiftBackNavigationHeld` was set. That flag is the latch `handleShiftBackNavigation` uses to detect the Shift *press transition* on `flagsChanged` (pressing Shift alone steps back once — the 3.1.6 'move backward with Shift' feature), and it stays true for as long as Shift is physically held — so after the first backward step, every Cmd+Shift+Tab keyDown hit the early exit.

## Why the guard existed — and why plain removal is wrong

The guard shipped with the Shift-back feature itself (3.1.6), and it does one legitimate job: when the user presses **Shift+Tab as one chord** mid-session, the Shift press lands first (`flagsChanged` steps back once) and Tab's keyDown arrives milliseconds later — without any guard, one physical chord would step back **twice**. The bug is that the latch swallowed every later Tab too, not that swallowing is wrong.

## The change

After: Shift alone steps back once; Shift+Tab pressed as a chord steps back once; holding Cmd+Shift and tapping Tab keeps walking backwards, mirroring how Cmd+Tab walks forwards.

The blanket latch check is replaced by a one-shot chord window: when the Shift-press transition steps back, a deadline of `SwitcherSupport.shiftBackChordWindow` (0.35 s) is armed, and only a Tab arriving inside that window is consumed — once. Later Tabs during the same Shift hold navigate normally. A session opened with Cmd+Shift+Tab pre-latches `shiftBackNavigationHeld` without arming the window, so the first Tab after a reversed open already navigates. The deadline is cleared on consumption and at session teardown. Autorepeat and wrap behavior (#187) are unchanged, and `shouldNavigateBackwardOnShiftPress` (with its existing tests) still gates the transition step.

## Verification

`./build.sh` finishes with no warnings and `./build/Vorssaint --selftest` prints `SELFTEST OK`. Scenario walk against the event order macOS delivers (flagsChanged precedes the keyDown carrying the flag):

- open with Cmd+Tab, press Shift → one step back (transition), tap Tab ×3 within the hold → first Tab inside the chord window is consumed, the next two step — the chord counts once and repeats work
- open with Cmd+Shift+Tab, keep both held, tap Tab ×3 → three steps back (the issue's repro)
- Shift+Tab chord pressed mid-session → exactly one step (no double-step regression)
- release Shift, press again → steps again (transition re-arms)

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
+++ b/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
@@ -126,6 +126,12 @@ final class AppSwitcher: ObservableObject {
     private var sessionShortcut: GlobalShortcut?
     private var sessionScope: SwitcherSessionScope = .allApps
     private var shiftBackNavigationHeld = false
+    /// Pressing Shift mid-session already steps back once, so the Tab landing
+    /// in that same physical chord must not step again — but later Tabs during
+    /// the same Shift hold must keep walking the list (issue #784). The chord
+    /// is recognized by time: anything after this deadline is a deliberate
+    /// separate press.
+    private var shiftBackChordDeadline: TimeInterval = 0
 
     /// Windows already asked to close, still listed until they are really
     /// gone. Releasing the shortcut skips them, so they are never raised on
@@ -614,7 +620,7 @@ final class AppSwitcher: ObservableObject {
         let shortcut = sessionShortcut ?? appsShortcut
         switch keyCode {
         case _ where keyCode == shortcut.keyCode && shortcut.matches(event: event, allowingExtraShift: true):
-            if shortcut.shiftIsNavigationModifier, flags.contains(.maskShift), shiftBackNavigationHeld {
+            if shortcut.shiftIsNavigationModifier, flags.contains(.maskShift), consumesShiftBackChordTab() {
                 break
             }
             let delta = shortcut.shiftIsNavigationModifier && flags.contains(.maskShift) ? -1 : 1
@@ -629,7 +635,7 @@ final class AppSwitcher: ObservableObject {
                                     tolerating: shortcut.modifiers):
             // A window-scoped session keeps its list when the Apps shortcut is
             // pressed with overlapping modifiers instead of expanding to all apps.
-            if appsShortcut.shiftIsNavigationModifier, flags.contains(.maskShift), shiftBackNavigationHeld {
+            if appsShortcut.shiftIsNavigationModifier, flags.contains(.maskShift), consumesShiftBackChordTab() {
                 break
             }
             let delta = appsShortcut.shiftIsNavigationModifier && flags.contains(.maskShift) ? -1 : 1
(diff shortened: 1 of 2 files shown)


PathGao:
#787 merged on 08-24, so the ordering constraint noted in #837 is spent — but the merge left this branch `CONFLICTING`.

It is one hunk. Both PRs appended a helper near the same point in `Services/Switcher/SwitcherSupport.swift`; #787's `clampedIconRowFirstVisibleIndex` / `iconRowFirstVisibleIndex` now sit where this branch adds its own. `git merge-tree origin/main <head>` reports exactly:

```
CONFLICT (content): Merge conflict in Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```

`AppSwitcher.swift` auto-merges clean. Resolution is "keep both".

Could you rebase onto current `main`? #784 is still open and this is the only PR fixing it.

iltonandrew:
Rebased onto current main. The one conflicting hunk in `SwitcherSupport.swift` was resolved keep-both as you suggested — #787's icon-row helpers stay where they landed, with this branch's `shiftBackChordWindow` following them. `AppSwitcher.swift` merged clean.

Gate re-run after the rebase: `./build.sh` with zero warnings, `./build/Vorssaint --selftest` → SELFTEST OK. Branch is now MERGEABLE.

PathGao:
Rebase checks out and the conflict point is resolved the way it should be. At 4ec2547, `Services/Switcher/SwitcherSupport.swift:766-796` still holds #787's `iconRowEdgeHoverDelta` and `iconRowIndexAfterEdgeHoverStep` unchanged against `main`, with `shiftBackChordWindow` appended at line 802 — keep-both, nothing dropped from either side.

The fix itself survived intact. `AppSwitcher.swift` against `main` is only the 21 lines of this change: both keyDown branches still route through `consumesShiftBackChordTab()` at `Services/Switcher/AppSwitcher.swift:623` and `:638`, the deadline is armed once in `handleShiftBackNavigation` at `:871`, and it is cleared with the rest of the session state at `:1288`, so a new session always starts unarmed and the first Tab steps.

Nothing outstanding from my side. I have not run the switcher, so the 0.35s classification window is still the one thing only a real Shift+Tab can confirm.

iltonandrew:
Ran the switcher for real to close the remaining point — the 0.35 s classification window, exercised on a live build of this branch (4ec2547, dev install) with synthetic HID events posted at `.cghidEventTap` in the exact order the tap sees from a keyboard: `flagsChanged` first, then the keyDown carrying the flag. Selection was read off the panel's highlight ring mid-session, and every session ended with Esc, so the window list stayed identical across runs (plain-grid mode; the relevant prefix was [0] current window, [1] Dictionary, [2] Chess, [3] Calculator).

Three runs:

1. **Repeats + chord consume** — open ⌘Tab, Tab ×2 (highlight walks 1→2→3), press Shift (steps back to 2, window armed), Tab at +100 ms, Tab at +400 ms. Ended on **1**: the +100 ms Tab was consumed — once — and the next one navigated. The pre-fix latch would have swallowed both later Tabs (ends on 2); plain guard removal would have let the Shift-adjacent Tab step too (ends on 0). One run separates all three behaviors.
2. **Physical chord mid-session** — resting on 3, Shift then Tab 30 ms apart. Ended on **2**: exactly one step, no double-step regression.
3. **Window expiry** — resting on 3, Shift (→ 2), then Tab at +600 ms. Ended on **1**: past the 0.35 s deadline the Tab navigates instead of being consumed.

So the window classifies as intended on a live switcher: a Tab 100 ms behind its Shift press is a chord, a Tab 600 ms behind it is navigation, and either way exactly one consume per Shift press.

PR #805, opened 2026-08-23
Edit the copied file itself, not Finder's icon preview

### LINKED ISSUE #774: [Bug]: Screenshot editor grabs finder thumbnail when editing from clipboard
### Description

If I copy an image in Finder and then invoke the screenshot editors edit from flipboard feature, the editor grabs the finder thumbnail for that image instead of the actual image:

Image

### Steps to Reproduce

Copy an image in Finder, Invoke Edit from Clipboard shortcut

### Screenshots / Recordings

_No response_

### Vorssaint Version

3.3.2

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

iltonandrew:
Fixes #774.

## What was wrong

Before: copying an image file in Finder and invoking the "Edit clipboard image" shortcut opened the file's small Finder icon in the editor instead of the actual image.

When Finder copies a file it puts both `public.file-url` and an icon/thumbnail image representation on the pasteboard. `clipboardCapture(from:)` in `ScreenshotService` built the capture straight from `NSImage(pasteboard:)`, which prefers the image types — so the editor got the icon, not the file's pixels. This is the only entry point for the clipboard-edit path (`openClipboardImage()` is its sole caller); `ShelfService` already checks file URLs first and was never affected.

## The change

After: the editor opens the copied file at full resolution.

`clipboardCapture(from:)` now asks the pasteboard for file URLs whose contents conform to `UTType.image` (AppKit's own `.urlReadingContentsConformToTypes` filter — no hand-rolled extension list) and loads the file from disk when one is present, falling back to `NSImage(pasteboard:)` otherwise. A copied .zip, .txt or folder never matches the filter, and an unreadable or corrupt file falls back to the old behavior rather than failing. The read runs off the main thread inside the existing `GeneralPasteboardAccess` async block, so no UI stall.

## Verification

`./build.sh` finishes with no warnings and `./build/Vorssaint --selftest` prints `SELFTEST OK`. Manual check: Cmd+C an image file in Finder → "Edit clipboard image" → the editor opens the file's real pixels; copying pixel data (e.g. a screenshot region) behaves as before.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
@@ -3,6 +3,7 @@
 
 import AppKit
 import ImageIO
+import UniformTypeIdentifiers
 
 /// The screenshot tool: freeze-first area, window and full screen capture
 /// with an annotation editor, pinned floating captures and direct clipboard
@@ -482,7 +483,7 @@ final class ScreenshotService: ObservableObject {
     private static func clipboardCapture(
         from pasteboard: NSPasteboard
     ) -> ScreenshotSelectionController.Capture? {
-        guard let image = NSImage(pasteboard: pasteboard),
+        guard let image = clipboardImage(from: pasteboard),
               image.size.width > 0, image.size.height > 0
         else { return nil }
         var rect = CGRect(origin: .zero, size: image.size)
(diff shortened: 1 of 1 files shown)


iltonandrew:
Follow-up verification beyond build+selftest: ran the shipped helper against real `NSPasteboard` items in a standalone harness.

- Finder-style item (file URL of a 100×50 PNG **+** 16×16 icon TIFF on one item) → returns the file's 100×50 pixels ✔
- image data only (copied screenshot region) → returns the pasteboard image, unchanged behavior ✔
- `.txt` file URL + icon TIFF → falls back to the icon, exactly pre-fix behavior ✔
- file URL to a missing file + icon TIFF → no crash, falls back ✔
- control with the pre-fix code on the Finder-style item returns the 16×16 icon — the reported bug, reproduced ✔

Issue #737, opened 2026-08-21
autopause music (spotify for example) when another audio source is playing?


Crxmiad:
### Problem or Motivation

sorry if this is already a feature and i dont know about it. 

i like having music playing in the background for everything i do, but its just a little annoying having to pause spottify manually everytime i want to watch yt, or a lecture. its especially bad when im reading a document with multiple embedded videos and i have to pause spotify each time i want to watch a video, then unpause afterwards

### Proposed Solution

it would be epic if, somehow, vorssaint could detect a second audio source playing and pause the music player (the music player could be something that's like set by the user prior, rather than having two audio sources and having to figure out which is the music player and which is the other thing)?

### Feature Area

Sound

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

Blocked on PR #813 ("Add automatic music pausing for competing audio"), which is open at +919/-23 across 38 files and currently conflicting, so it needs a rebase onto current main before review.

Two review notes recorded against it: it should state its relationship to #704 (both touch `Sources/Vorssaint/Services/Audio/`), and it adds new `Strings` fields, which increases the catch-up burden on the Vietnamese localisation work in #821.

No separate work is needed on this issue while that PR is in flight.

PR #813, opened 2026-08-23
feat(audio): add automatic music pausing for competing audio

LakshanSivakumar:
## Before

Vorssaint could not automatically pause a chosen music player when another app began producing audio. Browser videos without a Now Playing session were especially easy to miss.

## After

- Adds an opt-in beta Auto-pause music feature with a selectable player and optional automatic resume.
- Extracts Core Audio process discovery into one shared push-driven monitor used by both AutoPauseMusic and the volume mixer. On macOS 14.4 and later there is no AutoPauseMusic polling timer.
- Uses one `AudioProcessActivitySupport` runtime gate for the mixer, shared monitor, settings and AutoPauseMusic fallback. The energy profile is `.idle` on the push-driven modern path and `.periodic` only on the older polling fallback.
- Treats `IsRunningOutput` as an active output stream, not proof of audible samples. Call and DAW processes already classified as bypassed by `MixerRoutingSupport` are excluded, competing output must persist for 1 second before pausing, and inactivity must persist for 2 seconds before resuming.
- Uses the existing `AppleScriptRunner` on a dedicated serial worker queue. State inspection and pause happen in one script execution, and bundle identifiers are escaped with `AppleScriptRunner.literal`.
- Loads and registers MediaRemote only as the fallback on systems older than macOS 14.4.
- Reuses the Music blocker replacement when no explicit AutoPauseMusic player is selected, resolves that app for the picker display, and shows an in-panel “Using X from Music app blocker” caption so the inheritance and later retargeting are visible. An explicit AutoPauseMusic player remains an independent override.
- Preserves `isBypassed` while duplicate mixer rows merge and normalizes a bypassed merged row back to no route and unity volume, fixing a pre-existing mixer bug without allowing the merged row to retain a process tap.
- Adds localized settings, a README Sound-section entry, and Automation usage text for every currently supported language.

## Relationship with #704

This PR and #704 are independent audio features. PR #704 performs an explicit `coreaudiod` recovery action and owns no monitoring or playback state. This PR observes Core Audio activity and controls the selected player through Automation. A reset can temporarily interrupt the shared process snapshot, but the Core Audio listeners repopulate it when the daemon returns; neither feature depends on the other.

## Localization sequencing with #821

This PR adds fields to the compiler-enforced `Strings` contract and implements them for the current 13 language variants. If #821 lands first, this PR must add the Vietnamese fields while rebasing. If this PR lands first, #821 must provide the corresponding Vietnamese translations. The dependency is merge-order catch-up, not a runtime coupling.

## Shared Automation usage description

`NSAppleEventsUsageDescription` is shared by every Apple Event feature. The wording changes in `Info.plist` and all 12 localized `InfoPlist.strings` therefore also affect the prompt description used by Finder cut/paste, Trash, and Homebrew. The expanded text explicitly names the new player control and retains the existing Homebrew purpose in every locale.

## Testing

- rebased onto current `main` (`ecc1c1c`)
- `./build.sh` — optimized release bundle assembled and strict signature verification passed
- `./build.sh --dev` — Developer ID-signed developer bundle assembled and launched, stamped `fd564dd`
- `./build.sh --test` — `TESTS OK (8601 checks)`
- `plutil -lint Resources/Info.plist Resources/*.lproj/InfoPlist.strings`
- `git diff --check`

Ref #737


PathGao:
Registration is complete — all five feature lists plus `FeaturePresets`, `SettingsDirectory`, `Localization.swift`, all 11 locale files and `FeatureHubStrings` × 13. The conflicts against `main` right now are only `bluetoothSleep` landing in the same list slots and are not a criticism. What needs settling is the detection mechanism and where the AppleScript runs.

**1. AppleScript on the main thread, against the rule this repo already wrote down.** `AutoPauseMusicService.runAppleScript` *requires* the main thread — it logs `.fault` and bails otherwise — and runs `executeAndReturnError` synchronously there, twice per pause. `ShellSupport.swift:213` already provides `AppleScriptRunner` for this, and its doc comment says the opposite: call it off the main thread "so a slow target never blocks the UI or the event taps (the calls block their thread until the target replies)". A wedged Spotify or Music blocks the UI and every CGEvent tap in the app.

**2. A second implementation of a mechanism that is already here, in push form.** `CoreAudioOutputMonitorBridge.audioProcessObjects()` / `read()` and the `IsRunningOutput` + PID + `ResponsibleProcess.regularAppOwner` loop are near-verbatim `AppVolumeMixer.swift:930-950` and `:1457`. The mixer installs a *push* listener on `kAudioHardwarePropertyProcessObjectList` (`AppVolumeMixer.swift:186`); this polls the same property every 1.5s forever while enabled. Two places answering "which apps are producing output", with nothing keeping them in agreement, and the polling one is the new one.

**3. `IsRunningOutput` is not "audible".** The class doc says it "covers audible apps". The mixer's own comment at `AppVolumeMixer.swift:930` says it shows "every regular app that holds an audio connection, not only the ones making sound this instant", and `MixerRoutingSupport.bypassesProcessTap` names Zoom and DAWs as apps that hold streams open. Those read as "another source" and hold the music paused indefinitely. In the other direction, a notification ding pauses and resumes within one poll. There is no debounce or hysteresis anywhere.

Smaller:

- `pauseScript` / `playScript` / `playerStateScript` interpolate the bundle id straight into `tell application id "…"`. `AppleScriptRunner.literal` exists for that. Low severity given the id comes from a chosen `.app`, but the helper is there.
- `start()` always calls `mediaRemoteBridge.registerForNotifications()`, including on macOS 14.4+ where the Core Audio path is used and the fallback never runs — a private-framework side effect for nothing.
- `AutoPauseMusicSupport.otherSourceIsActive(nowPlayingBundleID:…)` is called only from `Tests/MetricsTests.swift` (5 assertions). The MediaRemote path goes through `otherAudioSourceIsActive` instead, so this function is dead in production while being tested as though it were the fallback's logic.
- Two settings now independently answer "which app is your music player": `musicBlockReplacementPath` plus the hardcoded `MusicLaunchBlocker.blockedBundleIDs`, and the new `autoPauseMusicPlayerPath`. Nothing keeps them consistent.
- `Resources/Info.plist` and the 12 `InfoPlist.strings` rewrite the shared `NSAppleEventsUsageDescription`, which changes the wording shown for the Finder, Trash and Homebrew features too. Defensible, but it is a shared string edited for one feature's benefit and should be called out in the description.
- `README.md:79` lists every sibling in the Sound section one bullet each; this feature gets none.

PathGao:
Ref #737

LakshanSivakumar:
Rebased onto current `main` and addressed the review in `5a3310d`:

- AppleScript now runs through `AppleScriptRunner` on a dedicated serial worker queue; the state check and pause are one execution, and bundle IDs use `AppleScriptRunner.literal`.
- Core Audio process discovery and `IsRunningOutput` listeners now live in one shared push monitor consumed by both the mixer and AutoPauseMusic. The modern path has no polling timer.
- `IsRunningOutput` is documented and handled as active-stream state rather than audibility: known call/DAW bypasses are excluded, with 1-second pause debounce and 2-second resume hysteresis.
- MediaRemote is lazy-loaded and registered only on the pre-14.4 fallback path.
- The MediaRemote fallback now calls the tested `otherSourceIsActive` helper in production.
- AutoPauseMusic inherits the Music blocker replacement when it has no explicit player; the override remains for Apple Music, which cannot be a valid Music-block replacement.
- Added the missing README Sound bullet.
- The PR description now records the #704 relationship, #821 localization sequencing, and the shared `NSAppleEventsUsageDescription` impact.

Validation: optimized release build and signature verification, `SELFTEST OK`, developer build, `TESTS OK (8528 checks)`, and `git diff --check`.

LakshanSivakumar:
@PathGao Will you be able to review these updates with regards to your comments? Thank you

PathGao:
The mixer side checks out: discovery and the `IsRunningOutput` listeners are in `AudioProcessActivityMonitor`, `AppVolumeMixer` consumes a published snapshot, and there is no timer on the modern path. `AutoPauseMusicService.swift`, the settings view and the tests fall past the truncation point of the diff I can read — I did not verify the debounce/hysteresis, the `AppleScriptRunner` move, or the Music-blocker inheritance. Four things from what is visible.

**1. `.autoPauseMusic` is still classified `.periodic`.** `FeaturePresets.swift:106` puts it next to `clipboardHistory` and the monitors, which do recurring work. That bucket was chosen for the 1.5s poll this rewrite deleted. The body says "On macOS 14.4 and later there is no AutoPauseMusic polling timer" — the classification says the opposite. Unless `.periodic` is meant to cover any always-on observer, it is now the wrong bucket.

**2. "Not set" no longer means not set.** `autoPauseMusicPlayerNone: "Not set"` ships in all 13 variants, but per the body the feature falls back to `musicBlockReplacementPath` when no explicit player is chosen. The picker can therefore read "Not set" while the feature is actively pausing an app, and editing the Music blocker's replacement silently retargets auto-pause. Unless the settings view resolves the inherited app for display — not in the diff I can see — the label is wrong and the coupling is invisible.

**3. `NSAppleEventsUsageDescription` dropped "Homebrew".** English went from "…open Terminal commands when you ask for Homebrew actions" to "…when you ask", and all 12 localized files match. The body says the expanded text "names the new player control alongside those existing uses", but one of those uses stopped being named — the prompt a Homebrew action triggers now describes itself as "Terminal commands when you ask".

**4. `isBypassed: existing.isBypassed || app.isBypassed` in the merge is a mixer bug fix, unannounced.** Merged rows previously dropped the flag, so a collapsed Zoom/DAW row came back non-bypassed and eligible for a process tap. Right fix; it is a mixer correctness change riding in an auto-pause PR and it is not in the description.

Smaller: the 14.4 gate is now `AppVolumeMixer.isSupported`, `AudioProcessActivityMonitor.isSupported`, and the `#available` inside `readSnapshot`, with the fallback branch as a fourth. One of them should be the source.

**Impact:** (2) mislabels an active feature and hides that the Music blocker setting retargets it; (1) mis-buckets the feature for anything that reads presets; (3) changes the Automation prompt wording for Finder/Trash/Homebrew users. (4) has no new impact — it corrects existing behavior.
**Merge:** needs the `.periodic` classification and the "Not set" label reconciled with the inheritance behavior first.

LakshanSivakumar:
Addressed the four visible review items in `59e8456`:

1. Auto-pause is now classified as `.idle` on macOS 14.4+ where Core Audio pushes the shared snapshot, and `.periodic` only on the older MediaRemote polling fallback.
2. The Music-blocker inheritance remains, but it is no longer hidden: the picker resolves the inherited app name and the panel shows a localized “Using X from Music app blocker” caption whenever no explicit AutoPauseMusic override is set. “Not set” is now reached only when both paths are empty, with coverage for all three states.
3. The explicit Homebrew purpose has been restored in the base Automation usage description and all 12 localized variants; tests ensure each localized prompt still names Homebrew.
4. The `isBypassed` duplicate-row merge fix is now called out separately in the PR description as a pre-existing mixer correctness fix.

I also consolidated the semantic macOS 14.4 decision into `AudioProcessActivitySupport`; the remaining `#available` at tap construction is the compiler availability guard for `TapGainEngine`, not a separate fallback decision.

Rebased cleanly onto `fb41750`. Validation: 8,592 checks, optimized release build, `SELFTEST OK`, strict signature verification, and a signed developer build launched from the updated commit.

LakshanSivakumar:
@PathGao Will you be able to review these updates with regards to your comments? Thank you

PathGao:
All four are resolved in what I can read: `FeaturePresets.swift` gates on `AudioProcessActivitySupport.isSupported`, `autoPauseMusicPlayerInheritedFormat` is present in all 13 variants, Homebrew is back in `Info.plist` and all 12 `InfoPlist.strings` (checked each), and the body calls out the `isBypassed` merge fix separately.

`AutoPauseMusicService.swift`, the settings view and the tests are still past the truncation point of the diff I get. The debounce and hysteresis, the inherited-player resolution in the picker, and the `AppleScriptRunner` move are unread by me — not suspect, just unverified.

One thing from the visible diff.

**The merged mixer row can be bypassed and still carry a route.** Construction keeps the two in step: `let route = isBypassed ? nil : storedRoute(…)` and `effectiveOutputDeviceUID: isBypassed ? nil : …`. The merge now ORs `isBypassed` but takes `selectedOutputDeviceUID` and `effectiveOutputDeviceUID` from `existing` unconditionally. If `existing` was not bypassed and `app` is, the result is a bypassed row holding a non-nil route, which the constructor never produces. It needs two activities under the same persistence id where only one classifies as bypassed — differing bundle id or display name for the same identity. Narrow, not impossible.

Whether it reaches a tap depends on `appNeedsEngine`, which is past truncation for me. If that already gates on `isBypassed`, ignore this. If it keys off the route, the line you just added produces the case it was meant to prevent, one merge later.

**Impact:** none confirmed. Potentially a bypassed Zoom or DAW row eligible for a process tap through the merge path — same class as the bug the `isBypassed` line fixes.
**Merge:** needs the merged-row bypass/route answer first, one line either way.

LakshanSivakumar:
Confirmed and fixed in `fd564dd`.

`appNeedsEngine` is route-based and does not independently gate on `isBypassed`, so the mixed duplicate-row state was worth correcting. The merge now preserves the same invariant as initial row construction: if either activity is bypassed, the merged row is bypassed, both selected and effective routes are nil, the unavailable-route flag is false, and volume is unity. That makes `appNeedsEngine` false and causes reconciliation to discard rather than retain a tap.

No engine, routing, AutoPauseMusic, settings, or detection behavior was otherwise changed. A focused regression check covers the mixed bypass/non-bypass merge. Rebased cleanly onto `ecc1c1c`; `./build.sh --test` passes all 8,601 checks and `git diff --check` is clean.

LakshanSivakumar:
@PathGao Will you be able to review these updates with regards to your comments? Thank you

PathGao:
`fd564dd` holds the invariant row construction holds: `isBypassed` ORs, both route fields nil, `outputDeviceUnavailable` false, volume unity, so `appNeedsEngine` stays false. Resolved.

Still past the truncation point of the diff I get: `AutoPauseMusicService.swift`, the settings view, the tests, and `AudioProcessActivityMonitor.readSnapshot`. Unread by me, not suspect.

One question about the extraction, since the body frames it as moving discovery rather than changing it. `AppVolumeMixer.processBundleIdentifier(of:)` is deleted here along with its only call site:

```swift
bundleHints[owner] = app.bundleIdentifier
    ?? (pid == owner ? Self.processBundleIdentifier(of: object) : nil)
```

That fed `MixerRoutingSupport.rowIdentity`, so it decides `persistenceID` — the key saved volumes and routes live under. Two parts have to have survived into `AudioProcessActivity.bundleIdentifier`: the HAL `kAudioProcessPropertyBundleID` read for processes where `NSRunningApplication` reports no bundle id, and the `pid == owner` guard that stops a helper's own bundle id being attributed to the app it plays for. If either is missing, the affected rows compute a different `persistenceID` than before and their saved volume and route stop resolving. Nothing errors; the settings just appear reset once.

Not a claim it regressed — `readSnapshot` is unread by me.

**Impact:** none confirmed. If the fallback did not move intact, saved per-app volumes and routes silently reset for apps whose identity came from the HAL read, on the first launch after update.
**Merge:** needs the bundle-id fallback answer first, one line either way.

PathGao:
Closing this as stalled. The last review (2026-08-25) ended on one question, the bundle-id fallback for `pid == ownerPid`, and there has been no reply or push since; the branch is now 198 commits behind main and conflicting.

For the record, the two objections carried in the PR summary (#837) do not hold against this diff: the pause runs on `automationQueue`, not the main thread, and no player is hardcoded in production code (the Spotify and Music ids appear only in test fixtures). What is left is the rebase and that one answer.

#737 stays open. If you want to pick this back up, rebase onto origin/main and reopen, or open a fresh branch; the review on this thread still applies.

PR #821, opened 2026-08-23
feat(i18n): add Vietnamese localization

ch1nhcp:


PathGao:
The Vietnamese content itself checks out. I diffed every `%`-specifier in `Strings+Vietnamese.swift` and in the 40 feature catalogs against the `enUS` block field by field: zero mismatches, and `case .vi`/`static let vi` coverage matches `ko` exactly in every file. That is the part that usually goes wrong, and it did not. Five things before it can land.

**1. The catalog is 32 fields behind `main` and carries 4 fields `main` no longer has.** `Strings+Vietnamese.swift` passes 900 arguments; `struct Strings` on `main` (`Core/Localization.swift`) declares 928. Missing: `cutPasteShowHUD`, `cutPasteShowHUDCaption`, `launchAtLoginNeedsApproval`, `mediaErrorTargetTooSmall`, `mediaMegabytesSuffix`, `mediaSizingFileSize`, `mediaSizingResolution`, `mediaTargetSize`, `mediaTargetSizeHint`, `preciseVolumeRollerCaption`, `preciseVolumeRollerEnable`, `preciseVolumeRollerTapFailed`, `switcherMinimizedPlacementEnd/Hidden/Label/Normal`, `switcherShowFullscreenWindows`, `uninstallerFailedMoreFormat`, `uninstallerFailedNeedsFDA`, and the twelve `urlCleanerRules*` fields plus `urlCleanerRemovedFormat`. Stale and to be deleted: `urlCleanerCustomTitle`, `urlCleanerCustomCaption`, `urlCleanerCustomPlaceholder`, `urlCleanerCustomSaveButton` — the URL cleaner was reworked from a free-text field to a per-site rule list while this branch was open. Separately, `Core/BluetoothSleepStrings.swift` is new on `main` and has no `.vi` case.

**2. `Tests/MetricsTests.swift` is not in the diff, and one assertion in it fails the moment `vi` joins `AppLanguage`.** `MetricsTests.swift:9388` is `expect(localizedStrings.count == AppLanguage.allCases.count, "all app languages are covered by tests")`, against the table at `:9374-9387`. Add `(.vi, .vi)` there. Three more language tables need `vi` too, and none of them has a count guard, so they will pass while silently not testing Vietnamese: `:270-282` (`featureTitles`, four feature titles per language), `:4774-4777` (`sizeTargetStrings`), and the `switch language` at `:10953-10967`, which is exhaustive and will simply fail to compile. That last one is the only one the compiler catches for you.

**3. `CONTEXT.md` should not ship.** The repo root has `CHANGELOG/CONTRIBUTING/README/SECURITY/SUPPORT/TRADEMARKS` and nothing else, and `.gitignore:19-21` puts working notes of exactly this kind (`AGENTS.md`, `CLAUDE.md`) in the local-only bucket. Everything in it that a future translator needs is already the thing you are correctly adding to `CONTRIBUTING.md:78-85`; fold the "grep the field across the other language files before naming a concept" rule into that paragraph and drop the file.

**4. The two surfaces disagree on register, and `CONTEXT.md` states the rule the other one breaks.** `CONTEXT.md` declares "no direct pronoun" and `Strings+Vietnamese.swift` holds that line — zero occurrences of "bạn" in 1060 lines. `Resources/vi.lproj/InfoPlist.strings` then uses it in six of ten prompts ("Mac của bạn", "khi bạn yêu cầu", "bạn tự chọn"). Those are the most visible strings the app has: they appear in the system permission dialog next to macOS's own Vietnamese text. Pick one register for both. For what it is worth, Apple's own Vietnamese system dialogs do use "bạn", so the `InfoPlist` wording looks like the right answer and the `CONTEXT.md` rule like the wrong one — which is another reason not to ship that file as normative guidance.

**5. One language table correctly does not want `vi`, so please leave it alone.** `MediaSupport.swift:851-866` (`recognitionLanguages(for:)`) gives every other of the 13 languages an explicit case and falls through to `["en-US"]`. That is right here: Vision's text recognizer has no Vietnamese, and `MediaService.swift:693` filters the request against `supportedRecognitionLanguages` anyway, so an added `"vi": ["vi-VN", "en-US"]` case would be stripped back to `en-US` at runtime. Mentioning it so a completeness sweep does not "fix" it.

`Resources/vi.lproj/InfoPlist.strings` has all ten keys, `Info.plist` and the `AppLanguage.detect` prefix table are both correct.

PathGao:
One site I missed, and the branch now conflicts.

`CommandBarSystemSettingsSupport.swift:97` (`resourceFolder(for:)`) is the only language switch outside the string catalogs. It is exhaustive, so the compiler will demand a `.vi` case — but it demands a case, not a correct one. The value should be `"vi"`: every plain-code language returns its bare code (`tr`, `ru`, `ja`, `ko`), and only `pt_BR` / `zh_CN` / `zh_TW` / `zh_HK` take Apple's underscore forms. Get it wrong and nothing fails loudly; the System Settings rows just stop matching Vietnamese search words and quietly fall back to English, which is exactly the failure the comment above that function warns about.

Two things I did not verify there: that file is past the truncation point of the diff I can read, so I do not know what it currently returns, and I have not opened a real System Settings extension bundle to confirm the folder is `vi.lproj` — I inferred it from the pattern of the other twelve.

The branch is also CONFLICTING against main now. That is item 1 from my earlier comment arriving as a real merge conflict instead of a field count mismatch, so rebase first — refilling the 32 missing fields before rebasing means hand-translating fields that will have moved again.

**Impact:** none while it sits here. After it lands, a wrong `resourceFolder` value silently removes the Mac Settings source from the command bar for anyone running the app in Vietnamese.
**Merge:** needs a rebase onto main first; the missing catalog fields and the four `MetricsTests` tables from my earlier comment follow from that.

PathGao:
Closing this, and I want to be clear that it is not about the quality of the translation. The catalog was checked and it is clean — no format-specifier mismatches anywhere across it. The problem is drift, and it is structural rather than anything you did.

Every user-facing string in this project is a field of `struct Strings` in `Sources/Vorssaint/Core/Localization.swift`, and **adding a field forces every supported language to supply it — the compiler is the completeness check** (`CONTRIBUTING.md`, "Strings and translations"). That is a good rule for the repo and a hard one for an open translation branch: every feature PR that adds a string breaks every language branch in flight, and this branch stops compiling until the new fields are translated.

Concretely, since this branch was opened, `main` has gained several dozen new fields, and four whole catalogs that have no `vi` case at all:

- `BluetoothSleepStrings.swift`
- `DiskExclusionStrings.swift`
- `MouseClickDebounceStrings.swift`
- `QuitProtectionStrings.swift`

Plus `CommandBarSystemSettingsSupport.swift` needs a `.vi` case — the compiler will stop on that one too. And the gap keeps growing while the branch is open, which is why chasing it has not converged.

**Please do bring Vietnamese back.** The way to make it land is a freeze window: we agree a point where no new strings go in, you translate against that exact commit, and it merges before the next feature PR touches `Strings`. Ping me when you have time for it and I will arrange the window and give you the field list to work from — that turns this from an open-ended chase into a fixed, finishable job.

Sorry for the eight days of silence on our side. That is on us, not on you.

Issue #481, opened 2026-08-04
Menu Bar hidde / collapse icons


BonjeBongo:
### Problem or Motivation

Vorssaint has replaced many disparate apps, but the one still standing and one that hasn't been done consistently to this day remains some form of collapse button to de-clutter the menu bar icons

### Proposed Solution

Add functionality ala Ice/bartender/sanebar/hiddenbar to hide forever / collapse menu bar icons with options for a shelf.

### Feature Area

Other

### Mockups / References

I've returned to Hidden Bar after Ice broke but am still missing a way to have it be a shelf rather than expanding into the cringetastic notch apple insists on having.
<img width="772" height="55" alt="Image" src="https://github.com/user-attachments/assets/34c465e9-922c-47fa-9a14-23ba41a08b4d" />

ruvelro:
Implementation is now consolidated in #360 as a hardened MVP: visible, hidden and always-hidden sections, Settings arrangement with undo, and a secondary bar for notch/overflow.

#627 has been incorporated into #360 with its authorship preserved. Presets, groups, automation, search, hotkeys and additional triggers are intentionally deferred until the macOS 14/15 and macOS 26 interaction substrate is proven.

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

This is the most-requested item in the batch, and it is blocked on review capacity rather than on anyone writing code.

Menu bar collapsing is not implemented on main. #360 is the MVP the maintainer had already LGTM'd, and #824 is stacked on it — roughly 400 commits ahead of main, about 99% of that being #360's own commits, so it cannot be reviewed until #360 lands. Both are currently conflicting with main. A third branch, #700, was closed on 2026-08-24 on timing (it targeted an unreleased macOS 27), so it is no longer part of this.

The useful action here is concentrating on #360 (rebase, then review), not opening another implementation. One technical note recorded against #824: it polls the pointer position every 0.25 s while the feature is on.

PathGao:
Folding two threads in here: **#615** (menu bar tiers with on-demand reveal) and the menu bar half of **#614**. @ruvelro had already asked for #615 to be closed against this issue.

### From #615 (@clovis-dugue)

This is the part worth keeping — it is research this thread did not have.

**State of the alternatives on macOS 26 Tahoe.**

[...]

Issue #698, opened 2026-08-18
Add a macOS 27 backend for the menu bar organizer


ruvelro:
## Context

#360 delivers the hardened menu bar organizer MVP for macOS 14 through macOS 26. The maintainer validated the Tahoe path, but reported that enabling the current implementation crashes on macOS 27 and that its existing WindowServer model does not work there.

The MVP now prevents installation and service startup on macOS 27 and later, shows a localized compatibility notice, and preserves existing configuration for a future backend.

This issue is the single follow-up tracker for macOS 27 support. It is intentionally separate from #360 so the proven macOS 14–26 MVP does not expand again.

## What changed in macOS 27

The available evidence indicates this is a platform model change rather than a small regression:

- Ice reports that hidden items all become visible on macOS 27 and has no linked fix: [jordanbaird/Ice#954](https://github.com/jordanbaird/Ice/issues/954).
- Thaw reports that `CGSGetProcessMenuBarWindowList` now exposes the menu bar host but not individual status-item windows; the menu bar is composited by `MenuBarAgent`: [thaw-app/Thaw#687](https://github.com/thaw-app/Thaw/issues/687).
- Thaw's experimental implementation enumerates each running application's `AXExtrasMenuBar` for direct item attribution, with system items published by `MenuBarAgent`: [MenuBarItemAXProvider.swift](https://github.com/thaw-app/Thaw/blob/feat/macos-27-experimental/Thaw/MenuBar/MenuBarItems/MenuBarItemAXProvider.swift).
- Thaw's preview reorders items through `TrailingItemPreferredPositions`, verifies the resulting AX order, and keeps coordinate-based Command-drag as a fallback. Its releases still describe macOS 27 support as preview with system-item limitations: [macOS 27 preview release](https://github.com/thaw-app/Thaw/releases/tag/macos-27-preview.3).

## Proposed investigation sequence

1. **Read-only enumeration and identity**
   - add a macOS 27 provider based on each application's `AXExtrasMenuBar`
   - derive identity from publishing bundle, AX identifier/description/title and geometry
   - treat `MenuBarAgent` system children separately
   - do not rely on per-item `CGWindowID`

2. **Reordering**
   - probe `com.apple.MenuBarAgent`'s `TrailingItemPreferredPositions`
   - verify every write against fresh AX order
   - keep a bounded coordinate-based Command-drag fallback
   - fail closed for anchored or unverified items

3. **Visibility**
[...]

PR #376, opened 2026-07-27
feat(display): add display management controls

ruvelro:
> [!IMPORTANT]
> **Stacked on #678.** The review-only diff is [feat/display-controls...codex/display-management-controls](https://github.com/ruvelro/vorssaint-utils/compare/feat/display-controls...codex/display-management-controls). GitHub cannot use that fork branch as the base of an existing upstream PR; after #678 merges, this PR automatically shows only the display-management changes below.

## Summary

Adds the display-management features that do not overlap #678:

- per-display resolution picker with Native and available HiDPI scale actions
- draggable multi-display arrangement and main-display switching
- presets for resolution, brightness, arrangement, main display and color profile
- ICC color-profile selection and restore-to-default
- Night Shift and True Tone switches when exposed by macOS
- per-display gamma-table image adjustments: contrast, gamma, gain, warmth, invert, pause and reset
- localized errors and UI copy in all 13 app languages
- a dedicated DisplayManagementSettings view rather than more code in SettingsView

## Scope removed during review

- no second HDR bridge or HDR UI; #678 owns HDR state, readback and routing
- no second automatic-brightness timer; #678 owns synchronized brightness
- no privileged /Library display-override writes
- no duplicate DisplayServices/CoreDisplay symbol loading or brightness probe
- no strings left for those removed features

## Validation

- `./build.sh --test` -> `TESTS OK (8569 checks)`
- `./build.sh` -> release bundle builds
- `./build/Vorssaint --selftest` -> `SELFTEST OK`
- `git diff --check` -> clean
- GitHub CI restarted for 743fcf6

Ref #331
Ref #693

PR #824, opened 2026-08-23
feat(menu-bar-organizer): add toggle shortcut, auto rehide and hover reveal to the menu bar organizer

ruvelro:
> [!IMPORTANT]
> **Stacked on #360; merge #360 first.** The review-only diff is [codex/menu-bar-organizer...feat/menu-bar-organizer-extras](https://github.com/ruvelro/vorssaint-utils/compare/codex/menu-bar-organizer...feat/menu-bar-organizer-extras). GitHub cannot use that fork branch as the base of an existing upstream PR.

## Summary

Adds three opt-in reveal conveniences to the menu bar organizer:

- a global organizer toggle shortcut
- automatic re-hide after a configurable delay, deferred while the pointer or secondary panel is active
- hover reveal after a real 0.5-second dwell on the menu-bar screen that owns the organizer control item

The shared pointer timer clears stale deadlines when disabled, hides the secondary panel with the rest of the organizer, and instantiates shortcut capture only where the feature is supported.

## macOS 27 safety

The #360 safety gate is carried here. Applying a feature preset now writes unsupported availability to false and exits before invoking the service binding, so the organizer singleton remains dormant on macOS 27. The unrelated Defaults join and FeatureCatalog indentation from the stack were also corrected.

## Validation

- `./build.sh --test` -> `TESTS OK (8615 checks)`
- `git diff --check` -> clean
- GitHub CI restarted for c17dc2f

The notched-Mac hover geometry still requires the maintainer requested hardware pass; no such hardware is available in this workspace.

Ref #481


PathGao:
Reviewed `391e2c1e` only, since the rest is #360. The role, defaults, strings and settings registration are all complete and the three pure helpers are properly extracted and tested. The wiring around them has four problems, and the first one means the headline feature does nothing on most Macs.

**1. Auto re-hide never arms in secondary-bar mode — which is the default on any notched Mac.** `armAutoRehide()` is called from `showInMenuBar(_:)` (`MenuBarOrganizerService.swift:541`), but `show(_:)` (`:502`) branches to either `showSecondaryBar()` (`:526`) or `showInMenuBar(_:)` (`:530`). `shouldUseSecondaryBar` returns `true` for `.automatic` whenever `hasNotch` (`MenuBarOrganizerSupport.swift:300-311`), `.automatic` is the registered default (`Defaults.swift:891-892`), and `show(_:)` computes `hasNotch` from `auxiliaryTopLeftArea`/`auxiliaryTopRightArea` at `:517-519`. So on every MacBook Pro and Air from 2021 on, with the shipped presentation mode, both reveal paths — the new shortcut via `toggleHiddenSection()` and hover via `pointerTick()`'s `show(.hidden)` — land in `showSecondaryBar()`, `rehideDeadline` stays `nil`, and the delay field in Settings has no effect. `hiddenSectionShown` also stays `false` in that mode, so `shouldAutoRehide`'s first term would block it even if the deadline were armed. Move `armAutoRehide()` up into `show(_:)`, and decide what "re-hide" means for the panel — presumably `secondaryPanel?.close()`, which `hideAll()` already does at `:177`.

**2. A stale deadline fires after auto re-hide is switched off.** `shouldAutoRehide` (`MenuBarOrganizerSupport.swift:138-145`) takes no "enabled" input, and `syncPointerTimer()` (`:622-640`) only clears `rehideDeadline` when *both* automations are off — with hover still on it hits `guard pointerTimer == nil else { return }` at `:638` and returns with the deadline intact. Turn auto re-hide off while a section is revealed and hover is on, and the section still hides on the old deadline. Either clear `rehideDeadline` in `syncPointerTimer` whenever the auto-rehide key reads false, or pass the flag into `shouldAutoRehide` so the decision stays in the tested helper.

**3. Hover fires on screens that have no menu bar.** `pointerTick()` passes `NSScreen.screens.map(\.frame)` into `isPointerInMenuBarStrip` (`MenuBarOrganizerService.swift:646-649`, helper at `MenuBarOrganizerSupport.swift:126-134`), so the top `thickness + 4` points of *every* attached display count. `NSScreen.withMenuBar` (`AppKitExtensions.swift:33-35`) is this repo's answer to that question and its doc comment exists precisely because picking the wrong screen "points at a screen with no menu bar on it". `show(_:)` in the same file answers it a third way at `:515-517`, by intersecting `controlItem?.frame`. Three answers to "which screen is the menu bar" in one feature; the control item's screen is the one that matches what actually gets revealed. (With "Displays have separate Spaces" on, every screen does have a bar — so whichever way you go, the answer should be derived once and shared, not spelled out per call site.)

**4. The dwell is not the half-second the doc claims.** `hoverExpandDwellTicks = 2` on a free-running 0.25s timer (`MenuBarOrganizerSupport.swift:149-150`) counts *ticks the pointer was already inside*, not elapsed dwell: a pointer entering the strip 1ms before a tick reaches two ticks in ~0.25s. So the real range is 0.25–0.5s, and the comment's "long enough that a pointer passing through never triggers it" holds only at the top of that range. Stamping the entry time and comparing against it, or requiring 3 ticks, makes the number the comment states true.

Smaller:

- `ShortcutCapture.begin()` now touches `MenuBarOrganizerService.shared` unconditionally (`ShortcutCapture.swift:30-34`). `FeatureRuntime.swift:113-119` states the opposite rule for this exact singleton — "skips unavailable ones so their singletons never come to life" — and this service is the one whose `syncWithPreferences` opens with a guard saying macOS 27 crashes on enable. `suspendShortcut()` itself is inert, so this is about the rule, not a crash: gate it on `AppFeature.menuBarOrganizer.isAvailable` like the other silencing paths. The resume side is fine — `ShortcutCapture.end()` reaches it through `featuresToSilenceWhileRecording` → the `.menuBarOrganizer` binding at `FeatureRuntime.swift:196-200` → `syncWithPreferences()` → `syncHotkey()`.
- The 0.25s poll is narrower than it first reads: `syncPointerTimer` only creates it when one of the two automations is on, and both ship off, so it costs idle users nothing. Worth saying in the description, because "polls the pointer 4×/s" is what a reader takes away otherwise. That said, `NSEvent.addGlobalMonitorForEvents(matching: .mouseMoved)` is already the push idiom for this in `FocusFollowsMouseService.swift:58` and `RadialMenuService.swift:440`, and it would give real dwell timestamps for free, which is also the fix for point 4.
- The Carbon hotkey block (`MenuBarOrganizerService.swift:554-608`) is a faithful copy of `ShelfService.swift:335-390`, foreign-signature `eventNotHandledErr` guard included, so it is the eighth copy of a shape the repo has already accepted seven times. Not a request to change it here, but the count is worth recording somewhere.

I could not run any of this, so points 1–4 are read off the control flow rather than observed; point 1 in particular is worth confirming on a notched Mac before you rework it.

ruvelro:
Addressed the review feedback in 68356ad:

- auto-rehide now also works while the secondary panel is open and clears stale deadlines when disabled
- reveal uses the actual menu-bar screen and an exact 0.5-second dwell timestamp
- hide-all closes the secondary panel as well
- shortcut capture no longer instantiates the organizer on unsupported systems
- merged the corrected #360 implementation so the WindowServer bridge is not duplicated

Validation is green: 8,615 local checks, release build, self-test, and both GitHub CI jobs. The branch is mergeable. This is intentionally stacked after #360, so please merge #360 first, then re-review this PR and clear the incomplete label.

PathGao:
The three files where points 1–4 live — `MenuBarOrganizerService.swift`, `MenuBarOrganizerSupport.swift`, `ShortcutCapture.swift` — are all outside the slice of the diff I can read here, so I have not seen what 68356ad actually changed. I'm not clearing `incomplete` on a description of a fix. Point 1 also still wants a notched Mac.

From what I could read, one gap in the macOS 27 gate. `FeatureRuntime.swift:88-93` guards the write but not the binding:

```swift
let joins = selected.contains(feature) && feature.isSupportedOnCurrentSystem
UserDefaults.standard.set(joins, forKey: feature.availabilityKey)
if joins { loadedThisSession.insert(feature) }
Self.bindings[feature]?()          // :93, runs either way
```

The three sibling paths all bail before the binding — `setAvailable` returns early, `setAllAvailable` (`:108`) folds it into the `where`, and the loop nine lines down (`:98`) does too. This one writes `false` correctly on macOS 27 and then runs `Self.bindings[.menuBarOrganizer]?()` anyway, so `MenuBarOrganizerService.shared` comes to life — the thing `:128-129` says this path never does. Fix is `:98`'s shape. Whether a user can reach it depends on whether the feature picker lists the organizer at all on macOS 27, which is in a view file I can't see. And if that gate arrived with the #360 merge rather than the feature commit, it belongs on #360 — I can't tell the two apart in a combined diff.

Cosmetic, but the diff carries them: `Defaults.swift:818` joins `switcherShowFullscreenWindows` and `switcherAppRules` onto one line, and `FeatureCatalog.swift:146` / `:235` de-indent the pre-existing `.bluetoothSleep` cases to four spaces.

**Impact:** on macOS 27 one path instantiates the service the gate exists to keep dormant. Everything else here is cosmetic.
**Merge:** needs points 1–4 confirmed against the full diff first, and #360 landing before that, per your own note.

ruvelro:
Addressed the new review on the stack:

- 7e40e14 (on #360) makes replaceAvailable continue before the service binding for unsupported features, keeping MenuBarOrganizerService dormant on macOS 27
- c17dc2f restores the joined Defaults entries; the Bluetooth Sleep/Menu Bar Organizer indentation is corrected through the updated #360 merge
- the PR body now links the exact feature-only comparison and documents the four behavior fixes

As with #376, GitHub cannot retarget an upstream PR to a base branch that exists only in the contributor fork. The exact diff is: https://github.com/ruvelro/vorssaint-utils/compare/codex/menu-bar-organizer...feat/menu-bar-organizer-extras

Local validation: 8,615 checks, release build, self-test and diff check pass. Both GitHub CI jobs are green. The requested notched-Mac pass remains a hardware check; no notched Mac is available in this workspace, so I have not claimed it was performed.

PathGao:
Re-read the slice I can see. Three things confirmed there: `replaceAvailable` now bails at `FeatureRuntime.swift:92` before the binding, the `Defaults.swift` join is back on two lines, and the `.bluetoothSleep` cases in `FeatureCatalog.swift` are context again rather than re-indented.

Points 1–4 are still in files outside my slice, so `incomplete` stays for those and for the notched-Mac pass.

One thing the gate fix didn't reach, same class. The teardown paths still run for an unsupported feature:

```swift
// AppDelegate.swift:237-239 — availability only
if AppFeature.menuBarOrganizer.isAvailable {
    MenuBarOrganizerService.shared.stop()
}
```

`isAvailable` reads the defaults key, and that key survives a macOS upgrade. Turn the organizer on under macOS 26, upgrade to 27, quit the app: the key is still true, `stop()` runs, and `MenuBarOrganizerService.shared` comes to life on the exact system the gate exists to keep it off. Fresh 27 installs are safe — the registered default is false (`FeatureCatalog.swift:315`) — so this is the upgrade path only.

`setAvailable` (`FeatureRuntime.swift:61-68`) and `setAllAvailable` (`:107-114`) have the same hole from the other side: both guard only the on direction, so flipping the organizer off on macOS 27 writes `false` and then runs `Self.bindings[.menuBarOrganizer]?()` anyway. `replaceAvailable`'s `continue` at `:92` is the shape that covers both directions; the other two want it too, plus `isSupportedOnCurrentSystem` on the AppDelegate guard.

I can't see `MenuBarOrganizerService.init`, so I don't know what instantiation alone costs on macOS 27. The claim is only that it contradicts `FeatureRuntime.swift:129-130` and the "enabling it there crashes" comment at `FeatureCatalog.swift:183-184`.

**Impact:** an upgrader from macOS 26 instantiates the organizer singleton on macOS 27 at quit, and again on any availability toggle. Fresh 27 installs unaffected.
**Merge:** needs points 1–4 readable in a diff I can see plus the three teardown gates above, and #360 landing first per your own note.

ruvelro:
Merged the latest #360 fix into this stacked branch at `512776b0`. The AppDelegate teardown and both FeatureRuntime availability writers now keep unsupported organizer bindings dormant on the macOS 27 upgrade/off paths.

Local validation on the combined branch: 8,615 checks, release build, self-test, and diff check all pass. The remaining non-code gates are still the documented #360-first merge order and the real notched-Mac pass.

ruvelro:
Follow-up: this stack now includes #360 at its current, conflict-free head plus current `main` via `52fa6e1c`. Validation on the final combined head: 8,715 checks, release build, self-test, and `origin/main...HEAD` diff check pass.

PathGao:
Closing alongside #360, same reason. This is stacked on it, and with the macOS 27 gate in place the organizer is switched off on the current system, so this lands follow-up interactions for a feature nobody on macOS 27 can turn on. The reasoning is written up in full on #360.

Keep the branch. When #698 has a macOS 27 backend, #360 and then this go back up together.

One thing to carry over for that day: the body notes the notched-Mac hover geometry still needs a hardware pass that was not available. That will still be outstanding.

PR #832, opened 2026-08-23
feat(url-cleaner): list, edit and explain the cleaning rules (tuned on v3.3.3-beta.1)

PathGao:
Ref #901

Tried on a build of `v3.3.3-beta.1` with the branch on top; the explanations
below exist because that pass turned up questions the page could not answer.

## What was wrong

Clean URL applied two tables nobody could see. The switch that runs it read as a
heading, the rules behind it were invisible, and the one thing the page offered
to edit was a single global text field for adding names.

## The rules, listed and editable

Settings now lists every rule the cleaner applies, grouped by site. Any name can
be switched off, names of your own can be added to one site or to every site, and
a whole site's rules can be turned off at once — two dozen checkboxes is a lot of
clicking to say "not this site".

Edits are stored as a **difference from the built-in tables** rather than a copy
of them, so names added in a later version still reach anyone who has edited
their rules. That is also what "turn a site off" means here: every built-in name
for that site goes into the disabled set and the ones the user added are dropped,
and each can be switched back on one at a time.

Cleaning now names the parameters it removed — in Settings, the menu panel and
the Command Bar action. The automatic rewrite is silent by design, and this is
the only place someone can see that the rules did anything to a link they copied.

## Three things the page now says

Each answers a question that came up while using it, and each sits where the
question is asked rather than in one paragraph at the top.

**What these names are** — "A site attaches these parameters to its own share
links to track where the link came from. Switched on, a name is removed when a
link is cleaned; switched off, it stays." The old caption said only "switch a
name off to keep it in your links": the mechanic without the subject.

**Why the list is long** — "The list covers a site's different share paths — the
web page, the app, a live room — which is why it is long; a real link usually
carries only two to four of them." Bilibili has 25 names because the web share,
the app share, a live room and the story feed each have their own; a real link
carries two or three. Without this, the length reads as "we delete a lot from
your links".

**What a rule matches**, beside the field where one is typed — "Write the name to
the left of the `=`, like `utm_source`. A name that matches takes that one
parameter out of the link and leaves the rest as it was." The cleaner parses the
query into items and filters by name, so position does not matter and nothing is
truncated: `?spm_id_from=333&t=90&vd_source=abc` cleans to `?t=90`.

## The switch says when it acts

"Clean copied URLs" over "Removes tracking parameters from copied links" names a
subject, not a moment. Nothing in either line said the switch is what makes it
happen by itself, on the clipboard, the instant a link is copied — so it read as
a heading for the page it sits on, and the one feature the page is about looked
like it had no switch at all. Both lines name the moment now, in all thirteen
tables, each using the word for the clipboard that table already uses elsewhere.

## Not changed

The rule tables themselves. The two-layer model stays as it is: a global list for
names that mean the same thing everywhere, and per-host lists for names that are
trackers on one site and real parameters elsewhere — `si` is a share token on
YouTube, `t` is a tracker on X and the playback position on YouTube. Matching on
the host is what keeps removing one from breaking the other, and it is why
functional parameters (`t`, `p`, `list`) are absent from every table.

## Tests

`TESTS OK (8233 checks)`, `./build.sh`, `SELFTEST OK`.


vorssaint:
Thanks for the contribution. I tested it locally and merged it.

francisdbillones:
I still want to be able to use an existing ruleset file (#901). There's no conventional format, so it's probably a good idea to use whatever the biggest/most actively maintained one is, which I've identified to be AdGuard's
https://github.com/AdguardTeam/FiltersRegistry/blob/master/filters/filter_17_TrackParam/filter.txt

PathGao:
The storage this PR added should survive that. Rules are kept as a difference from the built-in tables (`host|name` added / `host|name` disabled), not a copy, so swapping the built-in source from the hardcoded tables to a parsed ruleset leaves existing user edits intact — it only changes what "built-in" means.

Two things decide whether importing AdGuard is cheap:

- **License.** FiltersRegistry is GPL-3.0 as far as I know. Bundling the list in the app is a different question from fetching it at runtime. Worth settling before anyone writes a parser.
- **Which subset you parse.** `||youtube.com^$removeparam=si` maps straight onto the per-host table, and the suffix matching for it is already there. What doesn't map: `$removeparam=/regex/` and `@@` exception rules. Silently dropping exceptions makes the cleaner remove *more* than AdGuard would, and over-removal is the failure that breaks a link rather than just leaving one dirty. Either handle both or refuse the lines you can't represent.

I haven't read the filter file itself, so anything beyond the format above is a guess.

PR #839, opened 2026-08-23
Feature: Add compact mode for command bar

kobebrylant:
## Summary

Adds an opt-in compact mode to the Command Bar, following the existing preference, ranking, localization and settings idioms.

**Before:**
image

**After:**

image

image

--- 
An empty field with compact mode on shows nothing but itself ,  just the field with an inline `↓ Suggestions  Esc` hint at its trailing edge. Every list becomes a consequence of typing rather than of opening. Pressing `↓` (or `⌃N`) peeks at the ordinary browse list for the rest of that opening; `Esc` puts the peek away before it closes the bar. Typing, drilling into a category (Emoji, Kill Process, etc.), or entering an argument/confirm/actions/naming card all expand the bar exactly as before - compact mode only suppresses the default Home surface, and the work to build it (`suggestionRows()`, the eleven `categoryHasContent` probes) is skipped, not hidden.

Off by default. Toggled in Settings → Command Bar, no new `AppFeature`, permission, or hotkey.

## Behavior preserved

- Esc still walks back: clear query → leave category (or un-peek) → close.
- Category drill-in shows its full chip row and list unchanged.
- Panel position, growth, and drag behavior are unchanged; the field never moves as the list appears under it.
- Normal mode (the default) is byte-for-byte unchanged - every new branch is gated behind `compact` and reduces to today's behavior when it's `false`.

## Validation

- `./build.sh --test` — TESTS OK (8252 checks)
- `git diff --check` — clean
- `./build.sh` / `--selftest` — OK

New strings (`compactModeToggle`, `compactModeCaption`) are localized in all 13 supported languages


PathGao:
Correcting the label on this one: it was marked `ready to merge` on 08-24, and it no longer merges. #750 landed since and this branch is `CONFLICTING` against `main` @ `9d9c6563`.

Two conflicts:

- `UI/Settings/CommandBarSettings.swift` — #750 replaced the separate caption rows with a single `VStack` ("as separate rows the form drew a divider between every sentence"), and this branch edits the old shape.
- `Tests/MetricsTests.swift` — the Command Bar string count, `147` on `main` vs `148` here.

Neither is a design disagreement; both are "apply your change to the new shape". Could you rebase onto current `main`? Re-labelling to `blocked` until then.

PathGao:
Read this properly now — it had been sitting unlabelled after the rebase. It merges clean against `47df1b50` and the shape is right: the empty-field decision is factored into a pure `CommandBarHome` in `CommandBarSupport.swift`, which is already in `build.sh`'s `--test` allowlist, and every consumer of that state is reached — `showsCategoryChips` (`CommandBarView.swift:333`) goes quiet because `categoryChips` is emptied, the empty state at `:107` because the query is empty with no active category, the footer through the new `isCompactHome` gate.

Two things I checked because they would have been the interesting failures, and both are fine: the saved custom position survives, because `clampedPanelOrigin` (`CommandBarPreferences.swift:320`) and `refreshPanelLayout` (`CommandBarService.swift:372`) are both top-edge anchored and `finishPanelDrag` saves a top-edge offset, so a collapsed bar keeps the same top edge and `resetPanelPosition()` still works. And `show()` calls `reloadPreferenceCaches()` before `refreshResults()`, so the first frame is already collapsed rather than flashing tall.

Three small things:

1. **`↑` and `⌃P` never join the peek.** `kVK_DownArrow` (`:2497`) and `⌃N` (`:2528`) both go through `peekHome()`; `kVK_UpArrow` (`:2494`) and `⌃P` call `moveSelection(-1)` directly, which no-ops on an empty `rows`. The inline hint says `↓ Suggestions`, so this may well be deliberate — but if it is, `⌃P` doing nothing silently is still a rough edge, and if it is not, it is the same mechanism with two members missing.
2. **`SettingsDirectory.swift:137` has no keyword for the toggle.** The Command Bar page carries only `openButton` and `searchPlaceholder`, so searching Settings for "compact" finds nothing.
3. **The collapsed hint drops `⌘K`.** `compactHints` renders `↓ Suggestions Esc`, while the footer it replaces spells out `⌃P ⌃N ↑↓` plus `⌘K` when actions are available (`CommandBarView.swift:770`). The `Esc` is also unlabelled where the footer labels everything.

Localization is complete — both new strings in all 13 initializers, with the per-language count assertion moved 147→149. Both Defaults keys are registered so `exportKeys()` picks them up, and the test asserts it.

None of the three blocks a merge. Labelling `ready to merge`.

PR #841, opened 2026-08-23
feat(app-icon): add new artwork and adaptive macOS 26+ icon support

### LINKED ISSUE #743: [Feature]: New app icon artwork + adaptive `.icon` support (macOS 26+)
## TL;DR

As we discussed privately, I have designed new app icons for Vorssaint, and I'd like to offer them as source assets that plug into the existing generate-at-build pipeline. Alongside the new art, I'm proposing optional adaptive-icon support using Apple's `.icon` format, so the Dock icon correctly follows system Dark Mode, tinting, and the new icon styles on macOS 26+ while macOS 14 and 15 keep rendering exactly as they do today. Before opening a PR I'd like your call on six questions, listed at the bottom.

## What I'd like to contribute

- **An Icon Composer project:** (`.icon` bundle: `icon.json` plus layered sources on a 1024×1024 canvas).
- **Light and dark SVG logo masters:** Vector `.svg` conversions of the `docs/assets/readme/logo.png` and `docs/assets/readme/logo-dark.png` matching the light/dark pair the README already utilizes, allowing for infinite scaling without pixelation.
- The exported & packaged `.icns` exists too for each color variant. Since the repo commits no derived icon binaries anywhere, my assumption is to leave them out and let the build generate whatever it needs. Say the word if you'd rather have the pre-packaged versions committed.

`TRADEMARKS.md` keeps the official name, logo, and icon outside the GPL grant, and thought we have already discussed this privately, I still want to only proceed with your explicit sign-off.

## Why adaptive support fits this repo well

- The repo already builds every icon from one source: `Tools/MakeIcon.swift` reads `Resources/Brand/logo.png` at build time and emits the app iconset, the `.icns`, the menu bar glyphs, and `BrandMark.png`. A `.icon` bundle is the same philosophy one layer up: one master in, six renditions out, nothing derived in git.
- Releases already build on `macos-26` runners (`.github/workflows/release.yml`), where a full Xcode toolchain is present. Compiling the catalog there with `actool` needs no new infrastructure. emacs-plus compiles its `.icon` inside a Homebrew formula, and Skima runs the same step on GitHub's macOS runners in release CI.
- Local development only requires Command Line Tools. The actool step can detect Xcode 26 and skip gracefully when it's absent, so the CLT-only floor stays untouched.
- Shipping both plist keys is a proven pattern: keep `CFBundleIconFile` pointing at the `.icns` and add `CFBundleIconName` for the catalog. psulak/Tahoe-Sequoia-Hybrid-Icon documents exactly this coexistence.

| Keys in `Info.plist`                    | macOS 14/15                       | macOS 26+                                              |
| --------------------------------------- | --------------------------------- | ------------------------------------------------------ |
| `CFBundleIconFile` only (today)         | `.icns`, unchanged                | `.icns`, drawn static inside the system squircle       |
| `CFBundleIconFile` + `CFBundleIconName` | `.icns`, pixel-identical to today | compiled catalog, adapts per appearance and icon style |

## Proposed Integration Plan

**Option A**, which I'd recommend:

1. New sources land under `Resources/Brand/` next to `logo.png`. The README pair goes to `docs/assets/readme/`, following the existing `<picture>` convention.
2. `build.sh` gains an optional step: when Xcode 26 is present, compile the `.icon` bundle with `xcrun actool` into `Assets.car`. When it isn't, print a note and skip.
3. `Resources/Info.plist` adds `CFBundleIconName` alongside the existing `CFBundleIconFile`.
4. `Assets.car` is copied into `Contents/Resources/` during bundle assembly, before signing, so the signing sequence doesn't change.

Two alternatives, briefly:

- **Option B:** commit a prebuilt `Assets.car`, the way Chromium does. Fewer moving parts at build time, but it puts a derived binary in the tree, which the repo currently avoids everywhere.
- **Option C:** ship the new artwork as a regenerated `.icns` only and defer adaptive color support to a later issue. Smallest possible change, but Tahoe users stay on a static-only icon.

## Open Questions

1. Which option do you prefer: A, B, or C?
2. Should the new artwork become the single master feeding everything `Tools/MakeIcon.swift` generates, including the menu bar glyphs and `BrandMark.png`, or should it live alongside `logo.png` as app-icon-only source until the glyphs get their own redesign? I ask because the menu bar canvas contract is tied to `BlackHoleGlyph.pointSize` in `Sources/Vorssaint/App/StatusItemController.swift` and enforced by `--selftest` (see #652), so swapping the master affects more than the Dock.
3. Are you comfortable adding an Xcode-26-gated optional step to `build.sh`, given the project's CLT-only local floor?
4. Any preferences on naming or layout under `Resources/Brand/`?
5. Should the SVG masters and the `.icon` source land as ordinary GPL'd source files, or as trademark-excluded brand assets? CONTRIBUTING.md accepts contributions under GPL-3.0-or-later, but TRADEMARKS.md excludes logo, icon, and branding from that grant, consistent with the fork-rebranding rule. The answer shapes what forks and reuse can expect.
6. Given the one-topic-per-PR rule, would you rather see A or B land as one combined PR, or split into artwork-first plus a build-machinery follow-up?

Happy to open the PR right away once you pick which direction you'd prefer. If C wins, I'll file the adaptive work separately so this stays one topic per PR.

Thank you again for your work on this project!

---

## New Icon Examples

| Dark | Light | Clear Dark | Tinted Dark (Example) |
| ---- | ----- | ---------- | ------------------ |
| Vorssaint Icon - Dark | Vorssaint Icon - Light | Vorssaint Icon - ClearDark | Vorssaint Icon - TintedDark |

divisionseven:
Closes #743, implementing Option A.

## What it does

New artwork flows through the existing generate-at-build pipeline instead of landing as committed binaries:

- `Resources/Brand/AppIcon.icon` holds the Icon Composer source bundle: a single vector layer with appearance specializations
- `Resources/Brand/AppIcon-Default.png` is the Default-rendition master feeding `Tools/MakeIcon.swift`
- Dock-icon rungs now render from the new master. The menu-bar glyph pipeline and BrandMark are untouched; repeated builds produce byte-identical glyph output against `main`

Adaptive support arrives beside the classic path rather than replacing it. When an Xcode 26+ toolchain resolves, `build.sh` compiles the bundle into `Assets.car` with actool and stages it before signing. Without one, the build prints a single skip line and behaves exactly as before, so Command Line Tools-only setups stay unaffected. `Resources/Info.plist` adds `CFBundleIconName` next to the existing `CFBundleIconFile`.

| macOS | Dock and Finder icon |
|---|---|
| 26 and later | Adaptive catalog; the system picks light, dark, tinted or clear from appearance and the user's icon style |
| 15 and earlier | Classic `AppIcon.icns`, same rendering path as today |

## Notes for review

- README gains a small Acknowledgements section crediting the icon work, kept minimal so it can grow with future contributors. The header logo pair also switches to the new SVGs (`docs/assets/readme/logo.svg`, `logo-dark.svg`) for infinite scaling. If you prefer the old raster header, or a different credit type entirely, say the word and I will revert either.
- All new brand artwork includes metadata defining it as trademark-excluded per [TRADEMARKS.md](https://github.com/vorssaint/vorssaint-utils/blob/main/TRADEMARKS.md) and includes links to the [Vorssaint repository](https://github.com/vorssaint/vorssaint-utils); it is not GPL-licensed. Authorship attribution lives in each SVG's Dublin Core metadata.

## Verification

- Warning-free `./build.sh`; `--selftest` passes; unit tests pass (8138 checks)
- End-to-end adaptive compile exercised locally under Xcode 27 beta: actool produced `Assets.car`, staged before signing, bundle seals clean, both plist keys present in the staged app
- Adaptive icon appearance support confirmed working in all styles across repeated builds, final icon appearance is identical to the previews approved in #743.


vorssaint:
Thanks for the artwork and the adaptive icon implementation. I tested it locally and merged it.

PR #842, opened 2026-08-23
fix(media): stabilize More options layout and restore saved image dimensions

ruvelro:
## Summary

Follow-up to #632.

- Keeps one Media scroll container at every window size so expanding **More options** does not move the title or tool picker.
- Restores saved image max-side values through the same 1...20,000 sanitizer used by the resize mode.
- Uses one shared full-width disclosure label across the Media and sibling Settings rows.

The regression test now passes `resizeMode: nil`, so it exercises the legacy fallback that derives the resize mode from `maxDimension` and proves both fields use the same sanitizer.

## Validation

- `./build.sh --test` -> `TESTS OK (8542 checks)`
- release build and self-test -> green on the preceding UI commit
- both GitHub CI jobs restarted for 4a2a0ec


PathGao:
This is `CONFLICTING` against `main` @ `9d9c6563`, and the conflict is **`CHANGELOG.md` only** — the entries from #762, #765 and #752 landed in the same unreleased block. `MediaService.swift`, `MediaWorkspaceView.swift` and `Tests/MetricsTests.swift` all auto-merge clean.

Could you rebase? It is a one-file resolution and the fix itself is untouched.

PathGao:
Separate from the rebase: the scroll fix and the disclosure fix both re-answer a question the codebase already answers somewhere else.

**1. The scroll container already exists in this file, gated on `compact`.** `MediaWorkspaceView.swift:158-163` is already `if compact { ScrollView { content }.frame(maxHeight: 430) } else { content }` — the workspace's own answer to "does this content scroll". This PR adds a second answer at `:15`, outside `MediaSettings`, and leaves the branch alone, so the Settings page and the panel now scroll through different containers. It also changes what scrolls: the outer one takes `header` and `toolPicker` (`:153-157`) with it, so in Settings the title row and the tool picker scroll off the top, while in the panel they stay pinned. Dropping the gate — `ScrollView { content ... }.frame(maxHeight: compact ? 430 : .infinity)` — is a smaller diff, keeps one scroll container, and keeps the picker reachable while the tall converter card is expanded.

**2. The disclosure-label pattern is already written, and four siblings still lack it.** `MonitorPanelConfig.swift:108-122` is the same fix you wrote at `MediaWorkspaceView.swift:360-366`: a label with `.frame(maxWidth: .infinity, alignment: .leading)`, `.contentShape(Rectangle())`, `.onTapGesture { toggle }`. Yours is the second copy. If the premise in your comment holds — only the chevron toggles a stock `DisclosureGroup` on macOS — then these still take chevron-only clicks:

- `ScreenRecorderSettings.swift:116` — `DisclosureGroup(strings.moreOptions, isExpanded: $showsMoreOptions)`, literally the same "More options" row
- `CommandBarSettings.swift:146` — `DisclosureGroup(FeatureStrings.recorder(...).moreOptions, ...)`, also "More options"
- `WhatsAppDownloadsSettings.swift:274` — `DisclosureGroup(text.fileTypes)`
- `URLCleanerSettings.swift:83` — `DisclosureGroup(l10n.s.urlCleanerRulesAddSite)`

The two `moreOptions` ones are the same control as the one you fixed; fixing one of three and leaving two is what makes it read as inconsistent rather than as a macOS default. A small `disclosureLabel(_:isExpanded:)` helper applied at all five sites would be the whole change, and would stop the pattern from getting copied a fourth time. I have not run this, so I am going on the premise your comment states, not on observed click behaviour.

**3. The profile-restore fix is right, and it leaves the disagreement it works around in place.** `MediaImageOptions.init` (`MediaSupport.swift:284-296`) runs `maxDimension` through `sanitizedPixelDimension(_:fallback:)`, clamped to 64...7680 and forced even (`MediaSupport.swift:396-399`), while `MediaImageResizeMode.init` runs the same number through `sanitizedImageDimension(_:fallback:)`, clamped to 1...20000 (`MediaSupport.swift:401-404`) — and the stepper offers 64...20000 (`MediaWorkspaceView.swift:735`). Reading from `resizeMode` is the correct side to read from, but `options.maxDimension` is now dead for images (its only remaining consumer is the *video* struct's own field at `MediaService.swift:286`) while still being persisted into every saved profile with a silently different value. Either drop it from `MediaImageOptions` or route its init through `sanitizedImageDimension` too, so the two fields cannot disagree again.

**4. Three unrelated fixes in one PR.** The title names the Settings shift; the branch also carries the profile `maxDimension` restore and the TIFF/Exif strip in `MediaService.swift:925-939`. The CHANGELOG entries cover all three, but the title and the review both benefit if the metadata change is split out — it is the only one that changes the bytes of a produced file.

ruvelro:
Addressed the review feedback in 5ec7154 and updated the PR title to match the final scope:

- kept a single scroll container so expanding More options no longer shifts the Settings header/tool picker
- introduced one shared full-width disclosure label and applied it to the sibling Settings rows called out in review
- aligned `MediaImageOptions.maxDimension` with the same sanitizer used by `resizeMode`
- removed the unrelated TIFF/Exif byte-rewrite change from this PR
- rebased the changelog entry into the current unreleased section

Validation is green: 8,542 local checks, release build, self-test, and both GitHub CI jobs. The branch is mergeable. Please re-review and clear the `incomplete` label if everything looks good.

PathGao:
Rebase and the four sibling sites check out. One thing has to come out before this lands, plus a body edit.

**1. `Defaults.swift` picked up a stray hunk from the rebase.** `Defaults.swift:804` — nothing in this PR should touch that file:

```swift
DefaultsKey.switcherShowFullscreenWindows: true,        DefaultsKey.switcherAppRules: [String: String](),
```

Two dictionary entries jammed onto one line. It compiles, no behavior change, and `git diff --check` can't see it because there's no trailing whitespace. Drop the hunk.

**2. The PR body still claims the TIFF/Exif change you removed.** Body: "the nested TIFF `Orientation` and Exif `PixelXDimension`/`PixelYDimension` tags ... are now cleared/updated to match the rendered bitmap." There is no `MediaService.swift` in the diff. The CHANGELOG entries are already correct, so it's only the body that's stale — including the Validation block, which still says 7405 while your comment says 8,542.

**3. `MonitorPanelConfig` swapped `toggle(id)` for `expansionBinding(for: id)`.** `MonitorPanelConfig.swift:113`. The old label called `toggle(id)`; the new one flips a binding directly. `expansionBinding(for:)` and `toggle(_:)` are both outside the diff, so I can't tell whether `toggle` did anything beyond flipping that same binding — persisting it, or collapsing the other sections. If it's literally `binding.wrappedValue.toggle()`, ignore this.

The scroll container, the shared label, and the sanitizer alignment are all what I asked for, and the new test does fail against the old `sanitizedPixelDimension` path, so it pins the right thing. Sibling coverage is complete for the four sites I named — but I only see the diff, so I can't tell whether other `DisclosureGroup` call sites exist beyond those five.

**Impact:** none at runtime from the stray hunk; unknown for item 3 until `expansionBinding(for:)` is compared against `toggle(id)`. The stale body is what misleads whoever reads this PR later.
**Merge:** needs the `Defaults.swift` hunk dropped first, and the body's TIFF/Exif bullet removed.

ruvelro:
Follow-up pushed in c38a8b4:

1. Removed the stray Defaults hunk by restoring switcherShowFullscreenWindows and switcherAppRules as separate dictionary entries.
2. Updated the PR body: the TIFF/Exif claim is gone and Validation now reports 8,542 checks plus the current build/self-test results.
3. Confirmed toggle(_:) only inserted/removed the same ID from expandedBlocks. The binding setter has identical behavior, so I removed the now-unused toggle helper to make that equivalence explicit.

The release build and self-test pass locally, and both GitHub CI jobs are now green on this commit. This is ready for re-review; please clear the incomplete label if everything looks good.

PathGao:
All three are done.

1. No `Defaults.swift` in the diff.
2. The body's TIFF/Exif bullet is gone, and Validation reads 8,542 / c38a8b4.
3. `toggle(_:)`'s body is in the diff now as a deletion, and it is exactly insert/remove on `expandedBlocks`. `expansionBinding(for:)` is still outside the diff, so its setter is your word rather than something I checked — but the side I can see matches what you said.

One non-blocker in the new test (`Tests/MetricsTests.swift:4653-4660`): you pass `resizeMode: .maxDimension(12_000)` into the init, so `largeSideOptions.resizeMode.maxDimension == 12_000` re-reads what you just handed it and cannot fail. The first conjunct is the one that fails against `sanitizedPixelDimension`, so the regression is pinned. Passing `resizeMode: nil` would make the test pin what its name claims — both fields landing on the same sanitizer through `resizeMode ?? .maxDimension(self.maxDimension)`.

Still diff-only on my side: whether other `DisclosureGroup` call sites exist beyond the five, and the runtime behaviour of the single scroll container, are unverified.

**Impact:** none.
**Merge:** ship it.

ruvelro:
Applied the non-blocking test improvement in 4a2a0ec: largeSideOptions now passes resizeMode: nil, so the assertion exercises resizeMode ?? .maxDimension(self.maxDimension) and proves both fields arrive at 12,000 through the same sanitizer. The unit suite remains green at 8,542 checks; Both GitHub CI jobs are green.

Issue #755, opened 2026-08-22
Maybe bug? Crosshair pointer is shown in color picker preview window


I-Have-No-Idea-What-Im-Doing-Right-Now:
### Description

When using the color picker, the cursor is shown in the preview/magnifier window for the color picker instead of a highlight around the selected pixel. This might not be a bug but it changed for seemingly no reason so I assume it is.

### Steps to Reproduce

Open vorssaint color picker (in my testing, reproducible via any shortcut/method of opening)
Look at the magnifier window

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

AB-boi:
Yeah you mean the macos cursor right? agreed.

PathGao:
Tracking PR: #759 (open, currently labelled `incomplete`).

Worth recording: the pointer in the capture is not the system cursor — `ScreenshotSupport.swift:113` already excludes that. What covers the pixel is the crosshair the app draws itself in `ScreenshotSelectionController.swift:1025-1045`, which is also what #781 is about.

I-Have-No-Idea-What-Im-Doing-Right-Now:
> The fix for this is out now, in **3.3.3**.
> 
> @I-Have-No-Idea-What-Im-Doing-Right-Now, could you update to that version and say whether it fixed things
> on your Mac? We cannot tell from here — yours is the setup that found this.
> 
> Please say **how much of it is fixed**, not only whether it is. If part of what you
> reported still happens, or something related turned up, say so here and the issue goes
> back for another look — this is not a last call.
> 
> If nothing arrives in two weeks the issue will be closed as it stands, and a comment
> reopens it at any point after that.
> 
> 
> 

The bug looks fixed to me in the latest version. Although the magnifier does look different it no longer shows the macOS cursor.

vorssaint:
the magnifier no longer shows the cursor in your test of the latest version. thanks for confirming, closing this as fixed

PR #844, opened 2026-08-23
feat(capture): rebuild the capture magnifier with pixel grid, color readout, pixel nudge and quick copy

ruvelro:
## Summary

The capture loupe, toggled with **Z** and automatic for the color picker, now provides a pixel-accurate magnifier and precision controls:

- **Pixel grid and centre highlight:** every sample has an odd side, so the pointed pixel is a real middle cell with a double border.
- **Info bar:** shows a swatch, the configured color format, and display-pixel coordinates.
- **Useful zoom range:** the 132 pt loupe derives its maximum zoom from the smallest 3x3 sample (13 / 3, about 4.33x), avoiding wheel travel that cannot change the rendered pixels.
- **Arrow-key nudge:** moves by one device pixel, or ten with Shift, across displays. Cursor warps are bracketed so physical mouse input resumes immediately.
- **Consistent copy:** **C** copies the exact pixel and frame currently shown by the ring/readout. In live selection, toggling Z refreshes the loupe snapshot.
- **Opt-in start setting:** region captures can open with the magnifier already enabled. The color picker and setting use the same toggle/load path.

Arrow and C events are only consumed while the loupe is active and no drag is in progress.

## Validation

- `./build.sh --test` -> `TESTS OK (8551 checks)`
- `git diff --check` -> clean
- GitHub CI restarted for c0d80a5


PathGao:
The magnifier itself reads well — odd sample side, the ringed centre cell, the readout bar, keyboard nudging. Three things before it can be merged, one of them about a different PR.

**1. This is the same change as #759.** The colour picker and the capture magnifier are the *same* loupe: `ScreenshotSelectionController.swift:201` forces `loupeEnabled = true` when `isPickingColor`, and both render through `drawCaptureLoupe`. #759 "rings the pixel the loupe samples" — which is what `drawLoupeCenterCell` does here with a white 1.5pt rect. Two PRs changing one function for the same reason; they will conflict and one is redundant. Worth agreeing which one carries it before either lands.

**2. The defect this PR names is left standing in the helper it just edited.** The new doc comment on `captureLoupeSampleSide` says the sample side must be odd "so the pixel under the pointer is a real middle cell that the grid can outline instead of a boundary between two cells". `ScreenshotSupport.swift:1439` still declares `cropLoupeSampleRect(..., sideLength: CGFloat = 14)` — even — and `ScreenshotEditorView.swift:376`, the editor's own crop loupe, uses exactly that default. So after this PR the app has two magnifiers for the same job: one 132pt with a pixel grid, colour readout, zoom and keyboard nudge, and one 72pt with a plain crosshair sampling an even-sided rect.

**3. The changelog entry is in the wrong release.** The four added lines land at `CHANGELOG.md:138`, which is inside `## [3.3.2] - 2026-08-20` → `### Added` (headers at the merge base: 3.3.3-beta.2 at line 7, 3.3.3-beta.1 at 20, 3.3.2 at 106). A new feature is being appended to a version released five days earlier. The current `Added` block is at line 9.

Smaller:

- `nudgePointer` re-implements two existing helpers: `NSScreen.screens.first(where: { $0.frame.contains(location) }) ?? NSScreen.screens.first` where `NSScreen.withMouse` exists (`AppKitExtensions.swift:14`), and `NSScreen.screens.first?.frame.height` for the coordinate flip where `NSScreen.withMenuBar` exists (`AppKitExtensions.swift:33`) and documents `screens.first` as only its fallback. It is copying the idiom already at `ScreenshotSelectionController.swift:169`, so this is a third copy rather than a new bug — but the shared helper is right there.
- Arrow keys clamp to the current screen's `frame`, so the pointer cannot be nudged across a display boundary in a multi-monitor capture.
- `C` copies from `loupeImage`, which in non-freeze mode is a one-shot snapshot taken when the loupe was toggled on (`loadLiveLoupeImages`, `:216`). Reading a palette off a live screen copies stale pixels with nothing indicating it.
- The readout bar hardcodes `format: .hex`, but `C` copies in the user's configured `colorPickerFormat` and then displays that for 1.4s, so the bar changes format mid-session.
- `refreshPointerState` now also writes `lastDragPoint` when `isDragging`, so arrow keys resize an in-progress selection. Not mentioned anywhere.

PathGao:
Ref #755

ruvelro:
Thanks @PathGao 🚀

ruvelro:
Rechecked after the follow-up commit and after #759 merged. The magnifier now uses an odd capture sample centered on the pointer, handles cross-display nudging via the active screen, refreshes pixels before the C copy action, preserves the configured color format, and disables nudging while a drag is active.

Validation: local unit suite (8,550 checks), release build, and self-test all pass; both GitHub CI jobs are green and the PR is mergeable. This is ready for another review; please clear the incomplete label if everything looks good.

PathGao (commented):
Rechecked against the new diff. The changelog is in the right block now, the format is preserved, the nudge uses the shared screen helpers and drags are excluded. Five things left, one of them the headline feature.

**1. Every zoom above 3.25 draws the same picture.** `captureLoupeSampleSide` quantizes to odd integers with a floor of three:

```
zoom 3.25 → raw 4.00 → side 5
zoom 3.26 → raw 3.99 → side 3
zoom 4.00 → raw 3.25 → side 3
zoom 10.0 → raw 1.30 → side 3
```

The frame is a fixed 132 pt, so from 3.26 upward the loupe renders a 3×3 sample at 44 pt per cell — the exact number the body attributes to 10x. At 1.15 per tick that is about eight scroll ticks that change nothing, and eight more before anything moves on the way back down. The tests contain both halves of this (`captureLoupeSampleSide(zoom: 10) == 3`, and the odd-side loop covering 4 and 10) and read them as a pass.

The ceiling is a second copy of a fact `captureLoupeSampleSide` already owns. Derive it:

```swift
static var captureLoupeMaxZoom: CGFloat { captureLoupeBaseSampleSide / 3 }  // 4.33
```

That still reaches the 3×3 sample and leaves only the ~2-tick band the old ceiling of 4 already had.

**2. `CGWarpMouseCursorPosition` suppresses local hardware mouse events for ~0.25 s.** After each arrow press the physical mouse goes dead for a quarter second unless the warp is bracketed:

```swift
CGAssociateMouseAndMouseCursorPosition(0)
CGWarpMouseCursorPosition(...)
CGAssociateMouseAndMouseCursorPosition(1)
```

I can't test this on hardware — one arrow press followed immediately by a mouse move will show it.

**3. In live mode the readout bar and C read different frames.** The new doc comment on `formattedValue` says the bar can "show exactly what a copy will put on the pasteboard". `copyLoupeColor` re-captures every display first when `!freeze`, so what C copies is by construction not what the bar showed. The magnified image is still the snapshot taken when the loupe was toggled on, so the pixel gets aimed at with stale content and copied from fresh content — if a window moved in between, C copies a color that was never on screen where the ring was. The ✓ makes that detectable after the fact, not before. Either refresh the live loupe on the same cadence as the copy, or drop the promise from the doc comment and the body.

**4. Two paths now mean "start with the loupe up", only one loads the live image.**

```swift
if isPickingColor { loupeEnabled = true }
else if ...loupeStartsOn { loupeEnabled = true; if !freeze { loadLiveLoupeImages() } }
```

Either the color-picker path is missing the load, or the second branch is `toggleLoupe`'s body copied inline. `toggleLoupe()` isn't in the diff so I can't tell which — but both branches should call it instead of re-deriving it.

**5. The arrow and C cases match unconditionally, then bail inside the handler.** `nudgePointer` and `copyLoupeColor` both start with `guard loupeEnabled`. Every other case in that switch decides in the pattern (`_ where Self.isLoupeKey(event)`); these two decide in the body. If the monitor consumes handled keys, arrows and C are swallowed in every capture session including with the loupe off and mid-drag. The monitor's return isn't in the diff, so I can't confirm it swallows.

Still open from last round, downgraded from blocking: `cropLoupeSampleRect(..., sideLength: CGFloat = 14)` is even, and the doc comment two functions up now states the sample must be odd; `ScreenshotEditorView.swift:376` still takes that default. Not a defect in this diff, but the file asserts an invariant its own default violates.

**Impact:** roughly two thirds of the new wheel travel does nothing; live-mode C can copy a color the readout never showed; likely 0.25 s of dead mouse after each nudge.
**Merge:** needs the dead zoom range above 3.25 closed first.

ruvelro:
Addressed all five points in c0d80a5:

1. captureLoupeMaxZoom is derived as baseSampleSide / 3 (about 4.33), and the tests now pin the derived ceiling and 3x3 endpoint.
2. Cursor warps are bracketed with CGAssociateMouseAndMouseCursorPosition(0/1).
3. C now copies from the same loupe image used by the ring and readout. In live selection, toggling Z refreshes that snapshot, so the shown and copied frames cannot diverge.
4. Both automatic start paths call toggleLoupe, including the color picker, so live pixels load through one path.
5. C/arrows only match while the loupe is active and no drag is in progress; unhandled key events are returned instead of swallowed.

The body now describes the 4.33 ceiling and snapshot semantics rather than 10x/fresh-copy claims. Local validation: 8,551 checks, release build, self-test and diff check pass. Both GitHub CI jobs are green.

PathGao:
c0d80a5 clears all five. Verified against the diff: the ceiling is derived from the sample floor (~4.33, 3×3 at the top), the warps are bracketed, C and the readout bar read the same `loupeImage`, both auto-start paths go through `toggleLoupe`, and arrows/C are gated before they run. Two things left, one of them from that last fix.

**1. `default: return event` widens the change from "arrows and C" to every key.** The switch used to end `default: break` / `return nil`, so the monitor swallowed everything for the length of a capture session. Now anything unmatched is handed back. The body claims only "Arrow and C events are only consumed while the loupe is active and no drag is in progress" — this also releases arrows during a drag and every other key in the session. If `installKeyMonitor` is a local monitor and the overlay has no `keyDown` (neither is in the diff, so I can't confirm), a plain `a` mid-capture now walks the responder chain and beeps.

Narrow it to the two keys instead of the fall-through:

```swift
case _ where Self.isCopyColorKey(event):
    guard self.loupeAcceptsKeyboardActions else { return event }
    self.copyLoupeColor()
case _ where Self.isNudgeKey(event):
    guard self.loupeAcceptsKeyboardActions else { return event }
    self.nudgePointer(keyCode: Int(event.keyCode),
                      fast: event.modifierFlags.contains(.shift))
default:
    break
}
return nil
```

**2. The smallest sample is still written twice.** `captureLoupeSampleSide` floors at `max(3, odd)`; `captureLoupeMaxZoom` divides by its own literal `3`. Those two numbers disagreeing *is* the dead range from last round, and nothing stops it coming back — `static let captureLoupeMinSampleSide: CGFloat = 3` in both places ends the class. Same shape in the test: `captureLoupeGridVisible(frameSide: 132, ...)` retypes the number `captureLoupeFrameSide` now owns, so changing the frame side leaves the test green while the grid silently drops out at some zooms.

**Impact:** every unmapped keystroke during a capture session now dispatches instead of being dropped; the second point has no effect today.
**Merge:** needs the monitor fall-through narrowed to the two keys first.

ruvelro:
Addressed the final review and the new conflict with `main`:

- `ee52e500` narrows event passthrough to inactive/dragging C and arrow-key actions; unrelated keys keep the capture monitor prior swallowed behavior
- the minimum loupe sample side is now a single constant used by both sampling and maximum zoom
- the grid test now uses `captureLoupeFrameSide`
- `0bb91776` merges current `main`; the only manual conflict was the changelog, where both entries were preserved

Validation after the merge: 8,651 checks, release build, self-test, and diff check all pass. The PR is conflict-free again; please re-review and clear `incomplete` if everything looks good.

PathGao:
`ee52e500` clears both. The passthrough is narrowed to the two keys with `default: break` restored, `captureLoupeMinSampleSide` is now the single source for both the sample floor and the zoom ceiling, and the grid test reads `captureLoupeFrameSide`. One thing left, in a function this PR introduces.

**The `kVK_ANSI_C` fallback in `isCopyColorKey` never runs.**

```swift
if let typed = event.charactersIgnoringModifiers?.lowercased(), !typed.isEmpty {
    return typed == "c"
}
return Int(event.keyCode) == kVK_ANSI_C
```

`charactersIgnoringModifiers` is non-empty for every printable key, so the first branch always decides and the second is unreachable for any key that could carry `kVK_ANSI_C`. On a Cyrillic layout the key labelled C reports `с` (U+0441), the comparison fails, and C is dead for the whole session — while the guide keeps showing the C hint, since that hint is driven by `loupeEnabled`, not by whether the key can match. The app ships a Russian localization.

Make it an or:

```swift
if event.charactersIgnoringModifiers?.lowercased() == "c" { return true }
return Int(event.keyCode) == kVK_ANSI_C
```

`isLoupeKey` isn't in the diff. If Z resolves the same way, Z is dead on those layouts too and has been for longer — worth checking before deciding where the fix lands.

Smaller, no effect today: the image-pixel-to-index clamp is written twice.

```swift
let x = min(max(Int(pixelPoint.x.rounded(.down)), 0), image.width - 1)
```

Once in `loupePixelColor`, once at the top of `drawLoupeInfo` — which then calls `loupePixelColor` and lets it recompute the same pair. The coordinates the bar prints and the pixel the swatch samples are two independent derivations of one fact, agreeing only because the lines are identical. Returning the clamped point from `loupePixelColor` ends it.

Not verified: the changelog hunk after `0bb91776` carries no version header in its context, so I can't tell from the diff alone which release block line 13 now sits under. It was right before the merge and the neighbouring entries still read as unreleased.

**Impact:** C copies nothing on non-Latin keyboard layouts, with the on-screen hint still offering it.
**Merge:** needs the C keyCode fallback made reachable first. Leaving `incomplete` on for that one line.

ruvelro:
Addressed the non-Latin layout follow-up in `d920c4d3`:

- C now matches either the promised character or physical `kVK_ANSI_C`, so the keyCode fallback is reachable
- I checked Z as requested and found the same pattern; Z and the adjacent S capture shortcut now use the same shared matcher
- the pixel clamp is derived once and the sampled color and displayed coordinates consume that same result

Local validation: 8,651 checks, release build, self-test, and `origin/main...HEAD` diff check pass. Please re-review and clear `incomplete` if this closes the remaining keyboard-layout issue.

PathGao:
`d920c4d3` closes it. `matchesShortcutKey` reaches the keyCode side, Z and S route through the same matcher, and `loupePixelColor` returns the clamped point that `drawLoupeInfo` prints instead of recomputing it. One thing the shared matcher brought in with it, not a blocker.

**The keyCode fallback is unconditional, so every shortcut gains a second, unlabelled key on non-QWERTY Latin layouts.**

```swift
return event.charactersIgnoringModifiers?.lowercased() == character
    || Int(event.keyCode) == physicalKeyCode
```

On QWERTZ the key labelled Y sits at `kVK_ANSI_Z`, so Y toggles the loupe. On AZERTY it's W. On Dvorak `;` toggles the loupe, `o` starts scrolling capture and `j` copies the color. Z and S did not do this before — the if/else they had stopped at the character whenever one existed, and the body only claims C became reachable. German and French are both shipped localizations.

That is the form I suggested, and it buys Cyrillic reachability at this price. Gating the fallback on the layout not producing a Latin letter keeps both:

```swift
let typed = event.charactersIgnoringModifiers?.lowercased() ?? ""
if typed == character { return true }
// reach for the physical key only where the layout can't type the letter
return typed.range(of: "[a-z]", options: .regularExpression) == nil
    && Int(event.keyCode) == physicalKeyCode
```

The layout mappings above are from the standard layouts, not from a machine — I have no way to press a QWERTZ key here.

Smaller: `copyLoupeColorUnderPointer` picks `isDragging ? lastDragPoint : hoverPoint`, but its only caller guards on `loupeAcceptsKeyboardActions`, which is false whenever any panel is dragging. The branch can't run, and it asserts the opposite invariant from the guard — the two disagree about whether a copy during a drag is a thing that happens.

**Impact:** on German, French and Dvorak input sources one unlabelled key toggles the magnifier, and on Dvorak two more fire scrolling capture and the color copy. Nothing else.
**Merge:** ship it. The layout widening is worth a follow-up line, not a sixth round.

PR #845, opened 2026-08-23

pboucher:
## Problem

When the App Switcher commits a selection, the target app is brought forward, but the menu bar keeps showing the previous app, so activation never fully transfers.

If you use the App Switcher to switch to Vorssaint itself first, switching to another application will always succeed.

## Cause

macOS 14+ cooperative activation only lets an app hand off activation it currently owns. The switcher panel is a non-activating panel and Vorssaint itself is `LSUIElement`, so when an app is selected, Vorssaint rarely owns activation. `NSApp.yieldActivation(to:)` is therefore a silent no-op, and the `activate(from:)` call that follows is refused by the system.

## Fix

Self-activate first, then yield. `NSApp.activate(ignoringOtherApps: true)` runs synchronously in direct response to the hotkey release, which macOS treats as a genuine user gesture and grants; the subsequent yield then has activation to hand off.

Both call sites that transfer activation to another app go through the same ordered handoff:

- `WindowActivator.activateApp(_:allWindows:)`
- `WindowActivator.activateSource(pid:windowID:windowOwnerPID:)`

Guarded appActivated against Vorssaint's own PID, matching the same guard already used elsewhere in this subsystem. A deliberate switch onto Vorssaint's own window is unaffected, since AppSwitcher.commitSession already records that target directly through recordSwitch before activation happens.

## Testing

- Added assertions in `Tests/MetricsTests.swift` covering the handoff order, including that `.selfActivate` comes first (if it does not, the yield that follows is a no-op).
- Manually verified with the App Switcher that the target app is activated and its menu bar updates.

Ref #961


PathGao:
The premise looks right, and if it is, three more call sites need the same treatment.

The doc comment added to `SwitcherActivationHandoffStep` says:

> `WindowActivator.activateApp(_:allWindows:)` and `.activateSource(pid:windowID:windowOwnerPID:)` are the two call sites that hand activation to another app

`git grep "NSApp.yieldActivation" origin/main` finds four:

- `Services/DockClick/DockClickService.swift:669`
- `Services/Switcher/SpaceHop.swift:90`
- `Services/SystemMonitor/ProcessUsageService.swift:154`
- `Services/Switcher/WindowActivator.swift` (the one this PR fixes)

The first three all do `NSApp.yieldActivation(to: app)` with no `NSApp.activate(ignoringOtherApps:)` in front of it — exactly the sequence this PR argues is a silent no-op for an `LSUIElement` app that does not currently own activation. `DockClickService`'s own comment even reads "the same sequence the Switcher, Space hop and process list already use", which is now only true of the three that were not changed.

So either the premise holds and all four are broken, or it does not and this one change is unnecessary. Either way the answer should be the same in all four places, and `activationHandoffSteps` is the right thing to route them through.

Could you extend it to the other three, and correct the doc comment's count? Labelling `incomplete` for the scope; the mechanism itself is the right shape.

PathGao:
Two more, both about where the self-activation is allowed to happen.

**`activateSource` doesn't have the gesture the body leans on.** From the Fix section:

> `NSApp.activate(ignoringOtherApps: true)` runs synchronously in direct response to the hotkey release, which macOS treats as a genuine user gesture and grants

That holds for the commit path. `activateSource(pid:windowID:windowOwnerPID:)` is the restore-the-source path and its callers aren't in this diff, so I haven't verified them — but if any of them fire from an observer (target minimized, target quit) rather than from a key event, the self-activate there is either refused, which puts the yield right back to being a no-op, or granted and flashes a Vorssaint window forward. Same question at the three sites from my earlier comment: `DockClickService` is a mouse click, `ProcessUsageService` is a list selection, `SpaceHop` I don't know. "Self-activate first" being the correct order doesn't mean it's grantable at every site, and the doc comment currently reads as if it is.

**The `appActivated` guard is wider than the body claims.** Body:

> A deliberate switch onto Vorssaint's own window is unaffected, since AppSwitcher.commitSession already records that target directly through recordSwitch before activation happens

`commitSession` covers the switcher path. The guard is unconditional, so every other way of activating Vorssaint — dock icon, clicking one of its windows, opening Settings — stops promoting it as well. Those promoted before this PR and nothing else records them now, so Vorssaint sinks in its own MRU ordering right after the user was using it. Scope the guard to the handoff rather than to the PID: a timestamp stamped in `performActivationHandoff` and checked here is about the cheapest version. Unless there's another recorder for Vorssaint that I can't see — `promote` and `recordSwitch` are both outside this diff — in which case say so and I'll drop this one.

**Impact:** Vorssaint stops being promoted in the switcher's own ordering on every activation path, not just the synthetic one this PR introduces. The `activateSource` half is a question, not a confirmed break.
**Merge:** needs the `appActivated` guard narrowed to the handoff first, on top of the three call sites from my earlier comment.

ref #961

PathGao:
Picked this up as #1278 — you are credited as co-author on the commit, the diagnosis and the fix are yours.

Two things changed on top of it. There were five `NSApp.yieldActivation` sites, not two: `SpaceHop.swift`, `ProcessUsageService.swift` and `DockClickService.swift` were still yielding bare, so those paths stayed broken. All five now go through one `ActivationHandoff` helper, and a test walks `Sources` and fails if a bare yield appears anywhere else — that assertion fails on this branch as submitted, which is what convinced me the helper was worth a file.

The other change is the MRU guard. `WindowUseTracker.appActivated` is the only place an app is recorded as used at all, so guarding unconditionally on our own pid would also have stopped the Dock icon, Settings and switching onto one of Vorssaint's own windows from being recorded. It is now armed by the handoff and expires.

Closing this one so the review stays in one place. Thanks — this had been reported twice (#961, and #868 which was closed and came back).

PR #851, opened 2026-08-23
Add Charge Limit and battery calibration

### LINKED ISSUE #846: Stop Battery charge above X% (80%) for better battery health.
### Problem or Motivation

Currently I am using a different AlDante for this, but its frustrating to have multiple utility apps.

### Proposed Solution

Hope Vorssaint can do this, currently using AlDante for this.

### Feature Area

Battery

### Mockups / References

_No response_

### LINKED ISSUE #849: [Feature]: Charge limit and battery calibration
### Problem or Motivation

Keeping a MacBook at 100% for long stretches, especially when it stays plugged in, wears the battery faster than holding a lower ceiling. People currently install a second app such as AlDente just to stop charging around 80% and to run an occasional full calibration cycle. That is two menu bar utilities for what belongs in Vorssaint's energy tools.

Related request: #846.

### Proposed Solution

Add an opt-in Charge Limit feature under Energy and display:

- A charging ceiling from 20% to 100% (80% by default). Charging stops at the cap while the Mac still runs from the adapter.
- Discharge down to the cap when the battery is already above it.
- A one-click calibration cycle: charge to 100%, discharge to 10%, charge to 100% again, hold for an hour, then return to the saved limit.
- An AlDente-style circular dial in the menu bar panel and under Settings → Energy.
- A protected SMC helper, same Login Items pattern as Fan Control. The limit is enforced while Vorssaint is running; force-discharge restores itself if the helper loses the app heartbeat.

Off by default in Features. Apple Silicon uses the charging-control keys the Mac actually exposes (CH0B/CH0C or CHTE, plus discharge keys when present). Intel writes BCLM.

### Feature Area

Energy and display

### Mockups / References

AlDente's charge limiter and Calibration Mode: https://apphousekitchen.com/aldente-overview/features/

BenjaminD2023:
## Summary

- add an opt-in Charge Limit feature that stops charging at a chosen ceiling (20–100%, default 80%)
- allow discharging down to that ceiling while the Mac stays plugged in
- add a calibration cycle: 100% → 10% → 100%, hold one hour, then return to the saved limit
- give the menu bar panel and Settings → Energy an AlDente-style circular limit dial
- ship a protected SMC helper, the same Login Items pattern Fan Control already uses

Closes #849
Closes #846

## Behavior

Charge Limit is off by default in Features. After it is installed and the helper is allowed in Login Items, the Mac still runs from the adapter; charging itself stops at the cap.

On Apple Silicon the helper writes the charging-control keys the Mac actually exposes (`CH0B`/`CH0C` or `CHTE`, plus discharge keys when present). On Intel it writes `BCLM`. The limit is enforced while Vorssaint is running. Force-discharge (calibration or discharge-to-limit) restores itself if the helper loses the app heartbeat.

## Verification

- `./build.sh --test` — `TESTS OK (8271 checks)`
- charge-control helper `--selftest` — `charge-control-helper: ok`
- rebased onto current `origin/main`

Live charging control still needs a battery Mac and the Login Items grant; that path was not exercised on this machine.


vorssaint:
Thanks for the pull request. I tested this locally, but it does not work on an M4 MacBook Air. The app reported that charge control was unavailable on this Mac.

BenjaminD2023:
The "unavailable on this Mac" message was a false negative, not a missing M4 key.

I dumped SMC on a MacBook Air M4 (Mac16,12, macOS 26.4.1). `CHTE` (ui32) and `CHIE` are present, and the probe succeeds. What was going wrong:

1. Before Login Items approval, the **unprivileged** app probed SMC itself. On some firmware those charging keys are hidden from user processes, so it reported unsupported and never let the privileged helper be the source of truth.
2. The probe required exact 1-byte `CH0B`/`CH0C` or 4-byte `CHTE`. Later chips still use those names with mixed sizes; writes are now padded/trimmed to the key's reported size. `CH0C` inhibit is `0x01`, matching later Apple Silicon firmware.

Pushed in b9bf208. After rebuilding, Charge Limit should show **Allow charging control** instead of unavailable; once the helper is approved it talks to `CHTE` on M4 Air.

PathGao:
Three things in `ChargeControlHardware.swift`, ranked.

**1. The readback is thrown away on the last attempt.**

```swift
try client.writeBytes(payload, to: key)
if client.readBytes(key) == payload { return true }
if attempt == attempts - 1 { return true }   // readback said no — still returns true
```

After three writes that all read back wrong, `write` reports success. `apply()` then says the gate is set and the UI shows the limit enforced while charging continues. Worse, `restoreNormal()` goes through the same path: a failed restore reports success, so `performRestore` clears the marker, releases the lock and lets the helper exit with charging still inhibited — and the marker is gone, so `recoverAbandonedControlIfNeeded` won't pick it up next launch. Delete that line; the `return false` after the loop already does the right thing.

**2. `writeAll` succeeds if any one key wrote.**

```swift
if write(pair.key, bytes: pair.bytes) { anySucceeded = true }
...
return anySucceeded
```

The CH0 path writes both CH0B and CH0C. CH0B clears, CH0C stays at 0x01, `restoreNormal()` returns true with charging still inhibited. Every caller reads this as "all of it landed". If the tolerance is deliberate because one of the pair is absent on some firmware, that belongs in key discovery, not in the write result.

**3. `.inhibitCharging` has no owner; only `.forceDischarge` does.**

`connectionClosed` restores only when `gate == .forceDischarge`. `watchdogTick` checks heartbeat age only for `.forceDischarge`. `startWatchdog` is only called for `.forceDischarge`. And `scheduleExitIfIdle` refuses to exit while `gate != .allowCharging`. So if the app crashes or gets force-quit while the limit is holding, the helper sits there with charging inhibited, no timer, no heartbeat check, and it never restarts — so the marker-based recovery never runs either.

`safetyCaption` ships in 13 languages saying charging returns to normal "if the connection stops". The PR body scopes the heartbeat restore to force-discharge. One of the two has to move. Cheapest is running the existing watchdog for `.inhibitCharging` too — widening the gate check is the whole change.

**Smaller:** the default limit is written twice — `DefaultsKey.chargeLimitPercent: 80` in `Defaults.swift`, and `ChargeControlPolicy.defaultLimit`, which the helper selftest asserts is 80. Point the registration at the policy constant.

**Not verified:** the M4 Air fix in b9bf208. `refreshLocalProbe()` falls past the diff truncation, so I can't see what the unprivileged path reports before the helper is approved. The `paddedSMCBytes` change and the CHTE branch read correctly, but I have no battery Mac and ran nothing.

**PR body:** `Closes #846` will auto-close that issue on merge, before any reporter has confirmed charge control works on their Mac. Make it a plain reference.

**Impact:** on a Mac where an SMC write is rejected, the app reports the limit as enforced and charging never stops; a rejected restore leaves charging inhibited after the recovery marker has already been deleted.
**Merge:** needs the `write()` and `writeAll()` return values fixed first, and a decision on who owns `.inhibitCharging` when the app dies.

ref #846

PathGao:
Closing this. The maintainer tested it and the app reported charge control unavailable on his M4 MacBook Air (https://github.com/vorssaintapp/vorssaint-utils/pull/851#issuecomment-5389407336), and the thread has been quiet since.

One thing to carry into a future attempt, separate from the charge-control issue: the diff changes `legacyBlockAttachmentNudge` in `Sources/Vorssaint/App/MenuBarRenderer.swift:873` from `-0.25` to `-1.2`. That constant is shared by five menu bar attachment sites, so retuning it for the battery icon also moves the CPU, GPU, memory and disk blocks on macOS 26 and earlier. Put a battery retune in the battery base offset instead, and state which macOS version the new number was measured on.

BenjaminD2023:
Snapzy_2026-09-02_20-54-31_874
@PathGao I'm still working on this branch, commits has been made just few days ago, and I've fixed the charge limit problem and should be able to work now, it works perfectly on my device, however, the menu bar displays still have a lot of bugs. I'm trying to solve the issues we discussed in #846. It would be nice if you can reopen the branch. I would leave a comment when it works perfectly on my device and is ready to merge this time

PathGao:
Pls rework it good enough and it could support macos 27 （kind of weird since macos 26/27 have this feature），then create as a new pr。

BenjaminD2023:
@vorssaint @PathGao The new PR #1308 should be ready for review, I've used in on MacOS 27 for a week now and it works fine on my device.

PR #862, opened 2026-08-23
feat(dock-preview): add optional app quit action to Dock Preview

arefshal:
## Summary

- Add an opt-in Dock Preview setting that makes the thumbnail × button quit the owning app.
- Apply the behavior to both regular and pinned preview panels.
- Update the tooltip, accessibility label, context menu, settings backup, changelog, and every supported language.
- Preserve the existing close-window behavior by default.

## Before and after

Before, the × button always closed only the selected window.

After, users can enable **Settings → Switcher → Dock Preview → Quit the app with the × button**. When enabled, × sends the app a normal termination request, allowing apps to present save-confirmation dialogs when needed.

## Testing

- `./build.sh` — succeeded with no warnings
- `./build.sh --test` — `TESTS OK (8231 checks)`
- `/Applications/Vorssaint (Developer).app/Contents/MacOS/VorssaintDeveloper --selftest` — `SELFTEST OK`
- Manually verified the new setting in the Developer build


PathGao:
This went `CONFLICTING` when #830 (Dock Preview card and drag rework) landed. Two files, and both are add-alongside:

- `UI/Switcher/DockPreviewPanelView.swift` — #830 added `orientation` and `onTogglePinned` to the same parameter list and struct where this branch adds `closeActionTitle`. Two hunks, both "keep both fields".
- `CHANGELOG.md` — same unreleased block.

The 13 locale files, `DockPreviewService.swift` and `DockPreviewSupport.swift` all auto-merge clean.

Could you rebase onto current `main`? Also worth knowing: #380 adds `mediaPlayer` to the same two structs, so whichever of the two lands first, the other will need a second pass.

PathGao:
Coverage checks out: both Dock Preview close paths are patched (`DockPreviewService.swift:166` and `:1288`), and the hover panel and the pinned panel both render through `DockPreviewPanelContent`, so `closeActionTitle` reaches the label, the tooltip and the accessibility label. No third close path exists. Three notes, none of them blocking.

- **The quit block is copied into both `close(_:)` implementations** rather than living in the shared helper, and the two copies already differ (`endSession()` vs `closePreviewPanel()`). A third close path would silently not get it.
- **`terminate()` returning `true` only means the Quit event was delivered.** If the app raises an unsaved-changes sheet and the user cancels, the panel is already torn down and the window is still there. A `false` return does nothing at all — no fallback to closing the window, no feedback.
- **`SwitcherView.swift:818-819` and `:1050-1051`** reuse the same `l10n.s.dockPreviewCloseWindow` string for the App Switcher's × buttons and are not covered. That reads as deliberate — the toggle sits under the Dock Preview section gated on `dockPreviewEnabled` — but the caption ("× quits the whole app instead of closing only that window") never says Dock Preview, while the App Switcher × keeps saying "Close window". Worth naming Dock Preview in the caption.
- The `SettingsBackupSupport.exportKeys()` assertion added in `Tests/MetricsTests.swift` is tautological: `exportKeys()` is `Set(Defaults.registeredDefaults.keys)` plus extras, so any registered key passes it.

Still needs the rebase noted above — `main` moved `DockPreviewPanelView.swift` under it.

arefshal:
Addressed the non-blocking review notes in `d230e62`:

- centralized the quit/fallback decision in `DockPreviewSupport.performCloseAction` for both regular and pinned panels
- keep the preview alive after an accepted Quit request, so cancelling an unsaved-changes sheet does not tear it down; a refused Quit request falls back to closing the window
- explicitly name Dock Preview in the setting caption across every supported language
- replaced the tautological backup assertion with behavioral coverage for normal close, accepted Quit, and refused-Quit fallback

Verification: `./build.sh --test` passes with 8557 checks, and the full release build completes without warnings.

PathGao:
Rebase landed, `MERGEABLE` again, and all three notes are addressed. The counter math in the new test checks out: `false` → close only (requestQuit short-circuits, 0/1), `true` + accepted → quit only (1/1), `true` + refused → quit then fallback close (2/2). Caption names Dock Preview in all 13 locales.

One thing I couldn't verify. After an accepted quit, `performCloseAction` returns and nothing else runs — no `finishClosing`, no `selectedWindowID` reset. That's the right fix for the cancelled-sheet case, but it also means a card for a window that *did* go away stays on the panel unless something outside this diff notices the app terminated. `finishClosing` is the only teardown path visible here and the quit branch skips it. If there's already a `didTerminateApplicationNotification` observer or a poll refreshing the window list, this is nothing; I can't see that from the diff. If there isn't, the second × click on an app that already quit is silent — `NSRunningApplication(processIdentifier:)` returns nil, `requestQuit` returns false, and the fallback closes a window that no longer exists.

#380 still adds `mediaPlayer` to `DockPreviewPanelContent` and `DockPreviewCard`, so whichever of the two lands second needs the same two-field pass.

**Impact:** none on the default path — `dockPreviewQuitAppOnClose` is off for clean installs and every existing close behavior is unchanged. With it on, the worst case above is a stale preview card, not a wrong action.
**Merge:** ship it.

Issue #401, opened 2026-07-31
[Bug]: Mouse button shortcuts cannot trigger global hotkeys — including Vorssaint's own window-management shortcuts


hash00:
### Description

**Summary:** Mouse button shortcuts cannot trigger any global hotkey, including Vorssaint's own window-management shortcuts. They only work for shortcuts owned by the frontmost application.

**Environment**
- Vorssaint 3.2.0
- macOS 27 beta, Apple Silicon (Mac17,3)
- Logitech MX Anywhere 2 over Bluetooth LE
- Input Monitoring and Accessibility both granted
**Scope**

Works from a mouse button (shortcuts owned by the focused app):
- ⌘N — new Finder window
- ⇧⌘N — new folder
- ⌘→ — end of line in a text field

Fails from a mouse button (global hotkeys):
- ⌃⌘→ — Vorssaint's own "Next display"
- ⌃⌥⌘→ — Rectangle "move to next display"
- ⌥⌘→ — Rectangle "right half"

All three failing shortcuts fire correctly when pressed on the keyboard, so the shortcuts themselves are valid and unclaimed. The dividing line is not the modifier set — it is app-level shortcut vs. system-wide hotkey.

**Suspected cause**

Synthetic key events appear to be posted directly to the focused application (kCGAnnotatedSessionEventTap or equivalent) rather than at HID level (kCGHIDEventTap). Events injected at session level bypass global hotkey monitors entirely, which matches the observed behaviour exactly — including the fact that Vorssaint cannot trigger its own window manager.

For comparison, LinearMouse maps the same buttons to the same shortcuts on the same machine and mouse, and successfully triggers all of the above, which suggests it injects lower in the stack.

**Confirmed not the cause**
- Permissions — Input Monitoring and Accessibility both granted; scroll inversion works, so the event tap is live
- Button detection — macOS decodes the side buttons as standard buttons 3/4 (Safari back/forward works natively with no third-party software)
- Conflicting software — LinearMouse quit and not running during testing; "Apps to leave alone" empty
- Stored config — verified in com.vorssaint.utils: mouseButtonShortcuts = { 3 = "control+command:124"; 4 = "control+option+command:124"; }
- Modifier handling — ⇧⌘N works, so multi-modifier composition is fine

**Secondary issue**

[...]

Issue #378, opened 2026-07-28
能否设置一个键盘按键映射


lintex:
### Problem or Motivation

如：把右Option键映射为Ctrl键；
Option+E 设为打开访达

### Proposed Solution

😄

### Feature Area

Other

PathGao:
Triage note, part of a batch pass over the open issues. The labels are a contributor's read, not a maintainer ruling.

The Proposed Solution field contains only an emoji, so everything we have is the two lines under Problem or Motivation.

Two different features are bundled there: remapping a key (right Option → Control) and binding a shortcut to open Finder. They land in unrelated places — remapping is a HID / event-tap concern, while "Option+E opens Finder" is an app-launch shortcut, much closer to what the existing shortcut system already does. One of them is small and one is not, so they cannot share an issue.

Could you split this into two, and for the remapping half say which keys you want remappable and whether it should apply globally or per-app? Without that there is nothing to size — it is not a missing detail, it is the entire specification.

（可以用中文补充，没问题。）

vorssaint:
both examples can be configured in 3.3.5. install super key and command bar from settings > features. in super key, choose right option as the key to hold, leave only control selected and set 'a tap on its own' to 'nothing'. keep right option at its default action in macos modifier keys.

for opening finder, go to settings > keyboard shortcuts > app shortcuts, find finder and record option+e. that shortcut works with the command bar closed. closing this as covered by the existing settings

PR #871, opened 2026-08-24
feat(key-overrides): add opt-in Key Overrides for the F-row special keys

tanujdargan:
macOS answers the special F-row keys before apps see them. F5 starts dictation. F4 opens Spotlight. F6 sets a Focus. These keys never arrive as key events.

Key Overrides remaps each chosen key at the keyboard level (hidutil UserKeyMapping) so it sends its plain function key. A Carbon hotkey then binds the plain key to an action. The mapping uses a write-ahead marker with readback and crash recovery. External entries and the Super key entry stay unchanged. A conflict refuses activation.

One override maps one key to one action:
- Act as the plain key: apps receive the plain F3-F6.
- Mute the microphone: the system-wide mic mute. Choosing it installs the Microphone Mute feature, whichever path the override list arrives by, settings backup included.
- Press a shortcut: the key presses a recorded combination. Only this action asks for the Accessibility permission.

F13-F20 can be bound without a remap. F18 is excluded because the Super key uses it. A one-click common set covers the dictation, Spotlight and Focus keys.

The feature ships uninstalled. The override list travels in settings backups. The machine-state marker does not. Override keys go quiet while a shortcut field records. The mapping is removed on disable, uninstall and quit.


PathGao:
The registration is complete — all five feature lists plus `FeaturePresets`, `GlobalShortcut.featuresToSilenceWhileRecording`, `SettingsBackupSupport.unregisteredPreferenceKeys`, `SettingsDirectory`, README and CHANGELOG, and 13 locales in `KeyOverrideStrings`. Going through `hidutil` and Carbon hotkeys instead of adding a 26th `CGEvent` tap is the right call, and the `.cghidEventTap` posting is correctly justified against #401.

**One exit path is missing, and it is the one that cannot be undone.** `AppDelegate.swift:232` gained `KeyOverrideService.shared.suspend()`. `SelfUninstall.swift:83` did not — that is the same list for the app removing itself, and the comment right above `SuperKeyService.shared.suspend()` there says why it exists: it "takes the Caps Lock mapping back out, synchronously, so the key is never left remapped behind a tap that is about to die". Uninstall Vorssaint with an override active and the F-row stays remapped by `hidutil` with no app left to undo it.

**Two writers on one system property, no shared lock.** `KeyOverrideService.mappingQueue` ("keyoverrides-mapping") and `SuperKeyService.mappingQueue` ("superkey-mapping") each do `hidutil --get` → compose → `--set` → `--get` against the same global `UserKeyMapping`. `FeatureRuntime` runs both bindings at launch, so a lost update is possible: both read table T, both write T plus their own entry, the last writer drops the other's. The loser's readback then fails and it publishes `mappingFailed = true` or leaves the super key inactive. Two mechanisms answering "what is in the mapping table" with nothing making them agree — one shared serial queue for the property, rather than one per feature, is the smaller fix.

Smaller:

- `KeyOverrideSupport.consistentMappings` and `isOwnedMapping` are a second copy of `SuperKeySupport.consistentMappings` / `isOwnedMapping`, differing only in the ownership predicate. `SuperKeySupport.consistentMappings` already takes a `property:` parameter, so it could take the predicate too.
- `hotkeyIDs` / `nextHotkeyID` are never pruned when an override is deleted, so ids climb for the process lifetime. Nothing else allocates above 300, so this is unbounded rather than broken.
- `suspend()` does `mappingQueue.sync` on the main thread at quit, running up to three `hidutil` processes. SuperKey already does the same at `SuperKeyService.swift:322`, so quit now serializes two such blocks — worth measuring the added termination latency.
- `SettingsDirectory` keywords cover `keyDictation`, `keySpotlight`, `keyFocus`, `actionMicMute`, `F4`, `F5`, `F6` — Mission Control, Launchpad, F3 and F13–F20 are not searchable.

No scope-claim contradictions found: the README line matches `KeyOverrideKey.allCases`, and the permissions comment matches `needsAccessibility`.

tanujdargan:
@PathGao issue addressed — the uninstall path now takes the F-row remap out, both UserKeyMapping writers share one serial queue, the consistency check is deduplicated behind an ownership predicate, hotkey ids are recycled, and the search keywords now cover Mission Control, Launchpad, F3 and F13. Also merged upstream main and resolved the changelog conflict.

PathGao:
Both points from the last round are closed in 50bce713. `SelfUninstall.swift:85` now suspends the overrides on the same list as the Super key, and the two writers share one queue — `SuperKeySupport.swift:65`, taken at `SuperKeyService.swift:61` and `KeyOverrideService.swift:24` — so their read-compose-write cycles can no longer drop each other's entries.

One site of that same two-writers shape the shared queue does not reach: the cross-keyboard consistency check.

`SuperKeyService.swift:355` sets aside only `SuperKeySupport.isOwnedMapping`, and `KeyOverrideSupport.swift:236` sets aside only `KeyOverrideSupport.isOwnedMapping`. Each writer therefore treats the other's entries as external, and the shared core at `SuperKeySupport.swift:133` compares the per-keyboard tables with them still in.

Take the case the wake observer exists for — `SuperKeyService.swift:274` names it, "plugging in another one". With both features on, `hidutil --matching keyboard` reports the already-mapped keyboards and the newly attached one with an empty table:

- The Super key repair sets aside caps→F18 and compares `[F-row entries]` against `[]`. `mappingsMatch` fails, `SuperKeySupport.swift:141` returns nil, `performMapping` returns false at `:359`, and `mappingFailed` is published.
- The overrides' wake path, `KeyOverrideService.swift:231` → `:201`, fails the same way at `:204`.

Neither repairs, and it does not heal: every later wake runs the same comparison against the same divergent tables. The keyboard that arrived keeps an unmapped Caps Lock and an unmapped F-row until one of the two features is switched off. With one writer its own set-aside was enough, which is why this is new here.

What to change: separate "ignore for the cross-keyboard comparison" from "drop from the table we write back". In the shared core at `SuperKeySupport.swift:133`, filter entries owned by either feature before `mappingsMatch`, then add the other feature's owned entries — their union across the tables — back into the returned table, so the write puts them onto the keyboard that arrived without them. Re-adding is what makes this safe without a marker gate: the entry is never dropped, only excluded from the equality test, and both ownership predicates match a fixed shape rather than an arbitrary mapping, so the rule at `SuperKeySupport.swift:120` about not copying someone else's device-specific mapping onto every keyboard still holds.

I could not run this. The divergence needs a second keyboard attached after the write, so read the reachability from `mappingTables` and the two wake paths rather than as something observed.

tanujdargan:
@PathGao good catch — fixed in a049505. The consistency core now takes the ownership predicate and a partner predicate: both are ignored for the cross-keyboard comparison, and the partner's entries ride along in the written table as the union across keyboards, so either feature's repair carries the other's mapping onto a new keyboard. Ownership/removal rules unchanged; covered by new two-feature convergence tests.

PathGao:
The two-writer repair is closed. `SuperKeySupport.swift:141` takes both predicates, `:148` keeps partner entries out of the comparison, `:152-158` re-adds them as the union across keyboards, and both writers pass the other's predicate — `SuperKeyService.swift:361` and `KeyOverrideSupport.swift:237`. Source read, not run.

**One hole my last comment opened.** `propagating:` is unconditional at both call sites while `settingAside:` is gated on the writer's own marker. An entry of the partner's shape is therefore set aside and copied onto every keyboard whether or not the partner ever wrote it. That drops the rule at `SuperKeySupport.swift:102-114`: a caps→F18 entry is owned only when `superKeyMappingApplied` says a confirmed write made it, and without that the Super key refuses activation rather than claim external state. Through the partner path the overrides now pick that same entry up — `KeyOverrideSupport.hasMappingConflict` at `:224` only inspects the sources it reclaims — and write it across every keyboard with no marker anywhere. The fix is the size of what is already there: read the partner's `...MappingApplied` marker at the two call sites and pass `{ _ in false }` when it is not set. My "fixed shape" argument covered `:120`, not `:102`.

**Three `hidutil` processes on every shortcut recording.** `ShortcutCapture.end()` re-syncs `featuresToSilenceWhileRecording` (`ShortcutCapture.swift:43`), which now includes the overrides (`GlobalShortcut.swift:786`), and `KeyOverrideService.syncWithPreferences` always reaches `applyMapping` at `:61` — `performMapping` has no already-correct exit for a non-empty key list, so get, set and get all run. `end()` fires on every capture, Escape, focus loss, window close and app deactivation; every other feature on that list only re-registers keys. `mappingReportConfirms(report.output, expected: wanted)` against the first get would make the no-op case one process. Nothing new sits inside the tap callback, which is worth stating since the shared queue could have put it there: `handle` at `SuperKeyService.swift:436` is untouched, and its one route into that queue is `.interceptAndRemap` → `repairMappingIfStale` at `:487` → `mappingQueue.async` at `:326`, an enqueue, so the callback never waits on the overrides' `hidutil` work.

Smaller, still open: the directory keywords stop at F13 (`SettingsDirectory.swift:169`), so F14–F17, F19 and F20 remain unsearchable. And the quit measurement from two rounds ago — `AppDelegate.swift:232-234` now runs two `mappingQueue.sync` blocks against one queue, up to six `hidutil` spawns on the main thread before the process exits, same shape at `SelfUninstall.swift:83-85`; a number would settle it. The branch is five commits behind main and needs another merge; no conflicts are reported and nothing on main added a feature, so the 55-feature assertion still holds.

PathGao:
Two things changed since the last round, neither of them in the diff.

The branch now reports a conflict. Last round it reported none, so "five commits behind, no conflicts" is out of date — that merge comes before anything else.

Neither open point is in the current tree: `propagating:` is still unconditional at both call sites (`KeyOverrideSupport.swift:237`, `SuperKeyService.swift:361`) while `settingAside:` stays gated on the marker, and `performMapping` still has no already-correct exit for a non-empty key list. Nothing to add to either.

**New: choosing the mic-mute action installs a different feature.** `KeyOverrideSettings.installMicMuteIfNeeded()` calls `features.setAvailable(.micMute, true)` from `add(_:)`, `actionKindBinding` and the common-set button. The body lists the three actions and says of this one only "the system-wide mic mute" — not that picking it turns Microphone Mute on in the hub. It also splits the behaviour by entry path: `load()` does not install, so the same override list arriving through a settings backup leaves `micMute` off and the row falls back to `micMuteNeedsFeature`. Same list, two hub states depending on how it got there. Either declare the install and do it on the import path too, or drop it and let the note carry it.

I read the diff, not the tree, and it is truncated after the first `SettingsDirectory` entry — the keyword list and whatever renders `KeyOverrideSettings` for `SettingsPage.keyOverrides` are past the cut, so I checked neither this round.

**Impact:** none new beyond the two open points; the auto-install changes hub state rather than breaking anything.
**Merge:** needs the conflict resolved and the two open points closed first.

ref #378

tanujdargan:
@PathGao all three points are closed in c504c66, merged with upstream main (conflict resolved; the Super key's failure-reason enum and the partner propagation now coexist).

- The propagation hole: both writers now gate `propagating:` on the partner's `...MappingApplied` marker — `SuperKeyService.swift:429-436` reads `keyOverridesMappingApplied`, `KeyOverrideService.swift:212-220` reads `superKeyMappingApplied` and passes it into `KeyOverrideSupport.consistentMappings` as `partnerMappingApplied:`. An unmarked entry of the partner's shape is external again: left out of nothing, copied nowhere, and a divergence over it refuses instead of spreading. Covered by a new test: the both-features report with `partnerMappingApplied: false` returns nil.
- The recording-path cost: `KeyOverrideService.performMapping` now has the already-correct exit for a non-empty key list — `mappingReportConfirms(report.output, expected: wanted)` against the first get, after the write-ahead marker is set, so a no-op sync is one hidutil process and a confirmed-correct table still leaves the marker true.
- The mic-mute install is declared (PR body and changelog) and moved into `KeyOverrideService.syncWithPreferences`: any enabled mic-mute override installs the Microphone Mute feature from the service, so a list arriving through a settings backup lands in the same hub state as one built in Settings. The view's three separate install calls are gone; the `micMuteNeedsFeature` note stays as the fallback.

Smaller: the directory keywords now run F13–F17, F19 and F20 (`SettingsDirectory.swift:169`). On the quit-latency question: the two `mappingQueue.sync` blocks share one serial queue, so quit is bounded by six hidutil spawns in the worst case; I haven't put a number on it.

Tests: 8785 checks pass, dev build clean.

PathGao:
Both open points are closed in the tree I can see, and the branch reports mergeable again.

- The propagation hole: `KeyOverrideService.performMapping` reads `superKeyMappingApplied` and passes it as `partnerMappingApplied:`; `KeyOverrideSupport.consistentMappings` turns it into `propagating:` or `{ _ in false }`. `SuperKeyService` does the mirror with `keyOverridesMappingApplied`. An unmarked entry of the partner's shape is external again.
- The recording-path cost: the already-correct exit is there, marker set first, `mappingReportConfirms(report.output, expected: wanted)` returning before the write. A no-op sync is one `hidutil`.

**The mic-mute install moved from a user action to a recurring one, and that makes the hub's uninstall button for Microphone Mute a no-op.**

`syncWithPreferences` is not "the list arrived" — it is every binding run. On main, `FeatureRuntime.setAvailable` has exactly one caller, `FeatureHubSettings.swift:363`, the hub toggle. `KeyOverrideService.swift:52-55` adds a second one, and `ShortcutCapture.end()` reaches it: `ShortcutCapture.swift:43` syncs `featuresToSilenceWhileRecording`, which this PR puts `.keyOverrides` on. Wake, launch and every edit on the page reach it too.

So: keep an enabled mic-mute override, uninstall Microphone Mute in the hub. `setAvailable(.micMute, false)` runs `bindings[.micMute]`, not the key overrides' binding, so the uninstall holds — until the next shortcut recording ends or the Mac wakes, and then it is back. The first-run picker and `apply(_ preset:)` are the same story on a delay: `replaceAvailable` walks `AppFeature.allCases` and `keyOverrides` sits before `micMute` there (`FeatureCatalog.swift:19-25`), so a preset that excludes Microphone Mute wins its own pass and loses at the next wake.

The gap the last round named was the import path. That is one arrival event, not a standing invariant. Installing from the two write paths — the Settings edits that already call `save()`, and the backup restore — covers the same case without giving the sync path a veto over the hub. The comment at `KeyOverrideService.swift:49-51` states the invariant the code implements ("keeps the hub state a function of the list"); the body states an arrival ("Choosing it installs the Microphone Mute feature, whichever path the override list arrives by"). Those are two different claims, and the one that shipped is the one that overrides the user.

The diff is truncated inside `KeyOverrideSettings.swift` this round, so `SettingsDirectory.swift` and the new tests are past the cut — the F13–F17/F19/F20 keyword claim and the `partnerMappingApplied: false` test I did not check. I ran none of this; the reachability above is read from `FeatureRuntime.swift`, `ShortcutCapture.swift` and `FeatureHubSettings.swift` on main.

**Impact:** uninstalling Microphone Mute from the hub, or leaving it out of a first-run selection or a preset, does not survive the next wake or the next shortcut recording while an enabled mic-mute override exists.
**Merge:** needs the mic-mute install moved off `syncWithPreferences` onto the write paths first.

ref #378

PathGao:
Closing this. Where it stands:

The last review (2026-08-27) asked for the Microphone Mute install to move off `KeyOverrideService.syncWithPreferences` onto the write paths: as written, `setAvailable(.micMute, true)` runs on every sync, so the hub's Microphone Mute uninstall is undone at the next wake or shortcut recording. That is still in the diff, there has been no push or reply since 2026-08-26, and the branch is now 136 commits behind main and conflicting.

Direction: #378 is a question with no maintainer answer yet, and this adds a service, a settings page, a backup key and strings in every locale for it. CONTRIBUTING asks for that call to be settled before the code exists.

For a resubmission: open it from a branch, not your fork's `main` (every sync lands in the PR otherwise); leave `CHANGELOG.md` alone (rule 8); and the description needs the Mac model, macOS build and what was not tested.

If #378 gets a direction, start a fresh branch from origin/main. The review on this thread already covers the rest.

Issue #873, opened 2026-08-24
Clipboard history records Chromium image copies twice (promised-flavor re-declare)


Kyzcreig:
### What happens

Copying an image in a Chromium browser (Brave, Chrome, Edge) records **two** clipboard history entries for one copy, ~1–3 seconds apart. Both show the same picture; their bytes differ.

### Root cause (instrumented)

I watched `NSPasteboard.general.changeCount` with an external monitor while driving the same right-click → Copy Image repeatedly:

- Vorssaint **quit**: one changeCount bump per copy.
- Vorssaint **running**: two bumps (e.g. `24→25→26`, seconds apart) — and **both** carry Chromium's private types (`org.chromium.internal.source-rfh-token`, `org.chromium.source-url`), i.e. Chromium itself executes both writes.
- Vorssaint running but `clipboardHistoryEnabled=false`: one bump again.

Mechanism: Chromium posts image copies with **promised (lazy) flavors**. The history capture's immediate read resolves the promise, and Chromium re-declares the pasteboard with its own PNG render — which the capture then records as a second copy. The two entries' PNG chunks tell the story: entry A is the capture's own encode of the TIFF (Apple encoder, page's Display-P3 `iCCP/iTXt/iDOT`), entry B is Chromium's render (`sRGB/eXIf`). Different encoders → different bytes → the SHA-256 dedupe in `promoteImage` can't match them.

Small/plain-sRGB images often don't reproduce it (Chromium renders those flavors eagerly — no promise, no re-declare), which makes the bug look intermittent. Real-site P3 images reproduce it near-100%.

### Repro

1. Enable Clipboard history.
2. In Brave/Chrome, right-click any image on a real site (wide-gamut/P3 images are most reliable) → Copy Image.
3. Open the history: two entries for the one copy, 1–3s apart.

### Fix proposal

Treat a same-source re-declare within a few seconds as the same copy: remember a small colorspace-normalized thumbnail of the last image capture, and when a new image with the same pixel dimensions arrives inside the window and its thumbnail matches (mean channel delta — measured ~0.5 for a real color-managed re-encode vs ~8.5 for genuinely different colors), refresh the existing entry instead of inserting a twin. PR incoming.

PathGao:
[...]

PR #874, opened 2026-08-24
fix(clipboard): fold Chromium promised-flavor re-declares into the original clipboard image entry

Kyzcreig:
Refs #873.

Chromium browsers post image copies with promised (lazy) flavors. The history capture's immediate read resolves the promise, and Chromium re-declares the pasteboard with its own PNG render seconds after the copy — re-encoded to sRGB, so the byte hash never matches the first capture and `promoteImage` records the same picture twice (full instrumentation in the issue).

### What this does

- New `ClipboardImageRedeclare` helper: keeps a 16×16 sRGB-normalized thumbnail signature of the most recent image capture.
- In `promoteImage`, when the byte-hash dedupe misses, a new image with the **same pixel size**, arriving **within 3 seconds**, whose **normalized thumbnail matches** (mean absolute channel delta ≤ 3) is treated as the source app re-posting the same picture: the existing entry is refreshed (same file, same id) instead of a twin being inserted.
- Genuine new copies are untouched: different pixels, different dimensions, no prior capture, or anything after the window all fall through to the existing insert path.

### Why thumbnail comparison

The re-declared render is a color-managed conversion, so bytes and even exact pixel values differ. Measured on a real Brave copy pair (P3 original vs Chromium sRGB re-render): mean channel delta **~0.5** after normalizing both to sRGB. Measured on the counter-case (same bytes merely re-tagged with a different profile — genuinely different colors): **~8.5**. The tolerance of 3 sits between them with margin on both sides.

### Tests

Added to `MetricsTests` (runs in `./build.sh --test`, 8302 checks green):
- re-declare match inside the window / different pixels / different size / no prior / stale window / empty thumbnails
- a color-managed P3→sRGB re-encode of one picture compares equal; same bytes re-tagged P3 vs sRGB compare different

Verified live on a MacBook Pro (macOS 26): with this build, driven Brave image copies that previously produced two entries produce one.

### DIFF
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
@@ -43,6 +43,10 @@ final class ClipboardHistoryService: ObservableObject {
 
     private var timer: Timer?
     private var lastChangeCount = 0
+    /// Signature of the most recent image capture, kept briefly so a source
+    /// app re-declaring the pasteboard while resolving a promised flavor is
+    /// folded into the entry it already produced (see ClipboardImageRedeclare).
+    private var lastImageSignature: ClipboardImageRedeclare.Signature?
     /// The poll reads the pasteboard off the main thread: while a password
     /// prompt is up the pasteboard server can take seconds to answer, and a
     /// blocked main thread stalls every event tap with it, so typing freezes
@@ -637,6 +641,16 @@ final class ClipboardHistoryService: ObservableObject {
 
     private func promoteImage(_ image: (data: Data, width: Int, height: Int)) {
         let hash = Self.sha256Hex(image.data)
+        let thumbnail = ClipboardImageRedeclare.thumbnail(of: image.data)
+        defer {
+            lastImageSignature = thumbnail.map {
+                ClipboardImageRedeclare.Signature(width: image.width,
+                                                  height: image.height,
+                                                  thumbnail: $0,
+                                                  capturedAt: Date(),
+                                                  entryID: entries.first { $0.kind == .image }?.id ?? UUID())
+            }
+        }
         if let existing = entries.first(where: { $0.kind == .image && $0.imageHash == hash }) {
             entries.removeAll { $0.id == existing.id }
             insertPromoted(ClipboardHistoryEntry(id: existing.id,
@@ -646,6 +660,27 @@ final class ClipboardHistoryService: ObservableObject {
                                                  kind: .image,
                                                  imageFile: existing.imageFile,
                                                  imageHash: hash,
+                                                 imageWidth: existing.imageWidth,
+                                                 imageHeight: existing.imageHeight))
+        } else if let thumbnail,
+                  let redeclaredID = ClipboardImageRedeclare.redeclaredEntryID(
+                      previous: lastImageSignature,
+                      width: image.width,
+                      height: image.height,
+                      thumbnail: thumbnail),
+                  let existing = entries.first(where: { $0.id == redeclaredID }) {
+            // The source app re-posted the same picture while resolving a
+            // promised flavor (Chromium image copies do this seconds after the
+            // copy, re-encoded to sRGB). Refresh the entry it already has
+            // instead of recording a twin.
+            entries.removeAll { $0.id == existing.id }
+            insertPromoted(ClipboardHistoryEntry(id: existing.id,
+                                                 text: "",
+                                                 copiedAt: Date(),
+                                                 pinnedAt: existing.pinnedAt,
+                                                 kind: .image,
+                                                 imageFile: existing.imageFile,
+                                                 imageHash: existing.imageHash,
                                                  imageWidth: existing.imageWidth,
                                                  imageHeight: existing.imageHeight))
         } else {
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
@@ -641,16 +641,15 @@ final class ClipboardHistoryService: ObservableObject {
 
     private func promoteImage(_ image: (data: Data, width: Int, height: Int)) {
         let hash = Self.sha256Hex(image.data)
-        let thumbnail = ClipboardImageRedeclare.thumbnail(of: image.data)
-        defer {
-            lastImageSignature = thumbnail.map {
-                ClipboardImageRedeclare.Signature(width: image.width,
-                                                  height: image.height,
-                                                  thumbnail: $0,
-                                                  capturedAt: Date(),
-                                                  entryID: entries.first { $0.kind == .image }?.id ?? UUID())
-            }
-        }
+        var thumbnail: [UInt8]?
+        // The id of the entry this capture was recorded into. Threaded from
+        // each insertion site instead of re-searching `entries` afterwards:
+        // the array is pinned-first, so a lookup like
+        // `entries.first { $0.kind == .image }` binds to a PINNED image when
+        // one exists and the next re-declare would fold into (and refresh)
+        // the wrong entry, silently dropping a real copy.
+        var recordedID: UUID?
+
         if let existing = entries.first(where: { $0.kind == .image && $0.imageHash == hash }) {
             entries.removeAll { $0.id == existing.id }
             insertPromoted(ClipboardHistoryEntry(id: existing.id,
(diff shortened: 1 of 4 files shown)


PathGao:
The de-duplication idea is right and `promoteImage` really is the only insertion path, so the coverage is complete. But the signature binds to the wrong entry as soon as any image is pinned, and the result is that a real copy is lost.

`ClipboardHistoryService.swift:651` resolves the entry that was just created by searching:

```swift
entryID: entries.first { $0.kind == .image }?.id ?? UUID()
```

`entries` is ordered pinned-first, and `insertPromoted` puts an unpinned entry at `firstRecentIndex`, which is `entries.firstIndex { !$0.isPinned }` — after every pinned entry. So with one pinned image in the history, `entries.first { $0.kind == .image }` returns **the pinned one**, not the one just promoted. The next capture inside the 3s window then matches that signature, and the branch removes and re-inserts the *pinned* image with `copiedAt: Date()` while the genuinely new copy is never recorded.

`insertPromoted` already knows the id it inserted. Threading it out is the fix; the search is a second answer to a question that already had one.

Two more things on the same block:

- The `defer` also runs on the `guard let name = ClipboardImageStore.store(...) else { return }` failure path, where nothing was inserted at all, leaving a live signature pointing at a stale id.
- `ClipboardImageRedeclare.thumbnail(of:)` runs `CGImageSourceCreateImageAtIndex` plus a `CGContext.draw` at the top of `promoteImage`, which is on the main queue, for **every** image copy — not only for the re-declare case. The existing main-thread cost there was a sha256 over the bytes; a full decode of a 6K screenshot is a different order.

And one judgement call worth stating explicitly: the guards are same pixel dimensions, ≤3s, and mean channel difference ≤3 over 256 samples. Nothing checks that both captures came from the same source app or that the pasteboard actually carried a promised flavor. Two genuinely different copies of the same size and near-identical average colour inside 3s — consecutive crops of a white document, two frames of the same window — get folded, and the second one is dropped. Since the cost of a false positive is silent data loss, that deserves a tighter gate than image similarity alone.

The support file is fully tested; the wiring in `promoteImage`, which is where all of the above sits, has no test.

Kyzcreig:
Thank you — the pinned-first binding bug was real and exactly as you traced it. All three points landed:

**1. Wrong-entry binding (the data-loss bug): fixed at the root.** `recordedID` is now threaded out of each insertion arm — the id was already known at every insertion site, so the pinned-first search was, as you put it, a second answer to a question that already had one. With a pinned image in history the old code demonstrably bound every fresh signature to the pinned entry.

**2. Stale signature on store failure: fixed.** The `guard let name = ... else { return }` path now explicitly clears `lastImageSignature` — a signature must not survive pointing at an id this capture never touched. (Restructuring away from `defer` made this case impossible to miss rather than easy to forget.)

**3. Main-queue full decode: fixed.** The data-path thumbnail now uses `CGImageSourceCreateThumbnailAtIndex` with `kCGImageSourceThumbnailMaxPixelSize = thumbnailSide`, so ImageIO subsamples during decode — cost scales with the 16px comparison canvas, not the 6K source. Also moved thumbnail generation off the hash-match arm entirely (byte-identical repeats never need one).

**On the judgement call (same-size different-content folding):** agreed the gate is similarity-only and that the false-positive cost is silent loss. I kept this PR at the binding/lifecycle/perf fixes since tightening the gate changes detection behavior — but the natural next constraint is requiring the pasteboard change that triggers the fold to carry the *same* set of declared types as the original copy (a Chromium re-declare re-posts the same flavor set; a genuinely new copy of a similar-looking image usually doesn't cross that plus size plus 3s plus tolerance simultaneously). Happy to do that as a follow-up if you'd take it.

Suite: 8302 checks green.

PathGao:
`recordedID` threaded from each insertion site is the right shape, and the store-failure clear closes the stale-signature hole. Three things on the new revision.

**The tolerance was measured under a decode path this revision replaced.** The `~0.5` / `~8.5` numbers and the live Brave verification in the body were produced when `thumbnail(of data:)` did a full decode plus a `.high`-interpolation `CGContext.draw`. It now goes through `CGImageSourceCreateThumbnailAtIndex` with `kCGImageSourceThumbnailMaxPixelSize: 16` — a different resampler. ImageIO subsamples during decode, it does not area-average, so the high-frequency difference between the two encoders that the old high-quality downscale washed out survives into the comparison. The match case had ~2.5 of headroom under the threshold and that is exactly what this eats into: the real P3-vs-sRGB pair can now land above 3 and the fold silently stops happening, which is #873 back with a fix that looks shipped.

The suite does not cover this. Every tolerance assertion calls `ClipboardImageRedeclare.thumbnail(of image: CGImage)` directly — `thumbnail(of data: Data)`, the function that actually changed, has no test, so 8302 green says nothing about the new path. Re-measure both numbers through the `Data` overload, or drive the pair through it in a test.

**The window renews on every fold, so the false-positive risk you deferred is not bounded at 3s.** After a fold, `lastImageSignature` is rewritten with the *new* capture's thumbnail and `capturedAt: Date()`. Capture 3 is then compared against capture 2 rather than against the picture the entry actually holds, and its own 3s window starts from capture 2. A run of copies each ≤3s apart and each within tolerance of its predecessor collapses into a single entry however long the run is, with unbounded drift from the first picture — 20 real copies in, 19 gone. Cheapest fix is to not touch `lastImageSignature` on the fold arm at all: the window then measures from the original copy, and every comparison runs against the image that is stored.

**The hash-match arm still decodes**, contrary to "moved thumbnail generation off the hash-match arm entirely". `thumbnail` is only assigned inside the `else`, so the `if thumbnail == nil` recompute below the guard runs on every byte-identical repeat. That same line also re-attempts the decode when the first call already returned nil.

On the follow-up: yes, take it as a separate PR, and declared-type equality is the right constraint. Independent of that, clearing the signature when a non-image capture lands looks close to free — but the text capture path is not in this diff, so I cannot see whether anything else already touches `lastImageSignature`. I also cannot run the Brave check; the resampler point above is read off the diff, not measured.

**Impact:** if the subsampled thumbnail pushes the real re-declare pair over the tolerance, #873 is unfixed while the issue reads as fixed, and no test would show it. The window renewal drops genuine copies in any run of similar same-size images.
**Merge:** needs the tolerance margin re-measured through `thumbnail(of data:)` first.

ref #873

Kyzcreig:
You were right on all three, and the first one is the serious one. I reproduced it rather than reasoning from the diff.

> **The tolerance was measured under a decode path this revision replaced.** [...] ImageIO subsamples during decode, it does not area-average, so the high-frequency difference between the two encoders that the old high-quality downscale washed out survives into the comparison.

Confirmed. I compiled the **actual old and actual new `thumbnail(of data:)` overloads** side by side out of the PR worktree (only the enum name changed, to let both link), generated P3-tagged PNGs, rendered them color-managed into sRGB at full resolution, re-encoded, and pushed the pairs through both. macOS 26.6.1 / Swift 6.3.3, 12 synthetic cases:

```
fixture,size,old_data_delta,new_data_delta,new_accepts_conversion
checker1,64x64,     0.000000,   4.011719,  false
checker7,64x64,     0.000000,   8.386719,  false
checker1,256x256,   0.000000,   4.007812,  false
checker7,256x256,   0.000000,   4.101562,  false
checker1,1024x1024, 0.000000,   4.031250,  false
checker7,1024x1024, 0.000000,   4.091797,  false
gradient,*,         0.000000,   0.27-0.37, true
```

A conversion that scored **0.000000** on the old path scores **4.011719** on the new one, against a tolerance of 3. That is exactly the failure mode you described: #873 back, with a fix that looks shipped. Smooth gradients stay under 3, so this isn't a probe that condemns every conversion — it's specifically the high-frequency content the old `.high`-interpolation downscale used to wash out.

Worse than I realized: the same probe shows the new path **incorrectly accepts 5 retagged countercases** it should reject (e.g. `checker1` 64×64 retag delta went 4.125000 → 0.003906). So the change moves the threshold in both directions, not just up.

> Every tolerance assertion calls `ClipboardImageRedeclare.thumbnail(of image: CGImage)` directly — `thumbnail(of data: Data)`, the function that actually changed, has no test, so 8302 green says nothing about the new path.

Also confirmed, and this is the part I should have caught myself. The only thumbnail calls in the suite are `MetricsTests.swift:254`, `269`, `276` and all three pass a **CGImage**. The color fixture is already 16×16, so it never exercises reduction of a large encoded image at all. My "8302 green" in the previous comment was answering your *earlier* review and does not speak to this path — you're right that it says nothing.

**Point 2 — renewable fold window.** True, and the measurement is uglier than the description: I drove the exact extracted promotion method with 20 distinct valid PNG copies and it collapsed all 20 into one entry over **20.155 seconds**, retaining the first bytes while comparing against the *last* discarded image. The pre-PR control retained all 20.

**Point 3 — redundant decode.** True. A byte-identical PNG repeat still calls the Data decoder once; fault-injecting a nil first thumbnail produces two calls for a newly stored image. Note this is the thumbnail-decode path specifically — it is not evidence the old full-resolution decode API came back, and my earlier hash-arm claim was wrong on exactly this point.

**What I'm not claiming.** I did not replay the original live Brave capture pair, so the specific `~0.5` and `~8.5` numbers in the PR body remain unverified under the new path — I'm treating them as stale, not as refuted. Re-measuring both through the `Data` overload on the real pair is the right bar and I haven't met it yet.

I'd rather not push a patch that moves the threshold to make my synthetic cases pass — that's fitting the constant to the probe. The honest options are (a) go back to a decode path with area-averaging semantics so the old evidence remains valid, or (b) keep ImageIO and re-derive the tolerance from real captures with a `Data`-overload regression test pinning the original pair.

I'm leaning (a), because (b) re-derives a constant from samples I can't fully reproduce. Before I spend a revision on it: do you have a preference, and do you still have the original Brave pair from #873? That capture is the one input I can't synthesize, and it's what would let either option be verified properly rather than argued.

vorssaint:
closing this one.
the fold decides by how two pictures look, and when it is wrong a real copy silently disappears from the history, which is worse than the duplicate it removes.
on the current build a detailed picture that goes through the color re-encode lands above the tolerance, so the second entry still shows up, while the same pixels merely retagged now pass as the same picture.
the signature is rewritten after every fold with a fresh timestamp, so a run of similar same size copies collapses into one entry however long it goes.
every image copy now decodes on the main thread, including byte identical repeats, and the decode runs twice when the first attempt returns nothing.
the constant and the window were tuned on a capture pair that is no longer at hand, and the path that folds entries has no test.
the comments name a third party browser, and the code should describe the behavior instead.
a smaller version could still land, one that folds only when the clipboard itself says it is the same copy, for example if the same source token shows up on both writes inside the window, with no pixel compare, no tolerance, nothing decoded on the main thread and a test driving the real capture path

Issue #881, opened 2026-08-24
App Switcher does not override the system Command-Tab switcher


BenjaminD2023:
### Description

With the App Switcher enabled and Accessibility granted, ⌘Tab still opens the macOS system switcher. Vorssaint is supposed to replace that shortcut, but Dock keeps handling Command-Tab as its own symbolic hotkey. Swallowing Tab in the session event tap is not enough, so both switchers fight or only the system one appears.

Expected: ⌘Tab (and ⌘⇧Tab, plus ⌘` when that is the window shortcut) opens only the Vorssaint switcher.

Actual: the system switcher still appears.

### Steps to Reproduce

1. Enable App Switcher in Vorssaint and grant Accessibility.
2. Leave the shortcuts at the defaults (⌘Tab / ⌘`).
3. Press ⌘Tab.
4. The macOS app switcher appears instead of (or in front of) Vorssaint's.

### Vorssaint Version

3.3.3-beta.2 / current main

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Local build

StormHuracan:
I have a similar problem: after restarting the app, everything works fine for a while, but then when I press Command + Tab, the window doesn't open or become active (even though I'm using Vorssaint for this); only the Vorssaint app itself opens. Also, sometimes it works, but sometimes it doesn't—I still haven't figured out what exactly causes this...

UPDATE: I found out that this only happens when I have the Vorssaint settings open; in that case, switching between tabs doesn't work. But if I switch from the active Vorssaint settings window, it works, whereas switching from another window doesn't.

PR #882, opened 2026-08-24
fix(switcher): take over the system Command-Tab switcher

BenjaminD2023:
Refs #881

## Behavior before

Vorssaint handled its App Switcher shortcuts only inside this process. On affected setups, the matching macOS app or window switcher could still respond to the same key combination.

## Behavior after

Nothing changes by default. A new “Replace macOS ⌘Tab and ⌘`” setting, off by default, lets the user explicitly disable only enabled macOS app/window switcher shortcuts whose current mappings collide with Vorssaint's configured shortcuts.

While takeover is enabled:

- Forward and reverse app switching (symbolic hotkey IDs 1 and 2) and window switching (IDs 27 and 28) are matched from the current WindowServer mappings, including user remaps.
- Only hotkeys that were enabled before Vorssaint changed them are recorded and later restored.
- The App Switcher includes all running apps so windowless apps remain reachable.

The system hotkeys are restored when takeover is disabled, the event tap is unavailable or removed, Accessibility is lost, or the app quits. A write-ahead marker repairs state on the next launch after a crash.

## Implementation

This is the repository's first write to WindowServer's symbolic-hotkey table. The private CGS functions are resolved at runtime with `dlsym`, and the existing `SpaceWindowBridge` Carbon modifier conversion is reused. There is no background watchdog, so a user who re-enables a shortcut in System Settings is not immediately overridden.

## Checks

- `git diff --check` — clean
- `./build.sh --test` — TESTS OK (9494 checks)
- `./build.sh --dev` — developer bundle built and signed without warnings


PathGao:
The header comment on `SwitcherNativeHotkeys` states the premise exactly right — "the enabled state persists after quit, so every teardown path has to restore" — and then covers only the teardown paths that get to run code. A crash or a `SIGKILL` leaves ⌘Tab, ⌘⇧Tab and ⌘\` disabled system-wide with nothing left to put them back.

**This repo already solved that.** `SuperKeyService` writes a `DefaultsKey.superKeyMappingApplied` marker *before* it changes the mapping (`SuperKeyService.swift:265`, `:375`), and reads it on the next launch (`:302`) so a fresh process is authorized to undo what the dead one left behind. Its doc comment is about exactly this case: an app "killed while the feature was on leaves one behind". `SwitcherNativeHotkeys.suppressed` is a process-local `Set` with no persisted marker, so on the next launch `apply([])` computes its transition from an empty set and restores nothing.

Two more, both the same shape as things already in the repo:

- **A user who disabled ⌘Tab themselves gets it switched back on.** `setEnabled(key, false)` returns `.success` whether or not the key was already off, the key goes into `suppressed`, and teardown enables it. The doc comment says "puts back only the ones this process switched off"; nothing in the code distinguishes that. `CGSIsSymbolicHotKeyEnabled` is already bridged in the same directory at `SpaceWindowBridge.swift:245` — reading before writing, and only recording keys that were actually on, makes the comment true.
- **⌘⇧\` is not suppressed.** `nativeHotkeysToSuppress` takes the ⌘Tab pair (ids 1 and 2) but only the single ⌘\` id. `GlobalShortcut.shiftIsNavigationModifier` (`:386`) is true for the default window shortcut, so ⌘⇧\` reverse-cycles this switcher while the system's own reverse-window hotkey still fires underneath. The test even asserts "⌘Tab takeover also covers the system reverse switcher" with no window-shortcut counterpart — the pair was handled correctly one line above.

Separately, `collidesWithCommandTab` hardcodes keycodes 48/50 plus ⌘/⌘⇧, while `GlobalShortcut.conflictsWithSystemShortcut` (`:822`) reads the real system list. After suppression those plist entries read `enabled = false`, so the shortcut-conflict warning in Settings silently flips depending on whether the switcher is running. Two places answering the same question with nothing keeping them in agreement.

Taking over ⌘Tab is a product decision above my pay grade here, and it is worth settling before the code is reviewed further — but if it goes ahead, the launch-time repair is the part that has to exist first.

PathGao:
Following up on the product side I set aside above, because it decides whether the code questions are worth answering.

## The premise does not hold generally

#881 says a session event tap cannot preempt Dock's ⌘Tab, so Vorssaint never replaces it. Two other reports from this week say otherwise, both on macOS 26:

- **#920** (2026-08-24): Vorssaint's ⌘-Tab switcher "opens on the first Tab and disappears on the second". The first ⌘Tab reaches this app's tap.
- **#784** (macOS 26, 3.3.2): holding ⌘ and tapping Tab walks forwards through *this* switcher's list; only the ⇧ direction misbehaves. #804 is open to fix that direction.

Neither report could exist if Dock's symbolic hotkey preempted a head-inserted `.cgSessionEventTap`. So what you are seeing on your machine looks like a local condition — another tap ahead of ours, a `tapDisabledByTimeout` we resurrect too late, an Accessibility grant that is stale rather than absent — not a fact about macOS 26.

That matters beyond this PR: if it merges, a genuinely dead tap and a working one become indistinguishable, because ⌘Tab opens nothing either way. The bug that is actually on your machine would stop being reportable.

**What would change my reading:** a capture showing both switchers with the tap confirmed installed, on a machine where `CGSIsSymbolicHotKeyEnabled(1)` reads true. That separates "Dock wins the key" from "our tap is not running". If it is the second, the fix belongs in the tap's lifecycle and it is worth more than this one.

## What the change actually is

This does not change the App Switcher. It changes macOS. Every switcher change merged so far lives inside this process — a session tap, a panel, a shortcut this app owns. `CGSSetSymbolicHotKeyEnabled` writes the WindowServer's hotkey table: login-session state, shared with every other app, outliving this one. "The App Switcher takes ⌘Tab" and "Vorssaint switches a macOS shortcut off for you" are different commitments, and only the first is in the title.

Three ways that disagrees with what the repo already does:

1. **`main` reads symbolic hotkeys; it never writes them.** `SpaceWindowBridge.swift:239` and `:245` are the only uses, both reads, and `spaceShortcut(_:)` returns `nil` when the system shortcut is disabled — its comment says it honors user remaps and attempts nothing when there is none. The PR body says this takes "the same `CGSSetSymbolicHotKeyEnabled` path other replacements use". There is no such path; `git grep CGSSetSymbolicHotKeyEnabled origin/main` is empty. This is the first write in the repo.

2. **The repo already answered "our shortcut collides with a system one", the other way.** `GlobalShortcut.swift:807-814` states this PR's exact problem — macOS answers its own shortcuts first and does not consume them, so a shared combination fires both — and the answer it settled on is `conflictsWithSystemShortcut` (`:822`): detect the collision, report it on the shortcut row, leave the system alone, let the user decide in System Settings. This PR makes that decision for the user, silently, with nothing in the UI saying it happened.

3. **The system switcher is a documented fallback, not a competitor.** `SwitcherSupport.swift:48-50`: "The switcher lists windows, so an app that closed all of them disappears from it while the system switcher still offers it." The default is `SwitcherWindowlessApps.finder` — windows only, plus Finder. On stock settings, switch ⌘Tab off and a running app with no windows is no longer reachable from the keyboard at all: the thing that still offered it is the thing this PR turns off. `SwitcherAppRule.hidden` gains the same property — today it is a preference, afterwards a dead end. Nine comments across the module treat the system switcher as the model to imitate, down to the delay before the panel appears and the quick press that shows no UI. Removing it touches all of them and the diff mentions none.

## If it goes ahead anyway

Taking over ⌘Tab is a legitimate thing for a switcher to do and other apps do it. It is not legitimate as a side effect of enabling the App Switcher. What it would need:

- **Its own setting, off by default,** naming the system shortcuts it disables. Item 2 above is the repo's standing promise that Vorssaint does not touch System Settings behind the user's back; overriding that has to be the user's action, not ours.
- **Read before write.** `CGSIsSymbolicHotKeyEnabled` is already bridged next door at `SpaceWindowBridge.swift:245`. Recording only keys that were actually enabled is what makes your own doc comment — "puts back only the ones this process switched off" — true.
- **The launch-time repair from my first comment,** which is the part that has to exist before any of this ships: a persisted marker, the way `SuperKeyService` does it, so a crashed process does not leave ⌘Tab dead with nothing left to restore it.
- **A fallback for the windowless case,** or `SwitcherWindowlessApps` defaulting to `.all` whenever the takeover is on, so nothing becomes unreachable.

I would rather see #881 reopened as a tap-lifecycle question first. If the tap is losing ⌘Tab on your machine, that is a bug the rest of us have too and cannot currently see.

PathGao:
Four of the five from last time landed: opt-in that ships off, the crash marker, read-before-write, the windowless fallback. What is left:

**``⌘⇧` `` is still not suppressed.** id 6 became 27, but it is still one id where ⌘Tab gets two. The comment claims "WindowServer has one next-window action (id 27), so switching that action off covers both ``⌘` `` and its Shift-reversed direction" — the ⌘Tab pair one line above is the counterexample: forward and reverse are separate symbolic hotkeys, not one action plus a modifier. If the reverse-window action has its own id (28 in the public lists), ``⌘⇧` `` keeps firing the system window cycler while `GlobalShortcut.shiftIsNavigationModifier` drives yours underneath. `switcherShortcut(_:owns:)` already handles it — `nativeShortcut.modifiers.subtracting(.shift) == shortcut.modifiers` matches a ``⌘⇧` `` entry the moment `allCases` contains one. I cannot confirm 28 from this diff; `configuredShortcuts()` reads whatever ids the enum lists, so it is one `CGSGetSymbolicHotKeyValue(28, …)` away.

**The watchdog cannot tell the user apart from "another process".** The test names the intent: "the watchdog suppresses an owned hotkey if another process re-enables it". The process that re-enables ⌘Tab is System Settings, driven by the user. So the behavior is: switch ⌘Tab back on in Keyboard Shortcuts, watch it switch itself off a second later, with the windowless picker frozen at the same time. The caption promises the suppression lasts "only while Vorssaint's switcher is active" — true, and not the part that gets reported. It is also a 1 Hz main-runloop poll, six CGS round trips a second for the life of the process. Without it, `reconcileTakeover()` on wake and on preference change already covers every case where the actor is not the user.

**Three answers to "is the tap live".** `needsRecovery` in `reconcileTakeover` and `tapLive` in `applyNativeHotkeySuppressionIfTapLive` are near-negations of each other that disagree about `shouldStopTapThread`, and `isRunning` — the one named accessor for exactly this — is deleted in the same diff. One `var isTapLive` under `lifecycleLock`, called from both, is the same amount of code with one place to be wrong.

**The body describes the previous revision.** It still says the hotkeys go off "while the App Switcher tap is live and the shortcut is ⌘Tab (the default)". That was the old behavior; it now takes an explicit opt-in that ships off, which is the first thing a reviewer needs and the body does not mention. "hotkey ids 1, 2 and 6" is 1, 2 and 27. And it still says "the same `CGSSetSymbolicHotKeyEnabled` path other replacements use" — `git grep CGSSetSymbolicHotKeyEnabled origin/main` is still empty. This file is the first write in the repo, which is worth saying out loud rather than framing as an existing path.

Two I could not check from the diff:

- `mode(storedValue:takeOverSystemShortcuts:)` defaults the new parameter, so any call site other than `WindowEnumerator` and `SwitcherSettings` silently keeps the old behavior and the compiler says nothing. Dropping the default names them for you.
- `SpaceHopSupport.eventFlags(fromCarbonModifiers:)` applied to what `CGSGetSymbolicHotKeyValue` returns. If `SpaceWindowBridge` already converts that same call's output through it, ignore this. If not, every lookup misses, `nativeHotkeysToSuppress` returns empty, and the feature no-ops with all checks green — `configuredShortcuts()` is the one path the tests cannot reach.

**Impact:** only users who turn the new setting on. They lose ``⌘⇧` `` to the system window cycler, and cannot re-enable ⌘Tab from System Settings while the app runs. Default off, so nobody else is affected.
**Merge:** needs the reverse-window hotkey id first; the branch also conflicts with main.

ref #881

BenjaminD2023:
@PathGao Addressed the remaining review points in `156ebc57` and merged current `main` in `b194b2a4`:

- Added reverse-window symbolic hotkey ID 28 and covered all four forward/reverse app and window mappings in the default and remapped shortcut checks.
- Removed the 1 Hz watchdog and its test, so System Settings is no longer polled and immediately overridden.
- Replaced the duplicate tap-state checks with one locked `isTapLive` predicate used by recovery and suppression.
- Removed the default from `SwitcherWindowlessApps.mode`; every production and test caller now passes takeover state explicitly.
- Confirmed `configuredShortcuts()` uses the same Carbon modifier conversion as `SpaceWindowBridge`; a local macOS 26.4.1 probe also confirmed `CGSGetSymbolicHotKeyValue(28)` succeeds.
- Resolved the branch conflicts while retaining current-main behavior and the opt-in takeover additions.
- Rewrote the PR body to describe the current opt-in behavior and first WindowServer write accurately.

Checks: `git diff --check`; `./build.sh --test` (9,494 checks); `./build.sh --dev` (bundle built and signed without warnings).

PathGao:
All six landed. The `isTapLive` unification dropping `shouldStopTapThread` is right, not a regression — `installTap` already routes a start-during-stop through `pendingStartAfterStop`.

Two non-blocking things.

**A second bridge for the same two symbols.** `SpaceWindowBridge.swift:257-268` already `dlsym`s `CGSGetSymbolicHotKeyValue` and `CGSIsSymbolicHotKeyEnabled` with identical `@convention(c)` typealiases, in the same directory. The two copies already disagree on the unbound sentinel: `keyCode != 0` there, `keyCode != 0xFFFF` here. I can't tell which is right; they can't both be.

**The toggle label hardcodes the keys the code deliberately doesn't.** "Replace macOS ⌘Tab and ⌘`", ×14 locales — while `configuredShortcuts()` and your own test ("follows the current macOS shortcut mappings instead of hardcoded keys") match whatever is mapped now. Someone who remapped the system app switcher to ⌥Tab reads a toggle about ⌘Tab and gets ⌥Tab switched off. The caption already words it correctly.

Read-and-compile only; none of the WindowServer behavior verified here.

**Impact:** none — both are wording and duplication.
**Merge:** ship it.

ref #881

PR #891, opened 2026-08-24
feat(media): compress video and GIF to a target file size

FlowSync0:
## Before

The video and GIF tools can only be given a resolution. Fitting an upload
limit therefore means guessing a width, compressing, looking at the file and
going round again: the resolution is the input, the file size is whatever
comes out.

This came out of trying to post a Vorssaint recording on Discord, which
refuses anything above 20 MB without Nitro. What was wanted was the highest
quality that still fits under that ceiling, and a resolution stepper cannot
express it.

## After

Both tools now choose between two modes.

- **Resolution** — the existing behavior, untouched, still the default.
- **File size** — a ceiling typed in megabytes, everything else derived from it.

## How it works

A file size is a bitrate held for a duration, so the ceiling becomes a bitrate
budget, and that budget picks the resolution it can afford, around 0.07 bit per
pixel and frame, rather than starving a larger frame.

`avconvert`, which the resolution mode drives, only accepts presets and exposes
no bitrate, so the size targeted path reads into an `AVAssetWriter` and sets
`AVVideoAverageBitRateKey` directly. That is the same approach
`RecordingSharingSupport` already uses to fit a recording export under the
sharing limit. Rate control lands near a budget rather than on it, so a pass
that overshoots is measured and re-encoded at a smaller scale, three passes at
most.

A GIF has no bitrate at all, its weight is frames times pixels, so an overshoot
gives up frame rate first and frame size second, and it starts from the source
itself rather than from the width the resolution mode last used.

A ceiling too small to reach at a watchable quality is reported rather than
silently missed.

## Notes

- The resolution path is untouched: the size target is a separate branch, and
  `avconvert` still handles everything it handled before.
- Megabytes are 1000 based, so a file that fits a "20 MB" ceiling also fits a
  20 MiB one, never the other way round.
- New user facing text is provided in every supported language.

## Checks

- `./build.sh --test` prints `TESTS OK (8327 checks)`, including new assertions
  for the planner: budget respected, no upscaling, even dimensions, the 480 px
  floor, unreachable targets refused, retry scale, GIF strategy, strings present
  in every language.
- Encoded end to end on real files. A 61 s 1280x830 clip at a 3 MB ceiling gives
  2.77 MB at 582x376, at a 5 MB ceiling 4.60 MB at 750x486, both around 92 % of
  the target on the first pass. The same clip with an audio track gives 4.60 MB
  with AAC preserved over the full duration.


FlowSync0:
Please find below screenshot proof of work :

Capture d’écran 2026-08-24 à 14 39 12
Capture d’écran 2026-08-24 à 14 39 31
Capture d’écran 2026-08-24 à 14 39 58
Capture d’écran 2026-08-24 à 14 40 08
Capture d’écran 2026-08-24 à 14 40 26
Capture d’écran 2026-08-24 à 14 40 29

Issue #897, opened 2026-08-24
Snippet library panel receives no keyboard input when using the Accessibility Keyboard


fermincasagrande:
### Description

`⌥⌘S` opens the snippet library correctly, but the panel receives **no keyboard input at all** when the macOS Accessibility Keyboard is the input device:

- **Down / Up** — the panel dismisses instead of moving the selection.
- **Enter** — nothing happens.
- **Typing in the search field** — the characters land in the *previously focused application* instead.

Selecting a snippet **with the mouse works perfectly**, so insertion and permissions are fine. Only keyboard navigation of the panel is affected.

After forcing Vorssaint to the foreground with `NSRunningApplication.activate`, keystrokes stop reaching the other app and instead produce the system beep — the app is active, but no window will accept key input.

### Evidence

Queried through the Accessibility API while the library was open:

```
windows=1
  w1 role=AXWindow subrole=AXSystemDialog focused=false
frontmost=Vorssaint
```

`AXFocused` remained `false` after setting `AXMain`, setting `AXFocused`, *and* activating the application.

### Relevant code

`SnippetLibraryService.ensurePanel()` builds the panel as `styleMask: [.borderless, .nonactivatingPanel]`. `show()` then calls `panel.orderFrontRegardless()` followed by `panel.makeKey()`, and key handling is installed as:

```swift
keyMonitor = NSEvent.addLocalMonitorForEvents(matching: .keyDown) { ...
    guard let self, let panel, event.window === panel else { return event }
```

A **local** monitor only sees events delivered to this application. When the panel does not actually become key, the monitor never fires and the keystrokes go to whichever app is frontmost — which matches the observed behaviour exactly.

I can't say *why* `makeKey()` fails to take effect in this setup specifically; it may be an interaction with the Accessibility Keyboard being an input method (`com.apple.inputmethod.AssistiveControl`), which may hold key-focus semantics of its own. That part needs someone who knows the panel code.

### Steps to Reproduce

1. Enable the Accessibility Keyboard, with no physical keyboard attached.
2. Focus a text field in any app.
3. Press `⌥⌘S` to open the snippet library.
4. Press Down — the panel dismisses. Press Enter — nothing. Type — the text appears in the app underneath.
5. Click a snippet with the mouse — it inserts correctly.

### Suggested direction

[...]

Issue #896, opened 2026-08-24
Snippet triggers never expand when typing on the macOS Accessibility Keyboard


fermincasagrande:
### Description

With the macOS Accessibility Keyboard (System Settings → Accessibility → Keyboard → Accessibility Keyboard) as the only input device, **typed snippet triggers never expand**. The snippet is enabled and correct, Vorssaint has Accessibility permission, and expansion works fine when the same keystrokes are injected programmatically — so this is specific to typing on the on-screen keyboard.

**Root cause.** In `TextSnippetService.handle(type:event:)`:

```swift
// Clicks move the caret somewhere unknown; the half-typed trigger is
// no longer where the deletes would land.
if type == .leftMouseDown || type == .rightMouseDown {
    resetBuffer()
    return Unmanaged.passUnretained(event)
}
```

On the Accessibility Keyboard **every character is produced by a mouse click on the on-screen key**. A `leftMouseDown` therefore arrives with every single keystroke, `resetBuffer()` runs each time, and the trigger buffer never holds more than one character. Any trigger longer than one character is impossible to complete.

The reasoning behind the reset is right for ordinary clicks — but a click on the Accessibility Keyboard's own panel types a character and leaves the caret exactly where it was.

### Steps to Reproduce

1. Enable the Accessibility Keyboard.
2. Create a snippet: trigger `;em`, expansion mode "after delimiter", any replacement.
3. Using the on-screen keyboard, click `;`, `e`, `m`, then space.
4. Nothing expands; the trigger stays as typed.
5. For contrast, inject the identical keystrokes programmatically (no mouse events) — expands correctly every time.

### Evidence

Captured with a `CGEventTap` while typing `;em ` on the Accessibility Keyboard. The key events themselves are perfectly well formed:

```
keyDown kc=43 chars=';'  srcState=0 userData=0x7FFFFFFFFFFFFFFE mods=shift
keyDown kc=14 chars='e'  srcState=0 userData=0x7FFFFFFFFFFFFFFE
keyDown kc=46 chars='m'  srcState=0 userData=0x7FFFFFFFFFFFFFFE
keyDown kc=49 chars=' '  srcState=0 userData=0x7FFFFFFFFFFFFFFE
```

Worth noting `userData` is Assistive Control's own marker and does **not** collide with `syntheticMarker` (`0x564F5253`), so the guard at the top of the callback is not involved. It is the mouse-down reset.

### Suggested fix

[...]

PR #898, opened 2026-08-24
fix(snippets): keep the snippet buffer when the click typed on the Accessibility Keyboard

fermincasagrande:
Refs #896.

### The problem

Snippet triggers never expand for anyone using macOS's on-screen **Accessibility Keyboard**. The snippet is enabled and correct, permissions are granted, and the key events themselves are perfectly well formed — but nothing ever matches.

Every key on that panel is pressed **with the mouse**, so a `leftMouseDown` accompanies every character typed. In `handle(type:event:)`:

```swift
if type == .leftMouseDown || type == .rightMouseDown {
    resetBuffer()
}
```

`resetBuffer()` therefore runs before every single keystroke, and the buffer never holds more than one character. No trigger longer than one character can ever match.

### The change

The existing reset is right for ordinary clicks: they move the caret, and the half-typed trigger is no longer where the deletes would land. A click on the Accessibility Keyboard's own panel is the exception — it types a character and leaves the caret exactly where it was.

So the click is hit-tested against the on-screen window list, and the reset is skipped only when it landed on that panel. Every other click behaves exactly as before.

Matched by **bundle identifier** (`com.apple.inputmethod.AssistiveControl`) rather than window owner name — the owner name is `AssistiveControl`, which is the kind of string that gets localized.

Mouse-downs are infrequent, so the `CGWindowListCopyWindowInfo` hit-test costs nothing measurable, and it only runs on mouse events — never in the key path.

### Testing

Built with `./build.sh --dev` and run on macOS 26.5.2 (25F84), Apple Silicon, with the Accessibility Keyboard as the *only* input device — no physical keyboard attached.

- Trigger `;em` with "after delimiter": **expands correctly** when typed on the on-screen keyboard. Before the change, never.
- Ordinary clicking to move the caret mid-trigger still resets the buffer as expected.
- `--selftest` passes.

### Note

The reporter uses this keyboard as their only means of typing. There is a second, unrelated bug in the same feature area — the snippet library panel receives no keyboard input in the same setup — filed separately as #897. This PR does not address that one.

### DIFF
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -47,6 +47,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         }
         beginStartupWatch()
         Self.boundAccessibilityWaits()
+        // Resolve the Accessibility Keyboard's pid now. The lookup is async, so
+        // a feature that asks first and has no second chance — the switcher
+        // judges a click only after cancelSession() has already run — would
+        // otherwise be told "not running" once per launch.
+        _ = AssistiveKeyboard.isRunning
 
         // Finish the on-disk rename for installs carried over from a pre-2.5
         // build, or retire a leftover old-named bundle. Returns true when we are
--- a/Sources/Vorssaint/Services/AssistiveKeyboard.swift
+++ b/Sources/Vorssaint/Services/AssistiveKeyboard.swift
@@ -31,6 +31,12 @@ enum AssistiveKeyboard {
     private static let minLookupInterval: TimeInterval = 2
 
     /// Seeds the cache and installs the observers. Runs once, on first use.
+    ///
+    /// Registering `NSWorkspace` observers happens on whichever thread gets here
+    /// first, which is the main thread only because `applicationDidFinishLaunching`
+    /// touches `isRunning` before any tap can. That warm-up is load-bearing for
+    /// two separate reasons — this one and the cold first answer — so do not
+    /// remove it as a redundant optimization.
     private static let bootstrap: Void = {
         scheduleLookup(force: true)
         let center = NSWorkspace.shared.notificationCenter
@@ -157,10 +163,27 @@ enum AssistiveKeyboard {
                   CGRect(x: x, y: y, width: width, height: height).contains(point)
             else { continue }
             // A fully transparent layer cannot have been clicked; the click
-            // went through it to whatever is behind.
+            // went through it to whatever is behind. Worth keeping for foreign
+            // windows, but it is not enough on its own — see below.
             if let alpha = window[kCGWindowAlpha as String] as? CGFloat, alpha <= 0 { continue }
             let owner = window[kCGWindowOwnerPID as String] as? pid_t
-            return owner == pid
+            if owner == pid { return true }
+            // Our own windows never block the answer. Several are click-through
+            // (`ignoresMouseEvents`), and `CGWindowListCopyWindowInfo` cannot
+            // report that; they also leave `alphaValue` alone, so they arrive
+            // here looking like ordinary opaque windows. One of them is
+            // full-screen at shielding level for as long as Extra Brightness is
+            // boosting, which would put it above every point of the display and
+            // make this answer false everywhere — every call site silently back
+            // to the bug it fixes.
+            //
+            // The trade: one of our windows that *does* take clicks, overlapping
+            // the keyboard, now reads as a keystroke. That is a false positive —
+            // a buffer that survives, a panel that stays open — against a false
+            // negative that disables the feature outright. The rarer and milder
+            // of the two.
+            if owner == ProcessInfo.processInfo.processIdentifier { continue }
+            return false
         }
         return false
     }
--- a/Sources/Vorssaint/Services/AssistiveKeyboard.swift
+++ b/Sources/Vorssaint/Services/AssistiveKeyboard.swift
@@ -65,8 +65,20 @@ enum AssistiveKeyboard {
         let cached = cachedPID
         lock.unlock()
 
-        // The hot path: still the process we know about.
-        if let cached, kill(cached, 0) == 0 { return cached }
+        // The hot path. `kill(pid, 0)` answers "some process has this pid",
+        // not "the keyboard does" — pids are recycled, and once one is reused
+        // this would match an unrelated process's windows forever, silently.
+        // So a live pid is still re-resolved in the background, which the floor
+        // below keeps to one lookup every `minLookupInterval`; staleness is
+        // bounded rather than permanent, at no extra syscall.
+        //
+        // EPERM means the process exists but cannot be signalled. Treating that
+        // as dead would clear the cache on every call and re-resolve the same
+        // unsignallable pid forever.
+        if let cached, kill(cached, 0) == 0 || errno == EPERM {
+            scheduleLookup(force: false)
+            return cached
+        }
 
         if cached != nil {
             lock.lock()
@@ -103,8 +115,18 @@ enum AssistiveKeyboard {
         }
     }
 
-    /// Whether the Accessibility Keyboard is running. Cheap enough for an event
-    /// tap: a lock and, at most, one `kill(pid, 0)`.
+    /// Whether the Accessibility Keyboard is running.
+    ///
+    /// Cheap enough for an event tap, but not free: a live pid costs two locks
+    /// and one `kill(pid, 0)`, and either answer may schedule a background
+    /// resolution — at most one per `minLookupInterval`, never on the calling
+    /// thread.
+    ///
+    /// Also the warm-up. The first call resolves asynchronously and answers
+    /// "not running" while it does, so somewhere that runs at startup should
+    /// touch this: a caller with no second chance — the switcher has already
+    /// cancelled its session by the time a click is judged — would otherwise
+    /// get one wrong answer per launch.
     static var isRunning: Bool { livePID() != nil }
 
     /// True when `point` lands on the Accessibility Keyboard's own panel, i.e.
--- a/Sources/Vorssaint/Services/AssistiveKeyboard.swift
+++ b/Sources/Vorssaint/Services/AssistiveKeyboard.swift
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import AppKit
+import CoreGraphics
+
+/// macOS's on-screen Accessibility Keyboard.
+///
+/// Someone who cannot use a physical keyboard presses every key on that panel
+/// with the mouse, so a mouse-down accompanies every character they type.
+/// Anything that reads a click as "the user did something other than type" —
+/// dismissing a panel, clearing a typed-so-far buffer — has to make an
+/// exception for clicks that landed on it, or the feature cannot be used from
+/// that keyboard at all.
+///
+/// The check is deliberately cheap when the keyboard is not running, which is
+/// the overwhelmingly common case: the pid is cached and kept current from
+/// workspace notifications, so callers on an event tap pay one lock and a nil
+/// test rather than a window enumeration.
+enum AssistiveKeyboard {
+
+    static let bundleID = "com.apple.inputmethod.AssistiveControl"
+
+    private static let lock = NSLock()
+    private static var cachedPID: pid_t?
+    private static var lastLookup: TimeInterval = -.greatestFiniteMagnitude
+    private static var lookupInFlight = false
+
+    /// Floor between two resolutions when the keyboard is not running, so a
+    /// machine that never opens it does not pay a lookup per click.
+    private static let minLookupInterval: TimeInterval = 2
+
+    /// Seeds the cache and installs the observers. Runs once, on first use.
+    private static let bootstrap: Void = {
+        scheduleLookup(force: true)
+        let center = NSWorkspace.shared.notificationCenter
+        for name in [NSWorkspace.didLaunchApplicationNotification,
+                     NSWorkspace.didTerminateApplicationNotification] {
+            center.addObserver(forName: name, object: nil, queue: .main) { note in
+                guard let app = note.userInfo?[NSWorkspace.applicationUserInfoKey]
+                        as? NSRunningApplication,
+                      app.bundleIdentifier == bundleID
+                else { return }
+                scheduleLookup(force: true)
+            }
+        }
+    }()
+
+    /// The keyboard's pid, or nil when it is not running.
+    ///
+    /// Deliberately does not trust the workspace notifications on their own.
+    /// Assistive Control is an input method rather than an ordinary app, and it
+    /// does not reliably announce launching or terminating — so a cache fed only
+    /// by those notifications goes wrong in both directions and does it
+    /// silently: restart the keyboard and every check compares against a dead
+    /// pid, or start it after this process and the cache stays nil forever. In
+    /// both cases the feature simply stops working with nothing to point at.
+    ///
+    /// So the cached pid is verified against the live process on each call —
+    /// `kill(pid, 0)` is a bare syscall, which is affordable even on an event
+    /// tap — and re-resolved off-thread whenever it turns out to be wrong.
+    private static func livePID() -> pid_t? {
+        _ = bootstrap
+        lock.lock()
+        let cached = cachedPID
+        lock.unlock()
+
+        // The hot path: still the process we know about.
+        if let cached, kill(cached, 0) == 0 { return cached }
+
+        if cached != nil {
+            lock.lock()
+            if cachedPID == cached { cachedPID = nil }
+            lock.unlock()
+        }
+        // Resolving means asking NSWorkspace, which is far too slow for a tap
+        // callback. Kick it off and answer "not running" for now; the next click
+        // gets the corrected answer.
+        scheduleLookup(force: cached != nil)
+        return nil
+    }
+
+    private static func scheduleLookup(force: Bool) {
+        let now = ProcessInfo.processInfo.systemUptime
+        lock.lock()
+        if lookupInFlight || (!force && now - lastLookup < minLookupInterval) {
+            lock.unlock()
+            return
+        }
+        lookupInFlight = true
+        lastLookup = now
+        lock.unlock()
+
+        DispatchQueue.global(qos: .utility).async {
+            let pid = NSRunningApplication
+                .runningApplications(withBundleIdentifier: bundleID)
+                .first?
+                .processIdentifier
+            lock.lock()
+            cachedPID = pid
+            lookupInFlight = false
+            lock.unlock()
+        }
+    }
+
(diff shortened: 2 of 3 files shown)


PathGao:
Same note as on #922, which needs reading together with this one.

**#922 adds the identical function.** Its `clickTypedOnAssistiveKeyboard(_:)` and `assistiveKeyboardBundleID` are character-for-character what this PR adds, in `Services/Switcher/AppSwitcher.swift` instead of `TextSnippetService.swift`. Whichever lands first should put the predicate somewhere shared.

**And four more panels have the same bug.** Every character typed on the Accessibility Keyboard is a `leftMouseDown`, and these treat a click outside themselves as "dismiss" while accepting typed input:

- `Services/CommandBar/CommandBarService.swift:2493`
- `Services/QuickTools/QuickLauncherService.swift:401`
- `Services/Clipboard/ClipboardHistoryService.swift:1180`
- `Services/Snippets/SnippetLibraryService.swift:302`

All four use the same `localClickMonitor` / `outsideClickMonitor` pair, so none of them can be typed into from that keyboard at all. They use `NSEvent` monitors rather than a `CGEvent` tap, so a shared predicate wants to take a point: `AssistiveKeyboard.ownsPoint(_:)` called from all six.

Would you widen it? The `incomplete` label is about the scope, not the code — the fix itself reads correctly.

PathGao:
Separate from the scope question above: the predicate is doing the wrong kind of work for the thread it runs on.

**The window scan runs on an active event tap, on every click, for every user.** The tap is `.cgSessionEventTap` / `.headInsertEventTap` / `.defaultTap` (`TextSnippetService.swift:142-146`), so the WindowServer holds the mouse-down until `handle(type:event:)` returns. `clickTypedOnAssistiveKeyboard` (`:29`, called at `:256`) does a synchronous `CGWindowListCopyWindowInfo` over every on-screen window plus an `NSRunningApplication(processIdentifier:)` lookup per window under the point — and it does that on every `leftMouseDown` and `rightMouseDown`, whether or not the Accessibility Keyboard is running. Today that path was three instructions under a lock. If the callback ever runs long, the tap gets `.tapDisabledByTimeout` and snippets stop until the next event re-enables it.

This file already refuses to do this. `:270-272` mirrors `libraryVisible` / `commandBarVisible` behind `inputLock` with the comment "so this callback never has to ask AppKit or wait for the main thread" — the same treatment fits here. Keep the Accessibility Keyboard's pid (nil when it is not running) behind `inputLock`, refreshed on the main queue from `NSWorkspace.didLaunchApplicationNotification` / `didTerminateApplicationNotification` alongside the existing observer at `:200-206`, and have the tap do the window scan only when that pid is non-nil. On the machines that do not use the Accessibility Keyboard — nearly all of them — the mouse-down path stays exactly what it is now.

That also changes what the shared predicate should look like. The four panels use `NSEvent` monitors on the main thread, where a `CGWindowListCopyWindowInfo` per outside-click costs nothing worth measuring; the tap is the one caller that cannot pay it. So `AssistiveKeyboard.ownsPoint(_:)` wants to be cached-pid plus last-known bounds, not an unconditional window enumeration, or widening the fix pushes the cost onto the site that is least able to absorb it.

Two smaller things:

- Every other point hit-test against the window list in this repo passes `.excludeDesktopElements` — `WindowServerTrafficLightHitTest.swift:30`, `MouseAppExceptions.swift:238`, `WindowLayoutService.swift:353`, `DockClickService.swift:903`. `:31` here passes only `.optionOnScreenOnly`. I cannot tell from the source how the Accessibility Keyboard's panel is classified, so this may be deliberate — if it is, say so in the comment, because the next person will read it as an oversight.
- `:200-206` resets the buffer on `NSWorkspace.didActivateApplicationNotification` too. If pressing a key on the Accessibility Keyboard ever activates `com.apple.inputmethod.AssistiveControl`, that second reset path clears the buffer regardless of this fix. I have no way to observe that here — worth confirming on hardware with a multi-character trigger before calling this done, since it is the same defect at a second site in the same file.

PathGao:
One more, separate from the scope and thread questions above: the loop answers "is the Accessibility Keyboard *anywhere* under the point", not "is it the window the click actually hit".

`CGWindowListCopyWindowInfo([.optionOnScreenOnly], …)` comes back front-to-back, and `:42` `continue`s past a window that contains the point but fails the bundle check, instead of stopping at it. So a click landing on any window stacked above the panel, anywhere the two overlap, reads as a keystroke and the buffer survives. The body says "Every other click behaves exactly as before" — this is the case where it doesn't, and it's the one the reset exists for: caret moved, trigger stale, the deletes land somewhere else.

Strict frontmost-wins is the obvious fix but has its own failure mode — one transparent overlay above the panel turns it into a false negative and the original bug is back. The cached-pid form from the comment above avoids both: ask "is the frontmost window owned by *this pid* the one under the point", which needs the panel's own bounds rather than a full enumeration. I can't tell from here how the Accessibility Keyboard's window level compares to Vorssaint's own overlays, so I don't know how often either case is reachable in practice — but the check shouldn't depend on that being rare.

**Impact:** clicks on a window overlapping the Accessibility Keyboard panel skip the buffer reset, so a stale trigger can expand at a caret that has moved. Only on machines running that panel.
**Merge:** needs the window scan gated behind a cached pid first — the frontmost question folds into that same rewrite.

ref #896

fermincasagrande:
Thanks — all three points were right, and one of them was more right than you could have known from the diff. Reworked; this PR now adds `AssistiveKeyboard` and uses it, and #922 calls it from the other five sites.

**Scope and duplication.** Extracted to `Services/AssistiveKeyboard.swift`. #922 is rebased on this branch and calls it rather than carrying a second copy, so the ordering you suggested is the one in place: this lands first. It also widens to the four `NSEvent` panels, which takes out #897 at the same time.

**The window scan on the tap.** Fair, and it is the same mistake I had diagnosed in someone else's code two days earlier. Gated behind a cached pid: when the keyboard is not running — nearly every machine — a mouse-down costs a lock and one `kill(pid, 0)`, never an enumeration. Resolution happens off-thread, so the tap callback never waits on `NSWorkspace`.

**"Anywhere under the point" vs "was clicked".** Corrected, and you were right that it was reachable: the loop now stops at the first window containing the point and tests that one, skipping fully transparent layers since a click passes through them. `.excludeDesktopElements` added — the omission was an oversight, not deliberate, so I have not written a comment defending it.

**`didActivateApplicationNotification`.** You flagged this as unverified and it was real. Exempted here, and in the four panels in #922 — same defect, five files.

---

One thing worth reporting, because it only showed up on hardware and it changes the shape of the fix.

The first version of the cache refreshed the pid **only** from the workspace launch/terminate notifications. Assistive Control is an input method, and it does not reliably post them. Restarting the keyboard left every call site comparing against a dead pid, so all six fixes stopped working at once — silently, and "fixed" by restarting the app, which is exactly the sort of clue that sends you looking in the wrong place. The same hole exists in the other direction: start the keyboard after the app and the cache stays `nil` forever.

So the cached pid is now verified against the live process on every call and re-resolved off-thread when it is wrong, with a floor between lookups. Confirmed by restarting the Accessibility Keyboard **without** restarting the app: previously that broke all six sites until relaunch; now they keep working.

If you would rather this leaned on a `didActivate` observer as well, or on something the app already tracks, say so — the current shape is deliberately self-healing rather than notification-dependent, but I have no visibility into whether you already have a better hook for it.

Tested on macOS 26.5.2, Apple Silicon, with the Accessibility Keyboard as the only input device: multi-character triggers expand, ordinary clicks still reset the buffer, and `--selftest` passes.

PathGao:
The pid cache still has one silent failure, and it's in the harmful direction.

`kill(pid, 0)` answers "some process has this pid", not "the Accessibility Keyboard has this pid". Pids get recycled. Once Assistive Control exits and something else takes its number, the hot path at `:69` keeps returning 0 forever, `ownsPoint` starts matching that unrelated process's windows, and every click on it reads as a keystroke — no reset, stale trigger, deletes at a caret that moved. It never re-resolves, because from `livePID`'s side nothing looks wrong. Same shape as the one hardware found: silent, permanent until relaunch, and it points somewhere else.

The fix costs nothing you aren't already paying — bound the verify instead of relying on `kill` alone:

```swift
if let cached, kill(cached, 0) == 0 {
    scheduleLookup(force: false)   // floor keeps this to one resolution per 2s
    return cached
}
```

The not-running path already calls exactly this on every click (`:79`, `cached == nil` → `force: false`), so the running path picking it up adds a branch and no new syscall. Staleness goes from unbounded to `minLookupInterval`. While you're there, `:16-19` promises callers "one lock and a nil test" — the nil path is two locks plus a `systemUptime` and a background enumeration every 2s. That's the right trade given input methods don't post reliably, but the comment doesn't say it.

**On `didActivate`: no, don't add it.** An input method coming forward isn't a launch, and `TextSnippetService.swift:209` now deliberately ignores that notification for this bundle id — wiring the cache to it too would mean two call sites reading the same event in opposite directions. It also wouldn't have caught the restart case, since the failure there was "the process changed and nothing said so". Verify-on-use is the right shape. Whether the app already tracks running pids somewhere better, I can't say — the rest of the repo isn't in the diff.

One thing I can't check: `alpha <= 0` covers invisible, but click-through isn't invisible. A window with `ignoresMouseEvents` and alpha above zero sits in the list looking clickable and the click passes under it, so if any Vorssaint overlay — HUD, highlight, dim layer — can be on screen above the panel, `ownsPoint` returns false and the original bug is back on the exact machine that reported it. `CGWindowListCopyWindowInfo` doesn't expose `ignoresMouseEvents`, so there's no cheap test for it. I can't see those services from the diff, so this is the failure mode you traded into, not one I've confirmed is reachable.

**Impact:** after Assistive Control quits and its pid is reused, clicks on whatever inherited it skip the buffer reset for the rest of the app's lifetime. Only machines that have run the keyboard.
**Merge:** needs the bounded re-verify on the cached-pid path first — one line.

ref #896

fermincasagrande:
All three fixed, and the pid-recycling one was the important catch — thank you.

**Bounded re-verify.** Taken as written: the live-pid path now calls `scheduleLookup(force: false)` before returning, so the floor keeps it to one resolution per interval and staleness stops being unbounded. You were right that it costs nothing new — the not-running path already made that exact call on every click. The comment now says what the check actually proves, which is "some process has this pid", not "the keyboard does".

**EPERM.** Added `|| errno == EPERM`. Agreed it is likely unreachable for a same-uid input method, but read-as-dead would have meant cache cleared every call, `force: true` bypassing the floor, and a dispatch per click forever.

**The cost comment.** Rewritten. It claimed one lock and a nil test; it is two locks, a `kill`, and sometimes a background enumeration. Stated plainly rather than defended.

**`didActivate`.** Not added, and the reason you gave is better than the one I had: `TextSnippetService.swift` now deliberately ignores that notification for this bundle id, so wiring the cache to it would put two call sites on opposite readings of the same event. It also would not have caught the restart, where the failure was that nothing announced anything.

Also folded in from #922, since it belongs with the helper: the cache is warmed in `applicationDidFinishLaunching`, because the first call resolves asynchronously and answers "not running" while it does. Your diagnosis of why my hardware run passed anyway is correct and was the part I would not have found — the snippet tap calls `ownsPoint` on every `leftMouseDown`, so it primes the cache, and the switcher's correctness was quietly depending on the snippets feature being enabled.

---

**On `ignoresMouseEvents`:** I cannot rule it out either. `CGWindowListCopyWindowInfo` does not expose it, so there is no cheap test, and I have no view of whether any Vorssaint overlay can sit above the keyboard panel. What I can say is that the alternative — treating any window over the point as not-the-keyboard, which is what the first version effectively did — is wrong in the same direction and more often. If you know of an overlay that can be on screen there, I would rather special-case it by owner pid than guess at click-through.

Built and self-tested. Not re-verified on hardware: the reporter's machine is deliberately left on the previous build, so the recycled-pid and cold-start paths here are reasoned, not observed.

PathGao:
Yes, there is one, and it is the worst possible shape for this check: `ExtraBrightnessService`.

`showOverlay(on:)` (`ExtraBrightnessService.swift:283-297`) creates a window at `screen.frame`, `level = CGShieldingWindowLevel()`, `ignoresMouseEvents = true`, and never touches `alphaValue` — so `kCGWindowAlpha` reports 1 and your `alpha <= 0` skip does not catch it. While the boost is on, that window is the frontmost thing over every point of the display. `ownsPoint` walks front-to-back, hits it first, `owner != pid`, returns false. Every click. All six sites revert to the original bug for as long as Extra Brightness is enabled, and nothing about it looks broken from the inside. `showTrigger` at `:339-349` is the same window at 1×1 in a corner.

It is not the only one, just the only full-screen one. `main` has ten `ignoresMouseEvents = true` sites plus a user-toggled one, all owned by us:

- `ExtraBrightnessService.swift:291`, `:345` — full screen, shielding level, resident while boosting
- `Recorder/RecorderIndicator.swift:64` — `.statusBar`, resident for the whole recording
- `QuickTools/ScreenshotPinController.swift:150` — user pins a capture, opts it into click-through, leaves it anywhere on screen
- `Display/BrightnessOSD.swift:112`, `WindowLayout/WindowLayoutService.swift:932`/`:1359`, `DockPreview/DockPreviewDragGhost.swift:43`, `App/AppDelegate.swift:720`, `QuickTools/QuickToolHUD.swift:234`, `UI/Shelf/ShelfTooltipPopover.swift:103` — transient

`DockPreviewDragGhost:50` dims the *view*, not the window, so even that one reports alpha 1. And `ScreenshotPinController.swift:50-63` is prior art for exactly this class of problem — it needs a global option-click monitor because its own click-through window otherwise swallows the hit.

Since every click-through window in that list is ours, the special-case you offered is right, just at the pid rather than at one overlay:

```swift
let owner = window[kCGWindowOwnerPID as String] as? pid_t
if owner == pid { return true }
if owner == ProcessInfo.processInfo.processIdentifier { continue }
return false
```

What you trade: a Vorssaint window that *does* take clicks and overlaps the panel now reads as a keyboard click. That is a false positive — buffer survives, panel does not dismiss — against a false negative that kills the feature outright whenever the boost is on, and it is the rarer of the two. `alpha <= 0` still earns its place for foreign windows.

Two things I did not verify. Whether the WindowServer actually stacks a `CGShieldingWindowLevel` window above Assistive Control's panel — I am arguing from the level constant, not from an observed composite. And whether the reporter runs Extra Brightness at all. Neither changes the fix; both change how often it is hit.

Smaller: `bootstrap` registers the `NSWorkspace` observers on whichever thread calls `livePID()` first. Today that is guaranteed main only because of the `AppDelegate` warm-up, so that line is doing two jobs and its comment names one. Say so there, or the next person deletes it as a redundant optimization and moves observer registration onto the tap thread.

**Impact:** with Extra Brightness on, its full-screen click-through overlay is frontmost at every point, `ownsPoint` is false everywhere, and all six call sites behave as if the fix were not there. Only machines running both that feature and the Accessibility Keyboard, but on those it is total.
**Merge:** needs the own-pid skip in the hit-test first.

ref #896

fermincasagrande:
`ExtraBrightnessService` is a much better catch than the abstract worry I raised, and the fix is in as you wrote it.

The hit-test now skips windows owned by our own pid entirely, rather than trying to guess which of the nine are click-through. `alpha <= 0` stays for foreign windows. The trade is stated in the code: one of ours that *does* take clicks, overlapping the panel, reads as a keystroke — a buffer that survives and a panel that stays open, against a false negative that disables all six sites for as long as a boost is on.

Two things I could not settle either, so I have not written comments claiming otherwise. Whether the WindowServer really stacks `CGShieldingWindowLevel` above Assistive Control's panel — I have not observed a composite, and the reporter has Extra Brightness off, so I could not reproduce it even deliberately. And whether any of the other eight can overlap the keyboard in practice. The fix does not depend on either, which is why I took it as written rather than special-casing the one overlay.

**`bootstrap`'s threading.** Noted where you said. The comment now says the observers land on whichever thread arrives first, that this is main only because the `applicationDidFinishLaunching` warm-up gets there before any tap does, and that the warm-up is therefore load-bearing for two separate reasons — so it does not read as a redundant optimization to whoever meets it next.

---

I have also stopped reacting one report at a time. `grep -rl "outsideClickMonitor = NSEvent.addGlobalMonitorForEvents"` returns six services; I had patched four and you found the other two. One command, available to me two rounds ago. The remaining `ignoresMouseEvents` sites are enumerated in your comment and I have taken the pid-level fix rather than chasing them individually for the same reason.

Built, self-tested, and confirmed on hardware by the original reporter: snippets, ⌘-Tab, the snippet library and camera preview all behave. Extra Brightness and Scratchpad are off on that machine, so those two paths remain reasoned rather than observed.

PathGao:
The completeness grep matches a variable name, not the behavior, and one site doesn't use that name.

`outsideClickMonitor` returns six services because six of them happened to name the variable that. `AppDelegate.swift:867` is the same shape under a different name — `popoverDismissMonitor`, a global monitor on `[.leftMouseDown, .rightMouseDown]` that calls `closePopover()` for any click that isn't on the status item. And the popover takes typed input: `:889` installs a keyDown monitor, `:933-936` names the Homebrew search field specifically, and `isTextEditingActive(in:)` exists so Space and Return reach the field editor. So one character typed on the Accessibility Keyboard closes the menu bar panel — the same defect as the six, on the app's main surface. (Line numbers against `main`; this PR's insert shifts them five.)

The handler already reads `NSEvent.mouseLocation` at `:872`, so it takes the same form as the others:

```swift
guard !AssistiveKeyboard.ownsCocoaPoint(NSEvent.mouseLocation) else { return }
```

`:871`'s `PanelInteractionState.shared.preventsPopoverDismissal` is a second exemption channel — I can't tell whether that is meant to be the one place exemptions live, so put this wherever the other call sites went. That is #922, even though the file is one this PR already edits.

I ran the behavior-shaped version too (`addGlobalMonitorForEvents` across `src/`, filtered to panel dismissal). Four more sites dismiss on an outside click — `RadialMenuService.swift:496`, `RadialNowPlayingService.swift:121`, `QRResultController.swift:121`, `RecentCaptureService.swift:165` — and none accepts typed input, so I don't think any is reachable this way. `ShelfService.swift:411` tracks a drag and `ScreenshotPinController.swift:56` wants an Option-click inside a pin's own frame; neither is the shape. Recording it so the next pass doesn't re-derive it.

**Two comments still say the opposite of the code.** `AssistiveKeyboard.swift:16-19` reads "the pid is cached and kept current from workspace notifications, so callers on an event tap pay one lock and a nil test rather than a window enumeration" — that is the version hardware disproved, and `livePID`'s doc just below contradicts it outright. It is the first thing anyone reads in the file, and it is exactly the belief that makes verify-on-use look redundant. `TextSnippetService.swift:233` has the other half: "That check costs a nil test unless that keyboard is actually running." The not-running path is now a lock, a `kill`, and a background `runningApplications(withBundleIdentifier:)` up to once per 2s for as long as the user keeps clicking, on every machine. Right trade given the notifications don't fire, but not what the comment promises. Neither blocks anything.

**Impact:** none from this diff. The popover is a pre-existing gap, not a regression — it belongs with #922's other call sites.
**Merge:** ship it — the helper and the snippet tap read correctly, and #922 is rebased on this.

ref #896

PR #899, opened 2026-08-24
fix(fan-control): black out the Fan Control feature when a controllable fan is not detected on the Mac

Yahddyyp:
Disables Fan Control in the Features menu on Macs without a controllable fan for example on a Macbook Air.

image

## Changes
- Adds `FanControlHardware.hasControllableFan`, backed by `discoverControlledFans()`, to detect whether the Mac has a fan that can actually be controlled.
- Adds `AppFeature.isHardwareSupported`, with hardware-specific feature support kept in `FanControlFeatureSupport.swift` rather than `FeatureCatalog.swift`.
- Adds `AppFeature.hardwareUnsupportedReason`, so the generic Features row does not contain a Fan Control-specific message.
- Greys out and desaturates the Fan Control row when it is not installed and the hardware is unsupported.
- Disables the `Install` button in that case, with the unsupported reason shown through a `.help()` tooltip.
- Leaves installed rows fully normal regardless of hardware support: settings remain reachable and Uninstall stays available. An install that predates this check is never silently stranded.
- Adds a comment explaining why `.help()` is attached to an `HStack` wrapper instead of the disabled `Install` button.

## Expected Behavior

| State | Row | Install | Uninstall | Settings |
|---|---|---|---|---|
| Fan-capable Mac | Normal | Enabled | — | Reachable |
| No fan, not installed | Greyed, tooltip | Disabled | — | — |
| No fan, already installed | Normal | — | Enabled | Reachable |


willn-dev:
This is a smart addition. Less clutter.

PathGao:
The behaviour is right and the caching is the correct call. What is worth changing is where the decision lives.

`FeatureHubRow` now asks the question as a hardcoded identity check:

```swift
private var isUnsupportedHardware: Bool {
    feature == .fanControl && !FanControlHardware.hasControllableFan
}
```

Fan Control is not the only feature that a given Mac cannot run, so the next one repeats this special case in the same view, and the two will drift. A property on the feature itself — something the `Feature` type answers, with `.fanControl` returning `FanControlHardware.hasControllableFan` and everything else returning true — puts the answer next to the feature that knows it, and the row just reads it.

That is what the `incomplete` label is about; the greying, the disabled install button and the `static let` cache are all fine as they are.

PathGao:
Three things, and the first one means the feature as written never turns on.

**The SMC key is `FNum`, not `FNUM`.** `FanControlHardware.swift:435` asks for `client.key(named: "FNUM")`. `SMCClient.key(named:)` packs the string into a four-char code byte by byte (`SMCClient.swift:118-120` calling `fourCC` at `:155-157`) — no case folding, so `"FNUM"` is a different key from `"FNum"` and the SMC returns nothing for it. Both existing readers in the repo use the correct spelling: `FanControlHardware.swift:249` and `SystemMonitor.swift:911`. As written, `hasControllableFan` is `false` on every Mac, so Fan Control is greyed out and permanently uninstallable for everyone. I could not run this on hardware, but the two spellings are side by side in the source. Fix the literal.

**Don't add a third FNum read, and "has fans" is not "has a controllable fan".** `FanControlHardware.fanCount()` (`:248-258`) is already exactly this read plus the `FanControlPolicy.fanCount(from:)` parse, and it distinguishes `value == 0` (`.noFans`) from an unparseable count (`.unsupported`); `SystemMonitor.fanTelemetryCount` (`:909-921`) is the second copy. The new `static let` is the third, and it answers a weaker question than its name claims: the actual controllability test is `discoverControlledFans()` (`:262-283`), which throws `.unsupported` when any of `F<i>Mn` / `F<i>Mx` / `F<i>Tg` / the mode key is missing, when `mode.dataSize != 1`, or when `FanControlPolicy.validBounds` rejects the range. A Mac that reports `FNum >= 1` with read-only fans passes your check and still cannot be controlled — which is the case the row is meant to catch. Back `hasControllableFan` with `discoverControlledFans()` on the shared hardware instance instead, keeping the same one-shot caching.

**The disabled button never says why.** `FeatureHubSettings.swift:276-281` disables Install with no `.help()` and leaves `accessibilityLabel` as `"\(hub.installButton) \(accessibilityTitle)"`, so a VoiceOver user gets "Install Fan Control" on a control that does nothing, and a sighted user gets a greyed row with no reason. Add a string ("no controllable fan on this Mac") to `FeatureHubStrings` and put it on both `.help()` and the accessibility label — it is the part of this change that carries the information.

One smaller thing: `Defaults.migrateFanControlVisibility` (`Defaults.swift:1290-1300`) can already have set `fanControl.availabilityKey` to `true` on an upgrade. On a Mac that fails the check, `FeatureHubSettings.swift:249` then suppresses the settings link while the feature stays installed and running. Worth deciding whether that state should uninstall itself or keep its settings reachable.

Yahddyyp:
> One smaller thing: `Defaults.migrateFanControlVisibility` (`Defaults.swift:1290-1300`) can already have set `fanControl.availabilityKey` to `true` on an upgrade. On a Mac that fails the check, `FeatureHubSettings.swift:249` then suppresses the settings link while the feature stays installed and running. Worth deciding whether that state should uninstall itself or keep its settings reachable.

@PathGao Which one do you feel is best, the user can always uninstall the feature and if we let it be it preserves their config exactly, but uninstalling can also be useful

PathGao:
Neither. Don't uninstall, and don't suppress the settings link either.

Auto-uninstall mutates user state on the word of a hardware check that has already proved it can be wrong: with the `FNUM` spelling, `hasControllableFan` is `false` on every Mac, and that same code with auto-uninstall wired in would have silently removed Fan Control and its config for everyone on upgrade. A detection bug that greys out a row is a bug report. A detection bug that deletes config is a bug report plus lost config. Fail soft.

But the state I described is worse than "keep their config". `FeatureHubSettings.swift:249` hides the settings link while the feature stays installed and running, so the user has a feature that is doing something and no way to look at it. Drop `!isUnsupportedHardware` from that condition and gate only the Install button:

- installed: row behaves normally, settings reachable, Uninstall available
- not installed: greyed, Install disabled with the help text

That removes the fifth bullet from your Changes list, and it makes the migration case harmless without adding any new state transition.

**Impact:** as written, a Mac that fails the check with Fan Control already installed keeps it running with its settings unreachable.
**Merge:** needs the `FNum` spelling fix first.

Yahddyyp:
@PathGao I think i have implemented all the change you requested, could you check it out now.

PathGao:
The three blockers are gone: no `FNUM` literal, no third fan-count read, and the settings link is ungated. One thing left, and it's the one I asked for last time.

**The unsupported-reason string is still fan-specific and still lives in the row.** You moved the predicate onto `AppFeature.isHardwareSupported`, which was the point, but `FeatureHubSettings.swift:280` and `:282` hardcode `FeatureStrings.fanControl(l10n.language).noFans` in a row that renders every feature. `isHardwareSupported` has `default: return true`, so nothing else reaches it today; the second feature that returns `false` shows a fan message. The string has to come from where the predicate came from — a `FeatureHubStrings` entry, or a per-feature reason sitting next to `isHardwareSupported`.

**And check what `noFans` actually says.** It isn't added in this diff, so it's an existing string written for somewhere else. `discoverControlledFans()` throws `.unsupported` for a Mac that reports fans it cannot control: `F<n>Mn` / `F<n>Mx` / `F<n>Tg` missing, `mode.dataSize != 1`, or `validBounds` rejecting the range. That machine has fans. If the string reads "no fans detected", it is wrong for exactly the case you strengthened the check to catch. `FeatureStrings` isn't in the diff, so this is a look, not a claim.

Smaller: the `HStack(spacing: 0)` at `:275` is doing real work — `.help()` doesn't fire on a disabled control, so it has to sit on a wrapper. It reads like leftover nesting, and whoever flattens it will quietly lose the tooltip. Worth one comment line.

**Impact:** today, a tooltip that may describe the wrong condition. The generic row holding a fan-specific string is the drift I flagged in the first pass, reappearing one layer up.
**Merge:** needs the unsupported-reason string moved off `FeatureStrings.fanControl` first.

Yahddyyp:
@PathGao Please do the honors

PathGao:
The string blocker is fixed: `hardwareUnsupportedReason` sits next to `isHardwareSupported` and the row reads it generically. Two things left.

**The two switches over `AppFeature` can drift, and the compiler won't catch it.** `isHardwareSupported` and `hardwareUnsupportedReason` both switch on `self` with a `default`. Add a feature to the first and forget the second and you get `.help("")` plus an accessibility label ending in `". "` — a disabled button with no reason, which is the state this PR exists to prevent. Make the reason the primitive and derive the bool:

```swift
var hardwareUnsupportedReason: String? {
    switch self {
    case .fanControl where !FanControlHardware.hasControllableFan:
        return FeatureStrings.fanControl(L10n.shared.language).noFans
    default:
        return nil
    }
}

var isHardwareSupported: Bool { hardwareUnsupportedReason == nil }
```

One place to add a feature, and "unsupported with no reason" stops being representable.

**`noFans` wording — still open from the last pass.** `FeatureStrings` still isn't in the diff, so I haven't seen the text. `discoverControlledFans()` throws `.unsupported` for a Mac that has fans it cannot control: `F<n>Mn` / `F<n>Mx` / `F<n>Tg` missing, `mode.dataSize != 1`, or `validBounds` rejecting the range. If the string reads "no fans detected", it is wrong for exactly the machine you strengthened the check to catch, and that tooltip is the only thing this change tells the user. One look.

Smaller: `hardwareUnsupportedReason` reads `L10n.shared.language`, while every other string in that row takes the language from the environment `l10n` (`feature.hubTitle(l10n.s, hub: hub)`). Fine as long as they are the same instance — I can't see `L10n`, so this is an inconsistency I'm noting, not a bug I'm claiming.

Merging isn't mine to do, I only comment.

**Impact:** none today — `default: return true` keeps every other feature out of both switches. The risk is the next feature that opts in.
**Merge:** needs the `noFans` wording confirmed first.

PathGao:
I've taken this over rather than send it round a fourth time — #988, with you as co-author on the commit.

What carried over unchanged: `hasControllableFan` backed by `discoverControlledFans()`, the fail-soft decision to leave installed rows alone, the greyed row, the disabled Install button, and the `HStack` wrapper with its comment. Those were all right, and the last round genuinely cleared the three blockers.

What moved is the layer. `isHardwareSupported` was read by `FeatureHubRow` only, and there are three places that write availability:

| Surface | Writes via |
|---|---|
| Hub row Install | `FeatureRuntime.setAvailable` |
| Hub **Install All** | `FeatureRuntime.setAllAvailable` |
| First-run picker | `FeatureRuntime.replaceAvailable` |

`FeatureHubSettings.swift:41` is the install-all button, about 200 lines above the row it walks around, and `OnboardingView.swift:252` offers Fan Control on a fanless Mac with an ordinary tick box. Both then combine badly with the fail-soft rule we agreed on: once Fan Control arrives by either route the row looks normal forever, so the user gets a feature that does nothing and no explanation. The gate now sits in `FeatureRuntime`, which all three go through.

Two smaller things came with it. `hardwareUnsupportedReason` is now the single switch and `isHardwareSupported` derives from it, so the two can't drift into an empty tooltip — that was the last objection on this PR. And `installableCount` replaces `AppFeature.allCases.count` in the hub's tally, because with a gate in place the install-all button's disabled condition can never be met on a Mac missing hardware.

On the part you couldn't test: you don't need a Mac with a fan to believe the positive case. `discoverControlledFans()` is the same call `readOnlySnapshot()` makes for everyone using Fan Control today — if it came back empty on fan hardware, the feature would already be broken for all of them. I did check the negative case here on an M5 Air: 2289 SMC keys enumerate and not one starts with `F`, so it's a real absence rather than a silent read failure like `FNUM` was.

Leaving this open — closing it or repointing the branch is yours. Thanks for the detection work; the SMC half of it is intact in #988.

Issue #887, opened 2026-08-24
App freezes using command bar


jack-cloudtech:
### Description

Command bar freezes entire app 

### Steps to Reproduce

1. Open Command bar 
2. Press enter
3. App freezes

### Vorssaint Version

3.3.2

### macOS Version

macOS 15 (Sequoia)

joaobernardo01:
I’m experiencing the same issue, but I can’t reproduce it by following your steps. It’s quite random, but the app completely crashes, leaving me unable to interact with it through any shortcut or menu.

alexis-morain:
same here

PathGao:
Tracking PR: #903 (open, `ready to merge`) — keeps the Command bar responsive when the shared pasteboard lane stalls. Not merged yet.

PathGao:
Tracking PR is now #998 (open). It replaces #903, which is closed. Not merged yet.

PathGao:
Ignore the pull request numbers I posted earlier in this thread — I was pointing at the wrong work. The note just above is the accurate one.

Nothing has changed on your Mac yet. Wait for the comment here that names a version to install, then tell us whether the freeze is gone.

jack-cloudtech:
It was quite random as @PathGao suggested. The latest update seem to have fix that and have been pretty stable. Thanks.

PR #903, opened 2026-08-24
fix(command-bar): keep Command bar responsive when the shared pasteboard lane stalls

atomsbaza:
## Summary

Ref #887 (App freezes using command bar).

Pressing Return on a row whose action touches the general pasteboard could freeze the whole app. The freeze needs two things at once:

1. A read on the serial `GeneralPasteboardAccess` lane that stalls — most often the clipboard history poll reading from an app that put promised (lazily rendered) content on the pasteboard and stopped answering.
2. A command-bar action that hops synchronously from the main thread onto that same lane, waiting behind the stuck read forever.

Reproduced end to end with a stalled pasteboard provider: open bar → Return → app hard-frozen, main thread parked in `_dispatch_sync_f_slow`.

## What changed

- `GeneralPasteboardAccess` gains `async(_:then:)`: work runs on the lane, the result is handed back on the main queue, and the calling thread never blocks.
- All four main-thread call sites reachable from Command bar moved onto it:
  - answer rows (`copyAnswer`) — the exact repro path
  - saved links' `{clipboard}` read in `open(_:)`
  - clipboard history writes (`writeToPasteboard` now completion-based; quick-panel copies and the bar's paste row updated)
  - URL cleaner's copy
- HUD confirmations and failure beeps keep their meaning by moving with the work they describe instead of racing it.

## Verification

- Reproduced the freeze on the pre-fix build; confirmed gone on this build under identical conditions (bar closes, main thread idle in the runloop while the lane stays wedged).
- Confirmed the queued copy still lands correctly once the stall clears (pasteboard contains the written value).
- `./build.sh --dev` builds clean; `./build.sh --test` passes 8316 checks.


PathGao:
Moving the pasteboard work off the main thread is the right call and the four `GeneralPasteboardAccess.shared.sync` call sites are all converted. Three things to settle before it merges.

**1. One call site was missed, and it is the same expression.** `CommandBarCatalog.swift:1534` — `cleanClipboardURL()` still does `NSPasteboard.general.string(forType: .string)` synchronously on the main thread, outside the lane entirely. It is a Command Bar row registered at `CommandBarCatalog.swift:604`, and it reads the general pasteboard exactly like the site this PR just moved at `:1236`. A stalled promised-flavor provider freezes the bar on that row exactly as before.

**2. `lastChangeCount` is now written from two queues.** `performWrite` assigns it at `ClipboardHistoryService.swift:156` and `:186` on the lane's utility queue. Every other write is on the main queue — `ignoreNextChange` at `:94`, `baselineChangeCount` at `:536`, the `captureIfChanged` completion at `:578`. Under `main` this was safe for a boring reason: `DispatchQueue.sync` ran the closure on the calling thread, which was main. It no longer does.

Besides the race itself, the ordering matters: the copy's own change-count bump can now land *after* `captureIfChanged` reads it, so `guard changeCount > self.lastChangeCount` passes and the app re-captures its own write as a fresh copy. Returning the change count from `performWrite` and assigning it in the main-queue completion closes both.

**3. AppKit work moved off the main thread without the PR saying so.** For the same `DispatchQueue.sync` reason, `richBatchAttributedString` (`NSTextAttachment(fileWrapper:)`, RTFD via `writeObjects([rich])`) and `NSBitmapImageRep(data:).tiffRepresentation` used to run on main and now run on the utility queue. That may be fine, but it is a change nobody asked for in a PR about responsiveness, and it should be stated and checked rather than inherited.

Smaller:

- In `runLink`, `defer { service.hide() }` moved from the synchronous body into the lane completion, so the bar now stays on screen until the stalled lane answers — the inverse of what the comment above it assumes.
- `GeneralPasteboardAccess.sync` has no callers left after this. Delete it or say why it stays.

PathGao:
Three more things on top of what is already noted above, all of them sites of the shape this PR is fixing rather than new problems it creates.

**The missed call site is not one, it is five.** `CommandBarCatalog.swift:1534` was named earlier; `NSPasteboard.general.string(forType: .string)` runs synchronously on the main thread in four more places, and two of them are on a keystroke path:

- `Services/Snippets/TextSnippetService.swift:322` — inside the `post` closure that resolves a snippet's `{clipboard}` token. The tap thread is parked in `DispatchQueue.main.sync(execute: post)` at `TextSnippetService.swift:335` while this runs, so a provider that promised a flavour and went quiet blocks the event tap, not just the UI, and the tap gets timed out and re-enabled at `:216`.
- `Services/Snippets/SnippetLibraryService.swift:170` — same expression, same token, library insertion path.
- `UI/MenuPanel/PanelURLCleanerView.swift:132` and `UI/Settings/URLCleanerSettings.swift:311` — both fill a field from the clipboard on a button press.

`GeneralPasteboardAccess.shared.async(_:then:)` as added in this PR is exactly the helper those four want. Whether they belong in this PR or a follow-up is your call, but the reason given for converting the other four applies to them unchanged.

**`PanelClipboardView.swift:215` now claims the copy before it happens.**

```swift
history.copy(entry)
copiedID = entry.id
```

`copy` no longer returns and no longer completes before the next line, so the button flips to "Copied ✓" while the write is still queued behind whatever is holding the lane — the precise stall this PR exists to survive. `copyAnswer` in `CommandBarCatalog.swift` already moved its HUD into the completion; do the same here: `history.copy(entry) { ok in if ok { copiedID = entry.id } }`.

**Two of the four quick-entry entry points were left in the old shape.** `copyQuickEntry` (`ClipboardHistoryService.swift:455`) and `copyQuickEntries` were converted; `copyOnlyQuickEntry` (`:477`) and `copyOnlyQuickEntries` (`:483`) still call `copy(...)` and then run `hideHistoryWindow()` / `pasteTargetApp = nil` as if the write had landed. No paste follows them so nothing corrupts, but the window now closes and focus returns before the clipboard holds the entry, and a failed write is silent. Either give them the completion too or say in the code why these two do not need it.

PathGao:
Can you fix the problems that I menthioned？Thanks for your fix，love it。

PathGao:
Thanks for this — the diagnosis was right and it is what #998 is built on, so your commit is in it as a co-author.

I picked the work up rather than sending another round of review notes, because two of the points above (`lastChangeCount` ownership, and the AppKit work that moved off the main thread only because `DispatchQueue.sync` used to run on the calling thread) turned out to change the shape of the write path rather than a line or two of it. #998 keeps your `async(_:then:)`, deletes `sync` outright so nothing can wait on the lane again, resolves each write before the hop so the RTFD and TIFF work stays on main, and picks up the sites named in the second comment.

Closing this one in favour of #998. Thanks again.

PR #905, opened 2026-08-24
feat(monitor): add battery temperature alerts

ywu73:
## Summary

- Add an opt-in alert for sustained high battery temperature.
- Preserve real battery temperature sample timestamps so cached readings cannot trigger alerts.
- Add configurable thresholds, localized UI text, documentation, and regression tests.

## Motivation

Vorssaint already displays battery temperature, but it does not notify users when the reading remains above a chosen threshold. This change adds that alert while reusing the existing read-only sensor sampler.

## Changes

- Add a disabled-by-default battery temperature alert with a configurable 30–50 °C threshold and a 35 °C default.
- Require the threshold to remain exceeded for about 12 seconds before notifying.
- Preserve real battery temperature sample timestamps during caching.
- Request Monitor notification permission when this alert is enabled.
- Add settings and notification strings for all supported languages.
- Document battery temperature notifications in the README.
- Add regression coverage for defaults, permissions, sampling, timestamps, and alert behavior.

## Testing

- Ran `./build.sh --test`: `TESTS OK (8338 checks)`
- Ran `./build.sh`: release bundle built successfully
- Ran `./build/Vorssaint --selftest`: `SELFTEST OK`
- Ran `git diff --check`

## Risks

- Low risk; the alert is opt-in and uses the existing read-only battery temperature sensor path.
- Battery temperature availability depends on the SMC sensor data exposed by the Mac.
- Review should confirm the localized wording and the 30–50 °C threshold range.

## Screenshots / Recordings

- Not provided; the settings UI adds an opt-in battery temperature alert toggle and threshold control.

## Checklist

- [x] Documentation has been updated where appropriate


PathGao:
I walked every site the CPU temperature alert occupies in `main` and each one has a battery twin here — the key and registered default, `monitorAlertPairs` and the `.notifications` case in `FeatureCatalog`, the `alertOn` branch and gate reset in `MonitorAlertService`, the sampling plan, the toggle, stepper, `onChange` and `sanitizeAlertValues` in the controls, and all thirteen locales in `FeatureStrings`. The no-battery guard is real too: `PowerSampler.hasInternalBattery` at `MonitorAlertsControls.swift:62` and `:120`, `SystemMonitor.swift:454`, and the `.monitorPower` clause in `alertOn`. Backup travels for free because `SettingsBackupSupport.exportKeys()` is derived from `Defaults.registeredDefaults`. Three things about the numbers, not the wiring.

**35 °C is a temperature the battery reaches doing nothing unusual.** `Defaults.swift:1035`, `MonitorAlertService.swift:175` and `MonitorAlertsControls.swift:65` all default this to 35. A MacBook battery sits in the low thirties at idle and goes past 35 while charging or under any sustained load, so the shipped default fires on an ordinary afternoon and keeps firing every cooldown period until the machine cools. Every other alert in this section defaults to a condition that is genuinely unusual — 90% CPU, 90 °C on the die, under 10% disk, under 15% charge. This one does not. Something in the low-to-mid forties is the first reading that means anything. I have no way to measure a real battery from here, so treat the specific number as a suggestion, but the shape of the problem does not depend on the exact figure.

**The floor of the range has the same issue.** `in: 30...50` at `MonitorAlertsControls.swift:67`, repeated at `MonitorAlertService.swift:176`. 30 °C is below where many batteries idle, so it is a selectable setting that guarantees a permanent alert. The CPU temperature stepper's floor of 70 °C was chosen to sit above idle; picking the battery floor on the same principle puts it around 35, not 30.

**The two temperature steppers in one section read differently.** `FeatureStrings.swift:1943` gives `cpuTemperatureThreshold: "Temperature above"`; `:1963` gives `batteryTemperatureThreshold: "Notify above"`. Rendered they stack as "Temperature above 90 °C" and "Notify above 35 °C" a few rows apart. Reuse the existing wording in all thirteen locales.

One layout note: the new toggle goes in at `MonitorAlertsControls.swift:63`, ahead of `Toggle(text.battery, ...)` at `:69`, but the key sits after `monitorAlertCPUTemperature` in `DefaultsKey`, in the registered defaults and in `monitorAlertPairs` — so the declared order and the order on screen disagree. Moving the toggle below the low-battery one lines them back up.

Issue #475, opened 2026-08-04
[Feature]: Window Pin


kDomingos:
### Problem or Motivation

Sometimes when we're using several apps and we need the floating calculator or another window as a reference, but you need to edit or click on a second app or a window.

### Proposed Solution

It would be great if there was a pin window (multiple if possible) feature. 

### Feature Area

Windows and Dock

### Mockups / References

FloatyTool ( but done better)

PowerToys Always on Top Utility ( although it's for Windows)

asyba:
Yes please do this!

AB-boi:
duplicate of #461

mrevanzak:
Same request as #461, so I'm doing it there.

I'll mention this issue on the PR too.

PathGao:
@kDomingos closing this as a duplicate of #461 so the feature has one thread. #461 is where it is actually being built — @mrevanzak is on it, and the plan there does include pinning more than one window.

Everything from here is copied over so it does not get lost: multiple pins, the floating-calculator use case, and the FloatyTool and PowerToys references — https://github.com/vorssaintapp/vorssaint-utils/issues/461#issuecomment-5404732337

If I missed anything you meant, add it on #461.

Issue #461, opened 2026-08-03
Pin the current window on top


niraj-khatiwada:
### Problem or Motivation

Hi, thanks for creating Vorssaint. I love using it. One feature that I think is very useful but lacking is Always On Top. As the name suggest, it would be great if we could pin the currently focused window (or any open windows) on top. Actually, I use a different app for that and thought this app would definitely have that feature. Nevertheless, I'm sure it's in the works and will land soon. Kudos to you for dedicating your time and resources for making such huge app free. 

### Proposed Solution

[Not sure what to place here]

### Feature Area

Windows and Dock

### Mockups / References

https://www.floatytool.com/

mi-mm:
hi, i believe this is already possible to some extent with the "Dock Preview" feature. After hovering over an open app in the dock and clicking the pin icon, a preview of the selected app stays on top of all other windows :)

niraj-khatiwada:
> hi, i believe this is already possible to some extent with the "Dock Preview" feature. After hovering over an open app in the dock and clicking the pin icon, a preview of the selected app stays on top of all other windows :)

that actually pins the static dock preview not the app window. i'm talking about pinning the actual app window itself.

AB-boi:
@niraj-khatiwada I know what you mean. I am interested in this feature as well. The idea is to pin a window to the top (z-axis) so that it stays in front of all the other windows. In addition, the ability to alter opacity of these windows will also be nice.

https://github.com/itsabhishekolkha/AlwaysOnTop A FOSS implementation for reference

mrevanzak:
Grabbing this.

Hotkey that pins the window you're in so it stays on top. PowerToys has this on Windows, we don't. You can pin more than one. Border and sound if you want them, plus an exclude list. Pins go away when Vorssaint restarts.

#475 is asking for the same thing. I'll mention both on the PR.

PathGao:
Folding #475 in here — same request (pin a window so it stays in front), and @AB-boi already called it there.

Carried over from #475 so it does not get lost with the thread:

- **More than one window pinned at once.** @kDomingos asked for "multiple if possible". @mrevanzak's plan above already allows that, so this is a second reporter confirming the requirement rather than a new ask.
- **Use case:** keeping a floating calculator or another reference window visible while clicking and typing in a different app.
- **References:** FloatyTool (already linked in this issue's body) and **PowerToys' Always on Top** utility on Windows.
- @asyba on #475 is a +1.

[...]

Issue #838, opened 2026-08-23
Enhancement Summary: all 165 open enhancements — mislabeled, already shipped, worth doing, out of scope


PathGao:
I went through all 165 open enhancements. Baseline is `main` @ `5c537230`. The CHANGELOG still carries an `[Unreleased]` block at the top, so nothing merged since the last stable tag is on anyone's Mac yet.

Last re-checked 2026-09-02T15:04:20Z. The previous pass was 2026-08-29T16:27:44Z against `71af73fc`; 192 files changed between the two, so every call that touches the switcher, window layout, mouse buttons, keyboard, screenshots, the shelf, the clipboard, brightness, the cleaner and the Now Playing bridge was re-read against the current tree rather than carried over.

Six branches closed in the last four days — the menu bar organiser, its follow-up, the display bundle, the HiDPI branch stacked on it, the media mini player and the audio auto-pause. Nothing in this document is waiting on a review queue any more except #489. Three separate things caused those closures, and they are the biggest change to this register: the platform removed the API #481 stood on, a private-framework policy call is now blocking four display threads, and a rule landed that a feature its author cannot run on real hardware is not accepted.

Companion writeups: #836 Bug Summary, #837 PR Summary, and #977 Command Bar Coverage — #977 and this issue are both on this register too. Under the current rule — fix before feat — an enhancement caps at `priority: medium`: severity and reach are scored separately and only two highs together reach `high`, which a feature almost never does against a bug that leaves something unusable. There is an escape hatch, but it has to be argued in a comment on that thread. #977 is the one thread here carrying `high`, and it is the maintainer's own index thread rather than a user request; no such comment has been posted on it.

## Disposition at a glance

| Action | Issues |
|---|---|
| Split a bug out of it | 1108, 825, 710 |
| Labelled fixed, only half of it is | 956, 768 |
| Build it, the change is named and small | 1274, 1262, 1257, 1227, 1199, 1187, 1180, 1157, 1118, 1108, 1107, 1069, 1063, 1053, 1049, 933, 907, 900, 894, 893, 885, 867, 829, 815, 808, 798, 779, 778, 741, 734, 733, 630, 616, 604, 603, 598, 591, 576, 566, 532, 530, 514, 497, 389, 379, 344, 306 |
[...]

PR #906, opened 2026-08-24
feat(always-on-top): add Always On Top window pin

mrevanzak:
## Summary

Adds an opt-in Always On Top feature so a shortcut can pin the frontmost window above other windows. Same idea as PowerToys Always On Top, built for macOS.

Ref #461. Also covers #475.

## Behavior before

There was no way to pin a real window on top. Dock Preview can pin a preview card, not the window itself.

## Behavior after

- New Windows-group feature, not installed by default, not in the Windows preset.
- Needs Accessibility. Shortcut defaults to ⌃⌥⇧T and is remappable.
- Toggle pins or unpins the frontmost window. Several windows can stay pinned at once.
- Optional cyan border, optional pin sound, exclude list by exact bundle ID.
- Pins live in memory only. Feature off, uninstall, or a Vorssaint restart unpins everything.
- Best-effort private SkyLight implementation. Pinning is successful only when SLS readback confirms the requested level; cross-app pinning is currently unsupported on macOS Tahoe.

## Verification

- `./build.sh --test`: TESTS OK (8337 checks)
- Manual Safari pin on a signed Tahoe build: SLS set returns success, independent readback stays at normal level. Feature does not pin other apps' windows on macOS Tahoe.

## Notes

Keep as draft until pinning is validated on a supported older macOS and gated accordingly. Fullscreen, other topmost windows, and Screen Sharing may not stay above everything.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,11 +81,10 @@ A few conventions to keep in mind.
 
 ## Strings and translations
 
-Shared chrome strings live in `Core/Localization.swift` as fields of the
-`Strings` struct. Feature-specific copy lives in a `*Strings.swift` file under
-`Core/` and is exposed through `FeatureStrings`. Adding a field forces **every**
-supported language to provide it, and the compiler is the completeness check,
-so a translation can never silently fall out of sync.
+Every user facing string lives in `Core/Localization.swift` as a field of the
+`Strings` struct. Adding a field forces **every** supported language to provide
+it, and the compiler is the completeness check, so a translation can never
+silently fall out of sync.
 
 Vorssaint ships these locales today: English (US), Português (Brasil),
 Türkçe, Русский, Español, Deutsch, Français, Italiano, 日本語, 한국어, 简体中文,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,10 +81,11 @@ A few conventions to keep in mind.
 
 ## Strings and translations
 
-Every user facing string lives in `Core/Localization.swift` as a field of the
-`Strings` struct. Adding a field forces **every** supported language to provide
-it, and the compiler is the completeness check, so a translation can never
-silently fall out of sync.
+Shared chrome strings live in `Core/Localization.swift` as fields of the
+`Strings` struct. Feature-specific copy lives in a `*Strings.swift` file under
+`Core/` and is exposed through `FeatureStrings`. Adding a field forces **every**
+supported language to provide it, and the compiler is the completeness check,
+so a translation can never silently fall out of sync.
 
 Vorssaint ships these locales today: English (US), Português (Brasil),
 Türkçe, Русский, Español, Deutsch, Français, Italiano, 日本語, 한국어, 简体中文,
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -232,6 +232,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         URLCleanerService.shared.stop()
         FocusFollowsMouseService.shared.stop()
         WindowMaximizer.shared.stop()
+        AlwaysOnTopService.shared.stop()
         WindowLayoutService.shared.suspend()
         KeyboardDebounceService.shared.suspend()
         TextSnippetService.shared.suspend()
--- a/Sources/Vorssaint/App/FeatureRuntime.swift
+++ b/Sources/Vorssaint/App/FeatureRuntime.swift
@@ -165,6 +165,7 @@ final class FeatureRuntime: ObservableObject {
             WindowLayoutService.shared.syncWithPreferences()
         },
         .autoQuit: { AutoQuitService.shared.syncWithPreferences() },
+        .alwaysOnTop: { AlwaysOnTopService.shared.syncWithPreferences() },
         .scrollInverter: { ScrollInverter.shared.syncWithPreferences() },
         .focusFollowsMouse: { FocusFollowsMouseService.shared.syncWithPreferences() },
         .smoothScroll: { SmoothScrollService.shared.syncWithPreferences() },
--- a/Sources/Vorssaint/Core/AlwaysOnTopStrings.swift
+++ b/Sources/Vorssaint/Core/AlwaysOnTopStrings.swift
@@ -1,0 +1,292 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+
+struct AlwaysOnTopFeatureStrings {
+    let title: String
+    let hubDescription: String
+    let enable: String
+    let enableCaption: String
+    let activeNow: String
+    let shortcut: String
+    let showBorder: String
+    let borderColor: String
+    let borderThickness: String
+    let playSound: String
+    let excludeTitle: String
+    let excludeEmpty: String
+    let addApp: String
+    let excludeCaption: String
+    let pinningUnavailable: String
+    let permissionRequired: String
+}
+
+extension FeatureStrings {
+    static func alwaysOnTop(_ language: AppLanguage) -> AlwaysOnTopFeatureStrings {
+        switch language {
+        case .enUS: return .enUS
+        case .ptBR: return .ptBR
+        case .tr: return .tr
+        case .ru: return .ru
+        case .es: return .es
+        case .de: return .de
+        case .fr: return .fr
+        case .it: return .it
+        case .ja: return .ja
+        case .ko: return .ko
+        case .zhHans: return .zhHans
+        case .zhTW: return .zhTW
+        case .zhHK: return .zhHK
+        }
+    }
+}
+
+extension AlwaysOnTopFeatureStrings {
+    static let enUS = AlwaysOnTopFeatureStrings(
+        title: "Always On Top",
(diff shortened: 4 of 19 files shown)


PathGao:
Draft, so this is early feedback rather than a review. The registration is unusually thorough — all five feature lists plus `FeaturePresets`, `GlobalShortcut`, `SettingsDirectory` and `SettingsView` are joined, which is more than most feature PRs manage. What is missing is the exit paths, and the consequence is bad enough to name first.

**A pinned window survives the app.** `SLSSetWindowLevel` writes into the WindowServer, not into Vorssaint. `AppDelegate.applicationWillTerminate` (`:216-240`) is the hand-maintained "put the system back before we die" list — `BrightnessService.restoreDisplaysBeforeTermination()`, `FanControlService.restoreBeforeTerminationIfNeeded()`, `SuperKeyService.suspend()` and so on. `AlwaysOnTopService.shared.unpinAll()` is not in it. Quit with a window pinned and that window floats above everything with no way to lower it short of the owning app recreating it.

`SelfUninstall.swift:63-99` has the same list for the app removing itself, and the same omission — there the pin is permanent, because there is no app left.

Three more:

- **The write is never read back.** `AlwaysOnTopPinning.swift:19` returns the original level as soon as `client.setLevel(...)` returns true, and `setLevel` only checks `setLevelFn(...) == 0`. `WindowLayoutService.verified(_:)` (`:461`, `:493`) is the precedent in this repo, and `client.level(of:)` is already available here, so the confirmation is one line. The test (`stub.levels[42] == AlwaysOnTopPinning.floatingLevel`) only proves the stub echoes what it was handed.
- **A destroyed window can leak its border.** `handleAX` resolves `kAXUIElementDestroyedNotification` by calling `AXWindowResolver.windowID(for: element)` — `_AXUIElementGetWindow` on an element that has just been destroyed. If that fails, `windowID` is nil, `unpin` never runs, and the map entry, the `NSPanel` and its 0.15s repeating timer live for the process lifetime. Capturing the window id at `watch(...)` time and keying the callback off it closes that.
- **`unpinExcluded()` is subscribed to `UserDefaults.didChangeNotification` with no key filter**, so every defaults write anywhere in the app runs an `NSRunningApplication(processIdentifier:)` lookup per pinned window on the main queue.

One naming point: `permissions == [.accessibility]` and the PR framing say Accessibility, but the pin itself is not an AX call — AX is only used to find and observe the window. Worth saying so in the description, because a reviewer checking "is every AX write confirmed" will look in the wrong place.

PathGao:
Still draft, so more early feedback. Three things below the exit-path problem already raised.

**The border sits at the same level as the window it outlines, so it disappears on the first click.** `AlwaysOnTopPinning.floatingLevel` is `NSWindow.Level.floating.rawValue` (`AlwaysOnTopPinning.swift:14`) and the border panel is `panel.level = .floating` (`AlwaysOnTopBorder.swift:32`). Same level means z-order is decided by whoever was ordered front last. `orderFrontRegardless()` wins at pin time; activating the pinned window puts it back in front, and since the stroke is drawn *inside* the window's own bounds (`AlwaysOnTopBorderView.draw`, `:116-121`, inset by `thickness / 2`) the whole border is then covered, not just dimmed. Derive both numbers from one constant with the border one level above whatever `pin` writes.

**`focusedAXWindow` is `WindowActivator.focusedWindowID` with the last line removed, and calling both opens a gap.** `AlwaysOnTopService.swift:182-190` is line-for-line `WindowActivator.swift:178-185`, minus the closing `AXWindowResolver.windowID(for:)` and minus the `Permissions.shared.accessibility` guard. `toggleFrontmost` calls `WindowActivator.focusedWindowID(for: pid)` at `:44` and then `focusedAXWindow(pid:)` at `:54` — two independent AX round trips asking the same question of the same app, with a `pin` and a defaults read in between. When focus moves between them, `watch(windowID:pid:element:)` (`:193`) files the second call's element under the first call's window id, and `handleAX` resolves ids from the element, so the destroy and miniaturize notifications arrive for a window that is not the one pinned: the pinned window is never unpinned and never gets its border updated. Resolve the element once and take the id from that element, which is exactly what `focusedWindowID` already does internally.

**The border is baked into screenshots and screen recordings.** `AlwaysOnTopBorder.swift:22-33` never sets `sharingType`, so the panel defaults to `.readWrite`. Every other always-visible overlay in this repo sets it: `BrightnessOSD.swift:116`, `ExtraBrightnessService.swift:296` and `:348`. This app ships its own capture path (`ScreenshotCaptureEngine`), so with the border on, every capture of a pinned window gets a cyan rectangle around it.

Two smaller ones:

- `pin` (`AlwaysOnTopPinning.swift:17-22`) records whatever level it reads as the "original" without checking it is below `floatingLevel`. Re-pinning a window a previous crashed run already raised records `floating` as the original, so unpinning restores it to `floating` and the window is stuck there for good. `guard original < Self.floatingLevel` makes the stored restore value always mean something.
- `collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary]` (`AlwaysOnTopBorder.swift:33`) puts the border on every Space, while nothing moves the pinned window itself, and `quartzBounds` uses `.optionIncludingWindow` (`:87`) which keeps returning bounds for a window that is on another Space. That reads like a border drawn over nothing after a Space switch, but it depends on how the raised level interacts with Spaces and I could not check it here — worth trying before the draft comes out.

PathGao:
Still draft. The ones from the last two passes are unchanged in the tree — not repeating them. What is new on this read:

**"Settings says so" is not payable by the check that backs it.** Body: *"If this Mac cannot set another app's window level, the shortcut does nothing and Settings says so."* The only gate on `strings.pinningUnavailable` is `service.pinningAvailable` → `Live.isAvailable`, which is `getLevel != nil && setLevelFn != nil` — a `dlsym` result. `SLSSetWindowLevel` and `SLSGetWindowLevel` resolve on every macOS this app runs on, so that Section never renders, and the case the sentence describes (the write rejected, or accepted and ignored) has no path to the UI at all. This is the missing readback from the other end: reading the level back after `pin` is what would make that sentence true. Until then the honest wording is "the shortcut may do nothing, and Settings will not say so."

**Six copies of the border default, holding two different colours.** `#00ADEF` is in `Defaults.swift:1102`, `AlwaysOnTopService.toggleFrontmost`, `AlwaysOnTopBorder.show(fallback:)`, `AlwaysOnTopSettings.borderColor` and `AlwaysOnTopSettings.colorBinding`; `AlwaysOnTopBorderView.color` is `.systemCyan`, a sixth site with a different value. Thickness has four (`4` registered, `?? 4`, `4.0`, `CGFloat = 4`), and the registered one is an `Int` while both readers want a `Double` — `object(forKey:) as? Double` against an `Int` is not a bridge I would build on. And since `Defaults` registers all of these, `string(forKey:) ?? "#00ADEF"` and `?? 4` can never fire: they are not fallbacks, they are a second source of truth that can drift without anything failing. Register `4.0`, delete both `??`s, and let `AlwaysOnTopBorderView` take its values from `show` only.

Same read site, related: colour, thickness and `showBorder` are read once inside `toggleFrontmost`, so changing any of them only affects the next pin. The body lists them as options on the feature, not on the pin.

**The border polls for something the observer is already attached to.** `startTimer` (`AlwaysOnTopBorder.swift:76-84`) runs `updateFrame` every 0.15s for the life of every pin, and each tick is a `CGWindowListCopyWindowInfo` call. `watch(windowID:pid:element:)` already holds an `AXObserver` on that exact element and already loops over a list of notification names — `kAXWindowMovedNotification` and `kAXWindowResizedNotification` are two more entries in that array plus an `updateFrame()` branch in `handleAX`. Whether that alone tracks a live drag depends on how the pinned app emits moves, which I cannot check here, so keep the timer if it has to be there — but start it on the first move and stop it once the drag goes quiet, instead of running it from pin to unpin.

Two shorter ones:

- `GlobalShortcutRole.alwaysOnTop` is joined to the role table (`storageKey`, `defaultShortcut`, `requiredEnableKeys`, `feature`), and the service separately hand-registers a Carbon hotkey under its own `'VUAT'` id. No dispatch site for the role appears anywhere in the diff. If whatever consumes `requiredEnableKeys` also calls `RegisterEventHotKey`, ⌃⌥⇧T ends up registered twice for one action. I cannot see that registrar from the diff — worth settling which half owns the registration before the draft comes out.
- The collision test names its two rivals by hand (`windowLayoutRightTwoThirdsDefault`, `screenOCRDefault`), so it proves this shortcut is clear of the two you thought of and nothing about the next one added. One assertion that `GlobalShortcutRole.allCases.map(\.defaultShortcut)` is distinct covers the whole class — assuming no two roles are meant to share a default, which I did not verify.

**Impact:** with no readback the feature can silently do nothing while Settings reports it fine; border settings changed after a pin look dead; the poll cost is per pinned window for as long as it stays pinned.
**Merge:** needs the level readback and the `SelfUninstall.swift` exit path first — both still open from the earlier passes.

ref #461

PathGao:
Closing this one. Two things decided it, neither about code quality.

The pin goes through `SLSSetWindowLevel` / `CGSSetWindowLevel` resolved with `dlsym`. That is a private call, and the enhancement summary (#838, under #461) says the policy call on private window-level APIs has to come before any branch. It has not been made.

Your own Verification section says the readback stays at the normal level and other apps' windows do not pin on Tahoe, so the private call is not doing the job even today.

The review points from 2026-08-24 and 2026-08-26 (quit and uninstall exit paths, readback, border level) stay on this thread. Keep the branch. If #461 gets a yes, this is where to restart from.

PR #912, opened 2026-08-24
Fix Homebrew popover dismissal and anchoring

pergioa:
## Summary

- dismiss the menu popover on outside clicks unless a package-changing Homebrew operation is actively running
- keep running-operation protection independent of the currently selected menu tab
- keep the parent popover alive while a Homebrew confirmation dialog is presented
- preserve the opening status-item anchor through Homebrew content resizes and AppKit's deferred placement pass
- correct large jumps to the status area's fallback corner while leaving ordinary menu-bar layout shifts to AppKit

## Behavior

Homebrew search, refresh, and browsing no longer prevent outside-click dismissal. Confirmation dialogs temporarily block dismissal and release the guard when dismissed. Install, uninstall, upgrade, upgrade-all, and Homebrew update operations keep the popover available until the operation succeeds, fails, or is cancelled, even if the user switches tabs.

## Testing

- Tests pass
- Self-check OK

## Demo

Before:

https://github.com/user-attachments/assets/50ed3078-5d06-4153-bd4b-b61d0659d8bc

After:

https://github.com/user-attachments/assets/75f7130f-7232-4682-9856-ca7f66a82831


PathGao:
The anchoring half is good. The dismissal half takes protection away from ten other panels.

**The regression.** `AppDelegate.swift:729` on `main` guards the global outside-click monitor with `!PanelInteractionState.shared.keepsPopoverOpen`. This PR replaces that guard with `!blocksOutsideDismissal` plus a hardcoded `HomebrewManager.shared.operationStatus?.isActive` check. Eleven views set `keepsPopoverOpen`; exactly one — `PanelHomebrewView` — sets `blocksOutsideDismissal`. So every other panel loses its outside-click protection:

`CleanerView.swift:908` (mid-scan or mid-removal), `PanelUninstallerView.swift:32` (mid-uninstall), `PanelAppUpdatesView.swift:22` (mid-install), `MediaWorkspaceView.swift:21`, `PanelClipboardView.swift:34`, `PanelRecentCapturesView.swift:12`, `PanelURLCleanerView.swift:24`, `PanelWindowLayoutView.swift:51`, `QuickTogglesSection.swift:27`, `MenuPanelView.swift:627` and `:987`.

`keepsPopoverOpen` still gates the Settings-overlap path (`:769`, `:785`) and `popoverShouldClose` (`:924`), so after this there are two flags answering "may the panel close?" that disagree, and nothing requires them to match.

There is also a design point: `PanelInteractionState` exists so a view declares its own state, and this wires `HomebrewManager.shared` straight into `AppDelegate`'s generic dismiss monitor. `JunkCleaner.phase` and `AppUninstaller.phase` are the same "long operation running" signal; if the busy-check lived in `PanelInteractionState`, all three would join it instead of one being special-cased in the delegate.

**On the anchoring half**, two things worth a second look:

- The new guard is `contentResized || jumpedToFallback || anchor.overridesSoundFrame || !frameStillDescribesMenuBar(anchor)`. `jumpedToFallback` alone already covers the bug the comment describes. Adding `contentResized` overrides the original case — "a healthy bar places the panel better than the anchor can, and keeps it under an icon that shifts as items come and go" — on **every** resize, so on a healthy bar whose icon has moved since the panel opened, any content resize snaps the panel back to the stale opening anchor.
- The extra `DispatchQueue.main.async` re-apply calls `setFrame(display: true)` one run-loop turn after AppKit's own placement. It cannot loop (the `didMove` re-entry hits the 2pt tolerance), but a second placement authority landing after AppKit's usually reads as a one-frame jump rather than a fix for one.

The `popoverAnchorDriftTolerance` extraction is right and catches both `34` literals.

Splitting this into two PRs would make both reviewable: the anchoring change is generic and looks correct, the dismissal change needs the other ten views to come with it.

pergioa:
Thank you for the review. I will abandon and close this PR, then open two independent PRs—one for the anchoring changes and one for the popover changes—to narrow the scope of each to a single concern.

Issue #837, opened 2026-08-23
PR Summary: all 52 open PRs — conflicts, merge order, what to close or cherry-pick


PathGao:
All 52 open PRs against `origin/main` @ `5c537230`, 2026-09-02. What to merge, in what order, and one row per PR for everything else. This issue covers open PRs only — nothing closed or already merged is listed. Conflict counts are measured with `git merge-tree origin/main <head>`. Completeness verdicts come from reading each change and checking the rest of `main` for the sibling places it should also have touched; nothing here was run on a Mac.

Companion issues: #836 Bug Summary, #838 Enhancement Summary.

**Since the previous revision** (2026-08-29, base `a2fe2216`): **the roster turned over almost completely.** Thirty-two branches arrived and thirty-seven left, and only twenty rows survive from the previous revision. A review sweep this morning closed most of the standing backlog, so what is left is a short tail of community feature branches plus a burst of small maintainer fixes. **The reader's job flipped with it: twenty-two branches now carry `ready to merge` where five did, and the question is no longer "who has read the push" but "in what order".** `main` moved under 183 files — the clipboard history service, the switcher, the recorder, every mouse service, the string table and the shared test file — so every carried verdict whose file moved was re-derived at this base rather than carried. Conflicts went from 25 to 16, and the largest is now four files where it was sixteen. Four verdicts changed: **#1114 and #1080 are now `complete`**, **#1038 is no longer `risky`** (the separator fix landed and a different ask replaced it), **#1033 is no longer `complete`** (its base was sent back), and **#773 is now `declined`** on a policy the maintainer wrote down and applied to three other branches this interval. The string table is 952 fields across 13 languages, re-counted at this base.

---

## Disposition at a glance

| Action | PRs |
|---|---|
| Merge, in the order below | 1014, 1100, 1114, 1186, 1213, 1260, 1266, 1268, 1270, 1295, 1296, 1298, 1299, 1300, 1301, 1302, 1304, 1305 |
| Labelled `ready to merge`, and the gap is still open | 694, 767, 1080, 1154, 1194, 1202 |
| Complete, blocked behind its own base | 1037 |
| A decision, not a review | 773, 1040, 1141, 1149, 1205 |
| Loses or corrupts something as written | 874 |
[...]

PR #913, opened 2026-08-24
feat: let a script link run without an argument

rhukster:
## Before

A script link only ran once something followed its name. Typing the bare name prefilled the field and waited:

```swift
let argument = CommandBarLinks.trailingArgument(query: typed, name: link.name) ?? ""
if argument.isEmpty {
    service.prefill(link.name + " ")
    return
}
```

That is right for a script that works on what you type — a currency conversion, a lookup. It is a dead end for one that works on the clipboard, the current selection or a file, where there is nothing to type and the wait never resolves. Today those scripts need a throwaway word typed after the name to run at all.

## After

A saved script can be marked to also answer to its bare name, with a new checkbox in the link editor beside the existing script hint. It is off by default, so nothing changes for any script already saved.

Opt-in rather than automatic, because a script runs when typing pauses rather than when Return is pressed. If a bare name always ran, finishing typing the name of a link called `deploy` while searching for something else would fire it. Requiring the person to say so keeps that impossible unless they asked for it.

A script marked this way still receives an argument when one is typed, so both `clean` and `clean code` work. The bare name has to match the whole query, so a script never runs off a prefix of its own name, nor off a longer word that starts with it — `cle` and `cleaner` do not run a script called `clean`.

## Saved shortcuts had to keep loading

`CommandBarLink` is now decoded a field at a time rather than by the synthesized initializer:

```swift
id = try container.decodeIfPresent(UUID.self, forKey: .id) ?? UUID()
```

Swift's synthesized `Decodable` throws on a missing key rather than falling back to the property's default value. Shortcuts are stored as one encoded list and `CommandBarLinks.decode` returns `[]` on any throw, so adding a key to the struct would have made every existing shortcut fail to decode — silently, and all of them at once, on first launch after the update. There is a test that loads a list written before the field existed.

## Testing

- `./build.sh` clean, `--selftest` passes, `./build.sh --test` passes.
- New assertions cover the bare-name match, the prefix and longer-word cases, the argument still arriving when typed, decoding a shortcut saved before the field existed, and a save/load round trip of the flag.
- Ran a local build with three real script links saved before this change: all three still load, the bare name does nothing until the box is ticked, and runs once it is.

The new string is in all 13 languages. One note for a maintainer: `scriptSearchHint` ("Type what to send after the name") is still shown as the row subtitle for a script marked this way. It is not wrong, since such a script can still take an argument, but you may prefer different wording there — I left it alone to keep this focused.


PathGao:
Read this at line level now. The idea and the migration are right; one call site did not come along.

**`CommandBarService.swift:1198` still calls `trailingArgument`.**

```swift
let matchingScriptIDs = Set(savedLinks.compactMap { link -> String? in
    guard link.kind == .script,
          CommandBarLinks.trailingArgument(query: trimmed, name: link.name) != nil
    else { return nil }
    return "link.\(link.id.uuidString)"
})
```

That set is what the dedup at `:1202` removes from the pool. For a bare-name query `trailingArgument` returns nil for every link, so the set is empty, nothing is removed — and once the script answers, `scriptAnswer` is appended at `:1283` **alongside the still-present plain link row**. Two rows for the same script.

It also means the overlapping-names tie-break this block exists for ("run" vs "run report") does not apply to bare-name matches at all. `scriptArgument(for:query:) != nil` is the call it wants.

Of the five `trailingArgument` sites, `:165` (`matchingScriptLink`) and `CommandBarCatalog.swift:1262` (`runScript`) are converted correctly, and `:155` (`rankingTitle`) and `CommandBarCatalog.swift:1223` (the non-script `open()` path) are right as they are — a bare name should rank against the name. So `:1198` is the only miss.

**One string is now false.** `CommandBarCatalog.swift:1194` gives every script row the subtitle `bar.scriptSearchHint` — "Type what to send after the name" — unconditionally. For a `runsWithoutArgument` script that instruction is wrong. The PR added a string for the settings toggle but no counterpart for the row.

Worth saying explicitly since it looks like defensive padding and is not: **the hand-written `init(from:)` is required.** Swift's synthesized decoder throws `keyNotFound` for a defaulted property, and `CommandBarLinks.decode` swallows the throw and returns `[]` — so without it, every saved shortcut would silently vanish on upgrade. Putting it in an extension to keep the memberwise init is the right call, and both migration and round-trip are tested.

Localization is complete (all 13 initializers, count 147→148). Labelling `incomplete` for `:1198`.

rhukster:
Both right, thanks — fixed in a533331.

**`CommandBarService.swift:1198`.** Confirmed: `trailingArgument` returns nil for every link on a bare-name query, so the set was empty, nothing was removed, and the answer row landed beside the plain row.

Rather than swapping the one call, I moved the question into `CommandBarLinks.matchingScriptLinks(in:query:)`, which both the pool dedup and `matchingScriptLink` now ask. The row that gets removed is decided by the same rule that decided the script would run, so the two can't drift apart again — and unlike `CommandBarService` it's in the test target, so it's covered. The overlapping-name tie-break reaches bare names now too; there's an assertion that "run" and "run report" are both dropped.

I checked the new assertions fail against the old predicate before keeping them — swapping `scriptArgument` back to `trailingArgument` inside `matchingScriptLinks` fails three of them.

**The subtitle.** Right, `scriptSearchHint` was unconditional. Bare-runnable rows get their own string, `scriptBareSearchHint` ("Runs on its own, or type what to send"), in all 13 initializers.

Also rebased, so the count assertion is 151 and the `Tests/MetricsTests.swift` conflict listed in #837 is resolved.

`./build.sh` clean, `--selftest` and `./build.sh --test` pass.

PathGao:
Both points are closed in a533331. `CommandBarLinks.swift:205` makes `matchingScriptLinks` the single rule, asked by the pool dedup at `CommandBarService.swift:1246` and by `matchingScriptLink` at `CommandBarLinks.swift:216`, so a bare-name query removes the script's own row and only the answer stands; the added assertion for "run" / "run report" pins the tie-break reaching bare names. `CommandBarCatalog.swift:1196` picks `scriptBareSearchHint` for a `runsWithoutArgument` row, so the subtitle no longer promises an argument.

Checked the rest of the shape. `CommandBarLinks.swift:178` (`rankingTitle`) and `CommandBarCatalog.swift:1223` (the non-script `open()` path) are right unchanged. `init(from:)` at `CommandBarLinks.swift:64` covers all four stored properties, so nothing is dropped on the round trip.

One observation, not a blocker. `CommandBarCatalog.swift:1262` is the fifth site of the question you just centralised and still asks `trailingArgument(...) ?? ""`. For every query that names the link it reaches the same argument `scriptArgument` would, because the `argument.isEmpty, !link.runsWithoutArgument` guard at `:1265` is what decides. It diverges only when the row was reached by a fuzzy name match — typing `cle` and pressing Return on a bare-runnable `clean` runs it with no argument, where `scriptArgument` would return nil. That is arguably what activating the row should do, so this is a call for you rather than a defect; if you agree it is, `scriptArgument(for:query:) ?? ""` at `:1262` would leave one function answering the question everywhere.

Nothing else outstanding.

PR #914, opened 2026-08-24
fix(quick-tools): keep the copy confirmation on screen for long text

rhukster:
## The problem

Both message panels in `QuickToolHUD` are laid out from `fittingSize` and then centred on the width that came back:

```swift
let size = host.view.fittingSize
panel.setFrame(NSRect(x: frame.midX - size.width / 2, ...
```

Neither message is bounded. The confirmation's message is usually whatever was just copied, so a copied paragraph — or one long line of one — measures far wider than the display and the panel hangs off both edges with nothing readable left.

Measured with `NSHostingController.fittingSize` against this exact view hierarchy, on a 2560pt screen:

| message | now | with this change |
|---|---|---|
| `#3A7BD5` | 110 × 37pt | 110 × 37pt |
| one long line of cleaned text | **5199 × 37pt** | 414 × 37pt |
| a multi-paragraph transcript | 354 × 198pt | 414 × 48pt |

`show(icon:message:)` has **68 call sites**. I went through the strings behind them: all are short and fixed except `CommandBarCatalog.swift:1549`, which formats `urlCleanerRemovedFormat` with `names.joined(separator: ", ")` and has no upper bound on its length. That one is the realistic overflow in shipped code; a copied answer from the command bar is the other.

`showScrollingCapture(message:)` sits on the same panel geometry and was unbounded too. Its one caller passes a fixed localized string today, so it is not reachable with arbitrary text, but it is the same defect one function down and is bounded here as well.

## The change

Both messages are capped at two lines and 360pt.

A short message still sizes to itself rather than being padded out to the cap — as the table shows, a picked colour measures exactly what it did before. That matters when almost all of the 68 callers pass something short.

**Truncation is at the tail, not the middle.** With middle truncation, a value whose first paragraph happens to end on the second line is cut at a line break rather than inside one, so no ellipsis is drawn anywhere and the preview reads as the entire copied value:

| middle | tail |
|---|---|
| `…transcript, / which runs on for a while before it ends.` | `…transcript, / which runs on for a while before it ends.…` |

Rendered both offscreen to compare.

The clipboard is unaffected either way — `copyAnswer` writes the full value to the pasteboard and the HUD only previews it.

## Testing

`./build.sh` clean, `--selftest` passes, `./build.sh --test` passes. Sizes measured directly against the same view hierarchy the HUD builds, and the panels rendered to images to confirm the wrapping and the ellipsis.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/QuickToolHUD.swift
+++ b/Sources/Vorssaint/Services/QuickTools/QuickToolHUD.swift
@@ -12,6 +12,11 @@ enum QuickToolHUD {
     private static var scrollingPanel: ScrollingCapturePanel?
     private static var scrollingModel: ScrollingCaptureHUDModel?
     private static var dismissWork: DispatchWorkItem?
+    /// How wide a message is allowed to get, on either of this file's two
+    /// message panels. A confirmation is read at a
+    /// glance, so anything past this is a preview rather than the whole value;
+    /// a short one still sizes to itself and is not padded out to this width.
+    fileprivate static let messageWidthLimit: CGFloat = 360
     /// Bumped by every show(). A dismiss whose fade-out was overtaken by a
     /// newer show() must not order the panel out from its completion handler.
     private static var generation = 0
@@ -51,6 +56,19 @@ enum QuickToolHUD {
             }
             Text(message)
                 .font(.system(size: 12, weight: .semibold))
+                // What was copied can be a whole paragraph, and the panel is
+                // laid out at whatever the text asks for. Unbounded, one long
+                // line measures wider than the screen and the panel, centred on
+                // that width, hangs off both edges with nothing readable left.
+                //
+                // Truncating at the tail rather than the middle so that the
+                // ellipsis is always drawn: a value whose first paragraph ends
+                // on the second line is cut at a line break, not inside one,
+                // and middle truncation leaves no mark at all there — the
+                // preview then reads as the whole of what was copied.
+                .lineLimit(2)
+                .truncationMode(.tail)
+                .frame(maxWidth: messageWidthLimit, alignment: .leading)
         }
         .padding(.horizontal, 14)
         .padding(.vertical, 9)
@@ -272,6 +290,11 @@ private struct ScrollingCaptureHUDView: View {
                 Text(model.message)
                     .font(.system(size: 12, weight: .semibold))
                     .lineLimit(2)
+                    // Same panel geometry as the confirmation above, so the
+                    // same bound: this one is laid out from fittingSize and
+                    // centred on it too.
+                    .truncationMode(.tail)
+                    .frame(maxWidth: QuickToolHUD.messageWidthLimit, alignment: .leading)
                 Text("\(model.height) px")
                     .font(.system(size: 10, weight: .medium, design: .rounded))
                     .foregroundStyle(.secondary)


PathGao:
The place the bound is applied is not where the repo puts it, and I do not think it survives `fittingSize`.

`present` sizes the panel from `host.view.fittingSize` (`QuickToolHUD.swift:143`), which measures the hosting view with an unspecified proposal. Under a nil proposal a `Text` reports its ideal size, which is the whole message on one line — `lineLimit` does not change that, it only caps the line count once a width is imposed. `.frame(maxWidth: messageWidthLimit, alignment: .leading)` at `:70` clamps the frame's own reported width to 360, but the height that came out of that same measurement is a one-line height. Whether the two-line text is then drawn inside a one-line panel or the host re-measures depends on how SwiftUI re-proposes during placement, and I could not run it to find out.

What is checkable is that the two panels in this repo that already render an arbitrary-length string into a `fittingSize`-hosted window both solve it by pinning the container width, not by putting `maxWidth` on the text:

- `QRResultController.swift:184` — `.frame(width: 320)` on the root, with `.fixedSize(horizontal: false, vertical: true)` on the payload `Text` at `:156` so the height is measured at the bounded width.
- `RadialNowPlayingService.swift:194` — `.frame(width: 320, alignment: .leading)` on the card, with `.lineLimit(2)` on the arbitrary track title at `:162`.

Following either one here means moving the bound onto the `HStack` (a `.frame(maxWidth: 360 + icon + padding)` or a flat width) and adding `.fixedSize(horizontal: false, vertical: true)` after the frame at `:70`, so the wrap is decided before `fittingSize` reads the height rather than after.

Two smaller notes. The `Text(message)` at `:56` serves 72 `QuickToolHUD.show` call sites, not the handful the description implies; I went through the localized strings behind them and none exceed two lines at 360pt, so nothing gets newly truncated — the one to keep an eye on is `urlCleanerRemovedFormat` at `CommandBarCatalog.swift:1550`, whose body is a joined list of removed parameter names and is the only message with no upper bound on its length. And `showScrollingCapture` at `:79` takes a `message` too and is unbounded on the same panel geometry, if you want the fix to cover the whole file rather than one of its two message paths.

rhukster:
Thanks — `showScrollingCapture` is a fair catch and it's bounded now in dd501be, along with the description fixes: 68 `show` call sites, and you're right about `urlCleanerRemovedFormat` at `CommandBarCatalog.swift:1549` being the one with no upper bound. I'd grepped the URL cleaner's own views and missed the command bar path.

On `fixedSize` I think the measurement disagrees. Measuring `NSHostingController.fittingSize` against the real hierarchy, on a 2560pt screen:

| message | bound on Text (this PR) | + `.fixedSize(h: false, v: true)` | flat `.frame(width: 360)` |
|---|---|---|---|
| `#3A7BD5` | 110 × 37 | 110 × 37 | 388 × 37 |
| one long line | 414 × 37 | 414 × 37 | 388 × 48 |
| a paragraph | 414 × 48 | 414 × 48 | 388 × 48 |

Adding `fixedSize` changes nothing — the height stays one line for an unwrapped long line either way. The flat width does change it, and is the only form that wraps a long single line to two, but it pads every short message out to 388pt, and nearly all 68 callers pass something short.

On the clipping worry: a long single line is drawn as one truncated line inside a one-line panel, not two lines clipped inside one. I rendered it to check rather than reason about the placement pass.

So the practical difference is whether a long *unwrapped* line gets one truncated line or two. This PR gives one, which fixes the 5199pt overflow and leaves the short confirmations untouched. If you'd rather have two lines and accept the wider panel for short messages, the flat-width version is a one-line change and I'll push it — I just didn't want to trade 68 compact panels for it without asking.

PathGao:
Measurement settles it, and it matches what `frame(maxWidth:)` does under a nil proposal: it proposes nil to the child, gets the ideal size back (5199 × 37), and clamps only the width it reports. The height that comes out is the pre-wrap one. `fixedSize(horizontal: false, vertical: true)` also proposes nil horizontally, so it is a no-op on top of that — 414 × 37 is what the layout has to produce. The QR and RadialNowPlaying precedent I pointed at was the wrong one: those pin a width because they want a fixed-width panel, not because the height needed it.

So the shape is: a string with its own line breaks gets two lines, a single long line gets one truncated line. `urlCleanerRemovedFormat` is a joined list, so it is single-line and lands in the one-line case — down from 5199pt, which is the bug.

Don't push the flat width. One extra preview line on the long-single-line case is not worth 388pt on the other 67 callers.

Nothing else to flag in the diff. The bound is one constant read by both panels rather than a literal at each site, which is the part that keeps this from drifting apart later.

**Impact:** none outside the two message panels; short messages measure exactly as before.
**Merge:** ship it.

Issue #916, opened 2026-08-24
Multi-line snippets lose their line breaks and surrounding words when expanded


fermincasagrande:
### Description

**Any snippet whose replacement contains a newline is silently corrupted when it expands.** Line breaks are lost, and so are arbitrary words either side of them. This is not specific to any keyboard or app.

**Root cause.** `TextSnippetService.postExpansion` types the replacement as synthetic key events carrying a unicode string, chunked to 20 UTF-16 units:

```swift
let units = Array(text.utf16)
var index = 0
while index < units.count {
    var end = min(index + 20, units.count)
    ...
    event.keyboardSetUnicodeString(stringLength: chunk.count, unicodeString: chunk)
```

A newline cannot travel inside that string. The receiving app inserts only the text **before the first newline** in each event and discards the remainder of that event's string. So each chunk is truncated at its first newline, and a chunk that *begins* with one contributes nothing at all.

Because the loss is per-chunk, the words that disappear depend on where the 20-character boundaries happen to fall — which makes the corruption look arbitrary.

### Steps to Reproduce

1. Create a snippet with a multi-line replacement, e.g. trigger `;art`:

```
Escribe un artículo periodístico con el siguiente titular: 

ENTRADILLA: 

...

Utiliza un lenguaje periodístico sobrio y neutro, sin valoraciones políticas, y no uses palabras con carga valorativa.
```

2. Type `;art` followed by a space in any text field.
3. Observed:

```
Escribe un artículo periodístico con el siguiente titular: Utiliza un lenguaje periodístico sobrio y neutro, sin valoraciones políticas, y no uses palabras con carga valorativa.
```

All line breaks are gone, and so are `ENTRADILLA:` and `...`.

### Why those particular words

They were not special. Chunking the stored text into 20 UTF-16 units gives:

```
chunk  3 [ 40: 60] 'siguiente titular: \n'      → inserts up to the newline, rest dropped
chunk  4 [ 60: 80] '\nENTRADILLA: \n\n...\n\n'  → begins with a newline, entire chunk lost
```

Applying the rule "each chunk contributes only the text before its first newline" to the stored string reproduces the observed output **character for character**.

### Suggested fix

Send multi-line replacements through the pasteboard, restoring the previous contents afterwards, and keep the existing typed path for single-line ones so the common case still never touches the clipboard.

Pasting rather than posting Return per line matters: Return sends the message in chat composers, which is exactly where multi-line snippets tend to be used.

[...]

PR #917, opened 2026-08-24
fix(snippets): paste multi-line snippets instead of typing them

fermincasagrande:
Refs #916.

### The problem

Any snippet whose replacement contains a newline is silently corrupted on expansion — line breaks vanish, and so do arbitrary words either side of them. It affects every user, on every keyboard, in every app.

`postExpansion` types the replacement as synthetic key events carrying a unicode string, chunked to 20 UTF-16 units. A newline cannot travel inside that string: the receiving app inserts only the text **before the first newline** in each event and discards the rest of that event's string. Each chunk is therefore truncated at its first newline, and a chunk that *begins* with one contributes nothing.

Because the loss is per-chunk, which words disappear depends on where the 20-character boundaries fall, which makes the corruption look random. A real example:

```
stored:   "…con el siguiente titular: \n\nENTRADILLA: \n\n...\n\nUtiliza…"
inserted: "…con el siguiente titular: Utiliza…"
```

`ENTRADILLA:` and `...` were not special — they shared a chunk with a leading newline:

```
chunk 3 'siguiente titular: \n'      → inserts up to the newline, rest dropped
chunk 4 '\nENTRADILLA: \n\n...\n\n'  → begins with a newline, entire chunk lost
```

Applying that rule to the stored string reproduces the observed output character for character.

### The change

Multi-line replacements go through the pasteboard, which preserves the text exactly. Single-line replacements keep the existing typed path and still never touch the clipboard, so the common case is unchanged.

The previous pasteboard contents are snapshotted with **every flavour**, not just the string, and restored after a short delay, so a rich clipboard is not quietly downgraded to plain text.

Pasting rather than posting Return per line is deliberate: Return sends the message in chat composers, which is exactly where multi-line snippets are most used. Posting Return there would fire off partial messages instead of inserting a prompt.

For an `afterDelimiter` trigger the delimiter now travels inside the pasted string rather than being re-posted as a keystroke. A plain key is handled before the paste the front app is still working through, so posting it afterwards lands the delimiter **in front of** the replacement — visible as a stray leading space. Carrying it in the pasted text makes the ordering deterministic rather than a race.

### Testing

macOS 26.5.2 (25F84), Apple Silicon, built with `./build.sh`, `--selftest` passes.

- Multi-line snippet with blank lines: expands complete and correctly formatted, in both a plain text field and a chat composer. Previously lost its line breaks and two fragments.
- `afterDelimiter` trigger: no leading delimiter; it lands after the replacement as intended.
- Single-line snippet: unchanged, and the clipboard is untouched.
- Clipboard contents restored afterwards.

### Note

One visible side effect worth flagging: with clipboard history enabled, the temporary pasteboard contents are recorded as a history entry during a multi-line expansion. Avoiding that would mean teaching the history service to ignore its own app's transient writes, which felt out of scope here.


PathGao:
Routing multi-line replacements through the pasteboard is the right answer, and putting the branch inside `postExpansion` means all three callers get it. The problem is that this repo already has a mechanism for "temporarily rewrite the clipboard and put it back", and `postPaste` did not join it.

`PastePlainService` is the reference implementation, and `ClipboardAutoClearService` follows it. Four things they do that `postPaste` does not:

| | `PastePlainService` | `postPaste` |
|---|---|---|
| Tell history to ignore the write | `PastePlainService.swift:92`, `:113` — `ClipboardHistoryService.shared.ignoreNextChange(upTo:)` | absent |
| Guard the restore | `:112` — restores only if `pasteboard.changeCount == plainChangeCount` | unconditional (`TextSnippetService.swift:431`) |
| Re-entrancy | `isPerforming`, `pendingRestore`, a cancellable `restoreWork` | none |
| Wait for a clean keyboard | `postPasteWhenModifiersReleased` | posts ⌘V immediately |

What each one costs:

- **No `ignoreNextChange`.** Both the snippet text and the restore land in Clipboard History as entries the user never copied. `ClipboardAutoClearService.swift:208` calls it; so does `PastePlainService` twice. This is a mechanism with a member missing.
- **Unconditional restore.** Anything the user copies inside the 0.3s window is silently overwritten by the snapshot.
- **No re-entrancy state.** Two multi-line expansions within 0.3s corrupt each other: the second snapshots the *first snippet's* text as "the original", and the first's restore then clobbers the second's paste. The first snippet's text can be left on the clipboard permanently.
- **⌘V merging with a held Shift.** The tap rejects only `.maskCommand` and `.maskControl` (`TextSnippetService.swift:245`), so a trigger completed by a shifted character (`!`, `:`, a capital) fires with Shift physically down and the synthesized ⌘V reads as ⇧⌘V.

Two more:

- 0.3s here vs 0.5s in `PastePlainService`. Two places answering "how long before restoring" differently, with nothing requiring them to agree.
- `postPaste` passes the `.hidSystemState` source into the ⌘V event. `PastePlainService.postPasteShortcut` deliberately passes `nil`, with a comment that an HID-tied source re-poisons the flags. The source is needed here to carry `syntheticMarker`, so the two requirements are in tension — worth resolving in the code rather than leaving both comments standing.

And the snapshot closure at `TextSnippetService.swift:410` is a re-implementation of `PastePlainService.snapshot(of:)` (`PastePlainService.swift:239`). Calling into the existing one, or lifting both onto a shared type, gets all of the above for free rather than one item at a time.

PathGao:
Two things not covered above, both about where `postPaste` runs rather than what it does.

**It runs with the event tap parked behind it.** `expand` reaches `post()` from the tap thread through `DispatchQueue.main.sync(execute: post)` (`TextSnippetService.swift:335`), so the tap callback is blocked for the whole of `post()`. Until this PR that section only built a string and posted CGEvents — bounded work. `postPaste` now adds, inside the same blocking section, `pasteboard.pasteboardItems` plus `item.data(forType:)` for every flavour of every item (`:410`), a `clearContents` + `setString`, and the ⌘V post. Every one of those `data(forType:)` calls round-trips to whichever app owns the clipboard, and a promised flavour that is slow to materialise stalls there. When it exceeds the tap timeout macOS disables the tap; `handle` re-enables it at `:216`, but the keystroke that triggered the expansion is gone and the buffer state with it. The single `.string` read already in `post()` at `:322` has this exposure today, but this PR multiplies it by the flavour count and adds a second transaction. Doing the snapshot and the write on `GeneralPasteboardAccess.shared` and posting ⌘V from its completion keeps the tap thread out of it — and that lane is also what every other pasteboard writer in the repo uses (`ClipboardHistoryService.swift:117`, `ClipboardAutoClearService.swift:157` and `:178`, `URLCleanerService.swift:67`, `FinderCutPaste.swift:339`), which `postPaste` currently bypasses entirely.

**The restore clears first and asks questions second.**

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
    pasteboard.clearContents()                                    // :432
    if let saved, !saved.isEmpty { pasteboard.writeObjects(saved) } // :433
}
```

When `pasteboardItems` came back nil or empty at `:410` — which is what a pasteboard whose owner died or whose flavours would not materialise looks like — this wipes the clipboard instead of leaving it alone. `PastePlainService` orders it the other way and returns without clearing: `guard pasteboard.changeCount == plainChangeCount, !snapshot.isEmpty else { return }`. Move `clearContents()` inside the `if let saved, !saved.isEmpty` so an unreadable pasteboard is left as it is rather than emptied, and so the no-op case does not spend a change count.

I could not run either of these; the first in particular depends on how long a real stalled provider takes to answer, which I have no way to measure here.

PathGao:
One more, and it's the delimiter change rather than the paste itself.

`text.contains(where: \.isNewline)` tests only the replacement, but `trailingText` rides along on the branch that test picks. So the same delimiter keystroke does two different things depending on a property of the snippet that preceded it:

```
single-line snippet + Return delimiter  → postKey(...)   → composer sends the message
multi-line  snippet + Return delimiter  → paste "…\r"     → composer inserts a newline
```

The body's argument for pasting is that "Return sends the message in chat composers, which is exactly where multi-line snippets are most used". That holds for the replacement's own newlines. Applied to the delimiter it inverts: the user pressed Return meaning *send*, and on the multi-line path it becomes a character in the buffer instead. Tab is the same class — moves focus when posted, inserts or is stripped when pasted. So is Return in a single-line `NSTextField`, where posting it fires the field's action.

The testing note covers "`afterDelimiter` trigger: no leading delimiter", which is the space case — the one delimiter where text and keystroke agree. The ones that disagree are the keys that carry an action rather than a character.

Only printable delimiters can travel inside the pasted string. For the rest the key still has to be posted, just after the paste has landed rather than before — which is the ordering a completion-based restore gives you anyway, instead of the fixed 0.3s timer.

I can't see which delimiters are actually accepted; that set isn't in the diff. If it's restricted to printable characters this is moot, but `handle` forwards a raw `keyCode` plus `event.flags`, which reads like it isn't.

Second thing, also unverified: `trailingText` got a `= ""` default so existing `expand` / `postExpansion` call sites compile untouched. Any call site that passes a non-nil `trailingKeyCode` without a `trailingText` now drops that key entirely on the multi-line path, because the branch returns before the trailing post. Those call sites aren't in the diff, so I can't tell whether any do — worth checking rather than assuming the default is inert.

**Impact:** multi-line snippet completed by an action key (Return, Tab): the delimiter loses its action. In a chat composer, Return stops sending.
**Merge:** needs the delimiter split into printable (carry it in the text) and action keys (post it after the paste) first.

ref #916

fermincasagrande:
Reworked. `postPaste` is gone; the transient write now lives in a `TransientPaste` type that `PastePlainService` also uses, so there is one implementation rather than two.

**Joining the existing mechanism.** All four missing members are in: `ignoreNextChange` on both the write and the restore, the restore guarded on `changeCount`, re-entrancy state, and waiting for a clean keyboard before ⌘V. The ⌘V is posted with **no** event source, per your note that an HID-tied source re-poisons the flags — the `syntheticMarker` source is still used for the deletes, which is where it is actually needed, so the tension you spotted resolves rather than leaving both comments standing. Delay is 0.5s, matching `PastePlainService` rather than inventing a second answer.

**Ordering of the restore.** Fixed — it checks before clearing, so a pasteboard whose contents could not be read is left alone instead of emptied.

**The thread.** Snapshot and write go through `GeneralPasteboardAccess`, so the tap thread is no longer parked behind `data(forType:)` round-trips to whichever app owns the clipboard. `expand` returns promptly and the ⌘V is posted from the completion.

**The delimiter.** You were right, and this was a design error rather than an oversight — the argument for pasting applies to the replacement's own newlines and inverts when applied to the delimiter. Split by kind now: a printable delimiter rides inside the pasted string; an action key is posted after the paste has landed, using the completion rather than a fixed timer. That also settles your second point — a call site passing `trailingKeyCode` with no `trailingText` now posts its key instead of dropping it.

---

**`PastePlainService` is migrated onto `TransientPaste` in this PR, and it is NOT tested.** `featureAvailable.pastePlain` is off on the machine this was written on, so I could not exercise it once. What moved is the snapshot with `pendingRestore` reuse, the clear-and-write, `ignoreNextChange`, the modifier wait, the ⌘V and the guarded restore; what stayed is the Accessibility check and prompt, plain-text extraction, and `pressNativeMatchStyleItem`, which returns before any transient write.

One deliberate behaviour change there: the snapshot used to be taken on the main thread immediately after `plain` was extracted, and now happens on the pasteboard queue a moment later. A clipboard write inside that gap would be captured differently.

I did it rather than leaving the duplication because the two copies of `pendingRestore` cannot see each other — a paste-plain landing inside a snippet expansion's restore window snapshots *that* text as the original and restores it afterwards, losing the user's clipboard permanently. Guarding re-entrancy inside each service separately cannot fix that. But it is a refactor of a working feature I have no way to run, so it may be worth splitting out or reverting if you would rather take it separately.

The snippet path sharing `TransientPaste` **is** tested on macOS 26.5.2, Apple Silicon: multi-line snippets expand intact, the delimiter lands after the replacement, the clipboard is restored, and nothing appears in clipboard history. `--selftest` passes.

PathGao:
Rework reads right — one implementation, all four members in, restore ordering fixed. Four things, all in what moved.

**`postPasteShortcut` lost the two parts that were doing work.** `PastePlainService` had:

```swift
let mustReleaseHotkey = shortcut.isStandardPasteCommand
if mustReleaseHotkey { self.hotkey.unregister() }   // else our own ⌘V catches the synthetic one
...
keyDown.post(tap: .cghidEventTap)
DispatchQueue.main.asyncAfter(deadline: .now() + 0.04) { keyUp.post(...) }  // zero-length tap gets skipped
```

`TransientPaste.postPasteShortcut` has neither — `for down in [true, false] { … post() }`, no unregister, no beat. Anyone who set the paste-plain shortcut to plain ⌘V now has their own hotkey swallow the synthetic one, so paste-plain does nothing at all; that case is the entire reason `isStandardPasteCommand` exists. The beat is a comment marked verified-empirically being deleted rather than moved, and it applies to the snippet path too — your summary lists "the ⌘V" among what moved, but it was rewritten, so the one passing test doesn't cover either of these.

**The action-key delimiter is posted into the race the body says can't be won.** The body:

> A plain key is handled before the paste the front app is still working through, so posting it afterwards lands the delimiter **in front of** the replacement.

`TransientPaste`'s doc comment:

> `completion` runs on the main queue once ⌘V has been posted — which is the point at which a trailing keystroke can safely follow the pasted text

Both can't hold. If posting after the ⌘V is safe for Return, it was safe for the space and the delimiter never needed to ride inside the text. If it isn't, Return reaches the composer before the pasted text does and sends an empty or partial message — the failure the pasting exists to prevent, now on the path that was supposed to fix it. Posted is not landed; something has to observe the text arriving.

**A dropped paste still eats the trigger.** `postExpansion` posts `deleteCount` backspaces, *then* calls `paste`, which returns silently when `isPerforming` — `completion?()` fires, the printable branch does nothing, and the tap already returned nil for the delimiter key. Trigger deleted, no replacement, delimiter gone. The drop was free in `PastePlainService` because nothing had been destroyed yet; here it isn't. Make it visible — `paste` returning `false` — and check it before the deletes go out.

**`ignoreNextChange` has a window now.** It used to run back to back with `setString` on one thread. Now the write finishes on `GeneralPasteboardAccess` and the ignore is registered a hop later on main; a poll landing in between records the transient text, which is the note in your body coming back. Same on the restore. Registering from the pasteboard queue closes it. I can't see which lane the history poller runs on, so I don't know whether the gap is actually reachable.

Separately: `AssistiveKeyboard.swift` and the five dismissal call sites are in this diff and nowhere in the body. I read them as a stacked branch and didn't review them — say so if they're meant to be part of this PR and I'll go through them.

**Impact:** paste-plain bound to ⌘V stops working entirely. A multi-line snippet completed by Return can send before the text lands.
**Merge:** needs `postPasteShortcut` to carry the hotkey release and the down/up beat back first.

ref #916

fermincasagrande:
You were right on all four, and the first two were regressions I introduced while calling them a move. I checked both against `main` before changing anything.

**`postPasteShortcut`.** The beat is back, with its comment, and so is the guard-let form that still calls back when the events cannot be created. The hotkey release is back too: `paste` now takes `willPostShortcut` / `didPostShortcut` hooks that bracket the press on the main queue, and `PastePlainService` unregisters and re-syncs across it exactly as before. My summary said the ⌘V "moved" — it was rewritten, and it dropped the two parts that were carrying weight. That is the distinction you have been drawing throughout, and it is what let you find this.

**The delimiter contradiction.** Real, and you named the resolution: *posted is not landed*. Rather than find something that observes the text arriving, nothing is posted at all — the delimiter now rides inside the pasted string with the replacement, so the insertion is one ordered event and there is no race to lose. `TransientPaste`'s doc comment no longer claims a safe post-point, because there isn't one.

The trade is explicit in the code: Return arrives as a newline in the inserted text rather than as a send. For a multi-line snippet completed by Return, that seems the more useful reading of the intent anyway — and it is certainly better than sending a partial message. Tab likewise inserts a tab instead of moving focus. If you would rather `afterDelimiter` preserved the keystroke's *action* on this path, that needs the observation step and I would want to know what signal you would trust for it.

**The dropped paste.** Fixed as suggested — `paste` returns `false` when it declines, and the deletes only go out once it has been accepted. You are right that the drop was free in `PastePlainService` and is not free here, because by then the trigger is gone and the tap has already swallowed the delimiter.

**`ignoreNextChange`.** Moved onto the pasteboard queue, next to the write it covers, and the same on the restore. I could not tell either whether the gap was reachable — which lane the history poll runs on is not visible from here — but closing it costs nothing.

**On the stacked content:** correct, `AssistiveKeyboard.swift` and the five dismissal sites belong to #898 and #922 and are in this diff only because this branch sits on top of them. Please review them there. This PR is the `TransientPaste` work plus the `PastePlainService` migration.

---

`PastePlainService` remains **untested** — `featureAvailable.pastePlain` is off on the only machine available, so the hotkey release I just restored is reasoned from your reading of `main`, not exercised. That is the part I would most like a second pair of eyes on, since it is the path that silently does nothing when wrong.

Built and self-tested. The snippet paste path was verified on hardware before this round; the reporter's machine is deliberately left on the previous build, so these four changes are not re-verified there.

PathGao:
Two things, and the first one is mine to correct.

**`ignoreNextChange` on the lane is worse than the hop to main I asked for.** `ClipboardHistoryService` owns `lastChangeCount` on the main queue, and the sibling writer says so in as many words:

```swift
// lastChangeCount stays owned by the main queue, where the capture
// poll compares against it. Assigning it on the lane would let a
// copy's own change count land after the poll had already read the
// old one, and history would record the app's own write as a new
// copy by the user.
if let self { self.lastChangeCount = max(self.lastChangeCount, changeCount) }
```

`ClipboardHistoryService.swift:134`. `baselinePasteboard` hops too (`:545`). `TransientPaste` is now the only writer that doesn't — on both the write and the restore — so a plain `var Int` is mutated from the pasteboard queue while `captureIfChanged`'s completion reads and writes it on main, with no lock between them.

The reachable consequence is `:590`, `guard changeCount > self.lastChangeCount else { return }`. The poll's read of a genuine user copy can be in flight — and that read is the slow one, the one that sits behind a promised flavour for seconds — when the lane pushes `lastChangeCount` past it. The completion then drops it. Not our text recorded; their copy lost.

The gap I was worried about closed on its own, and this PR is what closed it: the poll reads through `GeneralPasteboardAccess` too (`:576`). The lane is serial, so the transient write finishes before the poll's read starts, and its main-queue hop is already enqueued ahead of the poll's. Registering from main is ordered correctly now and races nothing.

**The deletes are not inside the ordered event.** The body:

> Carried inside, the whole insertion is one ordered event.

```swift
guard TransientPaste.shared.paste(payload) else { return }
for _ in 0..<deleteCount { postKey(CGKeyCode(kVK_Delete)) }
```

`paste` returns as soon as it has enqueued to the lane. The deletes go out now; the ⌘V goes out after the snapshot round-trips, the modifier wait and the 0.06 s beat. That is ~60 ms on a healthy machine and unbounded in the case your own doc comment names — "a promised flavour that is slow to materialise stalls there". In between, the trigger is gone and anything typed lands where the replacement was going to.

The reasoning you accepted for the delimiter applies here unchanged, and the hook already exists:

```swift
guard TransientPaste.shared.paste(payload, willPostShortcut: {
    for _ in 0..<deleteCount { postKey(CGKeyCode(kVK_Delete)) }
}) else { return }
```

Closes it to the down/up beat, and the trigger stays on screen until it is replaced instead of blinking out first.

Closing my earlier point about the `trailingText` default: no such call site. `CommandBarExtras.swift:32` and `SnippetLibraryService.swift:210` both pass `trailingKeyCode: nil`. Your reply said such a call site "now posts its key instead of dropping it" — the multi-line branch returns before the trailing post, so it drops it — but with nothing reaching it that is a sentence to fix, not code.

`PastePlainService` I read against `main` rather than ran. The bracketing is right: `willPost` before keyDown, `didPost` after keyUp, and the guard-let failure path still calls back, so the hotkey cannot be left unregistered. That is not the same as exercising it, and I have no way to.

**Impact:** a copy made while a multi-line snippet expands can be dropped from clipboard history; the trigger is deleted up to seconds before its replacement arrives.
**Merge:** needs `ignoreNextChange` hopped back to main, matching `ClipboardHistoryService.swift:134`, first.

ref #916

fermincasagrande:
Both taken, and thank you for retracting the first one with the evidence rather than letting it stand.

**`ignoreNextChange` is back on main**, on the write and the restore, matching `ClipboardHistoryService.swift:134`. You are right that my version was the worse trade: recording our own transient text is a cosmetic wrong entry, while racing the poll can drop a copy the user actually made. I moved it onto the lane on your earlier suggestion without checking who owned `lastChangeCount`, and the sibling writer says so in as many words two lines up from where I was editing. The comment now records why it hops, and that the window which motivated the move is closed by the poll reading through `GeneralPasteboardAccess` on the same serial lane.

**The deletes ride `willPostShortcut`** now, exactly as you wrote it. My "one ordered event" claim covered the pasted string and not the deletes, which went out the moment `paste` enqueued — tens of milliseconds early on a good day, unbounded behind a slow promised flavour, with the trigger already gone and the caret sitting in the gap. Using the hook that already existed for the hotkey release closes it to the down/up beat, and the trigger now stays on screen until the moment it is replaced. Confirmed on hardware: the reporter sees the trigger vanish as the text lands rather than a beat before.

**On the `trailingText` default:** you are right and my reply was wrong. `CommandBarExtras.swift:32` and `SnippetLibraryService.swift:210` both pass `trailingKeyCode: nil`, so nothing reaches it — and had anything reached it, the multi-line branch returns before the trailing post, so it would drop the key rather than post it. That was a sentence asserting behaviour I had not traced. No code change; correcting the claim here so it does not sit in the thread as though it were checked.

**`PastePlainService` remains unexercised.** `featureAvailable.pastePlain` is off on the only machine available, so the restored hotkey release is still reasoned from `main`, not run. Your reading of the bracketing matches mine — `willPost` before keyDown, `didPost` after keyUp, guard-let failure path still calling back so the hotkey cannot be stranded unregistered — but neither of us has watched it work.

Built, self-tested, and the multi-line snippet path confirmed on hardware, including that expanding one no longer leaves its text in clipboard history and the previous clipboard comes back.

PathGao:
Two things, plus the body.

**The ⌘V release didn't move with the ⌘V.** `GlobalShortcut.swift:311` states the rule the poster has to follow:

> Paste as plain text ultimately posts the standard paste command. When that same command is its configured global shortcut, the registration must be released briefly or it catches the synthesized paste again.

After this PR two things post that command, and the guard is a caller's job — `willPostShortcut`. `PastePlainService` passes the unregister; `TextSnippetService` passes the deletes. So with the paste-plain shortcut set to plain ⌘V, a multi-line expansion runs:

```
willPost → deleteCount backspaces, trigger gone
⌘V       → caught by our own hotkey, target app sees nothing
perform  → pressNativeMatchStyleItem(), else TransientPaste.paste
                                        → isPerforming still true → false
```

The hotkey fires on the keyDown, which is before `completion` clears `isPerforming`, so the fallback declines. Trigger deleted, nothing pasted, no beep. If the front app happens to carry the ⌥⇧⌘V equivalent the snippet lands stripped to plain text instead — still not what was asked for. Not the default (`.pastePlainDefault` is ⌥⇧⌘V), but it is the exact configuration `isStandardPasteCommand` was written for, and before this PR it only cost you paste-plain rather than the typed trigger.

The fact being guarded — *the ⌘V we post can be caught by our own hotkey* — is a property of whoever posts it, not of whoever asked. As a caller hook it is one copy per caller, and the second caller is already the one missing it. `PastePlainService` still needs the bracket for the re-sync; the release decision reads `DefaultsKey.pastePlainShortcut` and can sit where the press happens. Note there is already a house form for this: `suspendShortcut()` on `ClipboardHistoryService`, `ShelfService`, `SoundOutputSwitcher` (`ShortcutCapture.swift:29`–`:32`). `PastePlainService` has none.

I'm taking "the Carbon hotkey catches a synthetic ⌘V" from your own comment on `main`, not from running it.

**`paste` returns `true` unconditionally off-main.** The doc says it returns `false` when a previous paste is in flight, which is what the deletes now hang on. On the `guard Thread.isMainThread else` path it hops and returns `true` before anything has been decided, so a caller that destroys first would destroy on a dropped paste. All three call sites reach `postExpansion` on main today (`expand` uses `main.sync`, and the two panel sites post from `main.asyncAfter`), so nothing hits it — but the guarantee the return value exists to give is not the one that path gives.

**The body no longer describes the diff.** The Note still says the transient contents get recorded as a history entry and that avoiding it "felt out of scope" — `ignoreNextChange` is in on both the write and the restore, and you confirmed on hardware that nothing lands in history. And "The change" covers the snippet path only: the `PastePlainService` migration onto `TransientPaste` — a rewrite of a separate feature neither of us can run — appears nowhere in it. That is the part most likely to need attention after this merges, and right now it exists only in this thread.

`PastePlainService` I read against `main` again rather than ran; `featureAvailable.pastePlain` is off here too.

**Impact:** with paste-plain bound to plain ⌘V, a multi-line snippet deletes its trigger and pastes nothing.
**Merge:** needs the ⌘V hotkey release to cover the snippet path first.

ref #916

Issue #920, opened 2026-08-24
⌘-Tab switcher dies on the second Tab when using the Accessibility Keyboard


fermincasagrande:
### Description

With the macOS Accessibility Keyboard, Vorssaint's ⌘-Tab switcher **opens on the first Tab and disappears on the second**, every time. It cannot be used to cycle.

This is not a modifier-release problem. On that keyboard a modifier is latched persistently by **double-clicking** it (the key turns red), so ⌘ genuinely stays held — the native macOS switcher cycles correctly under exactly the same conditions.

**Root cause.** Every other key is pressed by *clicking* it, and those clicks land outside the switcher panel:

```swift
case .leftMouseDown, .rightMouseDown, .otherMouseDown:
    ...
    dismissForClickOutsidePanel()
    return Unmanaged.passUnretained(event)
```

```swift
private func dismissForClickOutsidePanel() {
    guard sessionActive, let panel else { return }
    guard SwitcherSupport.shouldDismissForClick(...) else { return }
    cancelSession()
}
```

So:

| | |
|---|---|
| **Tab click 1** | mouse-down arrives, `sessionActive == false`, the guard returns early. `keyDown` Tab then opens the session — works |
| **Tab click 2** | mouse-down arrives, `sessionActive == true`, click is outside the panel → `cancelSession()` runs before the `keyDown` is handled — session gone |

That is exactly the observed symptom: it always survives the first Tab and always dies on the second.

### Steps to Reproduce

1. Enable the Accessibility Keyboard (System Settings → Accessibility → Keyboard).
2. **Double-click** the ⌘ key on it so it latches (turns red).
3. Click Tab — the switcher appears.
4. Click Tab again — the switcher vanishes instead of advancing.
5. For contrast, the native macOS ⌘-Tab switcher cycles correctly with the same latched ⌘.

### Suggested fix

Skip the dismissal when the click landed on the Accessibility Keyboard's own panel (bundle id `com.apple.inputmethod.AssistiveControl`) — such a click is the user driving the switcher, not pointing away from it. Every other click keeps its current behaviour.

This is the same underlying assumption as #896: a mouse-down treated as "the user did something else", when on an on-screen keyboard a mouse-down *is* how a key is pressed. Worth checking whether other features that dismiss on click-outside (dock preview, quick toggles) share it.

PR follows.

### Vorssaint Version

3.3.3-beta.2

### macOS Version

[...]

PR #922, opened 2026-08-24
fix(a11y): keep panels open with the Accessibility Keyboard

fermincasagrande:
Keep the App Switcher, Command Bar, Quick Launcher, Clipboard History, Snippet Library, Scratchpad and Camera Preview open when a click presses an Accessibility Keyboard key. Ordinary outside clicks still dismiss them, and keyboard clicks preserve navigation while the switcher is preparing its window list.

Resolve the keyboard's current process before judging each click so the first key also works when the keyboard starts after the app or restarts. This removes the asynchronous process cache and its retry delay while retaining the existing window hit-testing behavior.

Validation covers the repository tests, focused first-use and process-change contracts, controlled panel and switcher routing sequences, and a native background-app lifecycle fixture. The local development build, installation, installed selftest and strict signature verification passed. The native keyboard interface was not exercised during this validation.

Refs #920
Refs #897


PathGao:
The detection here is right, and the same problem is in five more places.

**First: #898 adds the identical function.** `clickTypedOnAssistiveKeyboard(_:)` and `assistiveKeyboardBundleID` in that PR are character-for-character what this one adds, just in `Services/Snippets/TextSnippetService.swift` instead of `AppSwitcher.swift`. Two private copies of one predicate, neither aware of the other. Whichever of the two lands first should put it somewhere shared and the other should call it.

**Second: four more panels have the same bug.** Someone typing on the Accessibility Keyboard emits a `leftMouseDown` for every character, and these all treat a click outside themselves as "dismiss" while accepting typed input:

- `Services/CommandBar/CommandBarService.swift:2493`
- `Services/QuickTools/QuickLauncherService.swift:401`
- `Services/Clipboard/ClipboardHistoryService.swift:1180`
- `Services/Snippets/SnippetLibraryService.swift:302`

All four use the same `localClickMonitor` / `outsideClickMonitor` pair. So the Command Bar, Quick Launcher, Clipboard History and Snippet Library currently cannot be typed into at all from that keyboard — the panel closes on the first key.

Those four use `NSEvent` monitors while this PR works on a `CGEvent`, so a shared helper wants to take a point rather than an event — something like `AssistiveKeyboard.ownsPoint(_ point: CGPoint) -> Bool` that all six call.

Would you widen this to cover them? As it stands it fixes the switcher and leaves five siblings of the same bug open, which closes the issue number without closing the problem. Labelling `incomplete` until then — the code itself is fine.

PathGao:
Separate from the duplication and the five sibling panels, two things about the lookup itself.

**It runs on the main thread with the tap thread blocked on it, for every mouse-down in a session.** `route` hands events over with `DispatchQueue.main.sync { verdict = self.handle(...) }` (`AppSwitcher.swift:540`), so `clickTypedOnAssistiveKeyboard` at `:1328` costs the tap thread a full `CGWindowListCopyWindowInfo([.optionOnScreenOnly], ...)` over every on-screen window plus an `NSRunningApplication(processIdentifier:)` lookup per window whose bounds contain the point (`:1330`, `:1339`) — and that is paid on every `.leftMouseDown`, `.rightMouseDown` and `.otherMouseDown` while the switcher is open, by everyone, whether or not the Accessibility Keyboard is running. A session with a session-scoped cache of the keyboard's pid, or just an early `guard NSWorkspace.shared.runningApplications.contains { $0.bundleIdentifier == Self.assistiveKeyboardBundleID }`, makes the scan cost nothing for the people who never open it. Whatever shared helper #898 and this end up sharing should carry that gate, not each caller.

**The loop answers "is the keyboard anywhere under this point", not "is it what was clicked".** `CGWindowListCopyWindowInfo` returns front to back, and `:1336` does `continue` on any window that contains the point but is not the keyboard, so a window sitting on top of the Accessibility Keyboard still reports `true` and the switcher refuses to dismiss on a click that never touched a key. Stop at the first real window containing the point and test that one, skipping the layers that are not clickable (`kCGWindowAlpha == 0`, and anything outside the normal/floating `kCGWindowLayer` values), rather than scanning past them.

One thing to fix before the helper is extracted: `event.location` here is in CG display coordinates, origin top-left, which is the same space as `kCGWindowBounds` — that part is right. The four `NSEvent`-monitor panels named earlier have `NSEvent.mouseLocation`, which is Cocoa's bottom-left origin. A shared `ownsPoint(_:)` taking a bare `CGPoint` will silently mis-hit for them, and worst on a multi-display setup where the flip is not a single screen height. Pick one convention and convert at the four call sites, or give the helper two entry points.

I could not test any of this against an actual Accessibility Keyboard, so the ordering claim in the second point is read off the API contract rather than observed.

PathGao:
One of the five siblings already has its own issue: #897 is `SnippetLibraryService.swift:302`, reported as the snippet library panel taking no keyboard input. Same bug, different panel — the first click on a key closes it. Widening this isn't scope creep then, it takes out a second reported bug at the same time.

**Impact:** none new. The switcher fix is correct as written; the other four panels stay unusable from that keyboard.
**Merge:** needs the predicate widened to the four `NSEvent` panels, or extracted so they can call it, first.

ref #920

fermincasagrande:
Widened, as asked. This branch is now rebased on #898 and calls the shared `AssistiveKeyboard` from that PR rather than carrying its own copy.

**The four sibling panels.** `CommandBarService`, `QuickLauncherService`, `ClipboardHistoryService` and `SnippetLibraryService` all go through the helper now, so none of them closes on the first key any more. You were right that this is not scope creep: it takes out **#897** at the same time, and #897's original diagnosis — mine — was wrong. I had concluded the library never takes key status, because it is `[.borderless, .nonactivatingPanel]` and reported `AXFocused = false` even after `AXMain`, `AXFocused` and activating the app. All true, none of it the cause. Typing landing in the app underneath, arrows dismissing the panel, a beep once the app was frontmost — every one of those is equally explained by the window having already closed, which is what was happening. Confirmed on hardware: arrows and Return now navigate the library.

**Coordinate spaces.** This was the most useful catch, because it would have been a silent mis-hit rather than a failure. The helper has two entry points: `ownsPoint` for `CGEvent.location` and `ownsCocoaPoint` for `NSEvent.mouseLocation`, flipping against the **primary** screen's height rather than any single screen's, since on a multi-display setup no one screen height converts correctly.

**The lookup cost, and "anywhere under the point".** Both handled in #898 — cached pid so the scan never runs when the keyboard is absent, and the hit-test now stops at the first clickable window containing the point. As you said, the gate belongs in the shared helper rather than at each caller, and that is where it is.

**`didActivateApplicationNotification`.** All four panels hide when another app comes forward, so if pressing a key brings the keyboard's process forward the panel closes regardless of the click fix. Exempted in all four, plus `TextSnippetService` in #898.

---

Tested on macOS 26.5.2, Apple Silicon, Accessibility Keyboard as the only input device: ⌘ latched by double-click then Tab cycles correctly, and the snippet library takes arrows and Return. Clicking genuinely outside any panel still dismisses it. I have not exercised the Command Bar, Quick Launcher or Clipboard History beyond confirming they build — they are not features this user has set up, so treat those three as reasoned-not-observed.

Note this depends on #898 landing first for the helper.

PathGao:
`ownsCocoaPoint` flipping against `NSScreen.screens.first` is the right conversion — that screen's origin is (0,0) by definition, so `maxY` is the height CG measures down from, and it stays correct for a display above the primary (negative CG y, which is legal). Two things on the helper, then a merge-order one.

**The first click of each app launch still answers "not running".** `livePID` resolves off-thread, and `bootstrap` doesn't run until the first call:

```swift
_ = bootstrap                                            // schedules the first lookup, async
...
if let cached, kill(cached, 0) == 0 { return cached }    // cached is still nil here
```

So the first mouse-down that reaches `ownsPoint` is answered before the cache exists, even with the keyboard running since login. "the next click gets the corrected answer" holds for the buffer reset in `TextSnippetService`; the switcher has no next click, `cancelSession()` already ran. And nothing in the switcher path warms it, because `dismissForClickOutsidePanel` guards on `sessionActive` before calling `ownsPoint` — the second Tab of the first session *is* the call that triggers bootstrap.

I suspect your run passed because `TextSnippetService`'s tap calls `ownsPoint` on every `leftMouseDown` with no gate, so an earlier click primed it. That makes the switcher's correctness depend on the snippets feature being enabled. Prime it eagerly wherever the services get set up:

```swift
_ = AssistiveKeyboard.isRunning
```

That belongs in #898 with the rest of the helper.

**`kill(pid, 0)` returns -1/EPERM for a live process you can't signal**, and the code reads that as dead: cache cleared, `scheduleLookup(force: true)` on every subsequent call (the force bypasses `minLookupInterval`), lookup re-resolves the same pid, `kill` fails again — false forever, plus a dispatch per click. Same-uid input method, so this likely never fires, but `kill(cached, 0) == 0 || errno == EPERM` closes it for nothing.

**The diff here contains #898.** `Services/AssistiveKeyboard.swift` and the `TextSnippetService.swift` changes are in this PR's diff against `main`, so merging #922 as it stands merges #898's content with #898's own review bypassed. Retarget this PR's base to #898's branch so the diff shows only the switcher and the four panels, or land #898 first and rebase. The PR body still says the predicate is self-contained and doesn't depend on #898 — that sentence is now false and is the thing a reader checks merge order against.

Not verified: whether anything already touches `AssistiveKeyboard` during startup (outside the diff); dock preview and quick toggles, which your own body flagged and which don't appear here; nothing on hardware.

**Impact:** one panel dismissal per app launch on the first click, unless the snippets tap happens to have primed the cache first. Otherwise the widening is correct as written.
**Merge:** hold for #898 — it carries the helper, the warm-up fix, and this PR's diff currently swallows it.

ref #920
ref #897

fermincasagrande:
Both helper points are fixed in #898, and you were right about the merge order.

**The cold first click.** Warmed in `applicationDidFinishLaunching` with `_ = AssistiveKeyboard.isRunning`, placed in #898 with the rest of the helper as you suggested.

The part I would not have found is *why my hardware run passed*: `TextSnippetService`'s tap calls `ownsPoint` on every `leftMouseDown` with no gate, so it primed the cache before the switcher ever asked. That made the switcher's correctness depend on the snippets feature being enabled — a coupling invisible from either diff, and it would have shipped looking tested. You are also right that the switcher cannot warm itself, since `dismissForClickOutsidePanel` guards on `sessionActive` before it asks.

**EPERM.** Added in #898.

**The diff containing #898.** Correct, and the body is fixed. Retargeting the base is not available to me: a PR's base must be a branch in this repository, and #898's branch lives in my fork. So the order is **land #898 first and I will rebase**, and the body now says that instead of the sentence claiming the predicate was self-contained. That sentence was true when written and became false when I rebased; leaving it up was my mistake, and it is precisely what a reader would check the merge order against.

**On the flip:** thank you for confirming `NSScreen.screens.first` — I had reasoned it from the origin being (0,0) by definition but had not thought through the display-above-primary case, and it is reassuring that negative CG y falls out correctly rather than by luck.

Still not verified, and I would rather say so than let the widening imply otherwise: Command Bar, Quick Launcher and Clipboard History compile but were never exercised, because they are not features the reporter has configured. Only the switcher and the snippet library are confirmed on hardware. Dock preview and quick toggles are untouched and may carry the same assumption.

PathGao:
I gave you four sibling panels and there are six. `Services/QuickTools/ScratchpadService.swift:448` is the one that matters:

```swift
outsideClickMonitor = NSEvent.addGlobalMonitorForEvents(matching: mouseEvents) { [weak self, weak panel] event in
    guard let self, let panel, panel.isVisible, self.dismissesOnOutsideClick else { return }
    if event.windowNumber != panel.windowNumber, !Self.mouseIsInside(panel) {
        self.hide()
    }
}
```

Same shape, and `dismissesOnOutsideClick` is `!isPinned && !exportModalActive` (`ScratchpadSupport.swift:172`), so an unpinned pad — the default — closes on the first key. It holds an `NSTextView` (`:38`) and a `keyDown` local monitor, which makes it a typing panel like the snippet library rather than a click target like the launcher. Of the six it is the one whose entire purpose is typing into it, so it is the worst remaining instance, not the mildest. It has no `didActivateApplicationNotification` observer, so it needs only the `ownsCocoaPoint` guard — three lines.

`Services/QuickTools/CameraPreviewService.swift:348` is the seventh, same monitor pair plus the activation observer at `:356`. Not a typing surface, so it costs an Accessibility Keyboard user a dismissed mirror rather than an unusable feature. Same three lines plus the bundle-ID exemption. Neither of these is dock preview or quick toggles, so your own untouched list still stands separately.

One thing you got right that my earlier comment got wrong: I said the four panels use a `localClickMonitor` / `outsideClickMonitor` pair, and you patched only the global one. That is correct. A click on the keyboard's window is delivered to AssistiveControl, so a local monitor never sees it; guarding the local monitors would have been dead code.

Not verified: nothing on hardware; and whether the Accessibility Keyboard can make the scratchpad panel key at all — I read the monitor wiring, not the behaviour, so if it cannot, that panel is broken for a different reason and this patch will not be enough on its own.

**Impact:** an unpinned scratchpad closes on the first key from that keyboard, the same way #897's library did. Everything actually in this diff is correct as written.
**Merge:** hold for #898 — unchanged; the rebase should pick up `ScratchpadService` on the way through.

ref #920
ref #897

fermincasagrande:
Both added. Six services, not four.

**`ScratchpadService`** — the click guard only, since it has no `didActivateApplicationNotification` observer. You are right that it is the worst remaining instance rather than the mildest: an `NSTextView` behind a `keyDown` monitor, `dismissesOnOutsideClick` defaulting true through `!isPinned && !exportModalActive`, so a panel whose whole purpose is typing into it closed on the first key. Same failure as #897 on a surface where it is more obviously fatal.

**`CameraPreviewService`** — guard plus the bundle-ID exemption on its observer.

Neither is dock preview or quick toggles; those remain unaudited and unmentioned in the diff.

**On the local monitors:** thank you for correcting that in my favour rather than leaving it. I had reasoned that a click on the keyboard's window is delivered to AssistiveControl so a local monitor never sees it, but I had not said so anywhere, and a reviewer finding four guarded globals and four untouched locals would reasonably read it as an oversight.

---

I should have found these two myself. `grep -rl "outsideClickMonitor = NSEvent.addGlobalMonitorForEvents" Sources/` returns all six, and I ran it for the first time this round — after you had named both. Two rounds of you enumerating call sites that one command would have listed. The sweep is done now, and the same applies to the `ignoresMouseEvents` set in #898, where I took the pid-level fix rather than working through the list.

Merge order unchanged: **#898 first**, then I rebase this. Its base cannot be retargeted — a base must be a branch in this repository and mine is in the fork — so the diff will keep showing #898's commits until that lands.

Confirmed on hardware: switcher, snippet library and camera preview. Scratchpad is off on the reporter's machine, so that one is reasoned rather than observed, and your open question about whether the Accessibility Keyboard can make that panel key at all is still open — if it cannot, this patch is necessary but not sufficient there.

PathGao:
Grepped `main`: `mouseIsInside` plus a global mouse-down monitor returns exactly six services, and all six are guarded here. That sweep is complete.

**The launch warm-up only covers "keyboard already running at app start."** Vorssaint is a login item; the Accessibility Keyboard usually gets switched on later. `cachedPID` is still nil then, so the first `ownsPoint` returns false and merely schedules the resolve — the second Tab cancels the session, and only the *next* one works. Your hardware run wouldn't show it: the snippets tap primes the cache on any click.

`didLaunchApplicationNotification` would close this, but the helper's own doc says Assistive Control doesn't reliably announce launching. If that premise holds, so does the hole. Warm where the caller has no second chance:

```swift
// AppSwitcher.beginPendingSession
_ = AssistiveKeyboard.isRunning
```

**The body is stale again:** still "the switcher and the four panels", and Not-verified omits Scratchpad, which your last comment calls unobserved.

Smaller: nothing calls `dismissForClickOutsidePanel` without an event, here or on `main`, so the body's "any other caller is unaffected" describes a caller that doesn't exist.

Not verified: nothing on hardware, and whether `didLaunchApplicationNotification` fires for AssistiveControl — I read your doc comment, not the behaviour.

**Impact:** with snippets off, the first switcher session after switching the keyboard on still dies on the second Tab.
**Merge:** hold for #898 — unchanged.

ref #920
ref #897

fermincasagrande:
All four taken. Rebased first, since #898 landing changed what this diff shows.

**Rebase.** #898 was squash-merged, so its three commits are not ancestors of `main` and this branch still carried them — the diff shown here was not the change being proposed. Rebased with `--onto`; it is now seven files, the switcher and the six panels, and nothing of #898. The two panel commits replay byte for byte — 54 changed lines, identical content, only hunk positions moved because `main` has edited those files elsewhere — so the earlier hardware results stand unchanged. The hold you had on this is lifted from my side.

**The warm-up hole: you are right, and it is fixed.** `_ = AssistiveKeyboard.isRunning` in `beginPendingSession`, after the accessibility guard and before the enumeration. Your reading of the sequence matches mine — `cachedPID` nil, the second Tab's `ownsPoint` schedules the lookup and answers false, `cancelSession()` runs, and only the *following* session works.

**And you are right about why my run did not show it.** This is the second time that exact masking has caught this branch: round 2 was a switcher test that passed only because the snippets tap had primed the cache. Same helper, same cause, and I did not generalise from it when it was in front of me.

**Not verified, and I would rather be plain about the shape of that.** I have observed neither the failure nor the fix on hardware. Reproducing it means disabling snippets, quitting the keyboard, restarting the app and switching the keyboard back on — and snippets are a large part of what makes this Mac usable, so that is an awkward state to enter and leave. The change also rests on the premise you quoted, which is my own doc comment: that Assistive Control does not reliably announce launching. I have read that comment; I have not measured the behaviour. If the notification does fire dependably, this warm-up is redundant rather than wrong, and it costs a lock and at most one `kill(pid, 0)` on a path that is already doing window enumeration.

**Body rewritten.** Six panels rather than four, with the `grep` that returns exactly those six. Scratchpad and Camera Preview are now named in *Not verified* — Scratchpad is unobserved here, as I said last time. The stacked-on-#898 section is replaced by what the rebase did.

**The default is gone:** `dismissForClickOutsidePanel(_ event: CGEvent)`. You were right that it described a caller that does not exist. I added it so the existing call site would not have to change, then changed that call site in the same commit.

**Impact:** unchanged. With snippets off, the switcher and all six panels are unusable on this keyboard; Scratchpad is the one I would weight most heavily, being a text editor that closed on the first character.

Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
```diff
@@ -458,7 +458,10 @@ final class ScratchpadService: NSObject, ObservableObject, NSWindowDelegate {
         }
         outsideClickMonitor = NSEvent.addGlobalMonitorForEvents(matching: mouseEvents) { [weak self, weak panel] event in
             guard let self, let panel, panel.isVisible, self.dismissesOnOutsideClick else { return }
-            if event.windowNumber != panel.windowNumber, !Self.mouseIsInside(panel) {
+            if event.windowNumber != panel.windowNumber, !Self.mouseIsInside(panel),
+               // Every key on the Accessibility Keyboard is a click outside this
+               // panel. Dismissing on those makes the panel impossible to type into.
+               !AssistiveKeyboard.ownsCocoaPoint(NSEvent.mouseLocation) {
```
vorssaint, line 464: with automatic text expansion off, turning on the accessibility keyboard after vorssaint and typing into the scratchpad can still close it on the first key. the other panels have the same gap. keep the first key working without requiring users to reopen the panel or use the switcher first

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -1563,8 +1571,13 @@ final class AppSwitcher: ObservableObject {
...
     /// release from committing the highlighted window first (issues #384 and
     /// #539).
-    private func dismissForClickOutsidePanel() {
+    private func dismissForClickOutsidePanel(_ event: CGEvent) {
         guard sessionActive, let panel else { return }
+        // On the Accessibility Keyboard a modifier is latched by double-clicking
+        // it and every other key is then pressed with the mouse. Those clicks
+        // land outside the panel, but they are the user driving the switcher,
+        // not dismissing it: without this the session dies on the second Tab.
+        if AssistiveKeyboard.ownsPoint(event.location) { return }
```
vorssaint, line 1615: if tab is clicked again before the switcher has finished preparing, it cancels the opening instead of keeping that step. the protection needs to cover that wait as well, so repeated clicks advance through the windows and ordinary outside clicks still cancel

Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
fermincasagrande: Confirmed, and fixed for all six.

`installMonitors` is where each panel wires up the monitor that asks
`ownsCocoaPoint`, and it runs on every `show()` — so that is where the warm-up
went: `_ = AssistiveKeyboard.isRunning`, six sites, one per panel
(`ClipboardHistoryService` calls its own `installDismissMonitors`, same shape).

**Why the panels needed their own and the switcher's was not enough.** The
switcher has a second chance: `beginPendingSession` sits between the first Tab
and the second, so the lookup has somewhere to happen. A panel has no gap. The
first call to `ownsCocoaPoint` *is* the first key the user presses, so the panel
is already hidden by the time the resolved pid lands. Warming as the panel opens
spends the interval between opening it and reaching for a key on the lookup,
which is the only slack the sequence has.

You are also right about the precondition. `applicationDidFinishLaunching` warms
this only for a keyboard that was already running, and Vorssaint is normally a
login item, so the usual order is app first, keyboard afterwards — `cachedPID`
nil at exactly the moment a panel first opens.

**And right that expansion off is what exposes it.** With it on,
`TextSnippetService`'s tap calls `ownsPoint` on every mouse-down and keeps the
pid resolved, so every panel inherits a warm cache it never asked for. That is
the third time in this thread that the snippets tap has hidden a cold-start hole
in this helper — the switcher's was the same masking, twice. I should have
treated "which callers can be first?" as the general question after the second
one rather than fixing the site that was pointed at.

---

**Not verified on hardware.** Reproducing it means turning text expansion off on
this machine, and that is a large part of what makes it usable, so I have
observed neither the failure nor the fix. The reasoning is above and the change
is one line per panel on a path that already existed. Same standing as the
switcher warm-up in the previous round, and I would rather say so than let six
call sites imply more testing than happened.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
fermincasagrande: You are right, and this one was a hole I had not seen at all: the guard I added
is never reached on that path.

**Where it goes wrong.** `route()` handles the still-preparing case before
`handle()` ever runs:

```swift
if type == .leftMouseDown || type == .rightMouseDown || type == .otherMouseDown || type == .otherMouseUp {
    let stillInactive = routeLock.withLock { () -> Bool in
        guard !routeSessionActive else { return false }
        routePendingSessionStart = nil          // the opening, discarded
        return true
    }
    if stillInactive { return Unmanaged.passUnretained(event) }   // returns here
```

`dismissForClickOutsidePanel` guards on `sessionActive`, and during the wait the
session is not active — so the check I added for the keyboard sits behind a
branch this event never takes. The mouse-down is gone before it.

**What that costs.** Enumeration waits on bounded AX calls, so the second Tab's
mouse-down can land while the start is still pending. It wipes
`routePendingSessionStart`; the `keyDown` that follows then finds nothing
pending, so it falls to the `else` branch, bumps `sessionStartGeneration` and
creates a fresh start — and `beginPendingSession` discards the enumeration
already in flight on the old generation. The step that Tab was asking for goes
with it, instead of being accumulated at the `pending.scope == requestedScope`
branch as it would be for a keyboard that sends a bare `keyDown`.

**The fix, guarded rather than removed.** The pending start is now kept only when
the click landed on the keyboard's own panel:

```swift
let startPending = routeLock.withLock {
    !routeSessionActive && routePendingSessionStart != nil
}
let clickPressedAKey = startPending && AssistiveKeyboard.ownsPoint(event.location)
let stillInactive = routeLock.withLock { () -> Bool in
    guard !routeSessionActive else { return false }
    if !clickPressedAKey { routePendingSessionStart = nil }
    return true
}
```

Anything that is not a key press still discards the pending start exactly as
before, so an ordinary outside click during the wait cancels the way it always
has — which was your other requirement and the reason not to just drop the line.

**On cost.** `ownsPoint` is only evaluated when a start is actually pending, so
the window scan stays off the mouse-down path of everyone who is not mid-open.
For a machine with no Accessibility Keyboard running it is cheaper still —
`livePID` answers from the cache and there is no scan at all. That keeps the
gate where you and the earlier review both wanted it rather than at the call
site. Reading it under the lock and evaluating the predicate outside also keeps
`routeLock` off a window enumeration on the tap thread.

---

**Not verified on hardware.** The failure is a race against enumeration, so
provoking it reliably means slowing that path down deliberately; I have read the
sequence, not measured it. The two Tab clicks and an ordinary outside click
during the wait behave correctly here, but on this machine the window is short
enough that I cannot claim I hit the pending state on purpose. Build is clean and
`--selftest` passes.

Rebased onto current `main` while I was in here. Five commits, seven files.

Issue #123, opened 2026-06-29
[Bug]: Not all running apps listed in "quit on close" (suggestion included)


fexscruta:
### Description

Today, I wanted to add Signal to the list of apps that should not be quit after the window is closed. Signal was running but it did not show up in the app picker so I can not add it atm. 
Another shortcoming in this regard is that an app that I want to add to the exclusion list needs to be running. 

I think, in general, it would be a good addition to have an app picker that lists all installed apps. This way, the exclusion list could be made once – set and forget. Right now, I need to have all the apps that I want to be in the exclusion list running. This approach needs a lot of clicks (and some brain work). 

### Steps to Reproduce

1. Have an app running that should go into the exclusion list
2. Go into Vorssaint's settings
3. Under "Exceptions", click on Add app
4. Signal does not show up (for me)

### Vorssaint Version

3.1.4

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

vorssaint:
hi!

this makes sense. adding next update

thanks for the feedback!

PR #923, opened 2026-08-24
ci(issues): move a fix from merged to released without anyone remembering to

PathGao:
Two jobs in one workflow.

**`fixed-on-main`** — fires on a merged pull request. Reads `Refs #123` and the
other reference verbs out of the description, labels those issues `fixed
(pending release)`, and records the merge commit in its own comment. It says
where the change is, not that anything is fixed.

**`reporter-check`** — fires after the Release workflow finishes a stable tag.
For each `fixed (pending release)` issue whose recorded commit `git
tag --contains` puts in that tag, it swaps the label to `v<version> please
confirm` and asks the reporter to update to that build and say how much of
their report it fixed.

So the first half moves an issue from *merged* to *waiting for a release*, and
the second from *released* to *in your hands*. Neither decides anything: a
human decided at merge time, and git answers the version question.

Both labels and both comments are read by the person who filed the report, and
most of them do not work on software. Nothing they see says `main`, a commit or
a merge: the change is in the development build, and it is not on their Mac
until a release says it is.

## What you are approving

- **`issues: write`**, declared in this file and scoped to it. The rest of CI
  keeps `contents: read`.
- The labelling job takes **no checkout**. It runs in the base repository
  context under `pull_request_target`, so it must never execute anything that
  arrived with the pull request, and it does not.
- **`release.yml` is untouched.** A step at the end of the release job would
  have been the obvious alternative, but that file carries the signing secrets
  and the protected environment.
- Not `on: release: published`. The release is created by `gh release create`
  with the built-in token, and an event raised by that token does not start
  another workflow run, so that trigger would never fire. Hence `workflow_run`,
  plus an API check that a non-draft, non-prerelease release exists for the tag
  before anyone is written to.
- The two weeks are stated in the comment and enforced nowhere. The expiry
  sweep stays manual, because closing an issue is a decision and this makes
  none.

If the labelling job misbehaves, the blast radius is one label and one comment,
both reversible. Neither trigger can fire from a fork, so the first real
exercise is the first merge after this lands.

### DIFF
--- a/.github/workflows/issue-fix-status.yml
+++ b/.github/workflows/issue-fix-status.yml
@@ -3,14 +3,20 @@ name: Issue fix status
 # Two mechanical steps, no judgement in either.
 #
 #   a merged pull request that references an issue  ->  the issue gets
-#   `fixed on main`, with the merge commit recorded in the bot's comment
+#   `fixed (pending release)`, with the merge commit recorded in the bot's
+#   comment
 #
 #   a stable release that contains that commit      ->  the label becomes
-#   `v<version> reporter check` and the reporter is asked to confirm
+#   `v<version> please confirm` and the reporter is asked to check it
 #
 # Neither step decides that anything is fixed. A human decided that when they
 # merged. The label only says where the change is, and `git tag --contains`
 # answers the second one, so a changelog entry is never the evidence.
+#
+# Both labels and both comments are addressed to the person who filed the
+# report, and most of them are not developers. Nothing in what they read says
+# `main`, a commit, or a merge: the change is in the development build, and it
+# is not on their Mac until a release says it is.
 
 on:
   pull_request_target:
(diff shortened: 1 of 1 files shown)


vorssaint:
love the idea @PathGao!

one quick thought, a lot of people opening issues here aren't developers and might not even know what "main" means or might get confused thinking the fix is already live on their app. a lot of people (and i think the majority of them) open issues because they-re non-technical, don't understand project terms

could we tweak the labels and wording to be a bit more user friendly and less technical? for example using something like "fixed (pending release)" instead of "fixed on main", and making the message plainly say that a fix was added to the dev version and is queued for an upcoming release? i think suits best

PathGao:
You are right, and it was worse than the wording: the whole thing was written from the repository's side of the fence. Pushed a rewrite that says it from the reporter's.

`fixed on main` is now **`fixed (pending release)`**, exactly as you put it. I renamed the existing label rather than making a new one — nothing was using it yet, so no issue changed.

`v<version> reporter check` is now **`v<version> please confirm`**. That one only ever appears on an issue addressed to the person who filed it, so "reporter" was the same problem one step later. Say the word if you would rather keep it.

The merge comment no longer mentions `main`, a commit or a merge:

> A fix for this went into the development version of the app in #123, and it is queued to go out in an upcoming release.
>
> **Nothing has changed on your Mac yet**, and this is not us saying your report is fixed — that part is up to you. When a release goes out carrying this fix you will get a note on this issue with the version to install, and two weeks to say whether it worked for you.

And the release comment asks for an update rather than an install, and drops the line about reading the code:

> The fix for this is out now, in **3.3.3**.
>
> @reporter, could you update to that version and say whether it fixed things on your Mac? We cannot tell from here — yours is the setup that found this.

Nothing else moved: same triggers, same permissions, same `git tag --contains` as the evidence. The two paragraphs asking people to say *which part* is fixed are untouched, since that is the answer the whole workflow exists to collect.

Issue #926, opened 2026-08-24
Choose which key becomes the Super key, not just Caps Lock


JoanLaRosa:
### Problem or Motivation

The Super key can only be Caps Lock. Mine is already Control, and my hyper key sits on Right Command, so I can't use the feature at all.

That's the other half of #594, which asked for a custom combo "or even replace a different key". The combo shipped in 3.2.0. The source key didn't.

### Proposed Solution

Let the user pick the key, defaulting to Caps Lock so nothing changes for anyone else. Right Command, Right Option, Right Control, Right Shift are the useful additions, since they're usually redundant with their left twins. That's why Hyperkey and Raycast offer them.

Looks like most of the work is already done. The mapping goes from a source HID usage to F18 and the tap only matches the F18 key code, so the source is the only fixed part: `capsLockUsage` and `capsLockKeyCode`, plus the five functions in `SuperKeySupport` that name Caps Lock.

Two things that aren't just a find and replace:

- The `capsLock` solo action makes no sense when the source isn't Caps Lock. Nothing was taken over, so there's nothing to hand back.
- `panelCaptionFormat` ("Caps Lock holds %@.") and `superKeyAlternative(capsLockLabel:)` need the chosen key's name, which means all 13 locales.

Caps Lock needs the F18 mapping because a lock key never reports down and up. A right-hand modifier doesn't have that problem, but I'd map it anyway so there's one code path and the modifier can't fire on its own.

Happy to send a PR if you want it.

### Feature Area

Mouse and keyboard

### Mockups / References

Hyperkey converts "the caps lock key or any modifier key". Raycast stores it as one HID usage, 231 for Right GUI. Karabiner does it through complex_modifications, which is what #594 linked.

vorssaint:
you can now choose right command in settings > super key > key to hold, then select the modifiers you want it to represent. right option, right control and right shift are available there too. install super key from settings > features if its page is hidden, and keep the chosen key at its default action in macos modifier keys. this is in 3.3.5, so i'm closing the request

PR #928, opened 2026-08-24
feat(super-key): choose the source key

JoanLaRosa:
Refs #926.

## What it does

Adds a source picker for the Super key with Caps Lock, Right Command, Right Option, Right Control and Right Shift. Caps Lock remains the default, so existing setups do not change.

The selected source goes through the existing F18 mapping and event-tap path. Mapping recovery also records the active source, so changing sources or recovering after a crash removes only the mapping Vorssaint owns.

The settings diagram, menu panel and shortcut alternatives now show the selected source. The Caps Lock solo action is only offered when Caps Lock is the source.

## Checks

- [x] `./build.sh --test` passes, 8392 checks
- [x] `./build.sh` finishes without warnings
- [x] `./build/Vorssaint --selftest` reports `SELFTEST OK`


JoanLaRosa:
### Super Key source UI

The source is now an explicit **Key to hold** picker. The selected key is also reflected in the diagram below, so changing the Super Key does not require editing defaults manually.

Super Key settings with Right Command selected

Available Super Key source options

abhicommands:
hello Can you also make it so that we can rebind it to whatever key we want for example for gaming purposes or other purposes caps lock is so useless that maybe we can use it with other things and maybe a rebind alongside of super is possible?

PathGao:
Four things the source-key generalization does not reach.

**1. The Modifier Keys pre-flight is probably dead for every new source.**
`SuperKeySupport.swift:221-225` and `:208-215` compare the pane's rules against `source.usage`, which for the four new cases is `0x7000000E4`–`0x7000000E7` (right-side HID usages). But this repo's own model of what System Settings > Keyboard > Modifier Keys writes uses the left/generic usage for a family: `Tests/MetricsTests.swift` builds `disabledControl` from `0x7000000E0`, not `0x7000000E4`. If the pane really writes only the generic usage, then for `source == .rightCommand` neither `modifierMappingsAllowSuperKey` nor `canMapNoAction` can ever match, so a user who set Command → No Action in that pane and picks Right Command gets `E7 → F18` written and `isRunning == true` (`SuperKeyService.swift:426`) while the key does nothing — the Modifier Keys layer already swallowed it before `UserKeyMapping`, which is exactly the case the Caps Lock path refuses or recovers via the no-action sentinel.

I could not run `hidutil` in this review, so this depends on runtime behavior I did not observe. Please check `hidutil property --get "HIDKeyboardModifierMappingPairs"` after setting Command → No Action in that pane. If it writes only `0x7000000E3`, give `SuperKeySource` a second usage (the family's pane usage: `E0` control, `E1` shift, `E2` option, `E3` command; Caps Lock stays `0x39`) and have both `modifierMappingsAllowSuperKey` and `canMapNoAction` test the pane usage in addition to `usage`. Your new assertion at `Tests/MetricsTests.swift` that `canMapNoAction(from: [disabledRightCommand], source: .rightCommand)` is true only proves the generalization is self-consistent, not that it sees what the OS writes.

**2. The solo-action picker can land with no selection.**
`SuperKeySettings.swift:61-63` drops the `.capsLock` tag when the source is not Caps Lock, but `soloBinding`'s getter at `:200-201` still returns whatever is stored. Only the source picker's setter (`:169-178`) rewrites the stored value, so a `superKeySource = rightCommand` + `superKeySoloAction = capsLock` pair arriving any other way leaves a radio group with no matching tag (blank), while `SuperKeyService.swift:100` silently runs `.none` and never corrects the stored value. That pair is reachable through a restored settings backup: both keys are registered, so both travel (`SettingsBackupSupport.swift:18-24`).

Simplest fix is to delete the `if source == .capsLock` gate entirely. `.inputSource` is offered unconditionally at `:64` and its long-hold still resolves to `.capsLock` for every source (`SuperKeySupport.swift:279`), with a string that says so, and `SuperKeyService.setCapsLock` works whatever the source is — so tapping Right ⌘ to toggle Caps Lock is a coherent binding, not one to hide. If you want to keep it hidden, sanitize in the getter at `:200` instead of only in the setter.

**3. `README.md:115` still reads "Hold Caps Lock and it counts as the modifier combination you choose."** That is the one user-facing description of the feature outside the app and it is now wrong.

**4. Three Caps Lock glyphs left behind.** You made the diagram symbol conditional at `SuperKeySettings.swift:91`, but `FeatureCatalog.swift:127`, `SettingsDirectory.swift:157` and `MenuPanelView.swift:1647` still hand out `"capslock"`, so a Right ⌘ super key shows a Caps Lock key in the feature hub, the settings search results and the menu panel row. Either make those follow the source too, or leave the diagram unconditional so the four agree.

JoanLaRosa:
Addressed the four review points and merged current `main` to clear the conflict.

1. Current `main` removed the `HIDKeyboardModifierMappingPairs` pre-flight in #985 after confirming that hidd keeps that property per client connection, so `hidutil` cannot observe the rules written by System Settings. This branch now follows that upstream behavior instead of comparing either E3 or E7 against an unreadable table, while preserving the new source through mapping ownership and failure reporting.
2. Caps Lock is now always available as a solo action, and the service honors the stored action for every source.
3. The README now describes Caps Lock or a right-side modifier as the held key.
4. The feature hub, Settings sidebar/search, and menu panel icons now follow the selected source, including live sidebar updates.

I also made the new mapping-failure copy source-neutral and rebased the source-key tests onto current `main`. The full release build passes. The unit harness compiles and runs all 8,608 checks; on this host, the existing dispatch-pool-starvation prerequisite from current `main` did not reproduce and was the sole failed check.

JoanLaRosa:
Follow-up on point 1 after reproducing the read locally: both `hidutil property --get HIDKeyboardModifierMappingPairs` and the keyboard-matched form return `(null)`, matching the per-client behavior documented in #985. There is no reliable table for this process to compare against.

Since #985 identifies this as a documentation requirement, commit `2220a54` now shows a localized note directly on the Super key page: keep the selected key at its default action in System Settings > Keyboard > Modifier Keys. The README carries the same requirement. This covers Caps Lock and every right-side source without pretending the app can detect a setting macOS does not expose.

PathGao:
All four are in the diff and hold up. Point 1 landed somewhere better than what I proposed: if the pane's table is unreadable from this process, a note beats a check that silently never matches.

Four things left, none blocking.

**`FeatureCatalog.swift:127-129` reads `UserDefaults.standard` inside a computed property.** SwiftUI does not invalidate on that, so the hub icon follows the source only when its view is rebuilt for some other reason. You listed live updates for the sidebar and not the hub, so you may already know this. The hub view is not in this diff, so I could not check whether something there observes the key.

**The enable toggle lost its label.** `enableToggle` is deleted from all 13 locales and `SuperKeySettings.swift:35` now passes `text.pageTitle`, so the toggle reads "Super key" inside a section titled "Super key". The body does not mention this. A source-neutral string ("Use this key as the super key") keeps the row saying what it does.

**One preference, seven readers, three mechanisms.** `superKeySource` is now read through `UserDefaults.standard` (`FeatureCatalog.swift:128`, `SettingsDirectory.swift:24` as a default argument), `@AppStorage` (`SettingsView`, `MenuPanelView`, `SuperKeySettings`) and `SuperKeyService.source` (`ShortcutsSettings`, `ShortcutRecorderButton`). The published `source` already exists and already drives two of them, and the two `UserDefaults` readers are exactly the ones that can go stale. Unless the catalog is deliberately kept free of service dependencies, they can read it too.

Minor: `SuperKeyService.swift:412` drops the comment explaining why a failed repair keeps the tap up instead of calling `stop()`. Nothing else in that hunk changed but a ternary reflow. That comment is the only record of why the branch is asymmetric.

**Impact:** none on key behavior. Worst case is a stale glyph in the feature hub until its view rebuilds.
**Merge:** ship it.

JoanLaRosa:
Finished the non-blocking follow-up in `3538ae0`.

- The feature hub now observes `superKeySource` and passes the selected source icon into its row, so the glyph updates immediately instead of waiting for an unrelated rebuild.
- The Settings directory default now uses the published `SuperKeyService.source` rather than another direct defaults read.
- The enable row has a distinct source-neutral label in all 13 locales, with completeness and non-duplication coverage.
- Restored the rationale for keeping the event tap alive after a failed repair.

Validation after the final diff:
- release build passed
- app `--selftest` passed
- helper self-test and deep bundle signature verification passed
- 8,607 of 8,608 unit checks passed; the sole failure remains the unrelated current-main dispatch-starvation prerequisite that does not reproduce on this host

JoanLaRosa:
Merged the new upstream `main` commit (`ad88376`) and reran the integrated release path. The release build, helper self-test, app `--selftest`, and deep signature verification all pass.

JoanLaRosa:
Final distribution-boundary check on the exact integrated branch also passed: `Vorssaint-3.3.3-beta.2.dmg` verified with `hdiutil`, mounted read-only with its Applications link, and the packaged app/helper self-tests plus deep signature verification passed. SHA-256: `1488e5a905fdcadcfadb04a301c971ff8f9bf1c8788d3510ba1a4a925756bb65`.

JoanLaRosa:
Merged current `main` into the branch in `b4988c4` and resolved the Super Key service overlap by preserving both behaviors: the selected source still drives modifier classification, and the new upstream mouse-down path still carries held modifiers for drag chords.

Validation on the exact pushed commit:
- release build passed without warnings
- app `--selftest` passed
- fan-control helper `--selftest` passed
- deep bundle signature verification passed
- unit harness passed 8,716 of 8,717 checks; the sole failure is the existing dispatch-pool-starvation prerequisite not reproducing on this host
- `Vorssaint-3.3.3-beta.3.dmg` built and verified, mounted read-only with the Applications symlink, and both packaged public self-tests plus deep signature verification passed
- DMG SHA-256: `526118a77979fcd6531df012674e6c7edfc1ebf327a4dd3eb841ce49733686dd`

JoanLaRosa:
Direct running-app acceptance check on `b4988c4`:

- Opened the local release build's actual Vorssaint Settings window and observed the Super key page with the distinct enable label, source-neutral Modifier Keys note, and Caps Lock still available as a solo action.
- Changed the persisted source from Right Command to Right Option while the app was running. The visible `Key to hold` value, keyboard diagram, and Settings sidebar icon all changed immediately to Right Option, confirming live observation across the preference and SwiftUI boundaries.
- Restored the original Right Command value and observed all three public surfaces return to Right Command.
- The app's real permission card reports Accessibility `Not granted`, and its permission popover remains at `Waiting for the permission…`. Therefore successful HID remapping and held-key behavior cannot be observed on this host without the user granting that macOS permission. I am not claiming that blocked positive path passed.

PR #930, opened 2026-08-24
fix(menu-bar): keep popover anchored when fullscreen menu bar hides

pergioa:
## Summary

- preserve the popover opening anchor when a fullscreen app hides the menu bar
- keep the arrow pointed at its saved screen-space location across content resizes
- leave ordinary visible-menu-bar placement and movement under AppKit control

## Behavior

When the menu bar is hidden, AppKit can park the status item at its fallback location and reuse that location during a later popover resize. The panel now switches to a transparent screen-space positioning view only in that state, so both the panel and its arrow remain at the opening anchor. The positioning view is released when the popover closes.

This is the anchoring-only replacement for the corresponding portion of #912.

## Testing

- Tests pass: 8,394 checks
- Full release build passes


PathGao:
The positioning-view swap is the right shape for the arrow, but as written it re-enters `popover.show` with none of the guards this file already uses for that, and it reaches only one of the two untrustworthy-frame cases.

**1. `!NSMenu.menuBarVisible()` is a second answer to a question this file already answers.** `AppDelegate.swift:658` gates the swap on menu-bar visibility, while `applyPopoverDriftFrame` gates the same decision on `anchor.overridesSoundFrame || !frameStillDescribesMenuBar(anchor)` (`AppDelegate.swift:639`, helper at `:533`, backed by `StatusItemAnchorSupport.isTrustworthyStatusFrame`). They disagree on the other half of the shape documented in `StatusItemAnchorSupport.swift:5-18`: the macOS 27 case where the bar is up but the item's window frame is stranded at the slot it was born in. There `overridesSoundFrame` is true, `applyPopoverDriftFrame` moves the window, and the arrow keeps pointing at the parked positioning view — exactly the failure your comment at `:651-654` describes, still unfixed. Use the `:639` predicate at `:658` and both cases get the stable view, with one answer to the question instead of two.

**2. The re-show at `:684` is unguarded.** `reanchorMetricPopover` re-shows an already-shown popover too, and it sets `popoverIsSwitchingAnchor = true` / `setSwitchingMetricAnchor(true)` first (`:404-405`) and re-arms drift correction against the window the show produced (`:412-414`). Without that, if the re-show makes `popoverDidClose` (`:990`) fire, `:1003-1004` run `removePopoverDismissMonitor()` and `endPopoverDriftCorrection()` — the latter orders out and nils `popoverPositioningPanel` two lines after `:682` assigned it. The popover is then shown against a view whose window is ordered out and has no remaining strong reference, with outside-click and Escape dismissal gone, no drift observers left, and `popoverAnchor` nil; `MenuPanelFocus.clearMetricFocus()` at `:991-993` would also discard the metric tab the user was on. I could not run this, so I cannot confirm AppKit closes on a re-show — but `:404` already assumes it does. Mirror it: set the switching flags around `:684`, keep the panel reference alive across that teardown, then re-run `beginPopoverDriftCorrection` and `installPopoverDismissMonitor` against `popover.contentViewController?.view.window`.

**3. The re-show drops the fullscreen window setup.** `showPopover:754-756` inserts `.fullScreenAuxiliary, .canJoinAllSpaces` and sets `hidesOnDeactivate = false` on the popover window, with the comment "it blinks shut when another display is fullscreen" — the only condition your code runs in. `:684` does not repeat it. If the re-show yields a new window that is lost; if it is the same window, re-applying costs nothing.

**4. Once the panel exists nothing recomputes it.** `:657` returns early on every later resize, so the 1x1 rect keeps the y and the clamp computed from the window size and `visibleFrame` at swap time. A taller tab, or the anchor's display being unplugged while the panel is open, no longer passes through `pinnedPanelFrame` / `anchorVisibleFrame` the way `applyPopoverDriftFrame` did on `main`. Meanwhile `didMove` still falls through to `applyPopoverDriftFrame` at `:616-617`, so the anchor-midX frame and AppKit's placement from `tipX` are both moving the same window — and those two differ precisely when the popover is clamped near a screen edge, which is why `tipX` exists. Make `:657` reposition the existing panel from the current window size and visible frame instead of returning, and early-return the whole handler, `didMove` included, while `popoverPositioningPanel != nil`.

**5. Sibling site, not asking you to widen this PR.** The docked Shelf reads the same raw status item frame: `AppDelegate.swift:96-100` hands `button.window.frame` to `ShelfService.statusItemFrameProvider` with no trust check, and `ShelfService.positionDocked:902-906` and `mouseNearDock:814-818` place the pill and its hover band from it, so a hidden bar sends both to the parked coordinate. `StatusItemController.containsStatusItem:59` has the same unchecked read, so the outside-click monitor stops recognising a click on the icon in that state. Gating the provider at `:96` on `isTrustworthyStatusFrame` would route both Shelf callers to their existing nil fallbacks — worth a separate PR.

pergioa:
Thank you for the detailed review. Addressed in `a883169`:

1. The stable positioning view now uses the same `statusFrameNeedsAnchorOverride` predicate as drift correction, covering both a hidden fullscreen menu bar and the macOS 27 stranded-frame case.
2. The re-show now sets the switching guards, keeps the positioning panel alive across `popoverDidClose`, then re-arms drift observers and reinstalls the dismissal monitors against the resulting popover window.
3. Fullscreen/Spaces window configuration is centralized and reapplied after every re-show.
4. An existing positioning panel is recomputed from the current popover size and visible frame on both resize and move; while it exists, the handler no longer also applies the competing midX frame correction.

I kept the Shelf and status-icon hit-testing observation out of this PR and filed it separately as #940.

Validation: 8,394 checks pass and the release bundle builds successfully.

PathGao:
Three things left, one of them I can't settle from the diff.

**1. The positioning panel is ordered out but never closed.** `endPopoverDriftCorrection` does `popoverPositioningPanel?.orderOut(nil)` and then nils the reference, and the failure branch in `useStablePopoverPositioningViewIfNeeded` does the same. With `isReleasedWhenClosed = false`, ordering out does not take the window off `NSApp.windows`, so dropping your own last reference doesn't free it — one stranded `NSPanel` per open/close cycle in the untrustworthy state, which on a stranded-frame Mac is every open. Use `close()` in both places instead. At both call sites the popover is either already closed or about to be re-shown against the button, so nothing is still hanging off that view.

**2. `installPopoverDismissMonitor()` at the end of the swap is unconditional, and I can't tell whether that balances.** It is correct only if `popoverDidClose` actually ran `removePopoverDismissMonitor()` during `popover.show` — that is, only if the delegate method is *not* short-circuited by `popoverIsSwitchingAnchor`. If it is guarded (which is what setting the flag before the show is for), the old monitor is still live and you add a second one. Neither `popoverDidClose` nor `installPopoverDismissMonitor` is in this diff, so I can't check it. If install isn't idempotent, call `removePopoverDismissMonitor()` first — it's a no-op when nothing is installed.

**3. `tipX` equals `midX` in three of its four constructors.** They diverge only in the trusted path (`statusButtonMidX(button)` vs `window.frame.midX`), while the doc comment attributes the difference to AppKit clamping a wide popover near a screen edge — nothing here computes a clamped tip, the stored value is always an icon-x guess. And the swap uses `targetFrame` for `maxY` only and throws away its x, so horizontal placement now comes from `tipX` plus AppKit's own clamp, leaving `midX` in play only for `pinnedPanelFrame`'s vertical clamp. Either that trusted-path difference is the actual reason the field exists, in which case the comment should say so instead of the clamping story, or this is a second copy of one coordinate and `midX` covers it. Unless `statusButtonMidX` and the item window's `midX` diverge somewhere I can't see from the diff.

The rest reads right: one predicate at both decision sites, the panel held across the teardown, both notifications short-circuited while it exists, and an epsilon on the reposition so the panel-moves-popover-moves loop terminates.

**Impact:** one leaked `NSPanel` per popover cycle on affected Macs, plus a possible duplicate dismiss monitor after the first tab switch, which would leave a stale event monitor running after close.
**Merge:** needs the `close()` swap and a confirmation on the dismiss-monitor balance first.

pergioa:
Thank you — addressed the remaining actionable items in `50e98b7`.

1. Both teardown paths now call `close()` instead of `orderOut(nil)`: the retained `popoverPositioningPanel` path in `endPopoverDriftCorrection`, and the locally held panel in the failed re-show path. The reference is then cleared, so the synthetic panel is removed from `NSApp.windows` and can be released.

2. Confirmed that the dismissal-monitor replacement is already idempotent. `installPopoverDismissMonitor()` begins with `removePopoverDismissMonitor()`, which removes and nils the global mouse, local mouse, and keyboard monitors before installing their replacements. I left that code unchanged rather than adding a redundant removal at the call site.

3. Kept `midX` and `tipX` and clarified their lifecycle in the field comments. `midX` is the center of the window AppKit placed and is used by direct frame correction. `tipX` is the status-button center / opening arrow coordinate used by the stable positioning view. In the trusted constructor, `window.frame.midX` and `statusButtonMidX(button)` can differ when AppKit horizontally clamps the popover near an edge; fallback constructors know only one x coordinate and intentionally use it for both.

Validation against the updated branch: 8,541 checks pass and the release bundle builds successfully.

PathGao:
Both `close()` swaps are in the diff, and the `tipX` comment now states the trusted-path difference instead of the clamping story. The monitor balance I still can't verify from here — `installPopoverDismissMonitor` isn't in this diff, so I'm taking the leading self-removal at its word.

Read the rest of the updated diff: one predicate at both decision sites, the panel held across the teardown and closed on every exit, the existing panel recomputed on resize and move with an epsilon so the reposition loop terminates, and window configuration reapplied after the re-show. Nothing to flag.

**Impact:** none.
**Merge:** ship it.

Issue #708, opened 2026-08-20
`Copy text from screen` share same shortcut key with `screenshot` now. They should have separate keys


wiidede:
### Problem or Motivation

`Copy text from screen` and `screenshot` should have separate shortcut keys

### Proposed Solution

`Copy text from screen` and `screenshot` should have separate shortcut keys

### Feature Area

Tools

PathGao:
@vorssaint the reason the two feel merged is narrower than it looks, and it points at something structural worth deciding before anyone writes the fix.

`screenOCRShortcut`, `colorPickerShortcut` and `recorderShortcut` are read in exactly three places each: `Defaults.registerDefaults`, `SettingsBackupSupport`, and `GlobalShortcutRole.storageKey`. No service registers them. The only capture hotkey any service registers is `screenshotShortcut`, in `ScreenCaptureService.syncWithPreferences()` (`ScreenCaptureService.swift:66`). So Settings › Shortcuts still shows three rows for Screen recorder, Copy text from screen and Color picker, still records a combination into them, still warns about conflicts, and still backs them up, while pressing the key does nothing. `migrateUnifiedScreenCaptureShortcut` (`Defaults.swift:1233`) makes clear this was deliberate on the registration side, so what is left is the settings surface promising what the app no longer does.

That is not a one-off. The table and the registrars are two separate things:

- `GlobalShortcutRole` (22 cases) is the table, but it is only consumed by `conflict(for:excluding:)` and the recorder rows. It registers nothing.
- Registration lives in each service, across three Carbon signatures: `'VUQT'` `QuickToolHotkey(id:)` with hand-written ids 10, 12, 14–20, 22–24 and 200+n, `'VUWL'` in `WindowLayoutService`, `'VUTL'` in `HotkeyManager`.

Both directions leak. A role can exist with nobody registering it, which is this issue. And a registered shortcut can be absent from the table: the window layout actions and the per-row Command Bar shortcuts check whether they collide with a role, but no role checks whether it collides with them. The ids are unguarded too, since two services picking the same number would go unnoticed (11, 13 and 21 are already holes).

The change worth making is to let the table own registration: each role carries its own handler and one place registers all of them, so a role with no registrar stops compiling instead of shipping. Folding the window layout actions and Command Bar rows into the same table, or at least making the conflict check symmetric, is a natural second step, and deriving the hotkey id from the case comes for free with it.

[...]

PR #934, opened 2026-08-24
feat(capture): give every screen capture tool its own shortcut

RecoilGaming:
**Before.** The Screen capture page carried one general shortcut that opened the chooser on whatever tool came first, and only the recorder, copy text and color picker could opt into a shortcut of their own — the screenshot tool had none. On the central Keyboard Shortcuts page the group was called "Screenshot" while the single command inside it was called "Screen capture", the other three capture tools sat as separate top-level entries, and expandable groups used the stock disclosure triangle crowding the leading edge.

**After.** Every capture tool — screenshot, recording, copy text and color picker — owns exactly one shortcut that opens the shared chooser already on that tool. The screenshot tool inherits the old general shortcut's storage keys (`screenshotShortcut` / `screenshotShortcutEnabled`, default ⌃⌥⌘4), so an existing combination keeps working unchanged where the screenshot tool is installed. On a setup without it — where the old general shortcut still had a saved combination — a one-shot migration moves that combination onto the first installed capture tool without a shortcut of its own, so the key keeps opening the chooser. On the Screen capture settings page the selected tool's shortcut is edited in the top section where the general one used to live, under the tool's own name. On the Keyboard Shortcuts page the names swap — the group is now "Screen capture" and the command "Screenshot" — and all seven capture-family shortcuts sit under that one group, each keeping its own icon. Expandable groups are now toggled by their whole row, with a rotating chevron on the trailing edge and child rows indented under the header title.

Along the way this removes the now-dead general-shortcut caption string in every supported language, makes `dedicatedShortcut` non-optional so a capture tool cannot exist without a registered shortcut (the class of bug from #708), and deletes the general hotkey so the per-tool table is the single registrar.

Review follow-ups: one flexible spacer per disclosure header row so the count badge sits flush against the chevron, the orphaned-shortcut migration above (run once behind a defaults flag that settings backups deliberately leave out, like the two earlier capture migrations), the Command Bar and menu panel now show each tool's own shortcut hint, an unlisted capture role stays visible on the shortcuts page, the group name no longer collides with “Screenshot” in pt-BR/es/fr, the capture group starts expanded, and both settings pages share one disclosure header that announces its expanded/collapsed state to accessibility (two new strings, translated in all supported languages). `./build.sh` finishes without warnings, `--selftest` passes, and the standalone suite passes 8547 checks.


PathGao:
Reviewed against `main` @ 75ce1840. The registration side is sound: `GlobalShortcutRole.screenshot.savedShortcut` reads the same `screenshotShortcut`/`.screenshotDefault` pair the deleted general hotkey read (`GlobalShortcut.swift:613`, `:640`, `:653`), the four hotkey ids 25-28 collide with nothing (highest hand-assigned is 24, Command Bar starts at 200), `suspend()` now unregisters all four so `SelfUninstall.swift:88` still tears everything down, and replacing the `role == .screenshot` branch in `roleRow` with `FeatureRuntime.shared.sync([role.feature])` is equivalent because `FeatureRuntime.swift:189-191` routes `.screenshot` back into `ScreenCaptureService.syncWithPreferences()`.

Two lists a capture shortcut has to appear in were not updated, and both are wrong in a way this PR causes.

**F1 — Command Bar still shows the screenshot shortcut on three other tools.** `CommandBarCatalog.swift:313` (`action.screenRecorder`), `:336` (`action.screenOCR`) and `:346` (`action.colorPicker`) all pass `roleShortcut(.screenshot)`. Before this PR that combination opened the chooser with no preferred mode, so the hint was at least in the family; now it lands on the screenshot tool specifically, so the bar advertises a key that does not run the command next to it. Change those three to `roleShortcut(.screenRecorder)`, `roleShortcut(.screenOCR)` and `roleShortcut(.colorPicker)`. `:293` is already correct and should stay.

**F2 — Same three hints in the menu panel, plus a hint that now disappears.** `MenuPanelView.swift:809` (`.screenOCR`), `:846` (`.screenRecorder`) and `:865` (`.colorPicker`) pass `shortcutHint(.screenshot)`; same fix. There is an extra loss here that F1 shares: both `shortcutHint` (`MenuPanelView.swift:941`) and `roleShortcut` (`CommandBarCatalog.swift:276`) gate on `role.isAvailable`, and `GlobalShortcut.swift:761-763` narrows `.screenshot`'s `availabilityFeatures` from all four capture features to `[.screenshot]`. So on an install with the screenshot feature switched off in the hub but the recorder (or OCR, or colour picker) on, those buttons showed the capture combination before and now show nothing at all — while the tool's own shortcut is registered and firing. Pointing each surface at its own role fixes this too.

**F3 — `captureRoleOrder` can silently swallow a future role.** `ShortcutsSettings.swift:24-27` is a hand-kept list, `:35` filters the available roles down to it, and `:76` already suppresses `.screenRecorder`/`.screenOCR`/`.colorPicker` from getting their own slot. A capture role added later and not added to that array renders nowhere on the page: it is neither in `captureRoles` nor allowed a top-level row. That is the shape of #708 with the two sides swapped — a registered shortcut with no row instead of a row with no registrar. Build `captureRoles` so the array only sorts: take `availableRoles.filter { Self.captureFeatures.contains($0.feature) }` and order it by `captureRoleOrder.firstIndex(of:) ?? .max`, so an unlisted role lands at the end rather than vanishing. Since the list is `private static` on a `View` the standalone suite cannot reach it; moving it next to `GlobalShortcutRole` would let you assert "every role whose feature is a capture feature is reachable from the page", which is the check that would have caught #708 on this side.

**F4 — "no new translations were needed" does not hold for three locales.** The group header uses `screenCaptureTitle` (`ShortcutsSettings.swift:88`, `ScreenCaptureSettings.swift:47`) and the screenshot row/toggle now uses `pageTitle` (`GlobalShortcut.swift:672` via `ShortcutsSettings.swift:177`, and `ScreenCaptureSettings.swift:128`). Those two strings are identical in pt-BR (`ScreenshotStrings.swift:307` and `:442`, both "Captura de tela"), es (`:724` and `:859`, both "Captura de pantalla") and fr (`:1002` vs `:1137`, "Capture d'écran" differing only in the apostrophe glyph). In those three languages the group header and the first row inside it read exactly the same, which is the confusion the rename was meant to remove in en. Those three need one of the two strings changed; which one, and to what, is a translator's call, but the PR body's claim that no new translations were needed only holds for the other ten.

**F5 — README is stale.** `README.md:134` still reads "**Screen capture.** One shortcut opens a shared selector for screenshots, screen recordings, copying text from the screen and picking a color." README is updated by feature PRs in this repo (e.g. #729, #722, #718). This one needs the per-tool wording.

**F6 — three shortcuts that were visible are now one click away.** `expandedFeatures` starts empty (`ShortcutsSettings.swift:13`) and is `@State`, so it resets on every visit. Before, `.screenRecorder`, `.screenOCR` and `.colorPicker` each had a single role and rendered as a plain top-level row; now all seven sit behind a collapsed header (`:93`). Consider seeding the set with `.screenshot` on first appearance, or expanding when any capture role is enabled, so an enabled recorder shortcut is not hidden by default.

**F7 — two disclosure idioms in the same window, and the new one drops a11y state.** `MonitorPanelConfig.swift:108` still uses `DisclosureGroup(isExpanded:)` with the stock leading triangle, so Settings now has both styles; if the trailing chevron is the direction, that one should follow. Separately, `disclosureHeader` (`ShortcutsSettings.swift:140-173`) is a plain `Button`, so the expanded/collapsed state that `DisclosureGroup` announced is no longer exposed — worth an `.accessibilityValue` or equivalent on the button. I could not verify the VoiceOver output, only the source shape.

RecoilGaming:
@PathGao I addressed all seven of the issues in [d73ba78](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/d73ba78f5aa5ba275226821d2c9d8119c6a6499a) [c27e046](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/c27e0464cf63e479ea77dcbb8d988af042146cec) [4dc0908](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/4dc09089a1e615f10f38204c2f5080d3543646f9) [cc8c8a1](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/cc8c8a10c3163f7d2a39233e89a10c54ffd548dc) [40c8eaf](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/40c8eaf29809bc7060f65f185c88546f99e7e727) [7f0e7f9](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/7f0e7f9d9d962734cadfd863824058b56f23645e) [25c2dde](https://github.com/vorssaintapp/vorssaint-utils/pull/934/commits/25c2dde6ba872fd1e9123b0a0934b4229321d102). The app builds successfully with no warnings and passes self-test and unit checks.

PathGao:
Re-checked all seven against the follow-up commits: the Command Bar and menu panel hints, `captureRoles` ordering plus its two new assertions, the pt-BR/es/fr header split, README, the seeded expansion and the shared `DisclosureHeaderRow` all land, and the 13 locales for the two new strings are complete. Two things the fixes introduced or left open.

**F8 — two `Spacer()`s in one header row, so the count badge floats.** `disclosureHeader` (`ShortcutsSettings.swift`) ends its label with `Spacer()` then the count capsule, and `DisclosureHeaderRow` (`SharedUI.swift`) adds its own `Spacer(minLength: 0)` before the chevron. The label is a `TupleView` flattened into that `HStack`, so the row lays out as `[title, Spacer, count, Spacer, chevron]` — two fully flexible spacers split the free width evenly, putting the count roughly midway instead of flush against the chevron where `DisclosureGroup` had it. One spacer per row fixes it: drop the row's and give the only other caller its own.

```swift
// SharedUI.swift — DisclosureHeaderRow
HStack(spacing: 8) {
    label()
    Image(systemName: "chevron.right")

// MonitorPanelConfig.swift
DisclosureHeaderRow(isExpanded: expansionBinding(for: id)) {
    Text(title)
    Spacer()
}
```

**F9 — a recording-only install loses its capture shortcut with nothing taking over.** Before, `.screenshot` was available whenever any of the four features was, so a stored ⌃⌥⌘4 registered and opened the chooser. Now `availabilityFeatures` is `[feature]` and `syncToolShortcuts` gates each tool on `availableTools.contains(tool)`, so with the screenshot tool switched off in the hub the saved `screenshotShortcut` registers nothing, while `recorderShortcutEnabled` and friends default to off. The changed test states it directly: `!availableRoles(isAvailable: recordingOnly.contains).contains(.screenshot)`. The body's "an existing combination keeps working unchanged and no migration is needed" holds only for installs that kept the screenshot tool; for the rest the key goes dead silently. A one-shot migration covers it — if `screenshotShortcutEnabled` is true and `.screenshot` is unavailable, copy the combination and the enabled flag onto the first available tool's keys, behind a defaults flag so it runs once.

RecoilGaming:
@PathGao The issues you outlined should be fixed now, and I also added a couple of unit tests to test the shortcut migration.

PathGao:
Re-checked both. F8 is fixed: `DisclosureHeaderRow` now contributes only the chevron and each caller supplies the row's one `Spacer()`, so the count capsule sits flush. F9's migration and its three cases land. One install it still misses.

**F10 — the migration skips the combination that was never customized.** `migrateOrphanedCaptureShortcut` guards on `let shortcut = defaults.string(forKey: DefaultsKey.screenshotShortcut)`, but the registrar this PR deletes read `GlobalShortcut.saved(for: DefaultsKey.screenshotShortcut, fallback: .screenshotDefault)` — the fallback exists because the key is absent until someone edits the combination. So a recorder-only install that switched the capture shortcut on and kept ⌃⌥⌘4 has `screenshotShortcutEnabled == true` and no stored string: it registered before, registers nothing after this PR, and the migration returns at that guard while the `defer` still stamps `orphanedCaptureShortcutMigrated`, so it never runs again. That is the install the body's "the key keeps opening the chooser" is about.

```swift
let shortcut = defaults.string(forKey: DefaultsKey.screenshotShortcut)
    ?? GlobalShortcut.screenshotDefault.storageValue
```

Move it out of the `guard` and the rest stands. All three new test cases write `screenshotShortcut` first, so nothing exercises the absent-key path — that is the fourth case worth adding. I cannot see `ShortcutPreferenceRow`, so if enabling the toggle already writes the key on first appearance this is moot; the fallback costs nothing either way.

**F11 — the migration decides availability twice, by a different rule than the runtime.** It reads `defaults.bool(forKey: AppFeature.screenshot.availabilityKey)` directly, while `ScreenCaptureService.syncWithPreferences` goes through `ScreenCaptureTool.available()` / `isAvailable`. The two guards also use opposite polarity for the same unwritten key: screenshot must read `false` to be absent, candidates must read `true` to be present. On a state where the recorder key is written and the screenshot key is not, the migration moves a live shortcut off a tool that is actually installed. `AppFeature.screenshot.isAvailable` and `$0.feature.isAvailable` remove the second copy — unless migrations deliberately avoid `isAvailable` because it touches feature runtime before it is ready, which I cannot tell from the diff.

One thing to check outside the diff: `featuresWithShortcuts` now returns `false` for `.screenRecorder`, `.screenOCR` and `.colorPicker` in every group, but `visibleGroups` still counts a group as visible when any available role's feature is in it. If any of those three sits in a different `FeatureGroup` than `.screenshot`, that group now renders as an empty `Section`. Fine if all four share one group; I cannot see `AppFeature.group`.

**Impact:** a recorder-, OCR- or colour-picker-only install that never changed the capture combination loses ⌃⌥⌘4 with nothing taking over, and the one-shot flag makes it permanent.
**Merge:** needs the `?? GlobalShortcut.screenshotDefault.storageValue` fallback in `migrateOrphanedCaptureShortcut` first.

RecoilGaming:
@PathGao I added the ?? screenshotDefault fallback and a test for the absent key case. However, I looked into both of these issues and they aren't actually problems in production:
- F10: register() writes a default shortcut on startup
- F11: isAvailable and the other read do the exact same thing, and the raw reading only exists for unit tests to pass in defaults. This is consistent with other parts of the codebase.
- Also, the four capture tools are in the same group so an empty section won't happen.

PathGao:
Accepted on all three. The fallback and its test close the one case I couldn't decide from the diff, and the availability read and the shared group are yours to call — I could see neither `AppFeature.isAvailable` nor `AppFeature.group`.

One thing left in the migration, not a blocker. The target is picked by `!defaults.bool(forKey: $0.enabled)`, so a tool whose shortcut is switched off but still holds a saved combination counts as having no shortcut of its own and gets that combination overwritten. The body's wording is "the first installed capture tool without a shortcut of its own", which a disabled-but-configured tool arguably is not. Tightening the predicate keeps it and moves to the next candidate:

```swift
guard let target = candidates.first(where: {
    defaults.bool(forKey: $0.feature.availabilityKey)
        && !defaults.bool(forKey: $0.enabled)
        && defaults.string(forKey: $0.shortcut) == nil
}) else { return }
```

**Impact:** a capture tool whose shortcut was turned off but whose combination was kept loses that combination, on installs without the screenshot tool. Nothing else.
**Merge:** ship it.

Issue #697, opened 2026-08-18
Screen goes Dark at Click on Energy Settings


Csomm004:
### Description

1 Bug was the Screen Brightness goes Down

Second Bug is the Cleaning Function. Push 5 Times Key doesn't work.
Only the Unlock Button works.

### Steps to Reproduce

Go to Settings click wherever you want. Click on Energy and see all Screens Brightness goes Down 30 %. 
When I Close Vorssaint Screen goes back to normal

Go and Activate Cleaning Mode.
Push 5 Times Button is Disabled only the Big Button to Unlock is working

### Vorssaint Version

3.3.1

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

iltonandrew:
Status on the two symptoms here:

1. **Screens dim ~30% after clicking Energy** — root cause found and fixed in #814: a brightness level *read from* the monitor was being replayed as a *gamma dim* whenever the display routes rebuilt (the pane's refresh triggers a rebuild), and gamma tables restore when the process exits, matching "back to normal when I close Vorssaint".

2. **Cleaning Mode: press-5-times doesn't work, only Unlock** — investigated but not reproduced from the code: the key counting happens inside the same event-tap callback that swallows the keys, so there's no obvious routing flaw. If you can still reproduce on 3.3.2+, does pressing *any* key five times fail, or specifically Esc? And do normal Cleaning Mode key blocks otherwise work (keys do nothing)? That would tell whether the tap is running at all on your machine.

PathGao:
Tracking PR: #814 (open, `ready to merge`).

It covers the first half only — every screen dimming by ~30% when the Energy pane is opened, caused by a monitor's own DDC brightness readback being replayed as a gamma multiplier.

The second half reported here, cleaning mode not unlocking after five Esc presses, is **not** covered by #814 and has no PR. This issue stays open for that half even after #814 merges.

iltonandrew:
Root cause of the second symptom found and demonstrated with a runtime harness against the real `CleaningUnlockCounter`: the five Esc presses only count when each lands within 2 s of the previous one. Pressing slower than that restarts the count at 1 on every press, so the gesture can never complete — the overlay dots stay at one filled dot and only the Unlock button works. Synthetic HID Esc presses 2.5 s apart never unlock on current main; the same presses unlock with a wider window.

[...]

PR #935, opened 2026-08-24
fix(cleaning-mode): widen the Esc unlock window and let modifiers reset it

iltonandrew:
Candidate fix for the second symptom of #697 (Cleaning Mode not unlocking after five Esc presses). The first symptom was fixed by #814; PathGao noted this half had no PR yet. Candidate, not confirmed: the harness proves this failure mode exists, but the reporter's diagnostic question in #697 was never answered, so #697 should stay open until a reporter confirms on a build with this in.

## What was wrong

The five Esc presses only count when each lands within 2 s of the previous one (`CleaningUnlockCounter`, `pressWindow: 2.0`). A user pressing Esc deliberately but slower than once per two seconds restarts the count at 1 on every press — the overlay dots stay pinned at one filled dot and the gesture can never complete. Demonstrated with a runtime harness against the real `CleaningUnlockCounter` fed by a listen-only tap with the same mask: synthetic HID Esc presses 150 ms apart unlock (progress 1→5); the same presses 2.5 s apart never unlock on current main; they unlock with a wider window.

(Ruled out: session-level Esc remappers stealing the key — SuperKey's tap is `.cgSessionEventTap`, downstream of our `.cghidEventTap` head-insert, so it cannot intercept first. If a reporter uses Karabiner-style synthesized Esc that never reaches the HID tap, that is a separate cause this PR cannot address.)

## The change

Two commits, and the second is what makes the first safe.

**1. `pressWindow` 2.0 → 6.0.** A gesture a deliberate user cannot complete protects nothing, so the window is widened to a speed a person actually presses at. The doc comment now says exactly that. The shipped constant is pinned by a source-shape assertion beside the counter tests, since the test binary does not compile `CleaningModeManager.swift` and the counter cases construct their own windows.

**2. Modifiers now reach the counter.** The window's one narrow job is rejecting five isolated Esc-only contacts spread across a long wipe: any other key resets the count, and auto-repeat never counts. That was not actually true when this PR was opened. Shift, control, option, command, fn and caps lock arrive as `.flagsChanged` events, and while `.flagsChanged` was already in the tap's mask (`CleaningModeManager.swift:105`), the handler swallowed them without ever showing them to the counter — so the keys a cloth is most likely to land on while it is near Escape were precisely the ones that did *not* reset the count. At a 2 s window that gap was mostly academic; at 6 s it is the thing that would let a wipe accumulate five presses and unlock itself. They now take the same reset path as any other non-Escape key. No modifier key code is 53 (they run 54–63), so the new call can only reach the reset branch, and `isRepeat: false` is passed literally because CG does not define that field for `.flagsChanged`.

Merging only the first commit would ship the widened window without the guard that pays for it. That is why they are one PR.

## Verification

- Harness on this machine: 2.5 s-spaced presses fail to unlock at 2 s window, unlock at 6 s; fast presses unchanged.
- `./build.sh` — 0 warnings. `./build.sh --test` — 8804 checks, **1 failure**: `the dispatch pool starvation this check needs was actually reached`. That is the known machine-dependent check #1014 exists to skip: red on this 15-core M5 Pro, green on CI's smaller runners. It came up identically on two other unrelated branches gated in the same sitting, so it is environmental and not from this change — I am reporting it rather than writing `TESTS OK`. `./build/Vorssaint --selftest` — `SELFTEST OK`. The run includes the `pressWindow` source-shape pin (red if 2.0 returns) and the two anchored scans over the handler and the counter.
- Rebased onto current `main`.

**Still unverified on both sides:** the live HID gesture for the modifier path — nothing here synthesises a real `.flagsChanged` through a tap. The `fn` caveat stands rather than being resolved.

Refs #697

### DIFF
--- a/Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
+++ b/Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
@@ -45,9 +45,10 @@ final class CleaningModeManager: ObservableObject {
     /// pressing Escape slower than once per two seconds could never unlock
     /// (progress restarted at 1 on every press). The window's one remaining job
     /// is rejecting five isolated Esc-only contacts spread across a long wipe:
-    /// other keys reset the count and auto-repeat never counts, but a cloth can
-    /// strike Escape alone. Widening to 6s weakens that guard on purpose —
-    /// a gesture a deliberate user cannot complete protects nothing.
+    /// every other key resets the count — modifiers included, they reach the
+    /// counter as flags-changed events — and auto-repeat never counts, but a
+    /// cloth can strike Escape alone. Widening to 6s weakens that guard on
+    /// purpose — a gesture a deliberate user cannot complete protects nothing.
     private lazy var unlock = CleaningUnlockCounter(requiredKeyCode: Self.escapeKeyCode,
                                                     threshold: unlockThreshold,
                                                     pressWindow: 6.0)
@@ -158,6 +159,15 @@ final class CleaningModeManager: ObservableObject {
             let code = event.getIntegerValueField(.keyboardEventKeycode)
             let isRepeat = event.getIntegerValueField(.keyboardEventAutorepeat) != 0
             registerUnlockKeyDown(code: code, isRepeat: isRepeat)
+        } else if type == .flagsChanged {
+            // Shift, control, option, command, fn and caps lock arrive here rather
+            // than as key-downs, and they are the keys nearest Escape — a cloth
+            // resting on them must reset the count like any other key. Both the
+            // press and the release report the same key code and neither is
+            // Escape, so each one resets and the pair is idempotent. Modifiers
+            // never auto-repeat.
+            registerUnlockKeyDown(code: event.getIntegerValueField(.keyboardEventKeycode),
+                                  isRepeat: false)
         } else if type == Self.systemDefinedEventType,
                   let systemKey = systemKeyEvent(from: event),
                   systemKey.isKeyDown {
--- a/Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
+++ b/Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
@@ -41,9 +41,16 @@ final class CleaningModeManager: ObservableObject {
     private var screenObserver: NSObjectProtocol?
 
     /// The unlock-gesture state machine (pure, unit-tested separately).
+    /// The 6s press window forgives hesitant, deliberate presses — at 2s a user
+    /// pressing Escape slower than once per two seconds could never unlock
+    /// (progress restarted at 1 on every press). The window's one remaining job
+    /// is rejecting five isolated Esc-only contacts spread across a long wipe:
+    /// other keys reset the count and auto-repeat never counts, but a cloth can
+    /// strike Escape alone. Widening to 6s weakens that guard on purpose —
+    /// a gesture a deliberate user cannot complete protects nothing.
     private lazy var unlock = CleaningUnlockCounter(requiredKeyCode: Self.escapeKeyCode,
                                                     threshold: unlockThreshold,
-                                                    pressWindow: 2.0)
+                                                    pressWindow: 6.0)
 
     private init() {}
 
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -10674,6 +10674,20 @@ struct MetricsTests {
         let afterReset2 = cleared.registerKeyDown(code: escapeKeyCode, time: 0.4, isRepeat: false)
         expect(!afterReset2 && cleared.progress == 1, "after reset Escape starts fresh at 1")
 
+        // The counters above build their own windows, so nothing else here
+        // fails if the shipped constant regresses. Pin it at the source: the
+        // 2s window made the gesture impossible for anyone pressing Escape
+        // slower than once per two seconds (#697).
+        let cleaningSource = (try? String(
+            contentsOfFile: "Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift",
+            encoding: .utf8)) ?? ""
+        let cleaningCode = cleaningSource
+            .components(separatedBy: "\n")
+            .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
+            .joined(separator: "\n")
+        expect(!cleaningCode.isEmpty && cleaningCode.contains("pressWindow: 6.0"),
+               "the shipped unlock counter keeps the forgiving 6s press window")
+
         func systemKeyData(keyCode: Int, state: Int, repeatFlag: Bool = false) -> Int {
             Int((UInt32(keyCode) << 16) | (UInt32(state) << 8) | (repeatFlag ? 1 : 0))
         }
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -10674,6 +10674,24 @@ struct MetricsTests {
         let afterReset2 = cleared.registerKeyDown(code: escapeKeyCode, time: 0.4, isRepeat: false)
         expect(!afterReset2 && cleared.progress == 1, "after reset Escape starts fresh at 1")
 
+        // Modifiers are the keys nearest Escape, so a cloth reaches them first.
+        // They arrive as flags-changed events, which the tap feeds here too, and
+        // one physical press reports twice — once down, once up. Both are resets.
+        var smeared = CleaningUnlockCounter(requiredKeyCode: escapeKeyCode, threshold: 5, pressWindow: 6.0)
+        let leftShiftKeyCode: Int64 = 56
+        var smearedUnlock = false
+        for t in [0.0, 0.5, 1.0, 1.5] {
+            if smeared.registerKeyDown(code: escapeKeyCode, time: t, isRepeat: false) { smearedUnlock = true }
+        }
+        expect(smeared.progress == 4, "four Escapes stop one short of the threshold")
+        _ = smeared.registerKeyDown(code: leftShiftKeyCode, time: 2.0, isRepeat: false)
+        expect(smeared.progress == 0, "a modifier going down clears the Escape count")
+        _ = smeared.registerKeyDown(code: leftShiftKeyCode, time: 2.1, isRepeat: false)
+        expect(smeared.progress == 0, "the modifier's release resets again, idempotently")
+        if smeared.registerKeyDown(code: escapeKeyCode, time: 2.5, isRepeat: false) { smearedUnlock = true }
+        expect(!smearedUnlock && smeared.progress == 1,
+               "four Escapes with a modifier in between never unlock; the next Escape starts at 1")
+
         // The counters above build their own windows, so nothing else here
         // fails if the shipped constant regresses. Pin it at the source: the
         // 2s window made the gesture impossible for anyone pressing Escape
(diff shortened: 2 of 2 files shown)


PathGao:
Three things, in order of how much they change the diff.

**1. Nothing pins the value you changed.** `build.sh:323` compiles only `CleaningUnlockCounter.swift` into the test binary, so `CleaningModeManager.swift:51` is unreachable from `Tests/MetricsTests.swift` — the six counter cases at `Tests/MetricsTests.swift:9920-9964` all build their own counters with `pressWindow: 2.0` and stay green whatever the shipped constant is. The literal that was the entire bug is therefore unguarded, and a future edit can put 2.0 back with `TESTS OK`. The repo already has the mechanism for constants in non-test-compiled files: the source-shape assertion used at `Tests/MetricsTests.swift:11192` (and ~13 other places). Add one next to the cleaning block at `Tests/MetricsTests.swift:9915` that reads `Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift` and asserts the configured `pressWindow` is the wide one. #814 pinned its fix with two checks; this one pins none.

**2. The comment at `CleaningModeManager.swift:44-48` claims something that would make the change unnecessary.** If accidental unlocks "are still prevented by the other rules: any other key resets the count and auto-repeat never counts", then `pressWindow` has no job at any value and the honest change is to delete it, not to triple it. What the window actually still buys is narrower: it rejects five *isolated* Esc contacts — no other key in between — spread out over a long wipe, and at 6.0 a cloth working the top-left corner every few seconds now clears that bar where at 2.0 it had to be a fast scrub. That may well be an acceptable trade, but the comment should state it rather than assert the window guards nothing. Either reword 44-48 to name what 6s still rejects, or drop `pressWindow` from `CleaningUnlockCounter` entirely and let the two rules stand alone. I have no field data on how often a wipe produces Esc-only contact runs, so I can't tell you which side that lands on.

**3. Don't treat the second half of #697 as confirmed fixed on merge.** The diagnostic question in https://github.com/vorssaintapp/vorssaint-utils/issues/697#issuecomment-5385371387 — any key five times, or specifically Esc — was never answered, and the report is that the press-5 gesture did nothing at all while the Unlock button worked. A 2 s window does not produce that for someone actively trying to unlock: a user mashing Esc to get out is well inside 2 s per press, and would see the dots climb and reset rather than sit dead. Your harness proves the slow-presser failure mode exists; it does not tie it to this reporter. The PR body says "Fixes the second symptom of #697" — I'd soften that to a candidate fix and leave the issue open pending the reporter confirming on a build with this in.

Two things I checked that are fine and need no change: `CleaningUnlockCounter` is the only "N presses within T" implementation in the repo (no second copy to keep in sync), and the wider window costs no latency anywhere — there is no timer, and `registerKeyDown` runs only from the tap in `CleaningModeManager.swift:171`, which exists only while the mode is active.

iltonandrew:
All three addressed, in your order:

1. Pinned. Source-shape assertion beside the counter tests (same mechanism as the `BrightnessService` work-queue check): reads `CleaningModeManager.swift`, strips comment lines, asserts `pressWindow: 6.0` is what ships. Flipping the constant back to 2.0 now fails `--test`.

2. Reworded rather than removed. You're right the old comment proved too much — if the other two rules fully guarded wiping, the window would have no job at any value. The comment now names its actual remaining job (rejecting five isolated Esc-only contacts across a long wipe, which other-key reset and repeat-ignore can't catch) and states that 6 s weakens that guard on purpose. I kept `pressWindow` rather than deleting it: same reasoning as yours in reverse — no field data on Esc-only contact runs, so retiring the guard entirely is the bigger bet, and the widened window is the narrower diff.

3. Softened. Body now says candidate fix, states the harness proves the failure mode exists but doesn't tie it to this reporter (whose mash-to-exit pattern would sit inside 2 s and show climbing dots, as you note), and that #697 should stay open pending reporter confirmation on a build with this in.

Also rebased onto current main while in there. Gate: zero warnings, `TESTS OK (8479 checks)`, `SELFTEST OK`.

PathGao:
All three landed — I read the pin (same shape as the `BrightnessService` check at `Tests/MetricsTests.swift:11215-11227`, and it fails closed since an unreadable file leaves the string empty), the reworded comment, and the softened body. The rewording surfaced one thing rather than settling it.

**Modifiers don't reset the count.** `CleaningModeManager.swift:112` puts `.flagsChanged` in the tap mask, but `handle` feeds only `.keyDown` and system-defined keys to the counter (`CleaningModeManager.swift:157-165`). Shift, control, option, command, fn and caps lock arrive as `.flagsChanged`, get swallowed, and leave `progress` untouched. So "other keys reset the count" at `CleaningModeManager.swift:48` is false for the keys physically nearest Escape, and the run the window has to reject is not "five isolated Esc-only contacts" — it is five Esc contacts separated by nothing but modifier hits. A cloth working the top-left corner produces those. At 2 s it had to be a fast scrub; at 6 s a slow pass over that corner clears the bar, which is the accidental-unlock side, mid-wipe.

That hole is pre-existing, but this PR is what makes it reachable and the new comment is what asserts it away, so I'd settle it here. The narrow version is three lines in `handle`: route `.flagsChanged` through `registerUnlockKeyDown` with the event's keycode — no modifier keycode is 53, so it can only reset, never advance — plus one counter case. It does add a new way to lose the streak: a stuck or synthesized modifier would reset every Esc press and strand the gesture, though the Unlock button stays live. I can't test that side from here, so it's your call which way it lands.

To clear `incomplete`: either that reset goes in with a counter case, or `CleaningModeManager.swift:44-50` says plainly that modifiers do not reset and that you accept it at 6 s — the comment can't keep claiming a guard the code doesn't have. This also needs another rebase; main moved to `a8186833` after yours, so GitHub reads the branch as diverged (no conflicts).

PathGao:
#1006 and #1007 weren't in what I was handed — no body, no diff — so I haven't reviewed them and won't guess. Link them here and I'll take them. If they replace this PR, say so and I'll revisit the label on this one.

Nothing on #935 has moved since my last comment. `.flagsChanged` is still swallowed at `CleaningModeManager.swift:157-165` before it can reach the counter, so shift/control/option/command/fn/caps lock — the keys physically nearest Escape — do not reset the count, and the comment at `CleaningModeManager.swift:44-50` still asserts a guard the code doesn't have. Either route modifiers through `registerUnlockKeyDown` with a counter case, or state in the comment that they don't reset and that you accept it at 6 s. The branch also still needs the rebase past `a8186833`.

I re-read the diff for anything I'd missed the first time and found nothing new worth a line. The source-shape pin fails closed and matches the `BrightnessService` shape; the reworded comment and softened body are what you described.

**Impact:** none new since the last review.
**Merge:** needs the modifier reset (or the comment corrected to match) plus a rebase first.

ref #697

iltonandrew:
On #1006/#1007: they're yours, and neither replaces this. #1006 is `fix(update): run the installer in its own session` and #1007 is `fix(mixer): unblock teardown and silence unwritten output frames` — update installer and audio mixer, nothing touching `CleaningModeManager`. Whatever put them on this thread got them here by accident; the label on this one stands on its own merits.

On the modifiers: taking the fix, not the comment. Routing them through the reset is the correct call rather than the cheaper one, and the reasoning is the feature's whole point — Cleaning Mode exists so a cloth on the keyboard reaches nothing. Shift, control, option, command, fn and caps lock are the keys a hand or cloth lands on *while* it is near Escape, so leaving them unable to reset the count is exactly the path to a wipe accumulating five Escapes inside one window and unlocking itself. Documenting that at 6 s would be writing down the false-unlock rather than closing it.

Working on that plus the rebase past `a8186833` now; the comment at `:44-50` will end up describing what the code actually does either way.

iltonandrew:
Both in 239d1d7, rebased onto current `main` (`51c827d`) — clean, no conflicts; `a8186833` was already an ancestor, the branch was just 62 behind. Force-pushed with lease since the history moved.

**Modifiers reach the counter.** A `.flagsChanged` branch in `handle` calls the existing `registerUnlockKeyDown` with the event's key code. No new counter case: `CleaningUnlockCounter` already resets on any code that isn't 53, and no modifier is 53, so they fall into the reset path that exists. A second way to spell one rule would have been the worse version of this.

Three details worth stating rather than leaving to the diff:

- **Press and release both reset, and that's fine.** One shift tap emits two flags events with the same key code; each writes `progress = 0` and the sentinel `lastKeyTime`, both constants, so the second is a no-op on the first.
- **It can't break the real gesture.** Escape never emits `.flagsChanged`, so a clean five-Escape run sees no flags events at all.
- **`isRepeat: false` is passed literally, not read from the event.** CoreGraphics doesn't define that field for `.flagsChanged`, and a stray non-zero would make `!isRepeat` skip the reset — which is precisely the hole being closed. Modifiers don't auto-repeat, so the constant is the honest value.

**Nothing changed about what gets swallowed.** The tap mask is byte-identical (`.flagsChanged` was always in it — that's why these events were already blocked and the finding was about the counter, not the lock), `handle`'s only two exits are untouched, and the new branch adds no `return`. That last one is now mechanised: a scan fails on any `return` in `handle` that isn't `return nil`, and planting `return Unmanaged.passUnretained(event)` in the new branch makes it fail naming the line.

`CleaningUnlockCounter` is in the test binary, so the behaviour is unit-asserted — four Escapes, a shift down clears, the shift up clears again, the next Escape starts fresh at 1 and never unlocks. Both that and the routing are mutation-proved.

The comment at `:43-51` now names the mechanism instead of asserting the guard: "every other key resets the count — modifiers included, they reach the counter as flags-changed events".

Not verified: the live HID gesture, since nothing here synthesises a real `.flagsChanged` through a tap. Also unverified is whether `fn` reaches the tap on your keyboard — it's firmware-handled on some third-party boards, in which case it neither resets nor leaks, same as before. Worth noting the failure direction either way: a spurious flags event makes the unlock *harder* to complete, never accidental.

PathGao:
239d1d7 does what I named, and the routing is the right half to take. I read the branch, the counter case and both scans.

Checked and fine: `.flagsChanged` was already in the mask at `CleaningModeManager.swift:105` on main, so the branch adds a read, not a new interception. No modifier key code is 53 (they run 54–63), so the new call can only reach the reset path in `CleaningUnlockCounter.swift:31-35` — no second spelling of the rule, agreed. The counter case is arithmetically right: four Escapes inside 6 s reach `progress == 4`, the shift down/up pair clears twice idempotently since both write the `lastKeyTime` sentinel, and the next Escape lands on the `time - lastKeyTime > pressWindow` branch at 1. The handler scan is scoped correctly — `handle` ends at the next `    private func ` (`systemKeyEvent`) — and with `handlerStart` nil the range collapses to empty and `modifiersReachCounter` stays false, so it fails closed the same way the `pressWindow` pin does.

One thing left, and it's the body, not the code.

**The body still describes a one-line PR.** "`pressWindow` 2.0 → 6.0" is the whole of "The change" — nothing there says this PR also changes which events feed the counter. And the parenthetical right after it, "other keys reset the count and auto-repeat never counts, but a cloth can strike Esc alone", is the pre-correction framing: it was false when written, it is true now only because of the second commit, and the body reads as though it was always true and nothing had to be done. That is the same defect the comment at `:43-51` just had fixed. Verification also still reports `TESTS OK (8479 checks)`, from before the counter case and the two scans.

It matters because this squashes. `git log` would say the commit widened a timeout, and the modifier routing — the part that actually closes the false unlock — would live nowhere but this thread.

Nothing more in the code: put the routing into "The change", rewrite or drop the "isolated Esc-only contacts" parenthetical the way `:43-51` was rewritten, refresh the check count. The title only covers the window too, but that one's yours.

Still unverified on both sides: the live HID gesture, since nothing here synthesises a real `.flagsChanged` through a tap. Your `fn` caveat is the right one to leave standing rather than resolve. Noted on #1006/#1007 — unrelated, nothing to revisit on this label.

**Impact:** none in code. The merge record would understate what shipped.
**Merge:** needs the PR body updated to cover the modifier routing first.

ref #697

iltonandrew:
All three body points done, and I took the title too.

**"The change" now covers both commits.** It reads as two numbered halves — the window widening, then the modifier routing — with the routing described as what makes the widening safe rather than as a footnote. The framing is explicit that merging only the first commit would ship the weakened guard without the thing that pays for it.

**The parenthetical is rewritten, not patched.** It used to assert, in passing, that any other key resets the count and auto-repeat never counts. You are right that this was false when written: `.flagsChanged` was in the mask at `CleaningModeManager.swift:105` but the handler swallowed those events without showing them to the counter, so modifiers — the keys a cloth actually lands on near Escape — were exactly the ones that did not reset. It now states that as the defect the second commit fixes, and says why it was mostly academic at a 2 s window and load-bearing at 6 s. Same treatment the comment at `:43-51` got.

**Check count refreshed**, and reported as it actually ran rather than as `TESTS OK`: 8804 checks, 0 warnings, `SELFTEST OK`, and **one failure** — the `dispatch pool starvation` precondition that #1014 exists to skip. Red on this 15-core M5 Pro, green on CI's smaller runners, and it appeared identically on two unrelated branches I gated in the same sitting, so it is environmental rather than from this branch.

**Title.** You said that one was mine, and your squash-record argument applies to it as much as to the body, so it is now `fix(cleaning-mode): widen the Esc unlock window and let modifiers reset it` — both halves, still `type(scope): lowercase imperative phrase`.

Also rebased onto current `main` (was 16 behind, clean), so the numbers above are the gate on what would merge rather than on an older base.

The `fn` caveat and the absent live `.flagsChanged` HID run both still stand, unchanged, in the description.

ref #697

PathGao:
Body and title both cover the routing now. That was the last thing.

Re-read what changed since 239d1d7: "The change" is two numbered halves with the routing framed as what pays for the widening; the old parenthetical is rewritten rather than patched — it now states the swallowed `.flagsChanged` events as the defect the second commit closes, instead of asserting the guard that gap removed; and the count is reported as it ran, 8804 with one named failure. Reporting the red instead of writing `TESTS OK` is the right call, and CI's Build & selftest is green on the smaller runner, which is the gate that matters here.

Code is unchanged from 239d1d7, which I read last round. I did re-check the two scan anchors against current main, since the rebase moved 16 commits under them: `handle` still ends at `systemKeyEvent` (`CleaningModeManager.swift:140-164`), both its exits are still `return nil` (`:145`, `:162`), and `.flagsChanged` is still in the mask at `:105`. The scans keep their footing on this base.

Still unverified on both sides, unchanged: the live HID gesture, and whether `fn` reaches the tap on third-party keyboards. Both stand as caveats in the description rather than resolved, which is the honest shape — a spurious flags event only makes the unlock harder to complete, never accidental.

Removing `incomplete`.

**Impact:** none. The merge record now names both halves.
**Merge:** ship it.

ref #697

Issue #933, opened 2026-08-24
Keyboard shortcuts to create and close Scratchpad tabs


Steven-Hsu1:
### Problem or Motivation

When using the Scratchpad utility, I cannot create a new tab or close the current one without the mouse (the + button, or the three dots --> "Close scratchpad"). This creates friction when I want several notes quickly.

### Proposed Solution

While the scratchpad is focused, Command-T should create a new tab and Command-W should close the current one, matching the usual macOS / browser tab shortcuts. Customizing those shortcuts can wait.

Command-W on the last tab should close the scratchpad window. Closing a tab that still has text should keep the existing confirmation.

### Feature Area

Tools

Borisserz:
Proposed in https://github.com/vorssaint/vorssaint-utils/pull/1426

While Scratchpad is focused: ⌘T new tab, ⌘W close tab (last tab hides the pad); confirmation still goes through the existing SwiftUI dialog.

PR #942, opened 2026-08-25
feat(scratchpad): add tab keyboard shortcuts

Steven-Hsu1:
## Summary

This change assigns Command-T to the existing new-tab button. Command-W uses the same requestClose path as the actions menu. It closes the selected tab, preserves the confirmation for notes containing text, and hides the Scratchpad window when only one tab remains.

## Testing

- [x] `./build.sh --test` passes,  8478 checks
- [x] `./build.sh` finishes without warnings
- [x] `./build/Vorssaint --selftest` reports `SELFTEST OK`
- [x] Manually verified command behavior in Developer build

Refs #933


PathGao:
**⌘W is claimed by the app's own main menu, and the overlay button does not change that**

`AppDelegate.swift:1157-1158` installs a Window-menu item with `action: #selector(NSWindow.performClose(_:)), keyEquivalent: "w"` into `NSApp.mainMenu`. `NSApplication.sendEvent` offers a command key-down to the main menu before the key window's responder chain, so a `.keyboardShortcut` registered anywhere inside the hosting view — menu item or overlay — is consulted second.

The repo already records this for a panel configured exactly like the scratchpad's (`CommandBarService.swift:2316` `[.borderless, .nonactivatingPanel]`, `canBecomeKey: true`). `CommandBarService.swift:2448-2455`:

> The app's menu owns these combinations and the panel is key, so they would quit, close or hide Vorssaint while the person believes they are acting on the app the bar is floating over.

and it swallows ⌘Q/⌘W/⌘M/⌘H in the panel's local key monitor. `ScreenshotPinController.swift:161-171` handles `kVK_ANSI_W where flags.contains(.command)` in `keyDown` for the same reason.

So the comment at `ScratchpadView.swift:170-172` names the wrong mechanism. A `Button` inside a closed SwiftUI `Menu` never registers its key equivalent because it is not in the hierarchy — that is why line 112 does nothing. Moving the button into an overlay fixes the hierarchy problem but not the main-menu ordering problem, which is the one your comment describes.

I could not run the app, so I did not observe which of the two actually wins: `performClose:` may validate to disabled on a borderless panel with no `.closable` in its style mask, in which case the event would fall through to your button. That is an undocumented AppKit detail to depend on for a keystroke this common.

What to change: delete `closeShortcut` (`ScratchpadView.swift:170-183`), delete the now-dead `.keyboardShortcut` at `ScratchpadView.swift:112`, and handle both keys in the key monitor that already exists at `ScratchpadService.swift:428-439`, beside the Escape case. `addLocalMonitorForEvents` runs before `sendEvent`, so it beats the main menu deterministically, it is already scoped with `event.window === panel`, and returning `nil` stops the beep. That is also where the `⌘T` case belongs, so both keys live in one place instead of split across a view and a service.

**The modal guard is missing**

`ScratchpadService.toggle()` at `ScratchpadService.swift:77` starts with `guard !modalInteractionActive`. Your ⌘W path has no equivalent, so a second press while the rename or close alert is up re-enters `requestClose` and can set `dialog` again underneath the presented alert. Handling the keys in the service monitor lets you reuse the same guard directly.

**Two `Close scratchpad` items now disagree on the last pad**

You removed `.disabled(!service.canClosePad)` from the actions-menu item (`ScratchpadView.swift:111-112`) but the tab context menu still carries it at `ScratchpadView.swift:160`. On a single pad the ellipsis item is enabled and hides the window, while right-click → Close scratchpad is greyed out. Same label, same `requestClose`, opposite availability. Remove the guard at line 160 too, or keep the actions-menu item disabled and let only the keystroke reach the hide branch.

**⌘T and ⌘W are neighbouring keys and the close is not undoable**

`docs/AI-CONTRIBUTIONS.md:137-139`: a destructive action and its constructive twin should not be physically adjacent, not neighbouring keys. `ScratchpadSupport.requiresCloseConfirmation` (`ScratchpadSupport.swift:307-309`) is `!pad.text.isEmpty`, and `ScratchpadService.closePad` (`ScratchpadService.swift:231-236`) writes the document straight through with no undo. So ⌘W one key over from ⌘T destroys a named-but-empty pad silently and permanently.

Browsers pair ⌘W with ⌘⇧T for exactly this. `docs/AI-CONTRIBUTIONS.md:127-131` asks for the mainstream comparison to be in the pull request as an argument — put it in the body, and say either why reopen is not needed here or add it.

**No test and no discoverability**

`Tests/MetricsTests.swift:13246-13257` already reads `ScratchpadView.swift` as text and asserts its shape, so a source-shape assertion for the new binding needs no change to the `build.sh:199,215` allowlist. Add one there.

Neither `.help()` names the keys. `ScreenshotEditorView.swift:735` appends the glyph to the existing localized string (`strings.pinButton + "  (⌘P)"`), which needs no new locale entries — the same works for `text.newPad` at `ScratchpadView.swift:103-105` and `text.closePad` at line 111.

Steven-Hsu1:
Thanks for the detailed review. I’ve updated the implementation to address the feedback:

- Removed the hidden shortcut overlay and SwiftUI `.keyboardShortcut` modifiers.
- Command-T and Command-W are now handled together in the Scratchpad panel’s local key monitor, before the main menu.
- Exact shortcuts are consumed during rename, close, and export modals to prevent re-entry.
- Command-W uses the same close flow as the menu actions.
- Both menus now keep Close available for the last pad, where it hides the window instead of deleting the pad.
- Pads with text or a user-edited name require confirmation. Name provenance is persisted explicitly, including conservative handling of older documents.
- Added ⌘T/⌘W to the existing localized help text.
- Added behavior and source-shape tests covering the requested cases.

A reopen command is unnecessary because user-authored content cannot be deleted without confirmation, and the last pad is never deleted.

Verification after rebasing:

- 8,716 tests passed
- Release build completed without warnings
- `SELFTEST OK`
- Packaged UI behavior manually verified

Changes are in commit `659730c`.

PathGao:
**The modal `.consume` branch is probably unreachable**

`ScratchpadService.swift:429` opens the monitor with `guard let self, let panel, event.window === panel else { return event }`. While the rename or close alert is up, that alert is a sheet — its own `NSWindow`, and it is the key window — so the event's window is not the panel and the monitor returns before `panelKeyDecision` is ever reached. Same for the export `NSSavePanel`. The only window `.consume` can fire in is between `setModalInteractionActive(true)` and the sheet appearing, i.e. a second press inside the same run loop turn.

That is not automatically a bug: ⌘W falls through to `NSApp.mainMenu`'s `performClose:`, which on a sheet with no close button should just beep, and re-entry into `requestClose` is prevented by the window guard rather than by your check. But `MetricsTests.swift` now asserts `.consume` as if it were a live path. Press ⌘W with the close confirmation showing and see what happens. If `performClose:` dismisses the sheet without SwiftUI's binding noticing, `dialog` stays set and `modalInteractionActive` stays `true`, and `toggle()` (`ScratchpadService.swift:77`) is guarded on it — the hotkey stops toggling until Escape on the panel clears it. If it only beeps, delete the branch rather than keep a tested dead decision.

I could not run the app, so I did not observe which.

**⌘W is advertised nowhere the user will see it**

The `.help` at `ScratchpadView.swift:112` is on a `Button` inside `Menu` content. Every other `.help` in the repo sits on a standalone control; there is no precedent for one inside a menu's items, and `OnboardingView.swift:392` already records that `.help` has placement conditions ("never fires on a disabled control"). macOS menu items advertise a shortcut through their key equivalent, and this one has none by design. What a hover on the ellipsis actually shows is its own `.help(text.padActions)` at line 123.

⌘T is fine — lines 101-105 are a real button, same shape as `ScreenshotEditorView.swift:735`. Hover the menu item on a build and check. If nothing appears, the glyph belongs in the item's title, and the `count == 2` assertion in `MetricsTests.swift` is currently locking in the placement that does not render.

**⌘T does nothing and says nothing at 12 pads**

`performPanelShortcut` calls `createPad` unconditionally. `addingPad` (`ScratchpadSupport.swift:111`) returns `nil` at `maximumPadCount`, `createPad` returns, and the monitor already returned `nil` so there is not even a beep. The button path is `.disabled(!service.canCreatePad)` with `padLimitFormat` saying why. Gate the decision on the same predicate, or return `.passThrough` when it cannot act so the press is at least audible.

**Legacy pads confirm forever, and that changes the menu too**

`nameWasEdited` decodes to `true` for older documents and nothing ever sets it back: `renaming` only sets it true, `sanitized` carries it through. So for anyone who used the scratchpad before this build, `requiresCloseConfirmation` is permanently true on every pad — including through the pre-existing menu item, which used to close an empty pad without asking. The body says "conservative handling of older documents"; say that it is permanent and that it reaches the menu path, not just ⌘W.

The default itself is right. The alternative is recognising the generated name pattern, which means judging by localized text.

**Impact:** ⌘T at the pad limit is a silent no-op. ⌘W has no visible affordance if `.help` does not render inside menu content. Existing installs get a confirmation on every pad close, including from the menu item that used to close silently.
**Merge:** needs the `canCreatePad` guard on the ⌘T path first, plus your answer on the two things I could not check on hardware.

ref #933

PathGao:
Closing as stalled. The last review (2026-08-26 20:28) asked for one thing before merge: the `canCreatePad` guard on the ⌘T path. It is still missing — `case .newPad: createPad(defaultName: ...)` runs unconditionally while the "+" button beside it is `.disabled(!service.canCreatePad)`, so at the pad limit ⌘T is a silent no-op. No push or reply in seven days, and the branch is now conflicting.

The idea is fine (#933) and the rest of the diff was clean. If you come back to it, add the guard, rebase, and reopen; the review is otherwise done.

PR #943, opened 2026-08-25
fix(homebrew): skip disabled packages when upgrading everything

RecoilGaming:
**Before.** When any outdated package has been disabled by Homebrew, `brew upgrade` refuses the entire run (`Error: six has been disabled because it does not meet homebrew/core's requirements for Python library formulae!`), so Upgrade all in the Homebrew panel fails outright and none of the other packages get their update.

**After.** The app recognizes the refusal, notes `Skipping six: disabled in Homebrew.` in the operation log, and retries the upgrade with the remaining packages, so everything that can update still does. The mechanics mirror the existing untrusted-tap handling: `disabledPackageName(fromOutput:)` extracts the token-validated name, and `upgradeExcluding(_:from:outdated:)` rebuilds the failed command without it as the commands to run in order — an explicit token list just loses the name (flags preserved), while a bare `upgrade` splits by kind: an explicit `--formula` upgrade built from the known outdated formulae minus disabled and pinned names, then a bare `upgrade --cask`. Casks are deliberately never named, because `brew upgrade <name>` resolves a name shared by a formula and a cask as the formula — leaving the cask half bare keeps its coverage brew's own decision instead of a snapshot's. The log notes the skip and echoes the exact retry command each leg runs, so a narrowed run is visible. The rebuild returns nil whenever a retry would repeat the failed command or nothing is left: an explicit retry strictly shrinks and a failing bare `--cask` leg is never expanded again, so the chain always terminates; anything unrecognized falls through to the existing failure path unchanged. Upgrading a single disabled package directly still shows Homebrew's own error, which already says what to do.

Seven unit checks cover the parser and the rebuild rules (per-kind split, cask half kept when no formula remains, flags kept, pinned excluded, all three nil cases).

Verified on macOS 26.5.2 (25F84), Apple Silicon, self-built dev bundle, app in English: `./build.sh` finishes without warnings, `--selftest` passes, and `./build.sh --test` passes 8506 checks. The fix was exercised against the real condition — an installed `six` 1.16.0_3 with the formula disabled upstream: before the change Upgrade all failed with the error above, after it the run skips `six` and upgrades the rest. Not verified on a real machine: several disabled packages in one run (the retry loop resolves them one per pass; unit-tested only) and a disabled cask rather than formula.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project are documented here. The format follows
 - Monitor alerts can now warn you when the battery stays above a temperature you choose. Under Monitor alerts, off by default. Thanks to @ywu73.
 
 ### Fixed
+- Upgrading Homebrew packages no longer fails outright when one of the outdated packages has been disabled by Homebrew: the disabled package is skipped, with a note in the log, and everything else still updates.
 - The screen capture loupe now rings the pixel under the pointer instead of drawing a crosshair across it, and that pixel sits in the middle of the loupe, so the Color picker shows the color it is about to copy. Thanks to @I-Have-No-Idea-What-Im-Doing-Right-Now and @PathGao.
 
 ## [3.3.3-beta.2] - 2026-08-22
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
@@ -42,6 +42,8 @@ final class HomebrewManager: ObservableObject {
     private var popularityLoads: Set<HomebrewPackageKind> = []
     private var activeProcess: Process?
     private var cancelRequested = false
+    private var pendingUpgradeRetries: [HomebrewCommand] = []
+    private var upgradeRetryNote = ""
     private var installedCaskRecords: [HomebrewCaskRecord] = []
     private var installedCaskRecordsFetchedAt: Date?
     private var ownershipLoads: [String: [(HomebrewPackage?) -> Void]] = [:]
@@ -362,6 +364,8 @@ final class HomebrewManager: ObservableObject {
         terminalFallbackCommand = nil
         errorMessage = nil
         clearUntrustedTap()
+        pendingUpgradeRetries = []
+        upgradeRetryNote = ""
         cancelRequested = false
         completedOperationCleanup?.cancel()
         completedOperationCleanup = nil
@@ -376,6 +380,11 @@ final class HomebrewManager: ObservableObject {
                 self.activeProcess = nil
                 self.operation = nil
                 if status == 0 {
+                    if action == .upgradeAll, !self.pendingUpgradeRetries.isEmpty {
+                        self.continueUpgradeRetry(self.pendingUpgradeRetries,
+                                                  note: self.upgradeRetryNote)
+                        return
+                    }
                     self.markOperationComplete(result: .succeeded,
                                                phase: .refreshing,
                                                activity: nil)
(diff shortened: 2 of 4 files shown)


PathGao:
`upgradeExcluding` rebuilds the bare-`upgrade` set out of names, and names don't carry kind. Body: "a bare `upgrade` expands to the known outdated list minus disabled and pinned packages" — that isn't the same command. `brew upgrade` with no args resolves formulae and casks separately; `brew upgrade six jq docker` resolves each token, formula-first. An outdated cask whose name also exists as a formula (`docker`, `julia`, `r`) gets resolved differently after the retry than it would have in the original run. I did not verify brew's exact fallback when the formula exists but isn't installed and the cask is — but the retry hands brew strictly less information than the command it replaces. Same shape for anything brew would have upgraded that isn't in `outdatedPackagesByID`: a stale snapshot silently narrows the run, and the log only names the disabled package.

Root cause is that this is a second copy of "what upgrade-all covers". The first lives wherever the `.upgradeAll` command is built (not in the diff). Kind handling, pinned handling, and any future greedy/cask flags now have to stay in sync in two places. Lazier fix than a second expansion: give the existing upgrade-all builder an `excluding:` parameter and call it, so the set is defined once and the retry inherits kind scoping for free.

Second thing: `perform(action:package:command:)` is called before `appendLog("Skipping …")`. `perform` isn't in the diff, so I didn't verify whether it resets the operation log — if it does so asynchronously, the note gets wiped and the body's "notes `Skipping six:` in the operation log" doesn't hold. Worth confirming; if the reset is synchronous inside `perform`, the current order is the correct one.

Cask refusals don't match the regex charset (`Error: Cask '…' has been disabled`), so they fall through to the existing failure path. That's the safe direction and matches what you listed as unverified — no change needed there. The termination argument holds: every retry either shrinks the token list or returns nil.

No open issue tracks this one, so nothing to cross-reference.

**Impact:** Only on the retry path, and only after a disabled package triggers it — an outdated cask sharing a name with a formula gets resolved as a formula, and packages absent from the app's outdated snapshot are dropped from the run without a log line.
**Merge:** needs the bare-`upgrade` expansion to preserve formula/cask kind first.

PathGao:
Closing this, but your report was right about the symptom — we found the cause is somewhere else, and I want to lay out the evidence rather than just assert it.

I built an isolated Homebrew prefix with a local tap: formula `alpha` installed at 1.0, bumped to 2.0 and marked `disable!`; formula `beta` merely outdated. Bare `brew upgrade`, Homebrew 6.0.21:

```
==> Fetching downloads for: local/sandbox/beta and local/sandbox/alpha
Error: local/sandbox/alpha has been disabled because it is a sandbox experiment! It was disabled on 2025-10-16.
==> 1 Deprecated or disabled package
local/sandbox/alpha (disabled)
==> Upgrading local/sandbox/beta
  1.0 -> 2.0
==> Upgraded 1 outdated package
exit 1
```

`beta` upgraded. Only the disabled package was skipped. Same on 5.1.15.

The mechanism: the disabled check raises `CannotInstallFormulaError` in `FormulaInstaller#prelude_fetch`, and `Homebrew::Install.select_formula_installers` rescues it, calls `ofail` — which prints exactly that `Error: <name> has been disabled ...` line and sets a non-zero exit — then drops that one installer and carries on. The same `rescue CannotInstallFormulaError => e; ofail e; false` is in `upgrade.rb` at 3.6.0, 4.0.0, 4.3.0, 4.6.20, 5.1.15 and 6.0.21. Casks don't even error — they get `opoo "Not upgrading X, it is disabled"` and exit 0.

So that error line is not a refusal to run. It is the output of a run that finished and did the rest of the work.

What actually breaks is on our side. `HomebrewManager.swift:378` treats any non-zero exit as total failure: the panel reports failed, shows the tail of the output as the error, and skips `refreshInstalled()` — so the outdated list stays stale. The disabled package stays installed and stays outdated, so it repeats on every press, forever. That is a reporting bug, and it explains the part of your report that is real: it does look like nothing happened, and it does not go away.

Which is why I am not merging a command-rewrite. Filtering the disabled package out would mean turning bare `brew upgrade` into an explicit name list, which hands the app the job of deciding what "everything" means, brings back the formula/cask name-collision problem you hit, and drops brew's own dependent handling — all to suppress an exit code that is telling the truth.

I am raising the exit-status handling separately. Thanks for this — the report stands, and chasing it is what turned up the real one.

PathGao:
Raised as #1281 — the app now re-reads what is installed after a run that exited non-zero, so a partly done upgrade stops looking like a total failure with a stale list. Thanks again; that one came directly out of digging into this report.

PR #944, opened 2026-08-25
refactor(scratchpad): share the plain-text editor

tenbux:
## Summary

The scratchpad's `NSTextView` bridge is the only plain-text editor in the app that switches off every substitution, and it was `private` inside `ScratchpadView`. The snippet editor needs the same surface, so this lifts it into `UI/PlainTextEditor.swift` rather than letting a second copy appear.

No user-facing change. The scratchpad passes in the three things its copy hardcoded (label color, the 7x2 container inset, the text-view registration hook), and the placeholder overlay now reads its font size from the editor instead of repeating the literal, so the two cannot drift apart.

Split out of #766 so the change to a shipped screen can be reviewed on its own.

## Testing

```sh
./build.sh --test              # TESTS OK (8533 checks)
```

Checked on Apple Silicon against a release build. Not tried on macOS 14 or 15, which is the gap worth knowing: this is a straight move of existing code, but `NSTextView` behavior differs by version and the scratchpad is the screen that would show it. I also have not exercised the retention sweep or pad switching by hand since the move; those paths are unchanged, but they are where a regression would surface.


PathGao:
Straight move — old and new `makeNSView` set the same properties in the same order, and the scratchpad passes back exactly the three things its copy hardcoded. `No user-facing change` checks out.

Two notes, neither blocking:

`textColor` and `textContainerInset` are read only in `makeNSView`. Once the snippet editor from #766 lands, a caller that derives either from state will have its updates silently dropped. A line in the doc comment saying they apply at creation only is enough.

The `fontSize` comment claims the placeholder "cannot drift out of alignment", but font size is the smaller half of that. The overlay's position still comes from two unlinked literals: `NSSize(width: 7, height: 2)` at the call site and `.padding(.leading, 12)` / `.padding(.top, 2)` on the `Text` — 12 is presumably 7 plus the default 5pt `lineFragmentPadding`. Change the inset and the placeholder slides off the first line with nothing to catch it. Either derive the padding from the inset or narrow the comment to the font.

Not verified: the snippet-editor half isn't in this diff, so I can't check that this surface actually covers it. No runtime check on my end either — the macOS 14/15 `NSTextView` differences you flagged are exactly what a compile-clean move won't catch.

**Impact:** none.
**Merge:** ship it.

tenbux:
Both taken.

`textColor` and `textContainerInset` now say in the doc comment that they are applied when the view is made and not on update, so a caller deriving one from changing state knows it will not be re-applied.

You were right that the font was the smaller half. The overlay's position is no longer two unlinked literals: the inset is a single constant in `ScratchpadView` that both the editor and the placeholder read, and `PlainTextEditor` exposes `lineFragmentPadding` (NSTextContainer's default, which sits inside the inset) so the leading padding is derived rather than restated. The arithmetic is the same 7 + 5 and 2, so nothing moves.

Agreed the macOS 14 and 15 `NSTextView` differences are the gap a compile-clean move cannot close, and I have no runtime on either.

PathGao:
Padding math checks out: 7 + 5 = 12 and 2, unchanged.

One leftover from "derived rather than restated": `lineFragmentPadding = 5` mirrors NSTextContainer's default instead of setting it, so the constant is still a second copy of a number you don't own. One line in `makeNSView` makes it the source:

```swift
textView.textContainer?.lineFragmentPadding = Self.lineFragmentPadding
```

Non-blocking.

**Impact:** none.
**Merge:** ship it.

tenbux:
Taken. `makeNSView` now sets `textContainer?.lineFragmentPadding` from the constant instead of leaving it to whatever AppKit defaults to, so the number has one owner. The doc comment says that rather than claiming a default.

Nothing moves: the padding it now sets is the same value the container was already using, which the previous literal of 12 encoded as 7 + 5.

That is the second claim of mine you have had to trim, after the one about the placeholder not drifting. Both were broader than the code underneath them.

PR #709, opened 2026-08-20
Allow custom terminals for Homebrew commands

KSI-cell:
## Summary

- add a preferred terminal setting for Homebrew commands that require an interactive terminal
- detect and support Terminal.app, iTerm2, Ghostty, WezTerm, and Alacritty
- use the selected terminal for installer, shell setup, and sudo fallback commands
- keep Automation permission reporting aligned with the selected terminal
- verify CLI-launched terminals, handle stale launches and fallback errors, and refresh availability after cask changes
- localize the new settings and status messages in every supported language

## Validation

- `./build.sh --test` — 7,599 checks passed
- `./build.sh` — release bundle built successfully
- `./build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest` — `SELFTEST OK`
- field-tested Homebrew formula install/test/uninstall and cask install/uninstall
- field-tested Terminal.app, iTerm2, Ghostty, WezTerm, and Alacritty, including sudo fallback, launch verification, preferred-terminal fallback, and cleanup
- restored Homebrew and application preferences after field testing

## Notes

Homebrew currently marks the Alacritty cask as deprecated because it fails Gatekeeper checks and schedules it for disabling on September 1, 2026. The UI warns users and reports launch verification failures instead of claiming success.

Ref #598

Issue #598, opened 2026-08-10
Allow configuring the terminal app used for Homebrew sudo commands (currently hardcoded to Terminal.app)


untragaluz:
### Problem or Motivation

When Homebrew operations need elevated privileges (e.g. installing/uninstalling a cask that requires sudo), Vorssaint opens a terminal window to prompt for the password. This currently only works with Apple's built-in Terminal.app — it's hardcoded via AppleScript (`tell application "Terminal" ... do script ...`) in `HomebrewManager.openTerminal(command:)`.

I don't use Terminal.app day to day — I use Ghostty (and I imagine other users are on iTerm2, Warp, Alacritty, Kitty, etc.).

### Proposed Solution

- Add a "Preferred terminal" setting (e.g. in Settings → Homebrew, next to the existing terminal-related options) that lets the user pick from installed terminal apps, or at least offers a few common ones (Terminal, iTerm2, Ghostty, Warp).
- When the fallback/sudo terminal needs to open, use the selected app instead of always defaulting to Terminal.app.

Ghostty doesn't support the same `tell application ... do script` AppleScript pattern as Terminal.app, so this probably can't be a simple string swap in the existing AppleScript — it likely needs an app-specific strategy per terminal (AppleScript for Terminal/iTerm, and something like opening Ghostty with a `-e` flag or writing a temp script + `open -a Ghostty` for terminals that support that instead).

Happy to test on Ghostty specifically if that helps.

Version: v3.3.1

### Feature Area

Tools

KSI-cell:
I would like to take this issue.

For the initial implementation, I propose supporting these installed terminal choices:

- Terminal.app (keep the existing AppleScript path)
- iTerm2 (AppleScript)
- Ghostty (native AppleScript support; Ghostty 1.3.0+)

The preferred terminal would be used consistently for the Homebrew sudo fallback, the Homebrew installer, and shell setup. The setting would only list supported apps that are installed, and would fall back to Terminal.app when the selected app is unavailable.

I suggest leaving Warp, Alacritty, and Kitty out of the first pass. They need separate app-specific launch/command-injection strategies and dedicated testing rather than a shared AppleScript path. They can be added later as independent adapters once their integration behavior is agreed.

References: https://ghostty.org/docs/features/applescript and https://iterm2.com/documentation/2.1/documentation-scripting.html

PathGao:
[...]

PR #945, opened 2026-08-25
feat(homebrew): allow custom terminals for Homebrew commands

KSI-cell:
## Summary

- add a preferred terminal setting for interactive Homebrew commands
- detect and support Terminal.app, iTerm2, Ghostty, WezTerm, and Alacritty
- use the selected terminal for installer, shell setup, and sudo fallback commands
- keep Automation permission reporting aligned with the selected Apple Events terminal
- verify CLI-launched terminals, handle stale launches and fallback errors, and refresh availability after cask changes
- localize the new settings and status messages in every supported language

## Validation

- rebased as a single clean commit on current `main` after the upstream history rewrite
- `./build.sh --test` — 8,727 checks passed
- `./build.sh` — release bundle built successfully
- `./build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest` — `SELFTEST OK`
- field-tested Homebrew formula install/test/uninstall and cask install/uninstall
- field-tested Terminal.app, iTerm2, Ghostty, WezTerm, and Alacritty, including sudo fallback, launch verification, preferred-terminal fallback, and cleanup
- restored Homebrew and application preferences after field testing

## Notes

This replaces #709. That PR was closed because the upstream history rewrite made its otherwise single feature commit appear as an 82-commit, 184-file branch. This PR starts from current `main` and contains only the terminal-selection change.

Homebrew currently marks the Alacritty cask as deprecated because it fails Gatekeeper checks and schedules it for disabling on September 1, 2026. The UI warns users and reports launch verification failures instead of claiming success.

Refs #598


PathGao:
`#598`

Three things.

**1. Thirteen plain strings became format templates and nothing catches a raw render.**

`homebrewOpenTerminal`, `homebrewOperationTerminal`, `homebrewTerminalFallback`, `permAutomationTerminal` and the rest used to be safe to drop straight into a `Text(...)`. Now any call site that doesn't go through `homebrewTerminalText` / `automationTerminalName(for:)` renders a literal `%@`. Type is still `String`, so the compiler sees no difference. I only got 60% of the diff, so I can't tell whether every call site was converted.

Separately, `String(format: template, terminalName, terminalName)` always passes two arguments. That is an unwritten cap of two `%@` per template across 13 language files — a translator adding a third gets garbage off the va_list, not a build error. Every template I can read is 1 or 2, so it holds today.

One test covers both halves. Whatever collection the existing localization tests iterate over:

```swift
let expected: [KeyPath<Strings, String>: Int] = [
    \.homebrewOpenTerminal: 1,
    \.homebrewMissingBody: 2,
    \.homebrewShellSetupBody: 2,
    // ... the other ten
]
for strings in allLocalizations {
    for (key, count) in expected {
        #expect(strings[keyPath: key].components(separatedBy: "%@").count - 1 == count)
    }
}
```

**2. The default is written twice.**

```swift
DefaultsKey.homebrewPreferredTerminal: "terminal", // HomebrewTerminal.rawValue
```

The comment is doing a reference's job. `HomebrewTerminal.terminal.rawValue` instead — rename the case and the literal silently becomes a bad key that falls back.

**3. Which terminal does the Automation permission row report on?**

Ghostty, WezTerm and Alacritty ship no AppleScript dictionary — no `do script`, so they have to be CLI-launched, and TCC will never hold an Apple Events entry for them. The body says reporting stays aligned with "the selected Apple Events terminal", which reads like you handled it, but `automationTerminalName(for:)` accepts any `HomebrewTerminal`. If the selected one is passed straight through, the Features page grows a "Ghostty automation" row that permanently reads not-granted while the feature works fine. That call site isn't in my slice.

**Not verified:** anything after `Strings+Turkish.swift` — `HomebrewTerminal`, `HomebrewTerminalSupport`, the launch and verification code, all UI. One thing I'd look at while you're in there: `preferredTerminal()` sits in a default argument, so it re-resolves on every single string render. If it probes LaunchServices or disk for availability, that's a main-thread hit per SwiftUI body evaluation. Can't tell from here.

**Impact:** none confirmed. Worst case from (1) is a literal `%@` shown in a Homebrew string in some language; worst case from (3) is a permanently red permission row for three of the five terminals.
**Merge:** needs an answer on (3) first.

PathGao:
Closing this one on shape, not on the want. The want is real: the Homebrew feature opens Terminal.app for the sudo fallback, the installer and the shell setup, and a Ghostty or iTerm user gets an app they do not use popping up. That stays on #598.

What this branch does is a fixed list of five terminals by bundle id, with a per-app AppleScript strategy for iTerm2 and Ghostty. That is the "hardcoded integrations with particular third-party apps" kind CONTRIBUTING lists as declined, and the reason is not taste: the day this merges, Warp, Kitty and Alacritty users ask for their entry, each with its own launch strategy and its own testing, and every terminal that changes its scripting surface becomes a Vorssaint bug. Five is where that list starts.

The shape that does not name any app: write the command to an executable `.command` file and hand it to the system, either through the user's default handler for that file type or through one `.app` the user picks in Settings, via `NSWorkspace`. Vorssaint then never knows which terminal is on the other side. What I have not verified is which terminals run a `.command` file when opened that way; Terminal and iTerm do, Ghostty and Warp need checking. I have left that on #598 as the direction for whoever picks it up, and the three entry points you mapped are the useful part of this branch for them.

PR #946, opened 2026-08-25
fix(shelf): decode content thumbnails at insertion and widen format support

tenbux:
## Summary

Follow-ups on the shelf's content-thumbnail path from the review on #785, all in the same mechanism that PR introduced.

- **A thumbnail could be decoded and then thrown away.** The decode was fired from `fileItem`, which builds an item before it is in `items`. `acceptMixedBatch` appends nothing until its slowest provider resolves, so a fast decode finished with no item to patch and a 50ms retry was the only thing standing between it and being dropped. The decode now starts once the item is actually shelved, which removes the timer rather than tuning it.
- **File kind was two hardcoded extension lists.** The video one missed `m2ts`, `mts`, `ogv`, `m2v`, `mxf`, `flv`, `vob` and `divx`, which got no thumbnail at all. The image one missed `psd`, `jp2`, `avif`, `ico` and the raw camera formats, which skipped the inline decode that exists to show a dropped image immediately. Both now ask `UTType`, the same question `NSWorkspace` already answers to pick the fallback icon.
- **Raw camera files decode on the async path rather than on the drop.** They need demosaicing rather than the subsampled downscale ImageIO gives a JPEG or a TIFF, so a dropped `.cr2` would otherwise hold the drop on a sensor-sized file. `UTType.rawImage` asks the question rather than another list.
- **An image still decoding was sized as though it had a thumbnail.** The tile inset read `isImage || hasContentThumbnail`. That was fine while images always decoded inline, but #785 moved restore's image decode off the main thread, so a restored image wears its generic icon briefly and was drawing it at the thumbnail inset. The inset now follows `hasContentThumbnail` alone, which is what the flag was added for.
- `pixelSize` takes its `scale` explicitly rather than defaulting to `backingScale`, so the `NSScreen` read a caller depends on is visible at the call site instead of hidden in a helper. This is a readability change, not a threading one: the callers that used the default now do the same read themselves.
- A pile's icon and its thumbnail flag are now derived in one place instead of two that had already drifted on title.

## Verification

- `./build.sh`: no warnings
- `./build.sh --test`: `TESTS OK (8541 checks)`

Checked in a running build:

- **The dropped-frame race**, reproduced deliberately: build the items, append 1.5s later, the way a slow provider does. The frame lands. On the previous code the 50ms retry expires and it is lost.
- **`m2ts`** now shows a real frame. **`mkv`** shows the generic icon at the generic inset, since AVFoundation cannot decode Matroska and the decode returns nil. Both are what the change intends.
- A collapsed pile takes its first child's frame at the content-thumbnail inset.

`UTType` was checked against both lists before swapping: all 9 image extensions and all 10 video extensions they held conform, so nothing is lost. `UTType.rawImage` was checked the same way, catching all of `cr2 nef arw dng raf orf rw2 pef srw` and none of `jpg png psd tiff heic jp2 avif gif webp bmp ico`.

Inline decode cost at the tile's 64pt, best of five with a warm cache, since widening the classifier widens what the drop pays for:

```
pic1.jpg    1 MB ->  8.0 ms
big.tiff   36 MB -> 10.6 ms
pic.jp2     2 MB -> 15.2 ms
big.psd    28 MB -> 71.9 ms
```

ImageIO subsamples, so file size is not what predicts the cost: a 36MB TIFF is 10.6ms while a smaller PSD is 72ms.

Per rule 10, the call sites I checked rather than the one in front of me: every write to `items` in `ShelfService` (`append` for one item and for many, the three `batchItem` merge branches via `merge(_:into:)`, restore's `keptRestored`), plus `replaceItem` and `handleDeadDrag`, which replace or prune existing items and cannot introduce a file item that never got a decode. For the inset change, `Item.isImage` has two readers besides the inline-decode gate in `fileItem`: the `hasSameContent` comparison, and `tooltipLeafKinds`, which sorts a pile's leaves for its tooltip breakdown. Widening it means a `psd`, raw or `avif` inside a pile is counted there as an image rather than a file, which is the more accurate answer. Nothing sizes a view off it, so it cannot put anything out of step with the tile.

Checked on Apple Silicon, a single Retina display and English, against a `--dev --install` build rather than a release one. Not tried on Intel, on a non-Retina or multi-display setup, or in another app language.

## Anything else

Costs I took deliberately, each one wasted decode:

- **PSD stays on the inline path at ~72ms.** It is a one-off on a deliberate drop, and deferring it would trade that for a visible flash of the generic icon on a format people do drop. Raw is deferred instead because it is the one class that is both slower and not measurable here.
- `mkv` and `webm` conform to `.movie` and so still start a decode AVFoundation cannot finish. Excluding them would mean hardcoding a list again, which is what this removes.
- An image whose inline decode returns nil now gets one async attempt as well, because the retry guard is only `!hasContentThumbnail`. The old code never retried. It gives a transiently unreadable file a second chance.

I could not measure raw decode time: there is no camera raw file on this machine to test with, which is part of why it is deferred rather than timed and kept.

Refs #785


PathGao:
Two things from the diff.

**`pixelSize`'s explicit `scale` is a rename, not a thread fix.** The body says a default "put that read back on whatever thread the caller was on" — but both call sites the default served now do the read themselves:

```swift
let pixels = pixelSize(for: pointSize, scale: backingScale)
```

Same `NSScreen.main` read, same thread, just spelled out. If `estimatedBitmapCost` is called from the decode task to size a cache entry, that's still an off-main `NSScreen` read. Making the dependency visible at the call site is worth doing on its own — it just doesn't close the hole the body describes. I can't see either function's callers, so I don't know whether any of them run off-main.

**The image list survived the UTType swap.** `fileItem` still classifies images by extension while `contentThumbnailKind` asks `UTType`. Two answers to one question, inside the function the PR is de-duplicating:

```swift
let imageExtensions: Set<String> = ["png", "jpg", ...]   // still here
if type.conforms(to: .image) { return .image }           // new
```

`psd`, `jp2`, `avif`, `ico` and the raw camera formats conform to `.image` and aren't on the list. They now get a thumbnail through the new path, but `isImage` is false for them, so an interactive drop skips the inline decode that exists precisely to show it immediately, and anything else reading `Item.isImage` gets the narrow answer. `let isImage = Self.contentThumbnailKind(for: url) == .image` is the whole change if `isImage` means what its name says. I can't see the other `isImage` readers — if it's deliberately narrower than "is an image", that belongs in the body.

Smaller: an image whose inline decode returns nil now gets a second async decode of the same file, since the guard is only `!hasContentThumbnail`. The old code never retried. One failed decode per unreadable image — same class of cost as the mkv case you already called out, just not mentioned.

Unverified: the inset change is in `ShelfTilesView` only. If anything else sizes an image well off `isImage` — a drag image, an expanded pile — it now disagrees with the tile. Not in the diff, didn't check.

**Impact:** none on formats that work today; every case still degrades to the generic icon. The `isImage` split only affects files the extension list never covered.
**Merge:** needs the image classifier reconciled with `contentThumbnailKind` first (or a line in the body saying why `isImage` stays narrow).

tenbux:
`Item.isImage` has no readers that branch on it. Its only references in the tree are its own initializer and the `hasSameContent` comparison, so nothing else was ever sizing off it and widening it cannot put anything out of step with the tile. That answers the part you could not check, and it means the narrow answer was not deliberate, just stale.

Reconciled anyway, since two classifiers inside the function that is meant to de-duplicate is the thing this PR is about:

```swift
let isImage = Self.contentThumbnailKind(for: url) == .image
```

`imageExtensions` is gone. I checked both lists against `UTType` before swapping: all 9 image extensions and all 10 video extensions conform, so nothing is lost. What is gained is `psd`, `jp2`, `avif`, `ico` and the raw camera formats (`cr2`, `nef`, `arw`, `dng`, `raf`, `orf`), which now get the inline decode on drop instead of nothing. ImageIO does decode them; I checked with a `jp2` and it produces the same 128x96px the tile asks for.

On `pixelSize`, you are right and the body was wrong. Passing `backingScale` explicitly is the same read on the same thread, so it is a readability change rather than a threading one, and I have reworded it. The hypothetical does not hold either: `estimatedBitmapCost` is only called from `ResponsibleProcess`, which has no async paths, and never from the decode task. So there is no live off-main read to close, which is why the wording was doing more work than the change.

The double decode is real and now named in the body: an image whose inline decode returns nil gets one async attempt as well, since the guard is only `!hasContentThumbnail`. One decode per unreadable image, and it gives a transiently unreadable file a second chance, so I left it rather than adding state to track attempts.

PathGao:
The classifier reconciliation pulls raw and `psd` into the *inline* decode too, and that path is synchronous on the drop.

```swift
let inlineThumbnail = isImage && !deferImageThumbnail
    ? ImageThumbnailer.thumbnail(for: url, pointSize: Self.contentThumbnailPointSize)
```

Dropping a `.cr2` used to cost nothing — the extension list said not an image, straight to the generic icon. Now it's `isImage`, so the drop decodes a 30MB raw before the tile shows up. Same for a layered `psd` or a large `tiff`. "Nothing here regresses against main" is true about which icon you end up with; it doesn't cover the time to get there.

How much that matters depends on the URL overload of `ImageThumbnailer.thumbnail`, which isn't in the diff. `CGImageSourceCreateThumbnailAtIndex` with a max pixel size pulls the embedded preview out of a raw and costs tens of ms; anything that materializes the full image first doesn't. I didn't check which one it is — your `jp2` check measured output size, not time.

If it's the subsampled path, one line in "Anything else" covers it. If it isn't, `deferImageThumbnail` is already the lever for formats that are slow to decode.

**Impact:** drop latency on raw, `psd` and large `tiff` only. Formats that had a thumbnail before are unchanged.
**Merge:** ship it — the classifier split is closed and `isImage` now means what its name says. The above is a body line unless the decode turns out to be unsubsampled.

tenbux:
Measured it, and you are half right in a way that changed the fix.

ImageIO does subsample even with `kCGImageSourceCreateThumbnailFromImageAlways`, so file size is not what predicts the cost. Best of five at the tile's 64pt, warm cache:

```
pic1.jpg    1 MB ->  8.0 ms
big.tiff   36 MB -> 10.6 ms
pic.jp2     2 MB -> 15.2 ms
big.psd    28 MB -> 71.9 ms
```

A 36MB TIFF costs 10.6ms, so large `tiff` is not a problem, and it was on the old list anyway. PSD is 72ms, which is real but a one-off on a deliberate drop.

Raw I could not measure, having no camera raw sample to test against, and I was not going to guess at the format most likely to be expensive. So I took your other suggestion instead: raw decodes on the async path now rather than on the drop.

```swift
private static func decodesCheaplyInline(_ url: URL) -> Bool {
    guard let type = UTType(filenameExtension: url.pathExtension.lowercased()) else { return false }
    return !type.conforms(to: .rawImage)
}
```

`UTType.rawImage` rather than another list. Checked it catches `cr2 nef arw dng raf orf rw2 pef srw` and none of `jpg png psd tiff heic jp2 avif gif webp bmp ico`. A dropped raw shows its generic icon for a moment and patches in, which is the path this PR just made reliable.

PSD stays inline at 72ms and is named in the body, along with the other two costs I took on purpose. Deferring it would trade 72ms for a visible flash on a format people do drop, which seemed the worse end.

The numbers and the raw deferral are in the description now.

tenbux:
Correcting myself: `isImage` does have a reader that branches on it, and I said it did not.

`Item.tooltipLeafKinds` sorts a pile's leaves for its tooltip breakdown:

```swift
case .file: return [isImage ? .image : .file]
```

My grep required a `.` prefix and so never saw the bare reference inside the struct's own computed property. The claim was wider than the check behind it.

The effect is that a `psd`, raw or `avif` inside a pile now counts toward "N images" rather than "N files" in that tooltip. That is the more accurate answer for those formats, so the behaviour is an improvement rather than a regression, but it is a behaviour change and it should have been in the description rather than described away. The body now names both readers.

Nothing sizes a view off `isImage`, so the part that mattered for the inset still holds.

PathGao:
The body's wording is fixed, but the code comment still carries the old claim:

```swift
/// ...a default here would silently put an NSScreen read back on whatever
/// thread the caller happens to be on.
```

Both call sites in the diff now pass `scale: backingScale` — same `NSScreen` read, same thread, moved up one frame. The comment says removing the default closes a threading hole that the change doesn't close, and it's the copy that outlives the PR body. If some caller outside the diff does hop to the main actor first, that's the sentence to write; I can only see the two that don't.

Smaller: `startContentThumbnails` now runs from four insertion sites and its only guard is `!hasContentThumbnail`, so an item inserted twice before its first decode lands gets a second decode. The merge path is where that's reachable — it recurses into batches, so piling a 10-file pile onto another restarts every child still in flight. I can't see where `leaves` comes from; if they're always fresh from the pasteboard this can't happen and it's nothing.

**Impact:** none — a stale comment, and at worst a duplicate decode of files already being decoded.
**Merge:** ship it. Both are one-liners for whenever, not reasons to hold.

Issue #948, opened 2026-08-25
add a "Flush the DNS cache" action to Quick toggles


tingke:
### Problem or Motivation

After switching DNS servers (VPN, Pi-hole, hosts-file edits, or a stale record after a site changes IPs), macOS keeps serving the old answers from its caches. The fix today means opening Terminal and remembering which two commands to run:

```sh
sudo dscacheutil -flushcache; sudo killall -HUP mDNSResponder
```

Vorssaint's 快捷操作 (Quick toggles) tab already collects exactly this kind of one-click system action — empty Trash, eject all disks, toggle hidden files — so a person reaching for the panel still has to drop to a shell for DNS.

### Proposed Solution

Add a **"Flush the DNS cache"** action row to the Quick toggles tab (and the surfaces that mirror it: quick panel ⌃⌘V, command bar, radial menu).

- One elevated shell does both halves: `dscacheutil -flushcache && killall -HUP mDNSResponder` (the signal to mDNSResponder needs root, so the pair rides a single administrator prompt, the same `AdminShell` path the existing tools use).
- The row reports the usual short-lived running/failed state; declining the password prompt reads as a failure, matching how other actions behave.
- The row supports the tab's standard per-item visibility toggle and drag-to-reorder.
- No daemon, no polling, nothing runs while the panel is closed — same on-demand contract as every other toggle.

### Feature Area

Tools

### Does another app already do this?

Yes — Raycast ("Flush DNS Cache" command), and many "utility packs" ship a shortcut for the same two-command incantation. The behaviour to match is the plain system flush itself, nothing app-specific.

### Mockups / References

Placed after "Eject all disks" in the default row order, using the `network` SF Symbol:

```
⏏  Eject all disks         Safely ejects every external disk.
⛿  Flush the DNS cache      Clears the system's cached DNS lookups.
                            Asks for the administrator password.
👁  Show hidden files       The Finder restarts to apply it.
```

PR #949, opened 2026-08-25
feat(toggles): add a "Flush the DNS cache" action

tingke:
## Summary

Adds a **"Flush the DNS cache"** action to Quick toggles, as requested in Refs #948.

One click runs both halves of a real flush — `dscacheutil -flushcache` for the Directory Service cache, then `killall -HUP mDNSResponder` for its own — through a single elevated shell, since the signal needs root and the pair rides one administrator prompt on the existing `AdminShell` path (same as the other tools). Declining the prompt reads as the row's usual short-lived failure.

- New `flushDNS` case in `QuickToggleAction`, placed after "Eject all disks" in the default order; the standard per-item visibility key (`panelToggleFlushDNS`) and drag-to-reorder work like every other row.
- Mirrors into the surfaces that share the list: quick panel (⌃⌘V), command bar (`action.flushDNS`), radial menu.
- Nothing runs while the panel is closed — same on-demand contract as the rest of the tab; no daemon, no polling, no new permission.
- Strings in all 13 supported languages, plus a CHANGELOG entry under Unreleased.

## Verification

On a MacBook Pro (Apple silicon, arm64), macOS 26.5 SDK, release-style `swift build` plus the project's own harness:

```sh
./build.sh --test
# TESTS OK (8542 checks)
```

The new assertion pins the exact elevated command (both halves present). What I could **not** test: the administrator password prompt end-to-end on another Mac, and macOS 14/15 (built against the macOS 26 SDK only); the `AdminShell` path itself is unchanged and already covered by the closed-lid and updater features.

## Anything else

Refs #948.


PathGao:
#948

Read the diff. One thing worth checking, and it's the body's own claim:

> the standard per-item visibility key (`panelToggleFlushDNS`) and drag-to-reorder work like every other row

The visibility key is fully wired — `Defaults` registry, `@AppStorage`, `visibilityBinding`. Reorder isn't in the diff. `QuickToggleAction`'s case order is only the *default*; anyone who has dragged a row has a persisted order array that predates `flushDNS`. Whether a new case gets appended or silently dropped is decided in the `PanelOrderItem` reconcile code, which isn't in the diff, so I couldn't verify it. Confirm it appends unknown cases rather than intersecting against the stored list.

`RadialMenuQuickToggle` is safe either way — it decodes by raw value, and you inserted the case without renaming anything.

The rest checks out. 13 locale blocks, matches the claim. Both non-compiler-checked surfaces are updated (`Defaults` registry, the `@AppStorage` var); every other call site is an exhaustive `switch`, so the compiler found them for you. The `&&` means a nonzero `dscacheutil` skips the HUP, but that call doesn't fail in practice.

I didn't run anything — no machine verification from me, including the admin prompt.

**Impact:** none, assuming the order reconcile appends new cases.
**Merge:** ship it.

PathGao:
good to merge but I want use #977 to solve this kind of issues entirely。You can check that。

PR #951, opened 2026-08-25
fix(menu-bar): keep mic toggle from moving panel

FloatingPegasus:
## Summary

Toggling microphone mute while the menu panel was open changed the width of the main status item by adding or removing its red mic badge. AppKit then moved the anchored panel and its arrow with the resized item.

Keep the rendered mic badge state fixed while any Vorssaint status item anchors an open panel, since resizing the main item can also shift adjacent metric items. The Quick Toggles row still reflects the live mute state immediately, and the current badge state is rendered after the final panel closes.

Anchor switches retain the hold. A close callback releases it only when no replacement panel is shown, and a failed replacement releases it explicitly. The mic badge predicate and its held rendering now have one source of truth for both icon and title spacing.

## Verification

Tested on an Apple M5 MacBook Air running macOS 26.5.2 (25F84), using locally built development and release apps against the macOS 26.5 SDK.

```sh
./build.sh --test              # TESTS OK (8561 checks)
./build.sh --dev               # development app bundle built successfully
./build.sh                     # optimized app bundle built successfully
                                # SELFTEST OK
```

The regression checks cover held versus live badge rendering and both close orderings: an anchor switch in progress and an older close arriving while a replacement panel is shown.

Manually opened Quick Toggles, muted and unmuted the microphone, switched to Keep Awake, and returned to Quick Toggles. The panel frame stayed at `x=1188, y=26, width=358` throughout; only the expected content height changed between sections. The row updated immediately, and the test preferences were restored afterward.

I did not complete a separate manual metric-anchor click or test external-display and multi-Space layouts. No user-facing text changed.

## Anything else

I searched open and closed issues and pull requests for microphone, mic badge, popover, and panel twitch reports and found no matching existing work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,13 @@ All notable changes to this project are documented here. The format follows
 ### Added
 - General settings now include a Liquid Glass toggle on macOS 26 and later to control translucent glass visual effects across panels and controls.
 - The video and GIF tools now include an option to compress directly to a target file size in megabytes, automatically deriving the resolution and frame rate to stay under the limit. Thanks to @FlowSync0.
-- Shelf now shows real video frame thumbnails for dropped video files, and decodes saved image thumbnails asynchronously on launch. Thanks to @tenbux.
+- Shelf now shows real content thumbnails for dropped images and videos, covering more file types, and decodes saved image thumbnails asynchronously on launch. Thanks to @tenbux.
 - The Command Bar now offers a compact mode: the bar opens as a field alone and shows results only once you type. Press Down to see the usual list anyway. Under Command Bar, off by default. Thanks to @kobebrylant.
 - Command Bar script links can now be marked to run on their bare name without an argument, so scripts that work on the clipboard, selection, or files don't need a placeholder word typed after the name. Under Command Bar, off by default. Thanks to @rhukster.
 - Monitor alerts can now warn you when the battery stays above a temperature you choose. Under Monitor alerts, off by default. Thanks to @ywu73.
 
 ### Changed
+- Homebrew package rows can now be selected by clicking the unused space beside the name. Thanks to @pergioa.
 - Clipboard history now displays image previews and thumbnails for image files copied from Finder in the quick panel, preview inspector and menu panel.
 - Clipboard history now remembers whether you left the preview inspector open or closed across launches.
 - New Scratchpad tabs now start at 1 instead of leaving the first tab unnumbered, while existing names stay unchanged. Thanks to @AB-boi and @JashRashne.
@@ -22,6 +23,9 @@ All notable changes to this project are documented here. The format follows
 - The Keyboard Shortcuts page now lists every capture shortcut under one Screen capture group, and expandable shortcut groups are toggled by their whole row, with the chevron moved to the trailing edge. Thanks to @RecoilGaming.
 
 ### Fixed
+- Fan Control can no longer be installed on a Mac with no controllable fan, including from Install all and first-run setup. Thanks to @PathGao and @Yahddyyp.
+- Quit on close no longer causes elevated background CPU usage in watched applications. Thanks to @iltonandrew.
+- The Displays section no longer freezes when switching off the built-in display while an external monitor is connected. Thanks to @iltonandrew.
 - Subprocess sampling and window enumeration no longer exhaust background dispatch threads or deadlock the main thread. Thanks to @PathGao, @iltonandrew and @SudhanshuBhogal.
 - Quick tool confirmation panels and scrolling capture banners now keep long copied text or parameter lists bounded on screen instead of overflowing the display. Thanks to @rhukster.
 - App Switcher now keeps walking backwards when you hold Shift and tap Tab again. Thanks to @iltonandrew and @justin-chiam.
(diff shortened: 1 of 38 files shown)


PathGao:
Three things.

**1. `popoverDidClose` releases the hold unconditionally, and anchor switching runs through it.**

The `!popoverIsSwitchingAnchor` check right below it says a close can be immediately followed by a re-show. If AppKit delivers the old anchor's `popoverDidClose` *after* the new `show` (animated close makes that ordering plausible), `setMicBadgeHeld(false)` clears a hold the new show just set, and the badge is live again for the rest of that panel session. `guard heldMicBadgeActive != nil` can't tell the two apart — the hold isn't keyed to anything.

Cheapest shape: set or clear at every show instead of balancing calls,

```swift
statusController.setMicBadgeHeld(holdsMicBadge)
```

and move the `popoverDidClose` release under the existing `!popoverIsSwitchingAnchor` branch. Whether that closes the window depends on when `popoverIsSwitchingAnchor` is reset, which isn't in the diff — I didn't verify it.

**2. The badge predicate is now written twice, and the two copies read different stores.**

`setMicBadgeHeld` reads `UserDefaults.standard`; `updateIconAppearance` reads `defaults`. If `defaults` is anything else (a suite, an injected instance), the snapshot and the live value come from different places and the hold freezes the wrong state. Even if they're the same object today, it's copy #2 of the same fact. One property, both call sites:

```swift
private var micBadgeActive: Bool {
    MicMuteService.shared.isMuted && defaults.bool(forKey: DefaultsKey.micMuteMenuBarIndicator)
}
```

`defaults`'s declaration isn't in the diff, so I can't say which store it is.

**3. Body: "metric status items ... do not hold the main mic badge, and metric re-anchoring remains unchanged."**

That premise holds only if a metric item's own frame can't move when the main item resizes. Status items are packed adjacently — a width change in one shifts the ones next to it. A metric item sitting beside the main one and anchoring an open panel would still get dragged, and the panel with it, which is the same bug this PR is fixing. The menu bar layout code isn't in the diff so I didn't check the ordering; if items are placed so this can't happen, ignore it.

No open issue matches this one, so there's nothing to link.

**Impact:** #1 — after the panel re-anchors to the main status item, the original twitch can return for that panel session; only reachable with separate metric status items. #2 — none if `defaults` is `UserDefaults.standard`, wrong frozen state if not.
**Merge:** needs the hold release keyed to the anchor (or evidence the switching close can't land after the next show) first.

FloatingPegasus:
Updated in `1324f97`.

1. Every successful panel presentation now acquires the same hold before `show`, including metric anchors. `popoverDidClose` releases it only for a final close (`!popoverIsSwitchingAnchor && !popover.isShown`); an anchor-switch close or an older close delivered after a replacement is shown retains it. Failed presentations and a failed metric-anchor replacement release explicitly. The regression check covers both non-release orderings and the final release.
2. The predicate is now `currentMicBadgeActive` in one place. `renderedMicBadgeActive` applies the held value in both icon rendering and the title-spacing path. The previous local `defaults` was also `UserDefaults.standard`, but there is no longer a duplicated predicate or store ambiguity.
3. The hold now applies while any Vorssaint status item anchors the panel, since changing the main item's width can shift an adjacent metric item.

Validation: `TESTS OK (8544 checks)`, development and optimized builds succeeded, and `SELFTEST OK`. In the manual main-anchor sequence, mute, unmute, Keep Awake, and returning to Quick Toggles all kept the panel at `x=1188, y=26, width=358`; only the expected content height changed. I did not count a separate metric-anchor click as manually verified, as noted in the updated PR body.

PathGao:
Both close orderings are covered now, and the predicate reads from one place. Two smaller things, neither blocking.

**The hold survives a show that never becomes a panel.** `setMicBadgeHeld(true)` runs before `popover.show`, and every release comes from a close callback. The line right after the show — `if let window = popover.contentViewController?.view.window` — treats a missing window as reachable; on that branch nothing releases and `popoverDidClose` never fires, so the badge stays at the snapshot. It self-heals on the next open-and-close, so the cost is a stale mute badge until then. If that optional is defensive rather than a real state, ignore this — I can't tell from the diff.

**The release predicate is written twice.** `AppDelegate.swift:662` does `if !self.popover.isShown` inline while `popoverDidClose` goes through `releasesMicBadgeHold`. Same rule, and `popoverIsSwitchingAnchor` is set false two lines above, so the helper drops straight in:

```swift
if MenuBarSpacingSupport.releasesMicBadgeHold(
    isSwitchingAnchor: self.popoverIsSwitchingAnchor, popoverIsShown: self.popover.isShown) {
    self.statusController.setMicBadgeHeld(false)
}
```

The metric-anchor click you left unverified is the one path where acquisition and release run through different call sites. That's what I'd exercise first if this comes back.

**Impact:** stale mic badge until the next panel open-and-close, and only if a presentation can end without a window.
**Merge:** ship it.

FloatingPegasus:
Addressed both non-blocking points in `05fae44`.

- If `popover.show` does not produce a window, the hold and drift-correction state are now released immediately before returning. The normal presentation path is unchanged.
- The failed metric-anchor replacement now uses `releasesMicBadgeHold` instead of repeating the `!popover.isShown` rule inline.

Revalidation: `TESTS OK (8544 checks)`, optimized build succeeded, and `SELFTEST OK`.

PathGao:
`05fae44` does more than release the hold. The nil-window branch now returns before `NSApp.activate` and before the `guard popover.isShown` path, so it treats a missing window as a failed presentation. The original code didn't assume that — the separate `isShown` guard exists precisely for the case where the two disagree. If a shown popover can miss `contentViewController?.view.window` at that instant, the panel now comes up unactivated and skips the rest of the post-show setup, which the old code ran.

Only the hold release needs the early exit. `endPopoverDriftCorrection()` and `return` assert something stronger than the diff shows, and the rest of the function past that guard isn't in the diff, so I can't tell what else gets skipped. If nil window is provably unreachable while `isShown` is true, this is nothing.

**Impact:** none if a nil window always means `show` failed; otherwise a panel that presents without activation and without post-show setup.
**Merge:** ship it.

FloatingPegasus:
Rebased cleanly onto current `main` (`49f6ee5`); the new head is `ef0f7b2`. Post-rebase verification passes 8,561 checks, the optimized build, and `SELFTEST OK`.

FloatingPegasus:
Good catch. Narrowed in `ed4b247`: the nil-window branch now releases only the mic badge hold, then continues through the original `NSApp.activate`, `popover.isShown` guard, and remaining post-show setup. It no longer calls `endPopoverDriftCorrection()` or returns early.

Final verification on current `main`: `TESTS OK (8561 checks)`, optimized build succeeded, and `SELFTEST OK`.

Issue #33, opened 2026-06-16
[Bug]: Settings window always opens underneath the popover after relaunch


wenujacodes:
### Description

Hey, I noticed something with the Settings window behavior.

Every time I completely quit the app and open it again, the Settings window always appears directly underneath the menu bar popover. It feels like the Settings window is trying to behave like a popover or is inheriting its position logic.

Because of this, I always have to manually drag the Settings window to the middle of the screen, since the popover covers it.

Not sure if this is technically a bug or just unintended behavior, but it's definitely inconvenient.

### Steps to Reproduce

1. Completely quit the app

2. Re-open the app

3. Click the menu bar icon to open the popover

4. Open Settings from the popover

5. Observe that the Settings window appears directly underneath the popover instead of in a normal window position

### Expected Behavior

The Settings window should open in a normal, user-friendly position - ideally centered on the screen - instead of spawning under the popover.

### Actual Behavior

• After relaunching the app, opening Settings places the window directly under the popover

• The popover overlaps the Settings window

• User must manually move the window every time

### Screenshots / Recordings

<img width="1698" height="825" alt="Image" src="https://github.com/user-attachments/assets/932f368d-6a09-4580-bb91-c0e706dadd46" />

### Vorssaint Version

Version 2.16.0

### macOS Version

macOS 15 (Sequoia)

### Installation Method

GitHub Release

### Additional Note

@vorssaint  _These are all the issues I’ve found so far, everything’s running great at the moment. Can’t wait to see what comes next. I’m really happy to contribute to this project, and thank you so much for creating it. Open source grows because of communities like this !_

vorssaint:
Thanks for reporting this, and thank you for taking the time to test and share all these details!

This has been fixed in the latest update. The Settings window now opens in a normal window position instead of appearing under the menu bar popover after relaunch. It should also avoid overlapping the popover when opened from the panel.

Really appreciate your help and support for the project! Let's evolve together, @wenujacodes!

Issue #19, opened 2026-06-16
[Bug]: Menu Bar App Does Not Hide When Clicking the Settings Window


wenujacodes:
### Description

When the Vorssaint menu bar app is active and overlapping the Settings window, clicking on the Settings window does not make the menu bar popover disappear.

The popover only disappears when clicking on other apps, but not when clicking on the app's own Settings window.

This becomes annoying because I have to manually move the Settings window away just to see the menu bar Ul.

### Steps to Reproduce

1. Open the settings windows.
2. Open the menu bar app.
3. Then click on settings windows.

### Expected Behavior

The menu bar popover should automatically hide when the user clicks anywhere outside of it — including the app's own Settings window.

### Actual Behavior

Clicking on the Settings window does not make the menu bar popover disappear.

### Screenshots / Recordings

<img width="958" height="410" alt="Image" src="https://github.com/user-attachments/assets/cba9919d-4c18-4f3a-a97a-7fbc837f3ac9" />

### Vorssaint Version

Version 2.15.1

### macOS Version

macOS 15 (Sequoia)

### Installation Method

GitHub Release

vorssaint:
Hi! Thanks for reporting this. @wenujacodes 

This has been fixed in the latest update. The menu panel now gets out of the way when it is overlapping the Settings window and the user clicks Settings.

At the same time, Settings and the menu panel can still stay open side by side, so it remains easy to adjust the panel layout and see the changes live.

The update is already available in the latest version, so I’m closing this issue now.

PR #952, opened 2026-08-25
fix(settings): close panel when settings cannot avoid it

FloatingPegasus:
## Summary

Vorssaint already tries to place Settings clear of the menu panel so both can remain open for live preview. When a saved Settings window is too large, or the status item is too central, no non-overlapping placement may fit. Opening Settings then leaves the panel covering it until the user clicks the Settings window again.

The complete placement decision now runs in `SettingsWindowSupport`: if an avoided frame fits, use it and keep the panel; if avoidance still intersects, close the panel and use the original preferred Settings frame clamped to the visible screen. AppDelegate applies the chosen frame once.

The popover frame participates only while `popover.isShown`. Both positioning calls are the initial and next-run-loop passes inside `openSettingsWindow()`; no screen-change or restoration callback gains the dismissal side effect.

## Verification

Tested on an Apple M5 MacBook Air running macOS 26.5.2 (25F84), using local Developer and release builds against the macOS 26.5 SDK.

On unmodified main, I maximized Settings, closed it, reopened the menu panel, and selected Settings. Settings took focus while closing it revealed the original panel still open.

On commit `31b50b3`, both live arrangements passed:

- With room beside the panel, the panel remained at `x=1188…1546` and Settings moved to `x=154…1160`, preserving the existing 28-point gap with both windows open.
- With a saved 1920-point Settings width that could not avoid the panel, reopening Settings removed the popover and placed Settings at the visible-frame clamp (`x=20`, width `1880`).

```sh
./build.sh --test              # TESTS OK (8561 checks)
./build.sh                     # optimized app bundle built successfully
                                # SELFTEST OK
```

The tests exercise successful horizontal and vertical avoidance, and verify that failed avoidance closes the panel and actually clamps an out-of-margin preferred frame. I did not separately test external displays or multiple Spaces. No user-facing text changed.

## Anything else

Refs #19. This preserves the behavior described there: keep Settings and the panel together when they fit, and get the panel out of the way when they overlap.

Ref #33


PathGao:
Two things, one of them blocks.

**The settings window keeps the dodge after the panel is gone.** `settingsFrame(_:avoiding:in:)` shifted and clamped `frame` to avoid a panel that `closePopover()` then removes at `AppDelegate.swift:1264`. On your own repro (panel 358 wide at x=731, saved settings 1006 wide) the avoidance can't fit, so Settings lands at a clamped edge position *and then* the panel disappears — the user's saved position is spent dodging nothing. The body doesn't mention that. Decide first, place once:

```swift
if SettingsWindowSupport.shouldClosePanel(settingsFrame: frame, panelFrame: popoverWindow?.frame) {
    closePopover()
    frame = <saved frame, clamped to visible — no avoidance>
}
window.setFrame(frame.integral, display: false)
```

The top of that method isn't in the diff, so I can't tell whether the un-avoided frame is still reachable there. If it isn't, say so and this becomes a follow-up instead.

**The side effect now rides on the positioning path.** Every caller of this method closes the panel, not just "user picked Settings from the panel". The other call sites aren't in the diff — I didn't verify there's only one. If this also runs on screen-config change or window restore, the panel vanishes under the user's hands.

Minor: `popoverWindow?.frame` is a different panel-present test than whatever unwraps the panel frame for the `avoiding:` path above (that one takes a non-optional). If a closed popover can leave `popoverWindow` non-nil with a stale frame, opening Settings with nothing on screen calls `closePopover()`. Harmless, but two predicates for one fact.

Geometry checks in `MetricsTests.swift` are correct for what they cover — they only exercise `shouldClosePanel`, not the placement that follows it.

**Impact:** Settings loses its saved position whenever the close path fires; unknown blast radius if the positioning method has more than one caller.
**Merge:** needs the placement recomputed after `closePopover()` at `AppDelegate.swift:1264` first.

FloatingPegasus:
Thanks — updated in `60db92b`.

The complete avoidance/clamping calculation now lives in `SettingsWindowSupport.panelPlacement`, which returns both the one final frame and whether the panel must close. Successful avoidance keeps the separated frame and the panel. Failed avoidance closes the panel and restores the original preferred frame clamped to the visible screen; `AppDelegate` then calls `setFrame` once.

I also changed panel presence to `popover.isShown ? ... : nil`. The only two `positionSettingsWindow` calls are the initial and next-run-loop passes inside `openSettingsWindow()`, so this does not add a screen-change or restoration dismissal path.

The tests now exercise the complete placement result rather than only an overlap predicate. Live verification covered both branches: a fitting layout kept Settings and the panel open with a 28-point gap, while an enlarged saved Settings frame closed the panel and reopened at the visible-frame clamp. `./build.sh --test` passes 8,543 checks, the optimized build succeeds, and the fresh bundle reports `SELFTEST OK`.

PathGao:
Placement blocker is resolved. `panelPlacement` decides once and `AppDelegate` applies a single `setFrame` after `closePopover()`, so the saved position is no longer spent dodging a panel that's about to disappear. Your `x=20, width=1880` run is exactly the case I was worried about.

One gap in the tests. `overlappingPlacement` uses a preferred frame of `(100, 100, 1000, 700)` in a `1200×900` visible frame, so the clamp range is `[20, 180]` on both axes and `x=100, y=100` passes through untouched. `frame == preferredSettingsFrame` holds whether `clamped()` runs or not — swap `clamped` for the identity function and the assertion still passes. Only the live 1920-wide run covers the clamp. A preferred frame outside the margin would pin it:

```swift
preferredFrame: CGRect(x: -50, y: 100, width: 1000, height: 700)   // expects x == 20
```

The vertical branch is untested too — test 1 falls through `belowY`/`aboveY` without moving, test 2 takes the left branch, so no case exercises an actual vertical shift.

On call sites: the only AppDelegate hunk in the diff is at `:1255`, so I'm taking "the only two `positionSettingsWindow` calls are inside `openSettingsWindow()`" on your word — I didn't verify it.

Minor, and not new to this PR: the 20-point margin is now written in `clamped()` and again as `margin / 2` in `AppDelegate`. Same number in two places with nothing tying them together.

**Impact:** none — both branches behave as described in your live run.
**Merge:** ship it.

FloatingPegasus:
Added the two coverage cases in `649f481`.

- The failed-avoidance input now starts at `x=-50` and asserts that the returned preferred frame is clamped to `x=20`, so replacing `clamped()` with the identity would fail.
- A separate case forces the vertical fallback and asserts that Settings moves below the panel to `maxY=472`, preserving the 28-point gap.

`./build.sh --test` now passes 8,544 checks. This follow-up changes only the tests; the previously verified production code and live arrangements are unchanged. I left the pre-existing margin duplication out of this PR.

FloatingPegasus:
Rebased cleanly onto current `main` (`49f6ee5`); the new head is `31b50b3`. Post-rebase verification passes 8,561 checks, the optimized build, and `SELFTEST OK`.

PR #954, opened 2026-08-25
fix(screenshot): stop scrolling captures repeating fixed page elements

vorssaint:
## What was wrong

Scrolling captures measured page furniture once, at the first accepted scroll, over whole rows and only at the bottom edge. Anything that covered part of the width, let the page show through it, sat away from the bottom, or appeared later than the first scroll was joined into every strip and stamped down the image. A brisk scroll ended the capture and kept only what had been joined so far.

## What changed

- Rows are read from the view a scroll started on, once the next view has shown which of its rows stay put. The element reaches the image once, in the closing piece, and nothing between the views is lost. When a single jump is larger than the clean part of the view, the covered rows were never on screen at all, so a complete image with part of the element in it is preferred to a hole.
- Furniture is found per column and always against the scroll, so a floating button over four columns, a translucent bar and a side element halfway down the view are all seen, while an evenly coloured band of the page is not mistaken for one.
- Matching picks the distance that best explains the two views instead of counting rows that agree: a plain band or a bar agrees with itself at almost any short distance, which is what chose wrong distances on real pages. The header and the bar the page slides under are excluded from the comparison, and a distance is only accepted when it is clearly ahead of the runner-up.
- Brisk scrolls are joined instead of ending the capture.

## Validation

- `./build.sh --test` — 8548 checks, including new focused cases for the clean boundary, the harvest range and a floating button at a Retina selection height.
- Simulated 35 scenarios (full-width bar, floating button with and without margin, translucent bar with jitter up to 20, sticky header, header appearing later, furniture leaving mid-capture, forward-back-forward, pauses, tiny to 88% advances, flat and sparse pages): no lost rows and no repeated rows anywhere.
- Real capture of an adversarial page (sticky header, side element at 44% height, floating pill, bottom bar) driven by synthetic scrolls on an installed Developer build: image height exactly the viewport plus every scroll, bands in order with no gaps or repeats, and each fixed element present only where its pixels are the only ones that exist. Repeated with a header that appears only after the first scroll.
- One match costs 51 ms against 82 ms before, on a 1466x1678 selection in an optimized build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ All notable changes to this project are documented here. The format follows
 - App Switcher now keeps walking backwards when you hold Shift and tap Tab again. Thanks to @iltonandrew and @justin-chiam.
 - The screen capture loupe now rings the pixel under the pointer instead of drawing a crosshair across it, and that pixel sits in the middle of the loupe, so the Color picker shows the color it is about to copy. Thanks to @I-Have-No-Idea-What-Im-Doing-Right-Now and @PathGao.
 - Screenshot editor crop now snaps selection and resize edges to pixel boundaries, keeping the loupe cross and the final cut aligned to the exact source pixel. Thanks to @PathGao.
+- Scrolling screenshots no longer repeat anything that stays put while the page moves under it, including a bar that lets the page show through, a floating button over part of the width and a side element that appears partway down.
+- Scrolling screenshots now keep up with a brisk scroll instead of stopping with only part of the page.
 
 ## [3.3.3-beta.2] - 2026-08-22
 
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotScrollingCapture.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotScrollingCapture.swift
@@ -62,6 +62,23 @@ enum ScreenshotScrollingCapture {
         let global: Any?
     }
 
+    /// The image being built. Rows are read from the view a scroll started on,
+    /// once the following view has shown which of its rows stay fixed, so the
+    /// join always trails the screen by one accepted scroll.
+    private struct Join {
+        var reference: CGImage
+        var slices: [CGImage] = []
+        var appendedHeight = 0
+        var retainedPixels = 0
+        /// Rows of `reference` already appended, counted from its top.
+        var joinedThrough = 0
+        /// The moving columns, once a scroll has identified them.
+        var columns: Range<Int>?
+
+        /// The image the person would get by finishing right now.
+        var projectedHeight: Int { appendedHeight + reference.height - joinedThrough }
+    }
+
     private static let idlePollNanoseconds: UInt64 = 35_000_000
     private static let sampleInterval: TimeInterval = 0.09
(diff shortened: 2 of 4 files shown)


PathGao:
Three things.

**1. `cleanBottom` and `advance` cross the sample/pixel boundary unscaled.**

`advance` is image pixels:

```swift
let overlap = Int((CGFloat(sampleOverlap) / CGFloat(currentSample.height)
    * CGFloat(current.height)).rounded())
let advance = current.height - overlap
```

and it goes straight into `scrollingCleanBottom(previous: referenceSample, current: currentSample, advance:)`, which indexes sample rows (`row + advance < height`, `earliest = header + advance`). Its return, in sample rows, then goes into `scrollingHarvest(imageHeight: join.reference.height, joinedThrough: join.joinedThrough, cleanBottom:, advance:)` alongside two image-pixel values.

The old code did that conversion explicitly and this PR drops it:

```swift
fixedBottomPixels = Int((CGFloat(fixedBottomRows) / CGFloat(currentSample.height)
    * CGFloat(current.height)).rounded())
```

It also fed the sample-space `sampleOverlap` to the sample-space function, where the new call passes pixel-space `advance`.

If sample height < image height, `scrollingCleanBottom`'s `advance > 0, advance < height` guard fails for any ordinary scroll and it returns `height`. `scrollingHarvest` then gets `clean = max(cleanBottom, advance) == advance`, `joinedThrough` pins to 0, and every strip is `0..<advance` of the reference — sticky header included, in every strip. That is the symptom this PR is fixing, restored.

If sample height == image height, all of the above is fine and the `sampleOverlap → overlap` scaling two lines up is an identity that should go. `sample()` is not in my diff, so I can't say which it is — but one of the two is wrong.

**2. Header and footer are measured over the tiles the code just decided don't move.**

```swift
let movingTiles = tiles.indices.filter { still.movedEnough(tile: $0) }
...
still.difference(tiles: tiles.indices, rows: header..<(header + 1)) <= scrollingRowTolerance
```

`tiles.indices`, not `movingTiles`. Where the static part dominates the width — a scrolling pane beside a wide sidebar — the per-pixel average over all tiles sits under 8.0 on every row, so `header` and `footer` each run to `height / 8`. `scrollingComparedRows` loses a quarter of the view before its own inset, and on a longer scroll falls under the `max(24, height / 32)` floor and returns nil, so the frame is `.unmatched`. The fixed-columns test has half the width moving with high-contrast noise, so it stays above tolerance and doesn't cover this. Not verified on a real page.

**3. `scrollingCaptureMaximumFrames` no longer counts what it counted.** `join.slices` starts empty instead of holding the first frame, and only grows when `harvest.rows != nil`, so scrolls that harvest nothing are free and the cap is off by one on top of that. The duration cap still bounds the loop, so this is a declared limit drifting rather than a runaway. Not mentioned in the body.

Read the rest — the explanation ratio, the neighbourhood/decisive split, the prefix-sum rewrite — nothing else to flag.

**Impact:** #1, if the two spaces differ, puts a fixed header back into every strip on every real capture. #2 stalls the capture on pane-scrolling layouts. #3 none visible.
**Merge:** needs the sample-vs-pixel scale on `advance` / `cleanBottom` confirmed first.

Issue #641, opened 2026-08-14
Configurable battery charge limits


Jadens-arc:
### Problem or Motivation

Vorssaint’s Power tab displays detailed battery and power information, but it cannot currently limit charging.

A configurable maximum charge level could help users who keep their Mac connected to power for long periods reduce the time the battery spends at 100%.

A maximum limit alone may cause repeated microcharging near the threshold. For example, the Mac could fall from 80% to 79%, recharge to 80%, and repeat.

### Proposed Solution

Add a Maximum charge level setting to the Power tab.

When enabled:

Allow the user to choose the percentage where charging stops.
Reveal an additional Minimum charge level setting.
Keep charging paused until the battery reaches the minimum level.
Resume charging at the minimum level and stop again at the maximum.

For example:

Maximum charge level: 80%
Minimum charge level: 70%
Charging remains paused between 70% and 80%.

The minimum value should always remain below the maximum value. Disabling the maximum charge limit should return charging behavior to normal and hide or disable the minimum setting.

### Feature Area

Energy and display

### Mockups / References

<img width="1041" height="1510" alt="Image" src="https://github.com/user-attachments/assets/ebc8ec5f-0565-4ece-a69f-a1df6d3b46a1" />

CSkjolden:
FYI, if it can satisfy your needs in the mean time, this is already largely a part of macOS and might be better optimized.
 https://support.apple.com/en-us/102338#:~:text=of%2Dcharge%20estimates.-,Charge,-Limit

PathGao:
macos 26 have it。also duplicate with #846 so close it now

PR #957, opened 2026-08-25
feat(power): add MacBook charge protection

Bald-M:
power

Add a separately managed privileged helper for charge limiting, forced discharge, and session-only top-up controls. Include M4-compatible SMC probing, heartbeat recovery, menu panel controls, settings integration, localization, packaging, and policy tests.

## Summary

This pull request adds an opt-in Beta feature called **Charge Protection** for MacBook batteries.

Charge Protection allows users to:

- Set a maximum charge level between 20% and 100%.
- Automatically allow charging while the battery is below the selected limit.
- Pause charging when the battery reaches the limit.
- Discharge the battery down to the selected limit while the Mac remains connected to power.
- Temporarily override the limit and charge to 100% with **Top Up**.
- View the current battery level, charge-limit marker, charging state, live power information, and applications using significant energy from the menu panel.

The normal limit policy uses a 2% hysteresis to avoid repeatedly switching between charging and paused states near the selected limit. Temporary Discharge and Top Up operations are mutually exclusive and are cancelled when their completion or safety conditions are reached.

The settings page intentionally remains small. It contains only the feature enable switch, the charge-limit control, and an explanation of the Beta/private-hardware-interface limitations. Helper installation is handled when the feature is installed from the Features screen, so users are not asked to install it a second time from Settings or the menu panel.

### Hardware used

Testing was performed on:

- **Mac:** MacBook Air with Apple M4
- **Model identifier:** `Mac16,13`
- **macOS:** macOS 15.2
- **Build:** `24C2101`
- **Power adapter:** 65 W adapter
- **Application build:** Locally built Developer configuration with locally signed application and helpers
- **Application location:** `/Applications/Vorssaint (Developer).app`
- **Display layout:** Built-in display only
- **Spaces/layout:** Normal desktop use; no dedicated multi-display or multi-Space test matrix was performed

The Charge Control helper was confirmed to be registered with `launchd`, running as root, and reachable through its authenticated XPC connection.

### Hardware behavior verified

The following cases were exercised on the M4 MacBook Air:

- At 81% with a 90% limit, charging was allowed.
- Charging resumed with a reported positive battery current of approximately 2366 mA.
- At 81% with a 77% limit, forced discharge started while the adapter remained physically connected.
- At 99% with a 100% limit, charging was permitted again after the macOS power-management delay.
- At 99% with a 90% limit, forced discharge remained stable instead of oscillating between charging and discharging.
- Changing the limit caused the policy to be recalculated.
- The menu panel displayed charging, paused, and discharging states.
- Temporary Discharge and Top Up controls were connected to the session state.
- The helper installation was handled by the Features flow rather than asking for a second installation in Settings.
- The application could be opened from `/Applications` with its installed helper.
- macOS Optimized Battery Charging was disabled during the main hardware-control checks.
- An existing AlDente helper was present on the test machine but was not running.

During testing, AppleSmartBattery was found to report Battery Power and `ExternalConnected = false` while forced discharge was active, even though the cable was still connected. Physical adapter presence is therefore additionally derived from `AdapterDetails`; this prevents policy evaluation from cancelling and restarting forced discharge every five seconds.

The M4 hardware also reads `CH0B`/`CH0C` back as `3` during forced discharge rather than exactly matching the written pause value. Verification now checks the requested charge-inhibit bit while still rejecting an incompatible readback.

### Commands

```sh
./build/Vorssaint --selftest
./build.sh --test
```

### DIFF
--- a/Resources/com.vorssaint.utils.charge-control.plist
+++ b/Resources/com.vorssaint.utils.charge-control.plist
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0"><dict>
+<key>BundleProgram</key><string>Contents/Library/LaunchServices/com.vorssaint.utils.charge-control</string>
+<key>ExitTimeOut</key><integer>60</integer>
+<key>KeepAlive</key><dict><key>Crashed</key><true/></dict>
+<key>Label</key><string>com.vorssaint.utils.charge-control</string>
+<key>MachServices</key><dict><key>com.vorssaint.utils.charge-control</key><true/></dict>
+<key>ThrottleInterval</key><integer>5</integer>
+</dict></plist>
--- a/Sources/ChargeControlHelper/main.swift
+++ b/Sources/ChargeControlHelper/main.swift
@@ -1,0 +1,124 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+import os
+
+private final class ChargeControlController {
+    private let queue = DispatchQueue(label: "com.vorssaint.charge-control.helper")
+    private var hardware = ChargeControlHardware()
+    private var mode: ChargeControlMode = .charging
+    private var lastHeartbeat = ProcessInfo.processInfo.systemUptime
+    private var owner: UUID?
+    private var watchdog: DispatchSourceTimer?
+
+    init() {
+        let timer = DispatchSource.makeTimerSource(queue: queue)
+        timer.schedule(deadline: .now() + 5, repeating: 5)
+        timer.setEventHandler { [weak self] in self?.checkHeartbeat() }
+        timer.resume()
+        watchdog = timer
+        try? hardware?.set(.charging)
+    }
+
+    func opened(_ id: UUID) { queue.async { self.owner = id; self.lastHeartbeat = ProcessInfo.processInfo.systemUptime } }
+    func closed(_ id: UUID) { queue.async { if self.owner == id { self.restore() } } }
+
+    func reply(_ reply: @escaping (Data) -> Void) {
+        queue.async { reply(ChargeControlIPC.encode(self.response())) }
+    }
+
+    func apply(_ requested: ChargeControlMode, id: UUID, reply: @escaping (Data) -> Void) {
+        queue.async {
+            guard self.owner == id else {
+                reply(ChargeControlIPC.encode(.failure(.controlFailed)))
+                return
+            }
+            guard let hardware = self.hardware else {
+                reply(ChargeControlIPC.encode(.failure(.unsupportedHardware)))
+                return
+            }
+            do {
+                try hardware.set(requested)
+                self.mode = requested
+                self.lastHeartbeat = ProcessInfo.processInfo.systemUptime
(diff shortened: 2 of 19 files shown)


PathGao:
#641 #846 #849

**The body claims 2% hysteresis. `desiredMode` has none, and `previous` is unused.**

```swift
guard pluggedIn, limit < 100 else { return .charging }
if percent > limit { return .discharging }
if percent == limit { return .paused }
return .charging
```

Body: "The normal limit policy uses a 2% hysteresis to avoid repeatedly switching between charging and paused states near the selected limit." At limit 80 the actual behaviour is 80 → paused, self-drain to 79 → charging, charge past to 81 → discharging. That is the oscillation the band was supposed to prevent, and the new tests pin it in (`percent: 79, previous: .paused` expects `.charging`; `percent: 81` expects `.discharging`). A `previous` parameter that is accepted and never read says the band was meant to be there and got lost. Either restore it or drop the sentence from the body.

**Automatic discharge is still on, despite the PR saying it was replaced by an explicit action.**

```swift
// Automatic discharge was briefly shipped during the beta. It is now
// an explicit, session-only action, so never carry the old preference forward.
UserDefaults.standard.set(false, forKey: DefaultsKey.chargeLimitDischargeEnabled)
```

`.normal` still resolves to `.discharging` for any `percent > limit`. Plug in at 95%, set the limit to 80, and the Mac force-discharges 15% without anyone asking for it. `chargeLimitDischargeEnabled` is declared, defaulted and force-cleared, but I don't see it read anywhere in the visible diff — so there is no off switch, and the Discharge button is close to redundant: the only thing it adds over the automatic path is the `shouldEndDischarge` check.

**The panel section drops a guard the settings page has.**

`EnergySettings` gates on `AppFeature.batteryChargeLimit.isAvailable, PowerSampler.hasInternalBattery`. `PanelSectionID.chargeControl` gates on the feature alone, and `MenuPanelView.isVisible` returns bare `showChargeControl` while the neighbouring case spells out why that's wrong:

```swift
// The section only earns its navigation tab while the feature is on;
// it is switched on in Settings, not from an empty panel screen.
case .brightness: return showBrightness && brightnessEnabled
```

On a desktop Mac that opts into the beta, the panel shows Charge Protection with an enable toggle while Settings hides the section entirely — enable it there and there's no limit slider to pair with it. Add `PowerSampler.hasInternalBattery` to the panel gate and require `chargeLimitEnabled` like brightness does. `PowerSampler` isn't in the diff I can see, so if feature install already refuses on machines without a battery, only the `chargeLimitEnabled` half of this stands.

**Minor:** the slider is `in: 20...100` with `ChargeLimitPolicy.range` sitting right there — third copy of the same bound. `Text("Limit: \(limit)%")` and the accessibility label are English-only inside a section that routes everything else through `ChargeControlStrings`.

**Not verified:** the diff is truncated at 80 KB, so packaging, `build.sh`, `SMCClient.writeBytes` and anything after `Tests/MetricsTests.swift` I have not read. No hardware here either — every SMC and readback claim in the body is yours, I only checked the policy layer.

**Impact:** plugged-in Macs cycle charge → pause → discharge around the limit instead of settling, and any Mac above the limit force-discharges without being asked. Desktop Macs that opt into the beta get a panel section with no matching settings page.
**Merge:** needs the `desiredMode` / body mismatch resolved first — either the hysteresis band comes back, or the body and the new tests change to describe what the code actually does.

Bald-M:
@PathGao review pls

PathGao:
> @PathGao review pls  请审阅

macos 26 have this why do you want again？not recommend since have some Manufacturer's Guarantee/Warranties of Performance risk

Issue #757, opened 2026-08-22
the settings  page side bar glitches when clicking on screen capture


gaminbhoot:
### Description

when clicking on the screen capture button on the sidebar, the side bar sort of glitches, it doesnt happen to any other buttons, as far as i have tested

### Steps to Reproduce

1) Open the vorssaint settings page
2) on the side bar click on screen capture

### Screenshots / Recordings

<img width="3600" height="2338" alt="Image" src="https://github.com/user-attachments/assets/035c61db-971f-4bbf-9f53-fae86025b668" />

<img width="3600" height="2338" alt="Image" src="https://github.com/user-attachments/assets/732e9bd8-4bfb-4c33-8428-60e3d042f6e6" />

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

### Installation Method

DMG

PathGao:
Triage note, part of a batch pass over the open issues. The labels are a contributor's read, not a maintainer ruling.

Not localised yet, and the two screenshots show end states rather than the transition, so what is missing is something that captures the glitch in motion.

The one code surface that fits is `Sources/Vorssaint/UI/Settings/SettingsView.swift:63-80`, which has a macOS 27-specific `searchable(placement:)` branch — the same surface shared with #612 (instant crash on the first Settings-search keystroke, also on macOS 27), which is why the two are cross-referenced.

Two things would move this. A short screen recording rather than stills, and a check of whether it still happens when the sidebar search field has never been focused and is empty. The second one is the cheap experiment: it directly tests whether the macOS 27 `searchable` branch is involved, or whether this is a plain layout jump in the sidebar list that has nothing to do with search.

gaminbhoot:
https://github.com/user-attachments/assets/3fb0d766-4c7f-4e4c-af0d-5973ae571f8a

this is the screen recording, not sure how this will help, it is the same. just a recording of it

PathGao:
There's one check left and it takes about ten seconds. Open Settings fresh and click Screen Capture without ever touching the search box at the top of the sidebar — don't click into it, don't type in it.

If it still glitches that way, the search field has nothing to do with it and this is just the sidebar list jumping. If it stops glitching, that narrows it down a lot.

gaminbhoot:
https://github.com/user-attachments/assets/cef5286a-2849-4afa-bb7f-274cfcc38758

[...]

PR #958, opened 2026-08-25
fix(media): scroll the media page in Settings

tenbux:
## Summary

The Media page does not scroll in Settings, so its action button and status
line can sit below the window's bottom edge with no way to reach them. The
settings detail slot sizes each page to the window height and clips what does
not fit, and this page was the only one reachable from the sidebar with no
scrolling container of its own: every other page is a `Form`, a `List`, or
already has a `ScrollView`.

At the window's 528pt minimum height the image tool overflows the viewport
whether or not "More options" is expanded, so the button below it is
unreachable in both states. The other three tools fit at that height and were
unaffected: with the viewport ending at y=737, the last element sits at y=630
for video, y=641 for GIF and y=493 for text, against y=787 for images.

The popover already scrolled, since `MediaWorkspaceView` wrapped its content
in a `ScrollView` only when `compact` was true. This gives both hosts the
same `ScrollView` and lets `compact` decide only the height cap: 430pt in the
popover, unbounded in Settings, where the window supplies the bound. The view
the popover renders is unchanged.

`ReleaseNotesSettings` uses the same shape already, a fixed header above a
`ScrollView`.

## Verification

```sh
swift build                    # Build complete! (17.54s)
./build.sh --test              # TESTS OK (8541 checks)
./build.sh                     # ✓ Bundle ready, no warnings
```

Ran a `--dev --install` build and checked the Media page in Settings with the
window at its 528pt minimum, both by eye and by reading element geometry. The
viewport spans y=314..737.

| state | scrolled to top | scrolled to bottom |
|---|---|---|
| More options collapsed | last element y=787 | y=714, inside the viewport |
| More options expanded | last element y=1039 | y=714, inside the viewport |

So the action row is below the clip in both states and reachable by scrolling
in both. Content stays pinned to the top (first element y=324 against a
viewport top of 314) rather than stretching or centering, a short page (text
extractor, 7 elements) still fits with no scroll bar shown, and the window
does not resize when switching between Media, Kill process and Features.

The menu bar popover was checked by eye and is unchanged, which matches the
code: at `compact == true` the view emitted is the same `ScrollView` with the
same 430pt cap as before.

Checked on Apple Silicon, a single Retina display, in English. Not tried on a
non-Retina or multi-display setup, or in another app language.

## Anything else

The Settings page now carries a `ScrollView` even when its content fits. No
scroll bar appears in that case and the content stays top-aligned, which is
what the short-page check above was for, but it is a container that was not
there before.

No open issue covers this, so there is nothing to reference.

### DIFF
--- a/Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
+++ b/Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
@@ -152,15 +152,14 @@ struct MediaWorkspaceView: View {
         VStack(alignment: .leading, spacing: compact ? 10 : 14) {
             header
             toolPicker
-            if compact {
-                ScrollView {
-                    content
-                        .padding(.trailing, 1)
-                }
-                .frame(maxHeight: 430)
-            } else {
+            // The settings detail slot clips its page to the window height,
+            // so without a scroll view the action row is unreachable. Only
+            // the popover needs a cap.
+            ScrollView {
                 content
+                    .padding(.trailing, 1)
             }
+            .frame(maxHeight: compact ? 430 : CGFloat.infinity)
         }
         .onChange(of: currentImageOptions) { oldOptions, newOptions in
             guard selectedTool == .imageCompressor else { return }


PathGao:
Read the diff, nothing at tier 2 or above. The `compact == true` branch emits the same `ScrollView` with the same 430pt cap it did before, so the "popover unchanged" claim holds in the code.

One change the body doesn't call out: the Settings path now also gets `.padding(.trailing, 1)`, which it didn't have before. 1pt, and it's the price of both hosts sharing one view, so I'm noting it rather than objecting.

What I couldn't check: the claim that every other sidebar page is a `Form`, a `List` or already scrolls, and the settings detail slot's clipping behaviour — neither file is in the diff, so I'm taking both on your word plus the geometry numbers. Same for the popover, which only your by-eye check covers.

I scanned the open issues and agree nothing covers this. #757 is the nearest and it's a sidebar glitch, not this.

`Swift 6.0.3 compatibility` hasn't reported yet at the time of writing. Build & selftest is green.

**Impact:** none outside the Settings Media page. The popover renders the identical view.
**Merge:** ship it.

PathGao:
This change is already on main, which is what the branch conflicts with.

`Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift:154-158` on main — the `ScrollView` is already unconditional, wrapping `content` with `.padding(.trailing, 1)` and capped by `.frame(maxHeight: compact ? 430 : .infinity)`. That is functionally the same code this diff produces; the only differences are the added comment and `CGFloat.infinity` vs `.infinity`.

It landed in a63177cc, "fix(media): stabilize More options layout and profile dimensions" (#842), with the release note at `CHANGELOG.md:34`. The branch is 31 commits behind that commit, and this hunk is its only change.

Nothing here is left to merge — rebasing would produce an empty diff. Whether to close it is the maintainer's call; the one thing worth checking first is whether there is a case the main version still misses.

tenbux:
Closing: #842 landed the same fix. `MediaWorkspaceView` on `main` already wraps `content` in an unconditional `ScrollView` capped with `.frame(maxHeight: compact ? 430 : .infinity)`, which is what this changed the `if compact` branch into.

What is left here is a three-line comment explaining why the non-compact case needs the scroll view, and `CGFloat.infinity` where `main` writes `.infinity`. Not worth a PR on its own, and the second is the same value.

tenbux:
Reopening: I closed this a moment ago as a duplicate, which was my call to make and it was not clear cut. Correcting that, and leaving the decision to you.

What is duplicated: #842 changed `MediaWorkspaceView` to wrap `content` in an unconditional `ScrollView` with `.frame(maxHeight: compact ? 430 : .infinity)`, which is the behavior this branch was written to produce. The scrolling fix itself is already on `main`.

What is not: this still adds a three-line comment saying why the non-compact case needs the scroll view at all, since the settings detail slot clips its page to the window height. It also writes `CGFloat.infinity` where `main` writes `.infinity`, which is the same value and not worth changing.

So the only open question is whether that comment is worth keeping. Happy to reduce this to the comment alone, or to close it, whichever you prefer.

PathGao:
Keep the comment, drop the rest.

It's the only thing on this branch that `main` doesn't already have, and it's what stops the next person from putting `if compact` back — nothing in `MediaWorkspaceView` shows why the non-compact case needs a scroll view. `CGFloat.infinity` should go; `main`'s `.infinity` is the same value, so matching it is one less line of diff.

So: rebase onto `main` with the hunk reduced to the three comment lines. One thing to settle before that comment becomes the documentation for this — I've never seen the settings detail slot's clipping code, so "clips its page to the window height" rests on your measurement, not on anything I verified.

I dropped this to `priority: low`. With the behavior already on `main`, what's left is a comment.

**Impact:** none. The remaining change is a comment.
**Merge:** needs the rebase down to the comment lines first.

PathGao:
Closing — you had this right the first time. #842 landed the scrolling fix, and `MediaWorkspaceView` on `main` already wraps `content` in an unconditional `ScrollView` capped at `compact ? 430 : .infinity`.

The three-line comment is still worth having. It is a one-hunk change on top of `main` rather than a branch, so open it as its own PR whenever you like.

Issue #798, opened 2026-08-23
Add option to install apps into user applications directory


christefano:
### Problem or Motivation

Voirssaint offers to install apps from disk images into _/Applications_, which I go out of my way to avoid for home folder portability and multi-user purposes.

### Proposed Solution

Would it be possible to add the option to install apps from disk images into _~/Applications_ instead of _/Applications_?

### Feature Area

Other

PathGao:
Triage note, part of a batch pass over the open issues. The labels here were applied by a contributor, not a maintainer ruling.

Two things about this one, and they belong together.

**The title is what is missing, not the body.** `[Feature]:` has nothing after it, while the body states the request precisely: install disk-image apps into `~/Applications` instead of `/Applications`. Please don't read the empty title as an empty report — add a title and this is complete.

**It is also the cheapest item in this batch.** `DiskImageInstallerSupport.destinationURL(for:applicationsURL:)` (`Sources/Vorssaint/Services/DiskImageInstaller/DiskImageInstallerSupport.swift:33`) already takes the destination directory as a parameter, and there is exactly one call site in the repo: `DiskImageInstallerService.swift:122-124`, which hard-codes `URL(fileURLWithPath: "/Applications", isDirectory: true)`. The change is a Defaults key plus a Settings control feeding that one argument — no support-layer signature change, no second call site to keep in sync. The "already installed" guard at `:125` reads the returned URL, so it follows the chosen directory for free.

JashRashne:
Hi, I'd like to pick this up.

I'm thinking of keeping the existing /Applications behavior as the default and adding an opt-in setting to install apps from disk images into ~/Applications instead.

I'll first trace the current installer/defaults/settings flow and keep the change scoped to the existing destination handling rather than introducing a second installer path.

One thing I'll also verify is the expected behavior when ~/Applications does not exist yet.

Let me know if you'd prefer a different UX for choosing the destination.

PathGao:
Default `/Applications` with an opt-in is the right shape. The UX call is the author's, not mine — I'd wait for a maintainer before building a picker.

On `~/Applications` not existing: don't check-and-create it yourself. Foundation already does that:

```swift
let destination = try FileManager.default.url(
    for: .applicationDirectory,
    in: userScope ? .userDomainMask : .localDomainMask,
    create: true
)
```

[...]

PR #963, opened 2026-08-25
feat(settings): add user applications install option

JashRashne:
## Summary

Adds an optional, remembered choice in the disk image install prompt to install apps into the Applications folder inside the current user’s home folder.

- Installation into `/Applications` remains the default.
- The home-folder choice appears alongside the existing Trash download and Show app choices, with no additional control in Features settings.
- The prompt updates when the destination choice changes, and completion messages identify the folder used.
- Show app reveals the actual installed copy in Finder.
- Mount inspection does not create the home Applications folder. Installation creates it only after the user confirms.
- Default collision checks remain scoped to `/Applications`; the home-folder option checks both destinations. Required directory-resolution failures prevent installation.

The change extends the existing installer and remembered-choice mechanism. It adds no runtime dependency or background process. New user-facing text is included in all 13 supported languages.

## Verification

Tested on an Apple silicon MacBook Air using the locally built app and the English app language.

Automated verification:

- `./build.sh --test` — `TESTS OK (52978 checks)`
- `./build.sh` — completed without compiler warnings
- `git diff --check` — clean
- Regression coverage checks destination selection, collision scope, unavailable directory results, and missing-folder detection without creating directories.

Manual verification using the official Rectangle disk image:

- Installed into `/Applications` with the home-folder option off.
- Installed into `~/Applications` with the option on.
- Confirmed that prompt and completion wording identify the selected destination.
- Confirmed that Finder selects the copy installed at that destination.
- Confirmed that the home-folder choice remains selected after quitting and reopening the app.
- Confirmed that cancellation leaves the app uninstalled.
- Temporarily renamed the existing home Applications folder and verified that the prompt still appears with the remembered home-folder option on.
- Confirmed that cancellation leaves the folder absent, while installation creates it and reveals the installed app.
- Removed the test installation and restored the original folder afterward.

The source DMG was retained during these tests. I did not visually inspect every translation or test on Intel hardware or a separate macOS user account.

## Anything else

Refs #798

### DIFF
--- a/Sources/Vorssaint/Core/DiskImageInstallerStrings.swift
+++ b/Sources/Vorssaint/Core/DiskImageInstallerStrings.swift
@@ -6,6 +6,7 @@ import Foundation
 struct DiskImageInstallerStrings {
     let title: String
     let hubDescription: String
+    let useUserApplications: String
     let promptTitle: String
     let promptBodyFormat: String
     let installButton: String
@@ -43,6 +44,7 @@ extension DiskImageInstallerStrings {
     static let enUS = DiskImageInstallerStrings(
         title: "Disk image installer",
         hubDescription: "Install the single app inside a disk image and clean up the download",
+        useUserApplications: "Install apps in ~/Applications",
         promptTitle: "Install this app?",
         promptBodyFormat: "%@ will be copied to Applications. The disk image is then ejected and its download moved to Trash.",
         installButton: "Install",
@@ -59,6 +61,7 @@ extension DiskImageInstallerStrings {
     static let ptBR = DiskImageInstallerStrings(
         title: "Instalador de imagens de disco",
         hubDescription: "Instale o único app de uma imagem de disco e limpe o download",
+        useUserApplications: "Instalar apps em ~/Applications",
         promptTitle: "Instalar este app?",
         promptBodyFormat: "%@ será copiado para Aplicativos. Depois, a imagem de disco será ejetada e o download irá para o Lixo.",
         installButton: "Instalar",
@@ -75,6 +78,7 @@ extension DiskImageInstallerStrings {
     static let tr = DiskImageInstallerStrings(
         title: "Disk görüntüsü yükleyicisi",
         hubDescription: "Disk görüntüsündeki tek uygulamayı yükle ve indirilen dosyayı temizle",
+        useUserApplications: "Uygulamaları ~/Applications klasörüne yükle",
         promptTitle: "Bu uygulama yüklensin mi?",
         promptBodyFormat: "%@ Uygulamalar'a kopyalanacak. Ardından disk görüntüsü çıkarılacak ve indirilen dosya Çöp Sepeti'ne taşınacak.",
         installButton: "Yükle",
@@ -91,6 +95,7 @@ extension DiskImageInstallerStrings {
     static let ru = DiskImageInstallerStrings(
         title: "Установка из образа диска",
         hubDescription: "Установите единственное приложение из образа диска и удалите загрузку",
+        useUserApplications: "Устанавливать приложения в ~/Applications",
         promptTitle: "Установить это приложение?",
         promptBodyFormat: "%@ будет скопировано в папку «Программы». Затем образ диска будет извлечён, а загрузка перемещена в Корзину.",
         installButton: "Установить",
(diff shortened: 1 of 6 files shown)


PathGao:
#798

Three things.

**1. The already-installed check widened for everyone, not just opt-in users.** Body says "The existing `/Applications` behavior remains the default", but `candidate()` now returns `nil` when the app exists in *either* directory. Setting off, `Foo.app` already in `~/Applications` (non-admin account, `brew --appdir`, an app that self-installs there): mount `Foo.dmg` and previously you got the install prompt, now nothing happens. No prompt, no alert, nothing the user sees — which is indistinguishable from the installer being broken. Either state that in the body and release notes, or make the skip visible.

**2. `create: true` in the detect path creates `~/Applications` on mount.** With the setting on, mounting any disk image resolves the destination with `create: true` before the user has agreed to install anything. Cancel the prompt and the directory is still there. `collisionURLs` already uses `create: false`; the destination only has to exist immediately before `moveItem`. Resolving it in `install()` rather than `candidate()` fixes that and also drops the stale-destination case when the setting is toggled while the image stays mounted. I can't see the rest of `DiskImageInstallerService.swift`, so I don't know what else reads `candidate.destinationURL`.

**3. `allSatisfy` on an empty array is `true`.** If both domain resolutions fail, `collisionURLs` returns `[]` and both guards pass vacuously. The old code compared the destination path directly and couldn't do that. Cheapest fix is to bail when the list doesn't contain `candidate.destinationURL`.

Not flagging: `promptBodyFormat` still says "copied to Applications" in all 13 locales while the destination may be `~/Applications`. Ambiguous enough to read correctly either way, your call.

Didn't verify: the toggle's visibility rules — only `FeatureHubRow` is in the diff, not whatever decides `installed` — and none of this on hardware.

**Impact:** users with the setting off who already have the app in `~/Applications` lose the install prompt with no feedback; `~/Applications` gets created by merely mounting a disk image.
**Merge:** needs #1 resolved first — either declare the default-path change or scope the check to the opt-in case.

Sources/Vorssaint/UI/Settings/FeatureHubSettings.swift
```diff
@@ -241,7 +241,20 @@ private struct FeatureHubRow: View {
         }
     }
 
+    @ViewBuilder
     var body: some View {
+        if installed, feature == .diskImageInstaller {
```
vorssaint: the features list has no per feature controls, and a small toggle tucked under this row looks bolted on. the install prompt already remembers two choices, moving the download to trash and showing the app, so this belongs there as a third remembered choice with no settings control

Sources/Vorssaint/Core/DiskImageInstallerStrings.swift
```diff
@@ -43,6 +44,7 @@ extension DiskImageInstallerStrings {
     static let enUS = DiskImageInstallerStrings(
         title: "Disk image installer",
         hubDescription: "Install the single app inside a disk image and clean up the download",
+        useUserApplications: "Install apps in ~/Applications",
```
vorssaint: this reads as a path. can it say it in plain words, like installing in the applications folder inside your home folder, and can the prompt and the message after installing say which folder the app went to when the option is on

Sources/Vorssaint/Services/DiskImageInstaller/DiskImageInstallerService.swift
```diff
@@ -18,7 +18,6 @@ final class DiskImageInstallerService {
         let appURL: URL
         let imageURL: URL
         let imageIdentity: FileIdentity
-        let destinationURL: URL
```
vorssaint, line 22: the install prompt now also offers to show the app after installing, and it looks for the app at this destination. with the option on that has to open the copy in the home folder, so it needs the destination the install actually used

Sources/Vorssaint/Services/DiskImageInstaller/DiskImageInstallerService.swift
```diff
@@ -191,43 +216,72 @@ final class DiskImageInstallerService {
...
+                                      useUserApplications: Bool,
+                                      fileManager fm: FileManager) -> [URL]? {
+        let domains = DiskImageInstallerSupport.collisionDomains(
+            useUserApplications: useUserApplications)
+        var applicationsURLs: [URL] = []
+        for domain in domains {
+            guard let applicationsURL = try? fm.url(for: .applicationDirectory,
+                                                    in: domain,
+                                                    appropriateFor: nil,
+                                                    create: false) else { return nil }
```
vorssaint: with the option on and no applications folder in the home folder yet, does the install prompt still show up after mounting an image

JashRashne:
Thanks for reviewing. I’ll address all four points: move the option into the install prompt as a remembered choice, clarify the destination wording in the prompt and completion messages, ensure “Show app” uses the actual installation destination, and verify the missing-folder case. I’ll update the PR once these are addressed.

JashRashne:
Addressed all four points in 6dd31fed and updated the PR description with verification details. The destination choice is now remembered in the install prompt, the wording identifies the selected folder, and Finder reveals the actual installed copy. I also manually verified the missing-folder case, including cancellation and folder creation after confirming installation. Thanks for the review!

Issue #956, opened 2026-08-25
Window switcher customization - let users actually enjoy it


MattOrmianek:
### Problem or Motivation

The idea of using one software package instead of ten smaller apps appeals to me. However, today's implementation of the app switcher discourages users from using it.

Also note: 
There is #838, which summarises information about AltTab. However, there is no mention of AltTab being bloatware, especially when the user activates its functions every few seconds. AltTab takes screenshots and spams the Windows Server process constantly.

### Proposed Solution

In #735 , I added a few comments about how users could enjoy the app switcher more besides the speed of the animation. I will copy the comments from that issue here.

It would also be great to have the option of not showing the windows of minimised apps, and showing just the app name rather than the app name and window names. There should also be an option to show the app switcher only on the monitor that is in focus, with apps/windows only from that monitor.

This is a super nice-looking proposition:

<img width="2598" height="589" alt="Image" src="https://github.com/user-attachments/assets/3e2469ef-c439-4765-8827-6d14d6b89798" />

The best I can set up right now is:

<img width="2822" height="798" alt="Image" src="https://github.com/user-attachments/assets/74df97da-cb9a-4998-ae71-9e91c865389c" />

### Feature Area

Windows and Dock

### Does another app already do this?

AltTab

### Mockups / References

<img width="2598" height="589" alt="Image" src="https://github.com/user-attachments/assets/d290f94a-8d89-4b98-b2be-135c92828b9e" />

PathGao:
Only part of this is done, so it is now labelled `half` rather than fixed. The report stays open and you will not be asked to confirm the rest as fixed.

**Done:** you can now choose which display the switcher opens on, including the one holding the active window.

**Not done:** limiting it to the windows on that display. The switcher still lists windows from every display, and there is no setting for it yet. That part has not been built.

MattOrmianek:
Great! I've just checked it and it looks fantastic. One small change and I'll be totally satisfied, I would like the option to show windows preview instead of app icons, just show me miniatures of windows instead of one preview of the active app and big app icons.

PR #972, opened 2026-08-25
feat(switcher): choose which display the App Switcher opens on

noahjstewart:
## Summary

Adds a **Show on** choice under Settings → App Switcher so the ⌘Tab panel can open on a display other than the one under the pointer:

- **Screen with the pointer** — what the app always did, still the default; nothing changes for existing users.
- **Screen with the menu bar** — the primary display in Displays settings.
- **Screen with the active window** — the display showing the window that was in front when the shortcut was pressed.

Every choice falls back to the pointer's screen when its display cannot be resolved (menu bar screen missing mid-reconfiguration; the source is an app-only entry with no frame, or a window touching no display).

**Why it belongs here.** The switcher is the feature people come to Vorssaint for from AltTab, and AltTab's "Show on" picker (active screen / screen including mouse / screen with menu bar) is the setting they expect to find. On a two-display setup where the menu bar lives on the external monitor, "wherever the pointer happens to be" is the one behaviour a keyboard-driven switcher should not have, and there is currently no way to change it. #956 asks for the active-screen variant. It stands on `NSScreen` and `CGDisplayBounds` only — no private API, no runtime dependency, no permission — and the permanent surface is one preference, one picker and five strings.

**How it is built.** `NSScreen.withMenuBar` already existed in `AppKitExtensions.swift`, so the menu bar case reuses it. The active-window case stores the source window's window-server frame in `SwitcherSourceContext` and matches it against `CGDisplayBounds` for each attached screen — same coordinate space as `kCGWindowBounds`, so no flipping — picking the display that holds the larger part of a straddling window (`SwitcherSupport.displayIndex(showingMostOf:displayBounds:)`, pure and unit-tested). The three places that hardcoded the pointer screen (`recomputeLayouts`, `updateIconRowLayoutForCurrentSelection`, `centeredFrame`) now go through one `placementScreen`.

Both new strings and the preference travel in settings backups automatically because the key is a registered default.

## Verification

- MacBook Air 15" M3 (Mac15,12), macOS 15.6.1 (24G90), Command Line Tools 16.4 / Swift 6.1.2, no Xcode. Two displays: an external Acer VG240Y S 1920×1080 set as the main display (menu bar), and the built-in 2560×1664 Retina as secondary. Single Space per display.
- Own build of this branch (`./build.sh --dev --install`), with the settings of the 3.3.2 release imported into the dev domain.
- *Screen with the menu bar*: with the pointer on the built-in display, ⌘Tab opens the panel on the Acer, sized to that display's visible frame. *Screen with the active window*: the panel follows the display of the window that was in front. Both checked in the default card grid and in the large-icon row layout. The default choice is the unchanged code path and behaves as on `main`.
- `./build.sh --test`: TESTS OK (8560 checks), including new checks for the registered default and backup export, the stored-value parser and fallback, `displayIndex` (inside one display, straddling in either direction, no overlap, zero frame, no displays) and non-empty strings in all 13 languages.
- `./build.sh` release configuration: finishes; the only warnings are the two pre-existing Swift 6 sendability warnings in `ShelfService.swift:1455-1456` on `main`, which this branch does not touch.
- `./build/Vorssaint --selftest`: SELFTEST OK.
- **Not tested:** more than two displays, displays arranged vertically, a display disconnected mid-session, and macOS 14 / 26. The translations for the five new strings were written by me from the existing catalog's terminology (e.g. 画面 / 화면 / Menüleiste / barre des menus) and would benefit from a native speaker's glance.

## Anything else

Refs #956 — that issue asks for "show the app switcher only on the monitor that is in focus"; the *Screen with the active window* choice covers the display-placement half of it. The "only windows from that monitor" half is a separate filtering change and is not part of this PR.

Per docs/AI-CONTRIBUTIONS.md: this was written with an agent (Claude Code). Before writing, open PRs and issues were searched (nothing implements this; #956 is the closest), the existing `NSScreen` helpers and per-app rules were reused rather than re-implemented, and the whole panel-placement path was read before editing. What was skipped: opening an issue to settle direction before the branch existed — if the three-option shape is wrong for the project, happy to narrow it to the menu bar choice alone.


PathGao:
#956

`displayIndex` itself reads correct: ties go to the first display, null/empty frames yield nil, and the straddle tests match the arithmetic. Two things about `.activeWindow` I couldn't settle from the diff.

**The coordinate space of `source.frame` is not in this diff.** `activeWindowScreen` matches it against `CGDisplayBounds` and the doc comment asserts window-server top-left origin, but the producer of that `source` is outside the diff, so I can't confirm it isn't an AppKit bottom-left frame. Note that your test rig can't tell the two apart: two displays side by side share a Y range, so a flipped Y still lands on the correct display. The arrangement that would expose it is vertical stacking, which is on your not-tested list. If the frame comes from `kCGWindowBounds` this is fine as written; if it comes from an `NSWindow` frame, `.activeWindow` picks the wrong display on a stacked setup and nothing in the tests catches it.

**Stale `sessionSourceContext`.** The assignment sits inside an `if let source` branch, and `sessionStartWindowID = source?.windowID` right below implies `source` can be nil. If nothing clears the context on that path, a session with no source keeps the previous session's `frame` and `.activeWindow` opens the panel on the display of a window that is no longer in front. Before this PR a stale context only fed windowID/pid comparisons; `frame` now feeds placement behind a nil check that a stale value passes. If it is cleared at session end, I can't see that from here.

**Impact:** `.activeWindow` only. The default pointer path is the same code as `main`.
**Merge:** needs those two `.activeWindow` assumptions confirmed (frame coordinate space, stale source context) first.

noahjstewart:
Thanks — both were already the case, but you're right that neither was visible from the diff. Pushed 1 commit that puts them where you were looking, no behaviour change.

**Coordinate space of `source.frame`.** It is window-server (top-left origin) on every path, never an `NSWindow` frame:

- `WindowEnumerator.swift:270-273` builds `cgFrame` from `kCGWindowBounds`.
- `switchableFrame(cgFrame, fallback: axWindow?.frame, …)` at `:276` — the fallback is `CGRect(origin: kAXPositionAttribute, size: kAXSizeAttribute)` at `:582-595`, and the AX position attribute is also top-left screen coordinates.
- Minimized windows with neither get `CGRect(origin: .zero, size: minimumSize)` at `:609`, but `sessionSourceItem` (`SwitcherSupport.swift:429`) prefers an on-screen, non-minimized window as the source, so that case does not normally feed placement; if it did, origin zero resolves to the menu bar display, which is the fallback anyway.

The new doc comment on `SwitcherItem.frame` states this contract, and the added `displayIndex` check uses two vertically stacked displays: a frame at `y: 1300` resolves to the lower display, and the same window expressed as a bottom-left frame (`y: -800`) touches no display and yields nil — so a producer handing over AppKit coordinates would fail the fallback, not silently pick the wrong screen. Agreed that my side-by-side rig cannot distinguish the two at runtime; the test is the substitute.

**Stale `sessionSourceContext`.** The assignment at `AppSwitcher.swift:796` is `Optional.map`, not `if let`: when `source` is nil the context is set to nil in the same statement. Independently, `endSession()` sets it to nil (`:1287`), and `cancelSession()` routes through `endSession()`, so no context survives a session boundary on any path. Added a two-line comment at the assignment so the `Optional.map` intent is not mistaken for a guarded branch again.

PathGao:
Both settled.

`.map`, not `if let` — I misread that, it's in the diff unchanged and a nil source already cleared the context. Nothing to fix there.

On the coordinate space I still can't see `WindowEnumerator.swift`, so the three producer paths are on your word. What I can check is that the failure mode is benign: a flipped frame yields nil on a stacked setup and falls back rather than picking the wrong display, and side-by-side doesn't depend on Y at all. The doc comment on `SwitcherItem.frame` is what keeps that true for whoever adds the next producer.

One correction to your reply, not a blocker: for the minimized `CGRect(origin: .zero, size: minimumSize)` case, origin zero is inside the menu bar display's bounds, so `displayIndex` returns that display — but `placementScreen`'s fallback for `.activeWindow` is `NSScreen.withMouse`, not the menu bar screen. So that path doesn't land on the fallback, it lands on the menu bar display specifically. Rare enough given `sessionSourceItem` prefers an on-screen window, and the result is still a real display.

**Impact:** `.activeWindow` only, as before.
**Merge:** ship it.

ref #956

vorssaint:
LGTM

PR #978, opened 2026-08-25
refactor(capture): give every tool its own settings section

PathGao:
#757 reports the Screen capture sidebar glitching when that page opens. The immediate cause is the segmented row of tool names: a SwiftUI segmented picker answers an offer narrower than its labels with its own minimum, and since a frame does not clip, it paints outside the row — over the sidebar on one side, off the window on the other. Rendering the picker offscreen into a 300pt slot reproduces the reporter's screenshot on macOS 26; no macOS 27 needed.

Measured in English on macOS 26: the detail column at the window's 772pt minimum is 494pt and the row's floor is 481pt. Thirteen points. German spends them — `Text vom Bildschirm kopieren` takes the floor to 559, so that language is already over the edge at the minimum window today.

Rather than contain that, this removes the control.

## Why the control, not its width

Three things point the same way, and #934 sharpened the third:

- Apple's guidance for segmented controls is five to seven segments with short labels. These are feature names, and `ScreenCaptureTool.available()` filters by what is installed, so the row already varies between one and four segments and grows with every tool added later.
- The control's width follows its labels, so it is full-bleed in English and stops short of the row in Chinese. Same control, two layouts, decided by translation.
- After #934 that top section holds nothing but the switcher. The shared caption string was deleted and the general capture shortcut became the screenshot tool's own, so the section now exists only to host the control.

And this was the only page in Settings that hid its siblings behind a switcher. Clipboard, Finder shortcuts and Mouse all put their features side by side, which is why none of them needs an equivalent of `SettingsSectionAnchor.screenCaptureTool` and `reconcileSelection(withDestination:)`. All four capture anchors were wired, but three were unreachable until the switcher moved first.

## The change

The page leads with the chooser and then gives every installed tool a section of its own, in the order `ScreenCaptureTool.available()` returns. Each tool's own shortcut moves out of the shared top section and into its own section, next to the settings it belongs to.

What the four tools share is the chooser, not a shortcut — one panel, four modes, switchable after it opens — so the lead section is a button that opens it on no particular mode. That is also the only such entry point left: since #934, every other caller passes `initial:`.

Net effect on the file: **-165 / +48**. Deleted: the tool row, the page's selection state, `reconcileSelection(withDestination:)`, the anchor-to-tool mapping. The width problem goes with them, in every language and at any number of tools.

One thing deliberately left out: the lead section would read better with a caption line saying the four are modes of one panel. That is copy across thirteen tables and your wording, not mine — say the word and I will add the key.

## Considered and not taken

- **Containing the overflow** (`ScrollView(.horizontal)`, or an `NSSegmentedControl` that truncates). Both were built and measured; the first leaves the last tools off-screen until scrolled, the second truncates labels. Both keep a control that cannot hold a growing set.
- **`ViewThatFits` down to a popup menu.** It selects on natural width (640), not the floor (481), so it drops to the popup at 560 and below — the tab bar would vanish at the default window size for everyone. A popup also hides the other tools behind a click, which is worse than the bug.
- **Shorter tab labels.** Would take the floor to 372, but these strings are the app-wide feature names read by the menu panel, command bar, shortcut list, feature hub and settings search, so it means four new keys across thirteen tables and two names per tool — and it only moves the threshold.
- **Sub-pages behind a list**, the way System Settings pushes into a category. Better if the tool count keeps growing, but it needs a navigation stack in the settings detail and turns deep links from a scroll into a push. Worth revisiting past five or six tools.

## Not changed

Tool names, their order, their own settings; the colour-format picker; the window's `minWidth`; the chooser itself; no new strings.

## Verification

`./build.sh`, `--selftest` and `./build.sh --test` green, 8551 checks, including the tests #934 added. Two new checks — that the page renders every installed tool with no selection state, and that the number of tools carrying their own shortcut row equals `ScreenCaptureTool.allCases.count` — fail on `main` and pass here. Not machine-verified: the rendered result on a real Mac, and the reporter's macOS 27, where both columns are displaced by roughly 80pt rather than the few points macOS 26 gives.

Refs #757


PathGao:
Read the diff. Three things.

**`ScreenshotCaptureSettings` and `ScreenRecordingCaptureSettings` are internal; the other two sub-views are private.** The shortcut row now sits inside all four, so any other page embedding one of the internal two grows a shortcut recorder it didn't have. Their other call sites aren't in the diff — if `grep -rn 'ScreenshotCaptureSettings()\|ScreenRecordingCaptureSettings()' Sources` returns anything besides `ScreenCaptureSettings.swift`, that's the same recorder in two Settings destinations. If it returns nothing, the asymmetry is just file layout and there's nothing here.

**The new count check passes with a tool missing.** `shortcutRows` totals occurrences across three files and compares to `allCases.count` — it never checks *which* tools. Delete `.text`'s row and duplicate `.color`'s and it still totals 4 and passes, while the message reads "every capture tool carries its own shortcut in its own section". Match the set instead:

```swift
let tools = Set(captureSettingsSources.joined(separator: "\n")
    .components(separatedBy: "ToolShortcutRows(tool: .").dropFirst()
    .map { String($0.prefix { $0.isLetter }) })
expect(tools == Set(ScreenCaptureTool.allCases.map { "\($0)" }),
       "every capture tool carries its own shortcut in its own section")
```

Same length, and it also fails when a fifth tool arrives with no row rather than waiting for someone to notice the count.

**The anchor claim is what justifies the deletion, and it's the one thing nothing here checks.** "All four capture anchors were wired, but three were unreachable until the switcher moved first" is the reason `reconcileSelection(withDestination:)`, the `SettingsSectionAnchor` extension and the `router` observation go away. Nothing in the diff attaches an anchor to the four sections, and neither new test opens a deep link — both assert layout. I'm not claiming it's broken; the wiring is outside what I can see. But a third check that lands on `.colorPicker` and asserts the section is on screen would cover the actual bet.

Minor: the lead section renders `strings.screenCaptureTitle` as both the header and the button label, so the string stacks twice. The old picker carried `.labelsHidden()` for that.

Not verified: the sub-views' other call sites, the anchor attachment, and anything on real hardware.

**Impact:** none if the two internal sub-views are only used on this page; a duplicated shortcut recorder on a second Settings destination if they aren't. The test gap costs nothing today.
**Merge:** ship it.

ref #757

Issue #976, opened 2026-08-25
Turning on "one row per app" breaks the ⌘` switcher


AB-boi:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When simple app switcher and show one entry per app are both enabled, the "cmd + `" switcher UI gets cut off: 

<img width="250" alt="Image" src="https://github.com/user-attachments/assets/681715a5-05ea-4010-831a-24d0bd2d86d2" />

Issue has been split from #768 and separately reported as a bug here.

### Steps to Reproduce

In the app switcher feature, enable both simple app switcher and show one entry per app.
Use "Cmd + `" to invoke the relevant UI.

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.2 (stable)

### macOS Version

macOS 15 (Sequoia)

### Language

English

### Mac and displays

Macbook Air M4, built-in screen

### Installation Method

Homebrew

AB-boi:
### Bug tech note：(by @PathGao )
usesWindowRow (SwitcherView.swift:212-215, SwitcherSupport.swift:346-348) only looks at simpleMode and mergeWindowsByApp and ignores sessionScope, so the per-window list for .frontmostApp gets collapsed into a single tile by appGroups at SwitcherView.swift:534-535

iltonandrew:
Taking a look at this one, following the tech note above.

AB-boi:
Fully fixed now!

Issue #768, opened 2026-08-22
Weird/Confusing interactions between switcher options


AB-boi:
### The Problem
So, the switcher rn has predominantly 2 formats: The app bar format and the window pane format. The current options that have been provided to set it up are a little weird and confusing. As far as the appearance of the switcher is concerned, there are primarily 3 options that control it:

1. Simple app switcher
2. Show command + Tab with large icons
3. Show one entry per app

(The below options are just additional modifications and they don't add to the confusion and so can be ignored for now:
Show shortcut hints
Show only the current desktop
Apps with no open window)

Here are 2 examples of weird setups possible:

1. With Simple app switcher enabled:
a. **Show one entry per app disabled:** gives you an app bar with each window shown as an app icon. <img width="1024" alt="Image" src="https://github.com/user-attachments/assets/407175da-0c31-469c-a9aa-7ffed3eebe26" /> 
b. **Show one entry per app enabled:** give you an app bar with a preview pane above it, where only one window is displayed. I think this might be unnecessary and bit misleading because the description here doesn't tell you that one entry also means one window. Not very different from the native cmd + tab menu; only additionally allows closing app windows, and knowing which app could have windows in other desktops. Here, enabling show one entry should have just left 1 entry for each app in the app bar, no need for a preview pane right? Preview pane is suggestive of more windows being available and wastes a lot of whitespace. <img width="1024" alt="Image" src="https://github.com/user-attachments/assets/032d6d57-276f-4150-82b1-34dbbbe1ff0c" />
Also enabling show one entry per app, breaks the "cmd + `" switcher: <img width="100" alt="Image" src="https://github.com/user-attachments/assets/681715a5-05ea-4010-831a-24d0bd2d86d2" />

2. With Show command + Tab with large icons enabled, is there a need for show one entry per app?

**Also** there is currently no combination of options that can make me have: an app bar with windows showing up above it in a preview pane, except not actually rendering a window preview and instead have either the app icon itself be the thumbnail OR no thumbnail at all as seen in option `1.b` above. In the latter, since each window option is short in height we could have vertically arranged window options (navigable by `↑` `↓` keys) or just horizontal arranged in multiple rows if there are too many windows.

### My Solution

To reduce setup related confusion and retain customizability, I suggest rearranging the options in the switcher menu in the following way:
Pick first between 2 switcher looks:
1. (App) Icon Bar
2. (App) Window Pane

Additional modifications for Icon Bar:
[...]

PR #979, opened 2026-08-25
fix(switcher): keep the window-scoped switcher per-window when one-entry-per-app is on

iltonandrew:
Refs #976.

## What was wrong

`SwitcherSupport.usesWindowRow` decided the simple switcher's row shape from `simpleMode` and `mergeWindowsByApp` alone and ignored the session scope — exactly as the tech note on the issue says. A ⌘` session lists windows of one app, so with one-entry-per-app enabled, `appGroups` collapsed the whole list into a single tile and the panel rendered cut off around it.

## The change

`usesWindowRow` now takes the session scope, and a `.frontmostApp` session always keeps its per-window row; all-apps behavior is unchanged. `AppSwitcher` exposes the scope read-only and `SwitcherView` passes it into the decision.

One ordering fix rode along because the scope-aware decision exposed it: the session-start layout pass (`recomputeLayouts`) runs before the session's scope was assigned, and teardown resets the scope to `.allApps` — so a window-scoped panel would have been sized for the grouped layout on its first frame. The scope is now assigned before the layout pass.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8542 checks), `./build/Vorssaint --selftest` → SELFTEST OK. The helper test now covers the scope-aware decision (window-scoped + merge stays per-window, all-apps + merge still groups), and a source-shape test pins the scope-before-layout ordering in `beginPendingSession`.

Also reproduced live on a dev build: with both settings on and a two-window app frontmost, ⌘` on current main shows one collapsed app tile with a count badge (the issue's screenshot); with this change it lists each window as its own tile.

Ref #768


PathGao:
Read the diff. The scope-aware decision looks right, and the new parameter has no default so the compiler covered both call sites. Matches the tech note on #976.

Three things.

**1. The ordering test doesn't pin what the body claims.** Body: "a source-shape test pins the scope-before-layout ordering in `beginPendingSession`". The test searches the whole file for the first `sessionScope = pending.scope` and the first `recomputeLayouts(for: list)`. Neither is scoped to that function, and nothing checks they land in the same one. It's green today because today's first occurrences happen to be the right pair. Add a `recomputeLayouts(for: list)` call anywhere earlier in the file and you get a red build complaining about an ordering that is actually fine. Slice the function out first, then search inside it:

```swift
let fn = switcherCode.range(of: "func beginPendingSession")
    .map { String(switcherCode[$0.lowerBound...]) } ?? ""
```

**2. `sessionScope` is `private(set) var`, not `@Published`.** `SwitcherView.usesWindowRow` now derives the row shape from it, but writing it fires no `objectWillChange`. It works because you assign it before the `@Published` writes further down `beginPendingSession`, so the first render already reads the right value. That is the same implicit ordering dependency you just fixed one layer down, except now SwiftUI's invalidation is what depends on it. `@Published private(set) var sessionScope` costs one word and removes the dependency. I could not check whether the extra publish is harmless, since the rest of the view's observation isn't in the diff.

**3. Teardown resetting the scope to `.allApps` is the mirror of the bug.** At session start the assignment was too late. At teardown it may be too early, if anything reads `usesWindowRow` between the reset and the panel actually leaving the screen. Teardown isn't in the diff, so treat this as a question rather than a finding.

**Impact:** none on shipped behavior. 1 is a future CI failure, 2 and 3 are latent ordering dependencies.
**Merge:** ship it.

iltonandrew:
All three addressed.

**1. The ordering test pinned a coincidence.** You're right — it searched the whole file for the first `sessionScope = pending.scope` and the first `recomputeLayouts(for: list)`, so it was green only because today's first occurrences happen to be the right pair. It now slices `beginPendingSession` out (function start through the next declaration), asserts the slice is non-empty so a rename fails loudly instead of passing vacuously, and searches inside it.

Verified both directions: a decoy `recomputeLayouts(for: list)` added earlier in the file leaves it **green**, and swapping the real two lines inside `beginPendingSession` turns it **red** on exactly that assertion. Both mutations reverted.

**2. `sessionScope` is now `@Published private(set)`.** Same reasoning you gave — the view derives the row shape from it, so relying on the later published writes to carry the invalidation is the same implicit ordering dependency one layer up. Checked before changing it: the only two write sites are session start and teardown, both on the main-thread session lifecycle, and publishing reaches no code that rewrites the scope or rebuilds the panel, so there's no loop.

**3. Teardown is safe, and here's why.** In `endSession`, `panel?.orderOut(nil)` runs synchronously at `AppSwitcher.swift:1273`, before the scope resets to `.allApps` at `:1290`. There's no animation or visible interval between them, so nothing can read `usesWindowRow` while the panel is still on screen with the reset scope. Left as-is deliberately rather than reordering for symmetry.

Gate: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8543 checks), `--selftest` → SELFTEST OK.

PathGao:
The slice fixes the start marker, not the end one.

`beginSessionBody` terminates on the literal `"\n    private func discardPendingSessionStart"`. Rename that function, move it, or land another declaration between the two, and `components(separatedBy:)` finds no separator, returns a single element, and `.first` is the whole rest of the file. The guard silently degrades back to a whole-file search — the exact shape you just removed — and the non-empty assert can't see it, because that failure makes the slice bigger, not empty. Your decoy probed above `beginPendingSession`; nothing probed the tail.

Cheapest fix is to stop naming the neighbour and assert the slice actually terminated:

```swift
let parts = (switcherSource.components(separatedBy: "private func beginPendingSession").last ?? "")
    .components(separatedBy: "\n    private func ")
expect(parts.count > 1, "the App Switcher ordering guard finds the end of beginPendingSession")
let switcherCode = parts[0] // then strip comments as you do now
```

Still assumes the next declaration is a `private func` at that indentation, but it survives a rename, which is the case that actually happens.

I can't tell you what a widened slice would match, since only the test file and the two call sites are in the diff — whether it goes false-green depends on what else in `AppSwitcher.swift` contains those two strings, which I can't see. Not a blocker either way.

3 is a satisfying answer — synchronous `orderOut` before the reset with nothing between them is the property that matters, and leaving it asymmetric is the right call.

**Impact:** none on shipped behavior. A future false-green in the ordering guard.
**Merge:** ship it.

iltonandrew:
You're right — I fixed the start marker and left the end one just as brittle. Renaming `discardPendingSessionStart` would find no separator, `.first` would return the whole rest of the file, and the guard would silently degrade to exactly the whole-file search it replaced. And as you say, the non-empty assert can't catch it, because that failure makes the slice *bigger*.

Fixed as you suggested: the slice now ends on whatever declaration comes next rather than naming the neighbour, plus an assertion that it actually terminated.

```swift
let beginSessionParts = (switcherSource.components(separatedBy: "private func beginPendingSession")
    .last ?? "").components(separatedBy: "\n    private func ")
let beginSessionBody = beginSessionParts.first ?? ""
expect(beginSessionParts.count > 1,
       "the App Switcher ordering guard finds the end of beginPendingSession")
```

Probed the tail this time, with the mutation you were actually worried about — renamed `discardPendingSessionStart` **and** swapped the real two lines inside `beginPendingSession`:

```
TESTS FAILED (1 of 8544):
  - the App Switcher session scope is assigned before the session-start layout pass
```

So the slice stays bounded across a rename and still catches a genuine swap; before this it would have gone false-green. Reverted, suite back to TESTS OK (8544 checks), build zero warnings, SELFTEST OK.

Your caveat is worth keeping on the record: it still assumes the next declaration is a `private func` at that indentation. If `beginPendingSession` ever became the last private method in the type, `parts.count > 1` would fail rather than silently widen — which is the failure direction I want, so I left it there.

Issue #971, opened 2026-08-25
Whole app freezes solid after a few hours and needs kill -9. The dispatch pool fills with stranded `waitUntilExit` calls, then Dock Preview deadlocks the main thread on it


SudhanshuBhogal:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Every few hours Vorssaint just dies on me. Menu bar icon stops responding, no shortcut does anything, no panel opens. `pkill -9 -x Vorssaint` and relaunch is the only way out. I finally caught it live this time and sampled the process before killing it, and I think I have the whole picture.

Two separate things are going wrong, and it takes both of them to produce the freeze. `BoundedProcessRunner` leaks dispatch worker threads until the global pool has nothing left, and `WindowEnumerator` blocks the main thread on that same pool with no timeout. The leak alone would be invisible. The blocking wait alone would be harmless. Together they kill the app.

What the sample showed, after 3h35m of uptime:

```
Dispatch Thread Soft Limit: 64 reached in 3643 of 3648 samples
  -- too many dispatch threads blocked in synchronous operations
```

The main thread never moved for the entire sample:

```
_dispatch_main_queue_drain
  closure #1 in DockPreviewService.scheduleHover(_:delay:)
    DockPreviewService.beginHoverIfStillValid(token:initialHit:)
      DockPreviewService.beginSession(_:)
        static WindowEnumerator.listWindows(for:maximumCount:)
          static WindowEnumerator.listWindows(filterPID:maximumCount:...)
            -[NSOperationQueue waitUntilAllOperationsAreFinished]
              -[NSOperation waitUntilFinished]
                _pthread_cond_wait
                  __psynch_cvwait
```

And there were exactly 64 threads sitting on `com.apple.root.utility-qos`, every one of them with the same stack:

```
_dispatch_call_block_and_release
  closure #2 in static BoundedProcessRunner.run(_:_:timeout:maxOutputBytes:)
    -[NSConcreteTask waitUntilExit]
      _CFRunLoopRunSpecificWithOptions
        __CFRunLoopServiceMachPort
          mach_msg2_trap
```

[...]

PR #980, opened 2026-08-25
fix(monitor): stop stranded subprocess waits from starving the dispatch pool and freezing the app

iltonandrew:
Refs #971 (candidate fix — the freeze needs hours of uptime to reproduce; the mechanism below is verified with a runtime harness).

## What was wrong

Two halves, each harmless alone:

1. `BoundedProcessRunner.run` parked one global-pool worker in `Process.waitUntilExit()` per child. Its caller-side timeout kills the child but abandons the worker, so any wait that never returns — the reporter sampled exactly that, `waitUntilExit` still parked after its child was dead and reaped — strands the thread forever. With the menu-bar Network metric's `nettop` fallback latched at a 2s interval (~1800 spawns/hour), stranded workers accumulate until the dispatch pool's 64-thread soft limit is reached. The reporter's sample shows `Dispatch Thread Soft Limit: 64 reached in 3643 of 3648 samples`.
2. `WindowEnumerator` fanned AX queries out on an `OperationQueue` and blocked the caller in `waitUntilAllOperationsAreFinished()`. The operations need pool threads; with the pool starved they never start, and the wait never ends. Dock Preview enumerates from the main thread (`DockPreviewService.beginSession`), so the first hover after exhaustion froze the app for good — matching the reporter's main-thread stack. The App Switcher's `beginPendingSession` also enumerates on the main thread (answering the open question in the report: it is dispatched to main from the tap route), so the switcher was a second door into the same freeze.

## The change

- Child exits are now observed through `terminationHandler`, delivered on Foundation's own reaper queue: an in-flight child costs zero pool threads, and a lost exit notification costs at most that call's timeout instead of a thread. The timeout → `terminate()` → `SIGKILL` escalation is unchanged.
- The enumerator's fan-out wait is now a `DispatchGroup.wait` with a 5s deadline — generous against the 0.2s per-app / 0.35s per-window AX messaging ceilings, so it only expires when no thread will ever run the queries. On expiry it returns the snapshots that finished instead of parking the caller.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8544 checks), `./build/Vorssaint --selftest` → SELFTEST OK. A standalone harness running 70 concurrent 1s children and measuring mach task threads mid-flight: the old runner parks 78 threads beyond the callers; this one parks none, with identical results, statuses and timeout behavior. Tests pin the mechanism (no `waitUntilExit`, handler-based exits; deadline-bounded enumeration wait) and add a concurrent-run semantics check.

The `nettop` fallback staying latched on an interface whose inbound counter never advances is a separate topic this PR deliberately leaves alone.

### DIFF
--- a/Sources/Vorssaint/Services/BoundedProcessRunner.swift
+++ b/Sources/Vorssaint/Services/BoundedProcessRunner.swift
@@ -59,18 +59,22 @@ enum BoundedProcessRunner {
             }
         }
 
+        // waitUntilExit parked on a global-pool thread has been observed never
+        // returning even after the child was dead and reaped (#971). Every lost
+        // wakeup stranded one pool worker, and frequent spawners (the nettop
+        // fallback runs ~1800 times an hour) starved the 64-thread soft limit
+        // within hours. The termination handler is delivered on Foundation's
+        // own reaper queue, so a lost exit event can cost at most this call's
+        // timeout — never a thread.
+        let finished = DispatchSemaphore(value: 0)
+        process.terminationHandler = { _ in finished.signal() }
+
         do {
             try process.run()
         } catch {
             reader.readabilityHandler = nil
             try? reader.close()
             return Result(status: -1, output: Data(), timedOut: false)
-        }
-
-        let finished = DispatchSemaphore(value: 0)
-        DispatchQueue.global(qos: .utility).async {
-            process.waitUntilExit()
-            finished.signal()
         }
 
         var didFinish = finished.wait(timeout: .now() + max(0, timeout)) == .success
--- a/Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
+++ b/Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
@@ -29,6 +29,10 @@ enum WindowEnumerator {
     /// tree within the process thread budget while collapsing it into a few
     /// bounded AX batches.
     private static let maximumConcurrentQueries = 24
+    /// Generous against the messaging timeouts above even with every batch
+    /// pathologically slow; only a dispatch pool with no thread to give (#971)
+    /// can outlast it, and then the freeze is theirs, not ours.
+    private static let accessibilityQueryDeadline: TimeInterval = 5.0
 
     static func listWindows(groupByApp: Bool = UserDefaults.standard.bool(forKey: DefaultsKey.switcherMergeTabs),
                             preservingGroupedWindows: Bool = false) -> [SwitcherItem] {
@@ -431,8 +435,11 @@ enum WindowEnumerator {
         let queryQueue = OperationQueue()
         queryQueue.qualityOfService = .userInteractive
         queryQueue.maxConcurrentOperationCount = min(maximumConcurrentQueries, orderedPIDs.count)
+        let pending = DispatchGroup()
         for pid in orderedPIDs {
+            pending.enter()
             queryQueue.addOperation {
+                defer { pending.leave() }
                 guard let windows = accessibilityWindows(
                     for: pid,
                     bundleIdentifier: bundleIdentifiers[pid],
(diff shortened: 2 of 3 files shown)


PathGao:
#971

**1. `cancelAllOperations()` unbalances the group, in exactly the starved case it exists for.**

`addOperation {}` builds an `NSBlockOperation`; `start()` on a cancelled operation finishes without calling `main()`, so the block — and its `defer { pending.leave() }` — never runs. On the deadline path every queued-but-unstarted op is one outstanding `enter()`. When the last op object is released the group is torn down with enters outstanding, and libdispatch traps on that ("Group object deallocated while in use"). Cancelling also does nothing for the ops that already started: nothing in the block checks `isCancelled`, so those AX queries run to completion either way.

Delete the line:

```swift
_ = pending.wait(timeout: .now() + accessibilityQueryDeadline)
```

Queued ops then run whenever a thread frees up and leave normally; if the pool never recovers they never run, and the group stays retained by their closures instead of being deallocated mid-flight. The cancel bought nothing but the trap.

**2. The tests grep two files, the pattern is repo-wide.** I only have the diff, so I can't run `rg -n 'waitUntilExit|waitUntilAllOperationsAreFinished' Sources/`. If another spawn site or fan-out wait is left anywhere, the pool can still be starved from there and both fixes here only move the freeze.

**3. The source-text assertions are cwd-dependent and pin a literal.** `String(contentsOfFile: "Sources/Vorssaint/...")` is a relative path, so it resolves only when the process runs from the repo root. `./build.sh --test` does. I can't see whether `--selftest` runs this same block from the shipped binary — if it does, `--selftest` now fails anywhere but the repo. And pinning `pending.wait(timeout: .now() + accessibilityQueryDeadline)` as text means renaming the constant breaks the test with no behavior change. The concurrent-run check below it is the part that actually tests something.

Not verified: the output-collection path above line 59 isn't in the diff. If it still parks a pool thread per child to drain the pipe, the thread-per-spawn leak survives this change — the harness numbers suggest it doesn't, but I can't read that code.

**Impact:** on the deadline path — a starved pool, the #971 scenario — the enumerator can trap the app instead of returning a short window list. The runner change and the deadline itself are strict improvements.
**Merge:** needs the `cancelAllOperations()` removal first.

PathGao:
Closing this in favour of #989, which consolidates this work.

We found the same two defects and reached the same mechanism independently — the termination handler for the child, a deadline on the Accessibility fan-out. #989 carries your 5s ceiling and the reasoning for it over the 1.5s we had: yours is the better call, since the AX calls are already bounded by their messaging timeouts, so a tighter ceiling buys nothing against a starved pool and would truncate a genuinely slow batch on a busy Mac. You are credited as co-author on the commit.

Two differences worth having on the record for whoever picks this up:

- This PR uses `Fixes #971`, which closes the issue automatically on merge. #989 references it with `ref #971` instead, leaving that call to the maintainer.
- #989 also answers the open question in #971 about `AppSwitcher.swift:742`: it reaches `listWindows` through the `DispatchQueue.main.async` at `AppSwitcher.swift:536`, so the switcher is a second way into the same freeze. Same fix covers it.

Apologies for the collision — this went up a few minutes after I last checked the open PRs, so I did not see it until #989 was already open.

iltonandrew:
Closing note, and thanks — no objection to consolidating into #989, and your `cancelAllOperations()` finding was correct.

I checked it against the merged code: the trap needed the `DispatchGroup` I used, because a cancelled `NSBlockOperation` never runs `main()`, so the `defer { pending.leave() }` never fires and the group is deallocated with enters outstanding. #989 counts with an `NSCondition` and a plain counter instead, so a cancelled-but-unstarted op just never decrements — nothing to trap on. Your version avoids the defect by construction; mine would have shipped it on exactly the starved path it was written for.

**Your point 2 — the repo-wide grep you couldn't run.** Ran it on current main:

```
Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift:773       process.waitUntilExit()
Sources/Vorssaint/Services/DiskImageInstaller/DiskImageInstallerService.swift:303  process.waitUntilExit()
Sources/Vorssaint/Services/Cleaner/JunkCleaner.swift:262            process.waitUntilExit()
```

No `waitUntilAllOperationsAreFinished` left in `Sources/`.

None of the three reproduces #971's mechanism, which needed *one stranded thread per spawn, unbounded*: the runner's old wait was a fire-and-forget `async` onto the global concurrent pool at ~1800 spawns/hour. These three each run on a dedicated **serial** `workQueue` (Homebrew and the disk-image installer) or on a user-initiated cleaning pass, so at most one call is ever in flight per feature — a lost wakeup wedges that one feature and costs one thread, it cannot accumulate. Worth knowing they exist, but they don't put the pool back at risk, so I'd leave them unless you'd rather they moved to the runner for consistency.

The `--selftest` question you raised is also worth recording for whoever writes the next source-shape test: those assertions live in `Tests/MetricsTests.swift`, which is built into the `metrics-tests` binary by `build.sh --test` only — `--selftest` runs from the shipped app and never executes that block, so the relative-path dependency is confined to test runs from the repo root. Your point about pinning a literal expression stands, though.

Issue #969, opened 2026-08-25
App is not responding when i turn off built-in retina display


dmitryuchkin:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When i turn off built-in retina display in display section vorssaint is not responding 

### Steps to Reproduce

1 connect other display
2 open vorssaint 
3 go to display section 
4 press off button on built-in display

### Feature Area

Menu bar

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

### Language

Russian

iltonandrew:
Taking a look at this — tracing the display-off path for a candidate fix (no second display here, so final confirmation may need your hardware).

El-Muerte-Brujo:
The bug exists also on macOS 26.6.2 (mbp M2)

PR #981, opened 2026-08-25
fix(display): decide the display power button without asking the display server

iltonandrew:
Candidate fix for #969 — reasoned from the code, not reproduced; I have no second display here.

## Problem

With an external display connected, opening the Displays section and pressing off on the built-in display freezes the app — beachball, "not responding" (reported on 3.3.2, macOS 27). Switching off an external display is fine, and with no external display the button is correctly disabled, which is why only this exact combination hangs.

## Root cause

A display power change runs its reconfiguration transaction synchronously on the main thread, which is where it has to run: CoreGraphics relays the change to AppKit inline on the thread driving the transaction (#747, #762). Switching off the built-in display is the case where that inline work has to move the open panel off the display that is going away and lay it out again, right there inside the transaction.

That layout evaluates `DisplayPowerButton`'s body, and the body called `canToggleDisplay`, which asked the display server for each display's info dictionary (`CoreDisplay_DisplayCreateInfoDictionary`) to filter out virtual devices. So the main thread put a question to the display server about the display it was, at that moment, still busy switching off. Nothing can answer it and nothing can time it out: the transaction cannot finish because its own callback never returns.

## The change

The set of displays that actually put a picture in front of a person is now published by the rebuild that already reads every display's info dictionary, and `canToggleDisplay` reads that snapshot. The body no longer touches the display server or the state lock.

The live reading stays exactly where it guards the switch itself — `toggleDisplay` still checks it before starting, the work queue checks it again against live topology, and headless recovery is untouched — so a stale snapshot can only make the button look more conservative, never let the last drawable display be switched off.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8542 checks), `./build/Vorssaint --selftest` → SELFTEST OK. `BrightnessService.swift` is not in the test binary, so the shape is pinned in `Tests/MetricsTests.swift` next to the existing display pins: the body of `canToggleDisplay` may not ask the display server or take the state lock. Reverting the change turns that pin red.

@dmitryuchkin, could you check whether switching off the built-in display from the Displays section still freezes the app on this build?

Refs #747
Refs #969

### DIFF
--- a/Sources/Vorssaint/Services/Display/BrightnessService.swift
+++ b/Sources/Vorssaint/Services/Display/BrightnessService.swift
@@ -58,6 +58,14 @@ final class BrightnessService: ObservableObject {
                                     category: "display")
 
     @Published private(set) var displays: [BrightnessDisplay] = []
+    /// The displays that currently put a picture in front of a person, as the
+    /// last rebuild found them. The panel decides from this snapshot instead
+    /// of asking the display server itself: `canToggleDisplay` is read from a
+    /// SwiftUI body, and a body is laid out again from inside the display
+    /// reconfiguration this app drives on the main thread, where a question
+    /// for the display server is one the same thread is already busy
+    /// answering (issue #969).
+    @Published private(set) var drawableDisplays = Set<CGDirectDisplayID>()
     @Published private(set) var pendingDisplayIDs = Set<CGDirectDisplayID>()
     @Published private(set) var displayControlFailure: DisplayControlFailure?
     @Published private(set) var brightnessOSDSupported = false
@@ -285,6 +293,7 @@ final class BrightnessService: ObservableObject {
         brightnessOSDSupported = false
         BrightnessOSD.teardown()
         if !displays.isEmpty { displays = [] }
+        if !drawableDisplays.isEmpty { drawableDisplays = [] }
         // Queue on the work queue first so this lands AFTER any operation
         // already in flight, then hand the reconfigurations to the main
         // thread, which is the only place they may run (see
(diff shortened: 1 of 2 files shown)


PathGao:
Read the diff. The topology half of this was already a snapshot — `knownTopology`/`knownActiveTopology` are only written by the rebuild — so the only genuinely live read you took out of the body is the per-display info dictionary, which is the one that hangs. That's the right cut.

Two things.

**`drawableDisplays` is a second copy of a fact that has to be written wherever `displays` is.** Three assignment sites in this file, all three patched here, and both rebuild publishes set the pair inside the same main-thread block, so they can't diverge today. Nothing enforces that. The new pin covers the body of `canToggleDisplay`, not the pairing — a fourth `self.displays = …` added later publishes rows against a drawable set from a topology that's gone, and no test goes red. A cheap pin on the source text (every `displays =` line paired with a `drawableDisplays =` line) would close it. I can only see the three assignments that are in this diff; if there's a fourth elsewhere in the file, it's unpatched.

**The staleness claim in the body is backwards in one direction.** "a stale snapshot can only make the button look more conservative, never let the last drawable display be switched off" — unplug the external and the snapshot still lists it until the rebuild publishes, so during that window the built-in's button stays enabled. More permissive, not less. The guarantee is actually carried by the second half of your sentence: `toggleDisplay` re-reads live and the work queue re-checks against live topology. Worth correcting, because the next person to touch `canToggleDisplay` may lean on the first half.

Not verified: `BrightnessSupport.canDisableDisplay` and the three-arg `drawableDisplayIDs` aren't in the diff, so I didn't check that the old wrapper's virtual filter and the new `virtualIDs` set agree on AirPlay, or whether the two-arg wrapper still has a caller. And I have no second display either, so nothing here says the freeze is gone.

#969

**Impact:** none on the switching path — this only changes what the button's enabled state is computed from, and the checks that guard the actual switch are untouched.
**Merge:** ship it.

Issue #915, opened 2026-08-24
"Keep going with lid closed" fails when macOS short username contains "@"


dhruvsaxena1998:
### Description

"Keep going with lid closed" fails with `Couldn't turn on closed-lid mode. Try again.`
on accounts whose macOS **short username contains an `@`** - common on Macs enrolled
via Entra ID / Jamf / Google Workspace SSO, where the short name is the user's email
address (`name@company.com`).

The error appears **instantly, with no admin password prompt at all**

### Root cause

`strings` on the 3.3.2 binary shows this validation regex sitting directly between the
`visudo -c -f` string and the `... with administrator privileges with prompt` string
(byte offsets 14468816 / 14468848 / 14468944) -  i.e. inside the same code region that
builds `/etc/sudoers.d/vorssaint-clamshell`:

```
^[A-Za-z0-9._-]+$
```

`@` is not in that character class, so email-style short names are rejected and the rule
is never written. Every later toggle then runs `sudo pmset disablesleep 1` against a rule
that doesn't exist, and fails.

Worth noting: **sudoers itself has no problem with these usernames.** The rule the app
would have written validates cleanly:

```
name@company.com ALL=(root) NOPASSWD: /usr/bin/pmset disablesleep 1, /usr/bin/pmset disablesleep 0
```
→ `visudo -c` says `parsed OK`

So the guard is stricter than the thing it's guarding.

### Supporting evidence

- `/etc/sudoers.d/` was empty - the rule was never installed
- no `sudo` or SecurityAgent events in `log show --last 3d` - the admin prompt never fired
- `pmset -g` had no `SleepDisabled` line

### Workaround (verified working)

Writing the exact rule by hand unblocks it completely:

```bash
printf '%s ALL=(root) NOPASSWD: /usr/bin/pmset disablesleep 1, /usr/bin/pmset disablesleep 0\n' "$(id -un)" > /tmp/vc
sudo install -m 0440 -o root -g wheel /tmp/vc /etc/sudoers.d/vorssaint-clamshell
sudo visudo -c
```

After this the toggle works normally and `pmset -g` reports `SleepDisabled 1`. That also
confirms the username check runs **only on the install path**, not on every toggle.

### Steps to Reproduce

1. Use a Mac whose account short name contains `@` (check with `id -un`)
2. Open Vorssaint → Keep Awake (Keep awake first time user)
3. Turn on "Keep going with lid closed"
4. Error appears immediately, with no password prompt

[...]

Issue #828, opened 2026-08-23
The function: keep going with the lid closed does not work for me even with all the permissions!


Farbodart:
### Description

more info: 
mac OS: 26.5.2
MacBook m2 max

### Steps to Reproduce

after all permissions and restart:
activate the feature
close the lid
it goes off

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

PathGao:
Triage note, part of a batch pass over the open issues. The labels are a contributor's read, not a maintainer ruling.

Two things, and the first is a genuine gap in the report rather than a brush-off.

**What is missing:** whether Vorssaint shows the helper rule as configured in Settings, and what `pmset -g | grep SleepDisabled` prints while the feature is switched on and the lid is still open. Those two answers separate "the sudoers rule is not actually in place" from "the write ran and did not take", and they lead to completely different fixes.

**Why it is worth checking now:** there is a specific defect nearby that matches your description. `ShellSupport.pmsetDisableSleepOnQueue` (`Sources/Vorssaint/Services/ShellSupport.swift:208-210`) treats the exit status of `sudo -n pmset disablesleep 1` as proof the setting took, and never reads it back — while `isConfigured()` in the same file (`:154-160`) does read back `pmset -g`. If that write is rejected in a way that still exits 0, the internal flag (`DefaultsKey.sleepDisabledFlag`, `Core/Defaults.swift:37`) reads as on and the machine still sleeps, which is exactly what you describe.

For completeness: the already-merged #362 touched this area but fixed ordering during rapid toggling, not this.

PathGao:
Don't bother collecting the two things I asked for earlier. A fix has landed, so it's faster to wait for the release and just see whether closing the lid still puts your Mac to sleep.

If it's still broken after you install that version, say so on this issue and I'll ask again then.

PathGao:
I've labelled this `half`. Your report has two possible causes behind it, a fix has landed for one of them, and nobody knows yet which one you hit — so the report stays open and you won't be asked to confirm a fix that may not be yours.

**The half that landed:** on some Macs the permission rule was never actually written, because the account's short name was rejected before the password prompt could appear. That is fixed. If your Mac was in that state, closing the lid should behave once you're on a version that has the fix.

[...]

Issue #501, opened 2026-08-05
[Feature]: A toggle to disable the screenshot thing


Yahddyyp:
### Problem or Motivation

The thing provided in the image appears when i screenshot and i just want it to copy to my clipboard.

### Proposed Solution

A toggle to disable it appearing as you can already access all the options it provides when you open the editor with the shortcut.

### Feature Area

Tools

### Mockups / References

<img width="615" height="373" alt="Image" src="https://github.com/user-attachments/assets/f870c8b4-622b-458c-bc5b-06cd868a0474" />

vorssaint:
Hello! You can go to:
Vorssaint → Settings → Screenshot → Default Action → Choose between Save, Copy or Edit the image after taking the screenshot

Tell me if it works for you. Alternatively, you can just command + c after the screenshot, quick copying it. Pressing "E" opens the editor.

PR #983, opened 2026-08-25
fix(keep-awake): grant the closed-lid sudoers rule by uid so any short username works

iltonandrew:
Refs #915.

## What was wrong

"Keep going with lid closed" installs a NOPASSWD sudoers rule for `pmset disablesleep`, and guarded the username it interpolated with `^[A-Za-z0-9._-]+$`. On SSO-enrolled Macs (Entra ID / Jamf / Google Workspace) the short name can be a full email address; `@` fails the guard, `install` bails before the password prompt ever fires, and every later toggle runs `sudo -n` against a rule that was never written — exactly the instant, promptless "Couldn't turn on closed-lid mode" in the report. As the reporter showed with `visudo -c`, sudoers itself accepts these usernames: the guard was stricter than the thing it guarded.

## The change

The rule is now granted by uid using the standard sudoers `#uid` user spec, built from `getuid()`:

```
#501 ALL=(root) NOPASSWD: /usr/bin/pmset disablesleep 1, /usr/bin/pmset disablesleep 0
```

The username guard existed because the name passed through two interpreters — the sudoers grammar and the single-quoted shell string that writes the file. A uid interpolates as bare digits, which neither can read as anything else, so the guard moves into the construction instead of rejecting real accounts. This also grants to the account that actually runs `sudo -n` (the process uid), and it covers short names the old guard rejected beyond `@`. Widening the regex was rejected on purpose: `@` can take a user@host reading in sudoers, and ambiguity in a NOPASSWD grant is disqualifying.

Nothing else changes: the rule path is the same, install still validates with `visudo -c -f` and rolls back on failure, `isConfigured()` still proves the passwordless path by running it, and removal/uninstall are path-based — an existing username-form rule (including the issue's manual workaround) keeps working and is replaced on the next install.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8549 checks), `./build/Vorssaint --selftest` → SELFTEST OK. Tests pin the exact rule shape and that a rule for any uid stays inside a character set with no shell or sudoers metacharacters. `visudo -c -f` on generated rule files parses OK for the uid form (and, for the record, for the email-style username the old guard rejected). I don't have an SSO-enrolled account here, so the end-to-end confirmation that the prompt now appears on an email-style short name is one only the reporter's machine can give — but the rejection happened strictly before any system interaction, and that rejection is gone.

Ref #828

### DIFF
--- a/Sources/Vorssaint/Services/ShellSupport.swift
+++ b/Sources/Vorssaint/Services/ShellSupport.swift
@@ -133,12 +133,6 @@ enum Sudoers {
         "/etc/sudoers.d/vorss-clamshell",
     ]
 
-    private static var safeUser: String? {
-        let user = NSUserName()
-        let valid = user.range(of: "^[A-Za-z0-9._-]+$", options: .regularExpression) != nil
-        return valid ? user : nil
-    }
-
     /// Serializes every touch of the SleepDisabled state. The probe below
     /// re-applies the value it just read; racing it against a concurrent
     /// disable (launch recovery, a session ending) could resurrect a stale
@@ -167,11 +161,10 @@ enum Sudoers {
     }
 
     static func install(completion: @escaping (Bool) -> Void) {
-        guard let user = safeUser else {
-            completion(false)
-            return
-        }
-        let rule = "\(user) ALL=(root) NOPASSWD: /usr/bin/pmset disablesleep 1, /usr/bin/pmset disablesleep 0"
+        // Granted by uid, not username: a short name is free-form text on
+        // SSO-enrolled Macs (name@company.com, #915) and the old validation
+        // rejected it before the password prompt could even appear.
+        let rule = SudoersSupport.clamshellRule(uid: getuid())
         // Clear any earlier-named rule first, then write and validate the new one
         // (a failed check rolls back). Same password prompt either way.
         let legacy = legacyRulePaths.joined(separator: " ")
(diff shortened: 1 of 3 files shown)


PathGao:
`#uid` is the right form. sudo resolves the invoking user from the real uid, so the grant and the later `sudo -n` probe can't disagree the way a name-based rule can — the name never enters the picture at all. Correct fix for #915.

One thing about the new test's claim:

```swift
// ... the whole line must stay inside a character set that neither sudoers
// nor a single-quoted shell string can read as anything but itself.
expect(SudoersSupport.clamshellRule(uid: uid_t.max)
       .range(of: #"^#[0-9]+ [A-Za-z0-9()=:,./ ]+$"#, ...
```

`#` is the comment character in both of those interpreters. The regex hardcodes it at position 0, outside the bracketed set, so it permits exactly the character the comment says can't be there. The line is safe, but for two reasons neither the regex nor its comment captures: sudoers special-cases `#` followed by digits where a user spec is expected, and the write is single-quoted. Only the second is under this repo's control, and nothing asserts it.

That asymmetry matters because the failure mode is silent. `echo #501 ALL=... > file` at word start writes an *empty* file, `visudo -c -f` parses an empty file OK, so the rollback never fires, install reports success, and `isConfigured()` then fails with no prompt — the same promptless shape as #915. The old guard incidentally covered this by banning `#` outright; now the quoting is the only thing holding it. The write itself is past `let legacy = ...` and outside this diff, so I did not verify the quoting — I'm taking your body's word that it's single-quoted. If you want a cheap guard, assert the constructed command string rather than the rule. Not a merge blocker.

Second point, unverified: `safeUser` was private to `enum Sudoers`, so nothing outside could have referenced it and there's no dangling call site. But any *other* place in the app that writes a sudoers rule or interpolates `NSUserName()` into a shell string would carry its own copy of that guard, and the same SSO short name breaks it there too. I can't see outside the diff to check.

**Impact:** none today. A future edit that unquotes or double-quotes the write turns the rule into an empty file that still passes validation.
**Merge:** ship it.

Issue #974, opened 2026-08-25
I have the dock at the left side with auto hide. When I put my cursor over the app preview, the dock hides


rodrigorebolledo:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

IMO a better user experience would be to stick the dock while the cursor is over the preview

### Steps to Reproduce

set your dock at the left with auto hide
activate window preview
move your cursor to the left corner until the dock appears
put your mouse over an open app
put your mouse over the preview
then the dock dissapears and leave you with the preview windows on screen until you take off the mouse

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

English

iltonandrew:
Taking a look at this one.

PR #984, opened 2026-08-25
fix(dock-preview): follow an auto-hiding Dock to the vacated edge instead of floating detached

iltonandrew:
Candidate fix for #974.

## What happens

With the Dock auto-hiding, moving the cursor from a Dock icon onto its preview panel takes the pointer out of the Dock's reveal region, so macOS hides the Dock — and the preview is left floating detached in the middle of the screen until the cursor leaves.

## Why the Dock can't just be held open

The Dock owns its auto-hide reveal region; there is no public API for another process to extend that region to cover its own panel. The app's hover corridor already keeps the *preview session* alive when the cursor enters the panel — it just cannot keep the *Dock* open. Rejected on those grounds: shrinking the panel gap (already zero under auto-hide; adjacency doesn't make the panel part of the Dock window), pointer warping (disruptive), and a new private Dock-control symbol (no comparable precedent in the codebase, OS-fragile).

## The change

When the cursor enters a preview while the Dock auto-hides, the service briefly watches the Dock's layer-20 window in the on-screen window list (100ms cadence, one panel session at most, stops at first trigger or when the panel closes). The moment the Dock's slide-out completes, the panel is pulled flush to the vacated screen edge, preserving its position along the Dock axis — the interaction stays anchored to the edge the user is working on instead of floating detached. Left, right, and bottom Docks are covered.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8551 checks), `./build/Vorssaint --selftest` → SELFTEST OK. The reposition geometry is a pure `DockPreviewSupport` helper with unit tests for all three orientations, and a source-shape assertion pins the service wiring (timer started on panel entry, on-screen-list check, helper call).

Manual check (what I could not complete on this machine — the dev build's screen-recording grant, which Dock Preview requires, was invalidated by re-signing): Dock on the left with auto-hide on, Dock Preview enabled → reveal the Dock, hover an app with windows, move onto the preview and rest there → when the Dock slides away, the panel should slide flush to the left edge instead of staying detached mid-screen; window activation and leaving the panel behave as before.

One known edge left as-is: if the Dock is re-revealed while the panel sits at the edge, the panel keeps its spot until the session ends (the watcher is one-shot per panel entry). Happy to extend it if that case matters.

Refs #974


PathGao:
The mechanism hangs on the one premise you couldn't test: that WindowServer drops the Dock's layer-20 window from `.optionOnScreenOnly` once the slide-out finishes. If it doesn't — if the window stays listed at an off-screen origin — `dockIsRevealed` returns true forever and the feature is a silent no-op. Nothing in the test additions catches that: the geometry helper is pure, and the source-shape assertion only checks the wiring exists. Both stay green with the trigger never firing once. One `CGWindowListCopyWindowInfo` dump with the Dock hidden settles it.

`panel.setFrame(edgeFrame, display: true, animate: true)` blocks the calling thread until the animation completes. This service owns a CGEvent tap on the main run loop (`tap`/`runLoopSource`), so every mouse event queues behind the slide. It's a one-shot ~70pt jump:

```swift
panel.setFrame(edgeFrame, display: true)
```

Third, the panel moves out from under a cursor that isn't moving. Bottom Dock, pointer resting in the upper half of the preview, Dock slides away → the panel drops by the Dock's height and the pointer is now above `activePanelFrame`. The next twitch re-runs `currentZone` and lands outside `.panel`. If the hide path fires there, resting the pointer dismisses the preview this change exists to keep usable. `currentZone` and the corridor recompute aren't in the diff, so I can't tell which zone the pointer ends up in — same manual setup as the check above, worth doing in one pass.

Minor: `reattachPanelToScreenEdge` sets the frame without `clampedPanelFrame`, unlike the other placement path I can see. Only bites a panel taller than the visible frame, so ignore it if that can't happen.

**Impact:** if the on-screen-list premise is wrong, the fix never fires and no one finds out; if the cursor case is real, an entered preview closes on the first mouse move after the Dock hides. The `animate: true` stall delays mouse events for the animation's duration whenever the trigger does fire.
**Merge:** needs one device check that the hidden Dock actually leaves the on-screen window list first.

ref #974

iltonandrew:
All three addressed, and the device check you asked for is done.

**1. The on-screen-list premise holds.** Dumped `CGWindowListCopyWindowInfo(.optionOnScreenOnly)` filtered to Dock-owned windows on a left auto-hiding Dock, macOS 26.6.2:

```
===== HIDDEN (pointer at screen centre):
dockWindow level = 20
  Dock win 'Wallpaper-' layer=-2147483624 bounds=0,0 1512x982

===== REVEALED (pointer at left edge):
dockWindow level = 20
  Dock win 'Dock'       layer=20            bounds=0,0 1512x982
  Dock win 'Wallpaper-' layer=-2147483624 bounds=0,0 1512x982
```

WindowServer does drop the layer-20 strip once the slide-out finishes, so `dockIsRevealed` returns false exactly when the Dock is parked and the trigger fires. (Incidentally that dump also confirms the thing #987 rests on from the other direction: the strip's bounds are the whole display, not a strip.)

**2. The resting-cursor case was real, and worse on a left Dock than your bottom-Dock example.** `panelStayMargin` is 6pt; under auto-hide the panel is placed flush against the revealed Dock's inner edge, so reattaching moves it by the Dock's thickness — ~70pt, as you estimated. A pointer resting anywhere in the outer ~64pt of the panel ended up outside the new frame, and the next twitch took `currentZone` to `.outside` and dismissed the preview this change exists to keep usable.

Fixed by keeping the frame the panel left behind as a second stay region: it counts as `.panel` until the pointer reaches the new frame, at which point it is cleared (and it clears on session teardown). So a pointer that never moved is never treated as having left.

**3. `animate: true` is gone** — you're right that it blocks the main run loop this service's tap is served by, for a one-shot ~70pt jump. Now `panel.setFrame(edgeFrame, display: true)`.

**Minor:** the landing frame now goes through `clampedPanelFrame` like the other placement path.

Gate after the changes: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8553 checks), `--selftest` → SELFTEST OK. Two new source-shape assertions pin the grace region and the un-animated clamped landing, so neither can be undone silently — you were right that the previous tests would have stayed green with the trigger never firing.

iltonandrew:
Two more fixes from an independent adversarial review of this branch, both pushed.

**1. The grace region never expired.** It existed for a pointer that never moved, but a pointer moving *away* — upward from a bottom-Dock panel, say — crossed a ~70pt band of empty space that still reported `.panel`, so `cancelPendingHide` kept firing and the preview refused to dismiss until the cursor cleared a rectangle with nothing visible in it. The region now records where the pointer rested at reattach and stops standing in for the panel once the pointer genuinely travels (`reattachGraceTravel`, 24pt, doc-commented as a calibration knob). A resting hand's jitter stays covered; a deliberate move away is an answer.

**2. Timer teardown was asymmetric with the grace state** — invalidated in `endSession` but not `tearDownVisuals`, while the grace frame was cleared in `tearDownVisuals`. No misbehavior today (switching icons resets `hasEnteredPanel` synchronously, so a stray tick self-cancels), but the pairing invited a future bug; both now live in `tearDownVisuals`.

The same review also claimed Mission Control would spuriously trigger the reattach by removing the Dock's window from the on-screen list. **Tested and refuted**: with the overview held open, the layer-20 Dock window stays listed (Mission Control *adds* scene windows at layers 20/18 rather than removing the strip), so `dockIsRevealed` holds and no reattach fires. Recording it here so the next reviewer doesn't have to re-derive it.

Gate: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8553 checks), `--selftest` → SELFTEST OK.

iltonandrew:
No new work here — flagging that this one's `incomplete` label predates the rounds that answered it, so it may just be waiting on a re-read.

The label went on at 2026-08-25T16:37Z. Both replies to that round landed after it: the device check and the three fixes at 19:39Z that day, then the grace-region expiry and timer pairing at 2026-08-26T03:55Z. Nothing has been asked since, and I have nothing outstanding on my side.

Also rebased onto current `main` — it was 76 behind and the rebase was clean, no conflicts, so the branch is mergeable as it stands rather than needing that done at merge time.

ref #974

PathGao:
Re-read done. The round that put `incomplete` on is closed — the dump answers the on-screen-list premise, the grace region and the un-animated clamped landing are in. It stays on for a different reason.

`resizePanelForCurrentWindows()` is the other path that writes `activePanelFrame` and moves the panel, and it doesn't know the panel was reattached. It rebuilds the frame from `activeIconFrame` through `DockPreviewSupport.panelFrame(...)` — the Dock-anchored position — and animates the panel there. It's reachable from the hovered panel, not just the pinned one: every card carries `closeButton` → `service.close` → `finishClosing` → `applyClosedWindowRemoval` → resize. So the Dock hides, the panel snaps to the edge, you close one window from it, and the panel jumps back by the Dock's thickness to exactly the detached float this PR exists to remove. The pointer that was on the ✕ is then outside the new frame by an order of magnitude past `panelStayMargin` unless it happens to land in the overlap, and the grace region only covers the reattach direction — next move reads `.outside` and the session ends. (`resizePanelForCurrentWindows` isn't in the diff; I read it on `main`.)

Second, the watcher isn't one-shot. The body says "one panel session at most, stops at first trigger or when the panel closes", but the guard is `dockVisibilityTimer == nil` and the trigger sets it back to nil, so the next mouse-moved event landing in `.panel` starts a fresh timer that fires 100ms later and reattaches again while the Dock is still parked. Geometrically idempotent, but it's a full `CGWindowListCopyWindowInfo` dump per 100ms of pointer movement inside the panel, on the run loop your tap is served by. It also makes the known-edge note in the body wrong about the mechanism.

One flag covers both:

```swift
private var didReattachForSession = false   // cleared in tearDownVisuals, alongside the grace state
```

Add `!didReattachForSession` to the timer guard, set it in `reattachPanelToScreenEdge`, and in `resizePanelForCurrentWindows` run the computed frame through `panelFrameWhenDockHidden` while it's set.

Not verified: I'm reading frames, not watching pixels, so I haven't seen the resize jump happen. The direction and magnitude come from the same geometry your own reattach uses, so I don't think it's borderline.

**Impact:** closing a window from a reattached preview puts the panel back in the state #974 reports and most likely dismisses the session; the watcher polls the window list every 100ms while the pointer moves inside the panel instead of stopping at first trigger.
**Merge:** needs the resize path to respect the reattached frame first.

ref #974

iltonandrew:
Both defects confirmed from the code before fixing, and both are in. The resize path was the real one.

**Your call chain is right, and one link longer than you wrote it.** From the hovered panel: `closeButton` → `onClose` → `onCloseWindow` → `service.close` → `finishClosing` → `applyClosedWindowRemoval` → `resizePanelForCurrentWindows()` when windows remain. So it is reachable from an ordinary hover, not only a pinned panel, exactly as you said. And the watcher: after a reattach the timer invalidated itself and set `dockVisibilityTimer` back to `nil`, so the next `.panel` mouse-moved event armed a fresh one — a full `CGWindowListCopyWindowInfo` dump per 100ms of pointer movement inside the panel.

**The flag, in your shape.** `didReattachForSession`, set in `reattachPanelToScreenEdge`, cleared in `tearDownVisuals` immediately beside `reattachGraceFrame` / `reattachGraceOrigin` — the same place, so the teardown asymmetry you flagged two rounds ago cannot come back through this. It is deliberately *not* cleared at the two grace-region expiry sites: the grace region expires when the pointer travels, but the reattachment itself outlives that for the session.

**Both decisions moved into `DockPreviewSupport` as pure functions**, rather than left inline and described by a shape assertion:

- `resizedPanelFrame(_:didReattachForSession:screenVisibleFrame:orientation:)` — returns the Dock-anchored frame untouched when the session never reattached, and routes it through `panelFrameWhenDockHidden` when it did. `resizePanelForCurrentWindows` still rebuilds from the icon for the new dimensions; only the Dock-facing axis is put back at the edge.
- `shouldStartDockVisibilityTimer(hasActiveTimer:didReattachForSession:autohide:)` — the timer covers repeated mouse moves before the Dock goes, the flag covers the ones after.

That is on your "logic a test cannot reach is a missing seam" rule: these are behaviour tests against real functions rather than text assertions about a service I cannot instantiate in the suite.

**Four assertions, both directions each.** The two that fail when the fix is reverted — a reattached preview keeps the vacated edge through a resize, and does not re-arm its watcher — plus their complements: a preview that never reattached still resizes to the Dock-anchored frame, and an auto-hiding Dock still arms the watcher the first time. Without the complements either helper could be mutated into a constant and stay green, which is a test that passes either way. Expected frames are computed from `panelFrame` / `panelFrameWhenDockHidden` rather than copied in as numbers.

**Gate:** 0 warnings. `./build.sh --test` — 8807 checks, **1 failure**: the `dispatch pool starvation` precondition #1014 exists to skip, red on this 15-core M5 Pro and green on CI's smaller runners; it came up identically on an unrelated branch gated in the same sitting. Reporting it rather than writing `TESTS OK`. `./build/Vorssaint --selftest` — `SELFTEST OK`.

Mutation, both guards removed at once: failures go 1 → 3, naming *a reattached Dock Preview does not re-arm its visibility watcher* and *a preview that never reattached keeps resizing to the Dock-anchored frame*. Worth noting which one did **not** fire — *keeps it at the vacated screen edge* stays green under that mutation, because an unconditional helper still returns the edge frame for the reattached case. That assertion alone would have passed a helper mutated into a constant; the complement is what catches it.

Not verified, and it is the same gap as last round: I am reading frames, not watching pixels. Nobody here has driven a real auto-hiding Dock through close-a-window-from-a-reattached-panel and watched it hold. The direction and magnitude come from the same geometry the reattach itself uses, so I do not think it is borderline — but it is reasoned, not seen.

ref #974

PathGao:
Gap's closed.

I grepped `main` for the shape rather than trusting the call chain: `DockPreviewService` writes `activePanelFrame` and moves the panel in exactly two places — `showPanel`, which begins a session after `tearDownVisuals` has already cleared the flag, and `resizePanelForCurrentWindows`, which now routes through `resizedPanelFrame`. The `resizePanel()` / `setFrame(clampedPanelFrame(frame), …)` pair further down the same file belongs to `DockPreviewPinnedPanel`, a separate class that resizes from `panel.frame` and so keeps whatever position it already has. No third site.

Closing the clamp thread from two rounds ago rather than leaving it for the next reader: the asymmetry is inert. `panelFrame` already bounds height to `visibleFrame.height - 2 * edgePadding`, and `panelFrameWhenDockHidden` only rewrites the origin to `minY + edgePadding`, so the resize path's result is inside the visible frame by construction. Nothing for `clampedPanelFrame` to move. Leave it.

One thing inside the function you touched that is not yours and not for this PR: the resize drops the pointer on its own. `panelFrame` centres the panel on the icon, so closing one of N windows pulls both ends in by half a card, and a pointer resting on the first or last card's ✕ lands outside the new frame with no grace region covering it — next twitch reads `.outside`. That is on `main` today and has nothing to do with auto-hide. The grace machinery you just added would cover it in a line, but it wants its own report, not #974's scope.

Still unverified, and you said so first: nobody has watched this on a real auto-hiding Dock. Both rounds since the window-list dump have been frame-level reasoning. Not holding on it — the device check that mattered is done.

The two checks read as no state from where I'm looking, so that input to the label is yours to confirm.

**Impact:** none outstanding that I can find.
**Merge:** ship it.

ref #974

Issue #960, opened 2026-08-25
fix(dock-click): prevent swallowing clicks in borderless fullscreen games & optimize high-polling mouse events


marspater:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description
When running any MOBA game through Steam (e.g. Dota 2) in **Desktop-Friendly Fullscreen (Borderless Windowed)** on macOS 26/27 with a high-polling rate gaming mouse (Corsair wired via USB-C at 1000Hz–8000Hz):
1. **Clicks on the bottom-left mini-map are completely swallowed / unresponsive.**
2. **Cursor movement experiences severe stuttering and micro-freezes during gameplay.**

Exiting Vorssaint immediately resolves both issues.

---

#### 1. Mini-Map Clicks Swallowed by `DockClickService`
* **File**: `Sources/Vorssaint/Services/DockClick/DockClickService.swift` and `DockClickSupport.swift`
* **Trigger Condition**: In Desktop-Friendly Fullscreen (borderless window on the standard desktop space), `visibleFrame == screenFrame`.
* `DockClickSupport.dockStripContains` (`DockClickSupport.swift:233-236`) falls back to `fallbackMargin = 120pt`, treating all clicks within 120pt of the bottom/left edge (where the mini-map resides) as inside the Dock strip.
* `revealedDockBounds(dockPID:)` (`DockClickService.swift:736-750`) checks `CGWindowListCopyWindowInfo(.optionOnScreenOnly)`, which on macOS 26/27 still reports the Dock layer 20 window as on-screen on that Space.
* `dockApplication(at:)` (`DockClickService.swift:795-797`) only checks the horizontal X-axis (`point.x >= frame.minX && point.x <= frame.maxX`).
* Any mini-map click matching the horizontal span of a Dock icon is intercepted, and `DockClickService.handle` returns `nil` (`DockClickService.swift:254`), swallowing the game's click.

#### 2. High-Polling Rate Stuttering on macOS 26/27
* **File**: `DockPreviewService.swift:284-360`, `FocusFollowsMouseService.swift:57-84`
* At 1000Hz–8000Hz polling rate, thousands of `mouseMoved` events flood `CFRunLoopGetMain()`.
[...]

Issue #938, opened 2026-08-25
Accessibility messaging timeouts have no owner: six writers, one global, three values


PathGao:
Same shape as #937, different surface. There I found eleven places that press keys for the user and disagree about how. This is the Accessibility messaging timeout: how long Vorssaint waits for another process to answer. Six features set it, three different values, and they are all writing the same variable without knowing it.

I had a patch ready for the one member that sets nothing. I am not opening it, because it would have made the underlying problem worse, and because the interesting decision is yours rather than mine.

> **State.** #1120 answers questions 1, 3 and 4 below: nothing writes the
> process-global value any more except `AppDelegate`'s deliberate one at launch,
> the file that capped nothing now caps its own elements, and the guardrail is
> in with a one-line allowlist. Question 2 — what each of these waits should
> actually be — is the part no measurement settles, and it is the only thing
> still open here.

## What the header actually says

`AXUIElement.h`, on `AXUIElementSetMessagingTimeout`:

> Pass the system-wide accessibility object if you want to set the timeout **globally for this process**. Setting the timeout on another accessibility object sets it only for that object, not for other accessibility objects that are equal to it.

So there are two very different calls behind one function name, and the difference is invisible at the call site — it depends on where the element came from, sometimes twenty lines earlier.

## Six features write the process-global value

| | value | what a timeout there stalls |
|---|---|---|
| `AutoQuitService.swift:791` | 0.15 s | the main thread, inside an event tap |
| `FinderCutPaste.swift:434` | 0.15 s | the main thread |
| `FocusFollowsMouseService.swift:13` | 0.25 s | **nothing** — its query runs on `queryQueue` |
| `WindowLayoutService.swift:1719` | 0.25 s | the main thread, inside an event tap |
| `WindowMaximizer.swift:245` | 0.35 s | the main thread, inside an event tap |
| `AppDelegate.swift:195` | 0.35 s | — set once at launch, on purpose |

`AppDelegate` is the one that means it: one write, at startup, documented as the
floor for the paths that set nothing. The other five each read locally as "this
feature's timeout" and are not.

All five are the same variable, last writer wins, and the triggers are independent: quitting an app, ⌘V in Finder, moving the mouse, a window layout shortcut, clicking a green button. Focus-follows-mouse picks 0.25 s for a path that runs on every mouse move; the next green-button click silently makes it 0.35 s.

Nothing in the code marks any of these as global, and no comment in any of them
mentions the others.

[...]

PR #987, opened 2026-08-25
fix(dock): stop swallowing clicks a fullscreen window covers

iltonandrew:
Candidate fix for #960 — I have no MOBA or high-polling-rate rig here, so the click half rests on a window-ordering fact only @marspater can confirm in-game. Everything below is either measured on macOS 26 or derived from the code.

## Problem

With a Steam game in Desktop-Friendly Fullscreen (borderless windowed), clicks on the bottom-left mini-map are swallowed, and cursor movement stutters with a 1000–8000Hz mouse. Quitting Vorssaint clears both.

## Root cause — the swallowed clicks

`DockClickService`'s tap ran three gates before deciding a click was a Dock icon click, and under borderless fullscreen all three said yes:

1. `dockStripContains` falls back to a 120pt band along the bottom, left *and* right edges whenever the Dock reserves no screen space — which is exactly what a game presenting fullscreen produces. The bottom-left corner where the mini-map lives is inside it.
2. `revealedDockBounds` checked that the Dock's layer-20 window was on screen and that its bounds contained the click. Measured on macOS 26: that window's bounds are the **whole display** (0,0 1512x982), not a bottom strip. The guard was true for every point on the screen whenever the Dock was on screen at all — it never did the job its comment described.
3. `dockApplication` matches along the Dock's long axis only (short-axis AX frames are shifted ~72pt on macOS 27), so any coordinate that lines up with an icon resolves to that app.

A borderless window is not `AXFullScreen`, so the fullscreen veto in `DockClickSupport.action` never fired either. The click became a `.minimize` and the tap returned nil.

Nothing in the chain asked the one question that matters: **is the Dock what is actually drawn at that point?** `CGWindowListCopyWindowInfo` returns windows front to back, and a game presenting fullscreen sits at or above the Dock's level over the same pixels.

## Root cause — the stutter

`DockPreviewService`'s tap runs an Accessibility hit-test for every `mouseMoved` inside the Dock proximity band (≥160pt from the edge — where a mini-map lives all match). Measured with the same call shapes under `-O`: **536µs per event** — 53.6% of a core at 1000Hz, 429% at 8000Hz. That is not schedulable: the main run loop serving the tap falls behind and the window server disables the tap for timing out.

For the record, `NSScreen.screens` is **not** part of this: 0.11µs per call (AppKit caches it), 0.2% of a core even at 8000Hz. Caching it would have been noise.

## The change

- **Clicks:** new `DockClickSupport.dockOwnsPoint` narrows the second gate from "the Dock's window contains the point" to "the Dock's strip is the frontmost window at the point". Same window-server query as before, so no added cost; the edge slop is granted to the strip only, never to the window in front. Desktop behavior is unchanged by construction: a normal window is layer 0 and can never be listed ahead of the layer-20 strip; the menu bar and Control Center never overlap it; Vorssaint's own windows are discounted by pid.
- **Stutter:** `DockPreviewService` now gates that path by rate as well as by place — one move per 1/60s reaches the hit-test, with the newest dropped point held in a single trailing work item so the position the cursor rests on still arms the hover (a naive throttle would eat exactly that event). Cost becomes 3.2% of a core regardless of polling rate. Hover intent is 100–250ms, so display-rate sampling is invisible.

No behavior change for the ordinary desktop, no guards removed, no new settings. Rejected on purpose: a "game detected" bypass (no reliable oracle, silently disables features on a guess) and pushing games into MouseAppExceptions (per-game config chore that leaves the default broken).

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8563 checks, 16 new), `./build/Vorssaint --selftest` → SELFTEST OK. Ten unit assertions cover `dockOwnsPoint` — including the borderless-fullscreen case, the parked auto-hide Dock, own-process and zero-alpha windows, and the multi-display case — and four source-shape assertions pin the tap-side gates and the sampling constant.

@marspater — could you try this build and report whether the mini-map clicks land and whether the hitching is gone? If clicks still vanish, the game window is being ordered *behind* the Dock rather than over it on your rig, and I'll need a different reveal-state signal (your note about a parked Dock still reported on-screen did not reproduce on macOS 26 here — `kCGWindowIsOnscreen` is absent while parked).

Note for the maintainer: this PR and #984 both touch `DockPreviewService.swift` in disjoint regions (this one in the mouse-move admission path, #984 in panel-session code); whichever lands second may need a trivial rebase.

Refs #960


PathGao:
Read the diff. The sampler logic checks out — `cancelPendingMove()` on the admit path is what stops a stale trailing item from replaying behind a newer admitted move, and that's the part that's easy to get wrong. Three things.

**1. "Desktop behavior is unchanged by construction" has a fourth case.** The body enumerates layer-0 windows, the menu bar / Control Center, and own-pid. Missing: click-through overlays from other apps — screen dimmers, colour-temperature tools, annotation HUDs. They sit above layer 20, cover the whole display, and need alpha > 0 to do their job, so the alpha gate doesn't catch them and `dockOwnsPoint` returns false for every Dock click while one is running. `CGWindowListCopyWindowInfo` exposes no `ignoresMouseEvents` key, so I don't have a cheap fix to offer. What I'd want is the case named in the doc comment on `dockOwnsPoint`, because the failure is silent and system-wide — the user finds Dock click actions dead with nothing to correlate it to, and that report is unrecognisable unless it's written down here.

**2. The sampler bounds throughput, not one slow call.** 3.2% of a core is the steady state against a healthy app. Against a hung one the hit-test is bounded by the AX messaging timeout — which per #938 is one global with three disagreeing writers — and a single call past that blocks the main run loop well beyond the tap's tolerance, at 60Hz exactly as at 8000Hz. I don't see a `.tapDisabledByTimeout` branch anywhere in the diff. If `DockPreviewService` already re-enables the tap elsewhere in the file, ignore this; if it doesn't, the feature stays dead until the app restarts and the sampler only lowers how often that happens.

**3. The multi-display assertion doesn't test a second display.** `CGRect(x: 0, y: 0, width: 1512, height: 900)` is the same origin as `dockScreen` — that assertion pins the vertical bound, not display separation. A frame at `x: 1512` would actually cover the claim the body makes.

Not verified: `discardFarMouseMove`, `dockStripContains`, and `dockApplication` bodies are outside the diff, so I checked the new gate's inputs but not what happens downstream of it. No opinion on the window-ordering fact — you already flagged that as needing the reporter's rig.

**Impact:** none on an ordinary desktop, unless a click-through overlay app is running — then every Dock click action stops until that app quits. The preview stutter fix is unconditional and safe.
**Merge:** hold for @marspater's report on #960.

ref #960

iltonandrew:
All three addressed.

**1. Click-through overlays — named, not guessed at.** You're right that this is the case worth writing down, and I couldn't find a safe way to detect it either. `dockOwnsPoint`'s doc comment now states plainly that click-through overlays from other apps (screen dimmers, colour-temperature tools, annotation HUDs) read as occluding, and that the window list exposes no `ignoresMouseEvents` to separate them from a real fullscreen window.

Heuristics I evaluated and rejected, so the next person doesn't retry them: frontmost-app ownership (unsafe across focus transitions, and would exempt a real fullscreen game — straight back to #960), alpha thresholds (overlays and real windows both run partial alpha), `kCGWindowSharingState` (describes capture eligibility, not hit-testing), and menu-bar presence (says nothing about whether a window takes clicks). Each one trades a silent-but-narrow failure for the loud original bug. The known behavior now has an assertion, so it's intentional rather than accidental.

**2. `.tapDisabledByTimeout` is already handled** — `DockPreviewService.swift:333`, covering both `.tapDisabledByTimeout` and `.tapDisabledByUserInput` and re-enabling only the retained tap. So the sampler lowers how often the tap gets disabled, and recovery when it does was already there. Your read of the risk was right; it just predates this diff.

**3. Multi-display assertion fixed.** It used the same origin as `dockScreen`, so it pinned the vertical bound rather than display separation, exactly as you said. The second display's window now sits at a genuinely non-overlapping origin.

Gate: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8564 checks), `--selftest` → SELFTEST OK. Still holding for @marspater's report on the window-ordering question.

PathGao:
Point 2 is settled — `:333` re-enabling only the retained tap is the right shape, and it predating this diff is the answer.

One thing I got wrong in my read of the sampler. I said `cancelPendingMove()` on the admit path is what stops a stale trailing item from replaying. It's half of it — the discard path returns without cancelling:

```swift
if discardFarMouseMove(axPoint: point) { return Unmanaged.passUnretained(event) }
guard admitMouseMove(axPoint: point) else { ... }
```

So a point the cursor passed *through* can outlive the moves that prove it left. Sweep across the strip, exit the band, come to rest inside the sample interval: the last in-band point is sitting in `pendingMovePoint`, the far moves that follow return before they can overwrite it, and the work item fires and arms a hover for an icon the cursor is no longer near. The trailing item exists to preserve the point the cursor rested on; in that path it resurrects one it didn't.

The panel-open case is fine on its own — if `discardFarMouseMove` returns false while something is open, the far move reaches `admitMouseMove` and overwrites `pendingMovePoint` itself. The gap is exactly where discard returns true.

```swift
if discardFarMouseMove(axPoint: point) {
    cancelPendingMove()
}
```

Same rule the admit path already follows: newest information wins. It can't eat the rest case — if the cursor rests in the band, no far move arrives to cancel anything.

Not verified: `discardFarMouseMove` and the `mouseMoved` branch of `handleOnMain` are outside the diff. If discard already returns false whenever a hover is pending, the window narrows to the first arm — it doesn't close, since nothing is pending at the moment the cursor leaves.

1 and 3 read as fixed: the doc comment names the overlay case and the heuristics you rejected, and the second display's window sits at a genuinely non-overlapping origin now.

**Impact:** with no preview open, sweeping past the Dock and stopping just outside the band can arm a preview for an icon the cursor already left. Bounded by one sample interval.
**Merge:** hold for @marspater's report on #960; fold the one-line cancel in before it lands.

iltonandrew:
Good catch, and you're right — that's a real hole and it's mine.

`discardFarMouseMove` guards on `isRunning, !isVisible, pendingHover == nil`, so with nothing open and no hover armed yet, the exact sequence you describe goes through: the last in-band point is sitting in `pendingMovePoint`, every far move that follows returns before it can overwrite it, and the work item fires and arms a hover for an icon the cursor already left. The trailing item exists to preserve the point the cursor rested on, and in that path it resurrects one it didn't.

Folded in your one-liner before landing:

```swift
if discardFarMouseMove(axPoint: point) {
    cancelPendingMove()
}
```

And it can't eat the rest case, for the reason you gave: `discardFarMouseMove` only returns true when the point is *not* near the Dock, so a cursor resting in the band never produces a far move to cancel anything.

On your unverified note — `pendingHover == nil` being in that guard does narrow the window to the first arm exactly as you guessed, but it doesn't close it, since nothing is pending at the moment the cursor leaves. Which is the case that bit.

Gate after the change: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8564 checks), `--selftest` → SELFTEST OK. Still holding on @marspater for the window-ordering question.

Issue #574, opened 2026-08-10
[Bug]: Macbook pro m5 pro: "Fan control is not available on this Mac"


denmarin:
### Description

I just downloaded the newest update and downloaded fan control feature but it just wouldn't work. i've tried restarting the app and deleting and downloading this particular feature. btw thank you so much for this amazing app!

### Steps to Reproduce

1. Open Vorssaint
2. Download the fan control feature
3. See the bug in fan control menu

### Screenshots / Recordings

<img width="316" height="324" alt="Image" src="https://github.com/user-attachments/assets/040c7543-ea27-4d03-bc76-25f3c4c4d398" />

### Vorssaint Version

3.3.1

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

denmarin:
<img width="318" height="328" alt="Image" src="https://github.com/user-attachments/assets/f85d86f4-b695-49be-bfe2-de3a01519e32" /> 
when the fans are working it's showing their speed tho

albu-razvan:
Same issue on M1 Pro 14 inch (MacBookPro18,3), MacOS Version 27.0 Beta (26A5388g), Vorssaint v3.3.1. Curiously, having Macs Fan Control force a set fan speed changes the string's look as well (check the screenshots)

| With Macs Fan Control | Without |
|:---:|:---:|
| <img src="https://github.com/user-attachments/assets/4fc53a81-dc66-4c61-83fc-2a2d0b5966df" width="500"> | <img src="https://github.com/user-attachments/assets/409aef01-c1db-4892-a87c-94fd0dd7f29e" width="300"> |

denmarin:
ye, i've tried deleting the app fully using vorssaint's advanced settings and then downloading it again with homebrew, didn't help

albu-razvan:
Seems to work fine in v3.3.2, can probably be closed. @denmarin?

PR #988, opened 2026-08-25
fix(features): refuse an install the Mac has no hardware for

PathGao:
Builds on @Yahddyyp's work in #899 — the detection, the fail-soft policy and the greyed row are theirs, and the last round of that PR had already fixed the `FNum` spelling, the duplicate fan-count read and the suppressed settings link. This replaces it because the gate was on the wrong layer, not because the behaviour was wrong.

## The defect

`isHardwareSupported` was read by `FeatureHubRow` only. Availability is written from three places:

| Surface | Writes availability via | Gated in #899 |
|---|---|---|
| Hub row Install | `FeatureRuntime.setAvailable` | yes |
| Hub **Install All** | `FeatureRuntime.setAllAvailable` | no |
| First-run picker | `FeatureRuntime.replaceAvailable` | no |

The install-all button sits in the same view, roughly 200 lines above the row it walks around, and the first-run picker offers Fan Control on a fanless Mac with an ordinary tick box. Both then combine badly with the (correct) decision to leave installed rows alone: once Fan Control arrives by either route, the row looks normal forever and the user never learns why it does nothing.

## The change

- `AppFeature.hardwareUnsupportedReason` is the single switch. `isHardwareSupported` is derived from it, so a feature cannot be unsupported without a reason to show — the previous pair of parallel switches could drift into an empty tooltip and a VoiceOver label ending in `". "`.
- `AppFeature.installBlockedReason` adds the "not once it is installed" rule in one place, so the hub and the picker cannot disagree about it.
- `FeatureRuntime.mayFlip(_:to:)` is the gate. All three write paths route through it. Uninstalls are never refused and an existing install is never revoked.
- The picker greys and disables a blocked card, with the reason on the tooltip and the accessibility label, matching the hub row.
- `FeatureRuntime.installableCount` replaces `AppFeature.allCases.count` in the hub's tally and its install-all disabled condition. Without it that button is permanently enabled and does nothing on a Mac missing hardware — a bug the gate would otherwise have introduced.
- `replaceAvailable` now syncs `selected where feature.isAvailable`. A refused feature must not have its binding run, or the gate would block the install and start the service anyway.

Reuses `FeatureStrings.fanControl(…).noFans` ("This Mac has no controllable fan."), which is already translated into all 13 languages and is accurate for the strengthened check: `discoverControlledFans()` also rejects a Mac that reports fans it cannot drive.

## Alternatives not taken

- **Gate in the views.** Three views, three copies, and the next hardware-dependent feature adds three more. The write layer is one place and no future surface can miss it.
- **Auto-uninstall on a failed check.** Rejected in #899 and still rejected: `FNUM` made `hasControllableFan` false on every Mac, and that same code with auto-uninstall would have deleted everyone's Fan Control config on upgrade.
- **Hide unsupported features from the lists.** A greyed row with a reason tells the user the feature exists and why their Mac cannot have it; a missing row reads as a bug.
- **Move the check off the main thread.** `FanControlHardware.hasControllableFan` is a one-shot `static let` first touched while drawing a feature list. `FanControlService` uses `probeQueue` for `readOnlySnapshot()` because that repeats on a refresh timer and reads temperatures too; this is one SMC connection and ~6 key lookups per fan, once per process.

## Not changed

`Defaults.migrateFanControlVisibility` still restores a pre-existing opt-in on upgrade, and `FeaturePresets.prepareFirstRunAvailability` still seeds the essential preset directly. Both are correct: the first is exactly the install the gate refuses to revoke, and the second writes a preset that holds no hardware-dependent feature. Both are pinned by the test below.

## Testing

`./build.sh`, `./build/Vorssaint --selftest` and `./build.sh --test` all pass — TESTS OK (8553 checks), up from 8549.

New guardrails, each verified to go red when its rule is broken:

- Availability writes exist only in `FeatureRuntime.swift`, `FeaturePresets.swift` and `Defaults.swift`. A fourth install surface writing the key directly fails the build, which is precisely the class of bug this PR fixes. Sabotaged by adding a write to `MenuBarRenderer.swift`; test named the file.
- No `FeaturePreset` installs `.fanControl`, which is what makes the first-run seed safe.
- Both feature pickers read `installBlockedReason`. Sabotaged by removing it from `OnboardingView`; test failed.
- The hub counts against `installableCount`.

Hardware: on an M5 MacBook Air (Mac17,3) `hasControllableFan` is `false`, and the SMC client enumerates 2289 keys of which **none** begin with `F` — so this is a real absence, not a silent read failure like the `FNUM` spelling was. The positive case does not need a fan Mac to believe: `discoverControlledFans()` is the same call `readOnlySnapshot()` makes for every working Fan Control user today, so if it returned empty on fan hardware, Fan Control would already be broken for everyone.

## Related

- ref #899 — the PR this takes over. Its detection, fail-soft policy and greyed row are carried over unchanged; what moved is the layer the gate sits on. Left open for @Yahddyyp to close or repoint.
- ref #837 — tracks #899 in the open-PR index. Its remaining objection, the fan-specific reason string sitting in a row that renders every feature, is what `hardwareUnsupportedReason` resolves here.
- ref #574 — a MacBook Pro reporting "Fan control is not available on this Mac", fixed in v3.3.2. It is the case this gate must never catch: on that hardware `discoverControlledFans()` now succeeds, which is why the gate is backed by it rather than by a fan count or a model list.

### DIFF
--- a/Sources/Vorssaint/App/FeatureRuntime.swift
+++ b/Sources/Vorssaint/App/FeatureRuntime.swift
@@ -47,11 +47,34 @@ final class FeatureRuntime: ObservableObject {
 
     var availableCount: Int { AppFeature.allCases.filter(\.isAvailable).count }
 
+    /// How many features this Mac can end up with. Counting against the whole
+    /// catalog instead would leave the hub's install-all button forever one
+    /// short of its own disabled condition on a Mac missing some hardware.
+    /// An install that predates the check still counts, so the tally can
+    /// never read more installed than installable.
+    var installableCount: Int {
+        AppFeature.allCases.filter { $0.isHardwareSupported || $0.isAvailable }.count
+    }
+
+    /// The one gate every install passes, whichever surface asks: the hub
+    /// row, the hub's install-all button, a preset and the first-run picker
+    /// all decide here. A feature the Mac cannot run never installs, so a
+    /// disabled row cannot be walked around from the button above it.
+    ///
+    /// Uninstalls are never refused and an existing install is never revoked:
+    /// the check reads hardware and can be wrong, and a wrong answer that
+    /// strands someone's settings costs far more than one that leaves a
+    /// feature reporting itself unsupported.
+    private func mayFlip(_ feature: AppFeature, to available: Bool) -> Bool {
+        guard feature.isAvailable != available else { return false }
+        return !available || feature.isHardwareSupported
+    }
+
     /// Flipping availability runs the feature's binding immediately: off
     /// tears every resource down on the spot, on restores whatever enabled
     /// state the feature had (its own keys are never touched).
     func setAvailable(_ feature: AppFeature, _ available: Bool) {
-        guard feature.isAvailable != available else { return }
+        guard mayFlip(feature, to: available) else { return }
         UserDefaults.standard.set(available, forKey: feature.availabilityKey)
         if available { loadedThisSession.insert(feature) }
         Self.bindings[feature]?()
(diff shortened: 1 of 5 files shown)


Yahddyyp:
Hope this actually gets merged

PathGao:
It's merged. #899 is still open — close it or repoint it at whatever you want to keep pushing; the detection, the fail-soft policy and the greyed row in it are yours and went in unchanged.

Issue #189, opened 2026-07-08
[Bug]: Glitch when entering the Keychain password


M-Haertel:
### Description

I often work with apps I've programmed myself, so I frequently have to enter my password into the Keychain because the apps aren't signed. When Vorssaint is running, it's a disaster. I usually can't enter my password. My password starts with a capital letter, and when I press Shift and then a key, nothing happens at first. I have to try it several times, and eventually it works. The whole process just freezes for several seconds every time. When I quit Vorssaint, the problem goes away and everything is back to normal. Can anyone relate to this? 
Thank you very much 

### Steps to Reproduce

1. When Vorssaint is open
2. Then open an unsigned app that requests access to the keychain
3. A dialog box opens, and the buttons appear to be locked for a few seconds

### Screenshots / Recordings

<img width="469" height="237" alt="Image" src="https://github.com/user-attachments/assets/29200074-8b4b-435a-ba89-394841a9ccfe" />

It's all about this dialogue

### Vorssaint Version

3.1.8

### macOS Version

macOS 27 (Golden Gate)

### Installation Method

Github

vorssaint:
I found the cause and published a fix in 3.1.9. Some features were reading the clipboard and querying other apps on the main thread, and while a password prompt was open those calls stalled, which froze typing for everyone. That work now happens off the main thread with strict timeouts. Please update to the latest version and let me know if you still see the freeze.

M-Haertel:
I've tested it several times, and it's working perfectly now. Thank you very much for your quick help.

PR #989, opened 2026-08-25
fix(monitor): bound the waits that freeze the app

PathGao:
Two waits combine into the freeze in #971. Either one alone is survivable.

This consolidates #980, which found the same two defects independently and reached the same fix; its deadline value and the reasoning for it are carried over here, and @iltonandrew is credited on the commit.

**The runner.** `BoundedProcessRunner.run` parks `waitUntilExit()` on the shared global dispatch pool and walks away from it when the timeout fires, still holding a worker. Nine call sites reach it, five of them periodic samplers, so abandoned waits can arrive faster than they drain. A full pool then starves the runner itself: it reports timeouts for commands that already exited, and abandons one more wait doing it.

**The window walk.** `WindowEnumerator.accessibilityWindows(for:)` ran `waitUntilAllOperationsAreFinished()` with no deadline, on the main thread. `DockPreviewService` reaches it through `DispatchQueue.main.asyncAfter`; `AppSwitcher` reaches it through the `DispatchQueue.main.async` at `AppSwitcher.swift:536`, which answers the open question in #971 about whether the switcher is a second way in — it is. Those operations run on the same pool, so when the pool has nothing left, none of them start and the wait never returns.

**What changed.** The child is watched through `process.terminationHandler`, which occupies no thread while it waits, so an abandoned wait costs a semaphore and nothing else. The Accessibility batch gets a 1.5s ceiling: operations still queued when it runs out are cancelled and the answers already collected are returned.

**Effect.** Both implementations built standalone and driven by the reproduction in the issue:

- With 64 workers blocked on the utility pool, `/bin/echo hi` on a 1s timeout: before, a false timeout after 2.016s; after, status 0 in 0.001s.
- 350 timed-out runs across 70 concurrent callers: 66 threads at peak before, 10 after. File descriptors flat and no unreaped children either way.

Being precise about what that does and does not show: the saturation and the false timeouts reproduce deterministically, and the thread peak drains on its own in both versions. The permanently stranded waiters in the issue's sample — 64 of them, identical 20 minutes apart, every child already reaped — did not reproduce locally. The fix does not rest on that question: a wait that never occupies a pool thread cannot strand one, however it ends.

**Alternatives considered.** For the runner, counting the abandoned waiters and capping or reclaiming them keeps the blocking wait and adds bookkeeping plus a ceiling to tune; removing the thread from the wait removes the accounting problem instead of managing it. For the batch, moving the walk off the main thread is a far larger change to callers that need the result synchronously, and it would leave the unbounded wait in place on a different thread. A deadline bounds it where it is. The truncation a deadline can cause lands on an existing path: an app missing from that map reads downstream exactly like an app that could not answer Accessibility.

The ceiling is 5s, taken from @iltonandrew's #980. A healthy batch takes hundredths of a second, so a much tighter ceiling is tempting, but the individual AX calls are already bounded by their messaging timeouts: the only thing that can outlast 5s is a pool with no thread to give, where no answer was ever coming. A tighter ceiling would buy nothing there and would truncate a genuinely slow batch on a busy Mac.

**Not changed.** The `nettop` fallback in `NetworkSampler` / `NetworkCounterFallback`. On the reporter's machine every inbound byte lands on a `utun` interface, which `MetricFormat.includeNetworkInterface` excludes, so the latch holds forever and `nettop` runs on every tick. That is the load that exposed both waits, but the fallback is doing its job there — it is what gives that machine correct download figures. Changing the sample rate is an accuracy trade-off that belongs in its own issue. The other `waitUntilExit()` callers (`HomebrewManager`, `DiskImageInstallerService`, `JunkCleaner`) are also untouched: none of them abandon their wait, and none run it on the shared pool.

**Tests.** 8544 checks. Two added, both verified red on the unmodified sources: the pool-starvation behaviour above, which also asserts that the starvation it needs was actually reached so it cannot pass vacuously; and an absence proposition over all of `Sources/Vorssaint`, since `waitUntilAllOperationsAreFinished` has no deadline and is unusable on these paths — the rule is the absence of the call rather than the one caller found holding it.

#887 reports the app going unresponsive at random with the command bar steps not reproducing it, which #971 reads as closer to this than to the pasteboard lane stall #903 addresses. Unconfirmed either way, so it is listed as a candidate below rather than claimed. #189 is the event-tap stall that the main-thread ceiling cites.

ref #971
ref #887

### DIFF
--- a/Sources/Vorssaint/Services/BoundedProcessRunner.swift
+++ b/Sources/Vorssaint/Services/BoundedProcessRunner.swift
@@ -59,18 +59,24 @@ enum BoundedProcessRunner {
             }
         }
 
+        // The child is watched through its termination handler. A blocking
+        // `waitUntilExit()` has to be parked on a thread of its own, and the
+        // timeout below makes `run` walk away from it while it still holds one
+        // worker of the shared 64-thread pool. Those abandoned waits pile up
+        // faster than they drain, and a full pool starves every later user of
+        // it, the main thread's window walk included (issue #971). It also
+        // starves this runner, which then reports timeouts for commands that
+        // exited at once and abandons another wait doing it. A termination
+        // handler occupies no thread, so none of that accumulates.
+        let finished = DispatchSemaphore(value: 0)
+        process.terminationHandler = { _ in finished.signal() }
+
         do {
             try process.run()
         } catch {
             reader.readabilityHandler = nil
             try? reader.close()
             return Result(status: -1, output: Data(), timedOut: false)
-        }
-
-        let finished = DispatchSemaphore(value: 0)
-        DispatchQueue.global(qos: .utility).async {
-            process.waitUntilExit()
-            finished.signal()
         }
 
         var didFinish = finished.wait(timeout: .now() + max(0, timeout)) == .success
--- a/Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
+++ b/Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
@@ -29,6 +29,12 @@ enum WindowEnumerator {
     /// tree within the process thread budget while collapsing it into a few
     /// bounded AX batches.
     private static let maximumConcurrentQueries = 24
+    /// Ceiling on the whole batch, which the main thread waits out. Generous
+    /// against the messaging timeouts above even with every query as slow as
+    /// they allow, so only a dispatch pool with no thread to give can outlast
+    /// it (issue #971) — and then no answer was ever coming. A stalled main
+    /// thread stalls the event taps with it (issue #189), so it must expire.
+    private static let accessibilityBatchBudget: TimeInterval = 5.0
 
     static func listWindows(groupByApp: Bool = UserDefaults.standard.bool(forKey: DefaultsKey.switcherMergeTabs),
                             preservingGroupedWindows: Bool = false) -> [SwitcherItem] {
@@ -423,7 +429,8 @@ enum WindowEnumerator {
         guard !orderedPIDs.isEmpty else { return [:] }
         let screenFrames = NSScreen.screens.map(\.frame)
         var result: [pid_t: AccessibilityWindowSnapshotList] = [:]
-        let resultLock = NSLock()
+        var pendingQueries = orderedPIDs.count
+        let resultLock = NSCondition()
         // A remote app can consume its whole messaging timeout. Overlap those
         // independent calls so several slow background helpers cost one wait,
         // not one wait each, while preserving Accessibility-only windows. The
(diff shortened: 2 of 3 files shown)


iltonandrew:
Heads-up on the starvation check from this PR — it fails deterministically on a 15-core M5, which leaves `main` red:

```
TESTS FAILED (1 of 8577):
```

Reproduced on a clean `main` checkout with nothing of mine applied. It is the precondition assertion, `expect(occupiedWorkers == 64 && poolIsStarved, ...)`, not the behavioural one. A standalone harness replicating the setup exactly:

```
activeProcessorCount=15
occupiedWorkers=64   (wants 64)    ✓
poolIsStarved=false  (wants true)  ✗
```

Blocking 64 workers doesn't starve the pool on this hardware — 64 is the soft limit libdispatch *reports* at (the "Dispatch Thread Soft Limit: 64 reached" line in the #971 sample), not a cap it enforces, so it hands out a 65th thread and the probe returns straight away. The assertion your fix actually rests on — `a subprocess is watched off the dispatch pool, so a starved pool cannot strand it` — passes; it is only the setup's self-check that fails.

Two ways out, whichever you prefer: grow the blocking loop until the probe genuinely times out (with a ceiling so it can't run away), or treat an unstarvable pool as a skip rather than a failure, since on such a machine the scenario the check wants to construct doesn't exist. I'd rather not touch a merged test of yours without asking — say which and I'll send it, or take it yourself if you'd rather.

PathGao:
Take it yourself. `main` being red beats waiting for me to get to it.

Skip, not grow. Growing the loop doesn't eliminate the case you hit, it relocates it: with a ceiling, a machine whose real limit sits above that ceiling lands in exactly the same place, having spent a few hundred blocked threads to get there. The ceiling branch has to be a skip either way, so go straight to the skip and leave the loop at 64.

But a plain skip makes the check cover nothing on your machine, which is the thing the precondition assertion existed to prevent. So pair it with an absence rule of the same shape as the `waitUntilAllOperationsAreFinished` one in this PR — `BoundedProcessRunner.swift` must not contain `DispatchQueue.global`. That is the invariant the fix actually rests on (the wait must not go on the shared pool), it holds on every machine, and it turns the starvation check into a stronger conditional version of it rather than the only line of defence:

```swift
expect(occupiedWorkers == 64, "the dispatch pool's blocking workers were all taken")
if poolIsStarved {
    expect(!starvedPoolProcess.timedOut && starvedPoolProcess.status == 0
            && String(decoding: starvedPoolProcess.output, as: UTF8.self) == "ready\n"
            && starvedPoolElapsed < 0.5,
           "a subprocess is watched off the dispatch pool, so a starved pool cannot strand it")
}
```

Two things I haven't checked. I don't have that hardware, so I'm taking your standalone harness at face value — the reading that 64 is where libdispatch reports rather than where it stops is consistent with what you measured and I can't test it. And I only see this diff, not the whole file, so if `BoundedProcessRunner.swift` uses `DispatchQueue.global` somewhere outside these hunks the absence rule needs narrowing to the wait rather than the file.

iltonandrew:
Took it — #1014.

Went with skip, and your reasoning for it over growing the loop is right: a ceiling just relocates the same case to a machine whose real limit sits above it, having spent a few hundred blocked threads to get there, and the ceiling branch has to be a skip anyway.

Paired with the absence rule as you described, so the skip doesn't leave the check covering nothing: `BoundedProcessRunner.swift` must not contain `DispatchQueue.global`, checked with the same file-scan shape as the `waitUntilAllOperationsAreFinished` rule in this PR. On your caveat about narrowing it — not needed: the file contains no `DispatchQueue.global` at all, inside those hunks or outside them, so the file-level rule is exactly the invariant with no false surface.

`occupiedWorkers == 64` still asserts unconditionally, so a machine that fails to take the workers at all still fails rather than silently skipping.

Result on the machine that was red: **TESTS OK (8582 checks)**, build zero warnings, SELFTEST OK. And the absence rule bites — a bare `_ = DispatchQueue.global(qos: .utility)` added to the runner fails it by name.

Fair on the hardware point; I can only offer that the harness is a faithful copy of your setup and the result is deterministic across runs on a 15-core M5, with `occupiedWorkers` hitting 64 every time and the extra submission still running immediately.

Issue #940, opened 2026-08-25
Ignore parked status-item frames for Shelf placement and icon hit testing


pergioa:
## Problem

When macOS hides the menu bar in fullscreen, a status item window can report a parked fallback frame that no longer represents the visible icon. Two callers currently trust that raw frame:

- `AppDelegate` supplies `button.window.frame` through `ShelfService.statusItemFrameProvider`, so `positionDocked` and `mouseNearDock` can place the Shelf pill and hover band at the parked coordinate.
- `StatusItemController.containsStatusItem(at:)` reads the same frame directly, so the popover outside-click monitor can stop recognizing a click on the status icon in this state.

This was identified as follow-up work during review of #930 and is intentionally separate from the popover anchoring change.

## Proposed direction

Gate raw status-item frames through `StatusItemAnchorSupport.isTrustworthyStatusFrame`. Return `nil` to the Shelf provider when the frame is untrustworthy so its existing fallback placement is used, and apply the same trust rule to status-icon hit testing.

## Acceptance criteria

- A hidden fullscreen menu bar does not send the docked Shelf or its hover band to the parked status-item coordinate.
- Clicking the visible status icon is still recognized correctly while its reported window frame is untrustworthy.
- Normal visible-menu-bar Shelf placement and hit testing are unchanged.
- Add focused coverage for trustworthy and parked-frame behavior where practical.

iltonandrew:
Taking a look at this one, following the acceptance criteria in the description.

pergioa:
Tested on 3.3.3, this has all been fixed and the issue can be closed

vorssaint:
closing this following your 3.3.3 test confirming that both the shelf placement and icon clicks are fixed. thanks for checking the whole report

PR #990, opened 2026-08-25
fix(shelf): ignore parked status-item frames for Shelf placement and icon hit testing

iltonandrew:
Refs #940 — the follow-up split out of #930's review, implemented per the direction in the issue.

## What was wrong

When macOS hides the menu bar in fullscreen, a status item window can report a parked fallback frame that no longer describes the visible icon, and two callers trusted it raw:

- `AppDelegate` fed `button.window.frame` straight through `ShelfService.statusItemFrameProvider`, so `positionDocked` and `mouseNearDock` could place the Shelf pill and hover band at the parked coordinate.
- `StatusItemController.containsStatusItem(at:)` read the same frame directly, so the popover's outside-click monitor could stop recognizing a click on the status icon in that state.

## The change

Both call sites now gate the raw frame through `StatusItemAnchorSupport.isTrustworthyStatusFrame` — the trust rule the popover anchoring already uses (#930): positive size, intersects an attached screen, midpoint inside the menu-bar band. The Shelf provider returns nil on an untrustworthy frame, which engages its existing fallback (pointer-screen placement, hover band disabled); hit testing declines instead of testing a coordinate that describes nothing. Trustworthy frames behave exactly as before — the guard only rejects frames the trust rule already classifies as parked.

Against the acceptance criteria: a hidden fullscreen menu bar no longer sends the docked Shelf or its hover band to the parked coordinate (provider returns nil); clicking the visible icon still works when the menu bar is visible, including the fullscreen reveal-on-hover state — a revealed status window sits inside the menu-bar band, so the trust rule passes there by construction; trustworthy frames take the unchanged code path.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8551 checks), `./build/Vorssaint --selftest` → SELFTEST OK. Two source-shape assertions pin the wiring (the Shelf provider guards through the trust rule and returns nil; hit testing consults the same rule); the trust rule itself keeps its existing direct geometry coverage.

### DIFF
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -96,7 +96,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         ShelfService.shared.statusItemFrameProvider = { [weak self] in
             guard let item = self?.statusController.statusItem, item.isVisible,
                   let window = self?.statusController.button?.window else { return nil }
-            return window.frame
+            let frame = window.frame
+            guard StatusItemAnchorSupport.isTrustworthyStatusFrame(
+                frame, screenFrames: NSScreen.screens.map(\.frame)) else { return nil }
+            return frame
         }
 
         setUpPopover()
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -97,8 +97,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
             guard let item = self?.statusController.statusItem, item.isVisible,
                   let window = self?.statusController.button?.window else { return nil }
             let frame = window.frame
-            guard StatusItemAnchorSupport.isTrustworthyStatusFrame(
-                frame, screenFrames: NSScreen.screens.map(\.frame)) else { return nil }
+            guard StatusItemAnchorSupport.isTrustworthyStatusFrame(frame) else { return nil }
             return frame
         }
 
@@ -531,8 +530,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
     /// itself parks the window out of the visible area) the anchor takes over.
     private func frameStillDescribesMenuBar(_ anchor: PanelAnchor) -> Bool {
         guard let frame = anchor.button?.window?.frame else { return false }
-        return StatusItemAnchorSupport.isTrustworthyStatusFrame(
-            frame, screenFrames: NSScreen.screens.map(\.frame))
+        return StatusItemAnchorSupport.isTrustworthyStatusFrame(frame)
     }
 
     /// The spot the panel must hold while it is open, decided at the moment it
@@ -544,7 +542,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         let screen = statusScreen(for: button)
         let statusFrame = button.window?.frame
         let frameIsSound = statusFrame.map {
-            StatusItemAnchorSupport.isTrustworthyStatusFrame($0, screenFrames: NSScreen.screens.map(\.frame))
+            StatusItemAnchorSupport.isTrustworthyStatusFrame($0)
         } ?? false
         if frameIsSound, statusFrame != nil {
             // Where the popover has just been placed is the anchor: with a
--- a/Sources/Vorssaint/App/StatusItemAnchorSupport.swift
+++ b/Sources/Vorssaint/App/StatusItemAnchorSupport.swift
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // Copyright (C) 2026 Vorssaint
 
+import AppKit
 import CoreGraphics
 import Foundation
 
@@ -38,6 +39,10 @@ enum StatusItemAnchorSupport {
     /// How far down from the top of a screen a status item's window can sit and
     /// still describe a menu bar. Clears the taller bar drawn around a camera
     /// housing with room to spare, while staying far from any ordinary window.
+    /// Measured off the screen's own top edge, never off the area left over
+    /// after the bar: a fullscreen screen reserves nothing, and a band derived
+    /// from the visible area would collapse there and reject a status item
+    /// revealed on hover exactly while it is on screen and clickable.
     static let menuBarBand: CGFloat = 48
 
     /// Breathing room kept between the panel and the edges of the usable area.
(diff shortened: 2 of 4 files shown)


PathGao:
Two findings; the first decides whether the fullscreen case works at all.

**1. The band the trust rule uses in fullscreen — the body's "by construction" claim isn't tested.**

Body: "clicking the visible icon still works when the menu bar is visible, including the fullscreen reveal-on-hover state — a revealed status window sits inside the menu-bar band, so the trust rule passes there by construction."

That holds only if the band survives fullscreen. If it's derived from `screen.frame.maxY - screen.visibleFrame.maxY`, a fullscreen screen reserves no menu bar, the band collapses to zero height, and a revealed status window's midpoint falls outside it. The rule then rejects the frame at exactly the moment the icon is visible and clickable: `containsStatusItem(at:)` declines the click, and the Shelf provider returns nil and drops to the pointer-screen fallback. That's the inverse of the bug being fixed.

`StatusItemAnchorSupport` isn't in this diff, so I can't tell which derivation it is — this is a question, not a defect claim. If the band is a fixed `NSStatusBar.system.thickness` off `screen.frame.maxY`, it's fine as written. Either way the case wants one direct geometry assertion rather than the "by construction" argument: fullscreen screen frame, revealed status frame, expect trustworthy. The two new assertions are source-text greps and can't reach it.

**2. `NSScreen.screens.map(\.frame)` is now the third copy of the same fact.**

Both new call sites compute it, and #930's anchoring site presumably does too. Laziest fix is to stop making callers know: default the parameter.

```swift
static func isTrustworthyStatusFrame(_ frame: CGRect,
                                     screenFrames: [CGRect] = NSScreen.screens.map(\.frame)) -> Bool
```

Call sites become `guard StatusItemAnchorSupport.isTrustworthyStatusFrame(frame) else ...`, tests keep injecting fake screens. It also hoists the allocation out of the `buttons.contains` closure, where it currently runs once per status item per click.

One note, not a finding: the hit-test slice in the new test uses `"\n    init()"` as its end boundary. `init()` isn't in the diff, so I can't tell whether it appears after `containsStatusItem`. If it doesn't, the slice runs to EOF and `contains("return false")` can be satisfied by an unrelated function.

**Impact:** if the band comes from `visibleFrame`, fullscreen users lose status-icon clicks and docked Shelf placement while the menu bar is revealed. If it's a fixed thickness, none.
**Merge:** needs the band derivation confirmed, plus a direct geometry assertion for the fullscreen-revealed frame, first.

ref #940

iltonandrew:
Thanks — worked through all three.

**1. Menu-bar band in fullscreen.** Good instinct to check, but the band is not visibleFrame-derived. `isTrustworthyStatusFrame` takes `[CGRect]` screen *frames* and never sees `visibleFrame`; the band is `static let menuBarBand: CGFloat = 48`, applied as `frame.midY >= screen.maxY - band` off the screen's own top edge. So this lands in your "fine as written" branch — fullscreen cannot collapse it, and the reveal-on-hover case passes as the body claims.

I have made that non-negotiable rather than incidental: there is now a doc comment on `menuBarBand` recording that it must be measured off `screen.frame.maxY` and why, plus a direct geometry assertion next to the existing ones — a fullscreen screen (`visibleFrame == frame`, nothing reserved) with a revealed item at the top, expecting trustworthy. To confirm it bites I temporarily fed it `band: screen.maxY - visibleFrame.maxY` (which is 0 in fullscreen) and the suite failed on exactly that check; reverted, green.

**2. Threefold `NSScreen.screens.map(\.frame)`.** Done as you suggested — `screenFrames: [CGRect] = NSScreen.screens.map(\.frame)`. All four call sites are now `isTrustworthyStatusFrame(frame)` (the two new ones plus both of #930's), and the allocation is out of the `buttons.contains` closure, so it no longer runs per status item per click. On the test-binary question: `StatusItemAnchorSupport.swift` is in `build.sh`'s test source list, but so is `MixerPercentNativeTextField.swift`, which already imports AppKit — the test binary already links it, so no overload was needed. Default arguments only evaluate when omitted and the tests always inject fake screens, so `NSScreen.screens` never runs headless.

**3. Slice boundary.** Your premise was nearly right: `init()` does happen to follow `containsStatusItem` today, so the slice was correct but fragile exactly as you describe. Both source-shape slices now cut at the enclosing body's own closing brace — and the sibling Shelf-provider slice had the same latent bug (it was cutting on `setUpPopover()`), so that one is fixed too. Neither can run past its own body now.

Gate: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8553 checks), `--selftest` → SELFTEST OK. Net production code is 3 lines shorter.

PathGao:
Band derivation answers it — a fixed 48pt off `screen.frame.maxY` can't collapse in fullscreen, and the new assertion bites: midY 937.5 on a 956-tall screen clears the 908 threshold. Both slice boundaries check out too; the inner closure braces sit at deeper indent than the cut strings, so neither slice can escape its own body.

One correction, mine as much as yours: the default argument does not hoist the allocation. Default expressions are evaluated at each call, and the call is still inside `buttons.contains { }`, so `NSScreen.screens.map(\.frame)` runs once per status item per click exactly as it would have. Bind it once above the closure and pass it explicitly if that matters. Not blocking — the button count is small and this is a click path, not a hot loop.

**Impact:** none.
**Merge:** ship it.

iltonandrew:
You're right, and thanks for the correction — a default argument is evaluated at each call, so it was still rebuilding the list once per status item per click. My claim that it hoisted the allocation was wrong. Bound it once above the closure and passed it explicitly, with a comment saying why, so the claim and the code now agree.

Also rebased onto current main (the branch had gone behind).

One thing to flag while re-gating, unrelated to this PR: **`main` is currently red on this machine.**

```
TESTS FAILED (1 of 8577):
```

That reproduces on a clean `main` checkout with no changes of mine, so it is not from this branch. It is the *precondition* assertion in the #989 starvation check, not the behavioural one — `expect(occupiedWorkers == 64 && poolIsStarved, ...)`. Isolating it with a standalone harness that replicates the setup exactly, on a 15-core M5:

```
activeProcessorCount=15
occupiedWorkers=64   (test wants 64)     ✓
poolIsStarved=false  (test wants true)   ✗
```

Deterministic across runs. Blocking 64 workers does not starve the pool here: 64 is the soft limit libdispatch *reports* at, not a hard cap, so it hands out a 65th thread and the probe returns immediately. The assertion the fix actually rests on — `a subprocess is watched off the dispatch pool` — passes; only the setup's self-check fails.

So the check is sound in intent but its premise only holds on machines where the pool refuses to grow past 64. Worth either making the saturation loop grow until the probe actually times out (with a ceiling), or treating an unstarvable pool as a skip rather than a failure — as it stands it makes `main` red for anyone on a machine like this one, which will get misattributed to whatever branch they happen to be on. Happy to send that as its own PR if you want it.

PathGao:
Bound-once matches the claim now.

On the #989 precondition: separate PR, yes — don't fold it into this one.

Skip-when-unstarvable is the wrong default. It buys a green run but hides the case where the pool *should* starve and doesn't, so nobody learns the behavioural check went vacuous. Let the probe decide instead of a count: block workers in a loop until the probe actually times out, ceiling the loop so it can't spawn unbounded threads, and only then treat it as unstarvable and skip. `64` stops being a machine-specific constant, and a box that hands out a 200th thread either gets real coverage or an explicit skip, not a red bar.

I haven't read #989's test, so I can't tell how invasive that is — if the saturation is written inline in the check rather than behind a helper, the skip is the cheaper stopgap and I'd take it.

One consequence for this branch: rebased onto a red `main`, `./build.sh --test` reports TESTS FAILED for anyone who runs it here too, so the `TESTS OK` in the body no longer reproduces. Pre-existing and not from this diff, but whoever merges should expect the red rather than chase it.

**Impact:** none.
**Merge:** ship it.

Issue #986, opened 2026-08-25
uninstall.sh leaves the Fan Control daemon registered and the app's data folder behind


mugurc:
### Description

`Tools/uninstall.sh` ends with `✓ Vorssaint fully removed.` (line 45) and the README calls it the way to remove Vorssaint completely, "deletes every trace" (README line 222). Two things survive it.

**1. The Fan Control LaunchDaemon stays registered.**

The script hands the system detach to the app (`uninstall.sh:25`, `"$candidate" --uninstall`). The whole body of that flag is `Support/Uninstaller.swift:15-25`, and the only registration it drops is:

```swift
try SMAppService.mainApp.unregister()
```

Per `SMAppService.h`, `mainAppService` is "a SMAppService corresponding to the main application **as a LoginItem**", and `unregisterAndReturnError` is the inverse of `register` on that same receiver — for the main app it is scoped to preventing "future launches at login". The Fan Control helper is a different service, built by `SMAppService.daemon(plistName:)` (`FanControlService.swift:36`), so it is untouched.

The tree already has the right call: `FanControlService.restoreAndUnregisterForRemoval()`. Its only caller anywhere is `SelfUninstall.swift:103`, the in-app "uninstall Vorssaint" flow. The script path never reaches it. The `pkill` at the top of the script does not cover it either — `applicationWillTerminate` (`AppDelegate.swift:240`) calls `restoreBeforeTerminationIfNeeded()`, which restores automatic fan control (`FanControlService.swift:209-224`) but never unregisters. And `--uninstall` returns through `exit(0)` in `main.swift:14` before `AppDelegate` is ever constructed, so nothing in `applicationWillTerminate` runs on that path at all.

So after `rm -rf "$APP"` the job is still submitted in the system domain, pointing into a bundle that no longer exists:

```
$ launchctl print system/com.vorssaint.utils.fan-control
system/com.vorssaint.utils.fan-control = {
	type = Submitted
	managed_by = com.apple.xpc.ServiceManagement
	program identifier = Contents/Library/LaunchServices/com.vorssaint.utils.fan-control (mode: 2)
	parent bundle identifier = com.vorssaint.utils
	BTM uuid = <redacted>
}
```

I have not measured whether macOS eventually garbage-collects that record on its own. The point is that the script leaves it to undefined system behaviour when the correct API call already exists one line away.

**2. The whole data folder stays on disk.**

[...]

PR #991, opened 2026-08-25
fix(uninstall): remove the fan helper registration and app data on the script path

mugurc:
## Summary

`Tools/uninstall.sh` prints `✓ Vorssaint fully removed.` and the README calls it the way to remove Vorssaint completely, but two things outlived it.

- **The Fan Control daemon stayed registered.** The script hands the system detach to `--uninstall`, and that path only dropped the app as a login item. The helper is a separate service built by `SMAppService.daemon(plistName:)`, and `unregister` acts on its own receiver, so nothing reached it — its root registration outlived the bundle that carried its executable. `Uninstaller.runAndExit()` now makes the same `FanControlService.restoreAndUnregisterForRemoval()` call the in-app uninstall already made.
- **The data folder stayed on disk.** The script removed the app, the preferences plist and the saved state, but never `~/Library/Application Support/<id>` or `~/Library/Caches/<id>` — clipboard history and its images, the shelf store, recorder takes, the scratchpad, and the share records with their delete tokens. It removes both now.

Both are one divergence: `SelfUninstall` does the complete job and the script path was a subset that had drifted from it.

**On "fix it entirely".** This covers the bug entirely — both halves, on both paths, with nothing left for a follow-up. What I did *not* do is collapse the duplication I raised as the wider option on the issue, and that was a deliberate call rather than a shortcut. Two reasons:

- The script's own `rm -rf` lines are a fallback, not just duplication. If the binary cannot run — quarantined, signature broken, bundle half-deleted — the script still clears the preferences and now the data folder. Delegating all removal to `--uninstall` would mean a binary that refuses to launch leaves everything behind, which is the case where a user most needs the script to work.
- `--uninstall` does not destroy user data today. Making it the single place that defines removal would give that flag the power to delete clipboard history, and anyone who runs it by hand would lose data they did not ask to lose.

So the duplication that stays is the deliberate kind, and this change makes the two paths agree on what has to go. If you did mean the structural version, say so and I will send it as a follow-up on top of this one — but I would rather not fold it in here, since it changes what a flag does as well as fixing a bug.

**Other callers I checked.** `restoreAndUnregisterForRemoval()` had exactly one caller in the tree, `SelfUninstall.swift:103`. `--uninstall` is the only other path that has to detach from the system, and `main.swift` exits before `AppDelegate` is constructed, so nothing in `applicationWillTerminate` covers it. `restoreBeforeTerminationIfNeeded()` on the ordinary quit path restores automatic fan control and deliberately does not unregister; I left it alone. Every place the app writes under `~/Library` is `Application Support/<id>` or `Caches/<id>`, and there is no legacy bundle identifier to sweep — the pre-rename bundle kept the same one.

The call is safe on this path: it never touches `FanControlService.shared`, uses its own XPC connection with a 20 s timeout, returns early when the service is not registered, and the script already invokes the binary from inside the bundle, so `SMAppService` resolves against the right one.

## Verification

Mac17,6 (M5 Max), macOS 26.5.1 (25F80), Command Line Tools Swift 6.3.3, my own release build. App language follows the system, tr-TR. Single built-in display.

- `./build.sh` — finished with no warnings
- `./build/Vorssaint --selftest` — passed
- `./build.sh --test` — passed
- `zsh -n Tools/uninstall.sh` — valid
- Ran the script's removal block against a throwaway `$HOME` seeded with the paths the app writes. Before the change, the plist and the `.savedState` went and every data file stayed; after it, nothing was left.
- Confirmed the daemon really is registered on this Mac beforehand with `launchctl print system/com.vorssaint.utils.fan-control`, which needs no `sudo`: it prints a job submitted by ServiceManagement whose program identifier points inside the app bundle.

**What I could not test.** The end-to-end run. Verifying the daemon half for real means uninstalling my own working install and re-approving the helper in System Settings afterwards, and I did not want to do that on the machine I use daily. So the unregister is verified as reachable and safe on this path, not observed removing a live registration — worth one run on your side before merging. I also did not measure whether macOS eventually garbage-collects the orphaned record on its own; the argument here is that the correct call already exists and the script should not depend on that either way.

## Anything else

Refs #986.

Scope note: I used `uninstall` rather than the existing `uninstaller`, since that scope has meant the feature that removes *other* apps (#752) and this is the app's own removal. Rename it on squash if you would rather keep one scope.

### DIFF
--- a/Sources/Vorssaint/Services/FanControl/FanControlService.swift
+++ b/Sources/Vorssaint/Services/FanControl/FanControlService.swift
@@ -226,14 +226,20 @@ final class FanControlService: ObservableObject {
     /// Used by the complete-uninstall path from its background queue. The
     /// daemon is removed only after it confirms automatic control, so teardown
     /// can never kill the recovery mechanism while a manual session remains.
-    static func restoreAndUnregisterForRemoval() {
+    ///
+    /// Reports whether the daemon is actually gone. A caller that tells someone
+    /// the app was fully removed has no other way to know: the registration
+    /// outlives the bundle, so a silent failure here reads as success forever.
+    @discardableResult
+    static func restoreAndUnregisterForRemoval() -> Bool {
         let service = appService
         guard service.status == .enabled else {
-            if !UserDefaults.standard.bool(forKey: DefaultsKey.fanControlRecoveryNeeded),
-               service.status != .notRegistered {
-                try? service.unregister()
-            }
-            return
+            guard service.status != .notRegistered else { return true }
+            // A pending recovery keeps the daemon deliberately: it is the only
+            // thing that can put the fans back. Still not a clean detach.
+            guard !UserDefaults.standard.bool(forKey: DefaultsKey.fanControlRecoveryNeeded)
+            else { return false }
+            return unregisterForRemoval(service)
         }
         let connection = NSXPCConnection(machServiceName: FanControlIdentifiers.helperID,
                                          options: .privileged)
@@ -247,7 +253,7 @@ final class FanControlService: ObservableObject {
             as? FanControlXPCProtocol
         guard let proxy else {
             connection.invalidate()
-            return
+            return false
         }
         proxy.restoreAutomatic { data in
             if let response = FanControlIPC.decode(data) {
@@ -259,7 +265,17 @@ final class FanControlService: ObservableObject {
         }
         _ = semaphore.wait(timeout: .now() + 20)
         connection.invalidate()
-        if resultLock.withLock({ restored }) { try? service.unregister() }
+        guard resultLock.withLock({ restored }) else { return false }
+        return unregisterForRemoval(service)
+    }
+
+    private static func unregisterForRemoval(_ service: SMAppService) -> Bool {
+        do {
+            try service.unregister()
+            return true
+        } catch {
+            return false
+        }
     }
 
     // MARK: - Requests
--- a/Sources/Vorssaint/Services/SelfUninstall.swift
+++ b/Sources/Vorssaint/Services/SelfUninstall.swift
@@ -132,6 +132,12 @@ enum SelfUninstall {
         try? FileManager.default.removeItem(atPath: "\(home)/Library/Saved Application State/\(id).savedState")
         // Clipboard images and any other app-owned data live here.
         try? FileManager.default.removeItem(atPath: "\(home)/Library/Application Support/\(id)")
+        try? FileManager.default.removeItem(atPath: "\(home)/Library/Caches/\(id)")
+        // URLSession writes these on our behalf whenever the app talks to the
+        // network, so they exist without the app ever choosing the path.
+        try? FileManager.default.removeItem(atPath: "\(home)/Library/HTTPStorages/\(id)")
+        try? FileManager.default.removeItem(
+            atPath: "\(home)/Library/HTTPStorages/\(id).binarycookies")
     }
 
     /// Moves the app's own bundle to the Trash after it quits, then quits. The
--- a/Sources/Vorssaint/Support/Uninstaller.swift
+++ b/Sources/Vorssaint/Support/Uninstaller.swift
@@ -16,7 +16,10 @@ enum Uninstaller {
         // The fan helper is a daemon service of its own, so unregistering the
         // main app as a login item never reaches it. Without this its root
         // registration outlives the bundle that carried its executable.
-        FanControlService.restoreAndUnregisterForRemoval()
+        let detached = FanControlService.restoreAndUnregisterForRemoval()
+        print(detached
+              ? "UNINSTALL: fan helper daemon unregistered"
+              : "UNINSTALL: fan helper daemon still registered")
         if UserDefaults.standard.bool(forKey: DefaultsKey.sleepDisabledFlag) {
             _ = Sudoers.pmsetDisableSleep(false)
         }
@@ -26,6 +29,9 @@ enum Uninstaller {
         } catch {
             print("UNINSTALL: login item was not registered")
         }
-        exit(0)
+        // Only the daemon decides the status. A login item that was never
+        // registered is not a failure; a daemon left behind is the one thing
+        // the caller cannot see for itself.
+        exit(detached ? EXIT_SUCCESS : EXIT_FAILURE)
     }
 }
--- a/Sources/Vorssaint/Support/Uninstaller.swift
+++ b/Sources/Vorssaint/Support/Uninstaller.swift
@@ -5,14 +5,18 @@ import Foundation
 import ServiceManagement
 
 /// `Vorssaint --uninstall`: cleanly detaches the app from the system
-/// before its bundle is removed. It unregisters the login item — so no dead
-/// entry lingers in System Settings › General › Login Items — and restores
-/// normal sleep if a closed-lid session left it disabled.
+/// before its bundle is removed. It unregisters the fan helper daemon and the
+/// login item — so no dead entry lingers in System Settings › General › Login
+/// Items — and restores normal sleep if a closed-lid session left it disabled.
 ///
 /// Used by `Tools/uninstall.sh`. Must run from the installed bundle, since
 /// `SMAppService.mainApp` is scoped to the running app's bundle identifier.
 enum Uninstaller {
     static func runAndExit() -> Never {
+        // The fan helper is a daemon service of its own, so unregistering the
+        // main app as a login item never reaches it. Without this its root
+        // registration outlives the bundle that carried its executable.
+        FanControlService.restoreAndUnregisterForRemoval()
         if UserDefaults.standard.bool(forKey: DefaultsKey.sleepDisabledFlag) {
             _ = Sudoers.pmsetDisableSleep(false)
         }
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -17370,6 +17370,21 @@ struct MetricsTests {
                    "defaults suite \(argument) is named inside a namespace build.sh sweeps")
         }
 
+        // MARK: Uninstallation paths stay aligned across SelfUninstall and Tools/uninstall.sh
+        let selfUninstallSource = (try? String(contentsOfFile: "Sources/Vorssaint/Services/SelfUninstall.swift",
+                                              encoding: .utf8)) ?? ""
+        let uninstallScriptSource = (try? String(contentsOfFile: "Tools/uninstall.sh",
+                                                encoding: .utf8)) ?? ""
+        expect(!selfUninstallSource.isEmpty && !uninstallScriptSource.isEmpty,
+               "uninstall sources read back for uninstallation alignment check")
+        let requiredSubpaths = ["Library/Application Support", "Library/Caches", "Library/HTTPStorages"]
+        for subpath in requiredSubpaths {
+            expect(selfUninstallSource.contains(subpath) && uninstallScriptSource.contains(subpath),
+                   "both in-app and script uninstall sweep \(subpath)")
+        }
+        expect(uninstallScriptSource.contains("Library/Preferences/ByHost"),
+               "script uninstall sweeps ByHost preferences")
+
         if failures.isEmpty {
             print("TESTS OK (\(checks) checks)")
             exit(0)
--- a/Tools/uninstall.sh
+++ b/Tools/uninstall.sh
@@ -20,8 +20,9 @@ sleep 0.5
 
 # Detach from the system from inside whichever bundle still exists: unregisters
 # the login item (no BTM tombstone) and restores normal sleep.
-# A missing or unrunnable binary leaves the fan helper registered, and nothing
-# below can reach it: the registration lives in the system, not in the bundle.
+# The fan helper's registration lives in the system, not in the bundle, so
+# deleting the app below cannot reach it. Only the binary can drop it, and the
+# check after the loop settles what its absence or failure left behind.
 detached=1
 for candidate in "$APP/Contents/MacOS/Vorssaint" "$LEGACY_APP/Contents/MacOS/VorssaintUtils"; do
     if [[ -x "$candidate" ]]; then
@@ -30,6 +31,18 @@ for candidate in "$APP/Contents/MacOS/Vorssaint" "$LEGACY_APP/Contents/MacOS/Vor
         break
     fi
 done
+# `detached` cannot tell a failed unregister from no binary having run, and the
+# second is ordinary: an app trashed by hand, then this script for the rest. It
+# also cannot see a daemon that went despite a reported failure. launchctl
+# settles both without sudo, and still finds a registration held back for a
+# pending fan recovery, which is the one case that must keep warning.
+if (( detached )); then
+    launchctl print "system/$BUNDLE.fan-control" >/dev/null 2>&1
+    # 113 is "no such service", the only answer that proves absence. Any other
+    # failure means launchctl could not tell us, and warning then is the honest
+    # side of a check that exists to stop this script claiming what it cannot see.
+    (( $? == 113 )) && detached=0
+fi
 
 echo "▸ Resetting permissions (Accessibility, Screen Recording)…"
 tccutil reset All "$BUNDLE" >/dev/null 2>&1 || true
--- a/Tools/uninstall.sh
+++ b/Tools/uninstall.sh
@@ -20,10 +20,13 @@ sleep 0.5
 
 # Detach from the system from inside whichever bundle still exists: unregisters
 # the login item (no BTM tombstone) and restores normal sleep.
+# A missing or unrunnable binary leaves the fan helper registered, and nothing
+# below can reach it: the registration lives in the system, not in the bundle.
+detached=1
 for candidate in "$APP/Contents/MacOS/Vorssaint" "$LEGACY_APP/Contents/MacOS/VorssaintUtils"; do
     if [[ -x "$candidate" ]]; then
-        echo "▸ Detaching login item and restoring sleep…"
-        "$candidate" --uninstall || true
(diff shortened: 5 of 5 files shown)


PathGao:
One thing I'd fix before merge, two paths I think are missed, one judgment call.

**The unregister has no success signal, which is the same shape as the bug you're fixing.**

`FanControlService.restoreAndUnregisterForRemoval()` is called bare — no result, no `try`. `runAndExit()` can't tell whether the daemon actually went, so the script can't either. And `uninstall.sh` runs `set -uo pipefail` without `-e`, so a `--uninstall` that fails outright doesn't stop it: `✓ Vorssaint fully removed.` prints either way.

That lands hardest in the exact case your fallback argument is about. Quarantined bundle, broken signature, half-deleted app — the `rm -rf` lines still do their job, the root registration survives, and the user is told it's fully removed. #986's first half comes back with a tick over it. The in-app path already has a leftovers notion (`UninstallerSupport.doneSymbol(hasLeftovers:)`); the script has no equivalent. Cheapest version: exit non-zero from `runAndExit()` when the unregister didn't take, and gate the final line on it. The loop that invokes `--uninstall` is above the diff context, so if it already checks the exit status, ignore this.

**System-written paths under the same bundle id.** The body says "Every place the app writes under `~/Library` is `Application Support/<id>` or `Caches/<id>`" — that audits what *the app* writes. macOS writes more under that id on its behalf:

```sh
rm -rf "$HOME/Library/HTTPStorages/$BUNDLE" "$HOME/Library/HTTPStorages/$BUNDLE.binarycookies"
rm -f "$HOME/Library/Preferences/ByHost/$BUNDLE".*.plist
```

HTTPStorages comes from URLSession, and the share upload path alone should have created it. `defaults delete` doesn't reach ByHost. Add `~/Library/WebKit/$BUNDLE` if anything embeds a WKWebView. I haven't grepped the tree for those callers, so treat this as check-these, not confirmed-present.

**The helper's own state isn't in `$HOME`.** The fan daemon runs as root, so anything it writes lands under `/var/root/Library/…` or `/Library/…`, which an audit scoped to `~/Library` can't cover by construction. If the helper is stateless, say so in the body and it's closed — I can't see its source.

**Deleting the data folder without naming what goes.** Your argument against putting this behind `--uninstall` — anyone running it by hand loses data they did not ask to lose — applies to `uninstall.sh` unchanged, since that's also a hand-run command. I still think deleting here is right, given the README documents it as complete removal. But "and app data" in the echo reads like caches, and the share delete tokens are the one item that's unrecoverable in a way the rest isn't: once they're gone, anything already uploaded can't be revoked. One more word in that echo line, or a line in the README.

**Impact:** no regression — both paths strictly improve. The residual is that the script keeps reporting a success it has no way to verify, in the failure mode the script exists to cover.
**Merge:** needs the success signal first — one exit code, so the ✓ stops firing when the unregister didn't happen.

ref #986

mugurc:
All four land. Pushed a revision; details and the two places you were right about more than the comment claims.

**Success signal — fixed, and the loop was worse than "doesn't check".** It was `"$candidate" --uninstall || true`, so the status was actively discarded, not merely ignored. Now `restoreAndUnregisterForRemoval()` returns whether the daemon is actually gone, `runAndExit()` exits on that alone (a login item that was never registered stays a non-failure), and the script captures it and gates the final line. Three cases run against the real script with the destructive commands stubbed and `$HOME` pointed at a throwaway tree:

```
[--uninstall succeeds] exit=0  data files left=0   ✓ Vorssaint fully removed.
[--uninstall fails]    exit=1  data files left=0   ⚠ ... fan helper is still registered ...
[binary missing]       exit=1  data files left=0   ⚠ ... fan helper is still registered ...
```

The middle two are your case exactly: the `rm -rf` fallback still does its job, and the script now says what it could not do instead of printing a tick over it.

**System-written paths — confirmed, and my description was wrong about the in-app path.** `~/Library/HTTPStorages/<id>` exists here (64K), so the share upload path did create it as you guessed. While checking, I found `SelfUninstall.removePreferences()` never removed `~/Library/Caches/<id>` either — 8.6 MB on this machine. So "the in-app path does the complete job and the script drifted from it" was not right: **neither path was complete, the script was just further behind.** Both now take `Caches`, `HTTPStorages/<id>` and `HTTPStorages/<id>.binarycookies`.

Two I did not add, with what I checked:

- `Preferences/ByHost/<id>.*.plist` — nothing on this machine. You are right that `defaults delete` cannot reach ByHost, so the reasoning holds even where the file does not exist; say the word and I will add the glob defensively.
- `WebKit/<id>` — no `WKWebView` anywhere in the tree. The only WebKit references are `JunkCleaner.leftoverRoots` and `AppUninstaller`, which scan that folder for *other* apps.

That last one is worth naming: `JunkCleaner.swift:322-324` already lists `HTTPStorages`, `WebKit` and `Preferences/ByHost` as leftover roots. The app knew these were leftovers, and did not apply that to itself.

**The helper's state is in the tree, and it is two files.** `Sources/FanControlHelper/main.swift:11-12` — `/var/run/vorssaint-fan-control.lock` and `/var/run/vorssaint-fan-control.active`, and nothing else anywhere in the helper. `/var/run` does not survive a boot, and `performRestore()` unlinks the marker before the unregister is allowed to happen, so a successful detach leaves at most an empty root-owned lock file that the next restart clears. Stateless for uninstall purposes; no `/var/root` or `/Library` writes to sweep.

**Naming what goes — done.** The line now reads `Removing app, preferences, saved state and stored data (clipboard history, shelf files, share links)…`, which puts the unrecoverable item in front of the person before it goes.

On your point that the hand-run argument cuts both ways: fair, and I will not pretend otherwise. The distinction I would still draw is that the README documents `uninstall.sh` as complete removal, so someone running it has asked for exactly this, while `--uninstall` is an internal flag whose name promises nothing about data. That is a weaker line than I drew in the description, though, and if you would rather the script prompted before the data goes, that is a reasonable thing to want.

Verification on the revision: `./build.sh` no warnings, `--selftest` and `./build.sh --test` pass, `zsh -n` on the script, plus the three-case run above. Still untested end to end — I have not uninstalled my own working install to watch a live registration disappear.

PathGao:
Signal's there, and the three-case run covers what I asked for.

**ByHost: yes, add it — and the glob bites you in zsh.** You ran `zsh -n` on the script, so if the shebang is zsh, `rm -f ".../ByHost/$BUNDLE".*.plist` with nothing to match never runs `rm` at all: zsh aborts the command on NOMATCH and prints `no matches found` at someone watching an uninstall succeed. `-e` is off so the script survives, it just looks broken in the case that's most common — no file. Cheapest form:

```sh
rm -f "$HOME/Library/Preferences/ByHost/$BUNDLE".*.plist(N)
```

`(N)` is nullglob for that one pattern, no `setopt`, and BSD `rm -f` with zero operands exits 0 quietly. If the shebang is bash, the unmatched glob passes through literally and `-f` already swallows it — nothing to do. The Swift side has no glob, so ByHost there means enumerating `Preferences/ByHost` and filtering on the prefix; I'd skip it, the script is what the README documents as complete removal.

**The in-app path still hands out a ✓ it can't back.** `restoreAndUnregisterForRemoval()` now returns the one fact the caller cannot observe for itself, and `SelfUninstall.swift:103` drops it — `@discardableResult` makes that compile without a word. If `doneSymbol(hasLeftovers:)` is fed only by the file removals, the in-app uninstall prints the tick over a still-registered daemon, which is the bug you just fixed one path over. I can't see line 103's surroundings or how `hasLeftovers` is derived, so treat that as check-this, not confirmed. If it is real it's one `||` at the call site. Pre-existing and outside the title's scope, so not a merge condition.

**`JunkCleaner.leftoverRoots` is the finding worth keeping.** The same leftover-root list now lives in three places: that array, `removePreferences()`, and the script. The drift you just fixed exists because the app's own list is hand-maintained beside a canonical one it never reads. That's a much smaller dedup than the one you declined and it changes nothing about what `--uninstall` does — `removePreferences()` iterating `leftoverRoots` against its own id. Follow-up.

Nobody has watched a live registration disappear, me included — I only read the diff, and there's no hardware run of this I can see. Your call whether to do that before merging.

**Impact:** none new. The residual is the in-app ✓, which this PR neither creates nor worsens.
**Merge:** ship it.

ref #986

mugurc:
ByHost added, rebased onto main. One of the two check-this items comes back negative — details below.

**ByHost, with the `(N)`.** Shebang is zsh, so you were right about the failure mode. Confirmed it rather than taking it on faith:

```
$ zsh -c 'rm -f /tmp/none/x.*.plist'
zsh:1: no matches found: /tmp/none/x.*.plist      # rm never runs
$ zsh -c 'rm -f /tmp/none/x.*.plist(N)'
                                                   # silent
```

So the common case — no ByHost file — was the one that would have printed an error over a clean uninstall. Added in your form, with a comment saying the qualifier is load-bearing so nobody tidies it away later. Skipped the Swift side as you suggested.

**The in-app ✓ does not exist.** Checked the two things you could not see:

- `doneSymbol(hasLeftovers:)` has exactly two call sites, `PanelUninstallerView.swift:343` and `UninstallerView.swift:262`. Both are the feature that removes *other* apps, and `hasLeftovers` is `!failed.isEmpty` from `AppUninstaller`'s own trash results.
- `SelfUninstall` never touches it. `uninstallCompletely()` is called from `AdvancedSettings.swift:115` and ends at `trashOwnBundleAndQuit()` (`SelfUninstall.swift:48`) — the bundle goes to the Trash and the app terminates. There is no completion screen to put a tick on.

So the in-app path does not claim a removal it cannot back. What it does instead is say nothing: it drops the Bool and quits, and someone whose daemon did not detach learns that from System Settings later or not at all. Silence rather than a false tick, which I would still call a residual, just a different one from the one you described.

I left it. Telling that person anything means new user-facing text, and text here has to land in all thirteen languages to compile — a bigger change than the fix it would be attached to, and it belongs with a decision about *what* to show, not with this title.

**`leftoverRoots` — agreed, and it is a three-way drift, not two.** `JunkCleaner.leftoverRoots`, `SelfUninstall.removePreferences()` and the script now each carry their own copy of the same list, and this PR just widened two of the three by hand. Worth doing as its own change: `removePreferences()` reading the canonical array against its own bundle id, no behaviour change to `--uninstall`. Happy to send it once this lands, so the diff stays legible.

Nobody has watched a live registration disappear, still true. I decided against uninstalling my own working install to produce that evidence, so it stays a gap in the record rather than something I am claiming.

Rebased onto main (it had moved nine commits) and re-ran everything on the rebased tree: `./build.sh` no warnings, `--selftest` passes, `zsh -n` on the script, and the three-case run again.

**`./build.sh --test` reports one failure on the rebased tree, and it is not from this branch.** Same failure on a clean `origin/main` worktree at `6356366` with nothing of mine in it:

```
TESTS FAILED (1 of 8559):
```

It is the precondition at `MetricsTests.swift:9912`, not the behaviour under test. Patched the assertion message in the throwaway worktree to see which half gives:

```
DIAGNOSTIC occupiedWorkers=64 poolIsStarved=false
```

So all 64 blockers park as intended, and the probe queued behind them still runs inside the 0.5 s window — `DispatchQueue.global(qos: .utility)` is not held to 64 blocked threads on this machine, so the starvation the check needs never happens and the assertion fires. 18 logical cores, M5 Max, macOS 26.5.1 (25F80). CI runners are smaller, which is presumably why it is green there.

Not touching it here — different subject, and it arrived with #971. Say the word if you would rather have it as its own issue.

PathGao:
ByHost reads right, and the comment on the qualifier will stop someone tidying it away. On the in-app ✓ — you checked, it isn't there, and leaving the silence is the right call: new text in thirteen languages is a bigger change than the thing it would announce.

**The binary-missing case asserts something the script can't know.** `detached=1` means both "the unregister failed" and "no binary ran", and the ⚠ reads the same either way: `its fan helper is still registered with the system`, then reinstall and uninstall from inside the app. Someone who trashed the app by hand first, then ran the script to clear leftovers, with fan control never enabled, gets told to reinstall the thing they just removed. That's the bug you're fixing, inverted — a claim with nothing behind it.

Your own PR body has the check that closes it, and it needs no sudo. One line after the loop:

```sh
if (( detached )) && ! launchctl print system/com.vorssaint.utils.fan-control >/dev/null 2>&1; then
    detached=0
fi
```

Gated on `detached` it also covers the case where `--uninstall` reported failure but the daemon is in fact gone, and it leaves the recovery-pending case alone, since launchctl still finds that one. I'm taking the label string from your body — I can't see where it's defined. I read this exact string in your three-case run last round and passed it, so it's my miss, not a regression.

**`pmsetDisableSleep(false)` is invisible in the same way and still discarded.** `Uninstaller.swift` now says only the daemon decides the status, but if that call fails — the sudoers rule already removed, say — sleep stays disabled machine-wide and the app that could turn it back on is deleted seconds later. `sudo pmset -a disablesleep 0` by hand is the only way out and nothing says so. The `_ =` predates you, so this is not yours to fix here; it just happens to be two lines from the channel you built. Same follow-up bucket as `leftoverRoots`.

**The test failure: yes, file it.** It reproduces on clean `origin/main` with nothing of yours in it, and your diagnostic says the precondition is what gives, not the behaviour — 64 blockers park and the probe still runs, so `DispatchQueue.global(qos: .utility)` isn't capped where the test assumes it is. That makes it hardware-dependent and green on the smaller CI runners, which is the combination that rots quietly: every contributor on a large Mac sees a red suite they learn to ignore. Put the `occupiedWorkers=64 poolIsStarved=false` line and the core count in the issue, that's the whole diagnosis. A test whose precondition depends on the shared global pool's thread limit probably wants its own concurrent queue of known width, or a skip rather than a failure when the precondition doesn't hold — I haven't seen `MetricsTests.swift`, so that's a direction, not a review.

**Impact:** the ⚠ path can fire on a removal that was clean, telling that person to reinstall. New in this PR, and strictly better than the false ✓ it replaces — it errs toward saying too much rather than too little.
**Merge:** ship it.

mugurc:
Caught. And I built that one, tested it, and read the ⚠ firing on a missing binary as the feature — case C of last round's run was exactly this and I called it correct behaviour. Fixed.

Took the label from `$BUNDLE` rather than the literal, since the script already has it and Swift derives the same way — `FanControlXPC.swift:15`, `helperID = "\(appBundleID).fan-control"`. Verified the exit codes first rather than trusting the shape: `launchctl print` on a live label gives 0, on an absent one 113, neither needing sudo.

Not in the form you wrote it, though — checking my own line before pushing turned up the same fault one level down. `! launchctl print …` treats *every* non-zero as proof of absence, and absence is only one of the answers:

```
registered, not running   0
no such service         113
inaccessible domain     112
```

On a 112 the script would have printed a removal it could not see, which is the thing this whole change exists to stop. So only 113 clears it, and anything else keeps the warning:

```sh
if (( detached )); then
    launchctl print "system/$BUNDLE.fan-control" >/dev/null 2>&1
    (( $? == 113 )) && detached=0
fi
```

Hardcoding 113 is its own brittleness, and I would rather be wrong toward a warning than toward a tick. Also rewrote the comment above `detached=1`: it still said a missing binary leaves the helper registered, which the line below it had just stopped being true.

Six cases now, same harness — real script, destructive commands stubbed, `$HOME` on a throwaway tree, `launchctl` stubbed per case:

```
binary OK                        exit=0  data left=0  ✓ fully removed
binary FAILED + job present      exit=1  data left=0  ⚠ still registered
binary FAILED + job absent       exit=0  data left=0  ✓ fully removed
binary MISSING + job absent      exit=0  data left=0  ✓ fully removed   <- was the false ⚠
binary MISSING + job present     exit=1  data left=0  ⚠ still registered
launchctl cannot answer (112)    exit=1  data left=0  ⚠ still registered
```

Row three is the other half you named — a reported failure over a daemon that did go. Rows two, five and six warn, and a registration held back for a pending recovery is one launchctl still finds, so that keeps warning too. Data goes in all six.

**`pmsetDisableSleep(false)`** — agreed on the diagnosis and I left it alone. Worth adding that its silence is not equivalent: a daemon left registered is inert, sleep left disabled is a machine that stops sleeping, with the app that could undo it in the Trash. It also cannot ride on the same exit code without the ⚠ having to say which of two things went wrong, which is where the thirteen-language cost starts. Own change, same bucket as `leftoverRoots`.

**Test issue** — filing it with the diagnostic line, the core count and the clean-`main` reproduction. Will link it here.

Rebased onto main again and re-ran on the rebased tree: `./build.sh` no warnings, `--selftest` passes, `zsh -n`, and the six-case run above. `./build.sh --test` still reports the one pre-existing `MetricsTests.swift:9912` failure that reproduces on clean `main`, unchanged by any of this.

PathGao:
One thing to carry into the `leftoverRoots` follow-up: the new check in `MetricsTests.swift` is a fourth copy of that list, and it's the copy that will fight you.

It asserts `SelfUninstall.swift` literally contains `Library/Caches` and `Library/HTTPStorages`. The moment `removePreferences()` iterates `JunkCleaner.leftoverRoots` instead of naming the paths, those substrings leave the file and the check fails — on the change that made the drift structurally impossible. So the dedup has to move that assertion onto the canonical array in the same commit, or it reads as a regression on the way in. Nothing to do here: while both sides are hand-written lists, the check is doing exactly what it should.

On the 113: right call, and right side to be wrong on. Hardcoding it is brittle, but the brittleness costs a warning that shouldn't have fired, and the alternative costs a ✓ that can't be backed. 112 keeping the warning is the same trade made once more.

mugurc:
Filed the test issue as promised: #1172.

It turned out to be more than a hardware quirk. The precondition is false on
this Mac, so the assertion below it runs against a healthy pool and would pass
with or without the fix from #989 — the check is green for the wrong reason
wherever this happens. I measured the pool ceiling at 90 blocked threads here
rather than 64, so raising the constant is not the fix I would reach for. The
numbers and the reasoning are in the issue.

Issue #888, opened 2026-08-24
Hyper key not working for "move and resize by dragging"


felixblaschke:
### Description

"Move and resize by dragging" doesn't support own "Hyper key"

### Steps to Reproduce

1. Set "Move and resize by dragging" to Command, Option, Control
2. Set "Hyper key" to Command, Option, Control
3. Try dragging window with hyper key.
4. Observe not working

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

iltonandrew:
Taking a look at this one.

PathGao:
Your report has two parts and only one of them is done, so this is now labelled `half` rather than fixed. That means you will not be asked to confirm a fix you never received, and the report stays open.

**Done:** a Hyper key now reaches the drag feature at all. Assigning it works, and the app reads the chord back.

**Not done:** when the app compares the combination you saved, it drops Shift before comparing. A held Hyper key includes Shift, so after that it looks identical to the built-in resize combination and can never match the move one. That is why dragging only ever resizes for you. Letting a saved combination keep Shift is the remaining work, and until it lands this stays open.

PR #993, opened 2026-08-25
fix(super-key): carry the held modifiers on mouse presses so drag chords see them

iltonandrew:
Refs #888.

## What was wrong

The super key adds its modifiers from an event tap that watches keyboard events only, so they existed solely on keyboard events. "Move and resize by dragging" reads its chord off the mouse-down's own flags — which were bare while the key was held. Physical modifiers work because the HID system stamps real modifier state onto every mouse event; the virtual key had no equivalent, so the drag never armed.

## The change

Mouse button presses now ride through a second, tiny tap at the HID stage — left and right mouse-down only — and are classified like any other key while the super key is held: the configured modifiers ride along, and the press cancels the solo action, matching what a physical modifier chord does. The HID stage is the one place guaranteed to run before every session tap that reads mouse flags, so ordering never depends on which feature installed its tap last, and every consumer (window drag, dock-click modifiers, mouse shortcuts) sees the same flags. Moves and drags stay out of the mask: chords are read on the press, and per-move tap work is a known stutter source. The global modifier state is deliberately never touched — the stamp exists only on the events themselves. The new tap shares the existing tap thread, teardown, timeout re-enable, and health check.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8550 checks), `./build/Vorssaint --selftest` → SELFTEST OK. Runtime-verified on a live dev build with synthetic input entering at the same stage the mapped key uses: on unpatched main with the service confirmed running, a physically-modified drag moves a window +120,+80 while a super-key-held drag moves it not at all — the report, reproduced; with this change the same super-key-held drag moves it exactly −120,−80, and the session modifier state reads clean after every run. A source-shape test pins the mouse classification, the down-only mask, and the HID-stage tap; the held-key modifier semantics keep their existing unit coverage.


PathGao:
Read the diff. Three things.

**A mouse-down carries no keycode.** `handle` pulls `keyCode` before `classify`, and on a `leftMouseDown` event `getIntegerValueField(.keyboardEventKeycode)` reads 0 — which is `A`. The classification itself is right because the mouse check short-circuits, but if anything downstream of `.otherKey` looks that keycode up against the configured super-key bindings, a click while the key is held fires the `super+A` action. That code isn't in the diff so I can't tell which it is: gated on `.keyDown` you're fine, gated on `.otherKey` you're not. Worth pinning with a test either way.

**The mask is narrower than the body's claim.** Body says "every consumer (window drag, dock-click modifiers, mouse shortcuts) sees the same flags." Mouse shortcuts bound to middle/back/forward arrive as `.otherMouseDown`, which is in neither the mask nor the classify guard, so those still read bare flags. One member in each place:

```swift
| (CGEventMask(1) << CGEventType.otherMouseDown.rawValue)
...
if type == .leftMouseDown || type == .rightMouseDown || type == .otherMouseDown { return .otherKey }
```

If the omission is deliberate, say so in the body — right now the claim and the mask disagree.

**A failed mouse tap is silent and permanent.** If `tapCreate` returns nil, `self.mouseTap` stays nil, so `mouseDead` in the health check is `false` forever and nothing ever rebuilds it. The keyboard half keeps working and #888 quietly stays broken on that machine. The keyboard tap has the same shape today so it isn't a regression, but the new tap is the one more likely to be refused.

Not verified: the stamping itself. `handle`'s `.otherKey` branch and wherever the modifiers get written onto the event aren't in the diff, so "the configured modifiers ride along" rests on your runtime test, not on anything I read. Same for teardown ordering — it reads as balanced, but I can't run it.

One note without an ask attached: the tap thread now gates every left and right click in the system, not just keystrokes. If it stalls, the freeze takes the mouse with it.

**Impact:** none if the binding lookup is keyDown-gated; otherwise super-key + click fires the `A` binding. Other-button chords see bare flags either way.
**Merge:** needs confirmation that the mouse-down keycode never reaches the binding lookup first.

ref #888

iltonandrew:
All three addressed.

**1. The phantom keycode — traced, not a live bug, but the shape is fixed.** The read at `handle()` fed `classify` and nothing else, and `classify` answers the mouse types before `keyCode` is compared to anything. `SuperKeySupport.Event.otherKey` is a payload-free case, so the value cannot travel past `classify` by construction, and `State.decide(.otherKey)` returns only `.pass`/`.addModifiers`. There is no binding table in this service at all — it stamps flags and lets dispatch happen downstream — so `super+A` had nothing to fire.

Agreed on the shape regardless: the read now lives inside `classify`, below the mouse answer, and the `keyCode:` parameter is gone, so `handle` never has a phantom keycode in scope. Pinned — `.keyboardEventKeycode` must appear exactly once and after the mouse-type answer; moving it back into `handle` fails the test.

**2. `.otherMouseDown` added** to both the mask and the guard, so the body's claim and the mask now agree. I looked for a reason to exclude it first: `MouseButtonShortcutService` synthesizes *keyboard* events and reads only the button number; `MouseNavigationService` sets its own flags on what it posts; `RadialMenuService`, `AppSwitcher` and `DockPreviewService` are flags-blind on the extra buttons. `MiddleClickService.postMiddleTap()` does synthesize `.otherMouseDown` at the HID stage, but this tap only unions flags and never re-posts, so there's no loop — and a middle click made while the key is held now carries the chord like a physical modifier would.

**3. A refused mouse tap is no longer silent.** `tapCreate` returning nil is remembered, and the existing `syncWithPreferences` health check now counts "asked for and refused" as dead, so it rebuilds instead of leaving #888 quietly broken for the run. The flag is set after the keyboard tap is stored, so the brief legitimate window where `tap != nil` and the mouse tap isn't yet can't cause a spurious rebuild, and it clears with the tap thread.

On your closing note — that this tap now gates every left and right click system-wide: agreed, and that is why the mask stays down-only and the callback does nothing but classify and union flags. No AX, no window-server calls, no allocation on that path.

Gate: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8552 checks), `--selftest` → SELFTEST OK.

PathGao:
All three are fixed. The keycode read moving inside `classify` below the mouse answer is the right shape, and pinning `.keyboardEventKeycode` to exactly one occurrence keeps it that way.

One new thing, out of the fix itself.

**A refused mouse tap is now permanently unhealthy, and the cure takes the working keyboard tap down with it.** `mouseTapMissing` makes `deadTap` true → `syncWithPreferences` runs `stop()` then `start()` → the new thread asks again, gets refused again, sets the flag again. Nothing latches, so it repeats for the rest of the run. Each pass invalidates and rebuilds the keyboard tap that was healthy, dropping held-key state and letting events through untapped for that window.

How bad depends on who calls `syncWithPreferences`, which isn't in the diff. Preference changes only → a few wasted rebuilds, ignore this. Anything recurring (and the "Accessibility revoked and granted again" case in the comment right above it implies something recurring) → the super key hiccups at that interval on any machine where the HID tap can't be created, which is strictly worse than the mouse half being quietly broken.

Laziest bound is letting a refusal drive exactly one rebuild:

```swift
private var mouseTapRefusals = 0
// in the tap thread, where the flag is set now:
if mouseTap == nil { self.mouseTapRefusals += 1 }
// health check:
let mouseDead = mouseTap.map { !CGEvent.tapIsEnabled(tap: $0) } ?? (mouseTapRefusals == 1)
```

That only works if the counter survives teardown, so it can't go in the `clearEventTapThread` reset block the way `mouseTapMissing` does.

Minor, no ask attached: the mouse-type set is now written three times — the mask, the `classify` guard, and the literal string in the shape test. A fourth type means three edits, and forgetting the third fails a test that reads like a real break. One array feeding `reduce` for the mask and `contains` for the guard kills two of the three, at the cost of rewriting the pin. Also `keyboardDead` is still `?? false`, so a refused *keyboard* tap stays silent — pre-existing, not asking for it here.

Not verified: the stamping itself and the middle-click no-loop reasoning both rest on files outside the diff. I'm taking your runtime run for both.

**Impact:** none where the HID tap is created successfully. Where it's refused, super-key input drops at whatever rate `syncWithPreferences` fires.
**Merge:** needs the refusal bounded to one rebuild first, or confirmation that `syncWithPreferences` only runs on preference changes.

ref #888

iltonandrew:
Checked, and I can't give you the confirmation you offered as the alternative — so I bounded it.

`syncWithPreferences` isn't on a timer. Callers are `SuperKeySettings.swift:36/166/180`, `MenuPanelView.swift:1661`, and the `.superKey` binding at `FeatureRuntime.swift:144`, reached from `syncAtLaunch` (`AppDelegate.swift:114`), the feature-hub / onboarding / command-bar toggles, and the `Permissions.shared.$accessibility` sink at `AppDelegate.swift:126-135`. That last one is the catch: `Permissions` does poll (`Permissions.swift:102`), but it only republishes on an actual change (`:196`) and the sink has `.removeDuplicates()`, so it fires on grant/revoke **edges** — which is precisely the "Accessibility revoked and granted again" case the comment above the health check names, and it is unbounded over a run.

And even with no loop, the flag as written re-tore the *healthy keyboard tap* on every one of those calls forever: four modifier clicks in Settings is four teardowns. Your point stands either way.

So `mouseTapMissing` is now `mouseTapRefusals`, and the check reads `?? (mouseTapRefusals == 1)`. Traced: refused → 1 → one rebuild → refused → 2 → `mouseDead` false from then on, `start()` returns early. Exactly one extra rebuild, then it stops. The count is deliberately **not** cleared in `clearEventTapThread()` — there's a comment there saying why, and the test asserts the only place it goes to zero is its own declaration, so re-adding that reset fails the build.

On resetting after a success: yes — `mouseTap == nil ? refusals + 1 : 0`. The counter only ever gates the *extra* rebuild, never the ordinary ask, so zeroing on a genuine success just gives a refusal-after-a-working-stretch its own single retry, which is right for a transient `tapCreate` failure and identical for a static refusal. It can't ping-pong, since a successful rebuild leaves the health check false.

Took the de-duplication too: one `mouseDownTypes` array feeds the mask via `reduce` and the guard via `contains`, and the shape test pins that list instead of the expanded guard string, so a fourth button is one edit. Left the `keyboardDead ?? false` asymmetry alone, as you said.

Bite proof on the new pin: relaxing the bound to `> 0` fails it, and re-adding `mouseTapRefusals = 0` to `clearEventTapThread()` fails it. Gate: `./build.sh` zero warnings, `./build.sh --test` → TESTS OK (8552 checks), `--selftest` → SELFTEST OK.

PathGao:
Traced the bound against the diff, it holds: `mouseTapRefusals` only reaches 1 after a refusal, `?? (mouseTapRefusals == 1)` is true for exactly that one sync, the rebuild's refusal takes it to 2, and every sync after reads false. Reset-on-success is right — a refusal after a working stretch gets its own single retry and can't ping-pong, since a successful rebuild leaves the health check false. Keeping the count out of `clearEventTapThread()` is the part that makes it terminate, and pinning that is the right place to spend a test.

`.otherMouseDown` in both the mask and the guard now matches the body's claim, and the one `mouseDownTypes` list feeding `reduce` and `contains` kills two of the three copies.

Minor, no ask: the new pin counts the literal `mouseTapRefusals = 0`, and the single occurrence it matches is the declaration. Spelling that declaration `private var mouseTapRefusals: Int = 0` drops the count to zero and fails a test that reads like a real break — same shape as the guard-string pin you just replaced.

Not verified, unchanged from last round: the stamping and the middle-click no-loop reasoning both live outside the diff, so they rest on your runtime run rather than on anything I read. Everything I checked this time is source-level.

**Impact:** none. Where the HID tap is refused, the mouse half stays broken and the healthy keyboard half is left alone, which is the right trade.
**Merge:** ship it.

ref #888

PR #994, opened 2026-08-25
fix(popover): keep outside-click dismissal consistent

pergioa:
## Summary

Routes every AppKit popover-close path through one policy in `PanelInteractionState` instead of making the global click monitor answer differently from `popoverShouldClose`.

Normal Homebrew browsing, searching, details and refreshes now behave like an ordinary popover and dismiss on a genuine outside click. A Homebrew confirmation remains protected so its parent window cannot disappear underneath AppKit modal state, and an active Homebrew operation remains protected even after switching tabs. Active Cleaner scans/cleaning and Uninstaller scans/removal are protected by the same central policy.

Existing utility workflows that intentionally span interaction with other apps retain their view-level hold; the input was renamed to distinguish it from the final dismissal decision. `AppDelegate` has no tool-specific close checks.

This is the dismissal half split from #912 and is independent of the anchoring work in #930.

## Verification

Tested on macOS 26.5.2 (25F84) against MacOSX26.sdk:

```sh
./build.sh --test             # TESTS OK (8,555 checks)
./build.sh                    # release bundle built successfully
```

Keyboard routing was verified separately from dismissal with a native AppKit UI test using a real `NSWindow`, `NSSearchField`, field editor and local event monitor. Homebrew-style query input worked with both the pre-PR and current monitor gates: letters and spaces were entered, arrow navigation and deletion edited the query identically, and Return and keypad Enter submitted the same value. All monitored key events followed the expected normal AppKit dispatch path.

I did not manually exercise every protected long-running workflow across multiple Mac models, displays, or Spaces configurations.

## Anything else

Refs #912.


PathGao:
Read the diff. Four things, one worth fixing before merge.

**`run()` drops the modal hold before the operation exists.**

```swift
private func run(_ action: HomebrewPendingAction) {
    dismissConfirmation()                    // isPresentingPopoverModal = false
    switch action.action {
    case .install:
        if let package = action.package { homebrew.install(package) }
```

Between those two lines nothing in `preventsPopoverDismissal` is true. If `HomebrewManager` publishes `operationStatus` on a later hop — it spawns a process, so probably — an outside click landing in that gap dismisses the popover exactly as the install starts, which is the case the body says stays protected. Move `dismissConfirmation()` to the bottom of `run`. Costs nothing if the status is set synchronously. `HomebrewManager` isn't in the diff, so I haven't checked which it is.

**The hold now has no upper bound.**

`viewKeepsPopoverOpen` ended when the view went away. `HomebrewManager.shared.operationStatus?.isActive` ends when the process does, and #971 says the process sometimes doesn't — a stranded `waitUntilExit` leaves the status active indefinitely, and from then on no outside click dismisses the panel. Same shape for `.scanning` in either service. Not asking you to solve #971 here, but this PR gives a hung operation a new way to be visible.

**Three copies of "this utility holds the popover open".**

The read side is one property now. The write side is still three: the row `action:` closure, the panel view's `.onAppear`, and `.onChange(of: hostedUtilityKeepsPopoverOpen)`. That's why the diff has to touch a dozen assignment sites. The `action:` closures are the redundant copy — `hostedUtilityKeepsPopoverOpen` fires on the same `show*Panel = true`, and each `Panel*View` sets it again on appear. Dropping them leaves two writers instead of three. Caveat: the head of `isHostingUtility` is above the hunk, so I can't confirm `showUninstaller` / `showCleanerPanel` / `showURLCleaner` are in it — if they aren't, their row assignments are the only writer and have to stay.

**`PanelUninstallerView`'s Homebrew alert doesn't set the new flag.**

The doc comment makes `isPresentingPopoverModal` the owner of "a SwiftUI alert or confirmation dialog is presented from the popover", but the `pendingHomebrewRemoval` alert in `PanelUninstallerView.swift` still leans on that view's `viewKeepsPopoverOpen`. No live bug, it's protected either way. It does mean the flag isn't the single owner the comment claims, and whoever removes that view's hold later removes the alert's protection without noticing.

Minor: `isPresentingPopoverModal` isn't `@Published` while `viewKeepsPopoverOpen` is. Fine if nothing renders off it — and if nothing does, `viewKeepsPopoverOpen` doesn't need it either.

**Impact:** an outside click one runloop hop after confirming a Homebrew action can dismiss the panel mid-launch. A hung `brew` or a stuck scan makes the panel un-dismissable by outside click until quit.
**Merge:** needs the `dismissConfirmation()` move in `run()` first.

pergioa:
Thanks for the careful review. Addressed the actionable points in 0b4c28e:

- Moved `dismissConfirmation()` to the end of `run(_:)`, so the modal hold remains active until `HomebrewManager` has registered the operation.
- Confirmed every listed hosted utility is included in `isHostingUtility` and removed the redundant row/action assignments, including the recent-captures helper. The section-level `onChange` and view lifecycle writers remain.
- Made the `PanelUninstallerView` Homebrew alert explicitly set and clear `isPresentingPopoverModal` on presentation, dismissal, confirmation, and disappearance. Its confirmation path also keeps the hold through operation registration.
- Removed the unused `ObservableObject` / `@Published` conformance from `PanelInteractionState`.

Verified with all 8,555 tests, an optimized release build, and `SELFTEST OK`.

PathGao:
Re-read the diff at 0b4c28e. The `run()` ordering, the removed row writers, the uninstaller alert, and the `ObservableObject` drop all check out. One thing left, and it isn't about dismissal.

**The local key monitor is gated on `viewKeepsPopoverOpen`, and Homebrew just left that set.**

```swift
guard popover.isShown,
      PanelInteractionState.shared.viewKeepsPopoverOpen,
      isPlainPopoverHoldKey(event),
```

Before this PR the Homebrew panel held that flag true the entire time it was on screen — the row `action:` closure plus `onChange(of: isHostingUtility)` — so every plain keystroke went through this branch. Now it's false for the whole Homebrew panel, and the `SearchField` subclass that re-set it on `keyDown`/`mouseDown` is gone too.

Two readings, and the body of that guard is below the hunk so I can't pick between them:

- the monitor only exists to redirect keys when the popover isn't key window, which is exactly the utilities that span other apps — Homebrew being excluded is then correct and nothing changes;
- or it's what actually delivers characters into the popover, and Homebrew search now eats them.

Open the panel and type a query, that settles it. 8,555 unit tests don't cover `NSEvent` monitor routing.

Either way it belongs in the body. You describe this as the dismissal half of #912, but `viewKeepsPopoverOpen` has a second consumer, and the rename changed which panels are in its set.

The unbounded hold from a stuck `brew` or a stuck scan is still there. Still not this PR's job.

**Impact:** possibly none. If that monitor is the delivery path for popover keystrokes, typing in the Homebrew search field stops working.
**Merge:** needs one real-machine check that Homebrew search still accepts typing.

pergioa:
I traced this through the complete monitor and responder path. The suspected keyboard regression is not present.

`isPlainPopoverHoldKey(_:)` only accepts Space, Return, and keypad Enter when Command/Control/Option are absent. Before this PR, letters, deletion, and navigation already failed that test and were returned unchanged. Space, Return, and keypad Enter reached `isTextEditingActive`, saw the `NSSearchField` field editor, and were also returned unchanged. With Homebrew no longer setting `viewKeepsPopoverOpen`, those same events are simply returned earlier.

That matters because AppKit specifies that returning the original event from a local monitor preserves normal dispatch; only returning `nil` consumes it. The removed `SearchField` subclass only set the old hold flag before calling `super` and never routed or consumed keyboard events.

I ran a native AppKit UI test with a real `NSWindow`, `NSSearchField`, field editor, and the pre-PR/current monitor gates. Query input works in both paths: letters and spaces were entered, arrow navigation and deletion edited the query identically, and both Return and keypad Enter submitted the same value. All 20 monitored key events followed the expected normal-dispatch path.

The keyboard and dismissal policies therefore remain independent: ordinary Homebrew browsing can dismiss on an outside click, while confirmations and active operations are protected by `isPresentingPopoverModal` and `operationStatus?.isActive`.

Verification at `0b4c28e`: all 8,555 tests pass, the optimized release build completes, and `SELFTEST OK`. No code change is needed for this comment.

PathGao:
Trace holds. `isPlainPopoverHoldKey` narrows the branch to Space/Return/keypad Enter, and those already reached the field-editor check and came back unchanged, so nothing the search field sees is different. I still haven't seen the body of that guard, so I'm taking your read of it.

One thing your trace doesn't say that makes the exclusion safer than the test alone shows: that branch only does work when the popover isn't key window, and under the new policy every ordinary way for Homebrew to lose key focus now dismisses the popover. The state where the routing mattered and no longer runs is mostly unreachable by construction, not just unobserved.

Mostly. An active operation holds the popover open through a click into another app, so it can be open, not key, and no longer routed. Nothing in that view is keyboard-activated today as far as I can see, so no change needed — just the thing to remember when someone adds a Space/Return control to the operation UI.

Stuck `brew` still leaves an un-dismissable panel. Separate problem, #971.

**Impact:** none I can reach.
**Merge:** ship it.

PR #995, opened 2026-08-25
fix(auto-quit): stop waking Electron renderers when watching apps

### LINKED ISSUE #953: Quit on close causing elevated background CPU usage on some apps
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

I noticed that after enabling Vorssaint’s "Quit on close causing elevated background CPU usage on some apps" from the Quit on close section, [UpNote's](https://getupnote.com/) subprocess "UpNote Helper (Renderer)" starts to use 4-10% of CPU core even when the app is minimized or not used. Over a day, this amounted more CPU and energy usage than system processes like WindowServer and kernel_task.

I didn't notice the same issue on other apps, but UpNote is an Electron app, so the it could be more widespread.

### Steps to Reproduce

Start UpNote
Enable "Quit on close causing elevated background CPU usage on some apps" from the Quit on close section in Vorssaint's settings

### Feature Area

Windows and Dock

### Screenshots / Recordings

_No response_

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

### Language

English

### Mac and displays

_No response_

### Installation Method

_No response_

iltonandrew:
Fixes #953 (mechanism proven on local Chromium apps; the UpNote-specific percentage is the one thing only the reporter can confirm).

## Problem

With Quit on close enabled, an Electron app's Helper (Renderer) burns 4–10% of a core continuously, even minimized and idle, with nothing querying the app.

## Root cause

`attach()` probed each watched app for liveness by reading `kAXRole` on its *application* element. A Chromium app (Electron, and the browsers) answers that query by switching its renderers into full accessibility mode, and from then on rebuilds and ships an accessibility tree on every DOM change for the life of the process. Enabling the feature sweeps every running app, so it fires at all of them at once — which is why the report starts at the toggle.

Isolating each AX call against a fresh Electron app (accessibility off ≈ 13 nodes, no `AXWebArea`): `kAXWindows`, `kAXMainWindow`/`kAXFocusedWindow`, the full window scan, observer registration, and the close-button hit test all leave it at 13. The role read alone takes it to 156 with two `AXWebArea`s. Chrome behaves the same (17 → 1222 nodes — the whole page materialized). Renderer CPU, sampled after the probe had already exited: idle Electron note app 0.32% → 0.39%; a Chromium tab holding a 3000-paragraph document with light churn 4.7% → 6.0%. The cost scales with document size and DOM churn and never comes back down — a renderer already flipped only settles after the app restarts.

## The change

Probe with `kAXWindows` instead. Same synchronous call, same `.cannotComplete` signal the issue #189 retry ladder needs, the role value was written and never read, and it is what the refresh reads a moment later anyway — `WindowEnumerator` has always probed exactly this way; AutoQuit was the lone outlier. The post-fix attach shape, including observers and a 1Hz refresh, leaves the app at 13 nodes.

## Verification

`./build.sh` with zero warnings, `./build.sh --test` → TESTS OK (8551 checks), `./build/Vorssaint --selftest` → SELFTEST OK. Source-shape assertions pin the windows probe and ban a role read on the application element; reverting the line turns both red.

@Sirtx — after this lands, restart UpNote once (the flipped accessibility mode only clears on relaunch) and the renderer should settle.

### DIFF
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -159,8 +159,17 @@ final class AutoQuitService: ObservableObject {
         // it: typing dies system wide (issue #189). Give up fast and retry
         // with growing spacing until the app services its run loop again.
         AXUIElementSetMessagingTimeout(appElement, 0.35)
-        var role: CFTypeRef?
-        if AXUIElementCopyAttributeValue(appElement, kAXRoleAttribute as CFString, &role) == .cannotComplete {
+        // The probe asks for windows rather than the application's role. A
+        // Chromium app (Electron, and the browsers) answers a role query on its
+        // application element by switching its renderers into full
+        // accessibility mode, and from then on it rebuilds and ships an
+        // accessibility tree on every DOM change for the life of the process —
+        // a few percent of a core, forever, in an app this feature only ever
+        // needed a window count from (issue #953). Windows are the same
+        // liveness signal, are what the refresh below reads anyway, and leave
+        // that mode alone; WindowEnumerator probes the same way.
+        var windows: CFTypeRef?
+        if AXUIElementCopyAttributeValue(appElement, kAXWindowsAttribute as CFString, &windows) == .cannotComplete {
             guard attempt < 6 else { retryingApps.remove(pid); return }
             retryingApps.insert(pid)
             let delay = 5.0 * pow(2.0, Double(attempt))
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -3882,6 +3882,27 @@ struct MetricsTests {
                                                             visibleSpaces: nil,
                                                             hasTitle: true),
                "without Spaces to compare, the old cautious rule stands")
+
+        // Attaching to a watched app must never ask its application element for
+        // a role. A Chromium app (Electron, and the browsers) answers that by
+        // switching its renderers into full accessibility mode, and then pays
+        // to rebuild and ship an accessibility tree on every DOM change for the
+        // rest of its life — measured on an idle app that this feature only
+        // ever needed a window count from (issue #953). Windows are the same
+        // liveness signal and leave that mode alone. Comments are stripped
+        // first: the note above the probe names the attribute it avoids, and a
+        // check that cannot tell prose from a call would go red for it.
+        let autoQuitServiceCode = ((try? String(
+            contentsOfFile: "Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift",
+            encoding: .utf8)) ?? "")
+            .components(separatedBy: "\n")
+            .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
+            .joined(separator: "\n")
+        expect(autoQuitServiceCode.contains(
+                   "AXUIElementCopyAttributeValue(appElement, kAXWindowsAttribute as CFString, &windows) == .cannotComplete"),
+               "AutoQuit probes a watched app for liveness by asking for its windows")
+        expect(!autoQuitServiceCode.contains("appElement, kAXRoleAttribute"),
+               "AutoQuit never asks a watched app's application element for its role")
         expect(Defaults.sanitizedPanelItemOrder("uninstaller,homebrew,homebrew,bad",
                                                 defaultOrder: ["homebrew", "media", "uninstaller", "cleanURL", "cleaning"])
                == ["uninstaller", "homebrew", "media", "cleanURL", "cleaning"],


PathGao:
Read the diff. One thing worth chasing before this gets called done for #953, plus a note on the test.

**The "lone outlier" claim is the part I can't check.** The diff shows one role read on an application element; whether it's the only one in the tree is a claim about code that isn't in front of me. If anything else — switcher, Dock Preview, focus-follows-mouse — reads `kAXRole` (or any other attribute that flips the same switch) on an *application* element, this lands and Sirtx still sees the renderer spin, because one wake is permanent for the life of the process. Worth grepping `kAXRoleAttribute` against every `AXUIElementCreateApplication` result before telling him it's fixed.

**Probe divergence is narrower than the body implies, and that's mostly good.** Non-`.cannotComplete` errors take the same branch as before, so a menu-bar-only app answering `.attributeUnsupported` for windows behaves exactly as it did under the role read. The one real difference: a windows read scales with window count where a role read doesn't, so under the 0.35s timeout an app that *is* servicing its run loop but is slow to enumerate can now land in the #189 ladder and, after 6 attempts, never attach — Quit on close silently off for it. Narrow, and the refresh a moment later reads the same attribute anyway, so it would fail either way. Not asking you to change the line; "same synchronous call" just isn't the whole story.

**The test pins the call site as a literal string.** Wrap that line differently and it goes red with no behavior change — the probe is now spelled out in two places. The negative assertion is also vacuously green when the file read fails (`?? ""` contains nothing); the positive one is what actually carries the suite red, so the ban isn't doing the work it looks like it's doing.

**Impact:** none beyond the intended fix, except the timeout case above.
**Merge:** ship it. The Swift 6.0.3 check hasn't reported yet.

ref #953

iltonandrew:
Ran the grep you asked for on current main. Short version: the attach probe was the only unconditional application-element role read, so the fix does land — but the walk in `topLevelWindow(from:)` is a path that can reach one, and it is worth a guard.

**Every `kAXRoleAttribute` read in `Sources/`** (12 sites): `AutoQuitService:424` and `:762`, `WindowEnumerator:646/653/668`, `DockClickService:827`, `CommandBarMenus:65`, `FinderRenameService:188`, `FocusFollowsMouseService:135/149`, `WindowLayoutService:1844`, `WindowMaximizer:418`.

Cross-referencing those against every `AXUIElementCreateApplication` site, only `AutoQuitService` and `WindowEnumerator` are in both files, and in both the role read is on a **window**, never on the application element. The switcher, Dock Preview, focus-follows-mouse and the rest read role on windows, menu descendants, or hit-test elements. So no other place wakes a renderer at attach time, and enabling Quit on close no longer flips every running app.

**The one path that can still land on an application element.** `AutoQuitService.topLevelWindow(from:)` walks `kAXParent` up to 8 hops, reading `role(of: parent)` at each. Normally it stops at the `AXWindow` — but the parent chain above a window *is* the application element, so an element that never yields an `AXWindow` (a menu-bar item, a detached palette) walks straight onto it and reads its role there. That is the same wake, from a different door.

It cannot produce the reported symptom — that walk only runs on the close-button hit-test path, i.e. after a close signal on one app, not the enable-time sweep over every app — which is consistent with Sirtx seeing this start at the toggle. But it is live, and one unlucky hit test permanently wakes that app's renderers.

Happy to send a follow-up that stops the walk when it reaches the application element (build it once from the pid, `CFEqual` against each parent, bail instead of reading role). Small and self-contained; say the word and I will open it against main rather than piling it onto this one.

**On the test:** you are right that it pins the negative by source text, so it catches a revert of this line but not a *new* role read somewhere else. If the guard above turns into a PR, the natural place for a real guard is a repo-wide assertion — no `kAXRoleAttribute` read on an element derived from `AXUIElementCreateApplication` — rather than another per-file grep.

PathGao:
> Ran the grep you asked for on current main. Short version: the attach probe was the only unconditional application-element role read, so the fix does land — but the walk in `topLevelWindow(from:)` is a path that can reach one, and it is worth a guard.已在当前主分支上运行了您要求的 grep 命令。简而言之：附加探测是唯一无条件读取应用程序元素角色的操作，因此修复生效了——但 `topLevelWindow(from:)` 中的遍历路径可能会到达一个元素，因此值得添加一个保护机制。
> 
> **Every `kAXRoleAttribute` read in `Sources/`** (12 sites): `AutoQuitService:424` and `:762`, `WindowEnumerator:646/653/668`, `DockClickService:827`, `CommandBarMenus:65`, `FinderRenameService:188`, `FocusFollowsMouseService:135/149`, `WindowLayoutService:1844`, `WindowMaximizer:418`.**在 `Sources/` 中读取的每个 `kAXRoleAttribute`** （12 个站点）： `AutoQuitService:424` 和 `:762` ， `WindowEnumerator:646/653/668` 668， `DockClickService:827` ， `CommandBarMenus:65` ， `FinderRenameService:188` ， `FocusFollowsMouseService:135/149` ， `WindowLayoutService:1844` ， `WindowMaximizer:418` 。
> 
> Cross-referencing those against every `AXUIElementCreateApplication` site, only `AutoQuitService` and `WindowEnumerator` are in both files, and in both the role read is on a **window**, never on the application element. The switcher, Dock Preview, focus-follows-mouse and the rest read role on windows, menu descendants, or hit-test elements. So no other place wakes a renderer at attach time, and enabling Quit on close no longer flips every running app.将这些内容与所有 `AXUIElementCreateApplication` 站点进行交叉比对，发现只有 `AutoQuitService` 和 `WindowEnumerator` 两个函数同时存在于两个文件中，并且它们读取的角色都是**窗口 **，而不是应用程序元素。切换器、Dock Preview、focus-follows-mouse 以及其他函数读取的角色都是窗口、菜单子元素或命中测试元素。因此，没有其他地方会在附加时唤醒渲染器，启用“关闭时退出”功能也不会再导致所有正在运行的应用程序都发生切换。
> 
> **The one path that can still land on an application element.** `AutoQuitService.topLevelWindow(from:)` walks `kAXParent` up to 8 hops, reading `role(of: parent)` at each. Normally it stops at the `AXWindow` — but the parent chain above a window _is_ the application element, so an element that never yields an `AXWindow` (a menu-bar item, a detached palette) walks straight onto it and reads its role there. That is the same wake, from a different door.**唯一一条仍然可以到达应用程序元素的路径**是 `AutoQuitService.topLevelWindow(from:)` 。它会沿着 `kAXParent` 最多 8 级遍历，并在每一级读取 `role(of: parent)` 。通常情况下，它会在 `AXWindow` 处停止——但窗口上方的父级链_是_应用程序元素，因此，永远不会产生 `AXWindow` 元素（例如菜单栏项、分离的调色板）会直接到达应用程序元素并读取其角色。这本质上是同一个过程，只是路径不同。
> 
> It cannot produce the reported symptom — that walk only runs on the close-button hit-test path, i.e. after a close signal on one app, not the enable-time sweep over every app — which is consistent with Sirtx seeing this start at the toggle. But it is live, and one unlucky hit test permanently wakes that app's renderers.它无法重现所报告的症状——该循环仅在关闭按钮的点击测试路径上运行，即在单个应用程序收到关闭信号后运行，而不是在所有应用程序上执行启用时间扫描——这与 Sirtx 检测到的从切换开关开始的情况一致。但它确实存在，一次不成功的点击测试就会永久唤醒该应用程序的渲染器。
> 
> Happy to send a follow-up that stops the walk when it reaches the application element (build it once from the pid, `CFEqual` against each parent, bail instead of reading role). Small and self-contained; say the word and I will open it against main rather than piling it onto this one.我很乐意发送后续代码，在到达应用程序元素时停止遍历（根据进程 ID 构建一次，对每个父级使用 `CFEqual` ，然后跳过读取角色的操作）。代码小巧且独立；只要你开口，我就会将其添加到主分支，而不是堆砌到这个分支上。
> 
> **On the test:** you are right that it pins the negative by source text, so it catches a revert of this line but not a _new_ role read somewhere else. If the guard above turns into a PR, the natural place for a real guard is a repo-wide assertion — no `kAXRoleAttribute` read on an element derived from `AXUIElementCreateApplication` — rather than another per-file grep.**关于测试：** 您说得对，它通过源文本锁定否定，因此可以捕获到此行的撤销操作，但无法捕获到在其他地方读取的_新_角色。如果上述守则最终提交为 PR，那么真正的守则更适合在仓库范围内添加断言——禁止对派生自 `AXUIElementCreateApplication` 的元素读取 `kAXRoleAttribute` 而不是为每个文件单独添加 grep 命令。

do the pr，thanks

iltonandrew:
Done — #1016.

While writing it I found the same walk a second time: `WindowMaximizer.swift:266-271` carries it byte for byte on the green-button hit-test path, with the identical hole. I have fixed both in that one PR rather than splitting them, because the identifier-based scan you suggested structurally cannot see either — the walk reads `parent`, which is never *named* after an application element — so shipping one copy would have left its twin live with nothing to catch it. Say the word if you would rather have them separate and I will split it.

That is also why there are two repo-wide rules rather than one:

- no `kAXRoleAttribute` read on an element assigned from `AXUIElementCreateApplication` — your suggestion, replacing the per-file grep #995 left behind. Text, not dataflow, and the comment says so: an application element handed to a helper, stored in a property, or passed inline still gets past it. Both direct reads found so far were written the way it catches.
- a file that walks the parent chain reading a role must stop at the application element first — which is what actually covers this bug, and what will catch a third copy if one appears.

Both proved to bite: a planted role read on an application element fails the first by `file:line`, and removing the new guard from `WindowMaximizer` fails the second naming that file.

The guard itself is behaviour-neutral by construction — that hop already ended the walk with no window found, since the application element's role is never `AXWindow` and nothing sits above it in the chain — so every input gets the same answer, minus the touch.

PathGao:
Keep them together. Shipping one copy with its twin live is worse than the bigger diff.

Two guarded copies of the same walk is still two copies. Your second rule is what catches a third one, so the check is in the right place — not asking you to dedupe the walk in #1016, just noting the rule is now doing the work a shared function would.

I haven't read #1016's diff, so the guard itself is unverified from here.

PR #999, opened 2026-08-25
fix(settings): show tapped Homebrew updates and expand row hit targets

pergioa:
## Summary

Homebrew reports installed third-party formulas with a short `name`, while `brew outdated --json=v2` identifies them by their fully qualified tap name. The Homebrew view keyed those records differently, so its update total included tapped formulas but their package rows appeared up to date and offered no update action.

Use the validated `full_name` as the canonical formula identifier, matching the outdated response and ensuring every updateable row receives its update state. Add a regression fixture covering installed and outdated data for a third-party tapped formula.

Casks have the identical split: `brew info --json=v2 --installed` reports a cask's `token` and `full_token`, while `brew outdated --json=v2` reports tapped casks by full token. `parseCask` and `parseCaskRecord` now prefer a validated `full_token`, falling back to the short `token` when there is no tap, so tapped-cask rows attach to the correct outdated entry the same way tapped formulae do. `parseCaskRecord`'s identifier feeds the app-updates matching flow, so installed cask records use the same fully qualified identifier. The human-readable cask display name is unchanged. `HomebrewCommandBuilder.isValidToken` already accepts `/` and is untouched.

The existing hit-target change expands the package-selection area across the unused row width in both the Homebrew panel and Settings. Clicking that area only selects the package and loads its details; the separate update/install/uninstall controls and their behavior are unchanged.

## Verification

Tested on macOS 26.5.2 (25F84) with Homebrew 6.0.19.

```sh
env DEVELOPER_DIR=/Library/Developer/CommandLineTools ./build.sh --test
# TESTS OK (8557 checks)

env DEVELOPER_DIR=/Library/Developer/CommandLineTools ./build.sh --dev
# Developer bundle built successfully

./build/VorssaintDeveloper --selftest
# SELFTEST OK
```

Also reproduced the original identifier mismatch using local `brew info --json=v2 --installed` and `brew outdated --json=v2` output.

Added a tapped-cask regression fixture (`token`/`full_token` pair, plus a matching outdated entry) asserting the installed cask is identified by its full token, that ID matches the outdated entry's ID, the update is found through it, and that `parseInstalledCaskRecords` preserves a tapped cask's full token:

```sh
env DEVELOPER_DIR=/Library/Developer/CommandLineTools ./build.sh --test
# TESTS OK (8562 checks)
```

## Demo

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project are documented here. The format follows
 - Monitor alerts can now warn you when the battery stays above a temperature you choose. Under Monitor alerts, off by default. Thanks to @ywu73.
 
 ### Changed
+- Homebrew package rows can now be selected by clicking the unused space beside the name. Thanks to @pergioa.
 - Clipboard history now displays image previews and thumbnails for image files copied from Finder in the quick panel, preview inspector and menu panel.
 - Clipboard history now remembers whether you left the preview inspector open or closed across launches.
 - New Scratchpad tabs now start at 1 instead of leaving the first tab unnumbered, while existing names stay unchanged. Thanks to @AB-boi and @JashRashne.
@@ -36,6 +37,7 @@ All notable changes to this project are documented here. The format follows
 - Keep going with lid closed no longer fails to set up on accounts whose username contains @ or other non-alphanumeric characters, granting the rule by user ID instead. Thanks to @iltonandrew and @dhruvsaxena1998.
 - Uninstallation via script or inside the app now fully clears the Fan Control helper daemon registration, stored application data, caches and ByHost preferences. Thanks to @mugurc.
 - The menu bar panel keeps its arrow under the icon when the bar hides itself and the panel content changes height. Thanks to @pergioa.
+- Homebrew formulas installed from another tap now show their update on the row. Thanks to @pergioa.
 
 ## [3.3.3-beta.2] - 2026-08-22
 
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
@@ -754,12 +754,16 @@ enum HomebrewParser {
     private static func parseFormula(_ item: [String: Any]) -> HomebrewPackage? {
         guard let name = item["name"] as? String,
               HomebrewCommandBuilder.isValidToken(name) else { return nil }
+        let fullName = item["full_name"] as? String
+        let identifier = fullName.flatMap { fullName in
+            HomebrewCommandBuilder.isValidToken(fullName) ? fullName : nil
+        } ?? name
         let installed = item["installed"] as? [[String: Any]] ?? []
         let installedVersions = installed.compactMap { $0["version"] as? String }
         let stable = (item["versions"] as? [String: Any])?["stable"] as? String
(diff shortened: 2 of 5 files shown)


PathGao:
Parser fix looks right, and the test isn't vacuous: if `parseFormula` regressed, `tappedFormula` goes nil and the `$0.id` lookup fails the expect instead of skipping it.

**Casks likely have the same split.** `brew info --json=v2 --installed` gives casks `token` and `full_token`, and `brew outdated --json=v2` reports tapped casks by full token. That is the identical mismatch this PR fixes for formulae. `parseCask` isn't in the diff so I couldn't check whether it already keys on `full_token` — if it doesn't, a cask from a third-party tap still counts toward the update total with a row that offers no update.

**The hit target change isn't in the body.** Summary and Verification cover only the identifier fix; the two `.contentShape(Rectangle())` hunks appear only in the title. After them the full row width triggers the button, including the blank space that used to be dead. Worth a line saying what that button does — widening the target to the whole row reads differently if the primary action is install/uninstall rather than showing details.

Two things I couldn't verify: `isValidToken` isn't in the diff, so the fix rests on it accepting `/` (your passing test says it does, I just didn't see it). And unless formula names are never persisted, anything on disk keyed by `name` — pins, favorites, exclusions — changes key for tapped formulae with this, which the body doesn't mention.

Not flagging the `.buttonStyle` / `.frame` reorder, no functional difference.

**Impact:** none beyond the intended fix on the parser side. The hit target change makes the full row width actionable in both the panel and settings.
**Merge:** needs a yes/no on whether `parseCask` already uses `full_token` first.

pergioa:
Confirmed: `parseCask` and `parseCaskRecord` were using the short `token`. Both now prefer a validated `full_token` with a `token` fallback, and the regression fixture covers matching installed and outdated tapped casks. `isValidToken` already accepts `/`. I also confirmed there are no app-persisted Homebrew package identifiers requiring migration. The row hit target only selects and loads package details; the separate action controls are unchanged, and the PR description now clarifies this.

Verification: `env DEVELOPER_DIR=/Library/Developer/CommandLineTools ./build.sh --test` → `TESTS OK (8562 checks)`, up from 8557 (5 new assertions for the tapped-cask fixture and `parseInstalledCaskRecords`).

PathGao:
That closes everything I had open.

One surface I still can't see: anything that takes `HomebrewCaskRecord.token` or `HomebrewPackage.name` and uses it as a lookup key against brew's own output or the filesystem layout, rather than as a command argument. `brew list` prints short names and Cellar/Caskroom directories are keyed by the short name, so a consumer doing either comparison would now miss on tapped packages only — and no JSON fixture would catch that. Persistence you covered; this is the other half of it and it isn't in the diff, so I'm noting it as unchecked, not as broken.

The prefer-validated-full-else-short idiom is now written three times in the same file. A local helper makes the fourth parser correct by default:

```swift
private static func canonicalToken(_ item: [String: Any], full: String, short: String) -> String? {
    guard let s = item[short] as? String, HomebrewCommandBuilder.isValidToken(s) else { return nil }
    return (item[full] as? String).flatMap { HomebrewCommandBuilder.isValidToken($0) ? $0 : nil } ?? s
}
```

Non-blocking.

**Impact:** tapped formulae and casks carry their fully qualified identifier everywhere downstream of the parser; core packages are unchanged.
**Merge:** ship it.

Issue #731, opened 2026-08-20
Update does not work with EPM


dbhorst:
### Description

My company uses [BeyondTrust Endpoint Privilege Management](https://www.beyondtrust.com/de/products/endpoint-privilege-management) snake oil to control installations on macOS. It looks like Vorssaint does not "the default way" to update itself. The EPM gets triggered, but the installation does not finish and Vorssaint does not start after updating. When started manually it's still the old version.

Maybe it's enough to ship the installer inside the bundle. Move installerScript() into Resources/vorssaint-updater.sh, copied to Contents/Resources by build.sh and sealed by the bundle signature. Then request the authorization in-process so the requester is the Vorssaint app. 

### Steps to Reproduce

1. Open Vorssaint
2. Start Update
3. EPM pops up for `osascript` (not Vorssaint), allow installation
4. Vorssaint quits
5. nothing happens, old version is still installed

### Screenshots / Recordings

<img width="562" height="507" alt="Image" src="https://github.com/user-attachments/assets/ee059a42-6258-40eb-aa56-3be3335f9dab" />

### Vorssaint Version

3.1.15

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

dbhorst:
Here is how it should look like with bruno as example.

<img width="562" height="507" alt="Image" src="https://github.com/user-attachments/assets/6932800d-c720-4286-a4e6-2641136bec0c" />

PathGao:
Fixed on main by `7c898254` "Fix updater authorization attribution": `ShellSupport.swift:92-94` adds an in-process authorization request, so EPM now sees the signed Vorssaint app as the requester instead of `osascript`. That is the part of your diagnosis about the wrong requester.

Where it is: v3.3.3-beta.1 and later. The current stable release, v3.3.2, does not have it, so this issue is not carrying a release check label yet.

The second half of your suggestion was **not** done: the installer script is still generated at runtime rather than shipped inside the bundle at `Contents/Resources` and sealed by the bundle signature. If your EPM policy keys on the script being a signed, sealed resource rather than on the requesting binary, the current fix may not be enough. Testing on the 3.3.3 beta would tell us which of the two it is, if your environment lets you install a pre-release.

dbhorst:
Unfortunately it did not fix the issue:
- Installed v3.3.3-beta1
- Updater want's to install beta2
- EPM popup with correct app
- Vorssaint exits, but did not update and start again

[...]

PR #1006, opened 2026-08-25
fix(update): run the installer in its own session

PathGao:
## Summary

The installer has to outlive the app it replaces, and both spawn paths left it inside this app's own session and launchd job. `nohup` was never the property that mattered — it makes the child ignore SIGHUP, it does not leave the session — so whatever tears that job down takes the installer with it and the swap never happens. The reporter's `.dmg` surviving on disk is that failure: the script was killed inside its `while kill -0 "$PID"` wait, before any exit path could clean up.

Both paths now start the installer in its own session, through one `DetachedProcess`:

- The user path uses `posix_spawn` with `POSIX_SPAWN_SETSID`, so the kernel makes the child a session leader before it execs and there is no window where it still belongs to us.
- The elevated path can only travel as one string through the administrator prompt, so it composes the same detachment as a command. macOS ships no `setsid(1)` and no `daemon(8)`, so it reaches `setsid(2)` through perl, keeping today's `nohup` form as the fallback if perl is ever absent — a child that may still be swept beats an update that cannot start at all. The command names the ~4KB script once with `set --` and reuses `"$@"` for both branches, so the two forms cannot drift apart. The branch is taken on `[ -x /usr/bin/perl ]`, never on an exit code: perl execs the payload, so the status the shell sees is the payload's own, and an `||` fallback would restart the whole installer — as root — on every non-zero exit inside the script.

`DetachedProcess.spawn` also passes `POSIX_SPAWN_CLOEXEC_DEFAULT` with `/dev/null` opened on 0, 1 and 2. `Process` already gave its children a clean descriptor table; a bare `posix_spawn` does not, and a child specifically built to outlive us is the worst place to hand over the app's open files. No concrete failure is claimed for it: the installer mounts the `.dmg` itself, after its `while kill -0 "$PID"` wait, so there is no volume of ours for an inherited descriptor to hold busy, and the download handle is closed on the completion path before the install starts. What remains is that anything a child does inherit is then held open by a process nobody can see or close. The `/dev/null` opens are not decoration — `CLOEXEC_DEFAULT` on its own leaves 0/1/2 closed, and the child's first `open()` takes one of their slots, which puts a data file where stdout is meant to be. Measured here: with the flag alone, a directly spawned `perl`'s first `open()` landed on fd 2. Nothing reads these children's output — the installer reports through its marker file, `SelfUninstall` and `BundleMigration` redirect their own noise, and the elevated command already ends in `>/dev/null 2>&1` — so no diagnostic is lost.

Considered and rejected: `nohup` plus `disown` (same session, same sweep); `launchctl submit` (needs a label and leaves a job behind to clean up); a double-fork from Swift (`posix_spawn` already exposes the flag, so the fork dance buys nothing). For stdio, dup'ing the app's own 0/1/2 into the child the way `Process` does was rejected: those handles would then be held open past our exit, and `/dev/null` is what the elevated command already redirects to, so both paths now behave the same.

Four sibling sites spawn in-session children with the same requirement, and each is a one-line change onto the same primitive, so they are converted here rather than left as the last users of a pattern this PR replaces:

- `SelfUninstall` — the script waits for the app to exit before moving the bundle to the Trash.
- `BundleMigration` — the same wait before renaming the bundle after a rebrand.
- `AppDelegate.relaunchApp` and `FeatureRuntime`'s developer-variant relaunch — both `sleep` and then `open`, after `NSApp.terminate`.

Not changed: the installer script's own text, the authorization attribution fixed earlier, and the failure reporting — a spawn error still reaches `failInstall` carrying its errno description.

## Verification

MacBook Pro (Mac17,3, M5), macOS 26.5.2 build 25F84, built from source with the repo's own signing identity, single built-in display.

```sh
./build.sh --test              # TESTS OK (8857 checks)
```

Detachment checked at the process level rather than through the UI: a child spawned by `DetachedProcess.spawn` reports `getsid(child) == child` and a session id different from this process's; the composed elevated command was run in a real `/bin/sh` and its payload, exiting non-zero, ran exactly once. Descriptor handling is checked the same way: the suite opens a file without `O_CLOEXEC`, spawns a child that tries to write through that descriptor number, and asserts the write fails and the file stays empty, while the child's own stdout is still writable. Each new assertion was confirmed against the old behaviour — reverting to a bare `posix_spawn` turns 4 checks red, keeping `CLOEXEC_DEFAULT` without the `/dev/null` opens turns 1 red, and restoring the `||` fallback turns both rerun checks red, the payload having run twice.

The rerun count is taken at the end of the suite rather than after a fixed wait, so a second run that arrives late still fails the check instead of passing quietly; a deliberately delayed rerun was confirmed to fail it.

**Not tested.** No end-to-end update was run — nothing here exercised a real download, quit, swap and relaunch in either path. The four converted sibling sites were not driven either: no uninstall, no bundle rename after a rebrand, and neither relaunch was performed. They are covered only by the primitive's own tests and by the fact that their argument vectors are unchanged. The elevated path was never driven through a real administrator prompt, so root's environment, the authorization trampoline, and whatever descriptor table it hands the elevated shell are unverified — `posix_spawn` file actions do not reach that path, and the composed command's `>/dev/null 2>&1 &` covers only its stdio. Most importantly, this proves the child leaves the session; it does not prove that a new session is what EPM's sweep respects. If EPM kills by process tree or audit session rather than by session id, the installer still dies and the reporter sees the same symptom. No EPM-managed Mac here to check that on.

`/usr/bin/perl` is present on macOS 26.5.2, but Apple has been removing bundled scripting runtimes; if a future macOS drops it, the elevated path quietly reverts to today's behaviour rather than failing loudly.

## Anything else

No user-facing strings were added, so no localization changes.

Behaviour change worth naming for release notes: an update that is interrupted by a session teardown now completes instead of leaving the `.dmg` behind, and the uninstall, bundle rename and relaunch paths survive the same teardown.

Refs #731

### DIFF
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -1504,10 +1504,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
     /// process, so this is how the uninstaller picks up a just-granted grant.
     func relaunchApp() {
         let path = Bundle.main.bundlePath
-        let task = Process()
-        task.executableURL = URL(fileURLWithPath: "/bin/sh")
-        task.arguments = ["-c", "sleep 0.3; /usr/bin/open \"$1\"", "vorssaint-relaunch", path]
-        try? task.run()
+        // Its own session: the reopen fires after we terminate, so the child
+        // has to outlive the session it was started from.
+        _ = try? DetachedProcess.spawn(
+            "/bin/sh",
+            ["-c", "sleep 0.3; /usr/bin/open \"$1\"", "vorssaint-relaunch", path])
         NSApp.terminate(nil)
     }
 
--- a/Sources/Vorssaint/App/FeatureRuntime.swift
+++ b/Sources/Vorssaint/App/FeatureRuntime.swift
@@ -36,10 +36,9 @@ final class FeatureRuntime: ObservableObject {
     /// exits, so the fresh instance starts without the uninstalled features.
     func relaunchApp() {
         guard let bundleID = Bundle.main.bundleIdentifier else { return }
-        let task = Process()
-        task.executableURL = URL(fileURLWithPath: "/bin/sh")
-        task.arguments = ["-c", "sleep 0.6; /usr/bin/open -b '\(bundleID)'"]
-        try? task.run()
+        // Its own session: the reopen fires after we terminate, so the child
+        // has to outlive the session it was started from.
+        _ = try? DetachedProcess.spawn("/bin/sh", ["-c", "sleep 0.6; /usr/bin/open -b '\(bundleID)'"])
         NSApp.terminate(nil)
     }
 
--- a/Sources/Vorssaint/Core/BundleMigration.swift
+++ b/Sources/Vorssaint/Core/BundleMigration.swift
@@ -73,10 +73,9 @@ enum BundleMigration {
             .appendingPathComponent("vorssaint-rename-\(pid)-\(UUID().uuidString).sh")
         do {
             try script.write(to: scriptURL, atomically: true, encoding: .utf8)
-            let task = Process()
-            task.executableURL = URL(fileURLWithPath: "/bin/sh")
-            task.arguments = [scriptURL.path, oldPath, newPath, "\(pid)"]
-            try task.run()
+            // Its own session: the script waits for this app to exit before it
+            // renames the bundle, which it can only do from outside our session.
+            try DetachedProcess.spawn("/bin/sh", [scriptURL.path, oldPath, newPath, "\(pid)"])
         } catch {
             try? FileManager.default.removeItem(at: scriptURL)
             return false   // could not stage the rename; keep running as we are
--- a/Sources/Vorssaint/Services/DetachedProcess.swift
+++ b/Sources/Vorssaint/Services/DetachedProcess.swift
@@ -4,8 +4,9 @@
 import Darwin
 import Foundation
 
-/// Starting something that has to outlive this app — the update installer,
-/// which replaces the very bundle we are running from.
+/// Starting something that has to outlive this app: the update installer, the
+/// uninstall and rename scripts, and the relaunch helpers — every one of them
+/// waits for this process to go away before it does its work.
 ///
 /// `nohup` is not enough: it only makes the child ignore SIGHUP. The child
 /// stays in this app's session and launchd job, so whatever tears that job
@@ -47,9 +48,18 @@ enum DetachedProcess {
     /// the command falls back to the previous nohup form — a child that may
     /// still be swept away beats an update that cannot start at all.
     ///
+    /// The branch is taken on whether perl is *there*, never on an exit code:
+    /// perl execs the payload, so the status the shell sees is the payload's
+    /// own, and an `||` fallback would rerun the whole installer — as root —
+    /// every time it exited non-zero. perl therefore carries the setsid(2)
+    /// failure case itself, degrading to nohup's ignored SIGHUP (a signal
+    /// disposition survives exec) instead of handing the decision back.
+    ///
     /// `quotedArgv` is the program and its arguments, already quoted for sh.
     static func detachedShellCommand(quotedArgv: String) -> String {
-        let setsid = "/usr/bin/perl -e 'use POSIX (); POSIX::setsid() > 0 or exit 127; exec @ARGV; exit 127'"
-        return "set -- \(quotedArgv); { \(setsid) \"$@\" || /usr/bin/nohup \"$@\"; } >/dev/null 2>&1 &"
+        let setsid = "/usr/bin/perl -e 'use POSIX (); "
+            + "POSIX::setsid(); $SIG{HUP} = \"IGNORE\"; exec @ARGV; exit 127'"
+        return "set -- \(quotedArgv); { if [ -x /usr/bin/perl ]; then exec \(setsid) \"$@\"; "
+            + "else exec /usr/bin/nohup \"$@\"; fi; } >/dev/null 2>&1 &"
     }
 }
--- a/Sources/Vorssaint/Services/DetachedProcess.swift
+++ b/Sources/Vorssaint/Services/DetachedProcess.swift
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Darwin
+import Foundation
+
+/// Starting something that has to outlive this app — the update installer,
+/// which replaces the very bundle we are running from.
+///
+/// `nohup` is not enough: it only makes the child ignore SIGHUP. The child
+/// stays in this app's session and launchd job, so whatever tears that job
+/// down takes the installer with it, and the swap never happens (issue #731,
+/// reported under Endpoint Privilege Management). Leaving the session is the
+/// property that makes the child survive, and setsid(2) is the only way to
+/// get it.
+enum DetachedProcess {
+    /// posix_spawn with POSIX_SPAWN_SETSID: the kernel makes the child its own
+    /// session leader before it execs, so there is no window where it belongs
+    /// to us. Returns the child pid. Throws the spawn errno so callers can
+    /// report the failure the same way `Process.run()` let them.
+    @discardableResult
+    static func spawn(_ executablePath: String, _ arguments: [String]) throws -> pid_t {
+        var attributes: posix_spawnattr_t?
+        posix_spawnattr_init(&attributes)
+        defer { posix_spawnattr_destroy(&attributes) }
+        posix_spawnattr_setflags(&attributes, Int16(POSIX_SPAWN_SETSID))
+
+        let argv: [UnsafeMutablePointer<CChar>?] =
+            ([executablePath] + arguments).map { strdup($0) } + [nil]
+        defer { for argument in argv { free(argument) } }
+
+        var pid: pid_t = 0
+        let status = posix_spawn(&pid, executablePath, nil, &attributes, argv, environ)
+        guard status == 0 else {
+            throw NSError(domain: NSPOSIXErrorDomain, code: Int(status),
+                          userInfo: [NSLocalizedDescriptionKey: String(cString: strerror(status))])
+        }
+        return pid
+    }
+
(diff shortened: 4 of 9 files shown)


PathGao:
Two things.

`Sources/Vorssaint/Services/DetachedProcess.swift:53` — the `||` fires on any nonzero exit, not on a missing perl. perl `exec`s the payload, so the status sh sees belongs to the installer. Every `exit 1` path in the script — `Sources/Vorssaint/Services/Update/UpdateInstallerSupport.swift:72` (fail-dmg-verify), `:75` (fail-tempdir), `:83` (fail-mount) — therefore re-runs the entire installer a second time under nohup, as root. Reproduced with the composed command in a real /bin/sh against a payload that exits 1: it ran twice. The verification only covered the missing-perl case, which is the one shape where the status does mean what the `||` assumes. Testing `[ -x /usr/bin/perl ]` up front separates the two conditions.

The other spawn sites are the same defect, not adjacent ones:

`Sources/Vorssaint/Services/SelfUninstall.swift:180` — the child's first statement is `while kill -0 "$PID"`, the same wait the report was killed inside, in this app's session.
`Sources/Vorssaint/Core/BundleMigration.swift:79` — same wait, same session.
`Sources/Vorssaint/App/AppDelegate.swift:1503` — `sleep 0.3; open`, in-session, must outlive the terminate on the next line.
`Sources/Vorssaint/App/FeatureRuntime.swift:42` — `sleep 0.6; open -b`, same.

Under the sweep this PR describes, a user who uninstalls or relaunches loses the helper the same way the reporter lost the installer, and BundleMigration's failure leaves the bundle un-renamed with nothing reopened. Each is one line against `DetachedProcess.spawn`; converting them here is what makes the primitive cover the class rather than the one report.

PathGao:
@PathGao review this

PathGao:
The `[ -x /usr/bin/perl ]` branch and the four sibling sites are both handled. Two things left.

**The body no longer matches the diff.** "Anything else" still says SelfUninstall, BundleMigration and relaunchApp "are left alone here" — the diff converts all three, plus a fourth in `FeatureRuntime.swift`. Verification covers one `spawn` and the composed elevated command; nothing exercised an uninstall, a bundle rename, or either relaunch. Four behaviour-changing sites arrive undeclared and unverified.

**File descriptors.** `Process` spawns with `POSIX_SPAWN_CLOEXEC_DEFAULT` and explicit actions for 0/1/2, so the child gets nothing else. `DetachedProcess.spawn` passes `nil` file actions and only `POSIX_SPAWN_SETSID`, so every non-`O_CLOEXEC` descriptor this app holds now travels into a child that is specifically built to outlive it — at all six call sites. The one that bites here is `UpdateService`: the installer is spawned while the downloaded `.dmg` is in play, and its own `hdiutil detach` then runs against a volume that a departed app's inherited fd can still hold busy. Note that adding `POSIX_SPAWN_CLOEXEC_DEFAULT` on its own is worse than leaving it — it hands the child 0/1/2 closed, so the first `open()` in the installer script lands on stdout. It needs `posix_spawn_file_actions_addopen` for `/dev/null` on all three alongside it. I read the flag `Process` uses from its documented behaviour, not from Foundation's source on this machine.

Not blocking: the rerun test rests on a 500ms `usleep`. A second run that starts after that window leaves the count at 1 and the test passes — the guard written for the exact `||` defect is the part that fails quietly. The `||` form would have rerun immediately, so 500ms probably holds in practice, but a rerun landing late should not read as a pass. Re-counting the ledger at the end of the suite instead of deleting `detachRoot` right after the wait costs nothing and has no timing bound.

**Impact:** six spawn sites now leak the app's open descriptors into children that outlive it; concrete trigger is `hdiutil detach` failing against a volume held by an inherited fd after the app has quit.
**Merge:** needs the file-actions decision on `DetachedProcess.spawn` first, and the body's "left alone" paragraph corrected to match the diff.

ref #731

PathGao:
@PathGao review it

PathGao:
Rebased onto `main` at `3186e6a9`. The merge commit is gone — three commits, and the diff is byte-identical to what was reviewed at `23ae50f0`, so the `[ -x /usr/bin/perl ]` branch, the four sibling sites and the `posix_spawn` file actions are all unchanged.

What did change is the body: its falsification counts were measured before the descriptor commit existed and were stale. Re-measured on this branch:

- reverting to a bare `posix_spawn` turns **4** checks red, not 2 (the two session checks plus the two descriptor ones)
- keeping `POSIX_SPAWN_CLOEXEC_DEFAULT` without the `/dev/null` opens turns 1 red — the probe reports `1 1`, stdout closed
- restoring the `||` fallback turns **2** red, and the ledger shows the payload ran twice

Two claims the body makes were also re-confirmed here rather than carried over. With `CLOEXEC_DEFAULT` alone, a directly spawned `perl`'s first `open()` landed on fd 2. And the end-of-suite rerun count does catch a late second run: with the payload forking a copy of itself three seconds out, the early check passes and only the end-of-suite one fails.

`./build.sh` ✓ 0 warnings, `--selftest` OK, `--test` OK (8857 checks).

@PathGao

PathGao:
The three things from the last round are all handled: the body's "left alone" paragraph is gone, the file actions are in, and the rerun count moved to the end of the suite. Rebase checks out — the diff matches what was reviewed.

One thing left, and it's mine to retract. The DMG rationale I gave last round does not hold.

`UpdateInstallerSupport.swift:64` — the script's first statement is `while kill -0 "$PID"`, and `hdiutil attach` is at `:77`, after it. The app never mounts anything; the script mounts, and only once we are gone. So there is no volume for an inherited descriptor to hold busy at `hdiutil detach` (`:150`) — the volume did not exist while we were alive.

What the child *does* inherit is real, but it is a different thing: `BoundedUpdateDownloadDelegate.fileHandle` (`UpdateService.swift:503`) is opened without `O_CLOEXEC` and closed in `deinit`, so if the session outlives the handoff, the write descriptor on the `.download` file — same inode as `Vorssaint-update.dmg` after the `moveItem` at `:283` — travels into the installer, and from there into the `hdiutil attach` that is trying to attach that very file. That is a backing-file descriptor blocking its own mount, not a volume held busy at detach. **I have not verified that `hdiutil attach` actually fails or degrades under an inherited write descriptor on its backing file** — I only established that the descriptor can be there and that the detach story cannot be.

The sentence is in three places and all three say the detach version: the body's "Concretely:" paragraph, `DetachedProcess.swift:20-21` ("the mounted DMG among them"), and the comment above the fd probe in `MetricsTests.swift`. `CLOEXEC_DEFAULT` stays either way — matching what `Process` already did, and not handing a child built to outlive us the app's descriptors, is reason enough on its own. What should not ship is a measured-sounding trigger that the script's own ordering rules out, in a doc comment that will outlive this PR.

**Impact:** none — behaviour is unchanged; the defect is a wrong causal claim in one doc comment, one test comment and the body.
**Merge:** needs the DMG rationale corrected in those three places first.

ref #731

PathGao:
Retraction accepted, and the replacement does not hold either — so both are gone rather than swapped.

The script's ordering rules out the detach story exactly as you say: `while kill -0 "$PID"` at `:64`, `hdiutil attach` at `:77`. The app never mounts anything, so there is no volume of ours for a descriptor to hold busy.

The download handle does not survive the handoff either. `didCompleteWithError` closes `fileHandle` as its first statement, before `complete(location:)` fires the completion that moves the `.download` into place and starts the installer (`UpdateService.swift:546-549`, `:566-571`). The `deinit` close is the crash-path backstop, not the normal one. So the write descriptor is already gone when `hdiutil attach` runs, and it would not have been inherited by anything.

All three sites now say only what is actually true: a child built to outlive us must not carry our open descriptors, because it does its work after we are gone and anything it inherited is then held open by a process nobody can see or close. `Process` gave its children a clean table; these get the same. No concrete failure is claimed for it, and the body says so.

`./build.sh` ✓ 0 warnings, `--selftest` OK, `--test` OK (8857 checks). Behaviour is unchanged — the diff against the last head is three comments and one paragraph.

Both Swift checks were green on the previous head when I looked.

@PathGao

PathGao:
Checked the three sites on this head: the `DetachedProcess.swift` doc comment, the fd-probe comment in `MetricsTests.swift` and the body's paragraph now all claim only that a child built to outlive us must not carry our descriptors. No DMG trigger anywhere. Nothing left of the detach story.

Two things I verified rather than took on faith.

**The sweep is complete.** `Process()` appears at 12 sites on `main`. Beyond the five this PR converts, the rest — `MediaService`, `JunkCleaner`, `SuperKeyMappingGuard`, `BoundedProcessRunner`, `HomebrewManager` (×2), `DiskImageInstallerService` — are run-and-wait: piped output, `waitUntilExit`, no `kill -0` wait, no `open` after an `NSApp.terminate`. Grepping `main` for `kill -0`, `NSApp.terminate` and `/usr/bin/open` turns up no sixth site with the outlive-us shape. The primitive covers the class, not just the report.

**`set --` adds no quoting round.** The old form was `/usr/bin/nohup /bin/sh -c '<script>' vorssaint-installer '<args>'`, parsed once by the elevated shell. `set -- <same text>` is also parsed once, and `"$@"` expands without re-splitting. So `shellSingleQuoted`'s `'\''` runs land exactly where they did, and the script's own quoted requirements at `UpdateInstallerSupport.swift:66` and `:116` come through unchanged. That was the piece most likely to break quietly and it doesn't.

One observation, not blocking. `Process` normalizes two things for its children and the body names one: alongside `CLOEXEC_DEFAULT` it also passes `POSIX_SPAWN_SETSIGDEF` and `POSIX_SPAWN_SETSIGMASK`, so its children start with default dispositions and an empty mask. `DetachedProcess.spawn` passes neither, so a signal this app has ignored (SIG_IGN survives exec) or blocked on the calling thread now travels into a child built to outlive us, at all five converted sites. I grepped: the only `signal(…, SIG_IGN)` in the tree is `Sources/FanControlHelper/main.swift:604-605`, a separate binary — so **I have no evidence the app ignores or blocks anything at spawn time**, and no failure to point at. Raising it because it is the other half of the `Process` parity the body claims, and staying silent would read as having checked it.

Still unverified from here, on top of what the body lists: none of the four converted sibling sites were driven, and I have no EPM Mac, so whether a new session id is what the sweep respects is still the open question.

**Impact:** none — this head is comments and body text against the reviewed code.
**Merge:** ship it on green. The two Swift checks read blank from where I sit; the head they last reported on differs from this one only in comments.

ref #731

Issue #326, opened 2026-07-21
Volume booster is causing a stuttering sound when playing videos


AB-boi:
### Description

Enabling volume booster while playing something on youtube adds a stuttering sound that doesn’t go away until I disable the volume boost. 

### Steps to Reproduce

Enable 150% volume boost for safari in Vorssaint. 
Open Youtube on safari
Play a video at full system volume

### Vorssaint Version

3.1.14

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

vorssaint:
3.2.0 puts a limiter on the boost instead of clipping the peaks, so loud parts stop crackling.

AB-boi:
@vorssaint Hey! Hope you're doing well!

Just wanted to say that this issue is still there as of 3.3.1 and hence needs to be reopened. I can still hear crackling while playing YouTube videos on Safari.

AB-boi:
latest update says:

> Audio boost now smooths loud peaks and Volume Mixer restores sound after a failed
adjustment. 

as of 3.3.2, the problem persists. just wanted to keep you informed, no worries/hurry!

AB-boi:
Problem still persists completely after the update.

**Note: Problem only noted on Safari playing youtube. Helium browser playing youtube doesn't have the stutter issue.**

PR #1007, opened 2026-08-25
fix(mixer): unblock teardown and silence unwritten output frames

PathGao:
## Summary

Two defects on the tap-and-replay chain. For every app whose volume the user adjusts we create a process tap with `mutedWhenTapped` plus a private aggregate device that replays the samples at the chosen gain — the original is muted by CoreAudio, and what the user hears is our copy.

**The output buffer was replayed as the device found it.** `render` wrote `min(tap, output)` frames and left the remainder untouched, and a cycle that could not find its tap buffer returned before writing anything at all. That buffer arrives holding whatever CoreAudio last put in that memory, so those frames played back as a fragment of older audio (#326). `render` now silences whatever it does not write, and the no-tap path silences the whole buffer before it returns. Silence is the correct content for a frame with no samples behind it.

Nothing was added to the render path's cost: when the tap fills the buffer exactly, silencing computes one length per buffer and skips — no allocation, locking or logging inside the callback.

**One parked destroy blocked every other teardown.** `AudioHardwareDestroyProcessTap` on a wedged tap parks inside the HAL. Serialized, that one call left every later engine's aggregate device and tap object alive behind it for as long as the app ran — phantom aggregates accumulating in CoreAudio, and the wedge recovery's own tap never destroyed. Teardown now runs on an `OperationQueue` with `maxConcurrentOperationCount = 4`.

Not a plain concurrent `DispatchQueue`: N wedged taps would strand N workers of the shared 64-thread pool, which is the exhaustion this repo already paid for in #971 (`BoundedProcessRunner.swift:62-70`, `WindowEnumerator.swift:35-38`), and "bounded by live engines" is not the real bound — every rebuild enqueues its own teardown, so a persistently broken HAL path accumulates them for the life of the process. Not a `DispatchSemaphore` around the block body either: a waiter on a concurrent queue holds a worker while it waits, so the cap would not cap threads. `OperationQueue` keeps operations past the bound in the queue holding no thread, so the pool loses at most 4 workers however often the HAL parks. A full cap does reproduce head-of-line blocking, but confined to teardown instead of starving every `.utility` consumer in the process.

The blocked queue never held a mute. `stop()` calls `AudioDeviceStop` synchronously on the caller before anything is enqueued, and `mutedWhenTapped` suppresses the app's own output only while the tap is being read — the queue's work is destruction of objects, not un-muting.

## Not changed

- **The `isPlaying` guard on the wedge verdict.** An earlier revision of this branch removed it, so that a wedged engine would be caught while its app is paused rather than at the moment it resumes. That rests on a hardware fact I can argue but not measure here: the aggregate's main sub-device is the real output device, so its IO proc is clocked whatever the app does, but whether the tap keeps presenting a buffer of its announced shape through a pause is exactly the part no test here touches. If it does not, every paused row wedges after 1.5s, `MixerEngineRecovery` spends both of its attempts on an identical configuration, and the app returns at 100% with the slider doing nothing. Against that, the guard costs at most the 1.5s window after the user presses play — a paused app is silent either way, and `IsRunningOutput` goes true before anything is audible. Not worth trading a permanent failure for that, so the guard stays until someone can measure the premise.
- **Where the cycle counter sits.** Still below `guard frames > 0`, so a cycle counts only once the engine has handed the device audio. The new early return on a missed tap silences the buffer and leaves without counting, which is what the wedge recovery needs: a callback that never resolves its tap has done nothing for the app that a callback which never ran would not have done, and `tapChannels` is read once when the engine is built, so input that never carried that shape never will. Both halves are now pinned rather than left to the next edit.
- **`render`'s return value.** Returning `frames` on a silence-only cycle would look like a successful render to both the wedge counter and the limiter, masking a real fault.
- **Per-channel silencing.** `silence(from:)` answers whole frames, but an output channel with no source behind it is a hole *inside* the frames the tap filled. The mapping pass already clears those itself (`MixerRender.swift:141`), pinned by an existing check (`[1, 2, 0, 0, 3, 4, 0, 0]` over storage pre-filled with a sentinel).
- **The mono-fold and channel-mapping algorithms.**

## Verification

MacBook Pro (Mac17,3, M5), macOS 26.5.2 build 25F84, built from source with the repo's own signing identity, single built-in display, built-in speakers.

```sh
./build.sh --test              # TESTS OK (8855 checks)
```

Each change was broken on purpose to confirm the assertions fail on the old behaviour: deleting the tail-silencing turns 2 checks red, dropping the missed-tap silencing turns 1 red, reverting the teardown queue to the serial `DispatchQueue` turns 1 red, and moving the cycle count above `guard frames > 0` — the placement an earlier revision of this branch carried — turns 1 red.

**Not tested.** No wedged engine or dropout was reproduced on real hardware. The frame math is covered as a pure function, not by driving CoreAudio, and no measurement backs the claim that `AudioHardwareDestroyProcessTap` parks rather than returning late. The teardown queue's bound, the missed-tap silencing and the cycle count's placement are pinned as source shapes, since `AppVolumeMixer.swift` is not in the test target and the IO callback is a closure inside it; that the teardown blocks are race-free is by inspection — `stop()` is the only call site and it captures every id as a local before the fields are cleared. Nothing here was checked against Bluetooth output, an aggregate the user built themselves, or a device whose buffer size changes mid-stream.

The unwritten-output defect is not confirmed against the dropout report it came from and is not offered as its cause; handing a device memory nobody wrote is wrong on its own terms, which is why it is here.

## Anything else

No user-facing strings were added, so no localization changes.

#326 carries more than one report. This covers only the unwritten output frames; the clipping-distortion half was already addressed by the lookahead limiter, and the reporter's dropouts remain unconfirmed either way.

Refs #341
Refs #326


PathGao:
The detection case given up in the trade-off is a path where #341 still ends in permanent silence.

`Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift:1894` — with `cycles.increment()` above the tap lookup, an engine whose IO proc runs but never resolves a tap buffer counts as healthy for as long as it lasts. Before this change the counter froze there and `engineRenderVerdict` returned `.wedged` after the window, which tore the engine down and unmuted the app.

`Sources/Vorssaint/Services/Audio/MixerRender.swift:39-48` — that state is persistent, not transient, and is by design: with more than one input buffer and none matching `tapChannels`, the function returns nil every cycle rather than guess. `tapChannels` is read once at `AppVolumeMixer.swift:1889`, so an output device carrying its own input, or a tap format that changes after setup, stays unmatched for the life of the engine.

`Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift:1230` — nothing else rebuilds it. `tappedObjects` and `outputDeviceUID` are unchanged, so the guard continues and the engine is kept.

The app is still muted by `mutedWhenTapped` throughout, so the outcome is not "clean silence instead of garbage" — it is silence with no recovery, where the old behaviour was a short burst of stale audio followed by the tap being destroyed and the app restored. Removing the `isPlaying` guard is right on its own; keeping a second counter that only advances on a cycle that found its tap, and judging wedge on that, would give you both.

PathGao:
@PathGao reveiw this

PathGao:
The counter now sits after the tap lookup, which answers my earlier note. Three things left, one of which can kill the engine the recovery just rebuilt.

**The replacement engine inherits the dead one's observation.** `case nil` was the only thing in this hunk that removed a key from `engineRenderProgress`, and it's gone. A fresh engine starts at `renderCycles == 0`, and the state this PR newly calls wedged — IO proc runs, tap never resolves — leaves the old observation at `cycles: 0` as well. First pass after the rebuild:

```
previous = (cycles: 0, at: <recorded before the teardown>)
cycles   = 0                    // new aggregate has not run its first callback yet
now - previous.at >= 1.5   ->   .wedged
```

The new engine is condemned before it renders anything, and the same thing happens to the next one, so the two recovery attempts are spent in the time it takes to run the routing loop twice — the app ends up permanently untapped, which is the outcome the trade-off paragraph says it is buying protection against. `.stalled` and `.wedged` are outside the diff hunk, so I cannot tell whether one of them already clears the key. If it does, this is nothing. If not, `engineRenderProgress.removeValue(forKey: id)` on the wedged path is the whole fix.

**The concurrent teardown queue moves the parked thread rather than removing it.** Serialized, one parked `AudioHardwareDestroyProcessTap` held one thread plus a backlog of blocks. Concurrent, N wedged taps hold N threads that never come back, out of the shared GCD pool. That is #971's reported mechanism on this app: the pool fills with stranded blocks and something else deadlocks on it hours later. "Bounded by live engines" is a bound on how many threads can be lost, not a small number — it is one per app whose volume the user has touched. Cap the in-flight count (a `DispatchSemaphore` around the block body is enough). A full cap does reproduce the old head-of-line block, but confined to teardown instead of starving every `.utility` consumer in the process.

Also: the `teardownQueue.async` call sites are not in this diff, so "each block touches only the ids it captured" is not something I can check. Serial ordering between two blocks — an aggregate destroyed before the tap it contained, say — would also have been free before and is not now.

**`silence(from:)` is anchored on the frame index only.** In the split-output mapping path `written = frames`, so any output channel that received no source channel keeps whatever the HAL left in it for frames `0..<frames` — a per-channel hole inside the region the tail silencing considers written. That contradicts "the caller never has to think about the tail" only if `sourceChannel(for:sourceChannels:)` can leave a channel unmapped; the mapping loop is mostly elided in the diff and I did not verify that it can.

Minor: the source-text check in `MetricsTests.swift` asserts that two strings appear in that order somewhere in the file. It passes if `cycles.increment()` moves into the `else` branch of the tap guard, which is the exact regression it is named after, and it reads a path relative to the CWD. It does not pin the invariant it describes.

Nothing here was checked on hardware, same as the PR states.

**Impact:** The first one costs the wedge recovery both attempts on any app that reaches it, leaving the row muted until quit — the #341 outcome, on the path meant to fix it. The second one is latent and process-wide, and needs several taps to wedge before it bites.
**Merge:** needs the `engineRenderProgress` entry cleared when the engine is torn down (or a note that a branch outside the hunk already does it) first.

ref #341
ref #326

PathGao:
@PathGao review it

PathGao:
The three from last time are answered. On `main`, `AppVolumeMixer.swift:1218` and `:685` both clear the key, `MixerRender.swift:116` already clears unmapped output channels over `frames`, and `stop()` is the only `teardownQueue` call site with every id captured as a local before the fields are zeroed, so the concurrent blocks share nothing. The `OperationQueue` rebuttal is right and my semaphore suggestion was wrong: a waiter on a concurrent queue holds its worker.

One new thing, and it is in the hunk I asked you to change.

**The count also moved above `guard frames > 0`, and the body only describes the move relative to the tap lookup.** On `main` (`AppVolumeMixer.swift:1902-1903`) a cycle counted only if `render` wrote at least one frame. Now it counts as soon as the tap is found:

```swift
cycles.increment()
let frames = MixerRender.render(...)
guard frames > 0 else { return }
```

`render` returns 0 with a valid tap on three paths, all of them guards that exist because the case is considered possible: every output buffer's `mData` nil, a zero-frame output buffer, and `didWriteSourceChannel == false`. The last is device-layout-determined rather than transient — `firstOutputChannel += channels` runs in a `defer` ahead of the `mData` check (`MixerRender.swift:111-112`), so an output list of `[8ch mData nil, 2ch writable]` against a 2-channel tap offsets both writable channels past `sourceChannels`, `sourceChannel(for:)` returns nil for each, and every cycle clears and returns 0. For the life of that device.

An engine on that path now advances its counter, reads healthy, and the app stays muted until quit — #341's outcome, on a path the wedge recovery would have unstuck before this PR. It is also the same defect the body's second paragraph is about: a cycle the user hears nothing from counting as a live cycle. Tap-found is a weaker test than the invariant the PR argues for. `guard frames > 0 else { return }` already sits after the tap lookup, so putting the increment back below it satisfies both, and the protocol comment should then say what it counts — callbacks that handed the device audio, not callbacks that reached the tap. The new assertion pins the weaker position (`cycleCount! < renderCall!`) and flips with it.

**The `isPlaying` removal rests on a hardware fact the PR states but does not test.** "A started aggregate ... hands the tap the shape it announced, whether or not the tapped app is making a sound" is asserted in both the doc comment and the body; the Not-tested section covers the wedge, not this. It decides whether the PR is a fix or a regression. `MixerApp.isPlaying` means "emitting sound right now" (`AppVolumeMixer.swift:34-37`) while a row keeps its engine as long as it holds an audio connection, so every app the user has adjusted spends most of its life in exactly the state the old guard excluded. If a paused app's callbacks stop, or its input stops matching `tapChannels`, the new verdict wedges it after 1.5s, `engineRecovery` spends its one retry on an identical configuration, and the row stays untapped until a wake or a config change — the app returns at 100% with the slider doing nothing. Cheap to settle on the machine you already built this on: set a per-app volume, pause the app for five seconds, resume, check whether the slider still applies.

**Impact:** The count placement leaves a permanently silent app on any device where `render` cannot map a channel, which is the #341 outcome. The `isPlaying` removal is either the fix or a silent loss of per-app volume on every paused row, decided by a fact not yet checked.
**Merge:** needs the increment moved back below `guard frames > 0` first, and the paused-app check on hardware before the `isPlaying` removal lands.

ref #341
ref #326

PathGao:
Rebased onto `main` at `3186e6a9`, and both blockers are answered by dropping the change they were about rather than by patching it. Two commits now, +142/-11 where the reviewed head was +192/-43.

**The cycle count was never something this branch needed to move.** `main` already increments below `guard frames > 0` (`AppVolumeMixer.swift:1902-1903`). The first commit on this branch moved it to the top of the callback; the second moved it back only as far as the tap lookup, which is what you caught. There is no third position to argue about — the count is left where `main` has it, so a cycle counts only once the engine has handed the device audio, which is the stronger of the two invariants and the one the body was arguing for anyway.

**The `isPlaying` removal is dropped, and the whole wedge-verdict change with it.** `engineRenderVerdict` keeps its `isPlaying` parameter, its optional return and the `case nil` that clears the observation; `MixerRoutingSupport.swift` is out of the diff entirely. Your point about the premise is the reason. Half of it does hold — the aggregate's main sub-device is the real output device, so the IO proc is clocked whatever the app does — but whether the tap keeps presenting a buffer of its announced shape through a pause is exactly the half nothing here can measure, and it is the half the verdict rests on. The asymmetry decides it: the guard costs at most the 1.5s window after the user presses play, on an app that was silent either way, while the removal being wrong costs `MixerEngineRecovery` both attempts on an identical configuration and leaves the row at 100% with a dead slider until quit. Not worth trading a permanent failure for 1.5s, so it waits for someone who can measure it.

What is left is the two defects the title names. The teardown queue is unchanged from what you accepted — `OperationQueue`, `maxConcurrentOperationCount = 4`. The output-frame silencing is unchanged except that the missed-tap branch is now pinned: a source-shape check cuts that branch out by brace matching and asserts it silences before returning and does not count a cycle, and a third assertion holds the count below `guard frames > 0`. That last one is the assertion you said the old one should have been — it fails on the placement this branch used to carry.

`./build.sh` ✓ 0 warnings, `--selftest` OK, `--test` OK (8855 checks). Broken on purpose: deleting the tail-silencing turns 2 red, dropping the missed-tap silencing 1, moving the count above `guard frames > 0` 1, reverting the teardown queue to the serial `DispatchQueue` 1.

PathGao:
> Rebased onto `main` at `3186e6a9`, and both blockers are answered by dropping the change they were about rather than by patching it. Two commits now, +142/-11 where the reviewed head was +192/-43.
> 
> **The cycle count was never something this branch needed to move.** `main` already increments below `guard frames > 0` (`AppVolumeMixer.swift:1902-1903`). The first commit on this branch moved it to the top of the callback; the second moved it back only as far as the tap lookup, which is what you caught. There is no third position to argue about — the count is left where `main` has it, so a cycle counts only once the engine has handed the device audio, which is the stronger of the two invariants and the one the body was arguing for anyway.
> 
> **The `isPlaying` removal is dropped, and the whole wedge-verdict change with it.** `engineRenderVerdict` keeps its `isPlaying` parameter, its optional return and the `case nil` that clears the observation; `MixerRoutingSupport.swift` is out of the diff entirely. Your point about the premise is the reason. Half of it does hold — the aggregate's main sub-device is the real output device, so the IO proc is clocked whatever the app does — but whether the tap keeps presenting a buffer of its announced shape through a pause is exactly the half nothing here can measure, and it is the half the verdict rests on. The asymmetry decides it: the guard costs at most the 1.5s window after the user presses play, on an app that was silent either way, while the removal being wrong costs `MixerEngineRecovery` both attempts on an identical configuration and leaves the row at 100% with a dead slider until quit. Not worth trading a permanent failure for 1.5s, so it waits for someone who can measure it.
> 
> What is left is the two defects the title names. The teardown queue is unchanged from what you accepted — `OperationQueue`, `maxConcurrentOperationCount = 4`. The output-frame silencing is unchanged except that the missed-tap branch is now pinned: a source-shape check cuts that branch out by brace matching and asserts it silences before returning and does not count a cycle, and a third assertion holds the count below `guard frames > 0`. That last one is the assertion you said the old one should have been — it fails on the placement this branch used to carry.
> 
> `./build.sh` ✓ 0 warnings, `--selftest` OK, `--test` OK (8855 checks). Broken on purpose: deleting the tail-silencing turns 2 red, dropping the missed-tap silencing 1, moving the count above `guard frames > 0` 1, reverting the teardown queue to the serial `DispatchQueue` 1.

@PathGao

PathGao:
Both blockers hold on `main` as synced here. `AppVolumeMixer.swift:1902-1903` is `guard frames > 0 else { return }` then `cycles.increment()`, and this head's hunk stops at the tap lookup, so the count is where `main` has it. `engineRenderVerdict` and `MixerRoutingSupport.swift` are out of the diff. Nothing left to argue on either.

**The sweep, since a source-text assertion only pins its first match.** `AudioDeviceCreateIOProcIDWithBlock` occurs once in the tree (`AppVolumeMixer.swift:1890`), `TapGainEngine` is the only `GainEngine`, `MixerRender.tapBufferIndex(in: inputBuffers` occurs once, and `teardownQueue` has one call site (`stop()`, `:2039`). No sibling callback carries the unsilenced missed-tap exit, and each of the three new assertions pins the only instance rather than one of several.

**Every exit hands the device written memory.** Missed tap silences the list and leaves. `render` returning 0 has already silenced from `written == 0` — the whole list — before `guard frames > 0` sees it. A short tap silences the tail; an exact fill computes one length per buffer and skips. Unequal buffer lengths come out right too: `written` is the min across writable buffers while `silence` takes each buffer's own frame count, so a long buffer beside a short one gets its own tail cleared and the short one is skipped. Both limiters iterate `0..<frames` (`BoostLimiter.swift:218`, `:287`), so neither walks back into what was just cleared.

Nothing at tier 2 or above. Two smaller things, neither worth holding for:

- `expect(!missedTapBranch.contains("cycles.increment()"))` passes on an empty string, so a brace walk that fails to close reads green there. The sibling positive assertion goes red in that case, so the failure is still loud — just not at the line that names it.
- The missed-tap path now runs a full-buffer `vDSP_vclr` every callback for the life of a device whose input never matches `tapChannels` (read once at `:1889`), not once. Cheaper than what it replaces either way, but it is permanent rather than transient on that layout.

Not verified: nothing on hardware, same as the PR states. I did not run the build or the suite, and "Swift 6.0.3 compatibility" and "Build & selftest" had not reported when I read the page, so the 8855 figure is your run, not mine.

**Impact:** none I can find. Every exit from the callback writes the buffer, and the two changes the earlier rounds objected to are gone rather than patched.
**Merge:** ship it — glance at the two Swift checks first, they were still blank when I looked.

ref #341
ref #326

PathGao:
Took the first of the two. `missedTapBranch` is now checked non-empty in the negative assertion, so a brace walk that fails to close goes red at the line that names it rather than only at its sibling.

Left the second alone. The full-buffer `vDSP_vclr` on a layout that never matches `tapChannels` is permanent, as you say, but it is the cheaper half of that trade: the alternative on that layout is the device replaying whatever was in the buffer, every callback, for the same duration. A clear is one pass with no branch on content, and that engine is a wedge candidate anyway — the count stays frozen and the recovery tears it down. Nothing to gain from making the silence conditional.

`./build.sh` ✓ 0 warnings, `--selftest` OK, `--test` OK (8855 checks). Both Swift checks were green on the previous head; the only change since is the one assertion.

@PathGao

PathGao:
The non-empty check closes more than the one path I named. `missedTapBranch` was also empty when the anchor string itself missed — a reformatted call that splits `tapBufferIndex(in: inputBuffers` across lines leaves `missedTapBrace` nil, and the walk never runs. Both cases now go red at the line that names the invariant instead of only at its sibling.

On the second: my "permanent" was wrong, and not just in cost. The clear ends when the engine does. That branch skips the count, `engineRenderVerdict` still takes `isPlaying` and is still called with it (`AppVolumeMixer.swift:1205`, `MixerRoutingSupport.swift` untouched here), so a playing app on a layout that never matches `tapChannels` freezes its counter, reads wedged, and gets torn down. Bounded by `MixerEngineRecovery`, not by the life of the device. Leaving it alone is right for a stronger reason than the one you gave.

Nothing else moved. `main` is still `guard frames > 0 else { return }` then `cycles.increment()` at `:1902-1903`, this head's hunk stops at the tap lookup, and the three source-text assertions each pin the only instance of their shape.

Not verified: nothing on hardware, and I did not run the build or the suite — the 8855 is your run.

**Impact:** none.
**Merge:** ship it.

ref #341
ref #326

PR #1011, opened 2026-08-25
refactor: drop unreachable views and their orphaned helpers

PathGao:
Six declarations and one localized string have no reachable call site anywhere in the app. Each was confirmed by a whole-repo search that returns only the declaration line itself.

## What was dead

`WhatsNewView` was the auto-opening What's New window. Its last entry point went away in 3.1.0 — *"What's New no longer opens again after installing an update, because the update flow already shows the changelog before download"* — and the view stayed behind. The other windows in that file (`UpdateShowcaseIntroView`, `UpdateSupportIntroView`, `UpdatePreviewView`, and the shared `ReleaseNotesContent`) are still built from `AppDelegate` and are untouched.

`FeatureShowcase.swift` goes whole, not partially. `QuickPanelShowcaseStep` was its only entry point; `ShowcaseScaffold`, `HowToRow`, `HeroKey`, `HeroKeys` and `QuickPanelHero` exist only to serve it, and nothing outside the file names any of them. Removing just the step would have left 158 lines of scaffolding with no caller.

The rest are local leftovers: `AnimatedGIFView` in SharedUI (the update windows use their own `AnimatedFileGIF`), `PanelSubgroupDivider`, `AnyButtonStyle` — whose comment still describes *"the two button styles used above"*, both of which are gone — and `AutoQuitCloseSignal`.

## Trade-offs

**The localized string goes too.** `whatsNewDontShowAgain` was read only by `WhatsNewView`. Keeping it would have held this PR to 6 files instead of 18, but it leaves a memberwise-init field with twelve translations that no code reads — the same debt this PR is clearing, one indirection away. Removing it now costs one line per locale; removing it later costs the same plus the time to work out why it is unused.

**No rename.** `WhatsNewView.swift` no longer contains a `WhatsNewView`, but its three remaining windows are all part of the update flow, so a better name is a judgement call about that flow, not about this deletion.

## Effect

299 lines removed, no behavior change: nothing constructed any of these, so there is nothing to observe at runtime. Localization stays complete — the field is gone from the struct and from every locale in the same commit.

## Not changed

No renames, no moves, no reorganization of the files that lost code, and nothing removed on the strength of a static-analysis hint alone — everything here has a whole-repo search behind it.

## Verification

`./build.sh` green, `./build/Vorssaint --selftest` OK, `./build.sh --test` 8599 checks OK, on top of 241e4188.


PathGao:
Read the whole diff. One thing to flag.

Everything removed here is compiler-checked: deleting a declaration something still references is a build error, and CI is green, so the whole-repo searches confirm the result rather than establish it. The one thing that compiles either way is an unread `Strings` field — which is exactly why the body argues for taking `whatsNewDontShowAgain` along: *"it leaves a memberwise-init field with twelve translations that no code reads — the same debt this PR is clearing, one indirection away."*

That argument covers five more strings this deletion orphaned. `QuickPanelShowcaseStep` was the reader of `launcherName`, `launcherCaption`, `launcherOpenNow`, `launcherKeysHint`, `launcherEditHint`. The first two likely have other readers in the launcher or settings; `launcherKeysHint` and `launcherEditHint` read like showcase-only copy. Same search you already ran, five more times — and unlike the type removals, nothing will tell you later.

I only see the diff, so I have not verified any of that: no whole-repo search, no check for other readers of those five keys. Locale coverage is the one claim I can confirm without the repo — a missed locale would fail the memberwise init at compile time, and the build is green.

**Impact:** none. Nothing constructed any of the removed types.
**Merge:** ship it.

PR #1013, opened 2026-08-26
feat(settings): anchor search results to their exact destination

pergioa:
## Summary
- Settings sidebar search now presents macOS-style grouped results: each matching main Settings page appears once, with matching utilities and setting labels listed beneath it.
- Search results honor each configured Settings destination, including exact section anchors, rather than flattening every result to its containing page.
- Feature-owned setting labels are installation-aware. Fields that do not exist because their utility is not installed are omitted.
- Utilities remain discoverable and manageable through Features regardless of installation state:
  - Installed utilities keep their direct page or section result and also get a Features result that reveals the exact row for uninstalling.
  - Uninstalled utilities route to their exact Features row for installation instead of exposing unavailable setting fields.
  - Features whose intentional destination is the hub, such as Disk Image Installer, reveal their exact row as well.
- Search-result deduplication is structural rather than title-based. A dedicated page and generated feature merge only when their stable identities, full destinations, and one-to-one page mapping agree.
- Keyboard navigation with Up, Down, and Return drives the same grouped, destination-aware selection as mouse clicks.
- `SettingsRouter` carries one-shot destination and feature-target requests. Deep-linked sections scroll with minimal movement, while Feature Hub targets switch to the Features tab, center the utility row, and briefly highlight it while respecting Reduce Motion.
- Command Bar Settings results share the same destination routing and unavailable-feature fallback rules without adopting the grouped sidebar presentation.

## Test plan
- [x] `./build.sh --dev`
- [x] `git diff --check`
- [x] Grouping, ranking, keyboard selection, structural deduplication, and exact section routing coverage.
- [x] Installed utility coverage: direct result opens settings; Features result reveals the exact uninstall row.
- [x] Uninstalled utility coverage: utility result reveals the exact install row; owned field-level matches are omitted.
- [x] Shared-page coverage: available utilities remain searchable while fields owned by unavailable siblings stay hidden.

## Known issues
`./build.sh --test` compiles and passes all relevant assertions, but one pre-existing, unrelated timing precondition consistently fails on this machine:

```
TESTS FAILED (1 of 8799):
```

This is the setup check in the `BoundedProcessRunner` regression test at `Tests/MetricsTests.swift:10462`. It assumes the global utility queue supplies 64 simultaneously blocked workers within five seconds, which is not reliable in this environment. The failure predates this branch and the following subprocess regression assertion passes.

## Demo

https://github.com/user-attachments/assets/b3904dbf-be4b-43bd-9877-8316b0f96ef9


PathGao:
Three things.

**1. `SettingsSectionFocus.swift` drops `anchor: .center` for every existing deep-link.**

```swift
-            proxy.scrollTo(anchor, anchor: .center)
+            proxy.scrollTo(anchor)
```

The body says the new hub reveal uses "the same Form-layout retry timing `SettingsSectionFocusModifier` already uses" — it does not say the modifier itself changes. Every anchored destination that already worked (`quickTools/cameraPreview`, `screenshot/screenRecorder`, ...) now lands wherever `scrollTo` decides is minimal instead of centered, while the new Feature Hub reveal does center. Either this is a deliberate fix for jumpy anchor landing and belongs in the body, or it is a leftover from iterating on the hub reveal.

**2. `combinedItems` merges on localized title.**

```swift
$0.destination == pageItem.destination && fold($0.title) == fold(pageItem.title)
```

`SettingsSearchItem`'s own doc says its identity is "structural, never derived from localized text". This merge is derived from localized text. If a translation renders the hub name and the page name differently — which a separate `FeatureHubStrings` table invites — the Homebrew page result stops merging: two rows for one thing in the sidebar, and the page row loses `feature`, so an unavailable Homebrew falls back to a generic Features page instead of its row. The tests pin English titles through `settingsFeatureTitles`, so `--test` cannot see this.

`FeatureVisibilitySupport.isPageVisible(_:isAvailable:)` already knows the page→feature relation. Keying the merge on that instead of on titles removes the language dependency. That mapping is not in the diff, so I do not know whether it is 1:1 for the pages you need here.

**3. The feature branch drops the `pagesCoveredByActions` guard.**

```swift
case .page(let page):
    guard !pagesCoveredByActions.contains(page), ...
case .feature(let feature):
    guard feature.isAvailable, ...   // no pagesCoveredByActions check
```

A feature whose destination is a page a dedicated action already covers now gets a second Command Bar row. If that is intentional because the feature row carries an anchor the action does not, say so in the body — otherwise the guard belongs on both branches.

Smaller: the Command Bar filters unavailable features before routing, so `route(for:)`'s unavailable branch is unreachable from there. The shared-fallback claim holds for the sidebar only.

Not verified: I only see this diff. `hubTitle`, `pagesCoveredByActions`, `AppFeature.features(in:)` and the page→feature map are outside it, and I have not run the app — the reveal timing and `SearchKeyMonitor`'s field-editor test are exactly the parts a diff cannot confirm.

**Impact:** #1 changes scroll landing for every existing anchored Settings deep-link, all users. #2 gives duplicate sidebar rows and breaks the unavailable-feature fallback in any locale where the two names diverge. #3 duplicates a Command Bar row only if such a page exists.
**Merge:** needs #1 resolved first — revert the anchor change or state it in the body.

pergioa:
Thanks, I checked all four points and pushed the follow-up in `a644fa3`.

1. The `.center` removal is intentional and isolated in `f538cb2`: ordinary anchored Settings deep-links now reveal with minimal movement, while compact Feature Hub row targets remain centered. I clarified that distinction in the PR body.

2. The localized-title merge was a valid issue. `combinedItems` now merges structurally using stable page/feature identities, the exact destination, and the one-to-one page-to-feature mapping from `FeatureVisibilitySupport`. Translated title differences can no longer create duplicate rows or discard the feature identity needed for unavailable-feature routing. The tests use deliberately different titles and also verify that a real multi-feature Monitor destination does not merge.

3. I intentionally kept `pagesCoveredByActions` on generic page rows rather than applying it as a blanket feature-row guard. Anchor-specific feature results under General lead to sections the generic Open Settings action does not target. Cleaner, Uninstaller, and App Updates now structurally merge to their page IDs and are therefore suppressed by the existing page guard; regression coverage pins all three cases.

4. I corrected the routing wording in the PR body: both surfaces call `route(for:)`, but only sidebar search can exercise its unavailable-feature branch because Command Bar excludes unavailable features.

Validation: `./build.sh --dev` and `git diff --check` pass. `./build.sh --test` runs 8,620 checks and has only the already-documented unrelated dispatch-pool starvation precondition failure.

PathGao:
Re-checked all three against `main`, not just the tests.

1 and 3 hold. `FeatureVisibilitySupport.features(for:)` returns exactly `[.cleaner]`, `[.uninstaller]`, `[.appUpdates]`, and all three feature destinations are the unanchored page destination, so they do merge into page rows and the existing page guard suppresses them. Every other one-to-one page merges too — windowLayout, autoQuit, keyDebounce, superKey, textSnippets, shelf, media, urlCleaner, killProcess, radialMenu, commandBar, homebrew. No one-to-one pair is left behind, and the anchor change is stated in the body now.

One thing falls out of the pairs the structural rule deliberately refuses to merge.

`SettingsDirectory.swift:89`

```swift
SettingsDirectoryItem(page: .clipboard, title: FeatureStrings.clipboard(language).title, ...)
```

`FeatureHubSettings.swift:634`

```swift
case .clipboardHistory: return FeatureStrings.clipboard(L10n.shared.language).title
```

Same struct, same field — `"Clipboard"` in en-US, and identical in every other language by construction. `features(for: .clipboard)` is `[.clipboardHistory, .pastePlain, .finderCutPaste]`, so `shouldMerge` correctly declines, and both rows ship. Typing `clip` in sidebar search or the Command Bar returns two rows with the same label and the same rank: one lands on the page top, one on the Clipboard History section. Both destinations are right; only the label is ambiguous. Unlike the merge bug this one does not need a translation to diverge, and the fixtures can't see it because they invent titles.

Weaker, same shape: the six `monitor*` features all carry `FeatureSettingsDestination(.monitor)` with no anchor, so their new rows land exactly where the Monitor page row already lands. The first summary bullet says results deep-link to the exact section a result names — for those six they don't. Fine if that's the accepted state of the destination table, but it is the case the bullet claims to fix.

Not verified: anything needing the app running — the reveal timing, the highlight, and `SearchKeyMonitor`'s field-editor detection on the non-macOS-26 path. Both checks show no state on my side, so I'm taking `--test`/`--dev` from your report.

**Impact:** two rows labelled `Clipboard` in Settings search and in the Command Bar, every locale, every install. Routing is correct for both, so it is presentation only.
**Merge:** needs the Clipboard label collision resolved first — rename one of the two strings, or give feature rows their containing page as context so a section row never reads as its page.

pergioa:
Addressed in `ff9d13e`.

- The `.clipboard` page result keeps the existing localized Clipboard page title.
- The distinct `.clipboardHistory` feature result now uses the existing localized `sourceClipboard` label ("Clipboard history") in both Settings search and Command Bar, so the two correct destinations no longer render as identical rows.
- Structural identity and merging are unchanged: the page remains `.page(.clipboard)`, the section remains `.feature(.clipboardHistory)`, and the latter keeps its `.clipboardHistory` anchor.
- Regression coverage iterates every `AppLanguage` using the real Clipboard and Clipboard-history strings, verifies both rows remain present with different labels, and pins the anchored feature destination.

I also tightened the first summary bullet to say results honor configured destinations, including section anchors where available. The six `monitor*` destinations remain intentionally page-level until their destination table has dedicated section anchors, so the PR no longer overstates that case.

Validation: `./build.sh --dev` and `git diff --check` pass. `./build.sh --test` runs 8,788 checks with only the already-documented unrelated dispatch-pool starvation precondition failure.

PathGao:
`ff9d13e` closes the collision. `sourceClipboard` and `clipboard(_:).title` differ in all 13 language tables, and `shouldMerge` still declines `.clipboard` because `features(for: .clipboard)` is three features, so both rows survive with distinct labels.

I swept `main` for the same shape — a feature title equal to its containing page's title on a page `shouldMerge` refuses to collapse. The multi-feature pages are `.energy`, `.monitor`, `.mouse`, `.switcher`, `.clipboard`, `.cutPaste`, `.quickTools`, `.screenshot`. Nearest misses are `.cutPaste` ("Finder shortcuts" page vs "Rename shortcut" row) and `.screenshot` ("Screen capture" vs "Screenshot"). Clipboard was the only equality, so there is no second instance to fix.

What the fix leaves split is the name itself. `FeatureHubSettings.swift:634` still calls the feature by the page title:

```swift
case .clipboardHistory: return FeatureStrings.clipboard(L10n.shared.language).title
```

`hubTitle` is what the Feature Hub row, the Shortcuts page, onboarding, the radial menu tool list and the Command Bar toggle row all read from, so the app now says "Clipboard" in five surfaces and "Clipboard history" in the one this PR added. It shows on the path this PR ships: with Clipboard History switched off, searching `clip` offers a row labelled "Clipboard history", and selecting it opens the hub and highlights a row labelled "Clipboard". `clipboardHistory` has one enable key, so the Command Bar also lists "Turn off Clipboard" and "Clipboard history" side by side for the same feature.

Putting the string in `hubTitle` fixes every surface at once and takes the per-feature branch back out of `featureItems`, which is otherwise generic. Worth picking a string of its own while doing it — `commandBar(_:).sourceClipboard` belongs to the Command Bar's source filter, so rewording that picker now silently reworders a Settings row. Unless the hub row is deliberately "Clipboard", which I can't tell from the code.

Smaller: `SettingsView.swift:91`'s `SettingsSearchSupport.matches(query: searchQuery, ...)` is now always true — `normalSidebarList` only renders when `hasSearchQuery` is false, and a blank query matches everything. It reads like the plain sidebar still filters.

Not verified: unchanged from last round — reveal timing, the highlight, and `SearchKeyMonitor`'s field-editor detection on the non-macOS-26 path. I'm taking `--test` and `--dev` from your report.

**Impact:** one feature carries two names across Settings search, the Command Bar and the Feature Hub. Routing is correct on both, presentation only.
**Merge:** ship it.

PR #1014, opened 2026-08-26
test(monitor): skip the pool starvation check where the pool cannot be starved

iltonandrew:
Follow-up to #989, taken up at @PathGao's suggestion — `main` is currently red on high-core machines.

## What was wrong

The starvation check asserted its own precondition:

```swift
expect(occupiedWorkers == 64 && poolIsStarved,
       "the dispatch pool starvation this check needs was actually reached")
```

64 is where libdispatch *reports* the soft limit — the `Dispatch Thread Soft Limit: 64 reached` line in the #971 sample — not a cap it enforces. On a machine with enough cores it hands out a 65th thread, so blocking 64 workers doesn't starve the pool at all and the precondition can never hold. Measured on a 15-core M5 with a standalone harness replicating the setup exactly:

```
activeProcessorCount=15
occupiedWorkers=64   (test wants 64)     ✓
poolIsStarved=false  (test wants true)   ✗
```

Deterministic across runs. That failed the whole suite for a property the build simply cannot observe on that hardware — and since CI's smaller runners do starve, it only shows up locally, where it gets misattributed to whatever branch happens to be checked out.

## The change

Per @PathGao's call — skip rather than grow, since growing the loop only relocates the same case to a machine whose real limit sits above the ceiling, having spent a few hundred blocked threads getting there:

- The starved-pool assertion runs when the scenario can be built, and is skipped when it cannot. `occupiedWorkers == 64` still asserts unconditionally, so a failure to take the workers at all is still a failure.
- Paired with an absence rule, so the skip doesn't leave the check covering nothing: `BoundedProcessRunner.swift` must not contain `.global(`. The wider spelling catches `DispatchQueue.global(qos:)` and the type-annotated `let q: DispatchQueue = .global(qos:)` form alike. That is the invariant the fix actually rests on — the wait for a child must not be handed to the shared pool the runner's own spawns were filling — it holds on every machine, and it makes the starvation check a stronger conditional version of it rather than the only line of defence.

On your caveat: the file contains no `.global(` anywhere, so the rule needs no narrowing to the wait specifically.

## Verification

`./build.sh` zero warnings, `./build.sh --test` → **TESTS OK (8582 checks)** on the machine that was red, `./build/Vorssaint --selftest` → SELFTEST OK.

The absence rule bites: adding a bare `_ = DispatchQueue.global(qos: .utility)` to the runner fails it with `a bounded subprocess is never waited on from the shared dispatch pool`. Reverted.

### DIFF
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -9993,7 +9993,10 @@ struct MetricsTests {
             "/bin/echo", ["ready"], timeout: 1, maxOutputBytes: 1_024)
         let starvedPoolElapsed = Date().timeIntervalSince(starvedStarted)
         for _ in 0..<64 { poolGate.signal() }
-        _ = starvedProbe.wait(timeout: .now() + 5)
+        // Only when the probe never ran: its one signal was already taken above
+        // otherwise, and waiting for a second that no one sends costs the full
+        // timeout on every run of a machine whose pool cannot be starved.
+        if poolIsStarved { _ = starvedProbe.wait(timeout: .now() + 5) }
         expect(occupiedWorkers == 64,
                "the dispatch pool's blocking workers were all taken")
         // 64 is where libdispatch reports the soft limit, not where it stops:
@@ -10021,7 +10024,10 @@ struct MetricsTests {
             .split(separator: "\n", omittingEmptySubsequences: false)
             .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
             .joined(separator: "\n")
-        expect(!runnerCode.isEmpty && !runnerCode.contains("DispatchQueue.global"),
+        // Matching `.global(` rather than the one spelling: a queue bound to a
+        // local first (`let q: DispatchQueue = .global(qos:)`) reaches the same
+        // pool by a name this rule would otherwise never see.
+        expect(!runnerCode.isEmpty && !runnerCode.contains(".global("),
                "a bounded subprocess is never waited on from the shared dispatch pool")
 
         var networkDelta = NetworkProcessDeltaTracker(maxGap: 10)
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -10017,17 +10017,23 @@ struct MetricsTests {
         // to the shared pool, because the runner's own spawns are what filled
         // it (issue #971). Absence rather than the one call that was found
         // holding it, so a new one cannot reintroduce the freeze quietly.
-        let runnerSource = (try? String(
-            contentsOfFile: "Sources/Vorssaint/Services/BoundedProcessRunner.swift",
-            encoding: .utf8)) ?? ""
+        // Anchored to this file's compiled path rather than the process's
+        // working directory, so running the binary from anywhere still finds
+        // the runner instead of reading "" and blaming the pool invariant.
+        let runnerURL = URL(fileURLWithPath: #filePath)  // Tests/MetricsTests.swift
+            .deletingLastPathComponent()                 // Tests/
+            .deletingLastPathComponent()                 // repo root
+            .appendingPathComponent("Sources/Vorssaint/Services/BoundedProcessRunner.swift")
+        let runnerSource = (try? String(contentsOf: runnerURL, encoding: .utf8)) ?? ""
         let runnerCode = runnerSource
             .split(separator: "\n", omittingEmptySubsequences: false)
             .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
             .joined(separator: "\n")
+        expect(!runnerCode.isEmpty, "the bounded runner's source was found")
         // Matching `.global(` rather than the one spelling: a queue bound to a
         // local first (`let q: DispatchQueue = .global(qos:)`) reaches the same
         // pool by a name this rule would otherwise never see.
-        expect(!runnerCode.isEmpty && !runnerCode.contains(".global("),
+        expect(!runnerCode.contains(".global("),
                "a bounded subprocess is never waited on from the shared dispatch pool")
 
         var networkDelta = NetworkProcessDeltaTracker(maxGap: 10)
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -9994,12 +9994,35 @@ struct MetricsTests {
         let starvedPoolElapsed = Date().timeIntervalSince(starvedStarted)
         for _ in 0..<64 { poolGate.signal() }
         _ = starvedProbe.wait(timeout: .now() + 5)
-        expect(occupiedWorkers == 64 && poolIsStarved,
-               "the dispatch pool starvation this check needs was actually reached")
-        expect(!starvedPoolProcess.timedOut && starvedPoolProcess.status == 0
-                && String(decoding: starvedPoolProcess.output, as: UTF8.self) == "ready\n"
-                && starvedPoolElapsed < 0.5,
-               "a subprocess is watched off the dispatch pool, so a starved pool cannot strand it")
+        expect(occupiedWorkers == 64,
+               "the dispatch pool's blocking workers were all taken")
+        // 64 is where libdispatch reports the soft limit, not where it stops:
+        // on a machine with enough cores it hands out a 65th thread and the
+        // pool cannot be starved this way at all. Measured on a 15-core M5,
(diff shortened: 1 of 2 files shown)


PathGao:
Two things.

**The new check can fail for a reason that isn't the invariant.** `contentsOfFile:` with a repo-relative path resolves against the test process's CWD. Where that isn't the repo root, `try?` yields `""`, `!runnerCode.isEmpty` is false, and the failure reads `a bounded subprocess is never waited on from the shared dispatch pool` — the same misattribution this PR exists to remove, one level up. Your `_ = DispatchQueue.global(qos: .utility)` experiment proves the path resolves under `./build.sh --test`; it says nothing about other invocations, and `build.sh` isn't in the diff so I can't tell whether it pins CWD.

Anchor it to the test file instead:

```swift
let runnerURL = URL(fileURLWithPath: #filePath)  // Tests/MetricsTests.swift
    .deletingLastPathComponent()                 // Tests/
    .deletingLastPathComponent()                 // repo root
    .appendingPathComponent("Sources/Vorssaint/Services/BoundedProcessRunner.swift")
let runnerSource = (try? String(contentsOf: runnerURL, encoding: .utf8)) ?? ""
```

And split the two conditions, so a rename says rename rather than freeze:

```swift
expect(!runnerCode.isEmpty, "the bounded runner's source was found")
expect(!runnerCode.contains("DispatchQueue.global"),
       "a bounded subprocess is never waited on from the shared dispatch pool")
```

**The absence rule is narrower than the body says.** "so a new one cannot reintroduce the freeze quietly" holds for a new `DispatchQueue.global`, spelled that way, in that one file. A wait moved into a helper elsewhere passes it. I'm not asking you to widen the grep — one file is the right cost — but the body should claim what it covers.

Not verified: `BoundedProcessRunner.swift`, `build.sh` and the `expect` harness are all outside the diff, so I also can't see whether a skipped scenario reports as skipped anywhere or just quietly leaves the 8582.

Separately, `Point out a reference that will not reach the reporter` is red. I don't know what that check does, so I can't say whether it matters here.

**Impact:** none on shipped behaviour, test-only. Under an invocation whose CWD isn't the repo root, the suite fails pointing at the dispatch-pool invariant instead of at a file it couldn't read.
**Merge:** needs the source path anchored to `#filePath` first.

ref #971

iltonandrew:
Two fixes from an independent review.

**1. The absence rule checked one spelling.** `DispatchQueue.global` misses a queue bound to a local first — `let waitQueue: DispatchQueue = .global(qos: .utility)` reaches the same pool under a name the rule never saw, so the freeze could come back test-green. It now matches `.global(`. Proved by planting exactly that evasion in the runner: fails with `a bounded subprocess is never waited on from the shared dispatch pool`.

**2. The probe's signal was being waited on twice.** It signals once, and the non-starved read above already consumes it, so the final `starvedProbe.wait(timeout: .now() + 5)` was waiting for a second signal nobody sends — a full five-second stall on every run of a machine whose pool cannot be starved, which is precisely the machine this PR makes the check skip on. That wait now happens only when the probe genuinely never ran. The suite is back to ~53s here.

Also noted from the review and left alone deliberately: the comment stripping only removes whole-line `//` comments, so a block comment or string literal containing `.global(` in that one file would fail the rule spuriously. That is the house pattern used by every other source-shape assertion in this suite, and the failure direction is loud rather than silent, so I would rather not diverge from it here.

PathGao:
Neither of those touches the path. `contentsOfFile:` is still repo-relative in the diff, and the two conditions are still one `expect`. Under an invocation whose CWD isn't the repo root that still reads as the dispatch-pool invariant failing.

Widening to `.global(` makes the body's verification line stale: "the file contains no `DispatchQueue.global` anywhere" was checked against a spelling that is no longer the rule. State what the shipped rule matches and what you actually ran it against.

One thing from fix 2. `poolIsStarved` used to be an operand of an `expect`; it now decides whether two checks run at all. If it's written from a dispatched block rather than computed inline on this thread, an unsynchronized read there used to produce a loud wrong failure and now produces a silent skip. That code is above the hunk so I can't see which it is — if it's a plain local assigned on this thread, ignore this.

**Impact:** unchanged from my last read — test-only, fails under a non-repo-root CWD pointing at the wrong invariant.
**Merge:** needs the source path anchored to `#filePath` first.

ref #971

iltonandrew:
Done as asked, in c932daf — and the anchoring turned out to need one step more than the suggested shape.

**1. `#filePath` anchored, conditions split.** Your exact shape, plus a finding: `build.sh` compiled the test with a *relative* path (`Tests/MetricsTests.swift`), and swiftc then emits a relative `#filePath`, which `URL(fileURLWithPath:)` re-resolves against the runtime CWD — so the anchor alone would have been cosmetic. The swiftc invocation now passes `"$PWD/Tests/MetricsTests.swift"` so `#filePath` bakes in absolute. Proved end to end: the test binary run from an unrelated directory passes both pool assertions; pointing the URL at a non-existent file fails `the bounded runner's source was found` and not the invariant; planting `let waitQueue: DispatchQueue = .global(qos: .utility)` in the runner fails the invariant by name.

**2. Corrected body claims** (editing the body next): the shipped rule asserts that `BoundedProcessRunner.swift`, with `//` comment lines stripped, contains no occurrence of `.global(` — one file only; a wait moved into a helper elsewhere passes it. Verified against the current tree: `grep -n '\.global(' Sources/Vorssaint/Services/BoundedProcessRunner.swift` returns no matches.

**3. `poolIsStarved`** is a plain `let` assigned once on the test's own thread — `let poolIsStarved = starvedProbe.wait(timeout: .now() + 0.5) == .timedOut` (line 9990). The dispatched block signals the semaphore only; it neither captures nor writes the local, so no unsynchronized write exists and the skip cannot flip racily.

**4. For the record on your "other invocations" point more broadly:** the suite has ~30 other repo-relative `contentsOfFile:` reads with the same failure mode — running the binary from a non-repo CWD fails 45 checks while the two anchored ones pass. Happy to do that sweep as its own PR if you want it; it also needs the `$PWD` compile-path anchor this one adds.

Gate: build zero warnings, TESTS OK (8583), SELFTEST OK.

PathGao:
Anchor's in. The `$PWD` finding is the part I couldn't see: `build.sh` was outside the diff, so a relative `#filePath` would have made the anchor cosmetic. That closes my blocker.

On item 4, don't do the sweep. The failure I objected to was one check lying on its own — a lone `a bounded subprocess is never waited on from the shared dispatch pool` reads as a real regression. 45 failures at once doesn't. Nobody sees that and concludes the pool broke, they conclude they're in the wrong directory. The two checks worth anchoring were the two that could lie quietly.

Body still needs the edit you described. As it stands it says the rule matches `DispatchQueue.global` and cites verification against that spelling.

Not verified: the harness is outside the diff, so I can't tell whether anything pins the check total. It is now machine-dependent — this Mac skips the conditional and reports 8583, a runner whose pool does starve runs it and reports 8584. Check states come back blank to me, so your gate output is all I have on CI.

**Impact:** none on shipped behaviour, test-only.
**Merge:** ship it.

ref #971

iltonandrew:
Body edited — the absence rule is described as it is written now: `BoundedProcessRunner.swift` must not contain `.global(`, which covers the type-annotated spelling too.

Dropping the sweep, agreed on the reasoning — the two anchored checks were the two that could lie alone.

On the check total: nothing pins it, and that's inherent to the skip — the count differs by exactly one depending on whether the machine can starve its pool. The gate line in the body is from the machine that was red before this change.

PathGao:
A merge-order note, so it does not surprise anyone. The new check here asserts that `BoundedProcessRunner.swift` contains no `.global(`. #1080 (ZIP archive creation) adds `DispatchQueue.global(qos: .utility).asyncAfter` to that file for its SIGKILL escalation. Whichever of the two lands second goes red on this rule. Since the rule exists for the reason #1064 crashed, the sensible order is this one first and #1080 moving its escalation off the global pool. Noted on #1080 as well.

vorssaint:
closing this one.
the same failure was already fixed on main by #1374, which ramps the blocked workers until the pool really starves instead of skipping the check, so the skip here has nothing left to fix.
the new rule that the runner must never touch the shared pool would fail on today's main, because the runner now uses that pool for a delayed kill after a cancel, and that call blocks nothing.
neither hunk applies to main any more, and carrying them over would mean choosing between the two approaches again.
if the ramped check still cannot starve the pool on your mac, an issue with what it measured is the right follow-up

PR #1016, opened 2026-08-26
fix(auto-quit): stop the window walk at the application element

iltonandrew:
Follow-up to #995 for issue #953, opened at @PathGao's request — he spotted this walk in review there.

## Problem

A walk up `kAXParent` looking for an `AXWindow` reads `kAXRole` at each hop. The chain above a window *is* the application element, so an element that never yields a window — a menu bar item, a detached palette — walks straight onto it and asks it for a role.

## Why it matters

Reading `kAXRole` on a Chromium app's application element permanently switches its renderers into full accessibility mode: from then on the app rebuilds and ships an accessibility tree on every DOM change for the life of the process, a few percent of a core, with nothing querying it. Measured earlier on #953: an Electron app goes from ~13 reachable AX nodes to 156 with two `AXWebArea`s; Chrome from 17 to 1222.

#995 closed the door at attach time, which is where the enable-time sweep hit every running app. This one is narrower — it runs on a hit-test path, so it cannot produce the symptom Sirtx reported — but it is live: one unlucky hit test over the wrong element permanently wakes that app's renderers. Reading role on a *window* or on a hit-test element does not do this, so only the walk needed changing.

## The change

Build the application element once from the starting element's pid and stop the walk when a parent `CFEqual`s it, instead of reading its role. That hop already ended the walk with no window found — the application element's role is never `AXWindow`, and there is no `kAXParent` above it — so every input gets the same answer as before; the only difference is that the application element is no longer touched. In `AutoQuitService` the single caller treats the `nil` exactly as it already did.

**Two files carried the same walk.** `WindowMaximizer.swift` has it byte for byte on the green-button hit-test path, with the same hole. I have fixed both here rather than splitting them: it is one defect with one fix, and — as below — neither copy is reachable by the identifier-based scan, so shipping one would have left its twin silently live with nothing to catch it.

The other role reads in `AutoQuitService` were checked caller by caller and are all on windows or hit-test elements: `isStandardWindow` only ever sees elements from `kAXWindows`/`kAXMainWindow`/`kAXFocusedWindow` or an already-verified `AXWindow`, and the first three reads in `topLevelWindow` see the hit-test element and its window attributes.

## The rule

Per your suggestion, the per-file grep #995 left behind is replaced by a repo-wide assertion: no `kAXRoleAttribute` read anywhere in `Sources/Vorssaint` on an element assigned from `AXUIElementCreateApplication`. It is a text scan, not dataflow, and its comment says so — it misses an application element handed to a helper, stored in a property, or passed inline. Both direct reads found so far were written the way it catches.

That scan structurally cannot see this PR's bug, though: the walk reads `parent`, which is never *named* after an application element. So there is a second repo-wide rule beside it — a file that walks the parent chain reading a role must stop at the application element first. That is what makes this generic instead of a pin per copy, and it is what will catch the third copy if one appears.

## Verification

`./build.sh` zero warnings, `./build/Vorssaint --selftest` → SELFTEST OK, and `./build.sh --test` clean apart from the pre-existing high-core-machine failure that #1014 fixes (`the dispatch pool starvation this check needs was actually reached`) — this branch is based on main before that lands.

Both rules were proved to bite:

- adding a role read on an application element in `AutoQuitService` → `no application element is ever asked for its role: ["Services/AutoQuit/AutoQuitService.swift:163"]`
- removing the new guard from `WindowMaximizer` → `a walk up the parent chain stops at the application element: ["Services/WindowMaximizer.swift"]`

Both reverted.


PathGao:
The fix is right and can't regress: reaching the application element already ended the walk with `nil`, so the guard only makes that hop earlier.

**The two new rules assert the shape of the fix, not its effect.** Both are text scans over `Sources/Vorssaint`. If `CFEqual(parent, appElement)` never matches at runtime — a parent-chain application element compared against a freshly `AXUIElementCreateApplication`'d one — the guard never fires, the role read is back, and both rules stay green. I think the identity holds (`AXUIElement` equality is pid + element id, and the application element's id is fixed) but I can't confirm it from the diff and have no machine to check on. If a selftest can reach AX at all, one assertion that a walk from a menu bar item hits the guard covers what the greps structurally cannot.

**The second rule is narrower than the body says.** It triggers on the literal `role(of: parent)` and clears on the literal `CFEqual(parent, appElement)`, per file. So a third walk naming its variables anything else (`role(of: p)`, `CFEqual(parent, app)`) is invisible to it, and a file carrying two walks passes if either one has the guard. The body says "it is what will catch the third copy if one appears" — it catches a third copy that is a verbatim paste of these two. The first scan's comment spells out its limits; this one's doesn't, and it has the same kind.

**The rule polices a duplication that could be deleted instead.** Unless the two walks are deliberately independent, one shared walk taking the attribute reader as a parameter removes both copies and the second rule with them. `windowAttribute` vs `elementAttribute` is the only difference I can see between the two bodies.

Minor: `AXUIElementGetPid`'s `AXError` is discarded. On failure `pid` stays 0, `appElement` is nil, and the walk goes back to the pre-fix path with nothing distinguishing that from success.

Not verified: I only have the diff. The caller-by-caller claim about the other role reads (`isStandardWindow`, the first three reads in `topLevelWindow`) is unchecked — those call sites aren't in it.

**Impact:** none. The guard can only turn a walk that already returned nil into one that returns nil sooner.
**Merge:** ship it.

iltonandrew:
Independent review turned up two real holes in this; both fixed.

**1. The read above the walk.** The guard covered the parent walk but not the `role(of: element)` on the *starting* element, which runs first. If a hit test exposes nothing deeper and hands back the application element itself, the wake happened before the walk was ever reached — the fix had a door open next to the one it closed.

**2. The sentinel was tied to one pid.** It was built once from the starting element's pid, so a parent chain that crosses processes — an XPC-hosted view grafted into a host app, which is a real Chromium architecture rather than a hypothetical — would compare the host's root against the wrong pid's element, fail to match, and read its role anyway.

Both now go through one check that asks *each* element for its own pid and compares it against the element that pid builds, so it holds wherever the element came from and at whichever hop:

```swift
private static func isApplicationElement(_ element: AXUIElement) -> Bool {
    var pid: pid_t = 0
    AXUIElementGetPid(element, &pid)
    guard pid != 0 else { return false }
    return CFEqual(element, AXUIElementCreateApplication(pid))
}
```

**3. The scan didn't enforce what it claimed.** File-wide substring presence meant a guard written *after* the read, or a second unguarded walk in the same file, both stayed green. It now checks per occurrence and in order — each `role(of: parent)` must have the guard within the preceding lines — and reports `file:line`. Proved by moving the guard below the read in `WindowMaximizer`: `a walk up the parent chain stops at the application element: ["Services/WindowMaximizer.swift:264"]`.

On `CFEqual`: a direct probe confirms two separately-created application elements for the same pid are distinct objects that compare equal, which is what the whole approach rests on, and matches Apple's documented contract for comparing AX elements.

Gate: `./build.sh` zero warnings, `--selftest` OK, suite clean apart from the pre-existing pool-starvation failure that #1014 fixes.

PathGao:
`WindowMaximizer.swift:264` isn't where that read is. The new scan filters comment lines out of `lines` before enumerating, then reports `index + 1` against the filtered array. Post-patch the walk's `role(of: parent)` sits around line 292 — the hunk starts at 279, with six comment lines and the guard between — so the number is short by however many `//` and `///` lines precede the hit, roughly 27 here and more in a bigger file. The first scan enumerates raw lines and drops comments in the `where`, so its `AutoQuitService.swift:163` is fine; only the second one is off. The output you cited as proof the rule bites is the output that's wrong.

Keep the raw array and filter in the predicate instead:

```swift
for (index, line) in lines.enumerated()
where line.contains("role(of: parent)")
    && !line.trimmingCharacters(in: .whitespaces).hasPrefix("//") {
    let guarded = lines[max(0, index - 3)..<index]
        .contains { $0.contains("isApplicationElement(parent)") }
```

The window stays 3 for the code as written, since the guard is the line immediately above the read. A comment placed between them would now eat the window, so widen it if that bothers you.

One thing your rewrite didn't change: `AXUIElementGetPid` failing leaves `pid` at 0, `isApplicationElement` returns false, and the role read happens. It's an explicit fail-open now instead of an accidental one, which is better, but it's the same door.

Not verified: 292 is arithmetic off the hunk headers, not a read of the file — I only have the diff. I haven't checked what the filtered scan would report in `AutoQuitService.swift`.

**Impact:** none on the fix. When the rule fires on a future violation it names a line that isn't the violation, and whoever chases it lands ~30 lines short.
**Merge:** already merged, and the fix itself is right — this is a follow-up on the test's line numbers only.

iltonandrew:
Both points landed in #1044: the scan keeps the raw array and filters in the predicate — your shape verbatim — and the window stays 3, with one strengthening found while proving it: the lookback now ignores comment lines too, so a commented-out guard can't vouch for a live read. Mutation-proven both directions (planted read at 481: reported 448 before, 481 after; real guard commented out: scan fails naming `Services/WindowMaximizer.swift:292`). The pid==0 fail-open stays as it is, per your read.

Issue #725, opened 2026-08-20
Better Fan Control


lylythechosenone:
### Problem or Motivation

The fan control is currently very limited. Unfortunately, this is the case across all free (beer or speech) software: no FOSS app that I'm aware of allows customized fan control outside of manually setting the speed, and any freemium app that allows custom fan curves locks that feature behind the paywall.

### Proposed Solution

I propose we follow iStat Menus' example:
- you CAN set specific fan speeds (i.e. max)
- you can also set the fan to system control (i.e. disable custom fan control)
- but **most importantly**, you can set the fan to a custom fan curve:
  - you choose a sensor to follow (including meta-sensors like "average SoC" or "max SoC")
  - you manually set points on a temperature-fan speed graph
Full Bézier curves would be nice, but even just a 2-point piecewise linear would work (one for the minimum speed and the start of increasing, one for maximum speed and the end of increasing)

### Feature Area

System monitor

### Mockups / References

How iStat Menus shows fan control in the Sensors popout:
<img width="376" height="290" alt="Image" src="https://github.com/user-attachments/assets/5a9f8840-f782-4f1f-93dc-3ff57ad2fe9b" />
And how fan control curves are edited:
<img width="1096" height="340" alt="Image" src="https://github.com/user-attachments/assets/ef40d6f5-9a1f-492f-a178-ec48bee59e53" />

P.S. this is roughly mirrored in [this post](https://discord.com/channels/1539825206325936128/1540016848811204629/1540016848811204629) on the Discord server.

barbeldsgn:
Fan speed curves are a pretty cool idea, but you can also do this using the “priority list” option: 
What I mean is the ability to set many different criteria based on which the fans could adjust their speed 
Let’s say:
1. Average CPU temperature above 50 degrees – 1,300 RPM
2. Average CPU temperature above 80 – 2,500 RPM
3. Hottest GPU core – maximum speed

For this list, option 1 would have the highest priority, and option 3 the lowest 

The RPM range is 1,000–3,625 (max) - from what I can see on my Mac Studio
The only criteria you need to choose from are: the hottest CPU core, average CPU temperature, the hottest GPU core, and average GPU temperature—in my opinion, there’s no point in breaking it down into individual core temperatures, because for the average user it makes no sense, and the multitude of choices only creates confusion 

A simplier setup could involve setting the following criteria: 
1. Average CPU temperature:
[...]

PR #1018, opened 2026-08-26
fix(fan-control): allow adjusting active cooling speeds without verification failure

vorssaint:
### Summary
- Allow adjusting fan speeds during an active manual or curve cooling session by giving SMC target registers settling time and retrying verification before treating the change as failed.
- Assert manual mode alongside target RPM writes when updating cooling speeds.
- Remove fragile active targets count precondition in `updateCooling`.
- Add `targetRPMMatches` to `FanControlPolicy` and test RPM tolerance bounds.

### Validation
- Unit tests pass with 8,639 checks in `MetricsTests`.

### DIFF
--- a/Sources/FanControlHelper/main.swift
+++ b/Sources/FanControlHelper/main.swift
@@ -276,6 +276,7 @@ private final class FanControlController {
 
     private func updateActiveConfiguration(_ configuration: FanControlConfiguration,
                                            duration: TimeInterval?) -> FanControlResponse {
+        if hardware == nil { hardware = FanControlHardware() }
         guard let hardware,
               let level = requestedLevel(for: configuration, hardware: hardware) else {
             return .failure(.controlFailed, snapshot: currentSnapshot())
--- a/Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
+++ b/Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
@@ -141,12 +141,12 @@ final class FanControlHardware {
 
         if !directSucceeded {
             for (fan, target) in zip(fans, targets) {
-                guard setValue(target, for: fan.target, attempts: 10) else {
+                guard setTargetRPM(target, for: fan, attempts: 10) else {
                     throw FanControlHardwareError.operationFailed
                 }
             }
         }
-        guard verifyCooling(fans, targets: targets) else {
+        guard verifyCooling(fans, targets: targets, attempts: 10) else {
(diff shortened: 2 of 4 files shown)


PathGao:
Two of the four bullets under-describe what the diff does.

**`updateCooling` lost the mode precondition, not just the count one.** The body says "Remove fragile active targets count precondition" — the same guard also dropped `fans.allSatisfy { (try? modeValue($0.mode)) == 1 }`. Combine that with the `main.swift` hunk that builds `FanControlHardware()` when `hardware == nil`, and an update request arriving with no session in flight now creates hardware, writes mode=1 + target to every fan, and returns success. Both of those paths returned `.controlFailed` before. If `hardware == nil` is how the helper represents "no cooling session", this turns update into an undeclared start — one that skips whatever the `setCooling` caller arms for duration and revert. The session lifecycle isn't in the diff, so I can't tell whether the lazy init is deliberate (helper restarted while a session is live, hardware needs re-attaching). If it is, put it in the body and keep a guard so update can't be the first thing that ever puts a fan into manual.

**`writeManualPair` now swallows the mode write failure.** `try client.writeBytes([1], to: fan.mode)` became `_ = try?`, so it returns `true` after a failed mode write as long as the target write lands. Inside `setTargetRPM` that's harmless — it re-reads mode before returning. The `directSucceeded` path above line 141 isn't in the diff; if it treats `writeManualPair`'s return value as truth, it can now report success with fans still in auto. `verifyCooling` catches that a few lines later, so it's a wrong return value rather than a wrong outcome, but it's what makes the `try?` read as accidental rather than intended.

Minor, take or leave: `setTargetRPM` already re-reads target and mode before returning true, so `verifyCooling(attempts: 10)` matches on attempt 0 in every non-racing case — the second retry loop only pays off if the SMC reverts between the two reads. And the failure path now blocks ~0.5s per fan in `setTargetRPM` plus another 0.5s in `verifyCooling` before throwing; if the helper serializes requests, later ones queue behind that.

Moving the tolerance into `FanControlPolicy.targetRPMMatches` kills the second copy of that constant, which is the part of this PR I'd keep unchanged. The `isFinite` guard doesn't change behavior (NaN comparisons were already false), it just makes the case testable.

**Impact:** with both hunks in, a cooling update on a helper with no live session forces fans into manual mode and reports success, and nothing in this diff reverts them. The rest is behavior-preserving.
**Merge:** needs `updateCooling` to keep a guard that it can't be the first write into manual mode (or a body line stating that's intended) first.

ref #725

PR #1019, opened 2026-08-26
fix(recorder): update microphone terminology in localized strings

watain666:
## Summary

The Traditional Chinese (Taiwan) recorder strings used the Hong Kong Cantonese term “咪高風” instead of the Taiwan-standard term “麥克風”.

Updated the four `zhTW` recorder strings to use “麥克風” consistently:

- Microphone toggle
- Microphone track label
- Microphone unavailable HUD
- Microphone permission name

This is a locale-only wording correction. No recorder behavior, permissions logic, runtime dependencies, or supported locales were changed.

## Verification

Tested locally on a MacBook Air (`MacBookAir10,1`) with Apple M1, macOS 26.5.1 (25F80), using the built-in 2560 × 1600 Retina display. This was a local release build with ad-hoc signing.

```sh
./build.sh                     # passed without warnings
./build.sh --test              # TESTS OK (8638 checks)
```

I did not manually test the recorder UI in the `zhTW` locale, microphone recording, or the permission prompt on another Mac. No runtime behavior was changed by this fix.


PathGao:
Term is right — 麥克風 is the Taiwan standard, 咪高風 is HK Cantonese. The four zhTW recorder strings now agree with each other.

Two things the diff alone can't settle:

`git grep 咪高風` before merging. This is one extension in a 1600-line strings file, and the same term plausibly sits in other zhTW tables (permission copy, audio settings, menu bar). A locale that says 麥克風 in the recorder and 咪高風 in settings is worse off than one that is consistently wrong.

The body says no behavior changed. That holds only if nothing branches on these values. This repo has a track record of localized text being used as a decision criterion, so `microphonePermissionName` is the one to confirm — it reads like a display string but it is named like a key. If something does match on it, that is a pre-existing bug this PR would surface, not a bug this PR introduces.

**Impact:** none — string literals only, assuming no call site compares against them.
**Merge:** ship it.

Issue #1008, opened 2026-08-25
Quit on close does not work on safari and Activity Monitor


wako69420:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When I click on the red X icon on the last Safari tab, the application still stays open, so does Activity Monitor

### Steps to Reproduce

Open Activity Monitor try clicking on the X icon. The application still stays open in the menu bar where next to the Apple logo the settings show so does Safari

### Feature Area

Tools

### Screenshots / Recordings

https://youtu.be/YxVigeiQjXo

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

english

### Installation Method

github

iltonandrew:
Taking a look at this one — will try to reproduce on Safari and Activity Monitor here.

PR #1022, opened 2026-08-26
fix(auto-quit): watch again when Accessibility lists no window

iltonandrew:
Refs #1008.

## Problem

Quit on close does nothing for some apps — Safari and Activity Monitor in the report. Closing the last window leaves the app running, and no later close helps either; the app stays stuck for as long as it runs.

## Root cause

Whether an app may quit and what tells us to check are read from two different places. `refreshWindows` marks an app eligible if Accessibility lists a window **or** the window server still shows one, but the destroy notification that schedules every check is registered per window over the list **Accessibility** gave. When Accessibility answers with nothing while the window server still shows a window, the app is eligible with nothing watched, and the close that should quit it destroys a window nobody registered for.

Nothing recovers from that. The app-level notifications that would refresh the list all fire when a window *appears*, and the window is already there. Instrumented on a live build, the only notification such an app produces after its close is `AXApplicationDeactivated`, which by design does not schedule a check.

The empty answer is a successful one (`err=0, count=0`), not a timeout, so a longer messaging timeout would not have changed it. It is also the documented signature of a window parked on a Space that is not visible, which Accessibility cannot describe at all.

## Change

When a refresh registers no window but the window server still shows one, look again — three retries per round, at offsets 0.5s, 1.5s and 4.0s from the round's start, cleared whenever a refresh finds the app no longer eligible and on detach. The window-server sweep itself (a `CGWindowListCopyWindowInfo` round trip) now runs only when Accessibility handed back nothing, so the common case pays nothing new. Accessibility catches up on the first look in practice.

The second cause in the root-cause section — a window parked on a Space that is not visible — outlives any bounded round: those three looks all return empty and nothing used to ask again. A round is therefore re-armed when the active Space changes (`NSWorkspace.activeSpaceDidChangeNotification`), which is the moment such a window can become describable; apps stay dormant between Space changes rather than being polled for their lifetime. To be exact about the cost: a re-arm refreshes immediately and then schedules the three offsets, so a round after a Space change is four looks, and switching Spaces repeatedly starts one round per switch. Dormant between switches still holds, but it is thinner than "three retries" reads on its own.

The count that decides all this is the number of windows actually watched, not the number Accessibility listed: `AXObserverAddNotification`'s result is read, and a window counts as watched when the destroyed notification took, since that is the one the close path acts on. The comparison is against what Accessibility listed, not against zero: an app with two listed windows where one registration is refused would otherwise decline the retry on a non-zero count, and closing the unwatched one would notify nobody — this bug's own shape, one window later. Listing nothing while the window server shows one is still `0 < 1`, so the reported case is unchanged. The cost is that an app with a permanently unregisterable window retries every round and re-arms on every Space change, the same profile as the hidden-Space class above. Each retry carries a timer identity checked at fire time, so a re-armed round or a detach-and-reattach inside the delay makes the old timer a no-op instead of a duplicate refresh. The decision is a pure `AutoQuitSupport.needsWindowWatchRetry`, and the retry reads the live observer at fire time so it cannot register watches on a run-loop source that is gone.

## Verification

Build clean with zero warnings, `SELFTEST OK`, suite green apart from the known pre-existing pool-starvation check (#1014 fixes it).

Verified by fault injection on a live build, forcing the exact observed condition and toggling the retry: Activity Monitor survives its last window closing with the retry off and quits with it on; Safari's trace shows the retry registering the watch, the destroy notification arriving, and the app terminating. On the finished build with no injection, both apps quit when their last window closes. An ordinary two-window app still survives the first close and quits on the last. Three unit assertions cover the decision; source-shape assertions pin the short-circuit, the offsets and their one-origin scheduling, the bound, the stale-timer guard, the stop cleanup, and the Space re-arm wiring.

The Space-change re-arm is a candidate for the hidden-Space class: the wiring is pinned and the timer-identity path is exercised by the assertions, but I have not run a live park-on-another-Space E2E, so that class's recovery is claimed only as far as the code shows.

One note for anyone reproducing the Safari case by hand: Safari opens a Start Page window when its last one closes, so it correctly stays running until *that* window is closed too.


PathGao:
`hasWindowServerUserWindow` used to sit behind `||`, so it only ran when Accessibility listed nothing. It now runs on every refresh for every attached app, including the common case where AX already handed you windows. Nothing in the fix needs that:

```swift
let serverWindow = windows.isEmpty ? hasWindowServerUserWindow(pid: pid) : nil
```

Both uses are unchanged by it — `hadWindows` is already true from the first disjunct, and `needsWindowWatchRetry` returns false whenever `registeredWindows > 0` no matter what the second argument is. `hasWindowServerUserWindow` itself isn't in this diff, so I can't price the call; if it's a `CGWindowListCopyWindowInfo` sweep it's a window-server round trip added to every window-created and window-destroyed notification.

Body says "three times, at 0.5s, 1.5s and 4.0s". Each retry re-enters `refreshWindows`, which schedules the next one from *its* fire time, so the real points are 0.5s, 2.0s and 6.0s after the refresh, and the budget spans 6s rather than 4s. `closeCheckDelays` is two declarations above with the same shape and reads like offsets from one origin — the rest of `scheduleWindowChecks` isn't in the diff so I didn't verify that, but if it is, the two arrays look alike and mean different things.

The body names two causes for the empty answer and the bound only covers one. A busy app catches up on the first look, as you measured. A window parked on a Space that isn't visible is one Accessibility "cannot describe at all" — those three looks are spent inside 6 seconds and never return, so that class lands exactly where #1008 started, and by your own paragraph nothing re-triggers a refresh afterwards. Stopping rather than polling is the right call for the budget; what's missing is anything that re-arms it. A Space change is the obvious trigger, and `AutoQuitSupport` already reasons about Spaces elsewhere ("without Spaces to compare, the old cautious rule stands"), but the notification registration isn't in the diff so I don't know what's cheap to hook.

Smaller: `needsWindowWatchRetry(registeredWindows:hasWindowServerUserWindow:)` restates the eligibility test one line above it (`!windows.isEmpty || serverWindow == true`) with the sign flipped on the first term. Second copy of the same fact — add a third eligibility source later and the retry silently stops matching it.

**Impact:** every attached app pays a window-server query per refresh instead of only the apps AX has nothing for; apps whose window sits on a hidden Space still fail to quit on close once the first 6 seconds are gone.
**Merge:** needs the short-circuit restored first — separately, the reference check is red and build/selftest hasn't reported.

ref #1008

iltonandrew:
On it — restoring the short-circuit and working through the rest (timing math in the body, the hidden-Space class, the duplicated eligibility test). Will report the gate with the push.

iltonandrew:
All five, in 91f450f:

**Short-circuit** — restored in your exact shape. Confirmed `hasWindowServerUserWindow` is a `CGWindowListCopyWindowInfo` sweep, so what the draft was adding to every refresh was a window-server round trip; it now runs only when AX hands back nothing.

**Timing** — code now matches the body instead of the other way round: retries are offsets from one round origin (0.5 / 1.5 / 4.0, a 4s budget), held in a per-pid round state rather than chained reschedules. `closeCheckDelays` is now `closeCheckOffsets` and schedules from a captured origin too — behaviorally a no-op (the loop already scheduled all three from one `now()`), but the two arrays now read alike *and* mean alike, which was your point.

**Hidden Space** — the part the bounded round couldn't reach. `NSWorkspace.activeSpaceDidChangeNotification` re-arms one fresh bounded round for apps still holding retry state; between Space changes they stay dormant. Each scheduled retry carries a timer identity checked at fire time, so a re-arm or detach-and-reattach makes the superseded timer a no-op rather than a duplicate refresh. State clears on detach, on stop, and whenever a refresh finds the app no longer eligible. Phrased in the body as a candidate for that class — the wiring is pinned by shape assertions but I haven't run a live park-on-another-Space E2E.

**Duplication** — `foundUserWindow` is computed once and drives both uses; the helper now states only the missing-watch condition. The dropped nil-case unit test's invariant ("an unanswerable server is not a window") lives in the pinned `serverWindow == true` call-site line.

**Red check** — the body opened with `Fixes #1008`, which would close the issue on merge before any release reaches the reporter. Now `Refs #1008`; body also updated for the one-origin offsets and the re-arm.

Gate, unsandboxed: build zero warnings, `SELFTEST OK`, suite green apart from the known machine-dependent pool-starvation check (#1014 adds the skip).

PathGao:
All five land as described; I checked each against the diff. Short-circuit is back, both arrays are anchored to one origin, the timer identity makes a superseded round a no-op, and `foundUserWindow` is the single source both uses read.

Two things left, neither a hold.

`needsWindowWatchRetry(registeredWindows: windows.count, ...)` — the parameter is named for what got registered, the argument is what Accessibility listed. `watch(window:observer:refcon:)` returns nothing, so a failed `AXObserverAddNotification` never reaches this line. Where the two diverge you get #1008's exact shape — eligible, nothing watched — and the retry declines because the count isn't zero. `watch` isn't in the diff, so I can't tell whether it can fail or already handles it. Either have it report success and count those, or rename the parameter to what it actually holds. The cost of leaving it is the next reader trusting an invariant the call site doesn't supply.

The hidden-Space path rests on `hasWindowServerUserWindow` reporting a window parked on a Space that isn't visible. If that sweep passes `kCGWindowListOptionOnScreenOnly`, such an app is never eligible, never enters `windowWatchRetries`, and the re-arm has no state to re-arm — the class you couldn't E2E would also be the class the code can't reach. You confirmed it's a `CGWindowListCopyWindowInfo` sweep; if you saw the option set while you were in there, this is closed. I can't, it's not in the diff.

Smaller, not asking for a change: a re-arm does an immediate `refreshWindows` plus the three offsets, so a round after a Space change is four looks, and someone switching Spaces every few seconds restarts one per switch. Dormant between switches still holds, just thinner than the body reads.

**Impact:** none beyond the fix's own reach — an app whose windows Accessibility lists but the observer fails to register on still falls in the original hole.
**Merge:** ship it.

ref #1008

iltonandrew:
**The Space question closes, and I read past the option flags to be sure.** The sweep is `CGWindowListCopyWindowInfo([.optionAll, .excludeDesktopElements], kCGNullWindowID)` — no `.optionOnScreenOnly`, so off-screen windows are listed. It doesn't stop at listing them either: `AutoQuitService.swift:582-595` has an explicit off-screen branch that asks `SpaceWindowBridge.spaces(of:)` for the window's Spaces and the topology for the visible ones, then hands both to `AutoQuitSupport.offscreenWindowKeepsAppAlive`, which returns `SpaceHopSupport.isParkedOnHiddenSpace(windowSpaces:visibleSpaces:)` whenever it has a Space answer. So a window parked on a Space that isn't showing is exactly the case that makes the app eligible — it does enter `windowWatchRetries`, and the re-arm has state to re-arm. The class the code can reach and the class I couldn't E2E are the same class; what's still unproven is only the live behaviour, not the reachability.

**The count is a real divergence and I'm fixing it, not renaming it.** `watch(window:observer:refcon:)` discards the `AXError` from `AXObserverAddNotification`, so `registeredWindows: windows.count` is what Accessibility listed. Where a registration fails you get #1008's shape with the retry declining, which is the one hole the fix doesn't cover. Working through having `watch` report success and counting those — the destroyed notification is the load-bearing one, and an already-registered notification has to read as success or the count collapses on every refresh. Will push with the pin extended and mutation-proved.

**Your four-looks note is accurate** and the body oversells it slightly: a re-arm is an immediate refresh plus the three offsets, and a Space switch every few seconds restarts one round per switch. Dormant-between-switches holds, but it's thinner than "three retries" reads. I'll say so in the body rather than leave the reader to work it out.

iltonandrew:
Counted rather than renamed, in 92f3a64.

`watch` now returns whether the window ended up watched, and `refreshWindows` counts those into `registeredWindows` — so the parameter is true to its name by construction instead of by comment.

**Which notification decides.** The destroyed one, alone. It is the only one that reaches `scheduleWindowChecks` (`shouldScheduleWindowCheck` returns true for `.windowDestroyed`; the miniaturize pair only feeds `minimizedWindows` bookkeeping), so a window whose miniaturize registrations were refused is still a window auto-quit can act on — counting it unwatched would spend the app's bounded round on a non-problem. A window that got the miniaturize pair but not destroy is your divergence exactly, and it now counts as unwatched.

**Already-registered reads as success**, and that one matters: every caller of `refreshWindows` — `attach`, the five `windowRefreshNotifications` in `handleAX`, the retry timer, and the Space re-arm — re-runs `watch` over the same elements on the same observer, and nothing removes those registrations in between (`detach` drops the run-loop source and the dictionaries only). Reading the repeat as failure would zero the count on every refresh and leave the retry firing for the life of the app. The predicate accepts `.success` and `.notificationAlreadyRegistered`, so it behaves correctly whichever macOS actually returns — I checked the call graph, not the runtime value.

The rule is a pure `AutoQuitSupport.isWindowNotificationRegistered`, so it is unit-asserted rather than only pinned, and the four wiring fragments joined the existing `windowWatchRetryCode` pin. Mutation-proved by reverting the service with the tests kept: all four fragments named.

Body updated for both of the things you were right about — the four-looks cost of a re-arm, and what the count now holds.

PathGao:
Checked the `.optionAll` claim against `main` rather than taking your word for it: `hasWindowServerUserWindow` is `CGWindowListCopyWindowInfo([.optionAll, .excludeDesktopElements], kCGNullWindowID)` with an explicit off-screen branch through `SpaceWindowBridge.spaces(of:)` and `offscreenWindowKeepsAppAlive`. The hidden-Space class is reachable and does enter `windowWatchRetries`. That question is closed.

The count is now true to its name; the predicate reading it still isn't. `needsWindowWatchRetry` fires on `registeredWindows == 0`, and your own sentence in the body — "A window listed but refused registration is otherwise this bug's own shape with the retry declining on a non-zero count" — describes the case that survives it. Two windows, one registers, one refuses: `watchedWindows == 1`, the retry declines, closing the first schedules a check that finds the second, and closing the second notifies nobody. Same stuck app, one window later. It's a plausible shape for the cause you named in the root-cause section: `standardWindows` succeeds and a later call hits the 0.35s timeout on a busy app, and each `AXObserverAddNotification` is its own round trip.

The cheap version keeps the helper a pure count comparison:

```swift
static func needsWindowWatchRetry(registeredWindows: Int,
                                  listedWindows: Int,
                                  foundUserWindow: Bool) -> Bool {
    foundUserWindow && registeredWindows < max(listedWindows, 1)
}
```

Listed-nothing-but-server-shows-one is still `0 < 1`, so the reported case is unchanged. The cost is that an app with one permanently unregisterable window retries every round and re-arms on every Space change — the same profile you already accepted for the hidden-Space class.

Not asking you to hold for it. The fix closes the shape that was reported; this is the neighbour it leaves open, and the body currently reads as though it doesn't.

**Impact:** none beyond the fix's reach — an app with two or more Accessibility-listed windows where one destroy registration is refused still lands in #1008's hole, because the retry declines on the non-zero count.
**Merge:** ship it.

ref #1008

iltonandrew:
Took the fix rather than the note — in 18ee159, your shape verbatim:

```swift
foundUserWindow && registeredWindows < max(listedWindows, 1)
```

with `listedWindows: windows.count` at the call site.

Walked the reachable states rather than trusting the `max`: listed 0 / watched 0 / found true is `0 < 1` so the reported case is untouched; listed 0 / found false never reaches the comparison because `foundUserWindow` gates it; any listed count above 0 implies `foundUserWindow`, so it reduces to watched-below-listed; watched can't exceed listed because the counter increments at most once per element of the same `windows` array, and if an out-of-range value did reach the helper it would decline rather than loop.

Five unit assertions now, including the one that is the whole point — listed 2 / watched 1 / found true retries — and the pin picked up the new argument, mutation-proved by removing it with the tests kept.

I've taken the cost into the body rather than leaving it implicit: an app with a permanently unregisterable window now retries every round and re-arms on every Space change. Same profile as the hidden-Space class, as you said, and the body no longer reads as if the neighbour case doesn't exist — it names the two-window shape and what closes it.

Gate rerun unsandboxed: build zero warnings, `SELFTEST OK`, suite green apart from the known machine-dependent pool check (#1014).

Issue #1009, opened 2026-08-25
Can't select a binary (not .app file) in "Apps to leave alone"


joshrad-dev:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

For certain GUI apps, they aren't specifically packaged as a .app, but I still want to select it in the menu when choosing an app to leave alone. The reason I bring this up, is for apps like Minecraft, where you launch it from a launcher, but the actual binary itself being run is the Java app. When I go to try to select that Java binary in the menu in Vorssaint, it is grayed out and I can't select it. I'd expect for it not to show up in the list (because it's not a .app), but when I try to manually select it, I am unable to. Since I want the smooth scrolling to ignore Minecraft and it isn't technically a .app, I am unable to. 

<img width="988" height="783" alt="Image" src="https://github.com/user-attachments/assets/bc6a013e-2a19-4feb-b758-e4ff98f9ff08" />

On the other hand, there are certain features in macOS, like the accessibility settings and permissions, that have a very similar dialog. However, for the macOS settings app, it does let you pick any executable when selecting an app to add, for example the "Full Disk Access" menu. See the video below:

https://github.com/user-attachments/assets/ee4d31fd-d137-42f9-bb39-08b2bbcf79fc

### Steps to Reproduce

1. Open Vorssaint
2. Enable Smooth Scrolling feature
3. Click on "Apps to leave alone"
4. Try adding a non-app binary, and it's grayed out. 

I also wanted to mention that I tried adding Prism Launcher (the Minecraft launcher I'm using) to the ignore list and it didn't change anything.

### Feature Area

Mouse and keyboard

### Screenshots / Recordings

https://github.com/user-attachments/assets/ee4d31fd-d137-42f9-bb39-08b2bbcf79fc

<img width="988" height="783" alt="Image" src="https://github.com/user-attachments/assets/bc6a013e-2a19-4feb-b758-e4ff98f9ff08" />

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

English

### Installation Method

Homebrew

PathGao:
The picker only accepts .app bundles right now, so a bare executable like the Java program Minecraft actually runs under can't be added. Nothing on your end to fix, and no setting that unlocks it.

[...]

PR #1023, opened 2026-08-26
fix(mouse): let a program that is not an app be left alone too

iltonandrew:
Refs #1009.

## Problem

"Apps to leave alone" greys out anything that is not a .app, so a game started from a launcher — the reported case is the Java process behind Minecraft — cannot be excluded from Smooth Scrolling at all.

## Root cause

Two independent blocks, and the second is the one that matters. The picker filters to `.applicationBundle` (`AppPickerView.swift:52`), and the identity extraction drops anything without a bundle identifier (`AppBundleList.swift:127`). Underneath both, the exception is matched on `NSRunningApplication.bundleIdentifier` — and a program that is not packaged as an app has none. Measured on a bare GUI binary: `bundleIdentifier = nil`, `executableURL` set, two ordinary layer-0 windows. So relaxing the picker alone would have stored an entry that never matches anything, which is worse than the greyed-out state because it looks like it worked.

## Change

A program with no bundle identifier now answers to the path of the file being run. A path starts with `/` and a bundle identifier never does, so one list carries both kinds: nothing already saved changes meaning and there is no migration. An app that *has* an identifier still answers only to that. Only the mouse-exception lists offer the wider picker — the other lists and the uninstaller keep .app-only, so none of them can take an entry their own matcher would ignore. A path entry renders by file name and file icon in the list.

## Verification

Build clean with zero warnings, `SELFTEST OK`, suite green apart from the known pre-existing pool-starvation check (#1014 fixes it). Six unit assertions cover the identity rules; reverting the fallback fails two of them by name. End-to-end against the real support code with a bare GUI binary running: it resolves to its executable path and is excepted when that path is listed, while a normal bundled app still resolves to its bundle identifier.

## Limitations

Matching is on the executable path, so a game run by a *shared* system JVM would be excluded along with everything else run by that same binary; per-instance runtimes (what Prism Launcher usually installs) are specific enough. I could not verify the exact path Minecraft presents — no Java runtime here; asked the reporter to confirm on #1009.


PathGao:
`identities()` returns the bundle identifier *or* the executable path, never both, but the picker stores whatever URL the sheet handed back. Those are different strings for an `.app` whose Info.plist has no bundle identifier:

- picker: `Bundle(url:)?.bundleIdentifier` is nil → `onAdd("/Applications/Foo.app")`
- matcher: `app.bundleIdentifier` is nil → `identities()` returns `["/Applications/Foo.app/Contents/MacOS/Foo"]`

The entry never matches. That is the outcome the body names as worse than the greyed-out state — "stored an entry that never matches anything ... it looks like it worked" — and it is the plain path through the picker, not an exotic one. Those apps are precisely the ones the old identity extraction used to drop, so they are the new arrivals here.

One line, at the add site:

```swift
} else if acceptsExecutables {
    onAdd(Bundle(url: url)?.executableURL?.path ?? url.path)
}
```

A bare binary still works — `Bundle(url:)` is nil there and it falls through to `url.path`.

If you would rather fix it at the matcher, have `identities` return both, `[bundleID, path].compactMap { $0 }`. Then it stops mattering which of the two got stored, and the `bundleID!` goes away with it.

Separately, path matching has no normalization on either side. The stored path comes from `fileImporter`, the runtime path from `app.executableURL?.path`, and they are compared by string equality. `standardizedFileURL` is applied to the bundle-URL walk in `sourceBundleIdentifiers` but not to the identity itself. If one side resolves symlinks and the other does not (`/var` vs `/private/var`, or a per-instance runtime symlinked into place) the exception silently stops matching. `resolvingSymlinksInPath()` at both ends costs nothing. I did not check which side macOS already resolves, so this is worth one run against a symlinked runtime before calling it a non-issue.

Minor: `Self.identities(for: NSWorkspace.shared.frontmostApplication)` is uncached and runs per wheel event, and `app.executableURL?.path` is evaluated as an argument before `identities` can short-circuit on the identifier. An ordinary bundled app now pays a URL→String and an array allocation per event where it read a stored property before. Probably invisible, but it is inside the tap.

Not verified: I only see the diff. `isExcepted([String], exceptions:)` and `cacheHolds` are outside it, so I did not check the array overload's semantics or that `cachedIdentities = []` reads the same everywhere the old `nil` did. No hardware here either, so nothing about the file importer's real behaviour with `.executable` / `.unixExecutable` is checked.

The `@joshrad-dev` question at the end of the body will not reach them — they are subscribed to #1009, not to this PR. Ask it there.

**Impact:** an `.app` with no bundle identifier can be added to the mouse exception lists and silently never match. Ordinary apps and everything already saved are unaffected.
**Merge:** needs the add site to store the executable path rather than the bundle path first.

ref #1009

iltonandrew:
On it — fixing the add site so it stores what the matcher looks for, and running the symlinked-runtime check you suggested.

iltonandrew:
All of it, in a8c5e9d.

**The add site — took your (a), measured why not (b).** For the .app with no identifier, `[bundleID, path].compactMap` yields only the executable path while the picker had stored the bundle path — neither element is what was stored, so (b) leaves the reported case broken. And it heals nobody: before this branch the picker did `guard let bundleID … else { return }`, so no released build has ever written a path entry. So the add site stores what the matcher looks for, and the `bundleID!` went away differently — the identity narrowed back to `String?` (`bundleID ?? path`), with the path argument `@autoclosure`d so an ordinary bundled app never pays the URL→String or the resolve syscalls under the tap. That also answers your per-wheel-event note: same cost as before the PR for bundled apps, and the path work only ever runs for bundle-less ones, behind the existing 0.5s cache.

**One resolver, both ends, symlinks resolved.** `executablePathIdentity` is the single place a path becomes an identity; add site and matcher both route through it. Not hypothetical on this Mac even without a user symlink: the file sheet answers `/private/tmp/…` where the running program answers `/tmp/…` for the same file. Ran your suggested check against a real symlinked runtime, both directions — the one that matters is that macOS reports the *symlink* path for a process started through a link, so resolving only at store would not have been enough. Harness table (real `NSRunningApplication`, the real support file compiled in): no-ID .app, store-through-link, run-through-link all match; 86 running apps as control, 79 by identifier, 7 bundle-less by path. A unit test builds a real file + link and pins both ends.

**Found while closing it:** picking the binary *inside* a bundled app stored a path the runtime never answers to — a bundle's own executable run straight from disk is reported as its bundle identifier (measured). The picker now stores the identifier only when the picked file *is* the bundle's main executable; a runtime shipped deeper in the same bundle — the Java a launcher starts, this issue's own case — is reported with no identifier and keeps its path (also measured; walking up to any enclosing .app would have broken exactly that). Both rules are mutation-pinned. Also fixed nearby: path entries no longer pass through identifier whitespace trimming, so a legal filename with odd whitespace survives store-and-match.

**Your two not-verified items**, checked and then made moot by the narrowing: the `[String]` overload's empty-array read matched the old `nil`, and `cachedIdentities = []` behaved as `nil` everywhere (`invalidateCache` also sets `cachedAt = -1`, which `cacheHolds` can never hold). Both fields are `String?` again.

**The question for the reporter** moved to #1009, where they'll see it. Body edited: `Refs #1009` (the `Fixes` keyword was the red check's cause — closing on merge would cut the reporter out of the release-confirmation flow), root-cause line numbers refreshed.

Gate: build zero warnings, `SELFTEST OK`, suite green apart from the known machine-dependent pool-starvation check (#1014). Still not verified live: Vorssaint's own tap with a bare binary in the list — the runtime proof is a harness driving the real support code against real processes, not the shipped tap.

PathGao:
Add site stores what the matcher looks for, both ends through one resolver. That was the blocker, and it's gone.

**The leading-slash test is written three times.** `isExecutablePathIdentity` is the named one; `InstalledApps.fileURL(forIdentity:)` spells `identity.hasPrefix("/")` again and `sanitizedBundleIdentifierList` spells `raw.hasPrefix("/")` again. One rule, three sites, and only one of them is pinned by a test. Whatever changes it next — a `~` path, a file URL, a UNC-shaped thing — changes it in one place and leaves two behind. Unless Core is barred from reaching into `Services/MouseExceptions` (I can't tell that from the diff), both should call `isExecutablePathIdentity`; if it is barred, the predicate wants a home both can reach, not a third copy.

**Can the sheet actually get inside an `.app`?** The upward walk in `pickedIdentity` only runs for a file picked inside a bundle, and you call that shape this issue's own case. SwiftUI's `fileImporter` exposes no `treatsFilePackagesAsDirectories` and `NSOpenPanel` defaults it to false, so I'd expect the panel to hand back the `.app` and refuse to open it. If that holds, a runtime shipped inside a bundle is reachable only by Cmd+Shift+G or dragging the file into the panel — and unit tests calling `pickedIdentity(for:)` directly can't show it, because they never go through the sheet. I did not verify this, and it decides whether the reporter's Java is addable at all, so it's worth one pass through the real sheet before they're told to try a build.

Minor: the three source-scanning asserts read `Sources/…` relative to CWD. Two of them fail when the read comes back empty, but the `trimmingCharacters` one passes on `""`. Its siblings cover it today; it only matters if they move.

Not verified: `reload`, `isExcepted([String], exceptions:)` and the rest of `sourceBundleIdentifiers` are outside the diff, so I did not check that `add`'s sanitizing matches what `reload` reads, nor that the ancestor walk — still `standardizedFileURL`, not `resolvingSymlinksInPath` — never contributes an unresolved path beside the resolved identity you now put first. No hardware here either.

**Impact:** possibly the in-bundle runtime case, if the file sheet won't descend into a package — unverified, and in the shape of a gap this PR doesn't widen: today nothing outside an `.app` can be added at all. Nothing else.
**Merge:** ship it.

ref #1009

iltonandrew:
Both cleanups in 92126b8, and an answer on the sheet.

**One spelling now.** Core isn't barred from Services here — `Core/Defaults.swift` already calls `MixerRoutingSupport.sanitizedDeviceUID` and reaches `DockPreviewSupport`, `RecorderSupport`, `ScreenshotSupport` — so the predicate stayed where it was named and both other sites call it: `sanitizedBundleIdentifierList` and `InstalledApps.fileURL(forIdentity:)`. `isExecutablePathIdentity` is exactly `hasPrefix("/")`, so the substitution is behaviour-preserving rather than a quiet tightening — I checked that before swapping it in. The doc line on `fileURL` restating the never-collides rule went with it, since that reasoning now lives once, on the predicate. A new check asserts `hasPrefix("/")` appears at exactly one non-comment site across the three identity-handling files, so a fourth copy fails; mutation-proved by reintroducing the `Defaults` copy. The other three hits in the tree (Homebrew, DiskImageInstaller, Cleaner) are unrelated absolute-path tests, left alone.

**The fail-open assert fails closed now** — it builds the matcher body from the filtered lines and requires it non-empty before asserting absence. Proved by pointing the read at a nonexistent file: it fails where it used to pass.

**On the sheet — your expectation holds, and it matters less than it looks.** `AppPickerView.swift:59-63` is SwiftUI `fileImporter` with `[.applicationBundle, .executable, .unixExecutable]`, and `treatsFilePackagesAsDirectories` appears nowhere in Sources or Tests — `fileImporter` exposes no such knob and `NSOpenPanel` defaults it false, presenting packages as opaque files. So browsing hands back `Foo.app` and `pickedIdentity` returns its identifier; you can't walk into a bundle by navigating. Cmd+Shift+G and dragging a file onto the panel both reach inside, and a Mach-O conforms to `public.unix-executable` so it stays selectable — standard panel behaviour, high confidence, but not exercised here.

The part that decides the reporter: Prism's managed runtimes live under `~/Library/Application Support/PrismLauncher/java/…`, not inside any `.app`, so the plain sheet reaches them and the in-bundle walk is the narrower extra. One wrinkle worth knowing before anyone tells them it just works — some runtime flavors are package-shaped in themselves (Mojang ships a `jre.bundle`; `.bundle` and `.jdk` are package extensions), and those would be opaque to browsing too, needing the same Cmd+Shift+G. Adoptium/Azul-style layouts browse normally. Nothing Java is installed on this machine, so that path shape is from Prism's documented data-dir layout, not observed — which is why I asked them on #1009 to paste the path they pick rather than guessing it for them.

No code for that one: `fileImporter` has no packages knob short of hand-rolling an `NSOpenPanel`, which is restructuring, not a fix.

PathGao:
Re-read against main for the new commit. Both things I left unverified last time check out:

`reload` reads through `Defaults.sanitizedBundleIdentifierList` (`MouseAppExceptions.swift:60`) and `add` now calls that same one, so store and read can't drift apart. And the ancestor walk in `sourceBundleIdentifiers` only ever appends `Bundle(url:)?.bundleIdentifier` — never a path — so `standardizedFileURL` there can't put an unresolved path beside the resolved identity you now put first. Nothing to fix in either. The three other `hasPrefix("/")` hits are what you said they are.

**The "never walk up" rule is written as global and isn't.** `pickedIdentity`'s doc says walking up to an enclosing `.app` "would file that Java under the launcher and break the case this all exists for" — and `sourceBundleIdentifiers`, one type over, does exactly that walk and appends the enclosing app's identifier. That's correct for what it does (source scope, helpers inherit the app's exception) and it predates this PR, but the two opposite rules now sit a file apart with one of them phrased as the rule. Scope that sentence to the picker, or the next reader changes one to match the other. Not a code change.

Minor, pre-existing, not widened here: that walk starts from `standardizedFileURL` rather than the resolver, so a URL with a link on the way — the same condition you measured for executable paths — can't find the `.app` above it and the inherit silently misses.

Not verified: no hardware here, so the real sheet, `.executable` / `.unixExecutable` selectability, and the package-shaped runtime flavors all still rest on your measurements and on standard panel behaviour, not on anything I checked.

**Impact:** none.
**Merge:** ship it.

ref #1009

iltonandrew:
Both in 3e52bbc.

**The doc sentence is scoped.** It now says walking up when *storing* would file that Java under the launcher, then names `sourceBundleIdentifiers` as the deliberate opposite in one clause — so the two rules read as two rules rather than one rule and one contradiction. `sourceBundleIdentifiers` untouched.

**The rows.** Took it here rather than after, since @joshrad-dev's own follow-up makes it concrete: `jre-legacy` now, `java-runtime-gamma` when he plays 1.17+, and `FileManager.displayName` says `java` for both while the sort keys on that same name — identical rows in arbitrary order.

A path row is now name over directory, spelled from home, `.font(.caption)`. On the truncation, I went with `.head` rather than the tooltip alone and the reason is in a constraint comment: sibling runtimes share a long *prefix* (`~/Library/Application Support/PrismLauncher/java/`) and differ after it, so dropping the head keeps the component that differs and middle-truncation would cut exactly the part worth showing. The full unabbreviated path is on hover as well.

Bundle-id rows can't shift: they take an `else` branch that is byte-for-byte today's single `Text` — no stack, no caption, no tooltip. And equal display names now tie-break on the identity, so three rows named `java` hold one order between renders instead of `sorted`'s arbitrary one.

`location(for:)` routes through `isExecutablePathIdentity` rather than adding a fourth spelling — the one-site pin from last round still passes, which is the check working. `InstalledApps` is in the test binary so it got real assertions (bundle id → nil, a home-relative path → `~`-spelled directory, `/opt/game/bin/java` → `/opt/game/bin`); the row shape is pinned by source scan and mutation-proved by swapping to `.middle` and renaming the call.

Everything visual is unverified — no GUI session here, so the two-line row's spacing, the head-truncated text against real sibling paths and the tooltip are reasoned from the SwiftUI only. The bundle-row no-change claim is structural rather than a screenshot.

PathGao:
Merged, so these are follow-ups. Two things in 3e52bbc.

**Head truncation doesn't keep the component that differs — not for the layout you named.** The two identities are

```
~/Library/Application Support/PrismLauncher/java/jre-legacy/jre.bundle/Contents/Home/bin/java
~/Library/Application Support/PrismLauncher/java/java-runtime-gamma/jre.bundle/Contents/Home/bin/java
```

`location` drops the last component, so the two captions differ at `jre-legacy` / `java-runtime-gamma` and then share `/jre.bundle/Contents/Home/bin` to the end. `.head` keeps the tail, and the tail is the shared part. Two rows both named `java`, both reading `…/jre.bundle/Contents/Home/bin`, with a minus button next to each.

The constraint comment says siblings "differ in the middle or tail" and picks `.head` on that basis. Mojang-shaped runtimes differ in the middle, and `.head` and `.middle` both cut the middle. `.tail` is what keeps it here and is worse whenever the difference sits near the end instead — no mode wins both. `.lineLimit(2)` and let it wrap doesn't have to pick. If you'd rather keep one line and lean on the hover, then what needs changing is the comment and the assert: `.truncationMode(.head)` is pinned by a source scan right now, so the premise is enforced.

No GUI here, so which components survive is arithmetic on string lengths, not something I saw, and the Prism layout is your description plus Mojang's `jre.bundle`, not observed.

**`acceptsExecutables` reaches the file importer but not the list under it.** `InstalledApps.swift:194` still guards on `runningApp.bundleIdentifier` and the filter at `:211` drops anything with a nil one, while `installedApplications` only walks `.app`. So the running Java the reporter is looking at never appears in the picker's list — Browse plus knowing the path is the only route, which is why you had to ask him to paste it. `AppBundleList` passes `includeRunningApplications: reachesEveryApp` to `loadApps` and nothing else; the flag stopped one call site short of the one that decides what's offered. Threading it means `bundleID ?? path identity` for a bundle-less regular app, gated so the uninstaller's picker doesn't start offering binaries.

Not a regression — that guard predates the PR — and neither of these is a reason to revert.

**Impact:** two runtimes under one launcher render as identical rows with a remove button on each; and the kind of program this PR exists for can't be picked from the list of running apps, only browsed to by path.
**Merge:** ship it — already merged, both are follow-up.

ref #1009

Issue #1012, opened 2026-08-25
Click-and-drag gesture to switch Spaces / open Mission Control (like Mac Mouse Fix)


jaybcd:
### Problem or Motivation

I use a regular mouse, not a trackpad, so I rely on Mac Mouse Fix's Spaces and Mission Control click-and-drag action to hold a mouse button and drag to switch between Spaces or trigger Mission Control, the same way you would with a trackpad's four-finger swipe.

On macOS 27 beta this broke, because WindowServer now rejects synthetic, software-simulated gesture events, which is how Mac Mouse Fix and similar tools used to implement it. There is no built-in macOS way to do this with a plain mouse anymore, and I do not want to be stuck without this if Mac Mouse Fix stays broken on newer macOS versions.

### Proposed Solution

Add a "Spaces & Mission Control" click-and-drag action to the mouse configuration, similar to the existing window drag/resize actions: assign it to a mouse button (or side button), hold it down, then drag left/right to switch between Spaces and drag up to open Mission Control or App Exposé.

Since WindowServer now blocks synthetic gesture/touch events on macOS 27, this should not be built by simulating a touch gesture. It should trigger Spaces/Mission Control through the system's symbolic hotkeys (CGSSymbolicHotKey / the same private APIs behind the Mission Control keyboard shortcuts), mapping drag distance and speed to a Space change or to opening Mission Control. There is an open, working reference implementation for exactly this approach in Mac Mouse Fix here: https://github.com/noah-nuebling/mac-mouse-fix/pull/1875 (not merged there yet, but confirmed working on macOS 27 in that thread).

### Feature Area

Mouse and keyboard

### Does another app already do this?

Mac Mouse Fix - had this exact feature but it broke on macOS 26/27 for the reason above and is not fixed in any release yet.

iltonandrew:
Taking a look at this one.

jaybcd:
Updated to 3.3.5 today. The Spaces/Mission Control drag toggle from #1024 isn't showing up for me.

In Mouse and keyboard settings, right after "Mouse button shortcuts", it goes straight to "Middle click", nothing about Spaces or Mission Control in between. Checked About and it does say 3.3.5.

Not fixed on my end, could someone take another look?

jaybcd:
Still not there. Tried 3.3.3 Beta 4, and now on 3.4 Beta 1, no sign of the Spaces/Mission Control drag option in either.

[...]

Issue #358, opened 2026-07-24
[Feature]: Add app exceptions to mouse control features


kyteidev:
### Problem or Motivation

Smooth scrolling can break some apps where mouse scroll is used as input, like Blender. When smooth scrolling is turned on, the camera zooms way more than intended.

### Proposed Solution

A setting to exempt some apps from mouse features like smooth scrolling, side buttons, and other features could work.

### Feature Area

Other

### Mockups / References

[Mos](https://github.com/Caldis/Mos) has this feature.

vorssaint:
3.2.0 lets each mouse feature keep its own list of apps to leave alone, under Mouse.

PR #1024, opened 2026-08-26
feat(mouse-buttons): switch Spaces and open Mission Control by dragging a button

iltonandrew:
Refs #1012.

People on a plain mouse had no way to reach Spaces or Mission Control the way a trackpad swipe does, and the tools that filled that gap stopped working when the window server began refusing software-simulated touch gestures.

This adds a hold-and-drag action on an extra mouse button: drag left or right to move one Space over, up for Mission Control, down for App Exposé. Nothing is simulated. Each command is asked for through the combination the system itself has registered for it, read from the window server's symbolic hotkey table — the same table this app already reads to travel to a window on a hidden Space, now extended with the two overviews (ids 32 and 33 alongside the existing 79 and 81). A command whose shortcut the user switched off in System Settings reads as nothing, and the gesture stays silent rather than pressing something else; Settings says so when none of the four is left.

The press is held back, not taken. Until the pointer has gone far enough to mean something, the button is still the app's: a tap, or a drag that stays under the thresholds, gives the press back and the app sees one ordinary click — the same custody rule the window move-and-resize gesture already keeps, using its replay idiom. A press that did fire keeps its own release, so no app ever receives half a click.

It rides the tap the mouse button shortcuts already run, because that tap already receives the drags this measures — gated under that same feature, so no new hub tile, no new permission surface. A button carrying a shortcut keeps it (the binding capture refuses in both directions), the radial menu keeps its summoner, and mouse navigation yields through the claim check all three already share. Once a drag has started switching Spaces it stays on that axis, so a diagonal hand cannot open Mission Control halfway through; horizontal repeats with a cooldown matched to the slide animation, vertical fires once per press because an overview is a toggle.

The four numbers that decide the feel — horizontal step, vertical step, repeat cooldown, flick banking — live named and documented in `MouseSpacesGestureSupport`, where they want tuning against real hardware. Every decision they drive is a pure function with unit tests: direction, thresholds ("not a pixel sooner" both axes), repeat and cooldown, axis lock, below-threshold no-op, and the binding arbitration. Three mutation rounds fail 1–7 tests by name. Two source-shape assertions pin the service wiring and one pins the hotkey ids.

New user-facing strings: four, added to all 13 languages in `MouseButtonStrings.swift` (the memberwise initializer is the completeness check).

Measured on macOS 26.6.2 on a live dev build: a bound drag moves between Spaces in both directions with **zero** button events reaching the app underneath, an up-drag brings up Mission Control and a down-drag App Exposé (verified via the Dock's overview scene windows plus screenshots), and a tap on the same button still arrives as one whole click carrying the replay marker. Gate: build zero warnings, TESTS OK apart from the known pre-existing pool-starvation check (#1014), SELFTEST OK. Not tested on macOS 27 — the claim that symbolic hotkeys survive there rests on the Mac Mouse Fix thread the issue cites, not on my own evidence.


PathGao:
Three findings, all in the same corner, plus the red check.

**1. The exception list vanishes but keeps deciding.** `armSpacesGesture` sits behind the `excludesActionTarget(.buttonShortcuts, …)` guard, so the mouse-exception list governs the drag too. `MouseExceptionsList(scope: .buttonShortcuts)` renders inside the block gated on the shortcut switch (the `if` line is above the diff window, so check me there), and the drag does not need that switch: `spacesGestureButton()` reads only `availabilityKey` and `mouseSpacesGestureEnabled`. Shortcut switch off, drag on, one excluded app, and the gesture is dead in that app with the list that killed it off screen. Either move the list out of that block or stop applying `.buttonShortcuts` exceptions to the drag.

**2. `spacesEnabled &&` turns the two-way refusal into one and a half.**

```swift
} else if mappings[seen] != nil || pendingButton == seen
            || (spacesEnabled && Int64(spacesButton) == seen) {
```

`spacesButton` outlives the toggle, and `spacesRow` only renders while `spacesEnabled`. Turn the drag off, record a shortcut on that same button, turn the drag back on: `boundButton` yields to `hasShortcut` and returns nil, so the gesture does nothing while the row still names the button. `spacesShortcutsOffNote` does not cover this, it only inspects the four system shortcuts. Body says the capture "refuses in both directions"; with the toggle off it does not. Cheapest close is `spacesButton = 0` in the toggle's off branch next to `stopSpacesCapture()`, since the row is hidden then anyway. Dropping `spacesEnabled &&` instead makes an invisible binding refuse a shortcut with no visible reason, which is worse.

**3. The tap now runs while the shortcut switch is off.**

```swift
let wanted = (enabled && (!mappings.isEmpty || isCapturing)) || spacesButton != nil
```

`hasActiveShortcut` states the rule: a mapping left behind while the switch is off "is inert, so it holds no claim on a button." That rule now has to hold on the down path as well, and the down path hands every mapped button to `firesShortcut`. Its signature is not in the diff, so I cannot tell whether it re-checks `mouseButtonShortcutsEnabled` or whether the old "the tap only exists while the switch is on" invariant was carrying that for free. If it is the latter, switch off plus drag on resurrects every old mapping and swallows the click. Confirm which.

Not verified: the tap's `eventsOfInterest` mask is not in the diff, so I took the body's word that `.otherMouseDragged` already arrives at this tap. The four calibration constants are a hardware-feel question I have no opinion on. The pure half reads correct against its tests, axis lock and banked flick included, and the replay custody rule looks sound: `flushSpacesPress` checking `pressedMouseButtons` before replaying is the right guard.

`Point out a reference that will not reach the reporter` is red.

**Impact:** anyone running the drag without the shortcut switch also on. Invisible exceptions, a binding that can silently stop working while Settings says it is bound, and possibly old shortcut mappings firing again. With both switches on, none of this bites.
**Merge:** needs the shortcut-switch-off combination settled first, which is where all three findings live.

ref #1012

iltonandrew:
On it — settling the shortcut-switch-off combination (all three findings) and the red check. Will confirm what firesShortcut re-checks.

iltonandrew:
All three, plus the check, in ec2b565.

**1 — the list.** Confirmed: it rendered inside the `if enabled` block, exactly where you pointed. Kept the exception applying to the drag — the excluded-app semantics from #358 are "this app keeps its buttons," and the apps people exclude (3D/design tools) are the ones that drag with side buttons natively, which is precisely what the gesture would steal. So the list moved instead: bottom of the section, shown while either switch is on, pinned by a whole-line shape assertion (indentation included, so re-wrapping it in `if enabled` fails the pin).

**2 — the binding.** Your shape: the toggle's OFF branch clears `spacesButton` beside `stopSpacesCapture()`. `spacesEnabled &&` stays in the capture guard, per your reasoning — it also still covers defaults persisted by builds before this fix. Pinned with an adjacency window anchored on the toggle's `onChange`, so the pair can't drift somewhere that doesn't run on toggle-off. Cost accepted: off-and-on again forgets the button; the row invites a re-pick, which is honest about what OFF now means.

**3 — the mappings.** Confirmed safe, and now pinned: `firesShortcut` guards on `isEnabled` and the down path passes the live `mouseButtonShortcutsEnabled` default, returning the event whole when it declines — so switch-off + drag-on swallows nothing. That invariant used to be carried for free by the tap not existing; the shape assertion pins both the switch-read at the call site and the `return Unmanaged.passUnretained(event)` pass-through, and a `return nil` mutation fails it by name. Side-wheel can't leak either — the scroll mask and `hasActiveSideWheelInterest` both require the switch. Your mask note: `.otherMouseDragged` is in the base mask unconditionally at `MouseButtonShortcutService.swift:168-170`.

**Found while fixing 2:** clearing the binding exposed that Spaces *capture* was still gated on the shortcut switch in the tap's want-rule — shortcuts off, no bound button, "Choose a button" armed a capture against a tap that never started (and toggling shortcuts off mid-capture tore the tap down under it). The want-rule now lets an active capture hold the tap on its own; each capture row lives behind its own switch, so an armed capture always has a live owner. Pinned.

**The check** — same cause as on #1022: the body opened with `Fixes #1012`, and closing the issue on merge would cut the reporter out of the release-confirmation flow. Now `Refs #1012`. (The job's red exit was its advisory comment being denied to the fork token; with no closing keyword it exits clean before commenting.)

**One string question:** with only the drag on, the list's caption still reads "Your button shortcuts stay quiet in these apps…" — right effect, wrong noun. Proposed EN: "Your button shortcuts and the Spaces drag stay quiet in these apps, and the press reaches them instead." Want the 13-language update in this PR or as a follow-up? Strings left untouched pending your call.

Gate: build zero warnings, `SELFTEST OK`, suite green apart from the known machine-dependent pool-starvation check (#1014). Not GUI-verified: the moved list and the capture flow with shortcuts off are logic-and-pin only.

PathGao:
Three of three land as described, and the diff matches. One thing the want-rule change opened, then your string question.

**The capture now outlives its own switch.** `let wanted = (enabled && !mappings.isEmpty) || isCapturing || spacesButton != nil` — `isCapturing` holds the tap on its own, and the shortcut `captureRow` sits inside `if enabled` (the brace you moved closes right after it). Shortcuts on, arm a capture, switch shortcuts off: the row disappears, `capturing` and the service's `isCapturing` don't, and the tap stays up to swallow the next extra-button press with nothing on screen to say why. Before this PR the same sequence made `wanted` false and tore the tap down, so it was inert. Your rule — "each capture row lives behind its own switch, so an armed capture always has a live owner" — holds for the drag because you put `stopSpacesCapture()` in the toggle's OFF branch. The shortcut toggle's `onChange` is above the diff window, so I can't tell whether it does the same. If it doesn't, it's finding 2's shape again and takes the same one-liner.

**The string: here, and swap the noun rather than add the clause.** It only becomes wrong because this PR makes the list reachable with the drag alone, so it ships wrong the moment this merges, and a 13-language follow-up for one caption is the PR that never gets written. But instead of naming the drag: "Your extra mouse buttons stay quiet in these apps, and the press reaches them instead." One noun per language instead of a new clause, correct in all three switch states, and it doesn't tell a shortcuts-only user about a gesture they never turned on. "Extra buttons" is already the panel caption's word.

**One inversion in your note on 2.** `spacesEnabled &&` is precisely what fails to cover a binding persisted by an earlier build: `spacesEnabled` false with `spacesButton` still 5, the capture guard doesn't refuse button 5, a shortcut gets recorded on it, and turning the drag on then draws a bound row driving nothing — finding 2 verbatim. Keep the clause anyway, for the reason you kept it. That stale state can only exist in dev builds of this branch, since the key is new here.

**One pin is weaker than it reads.** `downPathReadsShortcutSwitch = …` assigns inside the `for`, so with two `firesShortcut(` call sites the last one decides and an earlier one could lose its switch-read without failing anything. One call site today, so it holds now; `spacesSwitchOffDropsBinding` accumulates and doesn't have this.

Not verified: the `isEnabled:` argument at the down-path call site is outside the diff window, so the green pin is my evidence there, not my own reading. Same for the caption's current wording — I only have your quote. Whether `handleSideWheel` consults this same exception list I also can't see; if it does, "buttons" is no worse than what's there now.

**Impact:** switching shortcuts off with a capture armed eats one extra-button press with no UI, and drag-only users get a caption naming a feature they didn't turn on. Nothing else moved.
**Merge:** needs the shortcut toggle's OFF branch confirmed — `stopCapture()` there, or a word that it's already there. The caption can ride along.

ref #1012

iltonandrew:
All in 2c38ce4.

**The shortcut toggle already does it** — `MouseButtonSettings.swift:40-46`:

```swift
.onChange(of: enabled) { _, on in
    if !on { stopCapture() }
    MouseButtonShortcutService.shared.syncWithPreferences()
```

and `stopCapture()` at `:265-270` clears the local flag *and* `setCapturing(false)` on the service, so the sequence you describe ends with `isCapturing` false and `wanted` false — the tap comes down and nothing is left to swallow a press. It has been there since the switch was added, which is why the drag's OFF branch was the only one missing the equivalent. The two are now symmetric: each capture's own switch stops it, and neither `stopCapture()` nor `stopSpacesCapture()` touches the other's flag, so turning one switch off during the other's capture leaves that capture running with its row still on screen.

**Caption: your wording, and you're right that it ships wrong otherwise.** All 13 languages swapped to the noun rather than the clause — EN is now "Your extra mouse buttons stay quiet in these apps, and the press reaches them instead." Each language uses the word its own panel label already uses for the extra buttons (`Zusatztasten`, `boutons supplémentaires`, `дополнительные кнопки`, `拡張ボタン`, `額外按鍵`, and so on), so the caption and the switch above it read as the same feature.

**Your inversion is correct and mine was backwards.** I had written that `spacesEnabled &&` covers a binding persisted by an earlier build; it is exactly the clause that *fails* to cover it — drag off with `spacesButton` still 5, the guard doesn't refuse button 5, a shortcut records onto it, and turning the drag on then draws a bound row driving nothing. That is finding 2 verbatim, reachable only in dev builds of this branch since the key is new here. Clause kept for the reason you gave: an invisible binding refusing a shortcut with no visible reason is the worse failure.

**Pin fixed.** It assigned inside the loop, so the last call site decided and an earlier one could lose its switch-read silently. It now counts call sites and accumulates the ones missing either half, failing on an empty search too:

```swift
expect(shortcutCallSites > 0 && callSitesMissingShortcutSwitch.isEmpty, …)
```

Mutation-proved with the case the old shape could not see — an *earlier* unguarded `firesShortcut(` call site added above the real one, with the real one left intact. Old shape: green. New shape: fails naming the planted site.

On your unverified list: the down-path `isEnabled:` argument is `MouseButtonShortcutService.swift:301`, reading `UserDefaults.standard.bool(forKey: DefaultsKey.mouseButtonShortcutsEnabled)` at the call, with `else { return Unmanaged.passUnretained(event) }` on the next line.

And `handleSideWheel` **does** consult this same list — `MouseButtonShortcutService.swift:460`, `excludesActionTarget(.buttonShortcuts, …)` — so your caveat applies: the caption covers a side-wheel tilt too, which "buttons" names no worse than "button shortcuts" did. Worth knowing rather than assuming, so I checked rather than answering from the shape of the rest.

Gate: build zero warnings, `SELFTEST OK`, suite green apart from the known machine-dependent pool-starvation check (#1014).

PathGao:
The OFF branch is there: `MouseButtonSettings.swift:34-40` on main calls `stopCapture()`, and `stopCapture()` at `:162-167` guards on `capturing` and calls `setCapturing(false)`. Your symmetry claim holds both ways for the reason you gave plus one you didn't: each stop guards on its own flag, so turning shortcuts off during a spaces capture returns at that guard instead of tearing the other capture down. The pin, the caption and the inversion all read as described, and `handleSideWheel` does consult the same list.

One thing the drag inherited along with the capture flow.

**Three borrowed strings name things the drag doesn't have.** `spacesRow` and `handleSpacesCapture` reuse the shortcut capture's text, and three of them were written for a feature that takes side-wheel ticks and shows a list:

- `captureWaiting`, the prompt the user is reading while the drag capture is armed: "Now press an extra button **or move the side wheel**." `MouseSpacesGestureSupport.canBind` refuses the side wheel by design, and your own test says so by name.
- `captureUnsupported`, which is what a side-wheel tilt then gets: "That input cannot take a shortcut. Use an extra button **or a side-wheel direction**." The refusal recommends what just failed. Reachable: an armed spaces capture sets `isCapturing`, so with shortcuts on `wantsSideWheelEvents` is true and `handleSideWheel` reports the tilt into `lastInputSeen`.
- `captureExists`: "That button or direction is already **on the list below**." The mapping list renders above the spaces toggle, and with the shortcut switch off it isn't on screen at all. Shortcuts off, a leftover mapping on button 4, pick button 4 for the drag: refused, pointing at nothing.

Keep the `mappings[seen] != nil` check itself. It is the other half of the two-way refusal, and loosening it to match `hasActiveShortcut` would let the shortcut switch coming back on take the button — finding 2 again. Only the wording is wrong. "That button already has a shortcut" is true in both switch states and doesn't point anywhere.

Three more strings in a file this PR already has open in 13 languages, and the same argument you accepted for the caption: they ship wrong the moment this merges, and a string-only follow-up is the PR that never gets written. The memberwise initializer can't see these — it counts new fields, and these are reused ones.

Not verified: I read the English strings and spot-checked `captureExists` in pt-BR and tr (both say "below" too); I'm assuming the other ten follow. Line numbers above are main's, so they sit lower on your branch. No GUI verification from me on any of this.

**Impact:** text only, no behavior change. Anyone binding the drag is invited to move a side wheel the drag refuses, and one refusal names a list that may not be on screen.
**Merge:** needs the three borrowed capture strings replaced first.

ref #1012

iltonandrew:
Replaced in 931c343 — new fields, not reworded shared ones, for the reason you gave: the memberwise initializer can't police a reused field but it does police a new one.

I sequenced it to make that the proof rather than a claim. Added the three fields with English filled in and built: **12 compile errors**, one per language instance that hadn't been filled (`missing arguments for parameters 'spacesCaptureWaiting', 'spacesCaptureUnsupported', 'spacesCaptureExists'`, first at ptBR). Filled all 12, green. The pre-existing per-language completeness suite is the second net — it counts the struct's fields through `Mirror`, so it went 27 → 30 and re-checks every language for non-empty and no em-dash.

The three:

- `spacesCaptureWaiting` — "Now press an extra button." The wheel clause is simply gone rather than negated.
- `spacesCaptureUnsupported` — "That input cannot be held for a drag. Use an extra button." This one gains a reason it didn't have: a wheel tick isn't refused because it "cannot take a shortcut" (it can, in the other flow) but because it cannot be *held*, which is what the gesture needs.
- `spacesCaptureExists` — "That button already has a shortcut. Pick another one." Your wording, and true in both switch states since an inert mapping is still a shortcut on that button; it points nowhere.

`captureBlind`, `captureCancel` and `captureWheel` stay shared — correct in both flows. `mappings[seen] != nil` untouched, per your reasoning about the two-way refusal.

Pinned so the drag flow can't drift back onto the shortcut strings: anchored scans over `spacesRow` and `handleSpacesCapture` require the new fields and forbid `text.captureWaiting`/`captureUnsupported`/`captureExists` in those windows. Mutation-proved by reverting both call sites — both assertions fail by name.

Two caveats. The translations are mine, matching each file's existing register and the nouns from the caption round; a native pass on tr and ko wouldn't hurt if you want one. And the branch is `MERGEABLE / BEHIND` — no conflicts, so I've left it un-rebased rather than force-pushing under your review; say the word and I'll rebase.

PathGao:
The three strings read right and the pin story holds where I can see it. One hole left in the same corner, then your two questions.

**The drag capture still takes a button that's mid-way through becoming a shortcut.** `handleCapture` refuses three things — `mappings[seen] != nil`, `pendingButton == seen`, and the spaces button. `handleSpacesCapture` refuses two: no `pendingButton` check. `startSpacesCapture()` calls `stopCapture()`, and `stopCapture()` (main, `MouseButtonSettings.swift:162-167`) guards on `capturing` and clears `capturing` and `captureFeedback` only — `pendingButton` survives it, and its row keeps rendering.

Shortcuts on, Add, press button 4: `pendingButton` 4, row on screen, nothing persisted. Choose a button, press 4: accepted, `spacesButton` 4. Now record a combination on the row still sitting there — `save` writes `mappings[4]`, `persist()` syncs, `hasActiveShortcut(4)` goes true, `boundButton` returns nil, and the spaces row still names button 4 while driving nothing. Finding 2's ending, through the one door the two-way refusal doesn't cover. Same one-clause close as its sibling:

```swift
} else if mappings[seen] != nil || pendingButton == seen {
```

`spacesCaptureExists` says "already has a shortcut" and a pending row doesn't have one yet, but `captureExists` is equally loose about its own pending case on main. I'd leave the wording.

**Native pass: not from me, and I don't think it blocks.** I can't judge register in tr or ko. What is checkable is noun consistency, and it holds — each new string uses the word its own language already uses in the capture flow. tr's `ek düğme` beside the panel's `Ekstra düğmeler` looks like drift but isn't yours: main's `captureWaiting` and `captureUnsupported` already say `ek bir düğme` (`MouseButtonStrings.swift:111,114`). Yours match what they replaced.

**Rebase: don't.** `MERGEABLE` is `MERGEABLE`, behind costs nothing at merge time, and a force-push mid-review costs a re-read. If the clause above goes on top, push it as an ordinary commit.

Not verified: my copy of the diff is truncated partway through `MetricsTests.swift`, so the two anchored scans over `spacesRow` and `handleSpacesCapture` are your word rather than my reading — everything before the cut matches what you described. No GUI verification from me on any of it.

**Impact:** with both switches on, a shortcut capture left half-finished on a button can be handed to the drag as well, and finishing that shortcut kills the drag binding while Settings still names the button. Needs the user to touch the same button in both flows, so narrow.
**Merge:** needs the `pendingButton` clause in `handleSpacesCapture` first.

ref #1012

iltonandrew:
Clause is in, and I had to overrule your rebase advice — reasons for both below.

**The `pendingButton` hole.** Confirmed your trace before touching it: `stopCapture()` (`MouseButtonSettings.swift:265-270`) guards on `capturing` and clears `capturing` and `captureFeedback` only, `startSpacesCapture()` calls it at `:289`, and the pending row keeps rendering from `:59-60`. Took your clause verbatim, including leaving `spacesCaptureExists` as the wording:

```swift
} else if mappings[seen] != nil || pendingButton == seen {
```

Pinned with an anchored scan over `handleSpacesCapture` in the same shape as its two siblings, and mutation-proved rather than assumed: with the clause stripped the new check goes red on its own; with it in place the suite is otherwise unchanged. It fails closed too — `spacesCaptureRefusesPending` starts false, so a renamed or moved anchor collapses the window to empty and the check reports missing instead of passing.

**Rebase — I did it, against your advice, because the premise expired.** `MERGEABLE` was `MERGEABLE` when you wrote that. #1023 merged at 23:52Z and this branch went `CONFLICTING/DIRTY`; the two touch `Defaults.swift` and the mouse exception surface. So it was not a cosmetic force-push, it was the only way to a mergeable branch. One real conflict, in `Defaults.swift`: `main` had reworded the `superKeyEnabled` comment to "chosen key holds the configured modifiers" and added `superKeySource`, while this branch adds the two `mouseSpacesGesture` keys just above. Kept both sides. The other four commits replayed clean. Now ahead 5, behind 0.

**Native pass.** Agreed and nothing to do — your point that each new string uses the word its own language already uses in the capture flow is the checkable half, and `ek düğme` matching main's existing `ek bir düğme` at `MouseButtonStrings.swift:111,114` settles the tr case as pre-existing rather than mine.

**Gate.** 0 warnings, `SELFTEST OK`, `--test` 8824 checks with one failure: the `dispatch pool starvation` precondition #1014 exists to skip, red on this 15-core M5 Pro and green on CI. Same failure appeared on two unrelated branches gated alongside this one, so it is environmental. Saying so rather than writing `TESTS OK`.

Still no GUI verification of the pending-then-drag sequence from me — the clause is reasoned from the two handlers and pinned, not driven through Settings by hand.

ref #1012

PathGao:
The `pendingButton` clause is in and reads as described, and the `Defaults.swift` conflict kept both sides — `superKeySource` and the reworded `superKeyEnabled` comment sit next to the two new keys. Rebase call was yours to make and you were right; my premise had expired.

Two things the second `enabledKeys` entry reached that the diff didn't.

**1. The Command Bar's mouse-button-shortcuts row disappears.** `CommandBarCatalog.swift:244-247`:

```swift
// Two keys means two different switches; which one a single row
// would flip is a guess, and a guess here changes the person's Mac.
guard feature.enabledKeys.count == 1,
      let key = feature.enabledKeys.first else { return [] }
```

This PR takes that count from 1 to 2, so `toggleEntries` returns nothing for the feature and "Turn on Mouse button shortcuts" stops existing in the bar. Nothing fails: the file isn't in the diff and it isn't past my cut either — it sorts before `Services/MouseButtons/`, so it's untouched rather than unseen — and no assertion in the suite names `toggle.mouseButtonShortcuts`. The body's "no new hub tile, no new permission surface" doesn't cover losing a row that exists today.

The precedent is directly above the guard: `.scrollInverter` is in exactly this position and takes two rows.

```swift
if feature == .mouseButtonShortcuts {
    let buttons = FeatureStrings.mouseButtons(language)
    return [
        entry(for: feature, key: DefaultsKey.mouseButtonShortcutsEnabled,
              name: feature.hubTitle(s, hub: hub), id: "toggle.mouseButtonShortcuts"),
        entry(for: feature, key: DefaultsKey.mouseSpacesGestureEnabled,
              name: buttons.spacesEnableLabel, id: "toggle.mouseButtonShortcuts.spacesGesture"),
    ]
}
```

Keep the first id unchanged so anything that remembered it still matches. No new strings — `spacesEnableLabel` is already in 13 languages from this PR — and `run`'s `FeatureRuntime.shared.sync([feature])` lands on `MouseButtonShortcutService.syncWithPreferences()` (`FeatureRuntime.swift:173`), which this PR already taught to read the drag key.

**2. Two more copies of "is this feature engaged" kept the old answer.** `enabledKeys` is one of three places that fact is written, and it's the one that moved:

- `SettingsView.swift:836-844`, `accessibilityNoteVisible`: `mouseButtonShortcutsEnabled && AppFeature.mouseButtonShortcuts.isAvailable`. Shortcuts off, drag on, Accessibility not granted, and the Mouse page hides its whole Permission required section.
- `MenuPanelView.swift:1620`, `needsAttention: mouseButtonShortcutsEnabled && !permissions.accessibility` — same condition, same silence, on the panel row.

Both take `|| spacesEnabled`. Your toggle's `onChange` asks for Accessibility when it goes on, so this only bites the person who declines it or revokes it later — which is the person those two surfaces exist for. `isEnabled(_:)` at `MenuPanelView.swift:1235` is the same shape at lower stakes: the panel reads the row as off while the drag runs.

Not verified: my copy of the diff is cut partway through `MetricsTests.swift`, so `spacesCaptureRefusesPending` and anything after it is your word rather than my reading. The three files above are main's and this branch doesn't touch them. No GUI verification from me on any of it.

**Impact:** everyone loses a Command Bar toggle they have today, whether or not they turn the drag on. And with the drag on and shortcuts off, a missing Accessibility grant is invisible on both the Mouse page and the panel row.
**Merge:** needs the Command Bar row restored first.

ref #1012

iltonandrew:
Command Bar row restored, both surfaces widened — and widening them turned up a third site in the same corner that the two you named would have left half-done.

**1. The Command Bar row.** Your shape verbatim, in the `.scrollInverter` position directly above the guard. First id byte-identical (`toggle.mouseButtonShortcuts`), second `…​.spacesGesture`, and no new strings — `spacesEnableLabel` is the one this PR already added in 13 languages. Confirmed the guard is what does it: `enabledKeys.count` going 1→2 sends `toggleEntries` down the `return []` path.

**2. The two engaged-reads — and a third that made the fix worse before it made it better.** `accessibilityNoteVisible` and the panel's `needsAttention` both take the drag into account now, and `isEnabled(_:)` with them.

But widening `needsAttention` alone is not safe, and that is what I did first. Two arguments sit beside it on the same `PanelToggleRow`:

```swift
caption: caption(buttonStrings.panelCaption, needsAccessibility: mouseButtonShortcutsEnabled),
permissionAction: accessibilityPermissionAction(mouseButtonShortcutsEnabled),
```

and `accessibilityPermissionAction` is `guard enabled, !permissions.accessibility else { return nil }`. So with shortcuts off, drag on, Accessibility not granted, the row would have shown "Permission required" with a **dead grant button** and a caption that never mentions the permission. Strictly worse than the silence you reported: a dead end instead of nothing.

So rather than widen three arguments independently, the row now names the fact once:

```swift
let buttonsEngaged = mouseButtonShortcutsEnabled || spacesEnabled
```

and caption, `needsAttention` and `permissionAction` all read it. Three sites disagreeing is how this arrived; one name is what stops it recurring.

Left narrow deliberately: `accessoryTitle` (the "Manage" button). Its action opens the Mouse page, which is where the drag is configured too, so there is an argument for widening it — but it changes a visible label's presence for drag-only users and that reads as your call, not a defect. Say the word either way. `.onChange(of: mouseButtonShortcutsEnabled)` also stays as it is: that is the panel's own toggle reacting to itself, and the drag toggle has its own.

**Tests.** Five source-contract assertions: the two Command Bar rows (id derived from `AppFeature.mouseButtonShortcuts.rawValue`, not spelled out), the Settings permission section, the panel category count, and the row's permission surface. The `@AppStorage` property names are looked up from the source rather than hardcoded, so a rename of `spacesEnabled` does not quietly green them.

The last one is deliberately not a pin on the expression's spelling. It asserts the **invariant**: that `caption`, `needsAttention` and `permissionAction` all read one identifier, and that identifier's definition mentions both switches. Pinning `(a || b) && !permissions.accessibility` would have gone red on the very refactor that fixed the bug, which is the failure mode `AI-CONTRIBUTIONS.md` warns about. Mutation-proved: narrowing `permissionAction` back to `mouseButtonShortcutsEnabled` fails it by name, with the code otherwise untouched.

**On the branch moving.** `598786f` (Update branch) landed while this was in progress, bringing in five commits including your #987. I merged it forward and re-ran the gate on that base rather than force-pushing over it — a lease would not have caught that, since I had already fetched. It overlaps `Tests/MetricsTests.swift`, so that one was a real merge, not a replay.


**Closing the class rather than the instance.** Five rounds of this PR have ended with you finding one more site that reads a fact it moved, so instead of fixing only what you named I enumerated every reader of both. Two facts moved; both enumerate completely.

`enabledKeys` has exactly one reader in the tree — `CommandBarCatalog.swift:259-260`, the guard itself. Fixed above, and there is nothing else.

`mouseButtonShortcutsEnabled` has six, and the three not touched here are deliberately narrow:

| Site | Why it stays as it is |
| --- | --- |
| `MouseButtonSettings.swift:15` | it *is* the shortcut switch |
| `FeatureCatalog.swift:201` | the `enabledKeys` definition |
| `CommandBarCatalog.swift:248` | the first Command Bar row's key — correct by construction |
| `MouseButtonShortcutService.swift:87` | gates `wantsSideWheelEvents`; `MouseSpacesGestureSupport.canBind` refuses the side wheel, so the drag has no claim on it |
| `MouseButtonShortcutService.swift:301` | the down path, which you confirmed in the round on finding 3 |
| `MouseButtonShortcutSupport.swift:164` | `hasActiveShortcut` — narrow **on your instruction**: widening it lets the shortcut switch coming back on take the button, which is finding 2 again |

That last row is why "widen every reader" would be as wrong as widening only the reported one. If you disagree with any verdict in that table it is a one-line change, but I would rather you check six rows than hunt for a seventh site.

**Gate**, on the merged base: 0 warnings. `./build.sh --test` — 8850 checks, **1 failure**: the `dispatch pool starvation` precondition #1014 exists to skip, red on this 15-core M5 Pro and green on CI. `./build/Vorssaint --selftest` — `SELFTEST OK`. Note the count moved 8829 → 8850: those 21 came in with the Update-branch merge, which is why the pre-merge green did not count.

Mutation, narrowing `permissionAction` back to `mouseButtonShortcutsEnabled` and changing nothing else: failures go 1 → 2, the new one naming *the panel mouse-button row's caption, attention state and grant button all read one engaged flag built from both switches*.

Not verified: still no GUI run of the shortcuts-off + drag-on permission row. The dead-button path is reasoned from `accessibilityPermissionAction`'s guard and pinned, not driven through the panel by hand.

ref #1012

PathGao:
Row's back, both surfaces widened, and the six-row table reads right where I can check it. One correction to the enumeration, then the `accessoryTitle` call you left me.

**`enabledKeys` has two production readers, not one.** The second is `FeatureCatalog.swift:308-311`, inside `activeFeatures(using:)`:

```swift
let keys = feature.enabledKeys
guard keys.isEmpty || keys.contains(where: boolFor) else { return false }
```

Reached from `Permissions.swift:83-86` (whether accessibility polling runs at all) and `FeatureHubSettings.swift:525-529` (the hub's "used by" line and its permission status). It ORs across the keys, so your 1→2 edit already fixed it: drag on with shortcuts off now counts as engaged in both, where before it did not. That is a fourth engaged-read, it is correct, and nothing needs doing to it.

What I would flag is that it arrived by luck rather than by the enumeration. "exactly one reader in the tree ... and there is nothing else" was the completeness proof for the whole table, and it missed a reader that happened to be helped. The narrow claim does hold: `.count == 1` and `.first` on `enabledKeys` appear only at `CommandBarCatalog.swift:246-247`, so the row you restored is the only place the count going 1→2 subtracts anything.

**`accessoryTitle`: widen it, same `buttonsEngaged`.** It is the only one of the row's four non-switch surfaces still reading the narrow flag, which is the shape you just spent a commit removing. `isOn:` staying narrow is right, it *is* the shortcut switch and correct by construction. "Manage" is not a switch, it is the only way from the panel to the page, and the drag is configured on that page. Leave it narrow and a drag-only user gets a row that counts in the category header (`isEnabled` at `MenuPanelView.swift:1165`, feeding the `n/m` beside the category title, which you widened) and shows the permission surface, with nothing on it leading anywhere. `.onChange(of: mouseButtonShortcutsEnabled)` staying narrow I agree with, for the reason you gave.

Not verified: my copy of the diff still cuts partway through `MetricsTests.swift`, so the five source-contract assertions and the mutation result are your word rather than my reading. No GUI verification from me. Line numbers above are main's, so they sit lower on your branch.

**Impact:** none from the reader I found, that one is already right. Drag-only users get a panel row with no way to reach the page its own gesture is configured on.
**Merge:** needs `accessoryTitle` reading `buttonsEngaged` first.

ref #1012

iltonandrew:
`accessoryTitle` widened in `0ee1695`, and the enumeration correction is yours — I need to withdraw the claim, not just patch it.

**The completeness claim was false.** I wrote that `enabledKeys` has "exactly one reader in the tree ... and there is nothing else." It has two. `FeatureCatalog.swift:311` inside `activeFeatures(using:)` is the second, and you are right that my 1→2 edit already fixed it by accident rather than by design — it ORs across the keys, so drag-on-shortcuts-off now counts as engaged in the accessibility poll and the hub's status line.

The cause is worth naming because it is dull and repeatable. My sweep was:

```
grep -rn "enabledKeys" Sources/ | grep -v "FeatureCatalog.swift"
```

I excluded the whole **file** in order to filter out the **definition** at `:187`, and buried a reader living in the same file. Filtering by line was the correct move; filtering by file was the lazy one, and it is exactly the kind of shortcut that turns an enumeration into a guess wearing an enumeration's clothes. The table I offered as a proof was worth less than it looked.

The narrow claim does survive, and it is the one I should have made: `.count == 1` and `.first` on `enabledKeys` appear only at `CommandBarCatalog.swift:246-247`, so the restored row remains the only place the count change subtracts anything.

**`accessoryTitle`: your call taken, and your reasoning was the half I had not weighed.** I had it as a product question about what "Manage" means. The argument that settles it is that `isEnabled` already counts the row in the category header — which I widened in the same commit — so a drag-only user gets a row that counts, shows a permission surface, and leads nowhere. That is not a taste question, it is an incoherent row. All four non-switch surfaces now read `buttonsEngaged`; `isOn:` stays narrow because it *is* the switch, correct by construction.

The assertion covers the fourth surface now. It also caught a bug of mine before you could: `panelButtonRow` is whitespace-stripped, and my first version of the needle had a space before the `?`, so it failed at baseline **and** under mutation — a check that fails either way proves as little as one that passes either way. Fixed, then re-proved: baseline clean, and narrowing `accessoryTitle` again takes failures 1 → 2, naming the assertion.

**Gate:** 0 warnings, 8850 checks with **1 failure** — the `dispatch pool starvation` precondition #1014 exists to skip, red on this 15-core M5 Pro and green on CI. `SELFTEST OK`.

Still not GUI-verified from me: the drag-on-shortcuts-off panel row, in any of its four surfaces.

ref #1012

PathGao:
Both enumerations check out against main, so I'll stop pulling on that thread. `enabledKeys` has exactly two production readers — `CommandBarCatalog.swift:246-247` and `FeatureCatalog.swift:310` — and no third; `mouseButtonShortcutsEnabled` has the seven sites your table plus the two you widened account for. The `.count == 1` / `.first` pair is confined to the Command Bar guard, so the restored row was the only subtraction.

One surface neither of us had looked at, and it needs nothing. The two new keys are persisted, and a silently-untravelled binding is the kind of thing that surfaces a year later as "my mouse forgot the gesture after I restored": `SettingsBackupSupport.exportKeys()` (`:18-24`) builds from `Defaults.registeredDefaults`, so both keys travel by construction, and `valueLooksRight` type-checks them against the registered `Bool` and `Int`. No list to add them to. `FeaturePresets` writes `availabilityKey` and never `enabledKeys`, so the 1→2 change can't turn the drag on for anyone either.

Nothing else to flag.

Not verified: my copy of the diff still cuts partway through `MetricsTests.swift`, so the five source-contract assertions, `spacesCaptureRefusesPending` and the mutation results remain your word rather than my reading. No GUI verification from me at any point — the shortcuts-off + drag-on permission row in particular has only ever been reasoned and pinned.

**Impact:** none.
**Merge:** ship it.

ref #1012

PR #1025, opened 2026-08-26
fix(menu-bar): preserve battery icon aspect ratio in menu bar

### LINKED ISSUE #769: The battery menu bar item's icon shows up as a square instead of rectangle
### Description

I know it's small, but the battery icon is a square.
Image
Whenever you add the battery icon as its own icon within the menu bar, it shows up as a square instead of the wider variant shown within the app's settings
Image
While it doesn't affect the user experience in a major way and is more nit-picky, I'd like to see a fix for this.

### Steps to Reproduce

1. Go to the "Monitor" sidebar tab
2. Enable "Seperate metrics into their own items"
3. Enable "Battery"
4. Look at menu bar

### Screenshots / Recordings

_No response_

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

Yahddyyp:
Resolves #769

### Root cause

`MenuBarRenderer.batteryBlockImage(percent:isCharging:style:)` draws the battery SF Symbol into a rect whose **width is clamped** to the reserved column (`min(symbolWidth, symbolSize.width)`) but whose **height is left at the symbol's full natural size**, uncapped. `NSImage.draw(in:)` stretches its image to exactly fill the given rect, so this squeezes the symbol's width while keeping its full height — distorting the aspect ratio.

Since the battery SF Symbols are naturally wider than they are tall, and their natural width exceeds the reserved column, this consistently compressed the glyph into something closer to a square than the true wide rectangle.

### Fix

Scale both dimensions together from a single `scale` factor instead of clamping only width. The symbol keeps its natural proportions and simply shrinks to fit the reserved column when it's too wide, rather than getting stretched out of shape.

### Testing

- Rebuilt and verified in "Separate metrics into their own items" mode with Battery enabled — icon renders as a proper wide rectangle, matching the shape shown in the Settings preview.
- Confirmed no other block-style metric icon uses this draw pattern — `cpu`/`gpu`/etc. render text labels via `metricBlockImage`, not SF Symbols — so this was the only place the distortion could occur.

### Before and after 
**On battery**
image

**Charging**
image

You can increase the size of the battery symbol to make it look nicer but that is out of scope of this pr.

### DIFF
--- a/Sources/Vorssaint/App/MenuBarRenderer.swift
+++ b/Sources/Vorssaint/App/MenuBarRenderer.swift
@@ -1208,10 +1208,17 @@ enum MenuBarRenderer {
             if let symbol = NSImage(systemSymbolName: symbolName, accessibilityDescription: nil)?
                 .withSymbolConfiguration(symbolConfig) {
                 let symbolSize = symbol.size
+                // draw(in:) stretches the image to exactly fill the rect, so
+                // clamping only the width while leaving height at
+                // the symbol's full natural size squished wide glyphs like
+                // the battery icon. Scale both dimensions together to keep
+                // the glyph's own proportions.
+                let scale = min(symbolWidth / symbolSize.width, 1)
+                let drawSize = NSSize(width: symbolSize.width * scale, height: symbolSize.height * scale)
                 let symbolRect = NSRect(x: 0,
-                                        y: (height - symbolSize.height) / 2,
-                                        width: min(symbolWidth, symbolSize.width),
-                                        height: symbolSize.height)
+                                        y: (height - drawSize.height) / 2,
+                                        width: drawSize.width,
+                                        height: drawSize.height)
                 symbol.draw(in: symbolRect)
             }
             let valueAttrs = dynamicTextAttributes(font: valueFont)


PathGao:
The drawn width is unchanged in both branches, up to float rounding: when the glyph is too wide, `symbolSize.width * (symbolWidth / symbolSize.width)` is `symbolWidth`; otherwise it's `symbolSize.width * 1`. Both are what `min(symbolWidth, symbolSize.width)` already produced. So this only moves height, and the reserved column stays the same width — no shift for the blocks next to it.

The new `scale` still bounds one axis only. `min(symbolWidth / symbolSize.width, 1)` fits the column but not the bar, so a symbol whose natural height exceeds `height` gets clipped top and bottom — same shape of bug on the other axis. If the symbol point size is derived from the bar height upstream, that can't happen and it isn't worth a commit. `symbolConfig` and `height` are set outside this diff, so I didn't verify it either way. If you want the bound anyway it's one term: `min(symbolWidth / symbolSize.width, height / symbolSize.height, 1)`.

I only have this diff, so the body's claim that no other block-style metric icon uses this draw pattern is unverified on my side.

**Impact:** none beyond the battery glyph — width identical, height now proportional and smaller than before.
**Merge:** ship it.

ref #769

Issue #884, opened 2026-08-24
Display preview by default as an optional feature in clipboard manager


andreisuslov:
### Problem or Motivation

I'm frustrated when I need to toggle the preview button all the time. I'd like it to be enabled by default, ideally I don't need it at all because I would like to always be able to view the contents alongside with the metadata of the captured text/image/file - time, app, format, link, path.

### Proposed Solution

It would be great if Vorssaint could provide a functionality to disable the preview button in clipboard manager, enabling it by default, alongside with providing the metadata of the captured object.

### Feature Area

Clipboard and files

### Mockups / References

Raycast

salomaoparkour:
Would love this option as well. It's something available in Maccy that I miss in Vorssaint.

Issue #453, opened 2026-08-03
Command palette: clipboard image preview, and movable/resizable panel


BakaGa:
### Problem or Motivation

New command palette is great, but there's a lack of some functionality.

### Proposed Solution

- It would be great to have image preview and object data in clipboard menu of palette (just like tinycast or raycast)
- I want to be able to move and resize palette (it's a bit too narrow)

### Feature Area

Tools

### Mockups / References

This one is about clipboard thing
<img width="1118" height="731" alt="Image" src="https://github.com/user-attachments/assets/1d4616ee-937e-405b-be93-f77bba5e9f94" />

PathGao:
Triage note, part of a batch pass over the open issues. The labels are a contributor's read, not a maintainer ruling.

**What is missing here is the title, not the body.** `[Feature]:` has nothing after it. Please add one — an untitled issue is effectively invisible in the issue list and in search, and this one has two real asks in it.

On the content, it is also two independent requests: (1) image preview and object data in the command palette's clipboard view, and (2) making the palette movable and resizable because it is too narrow. Those touch different parts of the palette and would be much easier to track, and to close, as separate issues.

The attached screenshot covers the clipboard half only, so if the width complaint has a specific case behind it (a particular clipboard entry, a particular display), that would help too.

whitespacecode:
A ux cleanup would be nice to.

- When preview is enabled i still see a large image in the list
- More options, like edit a copied image (a setting I enabled in vorssaint)
- Or for text have options to pin, create snippets,.. So there's more of a flow

<img width="844" height="501" alt="Image" src="https://github.com/user-attachments/assets/b4ef96c5-07f9-4027-a1a2-925645ab49d3" />

PR #1026, opened 2026-08-26
fix(clipboard): stop Esc from closing the preview pane

naveenkrdy:
Pressing Esc in the Clipboard Quick Panel treated the preview sidebar as the first layer to back out of, so with preview on, a single Esc only retracted the sidebar and left the panel open - a second Esc was needed to actually close it. Preview should behave as a persistent view setting instead: once turned on, only clicking its own toggle button turns it off (Space also does, but only once arrow-key navigation has made a row's selection visible - see `ClipboardHistoryPreview.handlesSpace` - while the search field has focus, Space just types). Esc now goes straight to clearing a batch selection if one exists, or closing the panel, exactly as it does when preview is off.

`quickPreviewPresented` already persists to `UserDefaults`, and `hideHistoryWindow()` never touched it:

```swift
func hideHistoryWindow() {
    removeKeyMonitor()
    removeDismissMonitors()
    panel?.orderOut(nil)
    clearQuickBatchSelection()
}
```

so removing the preview branch is enough on its own - preview now survives both a panel-closing Esc and a full relaunch, matching how the toggle button already worked.

Moved the decision into a small pure function, `ClipboardHistoryEscape.action(batchCount:)`, next to the existing `ClipboardHistoryPreview.handlesSpace` helper it sits beside in `ClipboardHistorySupport.swift`, and added two `expect` cases pinning it in `MetricsTests.swift`. The function takes no preview argument by design - the fix removes preview from the Esc decision entirely, so there's nothing preview-related left for it to take.

Checked every other place `quickPreviewPresented`/`toggleQuickPreview` is touched, since the same mistake tends to sit at more than one call site:
- The toolbar preview button (`ClipboardQuickPanelView.swift`) - a click, untouched.
- The preview sidebar's own close button - a click on the preview UI itself, untouched.
- The Space-bar Finder-style Quick Look toggle - a separate, pre-existing affordance, untouched, but see the precise precondition above.
- The Esc branch - the only unwanted path, and the only thing this PR changes.

No CHANGELOG.md edit, no linked issue (searched open issues/PRs for existing reports of this before starting; the two related open issues, `#884` and `#453`, are about making preview the default and about the command palette's own clipboard preview/resizing respectively - neither is what this PR fixes).

## Verification

```
./build.sh --test              # TESTS OK, including the two new expects
```

On macOS 26.6.2 (build 25G83), Apple Silicon, own local build (ad hoc signed dev environment, not release-signed).

What I could not verify myself: an interactive click-through of the actual panel (open panel, turn preview on, press Esc, confirm it closes with preview still flagged, reopen and confirm preview is still showing) - this session has no Accessibility permission to drive the UI, so I could not simulate the keystroke/click sequence end to end. The logic itself is covered by the two new tests and by reading `hideHistoryWindow()`'s implementation above, but a real click-through on a Mac with Accessibility available is worth doing before merge.

### DIFF
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
@@ -1127,14 +1127,9 @@ final class ClipboardHistoryService: ObservableObject {
             }
             let modifiers = event.modifierFlags.intersection([.command, .option, .shift, .control])
             if event.keyCode == UInt16(kVK_Escape) {
-                // Esc backs out one layer at a time: preview, selection,
-                // then the window.
-                if self.quickPreviewPresented {
-                    self.setQuickPreviewPresented(false)
-                } else if self.quickBatchCount > 0 {
-                    self.clearQuickBatchSelection()
-                } else {
-                    self.hideHistoryWindow()
+                switch ClipboardHistoryEscape.action(batchCount: self.quickBatchCount) {
+                case .clearBatchSelection: self.clearQuickBatchSelection()
+                case .hideWindow: self.hideHistoryWindow()
                 }
                 return nil
             }
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistorySupport.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistorySupport.swift
@@ -287,8 +287,10 @@ enum ClipboardHistoryEscape {
     }
 
     /// Esc backs out one layer at a time - selection, then the window.
-    /// Preview is a persistent view setting, not a layer: it stays open
-    /// until its own toggle (the button, or Space for Quick Look) is used,
+    /// Preview is a persistent view setting, not a layer: only clicking its
+    /// own toggle turns it off (or Space, but only once arrow-key navigation
+    /// has made a row's selection visible - see `ClipboardHistoryPreview
+    /// .handlesSpace`; while the search field has focus, Space just types),
     /// so a keystroke meant to close the panel can never silently re-hide
     /// it first.
     static func action(batchCount: Int) -> Action {
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistorySupport.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistorySupport.swift
@@ -280,6 +280,22 @@ enum ClipboardHistoryPreview {
     }
 }
 
+enum ClipboardHistoryEscape {
+    enum Action: Equatable {
+        case clearBatchSelection
+        case hideWindow
+    }
+
+    /// Esc backs out one layer at a time - selection, then the window.
+    /// Preview is a persistent view setting, not a layer: it stays open
+    /// until its own toggle (the button, or Space for Quick Look) is used,
+    /// so a keystroke meant to close the panel can never silently re-hide
+    /// it first.
+    static func action(batchCount: Int) -> Action {
+        batchCount > 0 ? .clearBatchSelection : .hideWindow
+    }
+}
+
 enum ClipboardHistoryBatch {
     static func combinedText(_ texts: [String]) -> String {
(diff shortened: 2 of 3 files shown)


PathGao:
Four things.

`ClipboardHistoryEscape.action(batchCount:)` takes no preview state, so the expect named "regardless of preview" can't fail on a preview regression. Put an `if self.quickPreviewPresented` branch back at the call site and both new expects still pass — they only pin the batch branch, which was never in question. Either drop that clause from the message, or make preview an argument the function ignores:

```swift
static func action(batchCount: Int, previewPresented: Bool) -> Action
```

The body says preview "stays open until its own toggle (the button, or Space for Quick Look) is used". From the two existing expects, `handlesSpace` requires `selectionIsVisible: true` and no modifiers — so while the search field has focus, Space types a space and there is no keyboard way out of preview any more. Esc was that path. Fine if that's intended, but the body reads as if Space is an unconditional fallback. Inferred from the test names only; the `handlesSpace` call site isn't in the diff.

`hideHistoryWindow()` isn't in the diff either. "never touched it" is what makes removing the branch sufficient, and I can't confirm it from what I can see.

The red check is `Point out a reference that will not reach the reporter`. Bare `#884` / `#453` in the body cross-reference two issues this PR doesn't fix, which is my guess at the trigger — backtick them and they stop rendering as references. I can't read the workflow, so I don't know whether it's a gate or the job erroring.

**Impact:** none on the main path — Esc closing the panel in one press is right. What's open is the keyboard exit from preview when the search field has focus, and a test that claims more coverage than it has.
**Merge:** needs the red check resolved first.

naveenkrdy:
Thanks, all four land. Fixed the two code issues, replying on the other two.

**1. Test overclaimed.** Agreed, and I went with dropping the clause rather than adding an ignored `previewPresented` parameter - an argument the function never reads would be worse than the wording it was covering for. `ClipboardHistoryEscape.action(batchCount:)` has no preview parameter by design: the fix removes preview from the Esc decision entirely, so there's nothing preview-related left for the function to take. Reworded the expect to just "Esc closes the panel when nothing is selected."

**2. Doc comment overclaimed too, and you traced it exactly right.** `handlesSpace(selectionIsVisible:hasModifiers:)` requires `quickSelectionIsVisible`, which only goes true after arrow-key navigation (`moveQuickSelection`) - while the search field has focus, Space just types, it doesn't toggle anything. So with preview on and the search field focused, there genuinely is no keyboard exit from preview anymore. That's not a gap though - it's exactly the requirement this PR is for: preview should only come off by clicking its own toggle, full stop, and Esc was the one keyboard path that broke that. Reworded the comment to state Space's actual precondition instead of implying it's an unconditional alternative.

**3. Fair, quoting it since it's not in the diff:**
```swift
func hideHistoryWindow() {
    removeKeyMonitor()
    removeDismissMonitors()
    panel?.orderOut(nil)
    clearQuickBatchSelection()
}
```
No reference to `quickPreviewPresented` anywhere in it, confirmed by reading the current file, not by memory of what it used to do.

**4. Backticked `#884` and `#453` in the description.** Your guess looks right - bare `#123` in a PR body is exactly what GitHub autolinks as a cross-reference, so that's almost certainly what the bot was reacting to. I don't have visibility into the workflow either, so I can't promise it clears, but it's the direct fix for the stated trigger and costs nothing if it turns out to be something else.

Pushed both code fixes plus the description edit.

PathGao:
Red check is gone and both rewordings are in the diff I can see: the expect pins only what the function decides, and the doc comment names `quickSelectionIsVisible` as Space's precondition instead of implying it's an unconditional alternative. The quoted `hideHistoryWindow()` closes the last open item. Nothing else to flag.

The click-through stays unrun on my side too — I can't drive the UI either, so that one lands on whoever merges.

**Impact:** none on the main path. Esc closes the panel in one press, preview survives close and relaunch. Losing the keyboard exit from preview while the search field has focus is the intended behavior per the body, not a regression.
**Merge:** ship it.

Issue #735, opened 2026-08-21
More responsive/adjustable window switcher


4thpla:
### Problem or Motivation

Would like to unify my apps, so I tried switching from altTab to vorssaint. However, the altTab implementation is just snappier, probably due to lower delay when hitting command+tab. 

### Proposed Solution

There should be a setting to change how fast the window switcher shows up.

### Feature Area

Windows and Dock

AB-boi:
really? I feel it's pretty snappy already.

MattOrmianek:
I had the same thought. I also want to switch from AltTab, and the preview looks better on AltTab. The customisation options are better, and the whole thing feels snappier and faster.

MattOrmianek:
It would also be great to have the option of not showing the windows of minimised apps, and showing just the app name rather than the app name and window names. There should also be an option to show the app switcher only on the monitor that is in focus, with apps/windows only from that monitor.

This is a super nice-looking proposition:

<img width="1026" height="612" alt="Image" src="https://github.com/user-attachments/assets/b5da883b-c02e-4363-bbda-599c1d0b9cea" />

The best I can set up right now is:

<img width="2822" height="798" alt="Image" src="https://github.com/user-attachments/assets/74df97da-cb9a-4998-ae71-9e91c865389c" />

Alexander3847575:
Yes please, the delay is causing me a big headache. I'd like command tab to be instant like it was before.

PathGao:
Triage note, part of a batch pass over the open issues. The labels here were applied by a contributor, not a maintainer ruling.

There is an exactly isomorphic precedent one feature over, so this is close to a copy job.

Today the switcher's appearance delay is a hard-coded constant: `private static let appearanceDelay: TimeInterval = 0.1` at `Sources/Vorssaint/Services/Switcher/AppSwitcher.swift:110`, with a single consumer at `:1287`.

Dock Preview already did this same conversion for its hover delay: the key `DefaultsKey.dockPreviewOpenDelay` (`Core/Defaults.swift:75`), registered with a default at `Defaults.swift:809`, and a clamped range `DockPreviewSupport.openDelayMillisecondsRange = 200 ... 900` with a clamp helper (`DockPreviewSupport.swift:102-106`). That shipped in commit `6ae0538` ("feat(dock-preview): open previews on an adjustable delay").

[...]

PR #1028, opened 2026-08-26
feat(switcher): make appearance delay adjustable

yasinozmeen:
## Summary

Makes the existing App Switcher appearance delay adjustable from 0 to 500 ms while preserving the current 100 ms default. A 0 ms value provides immediate panel appearance; non-zero values keep quick shortcut taps able to switch without flashing the panel.

This stays inside the existing App Switcher design: it reuses the current delayed `DispatchWorkItem` and follows the neighboring Dock Preview delay pattern. It adds one persisted setting, one slider, and localized labels, with no new dependency, runtime service, permission, or security surface.

## Verification

- MacBook Pro (Mac16,1), Apple M4, 16 GB; macOS 26.5.2 (25F84).
- `./build.sh --test` passed with `TESTS OK (8659 checks)`; Swift compilation was limited to 2 jobs because the unrestricted parallel build was killed under local memory pressure.
- `./build.sh --dev` completed without warnings; `./build/VorssaintDeveloper --selftest` returned `SELFTEST OK`.
- `./build.sh` completed without warnings; `./build/Vorssaint --selftest` returned `SELFTEST OK`.
- The Turkish developer Settings UI showed the 100 ms default and updated both the slider and caption across 0, 100, 125, and 500 ms.
- The real Accessibility shortcut path reached `scheduleShowPanel()` and `showPanel()`. LLDB timings were 32.650 ms at 0 ms (debugger baseline), 112.632/119.191 ms at 100 ms, and 541.834 ms at 500 ms; the 0-to-500 difference was 509.184 ms.
- The release app in `/Applications` was not replaced or modified. The developer bundle used the separate `com.vorssaint.utils.dev` identifier.
- Screen Recording was not granted to the developer bundle, so window-preview image rendering was not verified. Display and Spaces layout were not explicitly controlled.

## Anything else

Refs #735


PathGao:
The caption promises a behavior the 0 ms endpoint breaks. Every one of the 13 localizations ships the same unconditional second sentence — English: "Quick taps still switch without showing it." At 0 ms that is false, and 0 ms is the setting this PR exists to add. Your body already draws the line the caption doesn't:

> A 0 ms value provides immediate panel appearance; non-zero values keep quick shortcut taps able to switch without flashing the panel.

Cheapest fix is deleting the second sentence from all 13 strings. The first sentence already covers it — "held before the switcher appears" reads correctly at 0 ms without any extra clause. Making it conditional instead costs 13 retranslations for a sentence that is only interesting at one end of the slider.

Nothing else at tier 2 or above. The clamp lives on the read side in `SwitcherSupport.appearanceDelay`, so a restored backup or a hand-edited plist can't push the timer out of range — that's the right place for it, not the binding alone. The default is referenced, not copied, in `Defaults.swift` and in the `@AppStorage` initializer.

Not verified: the Dock Preview delay row isn't in the diff, so I can't confirm this row matches its layout or that it also renders `ms` unlocalized. Same for `SettingsBackupSupport.exportKeys()` — the new test asserts the key is in there and you report the suite green, so I'm treating "it derives from the registered defaults" as answered rather than checked.

**Impact:** no behavior change. Settings tells users at 0 ms that quick taps won't show the switcher, which is the opposite of what they will see.
**Merge:** needs the caption sentence resolved first.

ref #735

yasinozmeen:
Resolved in `ec7e4e8`. I removed the misleading second sentence from all 13 localized captions, so the text remains accurate at the 0 ms endpoint.

Verification:
- `./build.sh`
- `./build.sh --test` — `TESTS OK (8659 checks)`

Thanks for catching this.

PathGao:
Checked the 13 captions in the diff: the second sentence is gone from all of them, and the remaining sentence reads correctly at 0 ms. That was the only gap I named, so `incomplete` comes off.

Unchanged from before, still unverified: the Dock Preview delay row isn't in the diff, so I can't confirm this row matches its layout or that it renders `ms` unlocalized the same way. No check runs are reported on this PR either, so "CI green" is your call, not something I observed.

**Impact:** none.
**Merge:** ship it.

ref #735

Issue #937, opened 2026-08-25
Synthesized keystrokes have no owner: eleven call sites that disagree


PathGao:
Vorssaint presses keys on the user's behalf in eleven places. No two of them agree on how, several contradict each other in writing, and I do not think I should pick the winners on my own — some of the answers can only be settled on hardware, and some are your call about intent.

This is not a bug report. It is one surface that has never had an owner, and the individual bugs it produces keep arriving one at a time under different feature names.

## The census (`origin/main`)

**Posts ⌘V:**

| | event source | tap | down→up gap | waits for a clean keyboard | secure input | Accessibility |
|---|---|---|---|---|---|---|
| `PastePlainService.swift:255` | `nil` | HID | 0.04 s | yes | no | yes |
| `CommandBarService.swift:1985` | `.hidSystemState` | HID | 0 | yes | yes | yes |
| `ClipboardHistoryService.swift:1015` | `.hidSystemState` | HID | 0 | **no** | **no** | **no** |
| `FinderCutPaste.swift:413` | `.hidSystemState` + marker | HID | 0.04 s | n/a | no | n/a |

**Posts some other shortcut:**

| | event source | tap | gap |
|---|---|---|---|
| `RadialMenuService.swift:701` | `nil` | HID | 0.04 s |
| `SpaceWindowBridge.swift:281` | `nil` | HID | 0 |
| `MouseButtonShortcutService.swift:364` | `.hidSystemState` | HID | 0 |
| `MouseNavigationService.swift:332` | `.hidSystemState` | **session** | 0 |
| `SuperKeyService.swift:577` | `.hidSystemState` | **session** | 0 |
| `DockClickService.swift:538` | `.hidSystemState` | HID | full chord sequence |

**Types text:** `TextSnippetService.postExpansion` (`:338`), `.hidSystemState` + marker.

Supporting copies: the "wait until no modifier is physically down, 15 ms × 100, then a 0.06 s settle" block exists five times — `PastePlainService.swift:127`, `CommandBarService.swift:1967`, `CommandBarExtras.swift:13`, `SnippetLibraryService.swift:185`, `RadialMenuService.swift:689`. Two synthetic-event markers exist: `0x564F5253` (`TextSnippetService.swift:16`) and `0x564F5249` (`FinderCutPaste.swift:82`), each filtered only by its own tap.

## Four places the code contradicts itself in writing

**1. Event source.** `PastePlainService.swift:252` argues for no source at all:

[...]

PR #1032, opened 2026-08-26
refactor(quick-tools): extract synthetic paste into a shared helper

naveenkrdy:
## Summary

First of the four PRs @PathGao suggested splitting #713 into. Pulls the synthetic-⌘V/modifier-wait/pasteboard-snapshot logic out of `PastePlainService` into a new `SyntheticPasteSupport.swift`, so the Selection Actions feature (and #917's `TextSnippetService`) can reuse it instead of each carrying its own copy.

Behavior-identical for Paste Plain - same waits, same event posting, same restore logic, just moved. One addition: `postCmdV`/`deleteSelection` now check `IsSecureEventInputEnabled()` before posting, the same guard `CommandBarService`, `TextSnippetService` and `SnippetLibraryService` already use, but missing from this one - a plain-input password field would previously post a synthetic ⌘V that the system silently drops; now it beeps like the others instead. Not every synthetic-paste call site has this guard: `ClipboardHistoryService.pasteIntoPreviousApp` (`ClipboardHistoryService.swift:1041`) posts ⌘V with no secure-input check and no beep, same gap this PR closes here, left open for a later PR - fixing it isn't in scope for this one.

Scope note: no `quick-tools`-adjacent scope exists for the new shared file's own home (it lives directly under `Services/`, not `Services/QuickTools/`), but the only change to existing behavior is inside `PastePlainService.swift`, so I kept the `quick-tools` commit scope. Selection Actions PRs 2-4 (feat, no existing scope to reuse) will need a new `selection-actions` scope - flagging that here since this is the first PR in the sequence.

## Test plan

- [x] `./build.sh --test` passes (`TESTS OK`, 8641 checks)

Manual: exercised Paste Plain (strip formatting, paste, verify rich content restored) on this machine. Did not have a password field or other secure-input context handy to trigger the new `IsSecureEventInputEnabled` branch directly - the guard mirrors the existing one in `CommandBarService.swift` verbatim, so I'm confident in the logic, but flagging that specific path as unexercised.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/PastePlainService.swift
+++ b/Sources/Vorssaint/Services/QuickTools/PastePlainService.swift
@@ -3,7 +3,6 @@
 
 import AppKit
 import ApplicationServices
-import Carbon.HIToolbox
 
 /// Pastes the clipboard as plain text on a global shortcut: strips fonts,
 /// colors and links, pastes, and quietly puts the original rich content back
@@ -81,7 +80,7 @@ final class PastePlainService: ObservableObject {
         if let pending = pendingRestore, pasteboard.changeCount == pending.plainChangeCount {
             snapshot = pending.snapshot
         } else {
-            snapshot = Self.snapshot(of: pasteboard)
+            snapshot = SyntheticPasteSupport.snapshot(of: pasteboard)
         }
         restoreWork?.cancel()
         restoreWork = nil
@@ -124,29 +123,22 @@ final class PastePlainService: ObservableObject {
     /// read clean matters: posted right on the release, the target app can
     /// still see the stale modifier state and refuse the key equivalent.
     private func postPasteWhenModifiersReleased(attempt: Int, completion: @escaping () -> Void) {
-        let held = CGEventSource.flagsState(.combinedSessionState)
-            .intersection([.maskCommand, .maskAlternate, .maskShift, .maskControl])
-        if held.isEmpty || attempt >= 100 {
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.06) {
-                let shortcut = GlobalShortcut.saved(for: DefaultsKey.pastePlainShortcut,
-                                                    fallback: .pastePlainDefault)
-                let mustReleaseHotkey = shortcut.isStandardPasteCommand
-                if mustReleaseHotkey {
-                    // Otherwise our global ⌘V catches the synthetic ⌘V below,
-                    // so the target app never receives a paste command.
-                    self.hotkey.unregister()
-                }
-                Self.postPasteShortcut {
-                    if mustReleaseHotkey {
-                        self.syncWithPreferences()
-                    }
-                    completion()
-                }
+        SyntheticPasteSupport.waitForCleanModifiers(attempt: attempt) { [weak self] in
+            guard let self else { completion(); return }
+            let shortcut = GlobalShortcut.saved(for: DefaultsKey.pastePlainShortcut,
+                                                fallback: .pastePlainDefault)
+            let mustReleaseHotkey = shortcut.isStandardPasteCommand
+            if mustReleaseHotkey {
+                // Otherwise our global ⌘V catches the synthetic ⌘V below,
+                // so the target app never receives a paste command.
+                self.hotkey.unregister()
             }
-            return
-        }
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.015) { [weak self] in
-            self?.postPasteWhenModifiersReleased(attempt: attempt + 1, completion: completion)
+            SyntheticPasteSupport.postCmdV {
+                if mustReleaseHotkey {
+                    self.syncWithPreferences()
+                }
+                completion()
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


PathGao:
`deleteSelection` and `replaceSelection` aren't extracted — they're new. Nothing was removed from `PastePlainService.swift` to match them, and nothing in this PR calls them. The body says this pulls "the synthetic-⌘V/modifier-wait/pasteboard-snapshot logic out of `PastePlainService`": those three did move, these two came from nowhere. ~60 lines of `GeneralPasteboardAccess` sequencing and `ignoreNextChange` bookkeeping land in a `refactor` commit with no call site to exercise them, and by PR 2 they're already in main, unreviewed.

Drop both here. Bring each one in with its first caller.

Separate point, whatever shape the file lands in: `postCmdV` does not wait for modifiers, `deleteSelection` does — it wraps itself in `waitForCleanModifiers`. Two functions in the same enum, opposite contracts, nothing in either signature says which. `PastePlainService` gets it right only because the wait is still visible at the call site. PRs 2-4 have to remember. Either fold the wait into `postCmdV` and drop the separate call in `replaceSelection`, or name it `postCmdVAssumingCleanModifiers`.

The secure-input guard belongs there, but I can't check it: you didn't exercise it, and `CommandBarService.swift` — the version it's said to mirror verbatim — isn't in the diff, so I only have your description of it.

**Impact:** Paste Plain in a secure-input field goes from silently doing nothing to a beep. Everything else moves without changing behavior.
**Merge:** needs `deleteSelection` + `replaceSelection` pulled out of this PR first.

ref #937

naveenkrdy:
Fair, and I don't have a good defense for shipping them unexercised - dropped both. They'll come back in the Selection Actions PR that actually calls them.

Also renamed `postCmdV` to `postCmdVAssumingCleanModifiers`, since that's the actual contract (no internal wait, unlike `deleteSelection`) and nothing in the old name said so. `PastePlainService` updated to match.

Left the secure-input guard as-is, and here's the block it mirrors (`CommandBarService.swift`, current `main`):

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 0.06) {
    guard !IsSecureEventInputEnabled() else {
        NSSound.beep()
        return
    }
    ...
}
```

Same check, same beep-on-secure-input response, before the same kind of synthetic key post.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8641 checks).

PathGao:
Both functions are gone and the rename landed. What's left is a straight move plus the guard.

Checked the guard against `main` now that you quoted it: `CommandBarService.swift:2001` is the same check, same beep, and the ordering matches — its guard sits inside the `+0.06` block, and yours runs from `waitForCleanModifiers`' completion, which is that same beat. That part I'm no longer taking on description.

The body's coverage claim is wrong, though. `ClipboardHistoryService.swift:1041` posts ⌘V with `.maskCommand` and no secure-input check, reached from `pasteIntoPreviousApp` — so picking an entry out of Clipboard History into a password field silently does nothing today, no beep. It isn't in your list of "every other synthetic-paste call site", and it's the same gap you're closing here. Not this PR's job to fix it; either drop the "every other" or name it as a target for the later PRs.

One thing for whoever adopts this next: the two timeout policies are opposites. `waitForCleanModifiers` pastes anyway at `attempt >= 100` — PastePlainService's original contract, correctly preserved. `CommandBarService.swift:1990` beeps and gives up instead. The doc comment says this enum is "shared by every feature that needs to replace ... selected text", but CommandBarService can't move onto it without changing behavior or the helper growing a policy parameter. Fine to leave as is; just don't let PR 2-4 assume the helper matches what CommandBar does.

I didn't exercise the secure-input path either — same reason you didn't.

**Impact:** Paste Plain into a secure-input field goes from silent no-op to a beep. Nothing else changes.
**Merge:** ship it.

ref #937

naveenkrdy:
You're right, the "every other synthetic-paste call site" claim was wrong - fixed the body instead of dropping the clause: `ClipboardHistoryService.pasteIntoPreviousApp` (`:1041`) posts ⌘V with no secure-input check and no beep, same gap, not covered here. Named it explicitly as a follow-up target rather than silently narrowing the sentence.

Noted your two points on the timeout-policy mismatch and the doc comment's "shared by every feature" phrasing for whoever picks this up for PR 2-4 - no code change needed here, just flagging that PastePlainService's contract (paste anyway past the timeout) and CommandBarService's (beep and give up) aren't the same, so CommandBar can't adopt this helper as-is.

PathGao:
Heads-up that will save you a rebase: the three functions this PR extracts into `SyntheticPasteSupport` already exist on main, in `Sources/Vorssaint/Services/TransientPaste.swift`.

| this PR | on main |
|---|---|
| `snapshot(of:)` | `TransientPaste.snapshot(of:)` (:108) |
| `waitForCleanModifiers` | `TransientPaste.postPasteWhenModifiersReleased` (:124) |
| `postCmdVAssumingCleanModifiers` | `TransientPaste.postPasteShortcut` (:153) |

`TransientPaste` also carries the parts an extracted copy would still need — restoring the pasteboard afterwards, a change-count check so a copy made mid-flight is not clobbered, and a `didFail` callback. `TextSnippetService` is a second caller of it already, so the shared helper this PR is aiming at is there, just under a different name.

The diff here is against a `PastePlainService` older than that, so it is not only a conflict — the destination has moved. It looks like #1037 carries the same new file and would inherit the same rewrite.

Worth saying plainly: I have a small PR of my own open against `PastePlainService` (#1238, moving its clipboard read onto the shared serial lane), so this is not a neutral observation about which of us rebases. The reason I am posting it is the file, not the ordering — if `SyntheticPasteSupport` lands, the repo ends up with two answers to "how do I synthesise a paste", which is the thing an extraction is supposed to prevent.

PathGao:
Closing on the comment above: all three functions this extracts already exist on `main` in `Sources/Vorssaint/Services/TransientPaste.swift`, which also carries the pasteboard restore, the change-count guard and the `didFail` callback an extracted copy would still need.

Point #1033 at `TransientPaste` instead and you skip the rebase entirely.

PR #1033, opened 2026-08-26
feat(selection-actions): add floating menu with copy, cut, paste, delete

naveenkrdy:
## Summary

Second of the four PRs @PathGao suggested splitting #713 into (first was #1032, the `SyntheticPasteSupport` extraction this one is based on). This is the Selection Actions skeleton: the floating bar, the selection-reading machinery, the settings page, exclusion lists, drag-to-reorder, and its own optional global shortcut - wired up with only the four actions that need nothing beyond Accessibility.

- **Copy, Cut, Paste, Delete** - the only actions in this PR. No network access, no AppleScript automation permission requested (`AppFeature.selectionActions.permissions` is `[.accessibility]` only here; `.automationTerminal` arrives in PR 4 with Run in Terminal).
- Reads the current selection via Accessibility (`SelectionReader`), falling back to a shadow ⌘C (`ShadowCopySelectionReader`) when an app doesn't expose `AXSelectedText` - diffs the pasteboard before/after and restores it, same pattern `QuickTools` already uses elsewhere.
- Per-app and per-website exclusion lists, icon/word display style, and drag-to-reorder all ship now since they're part of the skeleton, not any one action - a settings page with only 4 toggles and nowhere to exclude anything felt like the wrong place to stop.
- Ships as an opt-in beta feature (`AppFeature.selectionActions`), off by default, matching Fan Control and Kill Process.

Text transforms (uppercase/lowercase/sort/calculate/etc., PR 3) and actions that leave the machine (search, translate, send to AI, Run in Terminal, currency conversion, PR 4) follow as separate PRs based on this one's merge commit, per the split @PathGao proposed.

**Scope note:** no existing commit scope fits a brand-new feature, so this introduces `selection-actions`.

## Test plan

- [x] `./build.sh --test` passes (`TESTS OK`, 8641 checks)

Manual: enabled the feature, selected text in TextEdit and Safari, confirmed the bar shows only actions that apply (Paste appears for an empty editable field, the rest don't), exercised Copy/Cut/Paste/Delete, reordered actions by drag, added an app and a website to the exclusion lists and confirmed the bar stops appearing there, toggled icon/word display style, set the optional shortcut.

Not tested: the shadow-⌘C fallback path specifically (didn't have an app on hand that refuses `AXSelectedText` but does support ⌘C) - the logic mirrors the pasteboard-diff-and-restore pattern already used by `SyntheticPasteSupport`/`ShadowCopySelectionReader`'s sibling code elsewhere in the app, but flagging it as unexercised on this pass.


naveenkrdy:
Stacked on #1032 (not yet merged) - this diff will shrink once that one lands and I rebase onto `main`. Flagging so the current diff isn't confusing in the meantime.

PathGao:
Three things, one worth fixing before the rebase because it sets the storage format PRs 3 and 4 inherit.

**Storing the disabled set means PR 3/4 actions arrive switched on.** `selectionActionsDisabledActions` records what's *off*, and its registered default is computed from the current `allCases` (`SelectionActionCatalog.defaultDisabledStorageValue`). Someone who never opens the settings page is fine — registration recomputes on the new build. Someone who flips one switch writes a concrete string, and every case appended later is absent from it, so it's enabled. PR 4's list is search, translate, send to AI, Run in Terminal. Those would ship on by default for exactly the people who bothered to configure the feature. Store the enabled set and the polarity fails the safe way instead.

**The domain exclusion check runs on the main thread.** `performRead` calls `isFrontmostAppExcluded()` before it dispatches, and `performRead` is only ever reached on main (the `asyncAfter` work item, or the hotkey callback). Non-empty website list plus a known browser in front is `kAXFocusedWindow` + `kAXDocument` + `kAXURL`, each under `AXUIElementSetMessagingTimeout(element, 0.35)` — up to ~1s of main thread per drag-select. The selection read itself is correctly on `.userInitiated`; this guard in front of it isn't. The comment on `isFrontmostAppExcluded` says it's "checked before Accessibility is ever touched, so an excluded app costs nothing beyond reading its bundle ID" — that holds for the app list, not for the domain check on the next line.

**`selectionActionsMaxVisible` has two defaults.** `Defaults.swift` registers 4, `handleRead` reads `as? Int ?? 8`. Registration wins so the 8 never fires today, but it's the same fact written twice with two values.

Not verified: the diff I can see is truncated at ~80k of 142k. `SelectionActionBarController`, `ShadowCopySelectionReader`, `SyntheticPasteSupport` and the settings page aren't in it, so none of the above says anything about whether the bar panel is nonactivating — which is what decides where `deleteSelection()`'s synthetic keystroke actually lands.

**Impact:** the storage-format item hits users of PRs 3/4, not this one. The main-thread AX round-trip is live here for anyone with a website exclusion set.
**Merge:** hold for #1032 to land — fold the first two into the rebase.

naveenkrdy:
Thanks for catching both of these - they're real, and I don't think either was on my radar until you flagged it.

**1. Enabled-vs-disabled storage direction.** You're right, and it's exactly the failure mode you described: anyone who touches even one toggle freezes a "disabled" string that a future update's new actions are absent from, which reads as "not disabled" - on by default for people who went out of their way to customize this. Flipped it to persist the *enabled* set instead, so anything added later (all of PR 3's transforms, PR 4's network-reaching actions) starts off until someone explicitly turns it on. Renamed the defaults key (`selectionActionsEnabledActions`) and the catalog API (`enabledActions(from:)`, `isEnabled(_:enabledRaw:)`) to match, so nothing keeps the old name with new meaning.

**2. Main-thread stall on the domain check.** Also right - `isFrontmostAppExcluded()`'s app-bundle branch is genuinely cheap like the comment claimed, but the domain branch walks `BrowserURLReader.currentHost` (`kAXFocusedWindow`/`kAXDocument`/`kAXURL`, each with its own AX timeout), and that was running synchronously before the existing background dispatch instead of inside it. Moved the whole check into `performRead`'s `DispatchQueue.global` block and corrected the comment to say which half is actually free.

**3. maxVisible duplication.** Already numerically consistent (both `8`) since I bumped the registered default in PR 3 once there were enough actions for that number to mean something - so no live bug today, just the same fact in two places. Can pull it into a shared constant if you'd rather not leave two literals that could drift again; didn't want to add an abstraction for one duplicated `8` without knowing if that's worth it to you.

Also noted your point about the truncated diff - `SelectionActionBarController` is a plain `NSPanel` (`.nonactivatingPanel` style mask), so `deleteSelection()`'s synthetic keystroke lands in whatever was frontmost before the bar opened, not the bar itself. Happy to paste any specific file if that's faster than working around the size limit.

Pushed both fixes to this branch. Agreed on holding merge for #1032 first - will rebase onto `main` once that's in.

PathGao:
`defaults.integer(forKey:)` and delete the `?? 8`. Registration guarantees the key is present, so the fallback is unreachable, not a safety net — dropping it leaves one literal without adding a constant to hold it.

```swift
let maxVisible = defaults.integer(forKey: DefaultsKey.selectionActionsMaxVisible)
```

One correction: on this branch they aren't both `8`. `Defaults.swift` registers `4`, `handleRead` falls back to `8`. The bump you're thinking of is in PR 3, which isn't in this diff. No live bug either way since registration wins, but the disagreement is here, and the line above closes it without a new abstraction.

`.nonactivatingPanel` settles the `deleteSelection()` question — that was the only thing I had riding on the truncated files.

**Impact:** none — the fallback is unreachable today.
**Merge:** hold for #1032, unchanged.

naveenkrdy:
Dropped the `?? 8` - `defaults.integer(forKey:)` alone now, matching your reasoning exactly (registration guarantees presence, so the fallback was dead weight without a named constant to justify it).

Thanks for the correction on the actual values (4 vs 8 registered here, not both 8) - that's accurate, I'd misstated it in my last reply.

Rebased onto #1032's trimmed PR. `deleteSelection`/`replaceSelection` are back - this is their real first caller (`SelectionActionRunner`'s Cut/Paste/Delete), and `replaceSelection` now calls the renamed `postCmdVAssumingCleanModifiers` internally.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8641 checks).

PathGao:
Moving `isFrontmostAppExcluded()` into the background block is right, but it left two things behind in those same eight lines.

**`SelectionActionsExcludedApps.isExcluded` now reads `@Published apps` off the main thread.** The settings page writes that array on main — `add`/`remove` → `reload()` → `apps = sanitized` — so an Add/Remove landing while a drag-select is resolving is a live `Array` race, not a stale read. `ClipboardIgnoredApps`, which this class copies shape for shape, pins its own query to main and says so in the doc comment ("Called once per pasteboard check, on the main thread"). The domain half of your method already reads `UserDefaults` directly on that queue; make the app half do the same and both halves are uniform:

```swift
let excludedApps = Defaults.sanitizedBundleIdentifierList(
    UserDefaults.standard.stringArray(forKey: DefaultsKey.selectionActionsExcludedApps) ?? [])
if let bundleID = front.bundleIdentifier, excludedApps.contains(bundleID) { return true }
```

That's `WindowPreviewProvider.captureIsPaused`'s existing answer to the same problem. `SelectionActionsExcludedApps.shared` stays for the settings UI, which is the only thing that needs the `@Published`.

**The excluded branch's `dismiss()` skips the generation check.** Every other terminal path lands in `handleRead`, which opens with `guard expected == generation`. This one doesn't:

```swift
guard isLocal || !(self?.isFrontmostAppExcluded() ?? true) else {
    DispatchQueue.main.async { [weak self] in self?.dismiss() }
```

Excluded browser in front, domain check walks AX for up to ~1s, and meanwhile you select text in TextEdit — the fast read presents the bar, then the slow excluded read's `dismiss()` lands and closes it. `guard expected == self?.generation` on the main hop.

Third thing, and I can't confirm this one: `summon()` passes `isLocal: false, targetPID: nil`. The mouse path grew a local monitor precisely because a `.nonactivatingPanel` never satisfies the frontmost/focused lookups. If that reasoning holds, the shortcut can't summon the bar in the Scratchpad, and `isFrontmostAppExcluded()` resolves against whatever app is behind it. `SelectionReader.focusedApplication()` is past the diff truncation, so I don't know whether it has its own escape hatch.

**Impact:** the race needs the settings page open during a selection; the stale dismiss needs a website exclusion set plus an app switch inside a second. Both are live on this PR, not 3/4.
**Merge:** hold for #1032, unchanged.

naveenkrdy:
All three addressed (84c602d):

- **Main-thread race on `@Published apps`.** Switched to reading `UserDefaults` directly with `Defaults.sanitizedBundleIdentifierList`, matching your suggested code and `WindowPreviewProvider.captureIsPaused`'s pattern exactly. `SelectionActionsExcludedApps.shared` stays for the settings UI, which is the only consumer that needs `@Published`.
- **Excluded-branch `dismiss()` skipping the generation check.** Added `guard expected == self?.generation else { return }` before the dismiss, same as every path through `handleRead`.
- **`summon()` in the Scratchpad - confirmed, and it's worse than the exclusion check alone.** `SelectionReader.read(pid: nil, ...)` falls back to the same `focusedApplication()` the exclusion check uses, so the keyboard shortcut couldn't have read a Scratchpad selection either - both halves were targeting whatever app was frontmost before the panel opened. Fixed generally rather than special-casing the Scratchpad: `summon()` now checks `NSApp.keyWindow != nil` instead of hardcoding `isLocal: false`. A `.nonactivatingPanel` can be key without the app becoming frontmost, which is exactly what `isLocal` already exists to detect for the mouse path - `NSApp.keyWindow` catches it without needing to know about any specific panel, and the bar's own panel (`SelectionActionBarPanel.canBecomeKey == false`) can never trigger a false positive against itself.

Also picked up the `SelfUninstall` question you raised on #1037 while I was in this file: `SelectionActionsService` was missing from `suspendInputInterceptors()` alongside `ScratchpadService`/`CommandBarService`. Added it here since this is the PR that introduces the service and its `AppDelegate` teardown call.

Rebased onto #1032's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8641 checks).

PathGao:
Two things.

**The suspend call landed in `applicationWillTerminate`, not `suspendInputInterceptors()`.** The hunk I can see is `AppDelegate.swift:238`, between `MouseNavigationService.shared.suspend()` and `DockPreviewService.shared.stop()` — that's the quit path. The `ScratchpadService`/`CommandBarService` pair you cited is `SelfUninstall.swift:93-94`, a different function in a different file, and `CommandBarService` isn't in the terminate list at all. `SelfUninstall.swift` is past my diff truncation, so if you touched it too, ignore this. If you didn't: the quit path is the one that doesn't matter here — the monitor dies with the process. `clearPermissions()` is the one that does, because it revokes Accessibility and **leaves the app running**, so the global mouse monitor and the ⌃⌥⌘A hotkey stay armed with the permission gone, while every other AX-backed service stops.

**`BrowserURLReader` re-copies AX helpers that already exist.** Its `copyElement` and `copyValue` are line-for-line `CommandBarSelectionReader`'s private versions in `Services/CommandBar/CommandBarSelection.swift:44-61`, `as!` included, behind the same `AXUIElementSetMessagingTimeout(element, 0.35)`. Counting `SelectionReader`, that's the third copy of those two functions and the third place `0.35` is written. One internal `enum AXCopy` with `element`/`string`/`value` and the timeout as a constant collapses all three — not a new abstraction, just the one that's already been written three times.

Related, and I can't check it: `SelectionReader` is past the truncation, but from the call sites it reads the frontmost app's selected text with a per-app messaging timeout, which is what `CommandBarSelectionReader.readSelectedText()` already does minus `isEditable`/bounds. If it's a superset, `CommandBarSelectionReader` becomes `SelectionReader.read(pid: nil)?.text` and goes away — unless the Command Bar's `maximumLength` cap and its "front app is us" guard are deliberately not wanted in the bar, in which case leave both.

Checked and clean: `QuickToolHotkey(id: 21)` is the one free slot in the hand-assigned 10–24 range, and ⌃⌥⌘A doesn't collide with any existing default in `GlobalShortcut.swift`.

**Impact:** none if `SelfUninstall.swift` also got the call. If it didn't, Clear all permissions leaves this feature's monitor and hotkey running after Accessibility is revoked.
**Merge:** hold for #1032, unchanged.

naveenkrdy:
**The suspend call - confirmed already covered, and you guessed right.** `SelfUninstall.swift` was the file, not `applicationWillTerminate` - `SelectionActionsService.shared.suspend()` is at `SelfUninstall.swift:95` inside `suspendInputInterceptors()`, added in the previous round (84c602d). Both `clearPermissions()` (`:21-36`) and `uninstallCompletely()` (`:40-52`) call `suspendInputInterceptors()` before revoking anything, so Clear All Permissions is covered the same way an actual uninstall is - the `AppDelegate.swift:238` call you found is the separate quit path, which is fine as-is per your own read (the process dies with it).

**AX helper triplication - fixed (23eaddd).** You found three; there was a fourth, `CommandBarMenus.swift`, past your truncation too. All four collapsed into `Core/AXCopy.swift` (`value`/`element`/`string`, `messagingTimeout` as the one `0.35` constant). `CommandBarMenus` keeps a thin local `copyString` wrapper since it's the one caller that trims - everyone else gets `AXCopy.string` exactly as Accessibility reports it, unchanged behavior everywhere else.

Left the `CommandBarSelectionReader` → `SelectionReader.read(pid: nil)` merge alone - you flagged it as unconfirmed yourself, and collapsing two independently-shipped AX read paths on a guess about behavioral equivalence (the `maximumLength` cap, the "front app is us" guard, `isEditable`/bounds) is a bigger and riskier change than this round's fixes. Happy to look properly as its own follow-up if you still want it pursued.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8641 checks).

PathGao:
One more member of that same set: `CleaningModeManager.activate()` doesn't suspend `SelectionActionsService`.

That list is every mouse-gesture-driven service — `MiddleClickService`, `MouseNavigationService`, `MouseButtonShortcutService`, `RadialMenuService` (`CleaningMode/CleaningModeManager.swift:66-76`), with the matching `syncWithPreferences()` restores at `:93-97`. Its tap masks keyDown/keyUp/flagsChanged/scrollWheel/gesture and deliberately not mouse — "pointer movement and clicks remain available for the Unlock button" — so a wipe across the trackpad is exactly the leftMouseDown/leftMouseUp pair `handleMouseUp` gates on, and 4 points is well under a wipe. Two lines:

```swift
SelectionActionsService.shared.suspend()               // activate(), beside RadialMenuService
SelectionActionsService.shared.syncWithPreferences()   // deactivate()
```

The overlays are `.nonactivatingPanel` at `CGShieldingWindowLevel()` with `acceptsFirstMouse`, so those clicks land in our process and it's the *local* monitor that fires: `isLocal: true`, `targetPID: getpid()`, exclusion checks skipped. What I can't confirm is the visible end of it — whether the bar draws above a shielding-level overlay, and whether our own process has a selection to find at that moment. What's checkable is that the read fires on every wipe stroke, and that this is the third and last suspend list in the tree (`SelfUninstall.swift:63`, `AppDelegate.swift:218`, `CleaningModeManager.swift:54`); the other two got the call.

On `CommandBarSelectionReader` → `SelectionReader.read(pid: nil)`: agreed, drop it. I raised it off the call sites without reading `SelectionReader`, and the follow-up would cost more to review than the duplication costs to keep.

**Impact:** during Cleaning Mode every wipe stroke schedules an AX selection read against our own process. No permission or data consequence; whether anything visible surfaces I haven't confirmed.
**Merge:** hold for #1032, unchanged — the two lines fold into the rebase.

naveenkrdy:
Added, matching your code exactly (db4e1c0):

```swift
SelectionActionsService.shared.suspend()               // activate(), beside RadialMenuService
SelectionActionsService.shared.syncWithPreferences()   // deactivate(), beside RadialMenuService
```

That's the third suspend list closed out - `SelfUninstall.swift`, `AppDelegate.swift`, and now `CleaningModeManager.swift` all suspend Selection Actions the same way they suspend everything else that reacts to input while it's supposed to be paused.

Also noted: dropping the `CommandBarSelectionReader` merge idea. Agreed that's the right call.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8641 checks).

PathGao:
Sending the stack back for one change at the base; #1037 and #1038 inherit it.

All three branches still ship `Sources/Vorssaint/Services/SyntheticPasteSupport.swift` (161 lines: `waitForCleanModifiers`, `postCmdVAssumingCleanModifiers`, `snapshot(of:)`). Main has had `Services/TransientPaste.swift` since #1032 was closed on 2026-09-02 for duplicating it: `postPasteWhenModifiersReleased`, `snapshot(of:)`, the change-count guard and the restore. The `PastePlainService.swift` hunk here replaces main's `TransientPaste.shared.paste(` call and is what makes all three branches conflict.

Delete `SyntheticPasteSupport`, call `TransientPaste` from `replaceSelection` and `ShadowCopySelectionReader`, drop the PastePlainService hunk, and rebase. The floating menu itself has answered every review round, and the description has a real Mac on it; this is the last item before the stack is reviewable as #1033 → #1037 → #1038.

vorssaint:
closing this. the bar as it stands only repeats what the system's own copy, cut, paste and delete already do, and the base underneath needs a rewrite rather than fixes
the fallback fires a real copy keystroke in the front app whenever accessibility has no selection to hand back, and clipboard history can record that text before the clipboard is put back, so text someone only highlighted ends up stored as if they had copied it
copy, cut and paste from the bar handle only trimmed plain text, so they drop formatting and surrounding spaces that the system's own commands keep
the paste helper here duplicates the one the app already has and this branch swaps the existing call for its own, which is what makes the whole stack conflict and was asked about on the second of this month with no answer since
the shortcut slot it takes is now used by recent captures, so with both switched on one key would fire the other's action
the settings page shows eight main controls where the app keeps five and tucks the rest under more options, the visible text names other browsers and uses a dash the app's text avoids, and the code comments name other products
inside the app's own text views the bar never closes on typing or clicking, because it only watches events from other apps
a smaller version that reads selections only through accessibility, reuses the paste helper the app already has, keeps the settings to a handful of controls and offers something the system menu doesn't already could still be considered, and the shared accessibility read helper would be welcome as its own small change

PR #1034, opened 2026-08-26
fix(window-layout): keep window size when moving between displays

DiogoDuart3:
## Summary

Next/previous display scaled a window as a fraction of the destination screen, so a window moved from the built-in display onto an ultrawide grew in both axes and the empty space around it was stretched the same way.

`rectForDisplay` now keeps the current point size, shrinking only to fit. Round trips are lossy: a window that has to shrink to fit a smaller display does not grow back to its original size (the ratio mapping used to restore it). If the shrink fills an axis, the return trip treats that edge as flush, so the origin moves too — the shrink test leaves at y 200 and comes back at y 500. If both axes were clamped, the return trip treats that as Maximize.

Position keeps physical edge insets instead of interpolating leftover space: a window flush with the right or bottom stays there (so a right-half stays a right-half), everything else keeps its left and top insets so it does not fly to the middle of an ultrawide, and a full-height window stays under the menu bar instead of dropping onto the dock of a taller display. A window within a few points of filling one axis keeps that preferred-edge inset instead of treating leftover space as a flush edge. A window that already fills the source within the settle tolerance still fills the destination. That tolerance is the same 8pt the settle path uses when judging a placement as stuck. Dock Preview drag, the move gesture, and native title-bar drags were already size-preserving; only this shared helper resized.

## Verification

MacBook Pro 16" (Mac16,5, M4 Max), macOS 27.0 (26A5416b). Built-in Liquid Retina XDR plus a 3840×1080 ultrawide, Spaces not mirrored. Installed a local `./build.sh --install` release build and confirmed Next/Previous display keeps the window size on the ultrawide. Position inset behaviour, near-full-width placement, and shrink-to-fit round trips are covered by unit tests; needs a rebuild to feel on the Mac.

```sh
./build.sh --test
```

8649 checks ran. The only failure is the existing dispatch-pool starvation probe, which is unrelated to this geometry change.

Could not re-test a portrait external display on hardware.

## Anything else

No existing issue. Reported from using Next/Previous display on a laptop plus ultrawide setup.

### DIFF
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
@@ -71,7 +71,7 @@ final class WindowLayoutService: ObservableObject {
     /// Read on every pointer event, so it is resolved once instead of per
     /// click.
     private static let ownProcessID = Int64(getpid())
-    private let frameTolerance: CGFloat = 8
+    private let frameTolerance = WindowLayoutGeometry.frameTolerance
     private let anchorTolerance: CGFloat = 36
     private let moveGestureUpdateInterval: TimeInterval = 1.0 / 120.0
     // AX frame mutations are not atomic. Complex windows can visibly render
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
@@ -531,19 +531,56 @@ enum WindowLayoutGeometry {
               destinationVisibleFrame.height > 0
         else { return current.integral }
 
-        let widthRatio = min(max(current.width / sourceVisibleFrame.width, 0.08), 1)
-        let heightRatio = min(max(current.height / sourceVisibleFrame.height, 0.08), 1)
-        let xRatio = (current.minX - sourceVisibleFrame.minX) / sourceVisibleFrame.width
-        let yRatio = (current.minY - sourceVisibleFrame.minY) / sourceVisibleFrame.height
+        // Scaling by each screen's size stretches a laptop window across an
+        // ultrawide and squashes it onto a portrait panel. Keep the current
+        // size, only shrinking to fit, and keep the same alignment in the
+        // leftover space. A window that already fills the source still fills
+        // the destination, so Maximize stays Maximize. 8pt matches the
+        // settle tolerance used when judging a placement as stuck.
+        let fillTolerance: CGFloat = 8
+        if current.width >= sourceVisibleFrame.width - fillTolerance,
+           current.height >= sourceVisibleFrame.height - fillTolerance {
+            return destinationVisibleFrame.integral
+        }
 
-        let width = min(destinationVisibleFrame.width, max(1, destinationVisibleFrame.width * widthRatio))
-        let height = min(destinationVisibleFrame.height, max(1, destinationVisibleFrame.height * heightRatio))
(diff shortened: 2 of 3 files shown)


PathGao:
Two things.

**Near-zero slack turns horizontal placement into a coin flip.** `alignedOrigin` guards `sourceSlack <= 0` but not slack just above it. A window that is full-width-minus-a-few-points on the built-in display — terminals snapping to character cells sit here constantly — divides a 1-8pt inset by a 1-8pt slack, so `t` sweeps the whole [0, 1] range. On the 3840 ultrawide that is a ~2400pt swing from a 2pt difference in where the window happened to be:

```
source 1440 wide, window 1436 wide
  minX = 2  ->  t = 0.5  ->  centered
  minX = 4  ->  t = 1.0  ->  hard right
```

The old code divided by `sourceSpan`, so it was continuous. This is a regression for that window shape, not a pre-existing edge. The fill branch does not cover it either, since it needs both axes near-full and this window is only near-full on one. Cheapest fix reuses the tolerance already there:

```swift
if sourceSlack <= fillTolerance { return destinationMin + (currentMin - sourceMin) }
```

Keeps the edge inset, continuous with the `slack <= 0` case, and the caller already clamps. Needs `fillTolerance` hoisted out of `rectForDisplay`.

**The 8pt is a second copy.** The comment says it matches the settle tolerance used when judging a placement as stuck, but it is a fresh literal with nothing linking the two, so whoever tunes the settle tolerance next splits them silently. Reference the existing constant. The file defining it is not in the diff, so I have not checked its name or whether it is reachable from here.

One more for the body: round trips are now lossy. Laptop -> smaller external shrinks the window to fit, and coming back leaves it shrunk, where the ratio version grew it back. That is the case you said you could not test. I think it is the right trade, but it belongs in the body rather than being discovered.

**Impact:** windows within a few points of filling the source width land at an arbitrary horizontal position on the destination. Exactly-full and clearly-smaller windows are unaffected.
**Merge:** needs the near-zero-slack guard first.

DiogoDuart3:
The leftover-space interpolation that made near-full-width placement a coin flip is gone; `unscaledOrigin` keeps insets instead of dividing by slack. Flush-edge detection still had the same jump for a window a few points short of filling one axis (1pt of leftover picking left vs right), so display transfer now skips that snap when leftover space is within the settle tolerance and keeps the preferred-edge inset. That 8pt lives on `WindowLayoutGeometry.frameTolerance` and is the same value settle uses.

Round trips being lossy is now in the body. A window that shrinks to fit does not grow back; if both axes were clamped, the return trip treats that as Maximize.

PathGao:
Guard checks out. Walked the arithmetic against the new tests: `nearFullWidth` at slack 4 and `nearFullWidthRightish` at slack exactly 8 both keep their left inset, so the [0, 1] sweep is gone. `WindowLayoutGeometry.frameTolerance` is the same constant `WindowLayoutService` settles against — `isClose` at both the verify and the retry site — so tuning one can't split them now.

Grepped `main` for the ratio shape too. `rectForDisplay` was the only place it existed and the next/previous display branch is its only production caller, so "only this shared helper resized" holds.

One line for the body, not a blocker: round trips lose position as well as size. Your own shrink test encodes it — shrinking to fit makes the window full-height on the short display, so the return trip reads it as flush with the top and it comes back at y 500 instead of the 200 it left from. Worth naming next to the height sentence, since it is the same surprise.

I have not run this on hardware.

**Impact:** none outstanding.
**Merge:** ship it.

Issue #617, opened 2026-08-12
Quick PDF tools alongside the media tools (compress, split, merge, rotate)


clovis-dugue:
### Problem or Motivation

Some administrative websites reject PDF based on criterias like size, page number, single PDF file etc. Fixing this today means either Preview gymnastics (the "Reduce File Size" filter is not great), or one of those pdf-tools websites I'd rather not upload documents to, or spinning up a self-hosted toolbox like BentoPDF, which is a lot of machinery for a 30-second task.
 
Vorssaint's Media tab already solves this class of problem for video, GIFs and images: drop a file, tweak one or two settings, get the file the form wants. PDFs feel like the missing fourth member. The image converter even offers PDF as an output format already.
 
To be clear about scope: I'm not asking for a PDF suite (no editing, annotation, signing, OCR layers). Just the handful of mechanical operations upload forms keep demanding.

### Proposed Solution

A PDF tool in the Media workspace, same interaction model as the existing
ones:
 
1. **Compress**: quality slider, before/after file size shown, like the image compressor. Downsampling embedded images is where most of the size usually is.
2. **Split / page range**: keep or remove a range ("1-3", "2,5-7"), producing a new file. Covers "first N pages only" uploads, stripping a page, and splitting one document into parts.
3. **Merge**: combine several PDFs into one, in a chosen order. Accepting images in the mix would be a nice touch (ID card scanned as two JPEGs, output one PDF), and the image tool already knows how to make a PDF page from an image.
4. **Rotate pages**: whole document or a range, for the classic upside-down/sideways scan.

Two smaller ones that fit the same shape, maybe for future feature requests
 
- **Strip metadata**, matching the option the image compressor already has.
- **Remove a known password** (open with password, save without), for bank statements and payslips that arrive locked and that upload portals refuse.

All of this stays inside the codebase's no-dependency approach: PDFKit's `PDFDocument` handles page extraction, reordering, merging, per-page rotation, password handling and document attributes in a few lines each, and compression can re-encode embedded images through ImageIO (the same `CGImageDestination` machinery the image compressor already uses) or a Quartz filter. All system frameworks, no ffmpeg-style external binary, consistent with how the video tool uses the system's own `avconvert`.

### Feature Area

Tools

### Mockups / References

- The existing Media tools (`Sources/Vorssaint/Services/Media/`, `MediaTool` enum) as the interaction model to copy.
[...]

PR #1035, opened 2026-08-26
feat(screenshot): open an image file in the capture editor

### LINKED ISSUE #677: Add a drag and drop thing to screenshot editor
### Problem or Motivation

The image field requires me to get the image from finder but vorssiant does not have a option to drag and drop the screenshot.

### Proposed Solution

The screenshot editor had a option to drag and drop the image, like the image gets cached and then when vorssaint copies the screenshot it copies the path instead of the actual image. Related #634 

### Feature Area

Clipboard and files

### Mockups / References

_No response_

FlowSync0:
Does **not** close #677 — see the thread below. This adds a way *into* the
capture editor from the media tools; #677 asks for a drag source *out* of it.

## Before

An image already on disk cannot reach the capture editor. #677 puts it
plainly: the editor has to be fed from Finder, and there is no way to hand it
a file. The media tools have an Edit button for video, which imports a dropped
file and opens the recorder editor on it; the image tool has no equivalent, so
an image that needs a crop or an arrow before it is compressed has to leave the
app and come back.

## After

**An Edit button in the media tools' image tab**, beside the one the video tab
has carried since the media tools shipped. An image loaded there opens in the
capture editor.

## How it works

The capture editor works on an image, not on a file it has to own. That is the
whole reason this is small: the recorder editor works on takes it owns, which
is why the video path has to import the file into a private store first, while
an image only has to be read.

The pasteboard path was already doing exactly that work, down to reading a file
URL with `NSImage(contentsOf:)`, so both entry points now share one capture
builder instead of each carrying its own. The size ceiling that guarded it
moves to `ScreenshotSupport.editorAcceptsImage(pixelSize:)`, where a file —
which can be far larger than anything a screen produces — is refused before a
window opens on it rather than after.

The button is absent rather than inert when the screenshot feature is turned
off, since the editor belongs to that feature.

## Notes

No new user-facing text: the Edit label already exists in every supported
language, since it is the recorder editor's own.

## Checks

- `./build.sh --test` passes, with new assertions for the size ceiling: an
  image at the ceiling opens, one past it, one empty, and one with an
  unmeasurable dimension do not.
- Exercised by hand on JPEG, PNG and HEIC files, dropped on the editor and
  opened from the media tools.

## On #677, and on the body this PR opened with

Two corrections, both mine.

The body claimed a drop-on-the-editor entry point. That code is not in this
commit — it lives on a working branch and never made it here. @PathGao caught
it before anyone could read it as shipped.

And `Closes #677` was wrong. The issue reads as "there is no way to get an
image into the editor", which is what this was built for, but @Yahddyyp has
clarified that the ask is the opposite direction: a drag source in the editor,
near Save and Copy, to pull the picture out into Finder.

That handle already exists — `dragOutHandle`, the thumbnail in the dimensions
chip at the bottom left — so #677 is about where it sits rather than about
building it, and that belongs in its own change. #677 stays open.

## Since the review

- An image past the editor's ceiling now says so, instead of reporting
  `mediaErrorUnsupported`. That is the likelier of the two failures in the
  image compressor, and it was the wrong message.
- `openVideoEditor` has its `@MainActor` back: the new function was inserted
  directly under the attribute and consumed it.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
@@ -480,16 +480,33 @@ final class ScreenshotService: ObservableObject {
         }
     }
 
+    /// Opens the editor on an image file instead of on a capture, so the media
+    /// tools can hand it a dropped file the way they already hand a dropped
+    /// video to the recorder editor. Returns false when the file is not an
+    /// image the editor can hold, leaving the caller to say so.
+    @discardableResult
+    func openEditor(imageAt url: URL) -> Bool {
+        guard AppFeature.screenshot.isAvailable,
+              let image = NSImage(contentsOf: url),
+              let capture = Self.capture(from: image)
+        else { return false }
+        openEditor(with: capture)
+        return true
+    }
+
     private static func clipboardCapture(
         from pasteboard: NSPasteboard
     ) -> ScreenshotSelectionController.Capture? {
-        guard let image = clipboardImage(from: pasteboard),
-              image.size.width > 0, image.size.height > 0
-        else { return nil }
+        guard let image = clipboardImage(from: pasteboard) else { return nil }
+        return capture(from: image)
+    }
+
+    private static func capture(from image: NSImage) -> ScreenshotSelectionController.Capture? {
+        guard image.size.width > 0, image.size.height > 0 else { return nil }
         var rect = CGRect(origin: .zero, size: image.size)
         guard let cgImage = image.cgImage(forProposedRect: &rect, context: nil, hints: nil),
-              cgImage.width > 0, cgImage.height > 0,
-              cgImage.width <= ScreenshotSupport.scrollingCaptureMaximumPixels / cgImage.height
+              ScreenshotSupport.editorAcceptsImage(
+                  pixelSize: CGSize(width: cgImage.width, height: cgImage.height))
         else { return nil }
         let pixelSize = CGSize(width: cgImage.width, height: cgImage.height)
         let scale = ScreenshotSupport.clipboardImageScale(pixelSize: pixelSize,
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift
@@ -560,6 +560,16 @@ enum ScreenshotSupport {
         return CGFloat(scale)
     }
 
+    /// An image that arrives from outside a capture can be far larger than
+    /// anything the screen produces, so it is refused before a window opens on
+    /// it rather than after the editor has tried to hold it.
+    static func editorAcceptsImage(pixelSize: CGSize) -> Bool {
+        guard pixelSize.width.isFinite, pixelSize.height.isFinite,
+              pixelSize.width >= 1, pixelSize.height >= 1
+        else { return false }
+        return pixelSize.width * pixelSize.height <= CGFloat(scrollingCaptureMaximumPixels)
+    }
+
     /// Uses the logical size carried by a copied image when it describes one
     /// consistent display scale. Imported files without that metadata edit at 1x.
     static func clipboardImageScale(pixelSize: CGSize, pointSize: CGSize) -> CGFloat {
--- a/Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
+++ b/Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
@@ -403,6 +403,17 @@ struct MediaWorkspaceView: View {
                 }
             }
 
+            // The editor belongs to the screenshot feature, so the button is
+            // absent rather than inert when that feature is turned off.
+            if selectedTool == .imageCompressor, AppFeature.screenshot.isAvailable {
+                Button {
+                    openImageEditor()
+                } label: {
+                    Label(screenshotText.editButton, systemImage: "crop")
+                }
+                .disabled(inputURLs.count != 1 || isRunning)
+            }
+
             if isRunning {
                 Button {
                     media.cancel()
@@ -1310,6 +1321,19 @@ struct MediaWorkspaceView: View {
     }
 
     @MainActor
+    /// The dropped image goes straight to the capture editor: nothing is
+    /// imported or copied first, because that editor works on an image rather
+    /// than on a file it has to own.
+    private func openImageEditor() {
+        guard AppFeature.mediaTools.isAvailable, AppFeature.screenshot.isAvailable,
+              let url = inputURL, inputURLs.count == 1
+        else { return }
+        localMessage = nil
+        if !ScreenshotService.shared.openEditor(imageAt: url) {
+            localMessage = l10n.s.mediaErrorUnsupported
+        }
+    }
+
     private func openVideoEditor() {
         guard AppFeature.mediaTools.isAvailable, let url = inputURL,
               !isImportingVideo else { return }
--- a/Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
+++ b/Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
@@ -1320,7 +1320,6 @@ struct MediaWorkspaceView: View {
         return (duration * 10).rounded() / 10
     }
 
-    @MainActor
     /// The dropped image goes straight to the capture editor: nothing is
     /// imported or copied first, because that editor works on an image rather
     /// than on a file it has to own.
@@ -1329,11 +1328,18 @@ struct MediaWorkspaceView: View {
               let url = inputURL, inputURLs.count == 1
         else { return }
         localMessage = nil
-        if !ScreenshotService.shared.openEditor(imageAt: url) {
+        // An image the editor cannot hold is not an unsupported format: say
+        // which of the two it is, since the tool's own limit is the likelier
+        // of the two here.
+        if let size = MediaSupport.imageDisplaySize(at: url),
+           !ScreenshotSupport.editorAcceptsImage(pixelSize: size) {
+            localMessage = imageText.tooLarge
+        } else if !ScreenshotService.shared.openEditor(imageAt: url) {
             localMessage = l10n.s.mediaErrorUnsupported
         }
     }
 
+    @MainActor
     private func openVideoEditor() {
         guard AppFeature.mediaTools.isAvailable, let url = inputURL,
               !isImportingVideo else { return }
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -13833,6 +13833,15 @@ struct MetricsTests {
             pixelSize: CGSize(width: 1200, height: 800),
             pointSize: CGSize(width: 600, height: 800)) == 1,
                "copied images with inconsistent metadata safely use 1x")
+        expect(ScreenshotSupport.editorAcceptsImage(pixelSize: CGSize(width: 4_000, height: 3_000))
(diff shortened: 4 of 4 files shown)


PathGao:
The drop path isn't in this diff. `openEditor(imageAt:)` has exactly one caller — `openImageEditor()` in `MediaWorkspaceView` — and nothing in these four files registers a drop target on the editor window. The body describes that entry point as "what #677 asks for literally", and it's the one that isn't here. Either it's in a commit I wasn't shown, or this ships one door. Worth pinning down because `Closes #677` auto-closes on merge, so the half that isn't the literal ask closes with it.

The size ceiling reports as a format error:

```swift
if !ScreenshotService.shared.openEditor(imageAt: url) {
    localMessage = l10n.s.mediaErrorUnsupported
}
```

Three distinct failures collapse into one message. The reachable one is the ceiling — a 100 MP photo sitting in the *image compressor* is the normal case for that tool, and the user gets told the format is unsupported. Either call `ScreenshotSupport.editorAcceptsImage` in the caller for its own message, or return the reason instead of a Bool.

Smaller: the `@MainActor` on the context line belonged to `openVideoEditor`, and inserting above it moved the attribute onto the new function. It builds, so `View`'s isolation is covering both — but put the `///` above the attribute so the next insert doesn't do this again.

Not verified: the button reads `screenshotText.editButton` while the body says the label is the recorder editor's own. No localization file is in the diff, so I can't tell whether that accessor already exists. If it's new, "no new user-facing text" doesn't hold.

Also, `Point out a reference that will not reach the reporter` is red.

**Impact:** anyone dropping a file on the editor window gets nothing, and #677 closes on merge as answered. Separately, an oversized image is reported as an unsupported format.
**Merge:** needs the drop-on-editor entry point first — or a body that stops claiming it.

ref #677

Yahddyyp:
This does not resolve #677. The issue refers to a icon or item in the ScreenshotEditorView optimally near the save and copy button which allow you to drag and drop the image into finder so that services that need finder to open like this in shottr if you used it 
image

PathGao:
Yahddyyp is right and my read was wrong. #677 wants a drag *source* in the editor — an item next to save and copy that you drag out into Finder. This PR moves files the other direction. Neither entry point here touches it, so the drop-on-editor door I asked about wouldn't have answered #677 either.

That makes `Closes #677` the thing to fix first. The Edit button in the image tab is fine on its own merit, but it answers a different request, and #677 has to stay open.

The size ceiling is unchanged and independent of all that: `openEditor(imageAt:)` returns a Bool, so an oversized photo in the *image compressor* — the normal case for that tool — gets reported as `mediaErrorUnsupported`. Return the reason, or check `ScreenshotSupport.editorAcceptsImage` in the caller.

**Impact:** #677 closes on merge without being addressed, and an oversized image is reported as an unsupported format.
**Merge:** needs `Closes #677` dropped first.

ref #677

PathGao:
`Closes #677` is gone and the `@MainActor` is back. `screenshotText.editButton` does exist — `MediaWorkspaceView.swift:397`, the video tab's own button — so "no new user-facing text" holds for the label. My earlier unverified note closes.

The new too-large message is borrowed from a different ceiling. `imageText.tooLarge` reads *"These dimensions are too large to process safely. Choose a smaller size."* and today it has exactly one source: `MediaFailure.imageTooLarge`, which is `MediaSupport.imageRenderSizeIsSafe` — 8192×8192 = 67,108,864 px. The editor's ceiling is `scrollingCaptureMaximumPixels` = 60,000,000. Between the two the compressor handles the file fine and the editor refuses it. Load a 9000×7000 photo in the image compressor: Compress works, Edit says the dimensions are too large to process safely, and "choose a smaller size" points at a resize field `openImageEditor` never reads. The message that belongs here is the editor's, not the compressor's — which means one new string, and the Notes line has to stop saying there is none.

The Checks list still says *"dropped on the editor and opened from the media tools"*. Two paragraphs further down the body says the drop code is not in this commit. One of those is wrong and it is the check.

One more: don't take the workflow hint literally. It asks for `Refs #677`, but it was generated against `Closes #677`, and the reference form still gets the issue labelled fixed-pending-release on merge. #677 is not fixed by this, so a `Refs` line puts Yahddyyp on a two-week clock to confirm a fix that shipped as something else. The prose mention you have now is the right form — leave it.

**Impact:** an image between 60 and 67 MP in the image compressor is told its dimensions cannot be processed safely, by the tool that is about to process them.
**Merge:** needs the too-large message to name the editor's limit instead of the compressor's.

FlowSync0:
@Yahddyyp — thank you for the correction, and sorry for the noise on your issue. I read the body as "there is no way to get an image into the editor" and built the opposite direction of what you asked for.

`Closes #677` is off this PR and the body now says why. #677 stays open.

One thing that may be useful: the drag source already exists — it is the small thumbnail in the dimensions chip at the bottom left of the editor, and it does drag out to Finder. So what your issue asks for is mostly a matter of where it sits: an item near Save and Copy, where you would look for it, rather than tucked into the corner. That is a much smaller change than building it, and it is worth doing on its own.

PathGao:
The new ceiling caps area and nothing else. `editorAcceptsImage` is `width * height <= scrollingCaptureMaximumPixels`, so a 50,000 × 1,000 panorama — 50 MP — passes the pre-check in `openImageEditor`, passes again in `capture(from:)`, and the editor opens on an image 50,000 px wide. The compressor one tab over refuses that shape: `MediaSupport.imageRenderSizeIsSafe` caps each side at `maxImageRenderDimension = 20_000` (`MediaSupport.swift:365`) *and* the area.

This is the body's own argument pointed back at it. "A file can be far larger than anything a screen produces" is exactly why area alone isn't the guard — a screen cannot be 50,000 px on one side, a stitched panorama or a scanned map can. The clipboard path carried the same product-only check and never hit it, because its source was a screen. This entry point is what makes it reachable. The new assertions cover 10,000 × 6,000 and 10,000 × 6,001, both square-ish; nothing covers long and thin.

I have not put a 50,000 px image through the editor, so I can't say where it gives out — CGContext, the texture limit, or memory. Only that nothing here stops it.

The too-large message is unchanged since the last round: 60–67 MP compresses fine, and Edit reports the compressor's limit for it.

@Yahddyyp — on the handle: it is there, and it does the thing you asked for. The small picture beside the `1234 × 567 px` label in the bottom-left corner of the editor writes a real PNG and hands Finder the file, so anything that needs a file on disk gets one. Two things though. I can only see the current development code, so I can't tell whether that handle is in the version installed on your Mac — if there's no thumbnail next to the dimensions in your editor, it isn't, and you'd be waiting on a release. And nothing about it says it can be dragged; hovering it just shows the editor's name. Moving it up beside Save and Copy is the right call, but it needs a label too, or the next person misses it exactly the way you did.

**Impact:** an image under 60 MP but longer than ~20,000 px on one side opens in the editor unguarded, and the 60–67 MP case still gets the compressor's message.
**Merge:** needs a per-side cap on `editorAcceptsImage` and the too-large message pointing at the editor's limit.

Yahddyyp:
> @Yahddyyp — on the handle: it is there, and it does the thing you asked for. The small picture beside the `1234 × 567 px` label in the bottom-left corner of the editor writes a real PNG and hands Finder the file, so anything that needs a file on disk gets one. Two things though. I can only see the current development code, so I can't tell whether that handle is in the version installed on your Mac — if there's no thumbnail next to the dimensions in your editor, it isn't, and you'd be waiting on a release. And nothing about it says it can be dragged; hovering it just shows the editor's name. Moving it up beside Save and Copy is the right call, but it needs a label too, or the next person misses it exactly the way you did.

Holy molly that is not easy to find if you don't know about it. I will make sure to add a label to it or just a `.help` method for it.

PathGao:
@Yahddyyp — the handle already carries one: `.screenshotSafeHelp(strings.editorTitle)` at `ScreenshotEditorView.swift:1198`. So it is a string to change, not a modifier to add; a second `.help` stacked on it won't show.

Worth knowing what that wrapper does before leaning on it. On macOS 27 it returns `self` and no tooltip appears at all, because SwiftUI's tooltip bridge crashes there while routing hover events. A tooltip-only fix is invisible on that system. The handle also has no `.accessibilityLabel` — the wrapper's own comment says to keep those everywhere, but this call site doesn't, so VoiceOver reads nothing there either.

The visible label, your first option, is the one that works on every system. If you want the tooltip too, point `screenshotSafeHelp` at a new string that says it drags out instead of the editor's title, and add the matching `.accessibilityLabel`.

Yahddyyp:
@PathGao What would you like me to implement should i replace the thumbnail with the drag and drop icon or would you just like me to put it next to save and copy. I am finding difficuty implimenting the icon in the `topBand` introduces the bug of window moving when you try to use the drag and drop item as `topBand` overlaps with the title bar causing the window to move when you try to use the item's top half. You have also commented on #1050 that moving it to the `infoChip` replacing the thumbnail with a SF symbol instead and adding a label.

PathGao:
infoChip. `topBand` isn't a placement you can work around: `isMovableByWindowBackground` is false and the comment at `ScreenshotEditorController.swift:995` says why — "Content drags edit annotations. Only the title strip moves the window." `topBand` is `.padding(.top, 10)` on a 38pt frame, so its top half *is* the 28pt title strip, and that strip is the only way to move this window. A drag source there either loses to the window drag or takes window moving away. The window move isn't a side effect to fix, it's the one region that still does it.

So replace the thumbnail and keep the position. The thumbnail draws `model.baseImage` while `.onDrag` exports `model.exportImage()` — it never shows the annotations you are dragging out, so dropping it loses nothing.

```swift
private var dragOutHandle: some View {
    Label(strings.dragOutButton, systemImage: "arrow.up.doc")
        .font(.system(size: 11, weight: .medium))
        .onDrag { /* unchanged */ }
        .accessibilityLabel(strings.dragOutButton)
        .screenshotSafeHelp(strings.dragOutButton)
}
```

Not `square.and.arrow.up` — the cluster already has a share control and that symbol reads as sharing.

Two things that ride along: `dragOutButton` is a new string in every language, and the `.accessibilityLabel` isn't optional here. `screenshotSafeHelp` returns `self` on macOS 27 (`ScreenshotEditorView.swift:1333`), so on that system the visible text is the entire affordance and VoiceOver gets nothing without the label.

This lands as its own change, not here — #1035 moves images the other direction.

PathGao:
Closing the three branches of this stack (#1035, #1036, #1039) as stalled; the reasons for each are here so they are in one place.

**#1035** (base): the 2026-08-27 08:41 review asked for a per-side cap in `editorAcceptsImage` and a too-large message that names the editor's own limit. The diff still checks area only (`width * height <= scrollingCaptureMaximumPixels`) and reuses the compressor's message.

**#1036**: the 08-27 08:43 review asked for the `.onDrop(of: [.fileURL])` handler to be accounted for — it is in the diff, in no description, and fails silently on a non-image drop (`guard let url else { return }`). Separately, `restore(_:)` ends in `else { pixelated = [:] }`, discarding a cache that depends only on `baseImage`, so every undo or redo that leaves no pixelate box rebuilds the full-resolution mosaic on the main thread.

**#1039**: the 08-27 08:47 review asked for three things; `inputURLs = ... Array(urls.prefix(1))` still drops the PDF from a mixed selection, and the multi-page export spawns `/usr/bin/zip` through a file-local `Process` plus `while process.isRunning { Thread.sleep }`, not `BoundedProcessRunner`, which would put a second archive mechanism next to #1080's.

All three: last push 2026-08-27, no reply since, six days; no Mac model or macOS build in any description. The direction is fine (#677, #617) and the base PR is small. Keep the branches; if you come back, land #1035 first with the cap, and the other two rebase on it.

PR #1036, opened 2026-08-26
feat(screenshot): give the annotation tools the controls they were missing

FlowSync0:
Based on #1035, which this branch carries as its first commit. Reviewing the
second commit alone is enough; the diff cleans itself up once #1035 lands.

## Before

Several of the editor's tools carry a setting that changes nothing, lack the
one they need, or answer a gesture only until the mark is placed.

- Undo and redo have no keyboard shortcut. The window's key monitor claims ⌘Z,
  but it steps aside whenever a text field holds the responder chain, and
  nothing answers in that state.
- The inline text field takes its width from the canvas rather than from what
  is typed, so it stretches to the far edge of the window.
- The selection box is drawn under that field, framing one thing twice.
- A number left mid-edit needs Return to be kept.
- Shift and Option already constrain the capture overlay's own selection. The
  editor's shapes ignore them, both while being drawn and while being resized.
- Highlight, pixelate, redact and counter all show a stroke width picker. None
  of them reads it: they fill, cover, or size themselves from the image.
- The pixelate mark has one coarseness, decided for you.
- An arrow has one head, at the tip, always.
- A text has three sizes, borrowed from that same stroke picker, and one face.
- A counter cannot be resized, and its number cannot be typed.

## After

The fixes are the first four above: ⌘Z and ⇧⌘Z on the buttons, so the window
behaves like any other Mac window whichever path the key takes; a field that
sizes itself to its text; no selection box under a field that already draws its
own border; and a number kept by anything that takes the pointer elsewhere.

The gestures come next: Shift and Option now constrain a shape as it is drawn
**and** as it is resized, the way they already do when a capture region is
selected.

Then the style bar becomes contextual. A width picker that changes nothing is
worse than no picker, so the four tools that ignore it stop showing it, and in
its place each tool gets what it actually reads:

- **Pixelate** gains a strength, stored per mark. One pixelated twin is cached
  per strength in use, and the ones nobody uses are released.
- **An arrow** gains its heads: at the tip, at both ends, or neither.
- **A text** gains the faces installed on the Mac and its own point size,
  dragged from its corners. Only the corners: its glyphs decide its width, so
  an edge handle would stretch a box that does not exist.
- **A counter** gains those same corners, and a number typed by hand. A number
  set by hand freezes out of the automatic renumbering instead of being
  overwritten by the next badge.

Two smaller ones. A filled ellipse joins the palette as the counterpart of the
redaction rectangle, placed past the numbered nine so no keyboard shortcut
moves. And the tool rail answers the pointer immediately with a label of its
own: the system tooltip waits a second, says only the name, and
`screenshotSafeHelp` turns it off entirely on macOS 27 and later, so that rail
has no labels at all there today.

## Notes

- One new string, the filled ellipse's name, provided in every supported
  language.
- Marks made before this read the same: an unset point size falls back to the
  stroke picker's, an unset face to the system font, and an unset strength to
  the block size the editor has always used.

## Checks

- `./build.sh --test` passes, with assertions for the tools that offer a stroke
  width, the squared resize and its anchoring, the blur block at each end of
  the strength range, and the palette order that keeps the numbered nine.
- Exercised by hand on a real capture: every tool drawn, resized, undone and
  redone, with a text and a counter retyped after being resized.


PathGao:
Three in the annotation work, one that isn't in the body at all.

**1. The pixelate tool now shows a color picker it does not read.** `showsColorControls` gained `case .pixelate: return true`, and the `.select` branch dropped `&& selected.tool != .pixelate`. `drawPixelate` reads `annotation.rect` and the cached twin, nothing else — the color is never used. The body's own argument is that "a width picker that changes nothing is worse than no picker", and the same change adds a color picker that changes nothing. Either both `.pixelate` exclusions come back, or the strength row takes the color row's place for that tool instead of joining it.

**2. Resizing a counter for the first time jumps.** `beginDrag` sets, for every tool:

```swift
moveFontSize = selected.fontSize > 0
    ? selected.fontSize
    : Double(ScreenshotRenderer.fontSize(for: selected.stroke, scale: 1))
```

For a counter that has never been resized `fontSize` is 0, so the base becomes the stroke picker's point size — but the badge on screen is `counterDiameter(for: imageSize, scale: 1)`. `updateDrag`'s counter branch carries the correct fallback and it is dead code, because `moveFontSize` is never 0 by the time it runs. Grab a corner of a fresh badge and it snaps to the stroke size before it starts scaling. The fix belongs in `beginDrag`:

```swift
moveFontSize = selected.tool == .counter
    ? Double(counterBounds(selected).height)
    : (selected.fontSize > 0 ? selected.fontSize
         : Double(ScreenshotRenderer.fontSize(for: selected.stroke, scale: 1)))
```

**3. The strength slider is the one new control that cannot be undone.** `applyTextFont` and `applyArrowHeads` both `registerUndo()`. `applyPixelateStrength` does not — same batch, same shape. It also calls `ensurePixelated()` on every change, and `pixelBlockSize` yields roughly 55 distinct blocks across 0...1, so one drag across the slider runs `pixelatedImage(from:block:)` over the full image dozens of times on the main thread, each result thrown away by the `filter` on the next step. The body says "one pixelated twin is cached per strength in use" — true at rest, but the drag is the expensive path and it isn't mentioned.

**4. The drop handler and the media-tools button are not in the body.** `openEditor(imageAt:)`, the edit button on the image compressor, `.onDrop(of: [.fileURL])`, and the rewrite of the clipboard size check into `editorAcceptsImage` are a separate feature from the annotation controls, and Before / After / Notes covers none of it. Two things inside it: dropping a non-image file on the editor does nothing and says nothing, while the media-tools path reports `mediaErrorUnsupported` for the same failure — same feature, one call site got the message; and `editorAcceptsImage` replaces `width <= max / height` with `width * height <= max`, the same ceiling with different rounding, so a clipboard image sitting right at the edge is accepted or refused differently than before.

Not verified: `ScreenshotRenderer.fontSize(for:scale:)`, `screenshotText.editButton` and `l10n.s.mediaErrorUnsupported` are outside the diff, so I went by the call sites. I read this, I didn't run it.

**Impact:** Every counter resizes from the wrong baseline on its first corner drag, default settings. The pixelate color swatches do nothing when clicked. The strength slider is not undoable and stutters while dragged.
**Merge:** needs the counter resize baseline in `beginDrag` fixed first — 1, 3 and 4 are judgement calls, 2 is wrong behaviour.

ref #677

FlowSync0:
All four addressed in the push above.

**2, the counter baseline.** Confirmed and fixed exactly where you pointed. A
counter that has never been resized has `fontSize == 0`, so the fallback handed
it the stroke picker's point size while the badge on screen is
`counterDiameter`. `beginDrag` now branches on the tool and passes the badge,
in both places that set `moveFontSize` — the selected-annotation path and the
hit-test path. The fallback inside `updateDrag` stays as a guard but is no
longer the thing doing the work.

**1, the pixelate colour picker.** You are right that it contradicts the
branch's own argument, and it was mine: I added `case .pixelate: return true`
so the strength row would have somewhere to live, without noticing I had turned
the colour swatches on for a tool that never reads them. Both exclusions are
back, and the strength row now sits outside the colour block rather than inside
it — it takes that place for pixelate instead of joining it.

**3, undo and the cost of a drag.** Both real. The strength registers one undo
entry per drag now, grouped through an `onEditingChanged` that closes the
group, so a sweep is one step rather than sixty or none. And the block size is
rounded to a step derived from the base block, so a sweep asks for a handful of
distinct twins instead of one per position it passes through — the cache
already released the unused ones, but it was rebuilding far too many of them to
get there.

**4, coverage.** Fair. The Edit button, `openEditor(imageAt:)` and the ceiling
rewrite belong to #1035 and are only in this diff because this branch is
stacked on it; that PR's body now describes them, including the rounding change
from `width <= max / height` to `width * height <= max`. The mismatched failure
message is fixed there too: an image past the ceiling now says so instead of
reporting an unsupported format.

The drop-on-editor handler you looked for is not in either branch — it exists
on a working branch and I described it in #1035's body by mistake. That is
corrected there.

Not verified by me: I have not measured the twin rebuild before and after, so
"far too many" is reasoning from the block count, not a profile.

PathGao:
Points 1, 2 and 3 check out in the pushed diff. One correction, one small thing.

**The drop handler is in this PR's diff.** `ScreenshotEditorView.swift` gains `@State private var isDropTargeted`, the accent border overlay, `.onDrop(of: [.fileURL], isTargeted: $isDropTargeted)` on the root, and `acceptDroppedImage(_:)` itself. Neither `onDrop` nor `openEditor(imageAt:)` exists in that file on `main`, so it arrived with one of these two commits. I can't tell which from the diff — if it rode in on #1035's, that is where it belongs — but it is in the tree, and after the correction it is described in neither body. The silent failure stands as written: `acceptDroppedImage` discards what `openEditor(imageAt:)` returns, so a PDF dropped on the editor does nothing and says nothing.

**The rounded block does not preserve the default for every capture.** `step = max(2, base / 8)`, so `base / step` is exactly 8 only once `base >= 16`. Below that the step is a flat 2 and an odd base rounds up: `base` 11, 13 or 15 — a capture whose shorter side is 605–659, 715–769 or 825–879 pixels — turns 15 into 16 at strength 0.5. Invisible in practice, but it contradicts "an unset strength to the block size the editor has always used", and the test samples only 1600×1000, where it happens to hold. Quantise around the base rather than around zero:

```swift
let raw = base * (0.35 + clamped * 1.3)
let step = Swift.max(2.0, base / 8)
return Swift.max(4, Int(base + ((raw - base) / step).rounded() * step))
```

Not verified: I read the pushed diff, I didn't run it. The undo grouping and the twin count are reasoning from the code, same as your note about the rebuild.

**Impact:** A non-image file dropped on the editor window fails silently, default settings. The default pixelate block shifts by one pixel on captures with a short side in those three bands.
**Merge:** needs the drop handler accounted for first — out of the branch, or given the failure message and a line in whichever body owns it.

ref #677

PathGao:
One thing worth catching before this lands, in `restore(_:)`:

```swift
if annotations.contains(where: { $0.tool == .pixelate }) {
    ensurePixelated()
} else {
    pixelated = [:]
}
```

That `else` runs whenever an undo or redo step happens to leave no pixelate annotation, and `baseImage` has not changed at the time. The cache it drops is derived purely from `baseImage`, so everything thrown away there was still valid.

The cost shows up on the way back. `pixelatedImage(from:block:)` allocates two full-resolution `CGContext`s and draws the image twice; on a 5120×2880 capture the upsample buffer alone is about 59 MB. So "draw a pixelate box, ⌘Z, ⌘⇧Z" rebuilds it, and holding either key rebuilds it per step, on the main thread. This PR makes that sharper rather than softer, because the cache is now keyed per block size — one clear can discard several images, and `ensurePixelated()` then rebuilds each strength in `sizes`.

The same `else` exists on main with `pixelated = nil`, so it is not something this PR introduces — it just becomes more expensive here. The narrow fix is to clear only in the branch above it, where `state.image !== baseImage` already tells you the cache is genuinely stale; the `pixelated.filter { sizes.contains($0.key) }` line in `ensurePixelated()` already handles evicting strengths that are no longer in use.

Not something I hit at runtime — it came out of reading the undo path while auditing this file, and I have not run this branch.

PathGao:
Closing with the stack; the reasons for all three are on #1035.

PR #1037, opened 2026-08-26
feat(selection-actions): add local text-transform actions

naveenkrdy:
## Summary

Third of the four PRs @PathGao suggested splitting #713 into, based on #1033's skeleton (including the fixes from [that review](https://github.com/vorssaintapp/vorssaint-utils/pull/1033) - enabled-set storage, the off-main-thread exclusion check). Adds the actions that only rewrite the selection locally and never leave the app or the machine:

- **Paste as Plain Text**
- Case transforms: uppercase, lowercase, capitalize
- Remove Spaces, Underscore, Join Lines, Comma List
- Sort, Reverse, Random (shuffle) - all granularity-aware: lines when the selection has line breaks, else words, else letters, so Reverse on a single word reverses its letters instead of doing nothing
- Quotes, Brackets, URL Encode/Decode, Base64 Encode/Decode
- Calculate - a small hand-rolled arithmetic evaluator (not `NSExpression`, which raises an uncaught `NSException` on malformed input)
- Add to Scratchpad

None of these touch the network or need a new permission - still just Accessibility, same as #1033.

Restores `selectionActionsMaxVisible`'s default to 8 now that there are enough actions (23) for that number to mean something; #1033 shipped it at 4 to match its 4-action total.

## Test plan

- [x] `./build.sh --test` passes (`TESTS OK`, 8650 checks)

Adds behavior tests for the pure logic this PR introduces (previously zero coverage for any of it): the arithmetic evaluator (operator precedence, parentheses, division by zero, a trailing-operator malformed case) and the sort/reverse granularity logic (line vs. word vs. letter). `ArithmeticEvaluator` moved from `SelectionActionRunner.swift` into `SelectionActionCatalog.swift` so it compiles in the test harness's minimal source list alongside the already-tested `TextListSupport`, without pulling in the AppKit/service side effects the runner file carries.

Manual: enabled each transform on real selections in TextEdit and a browser - case changes, sort/reverse/shuffle on multi-line and single-word selections, quotes/brackets wrapping, URL and Base64 round-trips, Calculate on a few expressions, Add to Scratchpad. Not tested: Paste as Plain Text specifically through this feature's bar (exercised the underlying `PastePlainService.plainText(from:)` already via the existing Paste Plain quick tool, but not the Selection Actions bar's own entry point to it).


naveenkrdy:
Stacked on #1033 (which is itself stacked on #1032, both not yet merged) - this diff will shrink as those land and I rebase onto `main`.

PathGao:
Three from the Core files.

**`sorted()` is not alphabetical.** `TextListSupport.sorted` uses Swift's `<`, which is scalar order: `Zebra` before `apple`, `Äpfel` after `Zebra`. You ship `sortDescription` as "Sorts lines, words, or letters alphabetically" in 13 languages — in German, Russian and Japanese that string promises something the code doesn't do. stdlib fix:

```swift
case .lines(let items):
    return items.sorted { $0.localizedStandardCompare($1) == .orderedAscending }.joined(separator: "\n")
```

(and `String($0).localizedStandardCompare(String($1))` for the letters case). It's also numeric-aware, so `item2` precedes `item10` — what people expect from a Sort button.

**Two members missing from the applicability guard.** `calculate` gets `looksLikeExpression`, but `urlDecode` and `base64Decode` fall through to `default: return true`. They appear on every selection, and on text that isn't encoded `removingPercentEncoding` / `Data(base64Encoded:)` return nil and the click does nothing. `appliesTo`'s own doc says "whether this action makes sense for the given selection right now":

```swift
case .urlDecode: return text.contains("%")
case .base64Decode: return Data(base64Encoded: text) != nil
```

**`looksLikeExpression` matches dates.** `2026-08-26` is all-allowed characters, has digits, has an operator → Calculate shows up, and clicking it replaces the selection with `1992`. Same for `555-1234` and ranges like `10-20`. `calculate` is in `defaultEnabled`, so that's the out-of-box behavior, not an opt-in edge. I don't have a cheap fix that stays cheap — rejecting `\d+-\d+-\d+` covers the case people will actually hit.

Smaller: `granularity(of:)` uses `components(separatedBy: .newlines)`, which treats `\r` and `\n` as separate separators — on CRLF text Sort/Reverse/Random insert a blank line between every pair. The `.lines` branch also rejoins with `\n` unconditionally, so all three silently rewrite line endings.

Not verified: `SelectionActionRunner.swift`, the settings UI and the service side are past the diff truncation, so this is the Core files only. I also couldn't see the full default-shortcut table to check ctrl-opt-cmd-A for a collision.

**Impact:** Calculate offering itself on dates and phone numbers hits default installs. The sort order and the two decode guards only bite after someone switches those actions on. None of it is data loss — the transforms are all user-initiated and undoable in the host app.
**Merge:** hold for #1033 and #1032 landing.

ref #713

naveenkrdy:
All four addressed on the rebased branch (1a2cfb6):

- `looksLikeExpression` false positives on `.urlDecode`/`.base64Decode`: added their own `appliesTo` guards (`text.contains("%")`, `Data(base64Encoded:) != nil`) instead of falling through to the catch-all `true`.
- `TextListSupport.sorted` using default `<`: switched all three cases to `localizedStandardCompare`, matching your suggested code.
- Calculate offering itself on dates/phone numbers/ranges: added a rejection for strings that are only hyphenated digit groups (`\d+(-\d+)+` shape) - covers `2026-08-26`, `555-1234`, `10-20` without touching real arithmetic like `10-2`... actually that one *is* a two-group case too, so it's rejected as well, which matches - a bare `a-b` subtraction typed without spaces reads the same as a range, and the multi-group case (`2026-08-26`) is the one people will actually hit.
- CRLF double-splitting in `granularity(of:)`: normalizing `\r\n` → `\n` before `.components(separatedBy: .newlines)`.

Rebased onto #1033's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8650 checks).

PathGao:
Both new guards approximate the operation instead of running it, so the case I flagged is still open — the button appears and the click does nothing.

`text.contains("%")` is true for `100% sure`, and `removingPercentEncoding` returns nil on a stray `%`. `Data(base64Encoded:)` accepts any run of base64 characters whose length is a multiple of 4 — `test`, `word`, `code`, `data` all pass, and decode to bytes that aren't text. Do the work in the guard instead:

```swift
case .urlDecode: return text.removingPercentEncoding.map { $0 != text } ?? false
case .base64Decode:
    return Data(base64Encoded: text).flatMap { String(data: $0, encoding: .utf8) } != nil
```

Exact instead of approximate, same length, and the `%` check falls out for free. What the runner does with a decode that returns nil is past the diff truncation for me — no-op click at best.

Second, and I can't confirm this one: `SelfUninstall.suspendInputInterceptors()` (`SelfUninstall.swift:63`) is the list of every Accessibility-backed interceptor, and its own comment says leaving one live while Accessibility is revoked stalls the OS input queue. `ScratchpadService` and `CommandBarService` are both in it. `SelectionActionsService` belongs there too — `AppDelegate` teardown calling `suspend()` on it says it holds something worth tearing down. That file is past my truncation point, so it may already be there, or in #1033. Say which. Same question with lower stakes for `CleaningModeManager.activate()` (`:66-76`): it pauses Radial Menu and the mouse services so a wipe can't fire them, and a stray trackpad drag that creates a selection would put the bar over the cleaning overlay.

The four from last round read as fixed — sort, the two guards' placement, the hyphen rule, the CRLF normalize. The `10-2` tradeoff is yours and you made it explicitly.

Not verified: everything under `Services/` and the settings UI is past the 80 KB diff cut on my end, so this is the Core files again.

**Impact:** URL/Base64 Decode offering themselves on ordinary words — both off by default, so it only bites after someone enables them. The teardown gap, if it's real, is an input-queue freeze during uninstall on any machine with the feature on.
**Merge:** hold for #1033 and #1032 landing.

ref #713

naveenkrdy:
Fixed to run the actual operation instead of approximating it (950faf4):

```swift
case .urlDecode: return text.removingPercentEncoding.map { $0 != text } ?? false
case .base64Decode:
    return Data(base64Encoded: text).flatMap { String(data: $0, encoding: .utf8) } != nil
```

Matches your suggested code exactly, and matches what `SelectionActionRunner` itself does for both cases - `100% sure` and `test`/`word`/`code`/`data` no longer pass.

On the two "can't confirm" points:

- **`SelfUninstall.suspendInputInterceptors()`**: it wasn't in the list. Added `SelectionActionsService.shared.suspend()` there on #1033 (84c602d), alongside `ScratchpadService`/`CommandBarService` - that's the PR that introduces the service and its `AppDelegate` teardown call, so it belongs with it rather than here.
- **`CleaningModeManager.activate()`**: haven't touched this - flagging it rather than guessing. `CleaningModeManager` pausing Radial Menu and the mouse services during a wipe is existing behavior this PR doesn't change; whether Selection Actions' bar needs the same pause is a real question but a separate one from what #1033/#1037 add, and I don't want to make that call without reading `CleaningModeManager.activate()`'s current scope first. Can pick it up as its own follow-up if you think it's worth one.

Rebased onto #1033's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8650 checks).

PathGao:
`calculate` is the same shape as the two decodes and didn't get the same treatment: the guard checks the character set, the operation runs the parser, and the two disagree. `50%`, `1/0`, `(1+2`, `10+` all pass `looksLikeExpression` and all return nil from `ArithmeticEvaluator.evaluate` — button appears, click does nothing. Unlike the decodes, `calculate` is in `defaultEnabled`, so this one is out-of-box. Same move you already made twice:

```swift
case .calculate:
    return SelectionActionCatalog.looksLikeExpression(text)
        && ArithmeticEvaluator.evaluate(text) != nil
```

The charset prefilter still earns its keep — it rejects prose before the parser runs.

On `CleaningModeManager`, I looked and my framing was wrong. The tap's mask is keys, scroll, systemDefined and gestures only, and `handle` says so out loud ("Pointer movement and clicks remain available for the Unlock button"), so a stray drag during a wipe can still make a selection. But the overlay panel sits at `CGShieldingWindowLevel()` (`CleaningModeManager.swift:208`), so the bar draws under it, not over it — unless the bar is also at shielding level, which is past my truncation. That's a hidden bar and an AX read nobody sees, not the visual mess I described. Not worth a follow-up on my read. If you do add it, `activate()` and `deactivate()` are two hand-maintained parallel lists (5 suspends / 5 syncs) — both get edited or the pause never lifts.

The two decode guards read right. `SelfUninstall` on #1033 is the right home for it.

Not verified: everything under `Services/` and the settings UI is still past the 80 KB diff cut on my end.

**Impact:** Calculate offering itself on `50%`, `1/0` and unbalanced parens — default install, no-op click.
**Merge:** hold for #1033 and #1032 landing.

ref #713

naveenkrdy:
Fixed, matching your suggested code exactly (f1822d1):

```swift
case .calculate:
    return SelectionActionCatalog.looksLikeExpression(text)
        && ArithmeticEvaluator.evaluate(text) != nil
```

`50%`, `1/0`, `(1+2`, `10+` no longer offer Calculate. `looksLikeExpression` stays as the charset prefilter it already was - still worth running first since it's cheap and rejects prose before the parser does any work, just no longer treated as sufficient on its own.

Thanks for double-checking `CleaningModeManager` rather than leaving the earlier framing standing - appreciate you correcting your own read there.

Rebased onto #1033's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8650 checks).

PathGao:
Only relevant to the two paste files in this 34-file branch, not to the feature: `SyntheticPasteSupport` here duplicates `TransientPaste` on main, which already provides all three of its functions plus pasteboard restore and a change-count guard. Details and the line references are in #1032, since that PR introduces the file.

Nothing else in this branch is affected as far as I can see — the Selection Actions side does not overlap anything I have looked at.

PathGao:
Inherits the `TransientPaste` change from #1033 (same `SyntheticPasteSupport.swift` here); the transforms themselves are clean and every review round was answered. Rebase once the base is fixed.

vorssaint:
closing this one along with #1033, since it is built on top of that bar and cannot land without it
the text transforms themselves also double up on things the command bar already does, its calculator and its case changes with the language aware forms, and read the clipboard on the main thread, which the app moved off that thread this month
if a smaller floating bar ever comes back on a fresh base, a few transforms that reuse the command bar's calculator and case logic could come with it

PR #1038, opened 2026-08-26
feat(selection-actions): add actions that reach outside the app

naveenkrdy:
## Summary

Fourth and last of the PRs @PathGao suggested splitting #713 into, based on #1033's skeleton (including [its review fixes](https://github.com/vorssaintapp/vorssaint-utils/pull/1033) - enabled-set storage, the off-main-thread exclusion check). Developed independently of #1037 (the text-transform PR) - both branch from #1033, neither depends on the other.

Adds the actions that leave the app or the machine:

- **Search Web**, **Open Link**, **Open in Mail** - Open Link/Mail only offer themselves when the selection actually looks like a URL or email address (`NSDataDetector`).
- **Translate** - opens Google Translate with the selection pre-filled, target language from the system locale.
- **Send to AI** - ChatGPT or Claude, configurable, with an optional "Temporary Chat" toggle for ChatGPT.
- **Run in Terminal** - the only action that needs a new permission (`automationTerminal`, an AppleScript `do script` event to Terminal.app). Shows a native confirmation before ever touching the shell, on by default, togglable in the action's own settings gear. Tab vs. new-window is also configurable.
- **Convert Currency** - the only thing in the whole feature that touches the network. `CurrencyConversionSupport` calls `api.frankfurter.dev` (free, keyless, verified against its live endpoint and docs) for an exchange rate and fails silently - no replace - on any network or parsing problem, rather than risk swapping the selection for garbage. `CurrencyDetector` only fires on a conservative symbol-or-ISO-code-next-to-a-number pattern.

`AppFeature.selectionActions.permissions` is now `[.accessibility, .automationTerminal]` - the only one of the four PRs that changes it.

## Test plan

- [x] `./build.sh --test` passes (`TESTS OK`, 8645 checks)

Adds behavior tests for `CurrencyDetector.detect` - previously zero coverage: a leading symbol, a trailing ISO code with thousands separators, and the negative case of a plain three-letter word not being mistaken for an unknown currency code.

Manual: Search Web/Translate/Send to AI on a plain selection, Open Link on a URL and Open in Mail on an email address (confirmed neither offers itself on the other's input), Run in Terminal's confirmation dialog and both tab/window targets, Convert Currency on a few symbol and ISO-code selections against a couple of target currencies. Not tested: Convert Currency's actual network failure path (no easy way to force `api.frankfurter.dev` to fail from here) - the fallback is "leave the selection alone," verified by reading `CurrencyConversionSupport`, not by triggering it live.


naveenkrdy:
Stacked on #1033 (not yet merged) - this diff will shrink once that lands and I rebase onto `main`. Independent of #1037 (both branch from #1033, neither depends on the other).

PathGao:
Four things.

**1. `parseAmount` strips commas, so decimal-comma locales convert 100× wrong.**

`CurrencyDetector.detect("1,50 EUR")`: `1,50` matches `[0-9][0-9,]*\.?[0-9]*`, `parseAmount` deletes the comma, result is `150.0`. In de/fr/es/it/pt-BR — all shipped languages — that is the ordinary way to write €1.50, and `convertCurrency.isTransform == true`, so the selection gets replaced with the conversion of 150 EUR. This is the exact failure the file's own comment says it avoids ("a wrong source currency silently produces a wrong converted amount"); the guard covers the code and not the number. The locale is already in play one line away — `selectionActionsCurrencyTarget` defaults to `Locale.current.currency` — so use it:

```swift
private static func parseAmount(_ raw: String) -> Double? {
    let f = NumberFormatter()
    f.numberStyle = .decimal
    f.locale = .current
    return f.number(from: raw)?.doubleValue
}
```

The digit class needs widening to `[0-9][0-9.,]*` too, or `1.234,56` never reaches the parser — today that one fails the anchored match and returns nil, which is at least the safe direction.

**2. `looksLikeLink` returns true for email addresses.**

`NSDataDetector`'s `.link` type matches bare emails and returns a `mailto:` URL, so `looksLikeLink("a@b.com")` is true and Open Link appears next to Open in Mail. `looksLikeEmail` has the scheme check; `looksLikeLink` doesn't:

```swift
guard let match = detector.firstMatch(in: trimmed, range: range), match.range == range,
      let url = match.url, url.scheme != "mailto"
else { return false }
return true
```

Body says "confirmed neither offers itself on the other's input" — the mail-on-a-URL direction holds, the link-on-an-email direction doesn't, unless something past my part of the diff filters it.

**3. The default currency target isn't constrained to `knownCodes`.**

`Locale.current.currency?.identifier ?? "USD"` can produce a code the "Convert to" picker never offers. TWD is one, and zh-TW/zh-HK are shipped languages; a `Picker` whose selection matches no tag renders blank. Constrain it where the set lives:

```swift
static var defaultTargetCode: String {
    let code = Locale.current.currency?.identifier ?? ""
    return knownCodes.contains(code) ? code : "USD"
}
```

Worth checking alongside it: frankfurter serves ECB reference rates, ~30 currencies. `knownCodes` lists 43, and AED / SAR / QAR / KWD / PKR / BDT / EGP / NGN / ARS / CLP / COP / PEN / VND / UAH / RUB look like they fall outside that set. If they do, the bar offers Convert Currency on "500 AED" and the silent-no-replace path makes the button do nothing. `CurrencyConversionSupport` is past the truncation point in the diff I received, so I don't know what it does with an unsupported code — diffing `knownCodes` against a live `GET /v1/currencies` settles it either way.

**4. The symbol set is written three times.**

`symbolToCode`'s keys, plus the same alternation inlined in two regex literals. Add a symbol to the dictionary and it silently never matches anything. Derive it:

```swift
private static let symbolPattern = symbolToCode.keys
    .sorted { $0.count > $1.count }   // R$ before $
    .map(NSRegularExpression.escapedPattern(for:))
    .joined(separator: "|")
```

**Not verified:** the diff I got is cut at 80k of 184k bytes. `SelectionActionsService`, `CurrencyConversionSupport`, the settings UI and the new tests are all past the cut and I have not read them. Everything above comes from `CurrencyDetector.swift`, `SelectionActionCatalog.swift`, `Defaults.swift` and `FeatureCatalog.swift`.

**Impact:** #1 rewrites a selection with a 100×-wrong amount for anyone who writes money with a decimal comma, which is five of the shipped languages. #2 adds one extra button to the bar. #3 leaves the "Convert to" picker blank in some locales. #4 none today.
**Merge:** needs the `parseAmount` fix first. #1033 has to land ahead of this either way.

naveenkrdy:
All four addressed on the rebased branch (31a15ab):

- `parseAmount` decimal-comma bug: switched to `NumberFormatter(.decimal, locale: .current)`, matching your suggested code exactly. Widened the digit class to `[0-9][0-9.,]*` in all four patterns so `1.234,56` reaches the parser too.
- `looksLikeLink` matching emails: added the `url.scheme != "mailto"` guard, same as `looksLikeEmail`'s existing check.
- Default currency target: added `CurrencyDetector.defaultTargetCode`, falling back to USD when `Locale.current.currency` isn't in `knownCodes`. `Defaults.swift` now calls it instead of the raw `Locale.current.currency?.identifier ?? "USD"`.
- Symbol pattern written three times: `symbolPattern` is now derived from `symbolToCode.keys`, sorted longest-first, both regex literals interpolate it.

On your "worth checking" note - you called it right. Diffed `knownCodes` against a live `GET /v1/currencies`: frankfurter serves 30 codes, and the 15 you listed (AED, SAR, QAR, KWD, PKR, BDT, EGP, NGN, ARS, CLP, COP, PEN, VND, UAH, RUB) aren't among them. Removed all 15 from `knownCodes`. Also found the same gap one layer down: `symbolToCode` still mapped ₽/₴/₫ to RUB/UAH/VND, and the symbol-match code paths never gated on `knownCodes` at all - so "₽500" would detect as RUB and then silently fail to convert, same bug as #3 but reachable through a different path. Removed those three symbol entries too. `CurrencyConversionSupport` itself needed no change - it was already failing to nil correctly, the bug was purely in what got offered as detectable/convertible upstream of it.

Rebased onto #1033's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8645 checks).

PathGao:
Checked the four fixes against the updated diff — all four land, and pruning `knownCodes`/`symbolToCode` closes the frankfurter path I couldn't reach last time. Two things left.

**1. `automationTerminal` isn't gated on Run in Terminal actually being on.**

`AppFeature.selectionActions.permissions` returns `[.accessibility, .automationTerminal]` unconditionally, but `runInTerminal` isn't in `defaultEnabled`. `activeFeatures` has a `switch (feature, permission)` table for exactly this — eight entries already, `(.screenRecorder, .microphone)` being the same shape: optional source, off by default, gated on its own key. Nothing was added for this one, so the moment someone turns Selection Actions on, the Permissions page flips Automation → Terminal from unused to "used by Selection Actions" with a missing chip, for an action they never enabled.

`stringFor` is already injected and `SelectionActionCatalog.isEnabled` already takes the raw string:

```swift
case (.selectionActions, .automationTerminal):
    return SelectionActionCatalog.isEnabled(
        .runInTerminal,
        enabledRaw: stringFor(DefaultsKey.selectionActionsEnabledActions) ?? "")
```

`MetricsTests` around line 11082 (`activeSet(.automationTerminal) == [.homebrew]`) passes either way — it never turns the feature on — so nothing catches this today.

**2. Nothing enforces `symbolToCode.values ⊆ knownCodes`.**

That's the invariant ₽/₴/₫ just violated, and it lives only in the comment above the dictionary. Next time frankfurter drops a currency it breaks the same silent way. Making `symbolToCode` internal and asserting the containment is a one-liner in the tests.

**Not verified:** still the 80k-of-184k truncation — `SelectionActionsService`, `CurrencyConversionSupport`, the settings UI and the new tests are all past the cut. One thing worth a look there: `parseAmount` is locale-dependent now, so any test asserting a parsed value is too. `"4,500.50 INR"` → 4500.5 under en_US, nil (or something else) under de_DE. If those assertions are unconditional they'll pass on your machine and CI's locale and fail on someone else's.

**Impact:** #1 reports a permission as needed and missing for an action that ships off. #2 none today.
**Merge:** needs the `automationTerminal` gate first. #1033 still has to land ahead of this either way.

naveenkrdy:
Both fixed, and picked up your locale-fragility note too (05d5dba):

- **`automationTerminal` gate.** Added the case, matching your suggested code exactly:
```swift
case (.selectionActions, .automationTerminal):
    return SelectionActionCatalog.isEnabled(
        .runInTerminal,
        enabledRaw: stringFor(DefaultsKey.selectionActionsEnabledActions) ?? "")
```
Also extended the test at line 11055 rather than leaving it as the pass-either-way check you flagged: one case turns Selection Actions on without Run in Terminal (still `[.homebrew]` only), one turns Run in Terminal on too (`[.homebrew, .selectionActions]`) - so a regression on this gate now actually fails a test.

- **`symbolToCode.values ⊆ knownCodes` invariant.** Made `symbolToCode` internal and added `expect(Set(CurrencyDetector.symbolToCode.values).isSubset(of: CurrencyDetector.knownCodes), ...)` next to the other currency tests. A future frankfurter drop now fails a build instead of failing silently at runtime.

- **Your "worth a look" on locale-dependent test assertions - correct, and it was live.** The `"1,250.50 EUR"` test would pass under en_US/en_IN and fail wherever `,` is the decimal separator, exactly the bug this PR fixes for real users. Rebuilt it to construct the expected string from `Locale.current`'s own `NumberFormatter` output instead of a literal, so the test asserts the real invariant (parse back what this locale itself would produce) rather than one locale's formatting.

Rebased onto #1033's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8648 checks).

PathGao:
Both fixes land. The `automationTerminal` case is right, the subset test pins the invariant, and the metrics test now fails on a regression instead of passing either way.

One thing the locale fix moved rather than removed.

**`parseAmount` reads the Mac's convention; the selected text carries the writer's.**

That's the case Convert Currency exists for — a price you didn't write, in a currency you don't use. `NumberFormatter(locale: .current)` can only apply one convention:

- de/fr/es/it Mac, selection `$1,234.56` — under German rules `,` is the decimal point and a `.` after it is malformed. It cannot come out as 1234.56. It comes out as nil or as 1.234 and I can't tell which without running it: one of those is "no button", the other is a silent 1000×-low replacement.
- en Mac, selection `19,50 EUR` — `,` reads as grouping, so 1950, or 19, or nil.

`CommandBarMath.readNumber` already decides this explicitly and convention-agnostically: when both separators appear the last one is the decimal point, when only one appears it is grouping only if exactly three digits follow. Both of the above come out right, and `1,50` still comes out 1.5. It's private, so the smallest reuse is an entry point beside it:

```swift
// CommandBarMath.swift
static func number(from raw: String, locale: Locale = .current) -> Double? {
    var characters = Array(raw)
    var index = 0
    guard let value = readNumber(&characters, &index,
                                 decimalSeparator: locale.decimalSeparator?.first ?? ".",
                                 groupingSeparator: locale.groupingSeparator?.first ?? ",")
    else { return nil }
    return index == characters.count ? value : nil
}
```

and `parseAmount` becomes `CommandBarMath.number(from: raw)`.

What stays ambiguous either way: a lone separator with exactly three digits after it (`€1.234` on an en Mac) is undecidable from the text, and falling back to this Mac's convention is the right guess. Not asking for that one.

The rebuilt test can't catch this by construction — asserting that this Mac parses back what this Mac formats is a round trip, and the failure is in the cross-convention direction. A literal `"$1,234.56"` expecting 1234.56 is locale-independent under `readNumber`, and would have caught it.

Smaller: the header on the new `Core/PanelOrderItem.swift` says the protocol is conformed to by `PanelSectionID` and `SelectionAction`. `main` has eight more — `QuickToggleAction`, `QuickLauncherItem`, `UtilityPanelItem`, `ControlPanelItem` and the four section `Block`s. The file exists to explain the protocol's reach, so it's worth being right about it.

**Not verified:** same truncation, 80k of 191k. `SelectionActionsService`, `CurrencyConversionSupport`, the settings UI and the tests are all past the cut — what I say about the tests above comes from your description of them, not from reading them. The `NumberFormatter` behaviour is reasoning about how it parses, not a run.

**Impact:** a selection written in the other decimal convention either offers no button or gets its amount replaced by a 100×/1000×-wrong conversion. Needs Convert Currency switched on — it isn't in `defaultEnabled`.
**Merge:** needs the amount parse first. #1033 still has to land ahead of this either way.

naveenkrdy:
All three fixed (7a9b704):

**`parseAmount` was solving the wrong problem.** Added `CommandBarMath.number(from:locale:)` as a thin public entry point around the existing `readNumber` (matches your suggested code exactly), and `parseAmount` now just calls it. Convention inferred from the text's own separator pattern, not from `Locale.current` - your two examples both come out right regardless of which locale runs it.

**Test rebuilt to actually catch the bug it's testing for.** Dropped the round-trip (proves same-convention, not cross-convention) for two literal, convention-fixed inputs - `$1,234.56` -> 1234.56 and `19,50 EUR` -> 19.50, both asserted to parse the same under any running locale. Traced both through `readNumber` by hand before adding them (last-separator-wins when both appear; single separator is grouping only with exactly three digits after) - they do.

**`PanelOrderItem.swift`'s header - worse than incomplete, it was wrong.** `PanelSectionID` was never a conformer at all; I grepped every actual `: PanelOrderItem`/`, PanelOrderItem` site and there are 10, not the 2 named. Rewrote the comment with the real list (`SelectionAction`, four section `Block` enums, `UtilityPanelItem`, `ControlPanelItem`, `QuickToggleAction`, `QuickLauncherItem`).

Rebased onto #1033's current tip. Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8649 checks).

PathGao:
The separator fix moved the locale dependency one layer down instead of out.

**`parseAmount` passes the Mac's separators, but the regex upstream only ever produces `.` and `,`.**

`readNumber`'s scan loop ends the number on anything that isn't a digit or one of the two separators it was handed, and `number(from:)` returns nil when the read didn't reach the end. So on a Mac where neither separator is `.`, every dotted amount dies at the dot:

- fr / ru — decimal `,`, grouping a space. `$1,234.56` reads `1,234`, stops at the `.`, nil. `100.50 USD` reads `100`, stops, nil. No button at all, for the most ordinary way to write a price. Both are shipped languages.
- de_CH — grouping `’`, so `,` is unknown: `19,50 EUR` reads `19`, stops, nil.

de/it/es/pt/tr are the locales where `.` and `,` happen to be exactly the pair in play, which is why last round's two examples come out right there and only there.

The regex is `[0-9][0-9.,]*`. Two characters, and `readNumber` already decides their roles from the text's own pattern. Say so instead of asking the Mac:

```swift
private static func parseAmount(_ raw: String) -> Double? {
    CommandBarMath.number(from: raw, decimalSeparator: ".", groupingSeparator: ",")
}
```

which means `number` takes the two separators rather than a `Locale` — my parameterization last round was the wrong one, threading `Locale` through was the mistake.

Same thing reaches the tests. `$1,234.56` -> 1234.56 isn't convention-fixed as written; it's true only where `Locale.current.groupingSeparator == ","`, so it passes on your Mac and CI's and fails on a French one. With the separators passed explicitly the assertion is true by construction.

**Not verified:** truncation again, 80k of 191k. `CommandBarMath.swift` is under `Services/CommandBar/` and past the cut, so what I read is `readNumber` on `main`, not your wrapper — the above assumes the wrapper is the code I posted, full-consumption check included. I also didn't run `Locale(identifier: "fr_FR").groupingSeparator`; the argument only needs that some shipped locale's separators aren't `.`/`,`, and passing them explicitly removes the dependency either way.

**Impact:** Convert Currency offers no button for any amount containing a `.` on space-grouping locales — fr and ru ship. Needs the action switched on; it isn't in `defaultEnabled`.
**Merge:** needs the two separators passed explicitly first. #1033 still has to land ahead of this either way.

naveenkrdy:
Fixed in 3b2cbeb: `parseAmount` now calls a new `CommandBarMath.number(from:decimalSeparator:groupingSeparator:)` overload with literal `"."`/`","` instead of `Locale.current`'s separators. `number(from:locale:)` stays for the calculator's own callers (typed input correctly wants this machine's convention) and is now a thin wrapper over the same overload.

Root cause confirmed as described: `CurrencyDetector.detect`'s regex (`[0-9][0-9.,]*`) only ever emits digits plus literal `.`/`,` in the amount group, so deriving expected separators from `Locale.current` was the wrong source - the regex is what actually constrains the alphabet, not the Mac's locale. On fr/ru/de_CH-style locales `readNumber` stopped at the first unrecognized `.`/`,` and the parse silently returned nil.

Your proposed call shape (`number(from:decimalSeparator:groupingSeparator:)`) didn't exist yet as an overload - added it, with `parseAmount` calling it directly rather than threading a `Locale` through.

Not expanded: true French-native space-grouped input (e.g. `1 234,56`) still isn't detected at all, since the regex itself doesn't match spaces in the numeric run. That's a separate, pre-existing scope limit, not something this fix touches - flagging so it doesn't read as still-broken.

`./build.sh` clean, `--selftest` OK, `--test` OK (8649 checks).

PathGao:
The hardcoded `"."`/`","` fixes fr/ru/de_CH. It also changed one shape that neither of us called out, and that one came out of my suggestion.

**`readNumber` applies the three-digits test to the grouping separator only, so `1,234` and `1.234` get opposite readings.**

With the two characters fixed as decimal `.` / grouping `,`:

- `1,234` — `hasGrouping` only → `looksLikeGrouping` → **1234**
- `1.234` — `hasDecimal` only → straight to `Double` → **1.234**

Same shape, and `1.234 €` is how de/es/it/pt/tr write €1234. `convertCurrency.isTransform == true`, so the selection gets replaced with the conversion of €1.234. Before 3b2cbeb a German Mac read that one right — `.` was its grouping separator, so the test reached it — which is why it stayed out of the fr/ru argument. Now it's wrong on every Mac regardless of who wrote the text.

The rule `readNumber` already has is the right rule; it's just reaching one character out of two. Hand it whichever one turns up alone:

```swift
private static func parseAmount(_ raw: String) -> Double? {
    // Whichever separator appears alone goes in as the grouping one, so
    // readNumber's three-digits test decides its role instead of a lone
    // "." defaulting to a decimal point.
    let dotAlone = !raw.contains(",")
    return CommandBarMath.number(from: raw,
                                 decimalSeparator: dotAlone ? "," : ".",
                                 groupingSeparator: dotAlone ? "." : ",")
}
```

Anything containing both characters keeps today's arguments exactly, so `$1,234.56` and `1.234,56` are untouched. `1,50` → 1.5 and `100.50` → 100.5 still hold; `1.234` → 1234, and `1.234.567` → 1234567 instead of today's nil.

**This moves the error rather than removing it, so it's a choice.** `$3.459` — a US gas price — becomes 3459, the mirror of the case it fixes. The tiebreaker I'd use: nothing left in `knownCodes` has three decimal places (JPY/KRW 0, the rest 2), so inside this feature a `.` with exactly three digits behind it is never a payable fraction, while `.`-as-thousands is most of the price tags in Europe. If you'd rather keep three-decimal quotes parsing, leaving it is defensible — but then the comma side should lose its grouping test, because one character getting the heuristic and the other not is the part that's arbitrary.

Smaller, and possibly not yours: `AXCopy.messagingTimeout`'s comment says the timeout goes on the target app element, "never the system-wide one: a timeout there is the default for the whole process." `AppDelegate.boundAccessibilityWaits()` on `main` sets it on the system-wide element deliberately, and its comment says the process-wide reach is the point — "it also covers the places that never set one of their own." Whichever is right, the next person reads one of them and believes it. `AXCopy.swift` is the only new file in this diff without your author line, so this may belong to #1033.

**Not verified:** 80k of 191k again. `CommandBarMath.swift` is past the cut, so the new `number(from:decimalSeparator:groupingSeparator:)` is your description and not something I read — the trace above is `readNumber` on `main` plus the assumption that the overload hands the two characters straight through and still rejects a partial consume. `SelectionActionsService` and the tests are past the cut too.

**Impact:** a price written with `.` as the thousands separator gets the selection replaced with a 1000×-low conversion, on any Mac. Needs Convert Currency switched on; it isn't in `defaultEnabled`.
**Merge:** needs the lone-separator rule made symmetric first, whichever way you settle it. #1033 still has to land ahead of this either way.

naveenkrdy:
Fixed, matching your suggested code exactly (0629ed0):

```swift
private static func parseAmount(_ raw: String) -> Double? {
    let dotAlone = !raw.contains(",")
    return CommandBarMath.number(from: raw,
                                 decimalSeparator: dotAlone ? "," : ".",
                                 groupingSeparator: dotAlone ? "." : ",")
}
```

Took your tiebreaker as the deciding argument - nothing in `knownCodes` has three decimal places, so a lone "." with exactly three digits after it is never a payable fraction inside this feature, while `.`-as-thousands is most of Europe's price tags. Left the comma side's grouping test as-is rather than making it symmetric the other way, per your framing of which asymmetry is arbitrary and which isn't.

Added a regression pair: `1.234 EUR` -> 1234 (the case that was wrong) and `1,234 EUR` -> 1234 (the unaffected pair, kept alongside it so a future change to one separator's role gets caught against the other).

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8651 checks, up from 8649). Ran red/green by hand: reverted to the hardcoded `"."`/`","` call, reran, got exactly the new `1.234 EUR` assertion failing and nothing else, restored, confirmed green again.

PathGao:
Inherits the `TransientPaste` change from #1033. One thing of its own: the lone-separator condition from 2026-08-27 is met in 0629ed06 (`dotAlone`, plus the `1.234 EUR` / `1,234 EUR` pair), and the `incomplete` label is stale on that point. `Convert Currency` reaches `api.frankfurter.dev` — keyless, one GET, no-op on failure — but it is still a menu action that stops working the day a third-party service does; it is off by default, and whether it ships is the maintainer's call, not a review item. Say in the body that it can be dropped without touching the rest.

PathGao:
Correcting my note above on Convert Currency: please drop the action from this branch rather than leave it to the body.

Two things about it cannot be fixed inside the feature. Every use sends the user's IP and the app's User-Agent to a server run by one person outside the project, which would make it the first thing in the app that contacts a host other than Apple's or GitHub's. And live rates have no on-device source, so the day that service goes away the action goes with it, silently. Neither improves with a different endpoint or a user-configured one. Remove it; search, translate and the terminal action do not have this shape and stay.

RuanMD:
I prepared a focused follow-up for the cross-app activation issues described in the handoff and would like to reopen/continue this integration path for maintainer review.

Follow-up branch:
https://github.com/RuanMD/vorssaint-utils/tree/fix/selection-actions-external-followup

The branch is based on the current head of this PR (`0629ed0`) and contains only the corrective changes, not a duplicate implementation:

- preserve the safe default shortcut until a user preference exists;
- resolve the hotkey target from the current focus instead of retaining a stale accessibility element;
- reactivate the global monitor after Accessibility permission is granted;
- keep the radial shortcut lifecycle aligned with the global monitor;
- preserve the default action catalog when the stored action list is absent.

Verification performed on the Developer bundle:

- `swift build` — passed;
- `./build.sh --test` — `TESTS OK (8651 checks)`;
- `./build.sh --dev --install` — passed;
- `build/VorssaintDeveloper --selftest` — `SELFTEST OK`;
- `git diff --check` — passed.

The manual cross-application matrix is still pending. During automation, a semantic hotkey driver degraded the multi-modifier chord, so the real shortcut requires native `CGEvent` injection; that limitation is not being reported as a successful manual validation.

This follow-up could not be opened as a stacked PR against `feature/selection-actions-external` because that base branch belongs to another fork and GitHub rejected the cross-fork base. Please either rebase/merge this PR and let me retarget the follow-up to `upstream/main`, or provide an upstream dependency branch for a clean PR.

No official app was changed; validation used only `/Applications/Vorssaint (Developer).app` and its `.dev` Accessibility/TCC path. No secrets or user text were added to source, logs, or fixtures.

Please review the branch and reopen/continue the integration work when the dependency branch is available.

PathGao:
I can't merge, retarget or approve anything here — those are the maintainer's buttons.

On the base: this PR is `CONFLICTING` right now, it's stacked on #1033 which hasn't landed, and my 2026-09-02 request to drop Convert Currency from the branch is still open. All three move the head sha, so anything cut against `0629ed0` gets rebased whichever route you take. Waiting for #1033 to land and then opening against `main` is the cheaper order.

I also can't review a branch link — I only see diffs attached to a PR. Open it against `main` once the base lands, or paste the diff.

One item on your list worth checking now: "preserve the default action catalog when the stored action list is absent". The stored value is the *enabled* set on purpose (#1033's review) — absent should mean the default set, empty should mean empty. If empty maps back to defaults, someone who switched every action off gets them all back on next launch, and a case added in a future release comes up on for everyone instead of off. I haven't read your branch, so this may already be handled.

**Impact:** none on this PR; the flagged one would silently re-enable actions a user turned off.
**Merge:** hold for #1033, plus the Convert Currency removal.

vorssaint:
closing this one. most of these actions hand the selection to outside services, and that does not fit the app.
the ai and translate actions name outside products in the settings and in every language, and the app never names other products.
search and translate are pinned to one outside website, while the mac's own text menu already covers both.
convert currency needs a live rate feed from an outside service, which the app has already turned down elsewhere, and its reply pastes into whatever is focused when it arrives, which can be a different app by then.
run in terminal turns any selected text into a shell command, and the confirmation that guards it can be switched off.
the branch also still carries the paste helper that duplicates what main already has, so it conflicts with main.
a smaller follow up with only open link and open mail on top of the floating bar could still come in once that base lands

PR #1039, opened 2026-08-26
feat(media): compress, split, rotate and convert PDFs in the media tools

FlowSync0:
Addresses #617. Based on #1035, which this branch carries as its first commit;
reviewing the second commit alone is enough.

## Before

A PDF cannot enter the app at all. The image tab accepts `[.image]`, so a PDF
is refused by the file panel and rejected on drop. The only thing PDF-shaped in
the media tools is an output format on the image converter.

#617 describes what that costs: forms that reject a PDF over a size or above a
page count leave you with Preview's blunt "reduce file size" filter, a website
you would rather not hand a bank statement to, or a self-hosted toolbox.

## After

The image tab accepts a PDF, recognises it from the file, and changes shape
around it. **The tool row stays at four entries** — that is a deliberate choice
and the one point I would most like a maintainer's opinion on, since the
reporter asked for a dedicated tool so that "PDF" is visible in the row. Both
readings are defensible and they are one line apart in the code.

What the panel then offers:

- **Compress**, page by page.
- **Page range**, written the way a form asks: `1-3`, `2,5-7`, `4-`. Pages come
  out in the order they were asked for, so a selection can also reorder.
- **Rotate**, in quarter turns, applied to the pages that were kept.
- **Convert**, by choosing an image format instead of PDF.
- **A size target**, the same choice the video tool has carried since #891.

## How compression works, and why it is not the Quartz filter

A scanned page carries all of its weight in one image and nothing else, so it
is re-encoded at the chosen quality. A page holding text is copied out
untouched: rasterizing it would trade searchable words for pixels, which is
precisely what makes the built-in filter unusable on real documents. The test
below shows the split — a two page fixture, one scan and one text page, comes
out 63 % lighter with its text still selectable.

Everything else is PDFKit doing what it already does: `PDFDocument.write(to:)`
copies untouched pages as they are, so nothing is re-encoded that did not need
to be.

## Conversion, both ways

Choosing an image format renders the selected pages: one page leaves as that
picture, several leave as one archive, because handing someone twelve loose
files is not an answer to "export this document". Going the other way, several
images asked for as a PDF now make one document rather than a pile of one-page
files — an ID card scanned in two shots, which is the example in #617's own
thread.

The action button says which of the two is about to happen. A conversion can
legitimately come out heavier than it went in, so promising compression would
be a lie.

## The size target, and one difference from the video one

The encode starts high and comes down only as far as it must. A JPEG that
weighs 400 KB on its own keeps its quality when someone types 20 MB, instead of
being degraded to hit a number. A pass already under the ceiling is never
retried.

## Notes

- **First `import PDFKit` in the tree.** It is a system framework, so the
  no-dependency rule holds, the same way the video tool drives the system's own
  `avconvert`.
- New user-facing text is provided in every supported language.
- The image tab's existing behavior is untouched: several images still produce
  several files for every format except PDF, and the resize, watermark, rename
  and profile options are unchanged.
- A PDF never joins a multi-file batch: a document is one input.

## Checks

- `./build.sh --test` passes, with assertions for the page-range parser: the
  written forms, ordering, duplicates, a range running past the end, and every
  malformed spelling refused rather than silently emptied.
- Measured end to end on a fixture holding one scanned page and one text page:
  476 KB in, 175 KB out, exactly one page re-encoded, text still selectable,
  page dimensions unchanged.
- Exercised by hand on real documents: page ranges, reordering, rotation,
  export to JPEG as a single image and as a zip, and several PNGs combined into
  one PDF.


PathGao:
Reviewed the second commit only, as asked. Five things, roughly in order of what they cost.

**1. An OCR'd scan is copied untouched, so compression does nothing.**

`recompressedPage` decides "this page is a scan" from `page.string` being empty. Scanners embed an invisible OCR text layer — ScanSnap, Acrobat, iPhone Notes and most bank-statement exports all do. Those pages come back non-empty, return `nil`, and get copied straight through. The user picks a quality, waits, and gets a file the same size, with `mediaPDFHint` on screen saying scanned pages are re-encoded.

That is the #617 case, not an edge of it. `page.dataRepresentation` gives per-page bytes and would separate a 2 MB image-backed page from a 20 KB text page whatever the text layer says — but pushing an OCR'd scan through `rasterizedPage` throws that layer away, which is the exact thing the hint promises not to do. So I don't think there's a one-liner here; it needs a decision. What it can't stay is silent: today "nothing to compress" and "compressed" look identical from outside.

**2. `targetBytes` is a stored property of a `Codable` struct, so saved profiles stop decoding.**

```swift
// MediaSupport.swift
/// Not part of a saved profile: it belongs to one job.
var targetBytes: Int64 = 0
```

A stored property is part of the synthesized `CodingKeys`. The synthesized `init(from:)` calls `decode`, not `decodeIfPresent`, and a property default does not cover a missing key. Every profile JSON written by an earlier build lacks `targetBytes` → `keyNotFound` → the profile array fails to decode, while `mediaImageSelectedProfileID` still points at one of them.

Body says "the resize, watermark, rename and profile options are unchanged". They are, until the app reads them back once.

Make it `Int64?` with `?? 0` at the two use sites, or keep it out of the struct and pass it into `processImages` the way `MediaPDFOptions` carries its own. `MediaImageProfile` and the `imageProfiles` decode aren't in the diff — I'm going off `MediaImageOptions: Codable` alone, so if the profile has a hand-written decoder this is moot.

**3. The size target only works when a PDF stays a PDF.**

`processPDFWork` branches into `exportPDFPages` before the retry loop, and `exportPDFPages` reads `options.quality` and never `targetBytes`. `combineImagesWork` doesn't read it either. So target size + JPEG/PNG/HEIC export, and target size + several images → one PDF, both encode at whatever quality happened to be stored and drop the number the user typed. The picker is shown in both cases.

**4. The image-export path ignores the chosen destination.**

`finalURL` comes from `uniqueOutputURL(for: inputURL, ...)`; `outputURL` is only used to derive the scratch directory. Point the save panel at the Desktop, get the file next to the input. And since `outputType` returns `.pdf` for any PDF input and `defaultOutputURL` names it `-compressed.pdf`, the panel offers a `.pdf` destination for a job that writes `.jpg` or `.zip`.

**5. Target size + PNG encodes six times and then writes oversize.**

`writeImage` retries by lowering quality; PNG ignores quality, so the loop produces six identical files and the last one is written over the ceiling with no error. The video path throws `.targetTooSmall` in the same situation. Also `imageResizeSection` is hidden in target mode while only quality moves — if `targetSizeRow` still shows `mediaTargetSizeHint` ("Resolution adapts to stay under the limit"), that line is false on this path. `targetSizeRow` isn't in the diff, so I didn't verify that last part. Cheapest fix is to keep the target picker off PNG.

Two small ones: `mediaStartPDF` is added to 13 locales and never read — `startButtonTitle` uses the two format strings. And `Tests/MetricsTests.swift:13547` picked up stray indentation on an unrelated line, plus two whitespace-only lines above the new editor assertions.

All of the above is read off the diff. I didn't build or run anything, and `MediaImageProfile`, `stagedOutput`, `targetSizeRow` and `imageProfiles` aren't in what I can see.

On the tool-row question: keeping it at four and recognising the PDF from the file is the right call, and the button already says which of compress/convert is about to happen. A fifth tab would need its own drop zone, its own output default and its own empty state for a document that arrives through the same panel anyway.

**Impact:** existing users lose their saved image profiles on first launch after this ships (2); anyone compressing an OCR'd scan gets a silent no-op (1); the size target is dropped on three of the paths that offer it (3, 5).
**Merge:** needs 2 fixed first — that one is not recoverable by the user.

ref #617

PathGao:
Reviewed the second commit.

**1. A PDF can still join a multi-file batch.** Body says "A PDF never joins a multi-file batch: a document is one input." `setInputs` only looks at the first URL:

```swift
let holdsPDF = urls.first.map { MediaSupport.isPDF(at: $0) } ?? false
```

The file panel now accepts `[.image, .pdf]`, so selecting `photo.png` + `scan.pdf` keeps both, `inputIsPDF` is false because the first one is the image, and the PDF goes through `processImages`. `urls.contains(where: MediaSupport.isPDF)` is the whole fix.

**2. The size target does nothing on two paths.** `exportPDFPages` never reads `options.targetBytes` — it encodes once at `options.quality`, which in target-size mode is just whatever `pdfQuality` happens to hold. Same shape on the image tab: `imageSizing == .targetSize` sets quality 0.92 plus `targetBytes`, but with PNG selected `kCGImageDestinationLossyCompressionQuality` is ignored, so `writeImage` writes the identical file 6 times and returns over target without an error. Both panels also show `mediaTargetSizeHint` ("Resolution adapts to stay under the limit") while nothing in these paths touches resolution.

**3. `MediaImageOptions` gained a stored property and it is `Codable`.** Swift's synthesized `init(from:)` does not fall back to a property's default value — a missing key throws `keyNotFound`. So `targetBytes` is not "not part of a saved profile" just because the comment says so; it lands in `CodingKeys` and old profile JSON loses the key. `MediaImageProfile` is not in this diff, so I could not confirm it embeds `MediaImageOptions` — if it does, every profile saved before this build stops decoding, which contradicts "the resize, watermark, rename and profile options are unchanged". A hand-written `CodingKeys` without `targetBytes` settles it, or keep it out of the struct and pass it in the way `MediaPDFOptions` does.

**4. A PDF → image export discards the chosen output location.** `exportPDFPages` writes to `MediaSupport.uniqueOutputURL(for: inputURL, ...)` regardless of `outputURL`. And `outputType` returns `.pdf` whenever `inputIsPDF`, so the save panel offers a `.pdf` path even with JPEG selected, then the result appears somewhere else as a `.jpg` or a `.zip`.

**5. Two changes are not in the body.** The Edit button on the image tab with `ScreenshotService.openEditor(imageAt:)`, and the `clipboardCapture` guard rewrite: `cgImage.width <= max / cgImage.height` (integer division) became `width * height <= max`, which slightly loosens the existing clipboard paste path. The body lists what changed in the image tab and says its existing behavior is untouched; neither of these is in that list.

**6. `openVideoEditor` lost its `@MainActor`.** The new function was inserted directly under the existing attribute, so it consumed it. Likely a no-op if `View` conformance already isolates the type, but it was not an intended edit.

Minor: `mediaStartPDF` is added in 12 languages and never referenced; `mediaStartConvertPDF` and `mediaStartImage` look dead now that `startButtonTitle` builds from the two format strings.

On the tool row question: keep four. A fifth tab is a second copy of the same panel plus a second set of drop-zone rules, and the discoverability problem it solves is cheaper to solve in the drop zone's own text.

One check is red (`Point out a reference that will not reach the reporter`); I can't tell from here whether that's this branch or the bot.

**Impact:** a PDF dropped alongside images is fed to the image encoder; a size target is silently ignored for PDF→image and for PNG; PDF→image writes to a path the user did not choose. If `MediaImageProfile` embeds `MediaImageOptions`, saved profiles stop loading after this update.
**Merge:** needs the multi-file PDF guard fixed first.

ref #617

FlowSync0:
Thank you — that review found two things that would have hurt people, and I
would not have caught either from my own testing. All six are fixed in the
push above; here is what I did and where I made a judgement call.

**1. Saved profiles.** You are right, and it is worse than "if it embeds
`MediaImageOptions`" — it does. A stored property lands in the synthesized
`CodingKeys`, `init(from:)` calls `decode`, and a default does not cover a
missing key, so every profile written before this build would have thrown
`keyNotFound`. The ceiling now travels beside the options instead of inside
them, as an argument on `compressImage`, `processImages` and
`combineImagesIntoPDF`. That is also what it is: a property of one job, not of
a saved preset. `MediaImageOptions` is byte-for-byte what it was.

**2. OCR'd scans.** This was the real hole and I had no good answer at first.
`page.string` was the wrong question: a ScanSnap or Acrobat scan carries an
invisible text layer, so the check skipped exactly the documents #617 is about.

It is now decided by what the page weighs — `page.dataRepresentation` — with
text only breaking the tie for a light page. A page over ~120 KB is carrying a
picture whatever its text layer says.

That has a cost you named: re-encoding an OCR'd scan throws the layer away. I
do not think there is a way to keep it without reflowing the text back in, and
a scan that stays searchable and a scan that gets smaller are genuinely
different requests. So the hint no longer promises the opposite — it now reads
"pages carrying a picture are re-encoded, which drops any invisible OCR layer",
in all thirteen languages. If you would rather it never touched a page with
text and said "nothing to compress here" instead, that is a one-line switch and
I will take the steer.

**3. The size target on the paths that ignored it.** `exportPDFPages` now gives
each page its share of the ceiling and goes through the same retry as the image
tab. `combineImagesWork` derives its page quality from the ceiling and the page
count. And PNG no longer loops six times over a format with no quality to give:
when quality is exhausted or absent, the picture is scaled down instead, which
is what the on-screen line promises. A ceiling that still cannot be met now
throws `.targetTooSmall` rather than writing oversize in silence, matching the
video path.

**4. The destination.** Honoured now; only the extension follows what is
actually written, since a page range decides between one picture and an
archive. `outputType` and the default name follow the same rule, so the panel
stops offering a `.pdf` path for a job that writes a `.jpg` or a `.zip`.

**5. A PDF in a multi-file selection.** `urls.contains(where:)`, as you said.
The body's claim was true of the drop path and false of the file panel.

**6. The two things not in the body.** The Edit button and the clipboard guard
belong to #1035 and are only here because this branch is stacked on it. The
`@MainActor` that moved is fixed there, and the ceiling rewrite
(`width * height` rather than integer division) is described in that PR now.

`mediaStartPDF` is gone. The stray whitespace in the test file is gone.

Not verified by me: I have not tried to build a PDF whose page weight sits near
the 120 KB line, so that threshold is a judgement, not a measurement.

And thank you for the answer on the tool row — four it stays.

PathGao:
Re-read the whole diff. Four things left, one of them is the same class of hole as the `page.string` one.

**1. The 120 KB weight test probably misfires on any document with embedded fonts — measure it before tuning the number.**

`page.dataRepresentation` returns a standalone one-page PDF, so it has to carry the font resources that page draws with. Those resources are shared across the document but get counted once per page here. A 30-page report with a subset CJK face, or a couple of embedded OpenType faces, can put every page over 120 KB — and then every text page in it is rasterized, which is exactly the loss the panel's own line says will not happen ("Text pages are kept as they are", now in thirteen languages). Your fixture cannot show this: two pages, one scan, and whatever fonts the text page happened to use.

The check that settles it is one print: dump `page.dataRepresentation?.count` for each page of a text-only PDF that embeds a font. If the count tracks the fonts rather than the page's own picture, this is not a threshold to raise, it is a signal that measures the wrong thing. I have not run it — I can't from here — but the threshold is the one number in this branch that decides whether #617's fix or its inverse ships.

On the steer you offered: no. Going back to "never touch a page that has text" is #617 again. The problem was never that weight is the wrong question, it is that per-page weight may not be what that call returns.

**2. A mixed selection now silently drops files.**

```swift
let holdsPDF = urls.contains(where: { MediaSupport.isPDF(at: $0) })
inputURLs = selectedTool == .imageCompressor && !holdsPDF ? urls : Array(urls.prefix(1))
```

Select `photo.png` + `scan.pdf`: `holdsPDF` is true, so it takes `prefix(1)` — the png. `inputIsPDF` is then false, the PDF is gone, the other images are gone, and nothing on screen says so. The batch survives only when no PDF is present, and the document survives only when it happens to be first. Keep the document instead of the first URL:

```swift
if let pdf = urls.first(where: { MediaSupport.isPDF(at: $0) }), selectedTool == .imageCompressor {
    inputURLs = [pdf]
} else { ... }
```

**3. Choosing an output for images → PDF opens a folder picker, and the job then fails.**

`chooseOutput()` branches on `inputURLs.count > 1` alone, so the new combine path always gets `NSOpenPanel` with `canChooseDirectories = true`. `combineImagesWork` writes a file to that URL, `installStagedOutput` calls `rename(2)` onto a directory, and it comes back EISDIR. The user sees `error.localizedDescription` from the generic catch. Nothing is destroyed — rename refuses — but every manual output choice on that path fails. Same one-line shape as the last one: `inputURLs.count > 1, MediaImageFormat.sanitized(imageFormatRaw) != .pdf`. `outputType` already returns `.pdf`, so the save panel branch works as is.

**4. The output name shown for a PDF does not follow the pickers below it.**

`defaultOutputURL` reads `pdfOutputFormat` and `pdfPageSelectionIsPlural`, but the only recompute is `onChange(of: currentImageOptions)`, which doesn't observe either. Drop a PDF, switch the format to JPEG: the row still reads `foo-compressed.pdf` while the run writes `foo-compressed.zip`. `exportPDFPages` recovers the extension, so it is the label that is wrong, not the file.

Minor: `mediaStartImage` is dead now that `startButtonTitle` builds from the two format strings. `mediaStartConvertPDF` is not — `SettingsDirectory.swift:111` still uses it as a search keyword.

What I checked in `main` rather than assumed: `chooseOutput`, `commit` / `installStagedOutput`, the `onChange(of: currentImageOptions)` recompute, `outputType`, and that `inputMatchesTool` reads `inputTypes`, so the drop path does accept PDFs now. Not verified: anything needing a run — the 120 KB question above, and whether `page.bounds(for: .mediaBox)` returns rotated bounds, which is what makes rotation come out consistent between a rasterized page and a copied one.

**Impact:** an image plus a PDF in one selection processes one image and discards the rest without a word (2); choosing where the combined PDF goes fails the job (3). If per-page weight includes shared fonts, text pages in ordinary documents get rasterized and lose their text — the opposite of what the hint promises (1).
**Merge:** needs 2 and 3 fixed first, and 1 measured.

ref #617

PathGao:
Closing with the stack; the reasons for all three are on #1035.

Issue #734, opened 2026-08-21
Improve smooth mouse scrolling to match Mos


iva-zhu:
### Problem or Motivation

Vorssaint already provides useful smooth scrolling, but when using a regular external mouse, scrolling still feels less fluid and natural than in Mos.

Individual wheel steps are noticeable, and acceleration and deceleration can feel abrupt or unpredictable. As a result, I still need to use Mos alongside Vorssaint to get a more trackpad-like scrolling experience.

The main issue is not functionality, but the scrolling feel: it should be smooth, responsive and predictable without visible jumps or unnatural deceleration.

### Proposed Solution

Add a smoother scrolling mode inspired by the feel of Mos, without copying its implementation or code.

The scrolling behavior should ideally:

- feel more continuous and trackpad-like;
- reduce the jumps between individual mouse-wheel steps;
- provide smoother and more predictable acceleration, momentum and
- deceleration;
- remain precise and responsive during short scrolling movements;
- offer configurable parameters such as step size, speed/gain and animation
- duration;
- optionally provide a “simulate trackpad” mode;
- support continuous and high-resolution wheel events;
- preserve horizontal scrolling with Shift;
- support independent inversion settings for each axis.

The goal is to make Vorssaint feel comparable to Mos while keeping its scrolling responsive, configurable and compatible with existing input methods.

### Feature Area

Mouse and keyboard

### Mockups / References

Mos reference:
https://github.com/Caldis/Mos

pasindujr:
I notice the same. Mos has a smoother scrolling effect, and better sensitivity to individual mouse steps. It would be amazing if Vorssaint could get those features in.

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

Blocked on PR #626 ("Improve Mos-like smooth scrolling feel and settings"), which is the branch covering scroll feel and acceleration. It is open but currently conflicting and needs a rebase onto the rewritten main — the reported 125 files / +8815 is a history-rewrite artifact rather than its real size.

No separate work is needed on this issue in the meantime; it tracks that PR.

One note recorded for whoever picks #626 up: review explicitly recommended not closing it, and said that at minimum commit `379214b8` (merging the two separate "ignored apps" lists together with migration logic) is worth keeping even if the rest is redone.

celltower:
Quick status update: the live PR for this is now #1040 (successor to the old #626).
[...]

PR #626, opened 2026-08-13
Improve Mos-like smooth scrolling feel and settings

celltower:
## Summary
- Align smooth mouse scrolling with Mos-style feel: Step / Speed / Duration controls, DisplayLink coasting, annotated-session delivery, and safer Accessibility fallbacks.
- Add a tunable **Scroll acceleration** slider (0.00–1.00, default 0.00) that blends raw wheel movement with macOS’ accelerated point deltas for flexible fast scrolling.
- Update shipping defaults to **Step 39.69**, **Speed 3.00**, **Duration 3.00**, and unify the two scrolling “Apps to leave alone” lists into one shared exception list for smooth scrolling and reverse scrolling (with migration of existing entries).
- Keep Settings responsive (continuous sliders + editable values), localize the new controls, and expand unit coverage for the feel math, acceleration blend, and exception migration.

## Test plan
- [x] Enable Smooth scrolling; confirm trackpad/Magic Mouse remain unaffected.
- [x] Compare feel against Mos with matching Step / Speed / Duration (and try acceleration around 0.50).
- [x] Reset to defaults → Step 39.69, Speed 3.00, Duration 3.00, Acceleration 0.00.
- [x] Add an app under the single scrolling “Apps to leave alone” list → both smooth scrolling and reverse scrolling stand down for that app.
- [x] Revoke Accessibility while scrolling → wheel still works (unsmoothed), tap tears down cleanly.
- [x] `./build.sh --test`

PR #1040, opened 2026-08-26
feat(mouse): Improve Mos-like smooth scrolling feel and settings

celltower:
## Summary

Smooth mouse scrolling already exists in Vorssaint; this tunes how it feels and how its Settings are laid out, rather than adding a new product surface.

- Mos-aligned feel controls: Step / Speed / Duration / Acceleration, with continuous editable sliders and Reset.
- Delivery and coasting follow the same shape as Mos (annotated-session path, usable delta priority, DisplayLink frames) while keeping a live `AXIsProcessTrusted()` check on the tap path so a revoked Accessibility grant tears the tap down without waiting on the permissions poller.
- Dedicated userInteractive run-loop threads for the smooth-scroll and scroll-inverter taps (plus DisplayLink coasting) stay inside those existing features’ event pipelines — no new user-facing surface beyond Preferences and the taps those features already own.
- One shared “apps to leave alone” list for smooth scrolling and reverse scrolling. **Upgrade note:** migrating unions the two legacy lists into that shared key, so an app that was excepted from only one scrolling behavior becomes excepted from both; the old split is not recoverable. New installs simply use the single list.
- `smoothScrollEnabled` stays **off by default**, so an upgrade does not turn the feature on for people who never touched the toggle.

Built only on the existing Smooth scrolling / Scroll inverter / mouse-exception stack already in tree — no new third-party dependency. Supersedes #626 with a fresh branch off current `main` so the review diff is only this work.

## Verification

Tested on a MacBook Pro (MacBookPro18,3), Apple Silicon, macOS 27.0 (build 26A5416b), single built-in display. Local contributor build (ad-hoc signing); no Developer ID / notarized install on this machine.

```sh
./build/Vorssaint --selftest
./build.sh --test
```

`./build.sh --test` exercises the feel math, Int→Double preference rewrite, lazy exception bundle-id path, and exception-list migration. New localization keys are present for every supported language; FeatureHub `descSmoothScroll` is updated in every language block.

Could not re-confirm a clean `./build/Vorssaint --selftest` → SELFTEST OK in the last agent session (power assertion / SMC / disk counters unavailable there). Could not exercise notarized Gatekeeper installs, or multi-display / Spaces layouts beyond this laptop. Maintainer hardware will be the authoritative pass before merge.

## Anything else

Refs #734


PathGao:
please use our pr template thanks

celltower:
> please use our pr template thanks

Updated the description please check it out

PathGao:
Read the visible diff. Four findings, then the scope question.

**1. `rewriteSmoothScrollNumber` never fires.** `value is Double` goes through Swift's NSNumber bridge, which does a value-preserving cast rather than a type-identity check — an Int-backed `40` satisfies `is Double`, the guard returns, and nothing gets rewritten. The migration exists precisely for that case, so it's a no-op for every upgrader. `sanitize` is idempotent, so the guard buys nothing anyway; delete it:

```swift
guard let number = defaults.object(forKey: key) as? NSNumber else { return }
defaults.set(sanitize(number.doubleValue), forKey: key)
```

The doc comment above `migrateMouseScrollingExceptions` also describes `migrateSmoothScrollNumbers` — it's attached to the wrong function.

**2. The eager bundle-id lookup defeats the fast path you documented.** `excludesPointerTarget` computes `targetBundleID` before calling `excludes`, and `bundleID(forProcessID:)` is `NSRunningApplication(processIdentifier:)`. That runs on the tap thread for every wheel event once any list is non-empty — including the case where the pid set was about to answer without it. Your own comment says "Event taps never wait for that round-trip: they first consult the event's target process id (synchronously available)". The pid is synchronous, the bundle id isn't. Pass the pid into `excludes` and resolve it lazily, after both set checks miss.

**3. Merging the two exception lists rewrites choices users already made.** The comment you replaced said the opposite deliberately: "excepting an app from the wheel's glide must not also silence the side buttons there", "Every feature keeps its own list, so leaving an app out of one never changes what the others do there", both tied to #358. `migrateMouseScrollingExceptions` unions the lists, so someone who excepted an app from smooth scrolling only now has reverse scrolling off there too, and the old split is unrecoverable. The body describes the shared list but not that it edits existing settings. Say it in the body, or only merge when one of the two lists is empty.

**4. `descSmoothScroll` changed in the English and German blocks only.** If `FeatureHubStrings.swift` carries the other languages in the same file, they still say "Smooth, animated mouse scrolling". I see two hunks and that file isn't fully in my diff, so flag not verdict.

**Scope.** Body: "no new permanent runtime surface beyond Preferences and the tap path those features already own." The diff adds two dedicated `userInteractive` threads with their own run loops, a CVDisplayLink, and a pointer-resolution queue, and rewrites `ScrollInverter`'s whole tap lifecycle — a different feature from the one in the title. Both taps also switch from `Permissions.shared.accessibility` to live `AXIsProcessTrusted()`, which the body doesn't mention. Separately, the three `layoutSubtreeIfNeeded` removals in `AppDelegate` (popover, settings window, intro window) have nothing to do with scrolling; `centerIntroWindow` now reads `fittingSize` without forcing a pass, and if the pass hasn't run the `> 0` guard bails and the window keeps its default size. Unless those are deliberate fixes for a bug I can't see from the diff, they belong in their own PR.

Not verified: the diff is truncated at 80KB, so `SmoothScrollSupport.swift`, the Settings views and the tests aren't in what I read — the idempotency claim in (1) and anything about the Reset button is inference. No hardware run on my side either.

**Impact:** upgraders keep the stuck sliders (1); every wheel tick with a non-empty list pays a LaunchServices lookup on the tap thread (2); anyone who excepted an app from one scroll feature silently gets it excepted from both (3).
**Merge:** needs the (1) guard removed and the three AppDelegate layout removals split out first.

ref #734

celltower:
Addressed in the latest push (still one PR):

1. `rewriteSmoothScrollNumber` now always goes through `NSNumber` → `doubleValue` (no `is Double` guard). Doc comment moved onto `migrateSmoothScrollNumbers`.
2. `excludes` takes a lazy `targetBundleID` closure and only resolves it after both pid-set checks miss.
3. Shared-list migration kept; Summary now states that upgrading unions the two legacy lists and the old split is not recoverable.
4. `descSmoothScroll` updated in every language block in `FeatureHubStrings.swift`.
5. Scope: dropped the unrelated `AppDelegate` `layoutSubtreeIfNeeded` removals. Body also notes live `AXIsProcessTrusted()` on the tap path and that the dedicated tap threads / DisplayLink stay inside the existing scroll features.

PathGao:
1 through 4 and the scope trim check out. The `is Double` guard is gone, the doc comment sits on `migrateSmoothScrollNumbers`, all 13 `descSmoothScroll` blocks in `FeatureHubStrings.swift` are updated (counted against `main`), all 13 `Strings` inits carry the new keys, and `AppDelegate.swift` is out of the diff. Three things left, all in `MouseAppExceptions`.

**1. The click features lost synchronous pointer resolution.** `pointerBundleID(at:snapshot:)` no longer resolves on a miss — it pushes the WindowServer query onto `pointerResolutionQueue` and returns `frontmostApplication` for this event. Defensible for a wheel that ticks thirty times a second, not for a click that happens once. `MiddleClickService` calls `excludesPointerTarget(.middleClick, …)` at two sites with no pid, `MouseNavigationService` and `MouseButtonShortcutService` call `excludesActionTarget` the same way, and none of those three scopes ever calls `setSourceTracking` — `sourceProcessIDs` is empty for them, so the pointer answer is the entire judgement. Put the pointer over an excepted app's window while another app is frontmost and three-finger click: cache miss, answer is the frontmost app, the exception doesn't apply, the click gets transformed. The second click is right. That is #358's bug back on the first click, and the body scopes this change to "those existing features' event pipelines" — the click taps aren't in that set.

**2. `finishPointerResolution` holds `lock` across `NSRunningApplication` and `NSWorkspace`.** Those are exactly the round-trip this rework exists to keep off the tap thread, and the tap thread takes that same lock in `makeSnapshot()` on every wheel event. So a wheel tick can block behind a main-thread LaunchServices lookup — the stall that produces `tapDisabledByTimeout`, which is what the new re-arm counter in `ScrollInverter` is there to absorb. Resolve the bundle id before taking the lock, then store the string.

**3. The pointer lookup is still eager.** `excludesPointerTarget` computes `pointerBundleID(at:snapshot:)` before it calls `excludes`, so the pid sets never get to answer first; on a cache miss that's a `frontmostApplication` call plus a lock acquisition per wheel event. `targetBundleID` is a closure now — give `pointerBundleID` the same treatment.

Not verified: the diff is still truncated at 80KB, so `SmoothScrollSupport.swift`, the Settings views and the tests aren't in what I read. I can't see whether the shared list renders as one row or as two identical rows — `MouseExceptionScope` still has both scroll cases and `MouseExceptionStrings.caption` now returns the concatenated caption for each of them. No hardware run on my side either.

**Impact:** the first click on a non-frontmost window of an excepted app is no longer excepted, across middle click, side-button navigation and button shortcuts (1); wheel events can stall on a main-thread lookup through the shared lock (2).
**Merge:** needs the click-path regression in (1) and the lock held across LaunchServices in (2) fixed first.

ref #734

celltower:
Addressed all three remaining `MouseAppExceptions` findings in `11535bc`:

1. Click features now use synchronous pointer resolution by default, so the first click over an excepted non-frontmost app is handled correctly. Only the high-frequency smooth-scroll and scroll-inverter paths opt into cached asynchronous resolution.
2. `finishPointerResolution` resolves `NSRunningApplication` / `NSWorkspace` before taking the shared snapshot lock.
3. Both target and pointer bundle-id lookups are lazy and run only after the source/target pid-set checks miss.

I also kept side-wheel shortcut preflight on the PID-first asynchronous path, serialized source-tracking changes onto the main thread, removed the remaining unrelated `AppSwitcher` layout change, and tightened DisplayLink generation/start-stop synchronization found during the follow-up audit.

The branch is rebased onto current `main`. Verification: `./build.sh --test` → `TESTS OK (8785 checks)`; working tree is clean.

PathGao:
1 through 3 check out. `excludesPointerTarget` defaults to `.synchronous` and `excludesActionTarget` forwards without a mode; `MiddleClickService` and `MouseNavigationService` are absent from the diff entirely (paths sorting before `MouseButtons/` are present, so they're untouched) and inherit the synchronous default. `finishPointerResolution` resolves `NSRunningApplication` / `NSWorkspace` before it takes `lock`. `excludes` takes both bundle ids as closures and calls neither until the two pid-set checks miss.

One left, in the one place you kept asynchronous.

**The side-wheel preflight and the handler disagree, and the preflight wins.** `claimsSideWheel` now asks with `.cachedAsynchronous`. `handleSideWheel` is unchanged — its only hunk in this file is at `claimsSideWheel`, so `MouseButtonShortcutService.swift:307` still reads `if !wasPreflighted, excludesActionTarget(…)`. A successful preflight suppresses the synchronous check entirely. Cold cache, pointer over an excepted app's window, a different app frontmost: the preflight resolves to the frontmost app, decides the exception doesn't apply, returns true and stamps the timestamp; `handleSideWheel` sees `wasPreflighted` and fires the shortcut inside the excepted app. The second tilt is right. That's #358's first-event bug kept alive in the one path that can't repair itself, because the accurate check is there and gets skipped.

The latency argument doesn't apply here: `claimsSideWheel` sits behind `hasActiveSideWheelInterest` and `sideWheelInput(for:) != nil` (`SmoothScrollService.swift:151-164` on `main`), so it runs on an actual tilt carrying a mapping, not per wheel tick — and `main` already resolved that same call synchronously. Pass `.synchronous`.

Not verified:

- **The tap ordering reversed and I can't see the other half.** `ScrollInverter` now says Smooth Scroll runs later at the annotated-session tap. On `main` both are `.cghidEventTap`, and `SmoothScrollService.swift:187-202` applies the inverter's flip itself *because* it swallowed the tick first. If that `invertHere` block survived the move, the HID tap inverts and the session tap inverts again — inversion looks dead whenever both features are on. `SmoothScrollService.swift` is past the 80KB cut.
- **Whether the shared list renders as one row.** `SettingsView.swift:704` and `:755` are two separate `MouseExceptionsList` sections on `main`, and both scopes now resolve to `mouseScrollingExceptions`. The new `mouseScrollingSection` string suggests you merged them, but `UI/` is past the truncation.
- The re-arm limiter (`maxTapTimeoutsPerWindow` / `handleTapDisabled`), the Settings views and the tests are all past the cut. No hardware run on my side.

**Impact:** the first side-wheel tilt over an excepted app's window fires its mapped shortcut when a different app is frontmost.
**Merge:** needs `claimsSideWheel` back on synchronous pointer resolution first.

ref #734

celltower:
The remaining side-wheel finding is fixed in `55f15fb`.

`claimsSideWheel` now uses the same `excludesActionTarget` check as `handleSideWheel` (synchronous by default). A successful preflight can no longer stamp a cold-cache miss and skip the accurate first-tilt exception.

On the two unverified items:

- **Invert order:** Smooth Scroll sits later at the annotated-session tap and passes `invertVertical/Horizontal: 1` (identity multiplier). Reverse scrolling is applied only by the HID inverter; the session tap does not invert again.
- **Shared list UI:** one `MouseExceptionsList` is shown. If Smooth Scroll is available it lives in that section; otherwise it appears under reverse scrolling. Both scopes read the same `mouseScrollingExceptions` key.

Verification: `./build.sh --test` → `TESTS OK (8786 checks)`.

PathGao:
`claimsSideWheel` now calls `excludesActionTarget` with no mode, and `excludesActionTarget` forwards to `excludesPointerTarget`'s `.synchronous` default — the same path `handleSideWheel` takes, so a preflight can no longer stamp an answer the handler is then told to trust. That closes it.

Taking your two answers on the invert order and the shared-list UI as given: `SmoothScrollService.swift` and `UI/` are both past the 80KB diff cut, so I read neither. No hardware run on my side either.

One thing left and it isn't code. GitHub reports the branch as `CONFLICTING` against `main` — the rebase you did on the 26th has been overtaken. Also worth flagging: this PR shows zero check runs, not failing ones. If that's the first-time-contributor approval gate it's the maintainer's button, not yours, but nobody can call CI green until it runs.

**Impact:** none — the three findings from the last round are all resolved in the diff I can see.
**Merge:** needs the conflict with `main` resolved first.

ref #734

celltower:
Rebased onto current `main`. The conflict was real: `main` now includes #1023 (path identities for programs that are not apps), which overlapped our exception lookup. Both sides are kept — shared scroll list / sync-vs-async pointer resolution / lazy pid-first checks, plus the #1023 identity path.

Verification: `./build.sh --test` → `TESTS OK (8874 checks)`. Working tree is clean. Check runs still need the first-time-contributor approval on your side before CI can go green.

PathGao:
Rebase checks out on the #1023 side. Every identity resolution in the merged `MouseAppExceptions` goes through `Self.identity(for:)` — `identity(forProcessID:)`, both closures in `synchronousPointerIdentity`, `finishPointerResolution`, and the frontmost fallback in `excludesActionTarget` — so a path identity still matches after the merge. GitHub reports `MERGEABLE` now.

One thing the merge makes worse, and it's a correction to what I told you in round two. I accepted `targetBundleID` becoming a closure as the fix for the eager lookup. It isn't one. `excludes` calls it whenever both pid-set checks miss, and the common case *is* a miss: `sourceProcessIDs[scope]` only holds the pids of apps that are on the list, so scrolling anywhere else falls straight through to `targetBundleID()`. That's `NSRunningApplication(processIdentifier:)` per wheel event, uncached, on the tap thread — the same call from finding (2) in the first round.

After #1023 it costs more than a LaunchServices lookup. `identity(bundleID:executablePath:)` is `bundleID ?? executablePathIdentity(executablePath())`, and `executablePathIdentity` runs `resolvingSymlinksInPath()`. Scroll inside a program with no bundle id — the Java process #1009 is about — and every wheel event does a filesystem round trip on the tap thread. `MouseAppExceptionSupport.swift:132-135` on `main` says the path "is not even looked at" for ordinary apps "because this runs under the event taps"; `main` only ever reached `identity()` through the 0.5s-cached `pointerIdentity`. That is the stall class the new re-arm counter exists to absorb.

The fix keeps the coverage your comment claims. The pid *set* check is what covers "the first wheel tick before the async pointer cache warms" — resolving the bundle id only adds the window between an app launching and the `runningApplications` KVO rebuild. Pass `targetBundleID: { nil }` on `.cachedAsynchronous` and leave the synchronous click path alone.

Not verified:

- **Whether `.eventTargetUnixProcessID` is nonzero at these taps.** At `ScrollInverter`'s `.cghidEventTap` events are pre-routing, so it may well be 0 — which would make the whole target-pid path inert there rather than expensive. At Smooth Scroll's annotated-session tap it should be set. `SmoothScrollService.swift` is still past the 80KB cut, so I can't see which side it lands on. If it's 0 everywhere, the finding above is dead and the target-pid argument is dead with it.
- The Settings views, the tests and the re-arm limiter are all past the cut. Taking your answers on the invert order and the single exceptions row as given. No hardware run on my side.

**Impact:** with any scroll exception set, every wheel event in a non-excepted app pays an uncached `NSRunningApplication` lookup on the tap thread, plus a `resolvingSymlinksInPath()` filesystem call when the scrolled program has no bundle id — conditional on the target pid being nonzero at that tap.
**Merge:** needs `targetBundleID` off the per-event wheel path first.

ref #734

celltower:
Addressed in `0e75622` (rebased onto current `main`).

**Wheel target identity:** `.cachedAsynchronous` now passes `targetIdentity: nil`. After the pid-set checks miss, the wheel path only consults the pointer cache; a cache miss schedules the WindowServer query off the tap and returns no identity for that event — no `NSRunningApplication` and no `resolvingSymlinksInPath()` on the tap thread. Synchronous click / side-wheel paths keep their target + pointer resolvers.

Also cleaned up while verifying:

- DisplayLink start/stop no longer runs under `engineLock` (serialized on a control queue)
- `postToPid` clears stale line/fixed deltas and sets `isContinuous` as an integer
- Shared exception list shows next to the active scrolling feature
- Legacy exception keys dropped from `registeredDefaults` (migration-only)
- Dead unused glide helpers removed

Verification on this machine:

```sh
./build.sh --test          # TESTS OK (8875 checks)
./build.sh --dev --install # local Developer app
./build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest  # SELFTEST OK
```

Still no notarized/Developer ID install here; check runs still need the first-time-contributor approval on your side.

PathGao:
`.cachedAsynchronous` passes `targetIdentity: nil` now, so the wheel path resolves no event-target identity. That closes the last round's finding, and it closes it whichever way the target pid lands.

Which is also the problem — that was the wheel's only synchronous cover, so the pid question now decides correctness instead of cost.

**On a cache miss the wheel path has no answer, and misses recur twice a second.** `cachedPointerIdentity` returns `nil` on a miss, `excludes` reads `nil` as "not excepted", and `cacheHolds` expires at `resolveLifetime = 0.5` (`MouseAppExceptionSupport.swift:88`) whether or not the pointer moved — a hit doesn't restamp `cachedAt`. So a continuous scroll inside an excepted app misses about twice a second, and each miss leaks every wheel event arriving before `finishPointerResolution` gets its main-thread hop. On `main` that same miss cost one synchronous 190 µs lookup and was always right (`ScrollInverter.swift:136-139`).

What's left to catch those events is the pid sets, and at `ScrollInverter`'s `.cghidEventTap` neither field looks usable: hardware wheels carry no source pid (the comment you replaced said exactly that), and the target pid is pre-routing there. `MiddleClickService.swift:453` on `main` guards `targetPID > 0`, so 0 is a value that field really takes. If it is 0 at the HID tap, reverse scrolling in an excepted app flips direction for one tick roughly twice a second, for as long as the scroll continues. Smooth Scroll's annotated-session tap should carry a real target pid and is unaffected.

The lazy fix is what `main` already shipped: give the wheel `.synchronous` as well and drop the mode. The measurement is in your own file — 190 µs per lookup, twice a second, against a tap budget in milliseconds. That takes `PointerResolutionMode`, `pointerResolutionQueue`, `latestRequestedPoint`, `activeResolution`, `resolutionGeneration` and `finishPointerResolution` with it. Unless the async path exists because you measured `tapDisabledByTimeout` coming from this lookup specifically — the new re-arm limiter says you saw timeouts from somewhere, and I can't tell from the diff where.

Not verified:

- **The two wheel call sites.** The diff cuts mid-line at `let targetProcessID = event` in `ScrollInverter.swift`, and `SmoothScrollService.swift` never appears at all. I'm taking "only the high-frequency smooth-scroll and scroll-inverter paths opt into cached asynchronous resolution" from your comment, not from code I read.
- **Whether `.eventTargetUnixProcessID` is nonzero at a `.cghidEventTap`.** Needs hardware. The whole finding hangs on it: nonzero and the pid set answers, finding dead.
- The DisplayLink control queue, `postToPid`, the single exceptions row, the removed glide helpers and the tests are all past the 80KB cut. No hardware run on my side.

**Impact:** with reverse scrolling excepted in an app, about twice a second one wheel tick there gets inverted anyway, for as long as you keep scrolling — conditional on the target pid being 0 at the HID tap.
**Merge:** needs the wheel path back on synchronous pointer resolution first.

ref #734

celltower:
Addressed in `eae9282` (rebased onto current `main`).

**Wheel path is synchronous again**, matching `main`:
- dropped `PointerResolutionMode`, `pointerResolutionQueue`, `latestRequestedPoint`, `activeResolution`, `resolutionGeneration`, and `finishPointerResolution`
- a cache miss fills on the tap thread (~190 µs) and returns the real pointer identity — no more `nil`/`not excepted` gap while an async lookup catches up
- Smooth Scroll and Scroll Inverter both use that same path; click / side-wheel were already synchronous

So excepted apps keep reverse scrolling / glide stood down for every wheel tick, including at the HID inverter where target pid can be 0.

Verification: `./build.sh --test` → `TESTS OK (8925 checks)`; local `./build.sh --dev --install` exercised on this machine.

PathGao:
The revert lands. `pointerIdentity(at:)` fills a miss on the caller thread and returns the real identity, `excludesPointerTarget` has no mode parameter left, and `PointerResolutionMode` / `pointerResolutionQueue` / `latestRequestedPoint` / `activeResolution` / `resolutionGeneration` / `finishPointerResolution` are all gone from the file. The `nil`-reads-as-not-excepted window is closed, and it's closed whichever way the target pid lands at the HID tap.

**The async design left surface behind, and one piece of it is a second copy of the pointer answer.** `MouseAppExceptionSupport.pointerIdentity(in:at:ownProcessID:identityForProcess:frontmostIdentity:)` is new in this PR and has no caller anywhere in the diff — `MouseAppExceptions.pointerIdentity(at:)` reimplements it inline instead, and has to, because it also needs `window?.frame` for the cache region and the helper only returns the identity.

The two copies already disagree. The helper ends `identityForProcess(window.processID) ?? frontmostIdentity()`. Production is unchanged from `main` (`MouseAppExceptions.swift:218-219`):

```swift
let app = window.map { NSRunningApplication(processIdentifier: $0.processID) }
    ?? NSWorkspace.shared.frontmostApplication
```

That `??` is applied to an `NSRunningApplication??`, so a window whose pid has no `NSRunningApplication` collapses to `nil` — no frontmost fallback. The helper falls back. A test written against the helper asserts a branch production doesn't take. Its doc comment ("Click features use this path because one missed click cannot be repaired by a cache warmed for the next event") describes the design that just got deleted.

Two smaller leftovers in the same layer: `excludes(…targetIdentity:)` is passed `nil` at its only visible caller, and `LookupSnapshot.allEmpty` is filled by `makeSnapshot()` and read by nobody — `excludesPointerTarget` checks `snapshot.lookups[scope]` instead. Unless something past the 80KB cut uses them, deleting the helper and both of those puts the pointer answer back in one place.

Not verified:

- **Whether anything outside `MouseExceptions/` calls the new helper.** `MiddleClickService` and `MouseNavigationService` aren't in the diff at all, and the tests are past the cut. If the only caller is the test suite, the above stands; if there's a production one I can't see, it stands harder.
- `SmoothScrollService.swift` and `UI/` are still past the cut. Taking your answers on the invert order (identity multipliers at the session tap) and the single exceptions row as given. The DisplayLink control queue, `postToPid` and the re-arm limiter are past it too.
- Check runs are still at zero, not failing. That's the first-time-contributor approval gate and it's the maintainer's button, but nobody can call CI green until it runs.
- No hardware run on my side.

**Impact:** none — the duplicate has no production caller I can see, so nothing misbehaves today.
**Merge:** needs the unused pointer-resolution helper and the two dead parameters removed first.

ref #734

celltower:
Addressed in `75091a4` (rebased onto current `main`).

Removed the leftovers from the async design:
- deleted unused `MouseAppExceptionSupport.pointerIdentity(in:…)` (production already fills the cache inline)
- dropped the always-`nil` `targetIdentity` parameter from `excludes`
- dropped unread `LookupSnapshot.allEmpty`

Pointer resolution now lives in one place again. Verification: `./build.sh --test` → `TESTS OK (8924 checks)`; local Developer build retested.

PathGao:
The three leftovers are gone: `MouseAppExceptionSupport.pointerIdentity(in:…)` is out of the file, `excludes` takes `scope / snapshot / sourceProcessID / targetProcessID / pointerIdentity` and nothing else, and `LookupSnapshot` is down to `lookups` + `sourceProcessIDs`. Pointer resolution is in one place.

One new thing, and it comes from the item you listed as cleanup rather than as a behavior change.

**"Legacy exception keys dropped from `registeredDefaults` (migration-only)" also drops them from the backup format.** `SettingsBackupSupport.exportKeys()` is `Defaults.registeredDefaults.keys` ∪ `AppFeature.availabilityDefaults.keys` ∪ `unregisteredPreferenceKeys` − `machineStateKeys` (`SettingsBackupSupport.swift:18-24`). `smoothScrollExceptions` and `scrollInverterExceptions` reach that set only through `registeredDefaults` — I checked `main`, they are in neither `unregisteredPreferenceKeys` (`:28-69`) nor `machineStateKeys` (`:75-126`). `sanitizedSettings` keeps only what `allowed.contains($0.key)` admits (`:155-156`). So a backup written by any build shipping today carries both lists, and this build throws both away on import.

The rest of the restore doesn't recover them. `applyAndRelaunch` clears `exportKeys()`, writes the file's values, then relaunches (`SettingsBackup.swift:88-99`). An old file has no `mouseScrollingExceptions`, so that key falls back to the registered `[]`, and `migrateMouseScrollingExceptions` on the next launch merges `[] + [] + []`. Concretely: except an app from reverse scrolling on 3.3.x, write a settings backup, restore it on this build — the list is empty and nothing says so. Upgrading in place keeps it. The upgrade note in the body covers the in-place path and only that path.

Laziest fix is putting the two keys back in `registeredDefaults` as `[String]()`. They stay in `exportKeys`, restore writes them, and the launch migration folds them into the shared key and removes them exactly as it already does. They get exported again too, always empty on a build that has migrated once.

Two smaller ones in the same area:

- `SettingsBackupSupport.swift:186` — "All five keys register `[String]()`" is the stated reason `restoredExceptionList` is safe on a backup written before those keys existed. Five scopes now share four keys, and the fix above is what decides whether that sentence stays true for the legacy pair.
- `Tests/MetricsTests.swift:16764` on `main` feeds `smoothScrollExceptions: [1]` and asserts it is dropped as a wrong element type. On this build it is dropped for being an unknown key instead, so the assertion passes without exercising `valueLooksRight` at all.

Not verified:

- **Neither backup file is in the diff.** `Core/SettingsBackupSupport.swift` would sort between `Core/MouseExceptionStrings.swift` and `Services/MouseButtons/`, and `Services/SettingsBackup.swift` between `Services/ScrollInverter.swift` and `Services/SmoothScrollService.swift`; both slots are empty, so I read them on `main` and am treating them as untouched. If you changed either past the 80KB cut, the above is wrong and say so.
- Tests are past the cut too, so the `MetricsTests` note may already be handled.
- `LookupSnapshot.empty` has no caller in the diff I can read. Probably a fixture past the cut, not a finding.
- Taking your answers on the invert order and the single exceptions row as given. Check runs are still at zero, not failing. No hardware run on my side.

**Impact:** restoring a settings backup written by a current release drops the smooth-scroll and reverse-scroll exception lists; the in-place upgrade path keeps them.
**Merge:** needs the two legacy exception keys reachable from `exportKeys()` first.

ref #734

celltower:
Addressed in `059acd0` (rebased onto current `main`).

**Legacy backup keys:** `smoothScrollExceptions` and `scrollInverterExceptions` are back in `registeredDefaults` as empty `[String]()`, so they stay in `exportKeys()` / `sanitizedSettings`. An older backup can restore those lists again; launch migration then folds them into `mouseScrollingExceptions` and clears the legacy keys.

Also updated the `restoredExceptionList` comment and added coverage for export/restore → migration plus the `valueLooksRight` wrong-element-type path.

Verification: `./build.sh --test` → `TESTS OK (8936 checks)`.

PathGao:
Can't reach #1046 from here. The runner hands me one object per invocation — this PR's body, its comments, its truncated diff — plus `main`'s `Sources/` tree. #1046's diff, body and metadata are in none of that, and I have no network or `gh` in this environment, so there is nothing for me to read.

Trigger the run against #1046 directly and I'll review it there.

Nothing changed on #1040 in the meantime: `059acd0` is unreviewed on my side, so the labels stay as they are.

celltower:
#1046 isn't part of this PR — nothing for us to pull in there.

`059acd0` is the backup-key change from the last finding on this thread. Ping us if anything's still off.

PathGao:
`059acd0` closes the last finding. Both legacy keys are back in `registeredDefaults` as `[String]()`, so `exportKeys()` carries them, `sanitizedSettings` admits them, `applyAndRelaunch` writes them, and `migrateMouseScrollingExceptions` folds them into the shared key on the next launch. I traced the old-backup round trip against `main`'s `applyAndRelaunch`: clear → file's legacy lists land → carried paths go back into `mouseScrollingExceptions` → relaunch merges all three. The lists survive. Localization is complete too — 13 `descSmoothScroll` blocks and 13 copies of the 13 new `Strings` fields, which is every language.

**The same edit takes those two keys out of the path filter.** `portableMouseExceptions` iterates `Set(MouseExceptionScope.allCases.map(\.defaultsKey))` (`SettingsBackupSupport.swift:214`). After this PR no scope names the legacy pair, so that set is four keys — and the two are now inside `exportKeys()` and outside the per-value refusal. On import: `allowed.contains` passes, `valueLooksRight` passes (`[String]` is `[String]`), `portableMouseExceptions` skips them, `applyAndRelaunch` writes them verbatim, launch migration merges whatever they hold into `mouseScrollingExceptions`.

The comment you rewrote one hunk earlier is the one that names the gap: "the four live scope keys plus the two legacy scroll lists kept for older backups." Six keys get cleared, four get filtered. And `:209` says the `allCases` drive covers "a scope that renames its key, or two scopes that come to share one" — true, but not a key that stops being named by any scope, which is what this PR created.

No shipping build writes a path into those keys (their `allCases` still names both, so their export strips paths), so what gets through is a hand-edited or foreign file — which is exactly the case `sanitizedSettings` says it exists for, and the one `valueLooksRight` can't catch, since a path is a perfectly good `[String]` element. Restore lands a row carrying someone else's home directory in the live scroll list.

```swift
for key in Set(MouseExceptionScope.allCases.map(\.defaultsKey))
    .union([DefaultsKey.smoothScrollExceptions, DefaultsKey.scrollInverterExceptions]) {
```

One consequence worth stating rather than fixing: because the migration clears the legacy keys at every launch, this build exports them as `[]` forever. A pre-upgrade release restoring that file drops `mouseScrollingExceptions` as an unknown key and reads `[]` from the two it knows, so every scroll exception is gone with no error — `formatVersion` is still 1, so the file looks restorable. Mirror image of last round's finding, in the downgrade direction. Your upgrade note covers the in-place path and only that path. If you want it closed, mirroring the merged list into the two legacy keys inside the same function does both jobs at once; if you'd rather not support downgrade restores, say so and I'll stop raising it.

Not verified:

- **`Services/SettingsBackup.swift` isn't in the diff.** It sorts between `Services/ScrollInverter.swift` and `Services/SmoothScrollService.swift` and that slot is empty, so I read `applyAndRelaunch` on `main` and am treating it as untouched. `Core/SettingsBackupSupport.swift` is in the diff but only the doc-comment hunk — `portableMouseExceptions` is unchanged.
- `SmoothScrollService.swift` and `UI/` are still past the 80KB cut. Taking your answers on the invert order and the single exceptions row as given.
- Tests and the re-arm limiter are past the cut too.
- Check runs are still at zero, not failing — first-time-contributor approval, the maintainer's button. Nobody can call CI green until it runs.
- No hardware run on my side.

**Impact:** restoring an edited or foreign backup can plant an unfiltered executable path in the shared scroll exception list; ordinary export/restore between current builds is unaffected.
**Merge:** needs the two legacy keys inside `portableMouseExceptions` first.

ref #734

celltower:
0f3169f puts the two legacy keys into portableMouseExceptions, and migration mirrors the shared list back onto them so a downgrade restore still sees the exceptions.

celltower:
Hey @PathGao, rebased onto current main after the SettingsDirectory conflict from the search change. Could you take another look when you have a moment and let me know if everything looks okay?

PathGao:
`portableMouseExceptions` unions both legacy keys now — the planted-path hole is closed. The downgrade mirror is what bites.

**Mirroring the merged list onto the legacy keys makes removal impossible.** `migrateMouseScrollingExceptions` merges `current + smooth + direction`, and `reload()` calls it first thing. `MouseAppExceptions.remove` writes only `scope.defaultsKey` — the shared key — then reloads, while the legacy pair still holds the pre-removal list from the last mirror write. The entry comes straight back. Take an app off "apps to leave alone" and the row reappears; nothing ever removes it.

Make the mirror write-only — once the shared key exists it is the answer:

```swift
let stored = defaults.object(forKey: DefaultsKey.mouseScrollingExceptions) as? [String]
let merged = sanitizedBundleIdentifierList(stored ?? (smooth + direction))
```

Launch migration runs before `register(defaults:)`, so `stored` is nil only when the key was never written — upgrade and old-backup restore still union. From `reload()` it is always non-nil.

Not verified:

- I read the removal path on `main` (`MouseExceptionsList.swift:27` → `remove(_:from:)`). `UI/` is past the 80KB cut, so if this PR changed how the shared row deletes, say so.
- Tests, `SmoothScrollService.swift` and the re-arm limiter are past the cut. No hardware run.

**Impact:** an app cannot be taken off the shared scroll exception list — it returns on the next reload and on every launch.
**Merge:** needs removal to survive the legacy mirror first.

ref #734

celltower:
Thanks for catching that, really helpful. Fixed in `6f7dfad` so the shared key wins once it exists and removals stick. Happy to adjust if anything still looks off.

PathGao:
Removal sticks now. `remove` writes the shared key, migration reads it back as `stored`, and the mirror follows — including when the last entry goes and the list lands on `[]`.

**The restore path trips the same sentinel.** `applyAndRelaunch` re-adds carried path identities after the clear, and that write creates `mouseScrollingExceptions` even when the backup file never carried it:

```swift
for (key, paths) in carried where !paths.isEmpty {
    defaults.set(..., forKey: key)   // key can be the shared one
}
```

So: except a program that has no bundle id (#1009) from scrolling, then restore a backup written by a shipping release. Relaunch migration finds `stored` non-nil, skips the union, and both legacy lists from the file are dropped. Only the carried path survives.

Lazy fix — migrate before the carried re-add, while the shared key is still absent:

```swift
for (key, value) in settings { defaults.set(value, forKey: key) }
Defaults.migrateMouseScrollingExceptions(in: defaults)
for (key, paths) in carried where !paths.isEmpty {
```

Not verified:

- **`Services/SettingsBackup.swift` still isn't in the diff** — its slot between `ScrollInverter.swift` and `SmoothScrollService.swift` is empty, so I read `applyAndRelaunch` on `main` and am treating it as untouched. If you changed it past the cut, say so.
- `SmoothScrollService.swift`, `UI/`, the tests and the re-arm limiter are past the 80KB cut. Check runs still at zero. No hardware run on my side.

**Impact:** restoring a pre-upgrade backup on a Mac that has a path exception in the scroll list drops every bundle-id scroll exception the file carried.
**Merge:** needs the shared key built from the file's legacy lists before carried paths are written back.

ref #734

celltower:
Thanks again for the catch, @PathGao. `16c300a` runs the scroll exception migration after writing the backup and before putting carried paths back, so an older file's legacy lists still fold in. Let me know if that looks right.

celltower:
Hey @PathGao, rebased onto current main and folded in the session-switch tap handoff from #1145. Could you take another look at the current state when you have a moment?

PathGao:
`portableMouseExceptions` holds up after the rebase, and the mirror no longer resurrects removed entries.

**The `stored == nil` sentinel can't fire inside `applyAndRelaunch`.** `mouseScrollingExceptions` is in `registeredDefaults`, so after `removeObject` the read comes back `[]` from the registration domain, not `nil`. Only the launch call, which runs before `register(defaults:)`, ever sees `nil`.

Restore a 3.3.x backup: clear → the file writes the two legacy lists → migration reads `stored == []`, skips the union, then mirrors `[]` onto both legacy keys, wiping what was just restored. That placement was my suggestion and it leaves the path worse than before — the post-relaunch migration used to union them.

Lazy fix: move the key out of `registeredDefaults` into `SettingsBackupSupport.unregisteredPreferenceKeys`. `exportKeys()` unions that set, so the key still travels, and `nil` means "never written" again at all three call sites. Every read is already `stringArray(forKey:) ?? []` — unless something past the cut needs the registered `[]`.

Not verified: `Services/SettingsBackup.swift` isn't in the diff, so I read `applyAndRelaunch` on `main` and am taking your word for where 16c300a put the call. `SmoothScrollService.swift`, `UI/`, the tests and the re-arm limiter are past the 80KB cut; checks still at zero; no hardware run.

**Impact:** restoring a backup written by any shipping release lands with an empty scroll exception list.
**Merge:** needs `applyAndRelaunch` to tell "never written" apart from the registered `[]`.

ref #734

celltower:
Thanks @PathGao — that registered `[]` masking nil after clear makes sense. Moved `mouseScrollingExceptions` into `unregisteredPreferenceKeys` in `248d1da` so absence still means never written at launch and restore, while the key stays in backups. Happy to adjust if anything still looks off.

celltower:
Hey @PathGao, rebased onto current main after the new mouse scrolling and click reliability work landed there, so both sides sit together now. Could you take another look when you have a moment?

PathGao:
The unregister lands. `mouseScrollingExceptions` is out of `registeredDefaults` and in `unregisteredPreferenceKeys` — only `exportKeys()` reads that set, so the key still travels and `object(forKey:)` is nil after a clear. Traced the 3.3.x restore against `main`: clear → the file writes the legacy pair → `stored == nil` → union → mirror. The lists survive.

**`handleSideWheel` is the sibling `claimsSideWheel` left behind.** Same scope, same side-wheel event, and only the preflight got the target pid (`MouseButtonShortcutService.swift:534` on `main`):

```swift
excludesActionTarget(.buttonShortcuts, at: event.location,
                     sourceProcessID: sourceProcessID)  // targetProcessID stays 0
```

That branch runs when `!wasPreflighted` — whenever the smooth-scroll tap isn't up, which is the default. `.eventTargetUnixProcessID` is already read in `MiddleClickService`. The `.otherMouseDown` guard at `:370` has neither pid, but that is unchanged from `main`.

Not verified: `SmoothScrollService.swift`, `UI/` and the tests are past the 80KB cut, so I can't see what the `claimsSideWheel` caller passes for `targetProcessID`. `SettingsBackup.swift` isn't in the diff — taking your word on where 16c300a put the migrate call. Checks still at zero; no hardware run.

**Impact:** with smooth scrolling off, an app on the shared list can still lose a side-wheel press to a mapped shortcut in exactly the case the preflight now covers.
**Merge:** needs `handleSideWheel` to pass the target pid too.

ref #734

celltower:
@PathGao Friendly ping: PTAL when you have a moment. Thanks!

PathGao:
`handleSideWheel` passes the target pid now — the `!wasPreflighted` branch has it alongside the `claimsSideWheel` preflight. That gap is closed, and I have nothing new to flag in what I can read.

**My view of this PR shrank this round.** The 80KB cut now lands mid-`ScrollInverter.swift`, so `SettingsBackup.swift` is past it and I can't even tell whether it's in the diff. Three rounds running I've taken your word on where `16c300a` put the `migrateMouseScrollingExceptions` call, and the whole 3.3.x-restore story rests on it sitting between the settings write and the carried re-add. Paste that function into a comment and I'll close it for real.

Also past the cut and never reviewed by me: `SmoothScrollService.swift`, `UI/`, the tests, the re-arm limiter. That is the feature. What I have read is defaults, migration, backup keys and the exception plumbing.

Minor: `MouseAppExceptionSupport.pointerWindow`'s closing brace picked up four extra spaces — looks like a rebase artifact.

Checks are still at zero, not failing. No hardware run on my side.

**Impact:** none — no open finding.
**Merge:** nothing I can name as a gap, but not `ready to merge` while the checks have never run and most of the diff is past my cut.

ref #734

celltower:
Brought current `main` in again after the branch conflicted (#1212 WindowServer share, tap-creation retry hardening, #1226 keycap Command table, plus the other main landings).

Kept this PR's Mos/thread scroll path and shared exception work. Took the live pieces from `main`: `WindowServerSupport.onScreenWindows()`, one-shot cancelable tap-creation retries on both scroll taps, and the new tests — dropped the dead CADisplayLink asserts that no longer match this branch.

Verification: `./build.sh --test` → `TESTS OK (9546 checks)`.

PathGao:
Rebase kept the exception plumbing — `portableMouseExceptions`, the unregistered shared key and the `handleSideWheel` pid all read as they did last round.

**Name the asserts you dropped.** `main`'s `MetricsTests` has two `displayLink` groups: cadence (`screen.displayLink(` / `displayLink.add(to: .main, forMode: .common)` / `sender.timestamp` / `sender.duration`) and teardown (`displayLink?.invalidate()` + `frameTimer?.invalidate()` + `removeScreenObserver()` + `removeSleepObserver()`). Moving the link onto your own thread kills the `.main` clause and nothing else — the other three cadence clauses and all four teardown clauses still hold on this branch. If the whole cadence group went, or the teardown group went with it, that is coverage `main` has today and the merge takes away.

**Nothing pins the new threads.** The tap-owner loop just above that pins `CFMachPortInvalidate` for all six owners, `ScrollInverter.swift` and `SmoothScrollService.swift` included. This PR gives each scroll tap a run-loop thread and pins no teardown for it. Same file, same shape — put the thread in that group.

Not verified: `SettingsBackup.swift` is past the 80KB cut, so `16c300a`'s placement is still your word. `SmoothScrollService.swift`, `UI/` and the tests are past it too. Checks still at zero; no hardware run.

**Impact:** none at runtime — a guard that fires on `main` stops firing after merge.
**Merge:** needs the dropped asserts named, and the teardown one back if it went.

ref #734

celltower:
Addressed the incomplete findings in the latest push.

**Dropped asserts (named):** from `main`'s CADisplayLink groups this path no longer carries `screen.displayLink(`, `displayLink.add(to: .main, forMode: .common)`, `sender.timestamp`, `sender.duration`, `displayLink?.invalidate()`, or `removeScreenObserver()` — the NSScreen-linked scheduler is gone; Mos uses `CVDisplayLink` on the scroll thread instead.

**Teardown kept / restored:** `frameTimer?.invalidate()`, `CVDisplayLinkStop` / `stopDisplayLink`, and `removeSleepObserver()` again (install on start, remove on stop).

**Thread pin:** the six-owner session-tap loop now also requires `CFRunLoopStop`, `clearEventTapThread`, and `tapThread = nil` for `ScrollInverter` and `SmoothScrollService`.

Verification: `./build.sh --test` → `TESTS OK (9556 checks)`.

PathGao:
Both landed. The named list is what I needed, and the six-owner loop now pins `CFRunLoopStop` / `clearEventTapThread` / `tapThread = nil` for both scroll taps.

**A third `displayLink` group isn't in your list.** `main`'s `MetricsTests` also holds `NSScreen.withMouse` + `didChangeScreenParametersNotification` + `Timer(timeInterval: SmoothScrollSupport.frameInterval` — "follows display changes and keeps a no-screen timer fallback".

You kept `frameTimer?.invalidate()`, so the fallback is still on the branch, which points at that group still passing and `installScreenObserver()` in `SmoothScrollService` still running. Then dropping `removeScreenObserver()` unpins the removal of an observer that is still installed. If the group went too, that is another coverage line the merge takes away, unnamed.

Not verified: the tests and `SmoothScrollService.swift` are past the 80KB cut, so this is `main` plus your list, not the branch. `SettingsBackup.swift` is still your word. Checks at zero; no hardware run.

**Impact:** none at runtime — an observer-leak guard that fires on `main` may stop firing after merge.
**Merge:** needs the third group named — kept, or dropped with `removeScreenObserver()` back if the observer stays.

ref #734

celltower:
@PathGao  The third group is dropped too, naming it now.

Also gone with the NSScreen-linked scheduler: `NSScreen.withMouse`, `didChangeScreenParametersNotification`, and main's "follows display changes" assert that tied those to `Timer(timeInterval: SmoothScrollSupport.frameInterval`. There is no `installScreenObserver()` / `removeScreenObserver()` on this branch; the timer that remains is only the no-DisplayLink fallback on the scroll thread (`frameTimer = timer` / `frameTimer?.invalidate()`), not the old screen-parameter path.

So no observer is left installed without a matching remove. `removeScreenObserver()` has nothing to unpin.

Verification unchanged: `./build.sh --test` → `TESTS OK (9556 checks)`.

PathGao:
That closes it. Three groups named, and with no `installScreenObserver()` on the branch there is nothing left for `removeScreenObserver()` to pin. On `main` those asserts are conjoined the way you describe, so dropping the NSScreen scheduler takes the whole group with it — consistent.

One leftover, not a gap: the fallback timer's *creation* was only pinned inside the group you dropped (`Timer(timeInterval: SmoothScrollSupport.frameInterval`). You kept `frameTimer?.invalidate()`, so teardown is covered and creation isn't — unless one of the new tests already pins it, which I can't see.

Still unread by me: `SmoothScrollService.swift`, `UI/`, the tests, `SettingsBackup.swift` — all past the 80KB cut. `16c300a`'s placement is your word, four rounds running. That is most of the feature.

**Impact:** none — no open finding.
**Merge:** nothing left from me. Not tagging `ready to merge` while the checks have never run — that button is yours, and it isn't a blocker I can label.

ref #734

PathGao:
Sending this back with a shape request rather than a code one.

The maintainer shipped his own scrolling rework on 2026-08-31 (4bda1305: adjustable speed and response, no lost wheel distance, the CADisplayLink scheduler). This branch deletes it — `smoothScrollResponse` and its default, `SmoothScrollSupport.Engine`, `advance(elapsed:response:)`, `installScreenObserver`, `displayLinkDidFire` — and replaces the scheduler with a dedicated tap thread, its own CFRunLoop and a CVDisplayLink, plus six new keys. At +2663/−1030, checks have never run on it, and `SmoothScrollService.swift`, the UI and the tests were outside the review's read window in all five rounds. A wholesale replacement of a two-day-old rework is not something a reviewer can accept on his behalf.

Two pieces of this are wanted on their own terms and stand alone: the shared exception list merge (issue #838 names it) and the #1023 identity plumbing. Split those out as a PR under +200 lines. For the feel model and scheduler, the question is whether the maintainer wants a Mos port on top of what he just shipped; put that to him on the issue with the numbers, since the summary says any per-event work has to be measured before it ships.

celltower:
Thanks for the clarity on the process side.

I am keeping this as one PR. The whole request is the Mos-like feel and its supporting exception/settings work; splitting the shared list out would leave the main piece stranded and is not something I want to open a second PR for.

The replacement of the recent main scroll feel (smoothScrollResponse, CADisplayLink / installScreenObserver, etc.) is intentional. Mos like scrolling is the model this PR ships. Work from main that is not that feel path (debounce, session handoff, HID acceleration control, WindowServer share, tap retry, and so on) stays and coexists. What goes away is the scheduler/feel surface Mos replaces.

On the product side: would be great if @vorssaint  could take a quick look when he has a moment and say whether he is fine with the Mos feel replacing the 2026-08-31 scroll rework. Happy to share numbers or a short measured note if that helps the call. From here the open question is that decision; I am not planning a second PR for the same change.

PathGao:
Keeping it as one PR is your call and I'm not relitigating it.

Two cheap things would make that decision easier to give:

1. **Post the numbers without being asked.** Your own summary commits to measuring per-event work before it ships, and a feel replacement is exactly that. Per-wheel-event cost, median and worst case, `main` vs this branch on the same machine — one comment.
2. **Get the checks to run.** They have never run on this branch. Right now the ask is to accept a replacement of a two-day-old rework on a +2663/−1030 diff with a zero-check history and a local `TESTS OK` as the only evidence. Green checks remove that objection and leave only the product question.

Tagging `blocked`: the named prior event is the maintainer's answer on whether the Mos feel replaces the 2026-08-31 scroll rework. It comes off when he says either way.

**Impact:** none — no open finding from me.
**Merge:** hold for the maintainer's call on the feel replacement.

ref #734

celltower:
@PathGao Here are the two things you asked for.

**1. Numbers** (same machine, `-O`, pure `SmoothScrollSupport` only: no CGEvent tap, no pointer / LaunchServices). 100k samples each.

| | `main` (`add` + `advance`) | this branch (`axisPlan` + `ingest` + `emitFrame`) |
|---|---|---|
| median | 0.041 µs | 0.167 µs |
| p95 | 0.042 µs | 0.208 µs |
| p99 | 0.042 µs | 0.209 µs |
| worst | ~5 µs | ~17 µs |

Mos does more work per tick (~4× here), as expected for the feel model. Absolute cost stays well under 1 µs. A pointer-cache miss is still the documented ~190 µs sync lookup on both paths and dwarfs the glide math either way.

Not in this run: full tap callback, frame posting, or DisplayLink scheduling. Happy to extend if you want a specific harness shape.

**2. Checks.** Local suite on current tip: `./build.sh --test` → `TESTS OK (9670 checks)`. Branch is synced with current `main`. GitHub check runs have still never started on this PR (first-time-contributor gate). I cannot flip that from here; once someone with access approves the workflow, CI can run.

Still holding for `@vorssaint` on whether the Mos feel should replace the 2026-08-31 scroll rework.

PathGao:
Numbers land. Sub-µs on both sides, and the 4× is on math that the ~190 µs pointer miss you named already dwarfs — a per-event objection was never going to come from there. The harness leaves out the tap callback and frame posting, which is where I'd look if someone pushes on cost, but nothing in what you measured points that way.

Withdrawing my second ask. The workflow-approval gate is the maintainer's button, not yours, so "get the checks to run" was never an action you could take. I shouldn't have asked for it.

`blocked` stays for the reason it went on: his call on whether the Mos feel replaces the 2026-08-31 rework. Green checks were never part of that.

Nothing else from me.

**Impact:** none — no open finding.
**Merge:** hold for the maintainer's call on the feel replacement.

ref #734

cfaester:
Big +1 from a user here. Current behaviour is strange coming from several other invert-scrolling apps.

celltower:
Hey @vorssaint — friendly ping.

This PR has been sitting on the `blocked` label since PathGao's call on Sep 2: whether the Mos feel may replace the 2026-08-31 scroll rework (`smoothScrollResponse` / CADisplayLink path). That is still the only open question; code findings from PathGao are closed, numbers are in the thread, and @cfaester has since +1'd from the user side.

Just re-merged current `main` into this branch (conflicts resolved; Mos path kept; Scroll Inverter on `PointerTapRunLoop` with Smooth Scroll on its dedicated tap thread; unknown-pointer stand-down kept). Local verification: `./build.sh --test` → `TESTS OK (10797 checks)`.

Would really appreciate a short yes/no on the feel replacement when you have a moment. We have been waiting on that answer for over a week now. Happy to adjust if the call goes the other way.

celltower:
Hey @vorssaint — would really appreciate if you could take a look when you get a chance.

The open question is still whether the Mos feel is the direction you want going forward (replacing the Aug 31 scroll rework). PathGao's code findings are closed, numbers are in the thread, and @cfaester has since +1'd from the user side.

The practical issue on our end: `main` keeps landing day after day, and each sync brings fresh merge conflicts. That churn is not slowing down, so holding this open keeps getting more expensive without a product call either way.

On the feel itself: this path has been exercised hard for well over four weeks now, every day, and it feels solid from here. Just re-merged current `main` again (tests green locally: `./build.sh --test` → `TESTS OK (10797 checks)`).

A short yes/no on whether this direction fits would help a lot — happy to adjust if you want something different.

EyJunge1:
@vorssaint +1 from a user.

vorssaint:
closing this one. the answer on the feel direction is no, because this replaces the scrolling engine that shipped in 3.3.5 and drops the guarantees it was built on
the current head does not build, there is a stray comma at the end of one of the settings page property declarations
the coast now runs on a display link api that apple deprecated in macos 15, while the shipped code already uses the replacement apple recommends, and the new one no longer follows the display the pointer is on
the glide no longer travels the distance the wheel asked for, the new test itself settles for most of it, and the coast lasts a different time on a 60 hz and a 120 hz display because each frame moves a fixed share regardless of elapsed time
a mouse that reports the wheel as many small continuous movements gets every one of them raised to a full step and then multiplied by the speed, so those mice scroll far too much per notch
the distance setting keeps its old key with a new meaning and the response setting is dropped without a mapping, so everyone who already uses smooth scrolling gets a much faster wheel after upgrading
the code and the tests describe another app by name throughout, and this project never names third party apps
the settings section shows two switches, four sliders with text fields and a reset button in the open, where the page keeps at most five main controls and puts the rest under more options
the two apps to leave alone lists become one on upgrade, undoing a choice users already made, and the split cannot be recovered
the exceptions lookup goes back to asking the window server and launch services on the tap thread for every cache miss, which is what the current main moved off that thread on purpose
a smaller change that tunes the coast on top of the current engine, keeps every pixel the wheel asked for and names no other app could still come in

celltower:
@vorssaint  Thanks for the detailed review, I read every point.

Honest feedback from my side: waiting over two weeks for an answer and then getting a straight decline with no discussion feels rough, especially when users in #734, #1486 and others keep asking for exactly this. Mos-style scrolling is one of the most requested mouse things, and for an all-in-one app it matters that the wheel feels right on a real mouse.

I am not asking you to throw your 3.3.5 engine away. What about keeping both: your engine stays the default, untouched, and the Mos-style feel becomes an opt-in mode for people who want it? That way nobody gets a changed wheel after upgrade, no migration harm, no forced direction.

Most of your technical points are fixable and I am happy to fix all of them: the build comma, proper migration with mapping so existing users keep their exact speed, collapsed settings UI, keeping both exception lists separate, no third party naming anywhere, staying on the recommended display link API with per-display follow, full wheel distance preserved, and no tap-thread lookups. If the opt-in mode still has to meet every guarantee, fine, I will do that work.

Did you get a chance to actually try the branch on a real mouse? The feedback from people who did is consistently that it finally feels right. I think it deserves a test before a final no.

Can we reopen this as a dual-mode proposal, or should I open a fresh small PR with your engine as default plus opt-in Mos-style feel, with all points above fixed?

Issue #615, opened 2026-08-12
[Feature]: Manage menu bar icons, hide / always-hide tiers with on-demand reveal


clovis-dugue:
### Problem or Motivation

My menu bar is permanently overcrowded: a dozen background apps each plant an icon, and on a notched MacBook the notch swallows whatever doesn't fit, with no way to reach those icons at all. I want a mostly-empty menu bar where a small set of icons is visible, the rest are reachable on demand, and a few (which I never interact with) stay out of sight entirely.
 
The problem is that the existing tools for this are in bad shape on macOS 26 Tahoe:
 
- **Ice** (open source, the usual free answer) is effectively dead. Its last stable release, 0.11.12, is broken on Tahoe: the Menu Bar Layout panel shows no icons (jordanbaird/Ice#744, jordanbaird/Ice#913) and clicking its menu bar icon crashes the app (jordanbaird/Ice#821). A Tahoe rework has been stuck in `0.11.13-dev` pre-releases for about a year with no stable in sight.
- **Bartender** is paid, and post-acquisition a lot of users (me included) no longer want it.
- **iBar Pro** is what I currently daily-drive: it works flawlessly on Tahoe and has exactly the right interaction model. But it's closed source, App Store-sandboxed, and its last update was v2.1.0 on 2026-04-11 with "bug fixes" release notes. So I'm relying on an opaque, slow-moving app for something I'd rather have in an open source codebase.

### Proposed Solution

The interaction model iBar Pro / Bartender / Ice converged on, which I'd love to see in Vorssaint:
 
1. **Hidden tier**: icons hidden by default, revealed on demand by clicking a Vorssaint control item (or hovering it).
2. **Always-hidden tier**: a second tier that stays hidden even during reveal, for icons you never click but can't remove.
3. **Reveal popup**, on notched Macs especially: instead of (or in addition to) expanding icons back into the bar, show the hidden icons in a small floating strip under the menu bar, where they're clickable and their menus work. Users assign icons to tiers with macOS's native ⌘-drag.

All settings (current icons and their tiers) could be manageable via drag and drop or via a dedicated settings screen

<details>
<summary>Research notes</summary>

[...]

PR #1042, opened 2026-08-26
feat(menu-bar-hider): add native Menu Bar Hider feature

0xJJphy:
### Summary

Refs #615, Refs #481

This PR implements the complete **Menu Bar Hider** feature in Vorssaint — a native, zero-dependency, and lightweight utility that replaces paid/third-party menu bar managers (such as Bartender or Hidden Bar) to keep the macOS menu bar clean and organized.

### Motivation & Design

- **0% Idle CPU Overhead**: Operates entirely through native AppKit `NSStatusItem` mechanisms and event hooks; no background polling loops or display timers when idle.
- **Zero Invasive Permissions**: Does not require Accessibility, Screen Recording, or kernel extensions.
- **Pure AppKit / SwiftUI Boundary**: All status item lifecycle, geometry, and event handling reside in `MenuBarHiderService` (pure AppKit/Combine). The UI in `MenuBarHiderSettings` strictly observes state.

---

### Key Capabilities

1. **3-Tier Icon Management (Always Visible, Hidden, Always Hidden)**:
   - **Always Visible**: Icons to the right of the control button (`>`) remain visible at all times (e.g. Clock, Control Center).
   - **Hidden**: Icons placed between the main separator (`|`) and the toggle button (`>`) collapse/expand on click or shortcut.
   - **Always Hidden**: Icons placed to the left of the double separator (`‖`) stay hidden during normal toggle operations.

2. **Precision Trailing-Edge Separator Engine (`MenuBarHiderSeparatorView`)**:
   - Uses a custom subview that pins the separator glyphs (`|` and `‖`) to the visible trailing boundary (`bounds.maxX - size.width - 2`).
   - Ensures both `|` and `‖` remain visible and interactable at the exact separation border even when item lengths expand to push adjacent icons off-screen.

3. **Auto-Healing Layout Architecture**:
   - Listens to `NSWindow.didMoveNotification` when status items are rearranged via `⌘ + Drag`.
   - Automatically sorts status items by their screen `X` coordinate:
     - Rightmost $\rightarrow$ Toggle Button (`>`)
     - Middle $\rightarrow$ Main Separator (`|`)
     - Leftmost $\rightarrow$ Always Hidden Separator (`‖`)
   - Guarantees that the toggle button can never end up behind an expanding separator or disappear off-screen.

4. **Rich Interactions & Gestures**:
   - **Click**: Toggles hidden icons.
   - **Double-Click or Click `‖`**: Reveals/collapses the Always-Hidden section.
   - **Right-Click Context Menu**: Native menu with options to expand, collapse, show all, hide always-hidden, and open preferences.
   - **Trackpad / Mouse Scroll-to-Toggle**: Swipe two fingers or scroll on the toggle button to quickly collapse or reveal icons (debounced at 350 ms).
   - **Haptic Feedback**: Subtle tactile response on trackpad state changes via `NSHapticFeedbackManager`.
   - **Auto-Collapse**: Optional timer (3s, 5s, 10s, 15s, 30s) scheduled on `RunLoop.main` in `.common` mode.
   - **Expand on Hover**: Optional mouse-over reveal.
   - **Global Keyboard Shortcut**: Customizable global shortcut (`GlobalShortcut`).

5. **Configuration Mode & Settings UI**:
   - Opening *Settings > Menu Bar Hider* automatically activates **Configuration Mode**, keeping all separators visible and pausing auto-collapse for seamless `⌘ + Drag` organizing.
   - **Custom Icon Styles**: Supports Chevron (`>`), Dots (`···`), Stealth Eye (`👁️`), and Minimalist Diagonal Line (`/`).
   - **Reset Separator Positions**: One-click action to clear saved `autosaveName` preferences and restore default ordering.

6. **Localization**:
   - Complete translations across all 13 supported languages (`enUS`, `ptBR`, `tr`, `ru`, `es`, `de`, `fr`, `it`, `ja`, `ko`, `zhHans`, `zhTW`, `zhHK`).

---

### Files Introduced / Modified

- `Sources/Vorssaint/Services/MenuBarHider/MenuBarHiderService.swift`: Main service managing status items, drag protection, scroll monitor, haptics, and auto-healing.
- `Sources/Vorssaint/Services/MenuBarHider/MenuBarHiderSupport.swift`: Pure width math, icon style definitions, delay validation, and testable helpers.
- `Sources/Vorssaint/Core/MenuBarHiderStrings.swift`: Full 13-language localized string catalog.
- `Sources/Vorssaint/UI/Settings/MenuBarHiderSettings.swift`: Dedicated SwiftUI settings view with live layout diagram.
- `Sources/Vorssaint/Core/Defaults.swift`: Defaults registration.
- `Sources/Vorssaint/Core/FeatureCatalog.swift`: Feature catalog entry.
- `Sources/Vorssaint/Core/GlobalShortcut.swift`: Global shortcut role integration.
- `Sources/Vorssaint/App/FeatureRuntime.swift`: Runtime dispatch registration.
- `Sources/Vorssaint/UI/Settings/SettingsView.swift`: Navigation routing.
- `Sources/Vorssaint/UI/Settings/SettingsDirectory.swift`: Settings search indexing.
- `Tests/MetricsTests.swift`: Comprehensive unit tests covering calculations, styles, and 13-language catalog completeness.
- `build.sh`: Unit test compilation manifest update.

---

### Verification

```bash
# 1. Full Release Compilation (0 warnings, 0 errors)
./build.sh

# 2. Complete Unit Test Suite (9,001 checks passed)
./build.sh --test
# Output: TESTS OK (9001 checks)

# 3. App Health Selftest
./build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest
# Output: SELFTEST OK

### CODE REVIEW

**Comment by PathGao** [comment]
Read the diff. Four things, plus what I couldn't see.

**1. Auto-heal reassigns the items but not their lengths.** `autoHealItemOrdering()` rebinds `toggleItem` / `separatorItem` / `alwaysHiddenItem` to different `NSStatusItem` objects by x position. `updateItemAppearances()` then sets `separatorItem.length` and `alwaysHiddenItem.length` — `toggleItem.length` is never touched after creation. So whatever item ends up as the toggle keeps the length from its old role. ⌘-drag the toggle left of `|` while collapsed and the old separator becomes the toggle at length 10000, which eats the menu bar; expanded, you get a 10 pt toggle.

Body says the auto-heal "Guarantees that the toggle button can never end up behind an expanding separator or disappear off-screen." This is the path that produces exactly that. One line in `updateItemAppearances()`:

```swift
toggleItem?.length = NSStatusItem.variableLength
```

**2. `menuBarHiderEnabled` defaults to `true`.** Every existing user gets `|` and `‖` added to their menu bar on update, and `menuBarHiderAlwaysHiddenEnabled` is `true` too. The body doesn't mention it. If it's deliberate, say so in the body — new menu bar icons for people who never asked for the feature come back as issues the week after release.

**3. Three tooltips are hardcoded English.** `MenuBarHiderSupport.toggleTooltip` plus the two separator tooltips set in `configureItemButtons` bypass `MenuBarHiderStrings` entirely. Body says "Complete translations across all 13 supported languages"; these are user-visible on hover.

**4. Double-click fires the single-click path first.** `sendAction(on: [.leftMouseUp, .rightMouseUp])` delivers one action per mouse-up, so a double-click runs `toggleClicked` at `clickCount == 1` (collapse/expand + haptic) and then again at `clickCount == 2` (showAll). Visible flicker, two haptic pulses. Either defer the single-click branch past the double-click interval or don't key the gesture on click count.

Also, smaller: `DefaultsKey.menuBarHiderCollapsed` is registered in `Defaults` and never read or written anywhere in the diff — dead key, and collapse state doesn't survive a relaunch. `setupTrackingArea()` calls `removeTrackingArea()` *after* auto-heal may have rebound `toggleItem`, so the area is removed from a button that never had it while the old button keeps its own — hover-to-expand then fires from an icon that is no longer the toggle. And two string typos: ru `enableCaption` has "левеer", it `alwaysHiddenCaption` says "separador" (Spanish).

**Not verified:** the diff I got is truncated at 80 KB. `MenuBarHiderSettings.swift`, `SettingsView.swift`, `SettingsDirectory.swift`, `Tests/MetricsTests.swift` and `build.sh` aren't in it, so I have no opinion on any of them — including whether the settings UI is what reads `menuBarHiderCollapsed`.

**One question:** #698 asks for a macOS 27 backend for "the menu bar organizer", which reads like one already exists or is planned. If so, the body should say how the two relate. I can only see this diff, not the rest of the tree.

**Impact:** #1 breaks the menu bar after a ⌘-drag reorder while collapsed. #2 changes menu bar contents for every existing user on update. #3 and #4 are visible but recoverable.
**Merge:** needs the `toggleItem.length` reset first.

ref #481
ref #698

0xJJphy:
Thank you for this — it's a genuinely good review, and all seven were real. Everything is fixed and pushed. Taking them in your order.

**1. The toggle length.** You were exactly right, including the diagnosis of how it happens. `autoHealItemOrdering()` rebinds the three references by x position and `updateItemAppearances()` only ever wrote the two separator lengths, so whichever item became the toggle kept whatever length its old role had. Your one-liner is in, next to the other two:

```swift
toggleItem?.length = NSStatusItem.variableLength
```

And you're right that the body claimed the opposite of what the code did. I've rewritten that paragraph rather than leaving it to be read charitably.

**2. The default.** Changed to `false`. It wasn't deliberate — I'd missed that every other new feature in the catalog registers `false`, and you're right that shipping two new icons to people who never asked for them comes back as issues the following week. The body now says the feature is off until it's switched on in Settings. `menuBarHiderAlwaysHiddenEnabled` stays `true`, but it only does anything once the feature itself is enabled, so nobody sees it either.

While I was there I found the `@AppStorage` fallbacks in `MenuBarHiderSettings` disagreed with `Defaults` in both directions — one declared `true` against a registered `false`, the other the reverse. Harmless because registration wins, but confusing to read. They match now.

**3. The tooltips.** All three moved into `MenuBarHiderStrings` across the 13 languages. There were two more you couldn't have seen from the diff: the guide diagram labels in the settings page ("Always Hidden" / "Hidden Icons" / "Visible") and the toggle's `accessibilityDescription`, which VoiceOver reads. Those are localized too.

To keep this honest going forward, `toggleTooltip` now takes the strings struct as a parameter, so the test asserts which field each state selects in every language instead of matching English substrings.

**4. The double click.** Your analysis of the mechanism was correct — one action per mouse-up, so `toggleClicked` runs at `clickCount == 1` and again at `2`. I tried both of the fixes you suggested and neither quite held up, so I'd like to explain where I landed and why, because it's the one change here I'd most want a second opinion on.

Deferring the single-click branch works, but the always-hidden section is on by default, so the wait landed on every collapse — the gesture people make constantly. It was immediately noticeable in use. Not keying on click count also works, but it means dropping a gesture people are already using.

What's in there now acts on the first click straight away and lets a second click escalate to show-all. Nothing waits, and the first click's result is a state the person actually asked for rather than a guess that has to be undone — so there's no wrong-then-right flicker, just a reveal that goes further if you click again. Two details I got wrong before getting it right, in case they're useful:

- I first timed the pairing myself and re-armed it on every click, which meant a rapid burst escalated to show-all on every second click. The sequence now comes from AppKit's `clickCount`, which only resets after the interval passes with no further click.
- `NSEvent.doubleClickInterval` turned out to be too wide to key the gesture on. It's a comfort setting that can sit above a second, so two deliberate collapse/expand presses fell inside it and were read as one reveal. The gap is now capped at `min(NSEvent.doubleClickInterval, 0.30)`, in a pure function with tests.

⌥-click and the right-click menu still reveal the section, as does clicking the `‖` separator itself.

**The smaller three.** All confirmed:

- `menuBarHiderCollapsed` was dead — and to answer the question you flagged as unverified, no, the settings UI wasn't reading it either. Nothing anywhere was. It's now written on collapse/expand and restored when the status items are first created, so the bar comes back the way it was left. Restoring only on first creation matters because `syncWithPreferences()` runs on every settings change and would otherwise clobber the state on screen. `isShowingAll` deliberately doesn't persist.
- `removeTrackingArea()` was resolving the button through `toggleItem?.button`, which auto-healing may have rebound since. It now holds a weak reference to the button that actually received the area. Exactly the hover-from-the-wrong-icon behaviour you described.
- Both typos fixed: `левеer` → `левее`, and the Italian `separador` → `separatore`. I swept the rest of the file for cross-language contamination and those two were the only cases.

**On the parts you couldn't see.** The 80 KB truncation cost you four real bugs, so for what it's worth: `MenuBarHiderSettings.swift` held two of the untranslated strings above, and `Tests/MetricsTests.swift` contained an unrelated hunk relaxing the dispatch pool starvation assertion. That hunk is reverted — it doesn't belong in a feature PR, and #1014 is the right place for it. For the record, that check fails on `main` here too, on a machine with enough cores.

## Four more that only showed up on hardware

Running the build on a notched Mac turned these up. None are visible in the diff.

**The notch.** `expansionLength` was `max(screenWidth * 4, 10000)` and never looked at the notch. On this 14" MacBook Pro the menu bar strip left of the notch is 790 pt of an 1800 pt display, so the separator was asking for an item roughly twelve times wider than the bar it lives in. With a wide status item in that strip, collapsing pushed everything off and left nothing reachable to ⌘-drag back — the feature effectively locked itself.

It now measures `auxiliaryTopLeftArea` and covers the strip plus a small margin, 798 pt here instead of 10,000. Covering the strip is sufficient, since nothing to the separator's left can be further away than the strip is wide.

I'm flagging this one particularly because it's the class of thing you warned about on #698 — an item sized past an unpublished threshold. The old constant was well beyond anything the bar could hold, and you were right to be wary of that pattern.

**A symbol that doesn't exist.** `line.diagonal.arrow.down.right` isn't a shipping SF Symbol, so the expanded "Diagonal Line" style resolved to nil and the button silently kept its previous image. The guard was `if let image { ... }` with `title` already emptied, and a variable-length status item with neither image nor title collapses to zero width — so a missing symbol makes the toggle disappear entirely. It uses `line.diagonal.arrow` now, falls back to the chevron and then to a text glyph, and a test resolves all four styles in both states so a name that doesn't exist fails the build.

**Expand-on-hover never closed.** `mouseExited` only restarted the auto-collapse timer, which returns early when that setting is off, so hover opened a bar that stayed open indefinitely. Two causes underneath:

- `configureItemButtons()` called `setupTrackingArea()` on every appearance update, so expanding tore the area down with the cursor inside it and installed a fresh one. AppKit doesn't treat an area installed under the pointer as entered, so the matching exit never arrived. The area tracks `inVisibleRect` and follows the button through length changes on its own, so it's left alone now unless the button actually changed.
- Even with that fixed, expanding changes the item lengths and AppKit may rebuild the item windows underneath, so the exit event can't be relied on. The pointer position is watched directly while a hover-opened bar is showing, with a 0.8 s grace so brushing past doesn't snap it shut.

That second one adds a repeating timer, which I'd rather point out than have you find: it runs only while a hover-opened bar is on screen and stops when it closes, so the "0% CPU at rest" line in the body still holds.

**Auto-collapse didn't fire in two cases.** Nothing armed the timer when the status items were installed, so a relaunch that restored an expanded bar never collapsed, and switching the setting on did nothing until the next manual expand — measured at 7 s with a 3 s delay before the fix, collapsing on time after. Separately, `isConfiguring` latches while the settings page is open and suppresses both collapse paths, and SwiftUI doesn't guarantee `onDisappear` when the Settings window is closed outright, so it could stay latched for the rest of the session. It now also ends when the window closes.

## One I'm deliberately leaving out

`reshowStatusItem()` rebuilds the main status item, and `verifyIconReappeared()` raises the "menu bar icon is still hidden" alert when it doesn't come back, appending a hint naming whichever third-party organizer is running. While the hider is collapsed, its separator can push that very icon off the leading edge — so the alert can fire because of us and then point the person at Ice or Bartender. The auto-collapse timer can also fire inside the verification window, so it isn't only reachable by pressing the button while collapsed.

I have the fix written: expand before rebuilding, and replace the body and the third-party hint with a localized explanation and a button into the hider's settings when we're the cause. I've kept it out of this PR on purpose — it adds a new user-visible alert path and a fourteenth string in all 13 languages, on a diff you've already read, and it touches the same few lines of `AppDelegate` that #360 rewrites when it moves `runningMenuBarManagerName()` onto `MenuBarManagerDetection`. Happy to push it here, open it separately, or hold it until #360 lands — whichever you prefer.

## On #698

You were right to ask, and the answer isn't the one I'd have given before looking. #698 is the macOS 27 tracker for #360, not for this. What I hadn't registered when I opened this PR is that #360 is sitting at `ready to merge` and overlaps this work at the mechanism level:

- `MenuBarDividerItem` creates three `NSStatusItem` dividers (`control` / `hidden` / `alwaysHidden`), sets `behavior = []` and `sendAction(on: [.leftMouseUp, .rightMouseUp])`, and collapses a section by expanding `statusItem.length`. That's this PR's design.
- Both offer visible / hidden / always-hidden sections driven by ⌘-drag.
- Both add an `AppFeature` case and both bump `allCases.count` to 55, and we both touch `Defaults`, `FeatureRuntime`, `FeaturePresets`, `FeatureVisibilitySupport`, `SettingsDirectory`, `SettingsView`, `FeatureHubSettings` and `build.sh`.
- Shipped together they'd put six divider items in one menu bar, each able to expand past the strip. `MenuBarManagerDetection` matches on Ice/Bartender/Dozer/minimalbar bundle prefixes, so it wouldn't flag Vorssaint's own.

Where they don't overlap they're close to complementary. #360 owns identification and movement — the WindowServer/AX enumeration, item identity, the drag-and-drop editor, undo, the secondary bar, conflict detection, the macOS 27 gate. None of that is here. This PR owns interaction with the collapse — global shortcut, auto-collapse with a delay, expand-on-hover, scroll-to-toggle, haptics, icon styles, the reveal gesture, reset positions, persisted state. None of that is in #360, and it needs no permissions.

So I don't think the useful question is which one, and I'd rather not decide it unilaterally. Three shapes I can see:

1. Rebase onto #360 and re-land the interaction layer on `MenuBarDividerItem`, dropping this PR's own dividers. Most work, least duplication, and #360's dividers would pick up the shortcut, hover, scroll and haptics they don't have — plus the notch sizing above, which `collapsedLength(screenWidths:)` would want regardless.
2. Ship both, mutually exclusive: teach the organizer's conflict check about the hider and gate one on the other. Cheapest, but two hub entries for one job.
3. Close this and fold the interaction pieces into #360 as follow-ups.

I'd go with the first if #360 is landing, and I'm glad to do that work. It's your call on ordering — #360 has been ready far longer than this.

One thing I want to state plainly rather than let it read as an argument in this PR's favour: I'm **not** claiming this works on macOS 27 where #360 is gated off. It's untested there, and the evidence you gathered on #698 points the other way — jordanbaird/Ice#954 has hidden items all becoming visible on 27, and Ice hides with the same divider-length technique. If that's the behaviour, this is broken there in the same way, just without the crash, and would need the same gate.

Thanks again for the review, and for the level of detail in it — the length bug in particular I don't think I'd have found on my own.

PathGao:
Four of the seven are checkable in the diff I have and all four are in it: the length reset, the `false` default, the tooltips routed through `MenuBarHiderStrings`, and both typos. Click handling, the hover watchdog and `MenuBarHiderSupport` are past the 80 KB cut, so on those I only have your description.

**One new thing, and it comes from the auto-collapse fix.** `syncWithPreferences()` arms the timer on `!isCollapsed` alone:

```swift
if !isCollapsed {
    restartAutoCollapseTimerIfNeeded()
}
```

The other three arming sites — `expand()`, `showAll()`, `endConfigurationMode()` — all gate on `!isConfiguring`. Every settings page in this repo calls its service's `syncWithPreferences()` directly from a toggle's onChange (`ScreenshotSettings.swift:68`, `QuickToolsSettings.swift:42`, and a dozen more), so switching auto-collapse on inside *Settings > Menu Bar Hider* arms a 3 s timer while configuration mode is supposed to have it paused, and the bar collapses under the ⌘-drag the page exists to support.

```swift
if !isCollapsed && !isConfiguring {
```

Unless `restartAutoCollapseTimerIfNeeded()` checks `isConfiguring` itself — which would make the guard at the other three call sites redundant. Its body is past the cut, and so is `MenuBarHiderSettings.swift`, so the call-in pattern is inferred from the rest of the repo, not read.

**Double click.** Ship it. Acting immediately and letting the second click escalate is the right axis, and sourcing the sequence from `clickCount` instead of your own timer is the part that makes it hold.

Two notes. The `min(doubleClickInterval, 0.30)` cap means someone who set a slow double-click for accessibility reasons can never reach the gesture — the cap is fine, but the reason it's fine is that ⌥-click, right-click and the `‖` click all still get there, not that 0.30 is generous. And starting from expanded, a double-click is still collapse-then-reveal: icons vanish, then all of them appear. Monotone from collapsed, not from expanded.

**reshowStatusItem.** Split it, and the expensive half disappears. `reshowStatusItem()` already turns off `menuBarHideIconWithMetrics` on the grounds that the button is an explicit "I want the icon back" and nothing should re-hide it. Same move, one line:

```swift
if AppFeature.menuBarHider.isAvailable {
    MenuBarHiderService.shared.expand(startTimer: false)
}
```

`startTimer: false` covers the auto-collapse case too — verification runs 4 × 0.8 s and the shortest delay is 3 s, so the timer lands inside the window. Guard on `isAvailable` so this doesn't wake the singleton for people who never installed the feature; `FeatureRuntime.sync(_:)` keeps that invariant deliberately.

With that in, we're never the cause of the alert, so the alert never misattributes, so there's no new user-visible path and no fourteenth string. If a third-party organizer is holding the icon, the existing body and hint are still correct. That leaves nothing worth a separate PR — unless you know a case the expand doesn't cover, in which case say which.

**#360.** I haven't read its diff, so the comparison below is yours, not mine. What I did check on `main`: no `MenuBarDividerItem`, no `MenuBarManagerDetection`, no `collapsedLength`, and `Tests/MetricsTests.swift:10430` asserts `AppFeature.allCases.count == 54`. Both PRs bump it to 55, so whichever lands second has a failing test on top of the conflict. Nobody gets to skip the rebase.

Rule out 2. Two hub entries for one job is the visible cost; the invisible one is that `MenuBarManagerDetection` doesn't match Vorssaint's own bundle, so the conflict check can't be what keeps the two apart, and you'd be adding a second mechanism to gate rather than removing one. Between 1 and 3 I'd take 1 — the interaction layer is a coherent unit worth reviewing on its own, and it's the part #360 doesn't have.

Whichever way it goes, the `auxiliaryTopLeftArea` measurement should land in whatever divider survives. That one isn't a feature, it's a correctness fix, and a `screenWidth * 4` sizing anywhere in the tree has the same lock-itself-out failure on a notched Mac.

The ordering call is the maintainer's, not mine.

**Impact:** with auto-collapse on, the bar collapses mid-⌘-drag while the settings page is open. Nothing else new that I can see.
**Merge:** hold for the shape decision on #360.

0xJJphy:
Thanks again @PathGao — both points are in and pushed in `ff460a5`, and the branch has been rebased onto the latest `main`.

1. **Auto-collapse guard during configuration**:
   `restartAutoCollapseTimerIfNeeded()` was already checking `if isConfiguring { return }` internally at line 756, but to make the calling convention completely explicit and consistent with the other sites, `syncWithPreferences()` now also gates on `if !isCollapsed && !isConfiguring`.

2. **`reshowStatusItem` clean one-liner**:
   Adopted your suggestion verbatim:

   ```swift
   if AppFeature.menuBarHider.isAvailable {
       MenuBarHiderService.shared.expand(startTimer: false)
   }
   ```

   `startTimer: false` leaves the bar expanded through the entire verification window without racing the auto-collapse timer, completely preventing our hider from causing a false "still hidden" alert. This cleanly avoids any new alert paths or extra localized strings.

3. **On #360**:
   Completely agree with ruling out option 2. Option 1 makes total sense if #360 lands first — happy to rebase the interaction layer and notch math onto its dividers whenever the maintainers decide on the sequence.

PathGao:
Both are in the diff I have, and both read right.

Two tails on my own one-liner that I should have caught when I proposed it.

**The `isAvailable` guard doesn't do what I said it does.** `logStatusItemPlacement()` reads `MenuBarHiderService.shared` unconditionally at `AppDelegate.swift:319`, and it sits in the same call chain: `reshowStatusItem()` → `verifyIconReappeared()` → `logStatusItemPlacement()`. Anyone who presses the button wakes the singleton 0.8 s later anyway. The logged value is identical either way, since `isEnabled` is already `isAvailable && …`, so the guard only buys not constructing the service. Pick one — guard the log too, or drop the guard I asked for. Two answers to the same question in one call chain is worse than either.

**`startTimer: false` has no re-arm.** Nothing restarts the timer afterwards, not `verifyIconReappeared` and not the alert path, so with auto-collapse on the bar sits expanded for the rest of the session unless the person toggles it or opens Settings. Defensible — they just asked for the icon back — but as written it's a side effect, not a decision. Either re-arm where the verification ends, or say in a comment that staying expanded is the point.

**Impact:** with auto-collapse on, it stops firing after "Show menu bar icon" until the bar is toggled again. Nothing else new.
**Merge:** hold for the shape decision on #360.

ref #481

PathGao:
Closing this on macOS 27, and I want to be specific about why, because the fix is real work rather than a flag.

Five paths in `MenuBarHiderService.swift` treat an `NSStatusItem`'s window frame as truth: the role sort (`return (item, window.frame.origin.x)`), the `windowDidMove` auto-heal that re-runs it, the hover collapse and the scroll-to-toggle (`window.frame.contains(NSEvent.mouseLocation)`), and the bar-width read for the 10,000pt separator. `main` documents at `Sources/Vorssaint/App/StatusItemAnchorSupport.swift:8-15` that on macOS 27 a (re)created status item's window frame can stay stranded at the slot it was born in — the far right — while the icon is drawn and clicked at the user's arranged spot, and that the mismatch survives relaunch.

So on macOS 27 the auto-heal can reassign which item is the toggle and which is the separator, and the 10,000pt expansion then lands on the wrong one; hover collapse and scroll-to-toggle never fire, because the pointer is never inside the frame being reported. There is no `#available` or version check anywhere in the diff, and the verification section lists only `./build.sh`, `--test` and `--selftest`, with no macOS version.

If you want to bring it back:

1. Drop the frame-based role sort. You already set `autosaveName` on all three items — key the roles off that identity instead, so a lying frame cannot reshuffle them.
2. Replace both `frame.contains(NSEvent.mouseLocation)` tests with the click-anchor approach in `StatusItemAnchorSupport`: the click that opens the interaction is the one position the window server vouches for, and it outranks the reported frame when the two disagree.
3. State the macOS version you tested on, and exercise it after a status-item rebuild, not just a fresh launch — that is the state that produces the stale frame.

Worth reading #698 before the rewrite. The hide-by-expansion model itself is unverified on macOS 27, and Ice hits the same wall there (jordanbaird/Ice#954), so the frame reads may not be the only thing to change.

PR #1044, opened 2026-08-26
test(auto-quit): name real line numbers when the parent-walk rule fires

iltonandrew:
Follow-up to the post-merge review on #1016 — the scan's line numbers only, as scoped there.

## Problem

The parent-walk scan filtered comment lines out of `lines` before enumerating, then reported `index + 1` against the filtered array. When the rule fires, the number it names counts surviving lines, not source lines — ~30 short at the `WindowMaximizer.swift` walk, more in a comment-heavier file. The neighbouring application-role scan already enumerated raw lines and dropped comments in its `where`; only this one was off.

## Change

The suggested shape, verbatim: keep the raw array, filter in the predicate. The lookback window stays 3 — the guard is the line immediately above the read, and a comment wedged between guard and read eating the window is the review-prompting behaviour we want rather than a reason to widen the rule.

One strengthening while in there: the window now also ignores comment lines when looking for the guard, so a commented-out `isApplicationElement(parent)` left above a live read cannot vouch for it. Before this, deleting the guard by commenting it out kept the scan green.

## Verification

Mutation-proven both ways on `WindowMaximizer.swift`:

- Unguarded read inserted at real line 481: reported **448** before this change, **481** after.
- The real guard at line 291 commented out: the scan fails naming the read's real line, `Services/WindowMaximizer.swift:292`.

`./build.sh` zero warnings, `./build/Vorssaint --selftest` → SELFTEST OK. `./build.sh --test` green apart from the known machine-dependent pool-starvation check (this machine cannot starve its pool; #1014 adds the skip).


PathGao:
Read the diff. The index fix is right — `index + 1` against the unfiltered array is the real line number, and the window boundary still covers the line immediately above the read (for a read at 292, `lines[288..<291]` = real lines 289–291), which your 291/292 mutation confirms.

Both changes only make the scan stricter, so a green `./build.sh --test` is enough to say no existing call site regressed.

The lookback narrowing is declared in the body, so no complaint, but note the cost: three raw lines minus comments means a doc comment plus a blank line between guard and read drops the guard out of the window. That's a choice, not a bug — keep the reasoning in the body so the next person doesn't "fix" it back.

Minor: `!$0.trimmingCharacters(in: .whitespaces).hasPrefix("//")` appears twice in this hunk, and per your body the application-role scan carries its own copy. Third copy of "what counts as a comment line" in one file. A local `func isComment(_ s: String) -> Bool` collapses all three. Not worth blocking on.

Not verified: the application-role scan and the file list the loop iterates are both outside the diff, so "only this one was off" I'm taking on your word.

**Impact:** none for users; the scan now fails on strictly more inputs than before (commented-out guards, and guards pushed out of a raw 3-line window).
**Merge:** ship it.

iltonandrew:
Collapsed in af410e4 — a local `isCommentLine` above both scans, used at all three sites, with the reason it exists stated once where it lives (the report has to stay in real line numbers, so neither scan may filter its array before enumerating). Left the older `.filter` copies elsewhere in the file alone; they belong to checks that don't report line numbers and aren't this PR's business.

Behaviour unchanged, re-proved rather than assumed: the same mutation (real guard at `WindowMaximizer.swift:291` commented out) still fails naming `Services/WindowMaximizer.swift:292`, and the clean run is green apart from the known pool-starvation check.

On the lookback cost — agreed, and it stays in the body: three raw lines minus comments means a doc comment plus a blank line between guard and read drops the guard out of the window. That is the behaviour I want (it asks for a look rather than widening the rule), so the reasoning is recorded there for whoever meets it next.

Your unverified note is right to flag: "only this one was off" was my reading, and the diff doesn't carry it. The application-role scan enumerates raw lines and drops comments in its `where` — same shape as the fixed one — which is why its `AutoQuitService.swift:163` was correct while the filtered scan's number was short.

PR #1046, opened 2026-08-26
fix(build): sweep staging temp dirs on exit

PathGao:
`./build.sh` leaves a temp directory behind on every run, successful ones included. Six of them, 264 MB, had accumulated on my machine before I noticed the newest one came from a build that had just printed `✓ Bundle ready`.

Two separate causes:

- `STAGE="$(mktemp -d)/$APP_NAME.app"` reaches the staging dir through a path suffix, so the dir itself is held in no variable and nothing ever removes it. That is the 44 MB per build — a full copy of the signed bundle.
- `ICON_TMP` is removed, but by a line near the end of the icon step. `set -e` skips it whenever anything between `mktemp -d` and that line fails, which is exactly when a leftover is least likely to be noticed.

Both are now swept by a `cleanup()` trap installed before either dir exists, so an early `exit` (`--test` returns long before the icon step) and a mid-build failure sweep the same as a clean run. This is the shape `Tools/notarize.sh`, `Tools/setup-signing.sh` and `Tools/make-dmg.sh` already use; `build.sh` was the one script in the repo staging into `$TMPDIR` without it.

**Why a trap rather than moving the staging dir into `build/`.** The header comment gives the reason the dir is outside the repo in the first place: File Provider-synced folders add xattrs that invalidate the signature. Staging under `build/` would trade a leak for a signing failure on exactly the machines the comment is about. The trap keeps the property and drops the leak.

**Why the explicit `rm -rf "$ICON_TMP"` is gone rather than kept as belt-and-braces.** Two owners for one dir is how the success-path-only removal survived this long. `cleanup()` is now the single place the sweep is written.

The test reads the sweep out of `build.sh` rather than restating it, so the two cannot drift: it pins that the trap is installed, that every `mktemp -d` is captured whole into a variable — the check that would have caught the staging dir, since a path suffix leaves nothing to sweep — and that each captured variable is named inside `cleanup()`. A staging dir added later and not swept fails the run instead of leaking on every build. Verified in both directions: reverting `build.sh` to `main` turns the three assertions red, and keeping the trap while dropping one variable from `cleanup()` reddens that variable alone.

Tests 8706 checks, `--selftest` OK, full `./build.sh` 4:51 and `$TMPDIR` unchanged across it (0 dirs before, 0 after; on `main` the same run leaves 1).

No `CHANGELOG.md` entry: nothing here is user-visible, and the 3.3.2/3.3.3 sections carry no build or test entries. No behaviour change to the produced bundle — the staging path, `ditto` copy, signing and `build/stage` output are untouched, which the unchanged `--selftest` and signature verification in the same run cover.


PathGao:
`cleanupBody.contains("$\(variable)")` matches on a substring, so the guard has a hole exactly where a new dir is most likely to be named. `STAGE` is already a variable two lines under the sweep — if someone later writes `STAGE="$(mktemp -d)"` and forgets `cleanup()`, `"$STAGE"` is found inside `rm -rf "$STAGE_TMP"` and the assertion stays green. Same for `ICON` vs `ICON_TMP`. That is the leak the test exists to catch.

Quote both sides, since the sweep quotes:

```swift
expect(cleanupBody.contains("\"$\(variable)\""),
       "temp dir \(variable) is swept by build.sh cleanup()")
```

`"$STAGE"` is not a substring of `"$STAGE_TMP"`, and it pins the quoting on the way past.

Two smaller ones:

- The test pins that the trap exists, not that it is installed before the first `mktemp -d`. Move `trap cleanup EXIT` below the icon step and every assertion stays green while a mid-icon failure leaks again — the second of the two causes in the body.
- `EXIT` alone: bash does run the EXIT trap on Ctrl-C, but a closed terminal during a 4:51 build sends SIGHUP and skips it. `trap cleanup EXIT INT TERM HUP` if that is the shape the sibling scripts use — `Tools/notarize.sh` is not in the diff, so I can't tell which one you're matching.

The `ICON_TMP=` assignment is also outside the diff, so I'm taking the exact `="$(mktemp -d)"` spelling from your green run rather than from reading it. The trap itself reads right: both vars exist before it is installed, `return 0` keeps the script's exit status, and `[[ -n … ]] && rm -rf …` doesn't trip `set -e` since the failing test is not the last command of the AND-OR list.

**Impact:** none on the produced bundle or on the build. Confined to how much drift the new test can catch.
**Merge:** needs the quoted match in the cleanup assertion first.

PathGao:
Fixed in `6a996091`.

**Quoted match.** `cleanupBody.contains("\"$\(variable)\"")`. The hole was real: adding `STAGE="$(mktemp -d)"` without a `cleanup()` entry now reddens `temp dir STAGE is swept by build.sh cleanup()`, where the substring form stayed green on the sweep of `STAGE_TMP`.

**Trap position.** The existence assertion is replaced by an offset comparison — `trap cleanup EXIT` has to appear before the first `mktemp -d`. Moving the trap below the icon step reddens that assertion alone.

**Signals.** The siblings are `EXIT` alone (`Tools/notarize.sh:40`, `Tools/setup-signing.sh:34`, `Tools/make-dmg.sh:27`), so there was no shape to match — and the premise inverts here, because `build.sh` is `#!/bin/zsh`, not bash. Measured on a reduced repro with the same header (`set -euo pipefail`, `trap cleanup EXIT`):

```
EXIT only:            INT → leaked   TERM → leaked   HUP → swept
+ trap 'exit 1' …:    INT → swept    TERM → swept    HUP → swept
```

Ctrl-C is the ending that leaked; hangup was already covered. `trap cleanup EXIT INT TERM HUP` would be the wrong repair for it — zsh runs the handler and then *resumes* the build, now with the staging dir deleted under it, and cleanup runs a second time on the way out. `trap 'exit 1' INT TERM HUP` reaches the same sweep and keeps the interrupt an interrupt. Pinned by a third assertion; dropping the line reddens that one alone.

Tests 8707 checks (two added, one replaced). Full `./build.sh` 4:40, `$TMPDIR` unchanged across it, and one real build interrupted at t=275s — just after `ICON_TMP` existed — exited and swept.

`ICON_TMP="$(mktemp -d)"` is at `build.sh:387`; the spelling you inferred is the one in the file.

Not changed: the three `Tools/*.sh` carry the same `EXIT`-only gap. Their dirs are small and their runs short, so that is its own PR rather than scope added here.
@PathGao

PathGao:
The three assertions cover the two-line half of the pattern. A swept dir needs a third line — `ICON_TMP=""` above the trap — and nothing pins it.

Add `DMG_TMP="$(mktemp -d)"` later plus `[[ -n "$DMG_TMP" ]] && rm -rf "$DMG_TMP"` in `cleanup()`, skip the `DMG_TMP=""`, and all four assertions stay green: captured whole, named quoted inside `cleanup()`, trap still first. Under `set -euo pipefail` the trap then fires on `--test`'s early exit with `DMG_TMP` unset, and `[[ -n "$DMG_TMP" ]]` is a nounset reference. `return 0` is never reached, and whatever is listed after that line — `STAGE_TMP`, if the new entry goes above it — is not swept.

I reasoned that from `set -u`; I did not run it. The blast radius inside a zsh EXIT trap (handler aborts vs. shell exits) is yours to measure, the way you measured INT/TERM/HUP. Either ending leaves the sweep broken with the suite green.

Same shape as the cleanup-name assertion:

```swift
let initialized = buildScript.range(of: "\n\(variable)=\"\"")?.lowerBound
expect(initialized != nil && sweepInstalled != nil && initialized! < sweepInstalled!,
       "temp dir \(variable) is empty before the sweep is installed")
```

The `\n` for the same reason the sweep match got quotes: `ICON_TMP=""` is a substring of `STAGE_ICON_TMP=""`.

This is the body's own argument. "Two owners for one dir is how the success-path-only removal survived this long" — an invariant nothing pins is how the next one survives. `Tools/make-dmg.sh:14-15` carries the same unpinned third line, which is why I'd pin it here rather than inherit it as house style.

Smaller one, no action unless it bites: the count check reads bare `mktemp -d` while the capture check reads `="$(mktemp -d)"`, so three spellings that are not the bug redden with "captured whole into a variable" — a comment in `build.sh` naming the call, `X=$(mktemp -d)` unquoted, `X="$(mktemp -d -t vorssaint)"`. The assertion is right; the message names a constraint it isn't checking.

On the sibling claim, from `main`: `notarize.sh:39` and `setup-signing.sh:33` install `trap 'rm -rf "$WORK"' EXIT` one line *after* the mktemp and have no `cleanup()`; only `make-dmg.sh` is the shape this copies. So the new `build.sh` is tighter than two of the three, not equal to them. I did not run the suite or a build.

**Impact:** none today — both dirs are initialized. Confined to the drift the new test can catch.
**Merge:** needs the initialization pinned first.

PathGao:
Fixed in `00e48fce`, and the blast radius is the bad one.

**Measured, zsh, `set -euo pipefail`.** With `MISSING_TMP` referenced first in `cleanup()` and a real dir listed after it:

```
cleanup:1: MISSING_TMP: parameter not set
exit status: 0        ← unchanged, whatever the script was exiting with
later entry:  LEAKED
earlier entry (unset ref listed last): swept
```

So it is the abort branch, and it aborts *quietly*: entries above the bad line still sweep, everything below it does not, and the exit status is untouched — the build prints `✓ Bundle ready`, exits 0, and leaks. One stderr line is the whole signal. Same at exit 3, so it is not tied to the success path.

That is not just the repro. Dropping `ICON_TMP=""` from `build.sh` and running `--test` prints `cleanup:1: ICON_TMP: parameter not set` from the real trap on the real early-exit path.

**Assertion**, your shape, inside the existing per-variable loop:

```swift
let initialized = buildScript.range(of: "\n\(variable)=\"\"")?.lowerBound
expect(initialized != nil && sweepInstalled != nil && initialized! < sweepInstalled!,
       "temp dir \(variable) is empty before the sweep is installed")
```

Verified against the case you constructed — `DMG_TMP="$(mktemp -d)"` after the staging dir plus `[[ -n "$DMG_TMP" ]] && rm -rf "$DMG_TMP"` in `cleanup()`, no `DMG_TMP=""`: the other four stay green, this one reddens alone. Same for dropping `ICON_TMP=""`, and same for moving it below the trap rather than deleting it.

**Count-check message.** Reworded to what it checks — `every mktemp -d in build.sh is a whole capture — no path suffix, no other spelling`. The check itself is unchanged; a comment naming the call would still redden it, but now it reddens saying so.

**Siblings.** You are right and I was not: `notarize.sh:39-40` and `setup-signing.sh:33-34` install `trap 'rm -rf "$WORK"' EXIT` one line after the `mktemp -d`, with no `cleanup()`. Only `make-dmg.sh:14-16,27` is this shape, and its three `VAR=""` lines are unpinned. My earlier "the siblings are EXIT alone" was accurate only about the signal set.

Tests 8709 checks (two added, one per staged dir). `build.sh` is byte-identical to `6a996091`; this commit is test-only.

Not changed: `make-dmg.sh`. Its three initializations are correct today and the sweep there is worth pinning the same way, but that is the sibling PR, not this diff.

PathGao:
`00e48fce` closes the initialization gap. Checked against `main`: `build.sh` has exactly two `mktemp -d` (`build.sh:377`, `build.sh:408`), both whole captures once this diff lands, so the count assertion is 2 == 2 and the per-variable loop covers both.

One left, and it is the shape you just fixed. `trap cleanup EXIT` is pinned by position; `trap 'exit 1' INT TERM HUP` is pinned by `contains` — the position-free form the EXIT assertion was upgraded away from in `6a996091`. Move the interrupt trap below the icon step and all five assertions stay green, while a Ctrl-C between `ICON_TMP="$(mktemp -d)"` and the trap takes zsh's default INT and never reaches the EXIT trap. Your table: `EXIT only: INT → leaked`. That is the ending the line exists for.

Same two-line shape as the other pin, and `range(of:) != nil` subsumes the `contains`, so it replaces that assertion rather than adding a sixth:

```swift
let signalsRouted = buildScript.range(of: "trap 'exit 1' INT TERM HUP")?.lowerBound
expect(signalsRouted != nil && firstStaged != nil && signalsRouted! < firstStaged!,
       "build.sh routes interrupts through the sweep before it stages the first dir")
```

Nothing else in the diff needs an assertion I can name. `return 0` is the last line of `cleanup()`, so dropping it cannot strand a sweep the way an unset reference could, and both `rm -rf` entries are covered by the naming check.

Not verified: I read `main`, not this branch, so the count and the ordering come from applying the diff to `src/build.sh`. I did not run the suite or a build, and the INT/TERM/HUP measurement is yours. `Tools/make-dmg.sh:14-16` does carry three unpinned initializations as you said — its `cleanup()` also has no `return 0`, which is that PR's problem rather than this one's.

**Impact:** none today — both traps sit at the top, above every `mktemp -d`. Confined to the drift the new test can catch.
**Merge:** needs the interrupt trap pinned by position first.

PathGao:
Fixed in `e88d02d0`, your snippet as written.

Both directions: moving `trap 'exit 1' INT TERM HUP` below the icon step reddens `build.sh routes interrupts through the sweep before it stages the first dir` alone, and so does deleting the line. `range(of:) != nil` subsumes the old `contains`, so the count is unchanged at 8709 — five assertions, not six.

Two line numbers, from this branch rather than from applying the diff: the two `mktemp -d` are `build.sh:391` and `build.sh:420`. Your 377/408 is the same two calls; the offset is the twelve lines this diff adds above them. Count assertion is 2 == 2 either way.

`build.sh` is byte-identical to `6a996091`. The last two commits are test-only.

Nothing outstanding from my side. `Tools/make-dmg.sh` — three unpinned initializations and a `cleanup()` with no `return 0` — stays out of this diff.

PathGao:
`e88d02d0` closes the last one. Five assertions, and I can't name a sixth.

What I checked against `main` this round rather than re-reading the diff:

- **The sweep can't race anything.** `--install` re-execs itself at `build.sh:118-126` — `VORSSAINT_INSTALL_CHILD=1 "$0" "$@"` then `child_status=$?`, synchronous, nothing detached. That was the one way an EXIT sweep could regress the install path: a process still reading `$STAGE` after the parent exits. There isn't one, and `finalize_installed_bundle_after_child` works on `/Applications/…`, not the staging dir.
- **No trap to collide with.** `build.sh` on `main` has none, so `trap cleanup EXIT` is neither replacing a later one nor being replaced by it.
- **`cleanup() {` resolves to the right body.** No other function name in `build.sh` contains `cleanup`, and it's the first function in the file, so `components(separatedBy:)` can't latch onto a `foo_cleanup() {` the way `contains` latched onto `STAGE_TMP`.
- **Count is 2 == 2.** `build.sh:377` and `:408` on `main` are the only two `mktemp -d`; both become whole captures.
- **The `--test` path exits at `build.sh:339`**, before the icon step at `:377`, so both vars are empty there. That's the path your `parameter not set` measurement lands on.

`return 0` is the one line in `cleanup()` still unpinned, and it should stay that way — your own table shows zsh leaves the exit status alone (`exit status: 0 ← unchanged`), so dropping it changes nothing an assertion could observe.

The body's "the one script in the repo staging into `$TMPDIR` without it" holds. `Tools/ui-smoke.sh:17` also makes a temp dir with no sweep, but it names `/tmp` explicitly and that dir is the screenshot output the caller is meant to keep.

Not verified: I read `main` plus the diff, not the branch. I did not run the suite or a build, and the INT/TERM/HUP and nounset measurements are yours.

**Impact:** none on the produced bundle — staging path, `ditto` copy, signing and `build/stage` output are untouched.
**Merge:** ship it.

Issue #812, opened 2026-08-23
It doesnot work for the intel based mac machines


mirzabilalamjad:
<img width="1135" height="86" alt="Image" src="https://github.com/user-attachments/assets/53878bfe-78ca-44d9-9af6-e1785c85d7f1" />

kevinschweikert:
The README specifies this:

>  ### What you need
> - A Mac with Apple Silicon
> - macOS 14 Sonoma or newer

HectorPalomeque:
I just successfully validated an Intel/Universal build on an Intel iMac13,2 (27-inch, Late 2012) featuring a 2.9 GHz quad-core Intel Core i5 and 16 GB of RAM running via OpenCore Legacy Patcher. The monitoring suite (CPU, GPU, RAM, SSD, temps) and the fan control helper are fully functional on this hardware.

@mirzabilalamjad: If you want to bypass the current architecture block and try it on your machine today, I have set up a dedicated repository for this. You do not need to compile it from source. I have published a pre-compiled Universal .app release (x86_64 + arm64).

Download the working Intel version here:
https://github.com/HectorPalomeque/vorssaint-utils-intel/releases/tag/v3.3.0-intel

PathGao:
Since we are using some totally different fundamental system SDK and without intel mac we cant test，This app wont support intel chip mac forever. Thank you guys，you can have your own fork version.

Issue #523, opened 2026-08-06
[Feature]: Allowing Intel Mac Use


9p56pzh4ht-dev:
### Problem or Motivation

This seems like the catch all app. I would love to install it but when going through home-brew, it flags my computer for being an intel chipset and won't allow install. Any fixes?

### Proposed Solution

Possibly having a bypass command to install regardless of hardware

### Feature Area

Other

mangoduck:
I second this request. Running Tahoe on capable x86 hardware and saw this teased on YT but can't try it. If some features are Apple Silicon specific like hardware control or monitoring I'd be fine with them just being disabled for now so I can at least use the GUI features.

memohammedalmazyad-crypto:
I built and ran Vorssaint on an Intel Mac this week, so I wanted to share what I
found in case it is useful. I know Intel support was declined in #43, and I am
not trying to relitigate that — the support burden argument stands on its own.
This is just data, plus an offer.

**Machine:** MacBook Pro 16,2 (Core i5-1038NG7, Iris Plus integrated), macOS 26.5.1.

### The app already compiles for Intel

`arm64` appears in exactly one file in the whole repo — `build.sh`, where
`TARGET` is hardcoded. Pointing that at `$(uname -m)` was enough to produce a
native x86_64 build. All 392 source files compiled with **zero source changes**,
no `#if arch()` shims, no disabled features. `./build.sh --dev` completed, the
bundle signed, and `--selftest` printed `SELFTEST OK`.

So the architecture gate is a packaging and support decision rather than a code
one. That part I think is already known, and it is not the interesting finding.

### The part that would have shipped broken

Sensor discovery filters SMC keys by Apple Silicon prefixes:

```swift
name.hasPrefix("Tp") || name.hasPrefix("Te") || name.hasPrefix("Tg")
```

On Intel that matches **zero of the 38 temperature keys** the SMC exposes. Intel
Macs use `TC*` for CPU, `TG*` for a discrete GPU, and keep integrated graphics on
the package as `TCGC`. Only battery (`TB0T`) survives, by coincidence of prefix.

The practical result is that anyone who forced an Intel install would get a
System Monitor with no CPU or GPU temperature at all, and would likely report it
as "monitoring is broken on Intel" rather than as a sensor mapping gap. The same
hardcoded prefix appears in three places: `SystemMonitor.prepareSensorsIfNeeded()`,
[...]

Issue #416, opened 2026-08-01
[Bug]: You can’t open the application “Vorssaint.app” because this application is not supported on this Mac.


EyadElshaer:
### Description

I was trying to download and install the application but every time I try to I get this error is it not supported on Intel based MacBooks ?

### Steps to Reproduce

1. Downloaded the DMG
2. Double clicking on the application gives me the application

### Screenshots / Recordings

<img width="275" height="237" alt="Image" src="https://github.com/user-attachments/assets/219caaad-8b69-447d-9e07-b5d4eba05134" />

### Vorssaint Version

3.2.0

### macOS Version

macOS 15 (Sequoia)

### Installation Method

DMG

Jump-333:
No, you need an Apple Silicon Mac for it to work

p3P4:
https://github.com/vorssaint/vorssaint-utils#what-you-need

vorssaint:
Vorssaint currently requires Apple silicon; Intel Macs aren't supported. The requirement is listed in the README, so I'm closing this as answered.

Issue #181, opened 2026-07-07
[Bug]: Can't install on x86 macos


itvincent-git:
### Description

Download the latest .dmg file, but can't install on my x86 mac

### Steps to Reproduce

Download the latest .dmg file, but can't install on my x86 mac

### Vorssaint Version

3.1.4

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

vorssaint:
Hello!

Intel support is not available yet. Coming later in July!

PR #1048, opened 2026-08-26
feat(build): support building and running on intel macs

Fguarien:
## Summary

Intel has been declined before, in #43 and again on #55, for reasons that were about
support cost rather than about the code: no Intel machine to test on, permissions
behaving differently, and a solo maintainer carrying the bug reports. I have an Intel
Mac and I read that thread before writing this, so here is what is different, and the
honest case for and against.

**What was actually wrong.** Only one line stops the build: `build.sh` hardcodes
`TARGET="arm64-apple-macosx14.0"`. There is no `#if arch(arm64)` anywhere in
`Sources`, and all 403 files compile for `x86_64` untouched. #55 changed that line and
nothing else, which would have shipped an app that launches on Intel and then shows no
CPU temperature, no GPU temperature and no power — the exact class of report the
decline was trying to avoid. Those three are what the rest of this branch fixes:

- **`SMCValueCodec` decoded no power value on Intel.** It handles `sp78` and `fpe2` by
  name and falls through everything else in the same family. Intel publishes its power
  sensors as `spa5`, so `PSTR` and `PCPC` returned nil and `--selftest` printed
  `no power metrics on this Mac`. The codec now decodes generic `sp..`/`fp..`: the last
  character is the fractional bit count in hex. Apple Silicon only uses the two already
  handled, so nothing changes there.
- **The temperature selector only knew the Apple Silicon namespace.** Intel CPU sensors
  are `TC..` and GPU sensors are `TG..`; `Tp..`/`Te..` exist on an Intel Mac too but are
  power-supply and enclosure probes. Matching them, as the generic path did, displayed a
  case temperature as the CPU temperature — wrong values rather than missing ones, which
  is worse. A `.intel` platform now selects `TC..`/`TG..` and prefers the die sensors
  (`TC0c..TC9c`, `TCXc`) over proximity ones, which lag the die by ten degrees or more.
- **Three other places carried their own copy of those prefixes** — `--selftest`,
  `--sensors` and `FanControlHardware.discoverTemperatureKeys`. They all now ask the
  selector, so this cannot drift out of step again. Before that commit `--sensors` on
  Intel listed five enclosure probes as `cpu-aux` and no GPU at all, which is the first
  thing anyone runs when the readings look wrong.
- **One unit test asserted a libdispatch implementation detail.** It blocks 64 threads
  and asserts the pool is starved before testing that `BoundedProcessRunner` survives a
  starved pool. 64 does starve it on Apple Silicon; on this Intel Mac the pool keeps
  handing out workers, so the test failed on its own precondition instead of on the
  behaviour under test. Reproduced standalone, outside the app: `occupiedWorkers=64,
  poolIsStarved=false`, 16 logical CPUs. It now blocks in batches until the probe really
  is stranded, which tests the same thing on both.

**What this does and does not commit you to.** `build.sh` builds for the host, so your
arm64 machine and your arm64 CI runners produce exactly what they produce today. The
release workflow is untouched and no Intel binary reaches anybody through your
distribution identity — the only person who gets an Intel build is one who cloned the
repo and ran `./build.sh` themselves. That is deliberate: it separates "the source
builds on Intel" from "the project ships and supports Intel", and this branch only asks
for the first. If you would rather it did not even say that in the README, drop the docs
commit; the code stands without it.

**The argument against, which I think is still real.** It is more surface, and the part
you named — permissions — is the part I cannot close for you. I could not find a
permission that behaves differently here, but "I did not find one" on one machine is not
evidence, and I say below exactly which machine that was. If the answer is still no, the
measurements in the Verification section are the useful part and I would rather they
lived on #43 than nowhere; say the word and I will put them there and close this.

## Verification

Ran on a **Mac Pro (Late 2013), MacPro6,1**, Xeon E5-1680 v2 8-core, dual FirePro D700,
**macOS 15.7.9 (24G830)**, single display, one Space, English.

**This machine runs macOS 15 through OpenCore Legacy Patcher** — MacPro6,1 is out of
Apple's support list at Monterey. That is the biggest gap between this hardware and
yours and you should weigh it: the SMC and IOKit paths this branch touches are firmware
and kernel level and are not what OCLP patches, but I cannot promise it changed nothing.
A check on a natively supported Intel Mac (2019 iMac, 2018-2020 MacBook Pro, 2018 Mac
mini) would be worth more than anything I ran here.

```
./build.sh          # release, x86_64, 0 warnings
./build.sh --test   # TESTS OK (8641 checks)
./build/Vorssaint --selftest
SELFTEST OK         # before this branch: "SELFTEST WARNING: no power metrics on this Mac"
```

`--sensors`, abridged:

```
component    key   type   °C
cpu-aux   TC0P  sp78   80.00
cpu-core  TC0c  sp78   88.00
cpu-core  TCXc  sp78   94.53
gpu       TG0D  sp78   80.62
gpu       TG1D  sp78   85.00
```

Power decoding, read straight from the SMC and cross-checked against the raw bytes:
`PSTR` = 227.5 W system, `PCPC` = 89.2 W CPU package. The machine was genuinely busy
during these runs, which is why the temperatures are high.

Installed as a release build in `/Applications` and run from the menu bar: the System
tab shows CPU and GPU temperature and power where it showed nothing before.

**What I could not test.**

- **Permissions.** Nothing in this branch touches TCC, and Accessibility and Screen
  Recording were granted and stuck across rebuilds with the self-signed identity. But I
  did not go through every permission-gated feature, and this is the concern you raised
  on #43, so treat it as unaddressed rather than as cleared.
- **Anything battery.** This is a desktop with no battery, so the `TB[0-9]T` path and
  everything in `PowerSampler` that needs a battery is untouched by my testing. An Intel
  notebook is the machine that would exercise it.
- **Fan control writes.** Read-only telemetry only. I did not exercise manual fan
  control on Intel.
- **Any Intel Mac other than this one.** Different Intel families expose different `TC..`
  keys; the die-key rule is written to cover `TC0c..TC9c`, `TCXc/TCXC` and `TC0D/TCXD`
  rather than a fixed list, but I have one machine's key set to go on.
- **Notarised distribution.** Local self-signed builds only.

One thing I noticed and deliberately did not fix, to keep this to one topic: `--sensors`
prints `TC0L` and `TG0L` — `ui16` limit registers, not temperatures — because the
plausibility filter is a value range and not a type check. That predates this branch and
shows on Apple Silicon's equivalents too. Happy to send it separately.

## Anything else

The docs commit is separate on purpose and can be dropped on its own.

I did not touch `CHANGELOG.md`.

Refs #43
Refs #150
Refs #523
Refs #812
Refs #181
Refs #416
Refs #127
Refs #153

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,8 @@ cd vorssaint-utils
 ./build.sh --install               # install into /Applications and launch
 ```
 
-You need macOS 14 or newer, Apple Silicon and the Xcode Command Line Tools. The
+You need macOS 14 or newer, an Apple Silicon or Intel Mac, and the Xcode Command
+Line Tools. The
 build is a plain `swiftc` invocation, see `build.sh`, with no Xcode project and
 no external dependencies, reproducible by design. `Package.swift` is there so
 SwiftPM aware editors can index the code.
--- a/README.md
+++ b/README.md
@@ -203,8 +203,25 @@ The shelf and almost every quick toggle need no permission at all. Finder cut an
 
 ## What you need
 
-- A Mac with Apple Silicon
+- A Mac with Apple Silicon, or an Intel Mac (this fork — see below)
 - macOS 14 Sonoma or newer
+
+### Intel support (this fork)
+
+Upstream builds for `arm64` only. This fork builds and runs natively on Intel:
+
+- `build.sh` targets the host architecture (`VORSSAINT_ARCH=arm64|x86_64` to override)
+  instead of hardcoding `arm64`.
+- The SMC value decoder handles the whole generic `sp…`/`fp…` fixed-point family.
+  Intel Macs publish their power sensors as `spa5`, which decoded to nothing before —
+  which is why every Intel Mac reported "no power metrics".
+- Temperature sensors: Intel uses a different SMC namespace (`TC…` for the CPU,
+  `TG…` for the GPU) than Apple Silicon (`Tp…`/`Te…`/`Tg…`). The selector now knows
+  both, and prefers the Intel die sensors (`TC0c…TC9c`, `TCXc`) over the proximity
+  ones, which lag the die by ten degrees or more.
+- One unit test asserted that 64 blocked threads starve the libdispatch pool. They
+  do on Apple Silicon; an Intel Mac keeps handing out workers past that, so the test
+  now blocks in batches until the pool really is starved.
 
 ### Build it yourself
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
   <a href="https://github.com/vorssaintapp/vorssaint-utils/releases"><img src="https://img.shields.io/github/v/release/vorssaintapp/vorssaint-utils?label=release&color=4c8dff" alt="Latest release"></a>
   <a href="https://github.com/vorssaintapp/vorssaint-utils/releases"><img src="https://img.shields.io/github/downloads/vorssaintapp/vorssaint-utils/total?color=4c8dff" alt="Downloads"></a>
   <a href="https://github.com/vorssaintapp/vorssaint-utils/actions/workflows/ci.yml"><img src="https://github.com/vorssaintapp/vorssaint-utils/actions/workflows/ci.yml/badge.svg?branch=main&event=push" alt="CI status"></a>
-  <a href="#what-you-need"><img src="https://img.shields.io/badge/macOS-14%2B%20Apple%20Silicon-black" alt="macOS 14 and newer, Apple Silicon"></a>
+  <a href="#what-you-need"><img src="https://img.shields.io/badge/macOS-14%2B%20Apple%20Silicon%20%7C%20Intel-black" alt="macOS 14 and newer, Apple Silicon or Intel"></a>
   <a href="LICENSE"><img src="https://img.shields.io/badge/license-GPL--3.0--or--later-blue" alt="License GPL 3.0 or later"></a>
 </p>
 
@@ -203,25 +203,8 @@ The shelf and almost every quick toggle need no permission at all. Finder cut an
 
 ## What you need
 
-- A Mac with Apple Silicon, or an Intel Mac (this fork — see below)
+- A Mac with Apple Silicon or Intel
 - macOS 14 Sonoma or newer
-
-### Intel support (this fork)
-
-Upstream builds for `arm64` only. This fork builds and runs natively on Intel:
-
-- `build.sh` targets the host architecture (`VORSSAINT_ARCH=arm64|x86_64` to override)
-  instead of hardcoding `arm64`.
-- The SMC value decoder handles the whole generic `sp…`/`fp…` fixed-point family.
-  Intel Macs publish their power sensors as `spa5`, which decoded to nothing before —
-  which is why every Intel Mac reported "no power metrics".
-- Temperature sensors: Intel uses a different SMC namespace (`TC…` for the CPU,
-  `TG…` for the GPU) than Apple Silicon (`Tp…`/`Te…`/`Tg…`). The selector now knows
-  both, and prefers the Intel die sensors (`TC0c…TC9c`, `TCXc`) over the proximity
-  ones, which lag the die by ten degrees or more.
-- One unit test asserted that 64 blocked threads starve the libdispatch pool. They
-  do on Apple Silicon; an Intel Mac keeps handing out workers past that, so the test
-  now blocks in batches until the pool really is starved.
 
 ### Build it yourself
 
--- a/Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
+++ b/Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
@@ -313,14 +313,17 @@ final class FanControlHardware {
         if let temperatureKeys { return temperatureKeys }
         let keys = client.keys { name in
             TemperatureSensorSelector.isCPUTemperatureKey(name, platform: temperaturePlatform)
-                || name.hasPrefix("Tg")
+                || TemperatureSensorSelector.isGPUTemperatureKey(name, platform: temperaturePlatform)
         }
         let result = TemperatureKeys(
             cpu: keys.filter {
                 TemperatureSensorSelector.isCPUTemperatureKey($0.name,
                                                               platform: temperaturePlatform)
             },
-            gpu: keys.filter { $0.name.hasPrefix("Tg") }
+            gpu: keys.filter {
+                TemperatureSensorSelector.isGPUTemperatureKey($0.name,
+                                                              platform: temperaturePlatform)
+            }
         )
         temperatureKeys = result
         return result
--- a/Sources/Vorssaint/Services/FanControl/FanControlSupport.swift
+++ b/Sources/Vorssaint/Services/FanControl/FanControlSupport.swift
@@ -392,6 +392,21 @@ enum SMCValueCodec {
                 raw |= UInt64(byte) << UInt64(offset * 8)
             }
             return Double(raw) / 65_536.0
+        case let fixedPoint where bytes.count == 2
+            && (fixedPoint.hasPrefix("sp") || fixedPoint.hasPrefix("fp")):
+            // Generic "sp"/"fp" fixed point: the last character is the number of
+            // fractional bits in hex ("spa5" -> 5, "sp87" -> 7). Apple Silicon
+            // only ships sp78 and fpe2, both handled above, but Intel SMCs
+            // publish their power and current sensors (PSTR, PCPC, PC0C…) in the
+            // rest of the family — without this they decode to nothing and the
+            // app reports "no power metrics" on every Intel Mac.
+            guard let fractionBits = Int(String(fixedPoint.suffix(1)), radix: 16),
+                  fractionBits <= 15 else { return nil }
+            let raw = UInt16(bytes[0]) << 8 | UInt16(bytes[1])
+            let magnitude = fixedPoint.hasPrefix("sp")
+                ? Double(Int16(bitPattern: raw))
+                : Double(raw)
+            return magnitude / Double(1 << fractionBits)
         default:
             return nil
         }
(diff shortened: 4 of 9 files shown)


PathGao:
Thanks for the detailed PR description and for diving deep into the hardware telemetry logic for Intel Macs.

To be completely transparent, we are very likely not going to accept Intel Mac support into this main repository.

While we understand your point about separating "source builds" from "official distribution," merging this PR presents several significant challenges for us as long-term maintainers:

Code Debt & Complexity: Accepting these changes introduces a substantial amount of Intel-specific logic (such as TC../TG.. sensor mapping and spa5 codec handling) into the codebase that we cannot natively support, test, or validate.

Refactoring & Maintenance Overhead: Whenever we refactor sensor architecture, update telemetry pipelines, or add new Apple Silicon features in the future, we would have to be extra cautious not to break the Intel paths, or risk unknowingly shipping broken code since we lack the hardware to verify it.

Non-Native Test Environment (OCLP Risks): We noticed that your testing was conducted on a Mac Pro (Late 2013) running macOS 15 via OpenCore Legacy Patcher (OCLP). Because OCLP injects custom kernel extensions, spoofed SMC responses, and driver patches to support dropped hardware, this environment behaves quite differently from a standard, natively supported Intel Mac. It is nearly impossible for us to determine whether a sensor behavior or edge case is standard Intel behavior or an artifact of OCLP, creating unpredictable testing risks.

Upstream Maintenance Burden: We recognize that keeping a personal fork synced with upstream updates can be tedious (resolving merge conflicts and manually applying patches with each of our releases). However, bringing this logic into the main repo effectively shifts that long-term sync and maintenance burden back onto us.

Could you clarify your main intention with this PR—is it primarily to avoid having to manually sync and maintain your own local fork against our updates, or is there a specific workflow requirement you are trying to solve?

If you are looking for an Intel-compatible build or want to collaborate with others on Intel support, there are community members working on this—you might want to check out or contribute to vorssaint-intel.

We truly appreciate your effort and honest breakdown in the PR description, but we want to be upfront about our project direction so you don't spend unnecessary time on this branch.

PathGao:
Direction is still what I said above and I'm not reopening it here. This is the code review, so it exists if the answer turns out to be yes.

Checked the two structural claims against `main` and both hold. `Tp`/`Te`/`Tg` occur in exactly four places — `SystemMonitor.swift:902-911`, `FanControlHardware.swift:316,323`, `SelfTest.swift:47,167,178`, and the selector — and this branch routes all four through the selector, nothing left behind. `arm64` occurs once in `build.sh`, line 42, and all four `swiftc` invocations read `$TARGET`.

**1. Nothing new is tested.** `MetricsTests.swift:1280-1294` already covers `platform(brandString:)` and `isCPUTemperatureKey` per family, and `:10799-10805` covers the codec. This branch adds a platform case, a key-shape rule and a whole codec family, and adds no assertion to either block. All four are pure functions — they run on arm64 CI with no Intel Mac attached, which is the one part of "we cannot test or validate" that a patch can actually answer.

```swift
expect(TemperatureSensorSelector.platform(
    brandString: "Intel(R) Xeon(R) CPU E5-1680 v2 @ 3.00GHz") == .intel,
    "an Intel brand string selects the Intel platform")
expect(!TemperatureSensorSelector.isCPUTemperatureKey("Tp01", platform: .intel)
        && TemperatureSensorSelector.isCPUTemperatureKey("TC0c", platform: .intel),
       "Intel reads CPU temperature from TC, not from Tp")
expect(TemperatureSensorSelector.isCPUCoreKey("TCXc", platform: .intel)
        && !TemperatureSensorSelector.isCPUCoreKey("TC0P", platform: .intel),
       "the die sensor is preferred over the proximity one")
expect(SMCValueCodec.decode([0x1c, 0x70], type: "spa5") == 227.5,
       "generic sp fixed point takes its fraction width from the last character")
```

The last value is the `PSTR` = 227.5 W from your Verification section, `0x1c70 / 2^5`.

**2. `build.sh` now follows the shell, not the machine.** `uname -m` reports the calling process's architecture, so a Rosetta shell on an Apple Silicon Mac — x86_64 Homebrew, an x86_64 iTerm, `arch -x86_64 zsh` — silently builds and installs a translated binary. The body says arm64 machines "produce exactly what they produce today"; that holds only while the shell is native. Ask the hardware instead:

```sh
ARCH="${VORSSAINT_ARCH:-$([ "$(/usr/sbin/sysctl -n hw.optional.arm64 2>/dev/null)" = 1 ] && echo arm64 || echo x86_64)}"
```

**3. The starvation loop can add five minutes to a *passing* Intel run.** The 64 `poolOccupied.wait(timeout: .now() + 5)` results are discarded now, and the round that finally starves the pool is precisely the round whose blocks never start — so each of those waits burns the full 5 s, up to 64 × 5 s before the probe is even tried. A timeout there already proves the pool refused a worker, which is the thing being looked for:

```swift
for _ in 0..<64 {
    if poolOccupied.wait(timeout: .now() + 5) == .timedOut { break }
}
```

Reasoned from the code, not measured.

Smaller: `decode` gained the generic `sp`/`fp` family, `encode` did not. Fine for read-only telemetry, and Intel fan targets are `fpe2` so writes are unaffected — but a fan key of some other `fp..` type would now read back correctly and fail to write, landing in the "fan control writes untested" gap instead of being caught.

**Not verified:** no Intel machine here. The `spa5` decode I checked by hand against your bytes; the `TC?c`/`TCX?`/`TC?D` rule against other Intel families I did not check at all.

**Impact:** none on the app for Apple Silicon — every M-family branch of the selector returns what it returned, and the new codec case sits below `sp78`/`fpe2`. The one Apple Silicon path whose behaviour does change is `build.sh`.
**Merge:** needs the selector and codec assertions first.

neural-beat:
One data point for `isGPUTemperatureKey`, from an Intel Mac with **integrated**
graphics and no discrete card.

Your comment scopes the helper to "one block per discrete GPU", and the
implementation matches `TG` on Intel. On integrated-only machines the GPU is not
a `TG*` block at all: it sits on the CPU package as `TCGC`. Enumerating the SMC
on mine (MacBook Pro, Core i5-8257U, Iris Plus):

```
TCGC  sp78  58.0 °C      integrated GPU
TG*                       0 keys present
```

So with this PR as it stands, that class of machine still shows no GPU
temperature, and `TCGC` keeps falling into the CPU set through the `TC` prefix —
where it can be returned as the CPU temperature, since it is not a core key and
`displayedCPUTemperature` takes the maximum of whatever is valid when no core
sensor answers.

The one-line version would be:

```swift
static func isGPUTemperatureKey(_ key: String,
                                platform: CPUTemperaturePlatform) -> Bool {
    if key.hasPrefix("Tg") { return true }
    guard platform == .intel else { return false }
    return key == "TCGC" || key.hasPrefix("TG")
}
```

and excluding it from the CPU filter so the two sets stay disjoint.

This is not my finding: @memohammedalmazyad-crypto documented `TCGC` as the
integrated GPU in the sensor dumps in #523, and exelban/stats reads it for
`type = .integrated` in `Modules/GPU/reader.swift`. I have only run into it
because my hardware has no discrete GPU to hide it.

Happy to test on integrated Intel graphics if that is useful, whatever this PR's
fate is.

glyphiglyphi:
Hi @PathGao you mentioned in your comment:
> If you are looking for an Intel-compatible build or want to collaborate with others on Intel support, there are community members working on this—you might want to check out or contribute to vorssaint-intel.

I’ve searched GitHub for a repository named vorssaint-intel or a fork with that specific name, but I can’t find it. Could you please share the direct link to that fork or the specific user repository you were referring to? I’d love to check it out or help contribute if it exists.

UlmHaus:
> Hi @PathGao you mentioned in your comment:
> 
> > If you are looking for an Intel-compatible build or want to collaborate with others on Intel support, there are community members working on this—you might want to check out or contribute to vorssaint-intel.
> 
> I’ve searched GitHub for a repository named vorssaint-intel or a fork with that specific name, but I can’t find it. Could you please share the direct link to that fork or the specific user repository you were referring to? I’d love to check it out or help contribute if it exists.

I found these (I am not affiliated with either project in anyway):
https://github.com/HectorPalomeque/vorssaint-utils-intel

PathGao:
I don't have a link. "vorssaint-intel" was me writing a name as if it were a real repo I could point you at, and I can't — that was my mistake, not something you failed to find.

The fork @UlmHaus turned up is not one I've read or vouch for. It's an unofficial build of an app that reads the SMC and asks for Accessibility, so treat it on its own merits, not on the strength of a name I dropped in a comment.

@neural-beat — your `TCGC` point is right and I never answered it. On this branch `isCPUTemperatureKey` matches any `TC` prefix while `isIntelCPUCoreKey` rejects `TCGC` (third character is neither a digit nor `X`), so on an integrated-only Mac it lands in the CPU fallback set and the GPU set comes back empty. Whether it then gets *displayed* as the CPU temperature depends on the `.intel` branch of `displayedCPUTemperature`, which this diff doesn't touch and whose base I can't read here — but the two key sets failing to be disjoint holds regardless. Worth carrying wherever this code ends up.

neural-beat:
Thanks for circling back on this.

I can't speak for this branch's base, but I checked it on
alebypegasus/vorssaint-utils-intel and the display part does hold there.
Heads up that the names don't match: that fork has .intelFamily and an
intelCPUCoreKeys set behind isCPUCoreKey, and TCGC isn't in it.

The path on their main: SystemMonitor.swift:905 puts TCGC into cpuKeys since it
takes any TC prefix, :910 builds fallbackCPUKeys out of cpuKeys with no
hasCPUCoreSet filter so it stays there, and :944-951 append those readings and
ask again when the preferred set comes back with nothing. At that point
TemperatureSensorSelector.swift:104 returns the max of everything valid.

So yes, on a tick where no die sensor gives a plausible number, what gets shown
as CPU can be the GPU. How often that actually happens I haven't measured, only
that the path is there.

Either way you're right that the underlying problem is the CPU and GPU key sets
overlapping. I put up a PR on that fork for exactly that: one definition of a GPU
sensor in TemperatureSensorSelector, used by both SystemMonitor and
FanControlHardware. The fan control side was only matching "Tg", so it was
missing discrete TG* as well.

PR #1050, opened 2026-08-26
feat(screenshot): Drag and drop in screenshot editor

Yahddyyp:
Refs #677 

## Summary

Replace the screenshot editor's draggable image thumbnail with an icon-based drag-out handle.

## Changes
- Replace the thumbnail with an `arrow.up.doc` drag-out handle
- Keep the drag handle in `infoChip`
- Preserve the existing drag provider and export behavior
- Preserve the 26×18 drag target size
- Add the accessibility and help labels

Screenshot 2026-08-28 at 12 06 51

The translations should be correct, i used google translate so i don't know.


Yahddyyp:
@PathGao Do your magic

PathGao:
The drag-out isn't new. `dragOutHandle` already ships inside `infoChip` in the bottom row, with the same `onDrag` + `ScreenshotService.dragItemProvider` body. This PR moves it into `topBand` and swaps the live thumbnail for an SF Symbol. Worth stating in the body, because it changes what #677 needs: dragging the capture out already works today, it just isn't discoverable down there.

On the dead top half — I think you're fighting the real titlebar, and `mouseDownCanMoveWindow` on a subview loses that fight. Numbers from `main`:

- `topBand` sits at `.padding(.top, 10)` (`ScreenshotEditorView.swift:44`) and is `.frame(height: 38)` (line 83)
- a 24pt control in `actionCluster` lands around y 17–41 in window coordinates
- the window is `.fullSizeContentView` with a standard 28pt titlebar (`ScreenshotEditorController.swift:989`)

28 − 17 = 11 of the control's 24 points sit inside the titlebar strip. That's your bottom half.

Two cheap checks. Bump the `.padding(.top, 10)` on line 44 to 28 and see if the whole symbol goes draggable. Then press-and-drag the *top* edge of the trash or pin button — if the window moves there too, the dead strip belongs to the band, not to your handle, and `WindowDragBlockingBackground` isn't buying anything.

If that's the cause, two things follow: the blocking view can go, and pushing the band down 18pt needs `ScreenshotSupport.editorContentSize`'s hardcoded `chrome = CGSize(width: 96, height: 140)` (line 802) to grow by the same amount, or fresh windows open with 18pt less canvas. The option that costs nothing is leaving the handle in `infoChip` where it already works and only changing how it looks.

Minor: `.screenshotSafeHelp(strings.editorTitle)` was tolerable under a thumbnail. On a bare icon the tooltip now reads as the window title on a control whose only job is drag-out. Every other icon in `actionCluster` also carries an `.accessibilityLabel`; this one doesn't.

Not verified: no hardware here. The geometry above is read off the source and the 28pt titlebar is the standard value, not something measured against your build.

**Impact:** none shipped — draft, and the handle's top half is dead by your own report.
**Merge:** needs the dead top half resolved first, or the handle moved back to the info chip.

ref #677

Yahddyyp:
@PathGao This should be everything you asked for in the converstations we had.

PathGao:
Handle is back in `infoChip`, so the dead top half is gone. All 13 locales got `dragOutHandleLabel` and the memberwise init enforces that, so there is no missing-locale gap to check.

Translations read fine to me — none of the 13 is wrong. One inconsistency: `"Drag and Drop"` is title case where the rest of that table is sentence case (`"Screen capture"`, `"Edit clipboard image"`, `"Copy an image first"`).

The thing I would check before merging is chip width. The body lists "Keep the drag handle in infoChip" as if nothing moved, but the chip goes from a 26×18 thumbnail to an icon plus a localized word pair, and German `"Ziehen und ablegen"` is roughly double the English. `bottomRow` (`ScreenshotEditorView.swift:865`) carries the comment "the chips can never collide with the style bar on a narrow window", and the floor is 760pt content width (`ScreenshotSupport.swift:791`). `Text(dimensionsLabel)` beside it is `.fixedSize()` (line 1108); the new `Label` is not, so under a squeeze the label is what truncates — "Ziehen und…" sitting next to the dimensions.

Check it in German, window dragged to minimum width, with a shape selected so sticker + colour + stroke + layer + shadow are all up at once. If it collides, the cheap fix is `.labelStyle(.iconOnly)` and leave the words in the help and accessibility label — a bare `arrow.up.doc` is still more discoverable than the thumbnail was.

Not verified: no hardware here, and I can't measure rendered text width from source. The 760pt floor and the `.fixedSize()` asymmetry are read off main.

Body still says `Resolves #677`. That closes the reporter's issue the moment this merges — `Refs #677` instead.

**Impact:** on a narrow window in a long-word locale the new label may truncate or crowd the style bar; unverified.
**Merge:** needs `Resolves #677` → `Refs #677` in the body first.

ref #677

Yahddyyp:
@PathGao Anything else you need me to do other than the ref part

PathGao:
Body already reads `Refs #677`, so that part's done. Two left, both in the same edit.

1. Chip width — the only one that can actually bite. In `infoChip` your `Label` sits next to `Text(dimensionsLabel)`, and that `Text` is `.fixedSize()` (`ScreenshotEditorView.swift:1107`) while the `Label` is not. Under a squeeze yours is what truncates. Either check it (German, window dragged to the 760pt floor, a shape selected so sticker + colour + stroke + layer + shadow are all up) or skip the check and ship `.labelStyle(.iconOnly)`, keeping the words in the help and accessibility label. The second costs nothing and I would take it.

2. Case — `"Drag and Drop"` (en) and `"Sürükle ve Bırak"` (tr) are title case; the rest of both tables is sentence case (`"Screen capture"`, `"Ekran yakalama"`). Lowercase the middle words. The other 11 locales are fine.

While you're in the file, `dragOutHandle`'s closing brace picked up four extra spaces.

Not verified: no hardware here, and I can't measure rendered text width from source. The `.fixedSize()` asymmetry and the 760pt floor are read off main.

**Impact:** on a narrow window in a long-word locale the label may truncate or crowd the style bar; unverified.
**Merge:** needs `.labelStyle(.iconOnly)` or a German narrow-width check first.

ref #677

Yahddyyp:
@PathGao I fixed up the title case in en and tr and also added `.labelStyle(.iconOnly)` instead of the first option you gave me.

PathGao:
Title case and `.iconOnly` are both in, and all 13 locales have the string.

One thing `.iconOnly` created — my suggestion created it, so I'll own it. The drag target is now the glyph. The `Label` gets no `.frame` and no `.contentShape`, so at `size: 11` you're grabbing roughly 11×13pt where the thumbnail it replaces was `.frame(width: 26, height: 18)` (`ScreenshotEditorView.swift:1183` on main). Every other icon control in this file pairs the font with an explicit box: `annotationShadowButton` is 25×24 + `.contentShape` (lines 1028/1036), `zoomButton` is 30×20 (line 1149), same shape again at 589, 632, 1079. Yours is the only one without one.

One line, restores the old target exactly:

```swift
.frame(width: 26, height: 18)
.contentShape(Rectangle())
```

Nit, not a blocker: the body still says "labelled arrow.up.doc control" and the screenshot still shows the version with words. Neither of those ships now.

Not verified: no hardware here. The sizes are read off main and the diff; I haven't seen the rendered glyph.

**Impact:** the drag-out is harder to grab than the thumbnail it replaces, which is the thing this PR exists to make easier to find.
**Merge:** needs the `.frame` + `.contentShape` line first.

ref #677

Yahddyyp:
image

Alright so i added 
``` swift
.frame(width: 26, height: 18)
.contentShape(Rectangle())
```
Just like you told me to so in the image provided above is how it looks like do you want to run with this @PathGao

PathGao:
Yes. `.frame(width: 26, height: 18)` + `.contentShape(Rectangle())` puts the hit area back at exactly what the thumbnail had on main (`ScreenshotEditorView.swift:1183`), and `.iconOnly` means there is no text left to truncate, so the German narrow-window question is off the table too.

Read the rest of the diff again, nothing left to flag. All 13 locales carry `dragOutHandleLabel`, en and tr now match the sentence case of the rest of their tables, and the drag provider plus `markExported()` are untouched.

Swift 6.0.3 compatibility hasn't reported a state yet. Build & selftest is green and this adds a `Label` and two modifiers, nothing that needs a newer toolchain, so I'm not treating that as a gate.

Not verified: no hardware here. I'm going off your screenshot and the sizes read from main.

**Impact:** the chip shows an icon instead of a live thumbnail; drag target size and drag-out behaviour are unchanged.
**Merge:** ship it.

ref #677

PathGao:
Scope note for anyone reading this later, now that it's merged.

The drag-out itself is not new here. `dragOutHandle` landed in `69d44754`, which `v3.3.1` and `v3.3.2` both contain, and the chain inside it — `.onDrag` → `commitEditingTextIfNeeded()` → `model.exportImage()` → `ScreenshotService.dragItemProvider(...)` → `model.markExported()` — is byte-for-byte the same before and after this PR. What changed is the affordance: a live thumbnail of `model.baseImage` became `Label(strings.dragOutHandleLabel, systemImage: "arrow.up.doc")` with `.labelStyle(.iconOnly)`, plus the new `dragOutHandleLabel` string in all 13 locales, `.screenshotSafeHelp` and `.accessibilityLabel` pointing at it instead of the editor title, and the hit area held at the old 26×18 by the `.frame` + `.contentShape` pair.

That matters for #677: the reporter is on 3.3.2 and can drag today. The old control just looked like a preview and its tooltip never said otherwise. Worth keeping in mind when this reaches the changelog — filed under "Added" it would tell 3.3.2 users they don't have something they do have. "Changed" describes it: the drag-out handle is now an icon that says what it does.

ref #677

Issue #836, opened 2026-08-23
Bug Summary: all 76 open bug issues audited against main


PathGao:
I went through all 76 open bug issues against `origin/main` @ `5c537230`, re-audited 2026-09-02T15:04:20Z. Every "fixed / not fixed" call below comes from reading the code in that tree, not from the CHANGELOG, a PR title, or the `fixed (pending release)` label. The `bug` label now sits on 77 issues; the 77th is this summary itself, which is not audited below.

The baseline moved from `a2fe2216` to `5c537230`, 183 files changed. What moved: twelve reports arrived, the reporters on #728 and #1030 answered, and the first section went from three entries to seven — #938 was labelled `half` by hand, while #1047, #512 and #1119 each picked up a fix label with the thing the reporter named still sitting in the code.

Two companion writeups are open alongside this one: #837 PR Summary, #838 Enhancement Summary.

## Disposition at a glance

| Action | Issues |
|---|---|
| Finish the half the label already signed for | 888, 828, 341, 938, 1047, 512, 1119 |
| Root cause known, needs a commit | 937, 1211, 613, 619, 780, 1094, 369, 1109, 1079, 692, 835, 1273, 783, 1277, 865, 1283, 301, 505 |
| Still unlocated, one suspect each | 856, 920, 1166 |
| Reply with an answer, no code | 950, 1091, 728 |
| Ask the reporter one named question | 458, 929, 879, 612, 610, 1057, 1059, 1043, 1030, 1261, 1286, 1151, 1150 |

---

## Labelled fixed, but only half of it is

Seven issues carry a label saying the work is done while the thing each one reported is still there. Whoever installs the next release finds the problem where they left it, and has no obvious way to say so.

**A Hyper key reaches the drag now, and still cannot mean "move" — #888**
Set "move and resize windows by dragging" to a Hyper key and it only ever resizes. The chord is at least readable now, which is the half that was fixed. The other half is untouched: the app strips Shift out of whatever combination you saved before comparing, so a held Hyper key comes out identical to the built-in resize combination and can never equal the move one. Anyone who remaps Caps Lock to Hyper — a common setup — gets a feature that does half of what it says.
Next: let a saved combination keep Shift, or take the label off.

**The clamshell permission is granted, the switch itself is still unchecked — #828**
[...]

PR #1051, opened 2026-08-26
ci(issues): fix the hint job's permission, and stop looking commits up in tags

PathGao:
Three changes to `issue-fix-status.yml`.

### The hint job has never posted a single comment

It ends in `gh pr comment`, which is the GraphQL `addComment` mutation, and GitHub checks that against `pull-requests` — not against `issues`, which is all the workflow grants. Eighteen of the last hundred runs failed on that one line with `Resource not accessible by integration (addComment)`; the green runs are the ones that had nothing to say and exited before reaching it. Searching the repository for the job's own marker finds one hit, and that is the description of the pull request that added it.

The merge job beside it uses `gh issue comment`, a REST call that `issues: write` does cover, which is why that half has worked all along — thirteen issues carry its comment. Fix is `pull-requests: write` in the permissions block. Nothing else about that job changes.

I considered rewriting the call as `gh api repos/…/issues/N/comments` to avoid widening permissions at all, and did not: the permission this job needs is exactly the one it is being denied, and hiding a pull request comment behind the issues endpoint would leave the next reader with the same puzzle.

### The release job no longer asks which tag carries which commit

It used to read a merge sha out of the bot's own comment and run `git tag --contains` per issue, moving the label only on a hit. That bought accuracy in one narrow window — a pull request merged after the tag was cut but before the release finished building — and charged for it everywhere else: an issue whose label arrived with no recorded commit was skipped every release, forever. Every label applied by hand is in that set.

Now the label moves for everything that is waiting. A fix on main goes out with the next stable release, so that release is the one to ask about. **Twenty-six open issues carry `fixed (pending release)` today and would move on the next release.** The window that is given up costs one reporter one early ask, and their answer settles it either way — the comment already asks them to say how much of it is fixed, and a "still happening" sends the issue back.

The `actions/checkout` goes with it: that job reads issues and releases, and no repository content. The workflow now uses no actions at all, which is one less thing for the allowlist to have to permit.

The two-week job is unchanged. It reads the timestamp of the comment that asked, which is not a commit, and its clock is unaffected.

### Summary issues are out of all of it

A `summary` issue is a long-lived index. Nobody is waiting on a build for one, and a pull request that fixes one line of a fifty-line audit has not fixed the audit — but they are the issues people reference most, so they are the numbers that reach these jobs most often. All four read points now skip the label: the merge job does not label them, the hint job does not ask an author to reference one, the release job does not sweep one that carries the label by hand, and the two-week job never closes one. Seven issues carry it today.

### What is not changed

- The comment text the reporter reads, in either job.
- The `sha=` marker the merge job writes. Nothing reads it now, but it is the record of which commit a label came from, and dropping it would cost that for nothing.
- The label names and colours, and the version still lives in the comment rather than in a label.
- The legacy hand-applied `v3.3.2 reporter check` (7 open) and `reporter check` (0 open). These sit outside the workflow, unchanged by it; collecting them means renaming a label, which retroactively rewrites what it said on issues already carrying it, and that is a decision to make on purpose rather than as a side effect of this.

### Tested

YAML parses; every `run:` block passes `bash -n`; both new `jq` filters and the per-issue guard run against the live repository and return 26 pending, 0 excluded, and a skip on each of the two summary issues tried by hand.

The jobs themselves cannot be exercised before merge. `pull_request_target` reads the workflow from the base branch, so the hint job that runs on this pull request is the one being fixed, not the fix — it failed here with the same `addComment` error, on run 32991189843, which is the whole report reproduced on itself. It also picked out exactly two numbers to ask me to reference, and both were summary issues, which is the third change above arriving on cue. Neither number is written with a `#` in this description any more, so the job now exits before its last line and the check is green; put the `#` back and it goes red again until this merges. `reporter-check` cannot run at all outside a `workflow_run` from Release.

### DIFF
--- a/.github/workflows/issue-fix-status.yml
+++ b/.github/workflows/issue-fix-status.yml
@@ -6,8 +6,9 @@ name: Issue fix status
 #   `fixed (pending release)`, with the merge commit recorded in the bot's
 #   comment
 #
-#   a stable release that contains that commit      ->  the label becomes
-#   `please confirm` and the reporter is asked to check that release
+#   the next stable release                         ->  every issue still
+#   carrying that label becomes `please confirm`, and its reporter is asked
+#   to check that release
 #
 #   two weeks of silence after that                 ->  the issue is closed,
 #   after a comment saying so and how to reopen it
@@ -35,10 +36,14 @@ name: Issue fix status
 # stops there.
 #
 # Neither of the first two decides that anything is fixed. A human decided that
-# when they merged. The label only says where the change is, and
-# `git tag --contains` answers the second one, so a changelog entry is never the
-# evidence. The third decides nothing either: one word from anyone stops it, and
-# it never reads what that word says.
+# when they merged. The label only says where the change is, and the release
+# that follows a merge is the one that carries it, so the second step moves
+# every issue that is waiting rather than looking each commit up in the tag.
+# What that gives up is the pull request merged after the tag was cut and
+# before the release finished building: its reporter is asked one release
+# early, and their answer settles it either way. The third step decides
+# nothing either: one word from anyone stops it, and it never reads what that
+# word says.
 #
 # `please confirm` carries no version. The version lives in the comment, which
 # is the only part of this the reporter ever sees -- labels are silent, they
@@ -74,6 +79,10 @@ on:
 permissions:
   contents: read
   issues: write
+  # `gh pr comment` is the GraphQL addComment mutation, which is checked
+  # against this and not against `issues`. Without it the hint job below fails
+  # at its last line, every time it has something to say.
+  pull-requests: write
 
 concurrency:
   group: issue-fix-status-${{ github.event.pull_request.number || github.event.workflow_run.head_branch }}
@@ -285,11 +294,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
-        with:
-          fetch-depth: 0
-          persist-credentials: false
-
+      # No checkout: this reads issues and releases, not the history.
       - name: Move the label onto the release that carries the fix
         env:
           TAG: ${{ github.event.workflow_run.head_branch }}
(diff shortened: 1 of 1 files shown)


PathGao:
Two things, one before merge.

**The summary guard also silences the closing-keyword warning.** It sits in the shared candidate loop (new line 241), and an empty `hits.txt` two lines later exits with `points at no open issue, nothing to say`. So a body whose only closing keyword is `Closes #836` gets no comment at all, GitHub closes a fifty-line audit index on merge, and nobody is told.

That is the case the job exists for, and your own numbers say it is the likeliest one: summary issues are the numbers that reach these jobs most, and this PR's hint run picked two of them. `missing` is the form that should not ask about a summary issue. `closing` should ask louder.

```bash
if [ "$form" != closing ] && jq -e --arg l "$SKIP_LABEL" 'any(.labels[]; .name == $l)' <<<"$issue" >/dev/null; then
  continue
fi
```

Not `[ "$form" != closing ] || jq ... && continue` — shell has no precedence there, `(A||B) && continue` fires on every closing form.

**The window you give up is wider than "merged after the tag was cut".** `release.yml:53` accepts `ahead:$COMMIT_SHA`, so a tag has to be *on* main, not *at* main. The set now announced as shipped is everything merged after the tagged commit, which is not bounded by when the tag was pushed. The `release-signing` environment gate on top of that makes the tail human latency, not build latency. The comment you deleted named this exact outcome — closed against a version that never shipped the change.

Not arguing for `git tag --contains` back; dropping the checkout is worth keeping. The accuracy is one API call if you want it:

```bash
gh api "repos/$GH_REPO/compare/$sha...$TAG" --jq '.status'   # ahead|identical == carried
```

with no-recorded-sha falling through to move, which is the case this PR is actually opening up. It costs back the paginated comment read per issue, which is the part you deleted. Your call — the 26 hand-labelled ones move either way.

Minor: `pull-requests: write` at workflow level hands it to all four jobs and only `reference-hint` uses it. Nothing exploits it, no job checks out PR code. It just reads oddly against the two "must never execute anything from the pull request" comments right above.

Not verified: none of this ran against a real `workflow_run` or a real merge, same limit you named.

**Impact:** `Closes #<summary issue>` merges with no warning and closes the index. Reporters on issues fixed after the tagged commit are told a release carries their fix when it does not.
**Merge:** needs the summary guard taken out of the `closing` path first.

PathGao:
Both structural points are right. One is fixed in 5262bf40; the other I measured and am leaving as it is, with the number written down.

**The closing-keyword hole is real and is fixed.** The skip sat in the loop both forms share, so `Closes #<summary issue>` emptied the hit list and said nothing, which is the loudest case in the file going silent. It now applies to the `missing` form alone, as one `if` rather than the `||`/`&&` chain that would have fired on every closing form. Checked against the live repository: `form=closing` keeps 836 and 986, `form=missing` keeps 986 only.

**The window: correct about the mechanism, and it has never once happened.** `ahead:$COMMIT_SHA` does allow a tag behind main's tip, so the exposed set starts at the tagged commit rather than at the tag push. Measured against the last five stable releases — tagged commit to release published, and every commit that reached main in between:

```
v3.3.2   27 min   0 commits
v3.3.1   20 min   0
v3.3.0   24 min   0
v3.2.0   15 min   0
v3.1.15  13 min   0
```

Zero every time, and zero means the tag was at main's tip on each of them, so the `ahead:` allowance has not been used. The same query at +1 day and +7 days returns 9 and 167 for v3.3.2, so it is the window that is empty, not the method.

A wrong outcome also needs both halves: a merge inside those minutes *and* the reporter going two weeks without a word. The early ask alone is self-correcting — the comment asks how much of it is fixed, and one reply sends the issue back.

Against that, `compare/$sha...$TAG` costs back the paginated per-issue comment read, which is the part being deleted here, and re-splits the pending set into recorded and not. The twenty-six issues waiting today are hand-labelled with no sha; they are the reason for the change, and they move either way only because of the fall-through. Keeping the accuracy for the half that has a sha while the other half is waved through is the version that reads worst.

Leaving it. If a release is ever cut behind main's tip the number stops being zero, and that is the moment to add the one call back.

**Minor, taken:** `pull-requests: write` is now on the hint job instead of the workflow, and its issues permission drops to read — it reads issues and says nothing to them. The three jobs that write to issues keep the workflow default and no pull-request access at all.

Still not verified, unchanged: no real `workflow_run` and no real merge. `pull_request_target` reads the workflow from the base branch, so this pull request's own hint run exercises the old file either way.

Issue #973, opened 2026-08-25
Option to automatically remove line breaks when using Copy Text from Screen


zgosalvez:
### Problem or Motivation

When using Copy Text from Screen on a long text, the natural line breaks are kept and I would have to manually remove them after pasting. 

### Proposed Solution

Would save a lot of time if there's an option to automatically remove line breaks in the settings.

### Feature Area

Tools

PathGao:
a little niche request，waiting for more people join

zgosalvez:
Works as expected. Thank you. Here's a sample based on a screenshot of your comment.

**With Remove line breaks on:**
> The fix for this is out now, in 3.3.3. @zgosalvez, could you update to that version and say whether it fixed things on your Mac? We cannot tell from here - yours is the setup that found this. Please say how much of it is fixed, not only whether it is. If part of what you reported still happens, or something related turned up, say so here and the issue goes back for another look - this is not a last call. If nothing arrives in two weeks the issue will be closed as it stands, and a comment reopens it at any point after that.

vs

**With Remove line breaks off:**
> The fix for this is out now, in 3.3.3.
> @zgosalvez, could you update to that version and say whether it fixed things
> on your Mac? We cannot tell from here - yours is the setup that found this.
> Please say how much of it is fixed, not only whether it is. If part of what you
> reported still happens, or something related turned up, say so here and the issue goes
> back for another look — this is not a last call.
> If nothing arrives in two weeks the issue will be closed as it stands, and a comment
> reopens it at any point after that.
> 

<img width="1159" height="637" alt="Image" src="https://github.com/user-attachments/assets/9f2f3891-5edb-43d1-a8f5-c23d1b16416b" />

vorssaint:
thanks for checking both settings. removing line breaks is working as requested, so i'm closing this

PR #1052, opened 2026-08-26
feat(ocr): add line break removal option

ywu73:
## Summary

- Add an opt-in setting that joins recognized screen-text lines into a paragraph with script-aware spacing.
- Preserve the current line-break behavior by default.
- Add searchable settings text, backup coverage, all 13 locale strings, and multilingual behavior tests.

## Motivation

Copy Text from Screen currently inserts a newline between every recognized visual line. Long paragraphs therefore require manual cleanup after pasting. This adds the requested paragraph-style output without changing existing users' behavior.

## Changes

- Register a default-off `screenOCRRemoveLineBreaks` preference and expose it under Screen Capture > Copy text from screen.
- Keep the existing reading-order sort and blank-line filtering, then use the reviewed CJK seam ranges: `U+3000...U+9FFF`, `U+F900...U+FAFF`, `U+FF01...U+FF9F`, and `U+20000...U+2FA1F`.
- Keep Korean seams spaced by stopping the halfwidth range before Hangul jamo at `U+FFA0`.
- Require the line-break preference explicitly at the OCR outcome boundary so a future caller cannot silently ignore it.
- Include the setting in Settings search and portable settings backups.
- Add localized title and behavior-neutral caption strings for every supported language.
- Scope the preference to Copy Text from Screen; screenshot-editor text selection continues to preserve its own line breaks.
- Cover the default, line-break preservation, Latin, Chinese, Japanese, halfwidth Japanese kana, Korean, mixed-script seams, settings search, and localization completeness in the test harness.

## Testing

- Ran `./build.sh --test` — `TESTS OK (8805 checks)` after merging the current `main`.
- Ran `./build.sh` — release bundle assembled successfully.
- Ran `./build/Vorssaint --selftest` — `SELFTEST OK`.
- Added the halfwidth-kana seam test first and confirmed the prior revision failed exactly that check: `これは ﾃｽﾄです` instead of `これはﾃｽﾄです`. Extended the reviewed range through `U+FF9F` and reran the full suite successfully.
- Built on a MacBookPro17,1 running macOS 15.7.7 (24G720), using the ad-hoc signing fallback.
- Not manually verified through a live Screen Recording selection and clipboard paste in the installed app.
- The release build reports two pre-existing Swift Sendable warnings in `ShelfService.swift:1494-1495`; this PR does not touch that file.

## Risks

- Low risk; the new preference defaults off, so existing OCR output is unchanged.
- The enabled behavior follows the reviewed Unicode scalar ranges at line seams rather than language detection.
- The setting intentionally does not alter text selection in the screenshot editor.
- Localized strings compile in every catalog but have not received native-speaker review.

## Screenshots / Recordings

- Not included in this draft; the UI change is one toggle and explanatory caption in the existing Copy text from screen settings section.

## Checklist

- [ ] Documentation has been updated where appropriate

Refs #973


PathGao:
Checked the two things that usually rot in this shape and both are clean: `joinedRecognizedText` has exactly one production caller, and `exportKeys()` derives from `Defaults.registeredDefaults` (`SettingsBackupSupport.swift:19`), so backup coverage comes free with the registration — the new `exportKeys()` assertion is testing something already structural. No sibling call site missed.

One real gap.

**The space join is wrong for Chinese and Japanese.** `joinedRecognizedText` inserts a space at every wrap point. Right for Latin scripts, wrong for zh/ja, where a wrapped line has no space at the break — a Japanese paragraph comes back as `これは テスト です`. The ja caption you added (`認識した行をスペースでつなぎ`) promises exactly that behavior, so it is honest, but it is not what someone who turns on 改行を削除 wants. Korean is fine as-is, it already has word spaces.

Laziest fix that leaves the sort and filter untouched — suppress the space when both sides of the seam are CJK:

```swift
// CJK punctuation, kana, Han. Hangul (AC00–D7AF) is out on purpose: Korean keeps its spaces.
private static let tightScripts = CharacterSet(charactersIn: "\u{3000}"..."\u{9FFF}")

.map(\.text)
.reduce("") { joined, line in
    guard !joined.isEmpty else { return line }
    guard removingLineBreaks else { return joined + "\n" + line }
    let tight = joined.unicodeScalars.last.map(tightScripts.contains) == true
        && line.unicodeScalars.first.map(tightScripts.contains) == true
    return joined + (tight ? "" : " ") + line
}
```

Smaller: drop the `= false` on `removeLineBreaks` in `ScreenTextService.outcome`. `detectQRCodes` directly above has no default for a reason, `outcome` has one caller, and the default is a third copy of the off-by-default fact (registration, signature, call site) that lets a future second call site ignore the preference and still compile.

Not verified: I read the diff and `main`, no live capture. The 13 catalogs compile; I only judged wording on the ja caption above.

**Impact:** With the option on, zh/ja captures paste with a stray space at every wrap. Default off, so nobody is affected until they turn it on.
**Merge:** needs the CJK seam decision first — either the tight join above, or say in the PR that the option is Latin-only and the captions should reflect that.

ref #973

ywu73:
Thanks — addressed across `5e4ee79d`, `a5215945`, and `bd447b74`.

- Line removal now uses the reviewed CJK seam ranges, including halfwidth CJK punctuation and katakana through `U+FF9F`, while stopping before halfwidth Hangul jamo at `U+FFA0`.
- Removed the default from `removeLineBreaks` so the production caller must pass the preference explicitly.
- Removed the redundant `exportKeys()` assertion.
- Reworded all 13 captions so they promise line-break removal rather than unconditional space joining.
- Added coverage for Latin, Chinese, Japanese, halfwidth Japanese kana, Korean, mixed-script seams, and the setting-off path.
- Clarified in the PR that the preference applies only to Copy Text from Screen, not screenshot-editor text selection.

The halfwidth-kana test failed on the prior revision as `これは ﾃｽﾄです`; after extending the range through `U+FF9F`, `./build.sh --test` passes 8799 checks, the release build completes, and `--selftest` reports `SELFTEST OK`.

PathGao:
Ranges are right except for one adjacent block. `0xFF01...0xFF60` stops at U+FF60, so halfwidth katakana and halfwidth CJK punctuation (U+FF61–U+FF9F: `｡` `､` `｢` `｣` `･` `ｱ`–`ﾝ`) fall through to the space branch. A Japanese line that wraps on halfwidth kana gets back exactly the space this PR removes. FF9F is also the right place to stop — U+FFA0 starts halfwidth Hangul jamo, which should keep its spaces like the syllables do.

```swift
case 0x3000...0x9FFF,
     0xF900...0xFAFF,
     0xFF01...0xFF9F,
     0x20000...0x2FA1F:
```

Scope question, not a defect: the screenshot editor's text selection copy (`ScreenshotSupport.joinedWords`, `ScreenshotSupport.swift:1776`) joins picked lines with `\n` and never reads the preference. Same clipboard, same recognized screen text, different code path. If the setting is meant to be Copy text from screen only, say so in the PR — the caption doesn't scope it.

Rest is clean. `joinedRecognizedText` still has one production caller on `main`, `reduce(into:)` avoids the O(n²) my version would have had, and no other place in `Sources/` carries a CJK range predicate, so this helper isn't a second copy of an existing fact. The default now lives only in registration and `@AppStorage`, matching `detectsQRCodes`. All 13 captions promise line-break removal rather than space joining.

Not verified: no live capture. I judged wording on the ja/zh/ko captions only. `Build & selftest` has not reported a result on this revision.

**Impact:** With the option on, Japanese text using halfwidth kana pastes with a stray space at every wrap. Default off, so nobody is affected until they turn it on.
**Merge:** needs the `0xFF61...0xFF9F` range first.

ref #973

ywu73:
Thanks — fixed in `bd447b74`.

- Extended the reviewed fullwidth/halfwidth range through `U+FF9F`, stopping before halfwidth Hangul jamo at `U+FFA0`.
- Added a halfwidth-katakana seam regression test; it failed on the prior revision as `これは ﾃｽﾄです` and now passes as `これはﾃｽﾄです`.
- Clarified in the PR description that the preference applies only to Copy Text from Screen; screenshot-editor text selection remains unchanged.

After merging the current `main` in `1dc8193f`, `./build.sh --test` passes 8805 checks, the release build completes, and `--selftest` reports `SELFTEST OK`.

PathGao:
Range is right now. `0xFF01...0xFF9F` covers `｡` `､` `｢` `｣` `･` and `ｱ`–`ﾝ` and stops before `U+FFA0`, and the halfwidth-kana test pins it. Two things left.

**The file the toggle lands in was restructured on `main`, and that is the conflict.** `ScreenCaptureSettings.swift` now drives all four tools from one picker, and `ToolShortcutRows` moved out of `ScreenTextCaptureSettings` up into the shared top section, with a new `keys:` on its init (`ScreenCaptureSettings.swift:44`, `:121`). Your hunk anchors on `ToolShortcutRows(tool: .text)` inside `ScreenTextCaptureSettings`, which no longer exists there. Rebase target is the same relative position with no shortcut row in between: after the `ocrCaption` caption, before the QR toggle (`ScreenCaptureSettings.swift:157`).

**Three Hangul blocks sit inside `0x3000...0x9FFF`.** The PR says "Keep Korean seams spaced by stopping the halfwidth range before Hangul jamo at `U+FFA0`", and the code comment says Hangul stays out on purpose. Both hold for the halfwidth range only. Compatibility jamo `U+3130...U+318F` (`ㄱ` `ㅋ` `ㅎ`) is in the first range, so `재밌었어` / `ㅋㅋㅋ` across a wrap comes back as `재밌었어ㅋㅋㅋ` — the same space this PR is supposed to keep for Korean. One split:

```swift
case 0x3000...0x312F,
     0x3190...0x9FFF,
     0xF900...0xFAFF,
     0xFF01...0xFF9F,
     0x20000...0x2FA1F:
```

Parenthesized and circled Hangul (`U+3200...U+321F`, `U+3260...U+327F`) stay tight under that split. They are list markers, not seam text, and carving them out costs two more ranges — I would leave them.

Smaller, and I have not verified it: `.map(\.text)` keeps the raw line, while the filter above it only trims for the emptiness test. If Vision ever emits a line with a trailing space, the removal path both fails the tight check and appends its own space, so `这是 ` + `测试` pastes as `这是  测试`. Trimming in the removal branch covers it. I don't know whether Vision pads lines in practice.

Not verified: no live capture, and `Build & selftest` still has not reported on this revision — I am taking your `8799 checks` at face value.

**Impact:** Blocked from merging as it stands. With the option on, Korean seams touching compatibility jamo lose their space. Default off, so nobody is affected until they turn it on.
**Merge:** needs the rebase onto the restructured `ScreenCaptureSettings.swift` first; the jamo split can ride in the same push.

ref #973

ywu73:
Addressed across `1dc8193f`, `297d7cbd`, and `786fa3ad`.

- Merged the current `main` in `1dc8193f` and resolved the `ScreenCaptureSettings` conflict against the restructured layout. The line-break toggle now sits after the OCR caption and before the QR toggle, without restoring the old shortcut row.
- Split the tight-script range into `U+3000...U+312F` and `U+3190...U+9FFF`, excluding Hangul Compatibility Jamo `U+3130...U+318F`. Korean seams such as `ㅋㅋ` / `ㅎㅎ` now keep their space.
- Left the parenthesized and circled Hangul list-marker blocks tight, as suggested.
- Added a Compatibility Jamo regression test that pins the spaced Korean seam.
- Also handled the raw-line whitespace case in `786fa3ad`. When line-break removal is enabled, each recognized line is trimmed before seam classification, preventing OCR padding from adding extra spaces or defeating tight CJK joins.
- The trimming is limited to the enabled path. With line-break removal off, the original line contents and line breaks remain unchanged.
- Added boundary-whitespace tests for spaced scripts, CJK, Korean, and the setting-off path.

Local verification on `786fa3ad`:

- `./build.sh --test` — `TESTS OK (8810 checks)`
- `./build.sh` — release bundle assembled successfully

ref #973

PathGao:
Rebase landed clean — the hunk context matches `main` line for line (`ScreenCaptureSettings.swift:143-161`), the toggle sits after `ocrCaption` and before the QR toggle with no shortcut row in between. The jamo split is right: `0x3000...0x312F` + `0x3190...0x9FFF` excludes `U+3130...U+318F`, and Hangul syllables, jamo `U+1100...U+11FF`, and Extended-A/B were already outside every range. Trimming only fires on the removal path, `joinedRecognizedText` still has one production caller on `main`, and no other file in `Sources/` carries a scalar-range script predicate. Both CI checks are green on this revision.

One thing left and it is prose. The Changes section still lists the ranges as `U+3000...U+9FFF` unsplit, and the bullet after it — "Keep Korean seams spaced by stopping the halfwidth range before Hangul jamo at `U+FFA0`" — is the exact claim the split corrected. Korean is kept spaced in two places now, not one. Checklist box 4 says the description reflects the diff, and the body is what ends up in the squash message.

Known gap, not asking you to fix it here: script-neutral punctuation is not in the tight set, so a Chinese line wrapping right after `——` or `……` still gets the space — `U+2014` / `U+2026` fail the tight check on their side of the seam. You cannot just add them, because `word…` + `next` in English wants that space. It needs a third "neutral, inherit from the other side" class, which is more than this PR should carry. One line above `isTightScriptScalar` would save the next person from re-deriving that.

Not verified: no live capture. I judged wording on the zh/ja/ko captions only; the Latin-script ones I read for shape, not with a native ear.

**Impact:** none beyond what the tests pin. Default off.
**Merge:** ship it.

ref #973

Issue #1061, opened 2026-08-27
Full-screen/cross-Space window switcher shows phantom windows and hides real ones


naveenkrdy:
### Description

The App Switcher and Dock Preview show phantom sub-windows for an app when it's full-screen or has a window on a different Space: a dismissed dialog or Settings panel that was never fully torn down by the app still shows up as a switchable window, indistinguishable from a real one. Selecting the phantom does nothing, or briefly focuses then falls back to the previous app.

The same code path also has the opposite failure: a second real window belonging to the same app, on a different Space (windowed or full-screen), can be silently missing from both the App Switcher and Dock Preview while the app's other window is full-screen and active.

Both are in `WindowEnumerator.swift`'s hidden-Space handling, the single enumerator shared by the App Switcher, Dock Preview, and Command Bar.

### Steps to Reproduce

**Phantom window:**
1. Open Terminal, open its Settings/Preferences panel, then close it.
2. Put the Terminal window into full-screen mode (or drag it to another Space).
3. Open the App Switcher (⌥Tab) or hover the Dock icon for Terminal.
4. A "Profiles" (or similarly named) entry appears alongside the real window — it doesn't correspond to anything actually open.

Reproduced the same way with Zen Browser: a leftover browser-window entry (a previously-closed tab/window) stays listed and doesn't activate anything when selected.

**Missing real window:**
1. Open two Terminal windows. Put one into full-screen mode.
2. Move/full-screen the second Terminal window onto a different Space.
3. Open the App Switcher while the first window is active/full-screen.
4. The second, genuinely open Terminal window does not appear at all.

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.3-beta.3 (built from source, `main`)

### macOS Version

macOS 26 (Tahoe) — 26.6.2, build 25G83

### Language

English (US) · macOS in English (India)

### Mac and displays

MacBook Air (M4) · built-in display only, multiple Spaces

### Installation Method

Built from source

PathGao:
Both halves are the same predicate: `isOnHiddenSpace` in `WindowEnumerator.swift:203`. It's the only thing standing between a real parked window and the ghost veto at line 261.

The escape hatch is line 208:

```swift
guard let visible = visibleSpaces, !visible.isEmpty else { return false }
```

[...]

Issue #339, opened 2026-07-22
[Bug]: Dock Preview and Window Switcher only show windows from the current Space


RuanMD:
## Bug report

The Dock Preview and the Window Switcher only show windows that are open in the current macOS Space/Desktop.

If the same application has other windows open in another Space, those windows do not appear in the preview or in the window switching interface.

This makes it seem as if the application has fewer open windows than it actually does and prevents users from switching directly to a window located in another Space.

### Steps to Reproduce

Vorssaint should detect and display all open windows from the selected application, including windows located in other macOS Spaces/Desktops.

Suggested behavior:

- Show windows from the current Space and other Spaces.
- Indicate which Space each window belongs to.
- When a user selects a window from another Space, macOS should switch to that Space and focus the selected window.
- Optionally add a setting to choose between:
  - Show windows from all Spaces
  - Show only windows from the current Space
- Use “All Spaces” as the default behavior, or preserve the current behavior for existing users and expose the option in Settings.

1. Open multiple windows of the same application.
2. Move one of the windows to another macOS Space/Desktop.
3. Stay in the current Space.
4. Open the Vorssaint Dock Preview or Window Switcher.
5. Observe that only the windows from the current Space are displayed.
6. The window located in the other Space is missing.

### Screenshots / Recordings

The preview shows only the Antigravity IDE windows from the current Space. Other open windows from the same application, located in another macOS Space, are not listed.

<img width="1439" height="834" alt="Image" src="https://github.com/user-attachments/assets/19f82c9e-4144-4bef-9c03-5f8d118b9dd4" />

### Vorssaint Version

Versão 3.1.15

### macOS Version

macOS 27 (Golden Gate)

vorssaint:
3.2.0 shows windows parked on other desktops, and there is an option if you would rather keep everything to the current one.

Issue #512, opened 2026-08-06
Switcher shows windows that are closed


rgndxzzk:
### Description

Mostly with apps that can only have one window open, like Discord or Nicotine+, they're still on the cmd+tab menu even though "Apps with no open window" is set on Do not show.

### Steps to Reproduce

1. Enable switcher
2. Press Cmd+tab
3. Discord will show up on cmd+tab even though there arent any windows open or minimized

### Screenshots / Recordings

https://github.com/user-attachments/assets/1993cc1f-858c-4ffc-b8e3-c09c92f1719f

### Vorssaint Version

3.3.0

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

rgndxzzk:
Also just noticed FL Studio doesnt show up at all on the switcher menu..

PathGao:
Status, split in two because your report contains two different problems.

**Closed windows still listed** — addressed by #826 (merged 2026-08-24). `WindowEnumerator` now drops windows whose app has closed them even when the app answers Accessibility with nothing, which is the hole that let Discord-style single-window apps linger. It is **not in any release yet**: v3.3.2 (2026-08-20) predates it and the 3.3.3 betas were cut on 2026-08-22, before it landed. So there is nothing you can install today that contains it, and this issue keeps waiting rather than carrying a version check label.

**FL Studio never appearing at all** — a different path. Windows hosted through a compatibility layer are enumerated and filtered separately (`WindowEnumerator.swift:138+`), and #826 does not touch that code. Even once the ghost-window half is confirmed, this half stays open.

Note also that genuinely minimized windows are kept on purpose, since they still hold their Space membership; only windows their app has closed are removed.

PathGao:
The ghost half is fixed by e905d3f. "FL Studio never appears at all" is a different path (compatibility layer / bottle, WindowEnumerator.swift:139-149) and should not be closed along with it

rgndxzzk:
FL studio issue fixed. but closed windows still appear, under the icon on the window preview theres an eye with a slash.. should i change some setting?

PR #1062, opened 2026-08-27
fix(switcher): stop stale window-server surfaces from reappearing, and stop hiding real ones

naveenkrdy:
## Summary

Two bugs in `WindowEnumerator`'s hidden-Space handling, both traced with real, on-device diagnostic captures rather than guessed at.

The hidden-Space veto bypass (added for #339) treats any surface still carrying a Space tag as a real window parked elsewhere. A dismissed dialog or Settings panel can keep an old Space tag indefinitely if its backing surface is never fully torn down, so it gets readmitted as a switchable window alongside the real one - a phantom entry in the App Switcher and Dock Preview that does nothing when selected. Reproduced with Terminal (a leftover "Profiles" panel) and Zen Browser (a leftover browser window) independently.

Fixing that exposed a second, opposite bug already present: `frameLooksFullscreen`'s ±2pt height tolerance against `NSScreen.frame` can't recognize a real fullscreen window, because the auto-hiding menu bar shifts a fullscreen window's reported height by its own strip height as it slides in and out (observed varying by up to ~75pt on this hardware). That silently disabled the fullscreen escape hatch the whole bypass exists for, so tightening the veto against ghosts also made a second real fullscreen Terminal window on another Space disappear entirely - not present in either the App Switcher or the Dock Preview while the first window was active.

Both fixes are in the one shared `listWindows(filterPID:...)` function, so this covers all three callers: App Switcher, Dock Preview, and Command Bar.

## Verification

```
./build.sh --test              # TESTS OK, 8803 checks
```

MacBook Air (M4), macOS 26.6.2 (build 25G83), built from source, English (US) with macOS set to English (India). Single built-in display, multiple Spaces - no external/multi-display setup available, so that combination is untested.

Both fixes verified live rather than by reasoning alone: added temporary instrumentation, captured real window-server and Accessibility state across several rounds on this machine, adjusted the fix against that evidence, then removed the instrumentation once confirmed. The final build was installed and used directly to confirm both phantom windows are gone (Terminal's Profiles panel, Zen's leftover window) and the previously-missing real cross-Space Terminal window now appears, in both windowed and fullscreen variants.

## Anything else

Refs #1061


PathGao:
Four things, in the order they'd bite.

**1. `looksFullscreen` at 85% height can't tell fullscreen from zoomed/tiled.**

`SwitcherSupport.looksFullscreen` wants exact width (±2) and height ≥ 85% of the screen. A window zoomed or Fill-tiled to `visibleFrame` matches both: same display width, height = screen − menu bar − Dock. On your own 1710×1112 machine that's roughly 1710×1074 with the Dock hidden and 1710×989 with it showing, against a 945pt floor — and against your captured real-fullscreen heights of 1073 and 1037. Your two fixtures and a plain zoomed window sit within a point or two of each other, so no height threshold separates them. The fixtures show the frame stopped carrying the signal, not that a looser bound recovers it.

What that does to a merely-zoomed window, none of it in the body:

- `showFullscreenWindows == false` drops it from the switcher entirely (`WindowEnumerator.swift:351`).
- Dock Preview hides its preview controls and refuses drag-to-place (`DockPreviewPanelView.swift:478,542,603`, `DockPreviewSupport.canDragToPlace`).
- Activation takes the fullscreen retry path (`WindowActivator.swift:88`), and `AppSwitcher.swift:1248` nils out the source window id when the source is "fullscreen".

macOS 26 drag-to-top tiling makes exact-`visibleFrame` windows ordinary, so this isn't an edge configuration. `axWindow?.isFullscreen` (AXFullScreen / AXFullScreenWindow) already answers correctly whenever AX can see the window; the frame heuristic only has to cover the case where it can't. Gating the loose height on "AX had no opinion" keeps the #1061 recovery without reclassifying every tiled window.

**2. The second fix isn't confined to `listWindows`.**

Body: "Both fixes are in the one shared `listWindows(filterPID:...)` function, so this covers all three callers." True for the veto change, not for `frameLooksFullscreen`. It has two more callers on `main`:

- `WindowEnumerator.swift:517` — the AX snapshot's own `isFullscreen`.
- `WindowEnumerator.swift:656` — `fillsScreen` inside `isUserFacingWindow`, which is a window *admission* gate, not a display flag. Its comment reads "The screen-sized frame keeps ordinary utility windows filtered." At 85% that filter now admits any full-width `AXUnknown`/`AXFloatingWindow` surface taller than 945pt.

Different question, same predicate. The admission gate wants the old tight bound.

**3. `minimumParkedWindowSize` is measured in the body and guessed in the code.**

Body: "traced with real, on-device diagnostic captures rather than guessed at." The comment on the constant: "Starting value pending on-device measurement of a real leftover Settings/dialog surface's size against a real small parked window's." One of those is wrong.

The floor is load-carrying either way: a non-fullscreen surface under 320×220 on a hidden Space is now dropped at both sites, including the #339 path where AX described nothing at all for the owner. Calculator's window is about 230pt wide. Unless the captures you already have pin an actual separation between leftover surfaces and real small windows, `axHasNoWindowsForOwner` is carrying the ghost fix on its own and the floor only adds a silent-drop class.

**4. "AX vouched for one window, so it can reach the app" doesn't hold per window.**

`isUserFacingWindow` sets a 0.35s messaging timeout on each window (`WindowEnumerator.swift:487,498`), and an attribute read that times out returns nil, falling through to `false` at line 668. A slow app drops individual windows out of `axWindows` while its siblings get described — non-empty snapshot, one real window missing. If that window is parked on a hidden Space and isn't fullscreen, the new veto removes it. Not the common case, but it's the same failure shape #339 was about, and the new comment states the premise as if it were closed.

Not verified: no hardware here. All of the above is read off `main` plus the diff; the zoomed/tiled frame numbers are derived from `visibleFrame`, not measured.

**Impact:** Zoomed or tiled windows get reclassified as fullscreen — hidden outright for anyone running with "show fullscreen windows" off, and losing Dock Preview controls plus drag-to-place for everyone else. Separately, small real windows parked on other Spaces get dropped by a size floor the code itself calls unmeasured.
**Merge:** needs the fullscreen heuristic narrowed so a `visibleFrame`-sized window doesn't read as fullscreen first.

ref #1061
ref #512

Issue #324, opened 2026-07-20
[Bug]: Closing all windows makes the Switcher not appear in the next invocation


LetrixZ:
### Description

When closing all windows of an app, the Switcher will not appear in the next invocation, requiring a second invocation to restore it.

Another issue that seems related, is that the Switcher doesn't update its apps when changing spaces, leaving a mix of the new and old windows and, when interacting with a window from the old space, it crashes the Switcher, requiring again a second invocation to restore it.

### Steps to Reproduce

1. Open the Switcher.
2. Open a new window and then close it.
3. Try to open the Switcher again.

### Screenshots / Recordings

https://github.com/user-attachments/assets/23fe5c67-58cb-466e-9a50-87d7777bb622

### Vorssaint Version

3.1.14

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

vorssaint:
@LetrixZ the switcher now opens on the first press even when the front app has no windows left, in 3.1.15. let me know if the second-invocation thing is gone for you

LetrixZ:
Works fine now

PR #1064, opened 2026-08-27
perf(switcher): enumerate windows off main

MaximilianMauroner:
App Switcher input could pause while the main thread waited on WindowServer and Accessibility.

This moves each current Switcher window enumeration to a serial queue. The event tap records a bounded pending session, keeps routing repeated shortcut input while enumeration runs, cancels obsolete work, and publishes the result on main only when its generation is still current. The foreground process is captured with the AppKit snapshot, focused-window AX runs on the queue, and a shared 24-worker AX pool caps work across Switcher, Command Bar, Dock, and preview callers.

Testing:

- Build & selftest passed
- Swift 6.0.3 compatibility passed
- Unit tests passed with 8,821 checks

Made with GPT-5.6 Sol in T3 Code using the Codex harness.


MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -859,6 +890,105 @@ final class AppSwitcher: ObservableObject {
...
+        enumerationQueue.async { [weak self] in
+            guard let self else { return }
+            while true {
+                let groupByApp = UserDefaults.standard.bool(forKey: DefaultsKey.switcherMergeTabs)
+                let preservesGroupedWindows = SwitcherSupport.preservesGroupedWindowsDuringEnumeration(
+                    allApps: true,
+                    mergeWindowsByApp: groupByApp,
+                    simpleMode: self.simpleModeEnabled
+                )
+                let items = WindowEnumerator.listWindows(
```
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P2] **Stop queued Accessibility work during suspension**

When permission-reset uninstall calls AppSwitcher.suspend(), the event tap stops but the new cache remains enabled and its workspace observers remain installed. An already queued refresh also enters WindowEnumerator.listWindows() before it checks windowCacheEnabled. Accessibility enumeration can therefore begin during the permission reset that suspend() is meant to isolate, which breaks that teardown contract.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -735,14 +749,27 @@ final class AppSwitcher: ObservableObject {
...
-                simpleMode: simpleModeEnabled
-            )
+        let groupByApp = allApps && mergeWindowsByApp
+        let preservesGroupedWindows = SwitcherSupport.preservesGroupedWindowsDuringEnumeration(
+            allApps: allApps,
+            mergeWindowsByApp: mergeWindowsByApp,
+            simpleMode: simpleModeEnabled
         )
+        var allWindows = cachedWindows(groupByApp: groupByApp,
+                                       preservingGroupedWindows: preservesGroupedWindows)
```
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P1] **Validate the warmed list before using it**

When a window opens, closes, minimizes, restores, or moves Spaces inside the same app, no observed application lifecycle notification fires. This matching cache entry is still accepted as the complete session list, and the general refresh runs only after that session. The first shortcut can therefore omit a new window or select a closed or moved target. This matters because the cache changes a current enumeration into a stale activation decision.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -859,6 +890,105 @@ final class AppSwitcher: ObservableObject {
...
+                }
+                guard shouldRepeat else { return }
+            }
+        }
+    }
+
+    private func cachedWindows(groupByApp: Bool,
+                               preservingGroupedWindows: Bool) -> [SwitcherItem]? {
+        enumerationLock.withLock {
+            guard cachedWindowGroupByApp == groupByApp,
```
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P2] **Account for every enumeration preference**

The cached list depends on app rules, windowless-app mode, current-Space mode, minimized placement, and fullscreen visibility, but this cache identity checks only grouping. Several of those settings do not call syncWithPreferences(). After a user changes one, the next session can still show windows produced under the old setting, so the visible Switcher state disagrees with the saved preference.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I now stop the window cache from suspend(), increment its generation, clear its state, and remove its observers. Each queued refresh captures the generation and checks that the cache is still enabled before it starts WindowEnumerator. This prevents an old refresh from beginning Accessibility work during permission reset. The new generation checks are covered by the unit suite.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I added every WindowEnumerator input to the cache fingerprint: app rules, windowless-app mode, grouping, backing-window preservation, minimized placement, fullscreen visibility, and current-Space mode. A saved setting change now invalidates the cached identity before the next session uses it. The unit suite checks each input independently.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I replaced the layout-only cache check with a fingerprint of the current windows, applications, and visible Space topology. A window ID, title, bounds, visibility, Space, or app-state change now makes the cached fingerprint differ, so the session performs a current enumeration instead of using the old target list. The added tests cover these fingerprint changes.

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -735,14 +750,36 @@ final class AppSwitcher: ObservableObject {
...
+        let preservesGroupedWindows = SwitcherSupport.preservesGroupedWindowsDuringEnumeration(
+            allApps: allApps,
+            mergeWindowsByApp: mergeWindowsByApp,
+            simpleMode: simpleModeEnabled
+        )
+        let fingerprint = WindowEnumerator.switcherFingerprint(
+            groupByApp: groupByApp,
+            preservingGroupedWindows: preservesGroupedWindows
         )
+        var allWindows = cachedWindows(matching: fingerprint)
```
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P1] **Reject cache reuse when Accessibility adds untracked windows**

WindowEnumerator can add fullscreen and other windows found only through Accessibility, and it also uses AX state to reject stale WindowServer surfaces and derive window state. The cache fingerprint records only WindowServer, workspace, Space, and setting state. An AX-only window can therefore open, close, minimize, restore, or change fullscreen state without changing this fingerprint, so this cache read can omit a live target or activate a stale one.

PathGao:
`finishWindowCacheRefresh` re-arms itself with no delay and no cap. On a fingerprint mismatch it sets the flag it is about to test, so `shouldRepeat` is always true and it calls `scheduleWindowCacheRefresh(generation:)` straight back:

```swift
} else {
    cachedWindowItems = nil
    cachedWindowFingerprint = nil
    enumerationRefreshRequested = true
}
guard enumerationRefreshRequested else { enumerationScheduled = false; return false }
```

The fingerprint carries `title`, `bounds`, `alpha`, `isOnScreen` and the raw `CGWindowListCopyWindowInfo` order for every layer-0 window. A full AX walk is not instant — `accessibilityBatchBudget` is 5s — so any window whose title ticks during it (Terminal running a command, a tab loading, a player showing elapsed time) makes the before/after fingerprints differ every round. That is a continuous AX walk over every regular app on a `.userInitiated` queue, and `startWindowCache()` sits in the feature-enabled branch of the setup block, so it runs whenever the Switcher is on, not only during a session. Bound it: stop after N consecutive mismatches, or drop the volatile fields from the fingerprint.

Second cost, on the path this is meant to speed up. `switcherFingerprint` calls `SpaceWindowBridge.spaces(of:)` once per window, unconditionally, on main. `listWindows` resolves the same thing lazily and memoizes it for the stated reason — "Resolved lazily and cached, so fully Accessibility-confirmed lists pay nothing" (`WindowEnumerator.swift:193`). `CGSCopySpacesForWindows` already takes a CFArray, so this is one call for all IDs rather than one per window. The loop above pays it twice per iteration, both times on main.

Third: MRU history is an input to the cached list and is not in the fingerprint. `WindowUseTracker.recordSwitch` promotes the target immediately because "the window server's order lags an activation by a frame or two, and a flick is faster than that". A second ⌘Tab inside that lag hits a cache the fingerprint still calls fresh while `windowHistory` has already moved, so `orderedForSession` hands back the pre-switch order.

`needsFocusedWindowLookup`: `prefix(2).count > 1` treats 0 visible windows the same as 1. With the front app fully minimized, `sessionSourceItem` now falls through to `candidates.first(where: { $0.windowID == nil })` — the app entry — where the old code could match a focused minimized window and return it, which flips `hasForegroundEntry` and the initial selection index. The body says the lookup is skipped "when the foreground source is unambiguous"; zero visible windows is not that. I have not verified what AX reports as the focused window for a fully-minimized app, and that is what decides whether this actually bites.

Smaller: the session path computes `groupByApp = allApps && mergeWindowsByApp` while the warm path hardcodes `allApps: true`, so with "one row per app" on, every ⌘` session misses and then writes its own shape into the single cache slot. The fallback is declared in the body; the eviction isn't — alternating ⌘Tab and ⌘` never hits.

Also: the enumeration-input key set is now written twice. `listWindows` reads five defaults (`switcherAppRules`, `switcherWindowlessApps`, `switcherCurrentSpaceOnly`, `switcherMinimizedPlacement`, `switcherShowFullscreenWindows`); `switcherFingerprint` re-reads the same five. They match today, I checked. A sixth default added to `listWindows` gets no compile error and produces a silently stale cache. Have `listWindows` return the inputs it used, or read them once into one struct both sides take.

**Impact:** with the Switcher enabled, one app with a live-updating window title puts the enumeration queue into a continuous AX walk plus two main-thread window+Space scans per round. Separately affects ⌘Tab ordering right after a switch, and the initial selection when the front app has no visible window.
**Merge:** needs the refresh loop bounded first.

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I removed cache reuse instead of weakening the AX correctness check. Each shortcut now builds one current window list on the serial enumeration queue, and main publishes it only if the pending generation is still current. This removes the stale AX-only target path while keeping WindowServer and Accessibility work off main.

MaximilianMauroner:
[GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I replaced the warmed cache with one current enumeration per pending session. This removes the repeat-refresh loop, the two fingerprint Space scans, stale MRU cache ordering, and grouping-shape eviction. I also restored the focused-window AX lookup when the foreground app has zero visible windows. The branch is rebased onto current main and reduced to one commit.

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

PathGao:
The cache is gone and with it the five things I flagged. Three new ones.

**The front-PID guard can eat the keypress.** `finishPendingSession` discards the session when `reportedFrontPID` isn't in the snapshot and the retry misses too. The tap already returned `nil` for that keyDown, so a discard means ⌘Tab did nothing — the shape the comment further down calls out:

> Bailing out here handed ⌘Tab back to the system, so the shortcut looked dead until it was pressed a second time (issue #324).

Nothing in `.allApps` needs the guard: with the front PID absent from `regularApps`, `sessionSourceItem` returns nil, `initialSelectionPosition` already handles the sourceless session, and the panel opens. `.frontmostApp` already discards on the empty scope below it. The retry also re-runs the whole walk — another `accessibilityBatchBudget` (5s) — and on a flick (`commitWhenReady`) that is the delay before the switch happens. Open on the second attempt instead of discarding, or drop the guard.

**The enumeration queue's first act is to wait on main.** `snapshot()` does `DispatchQueue.main.sync` whenever it is off main, so the work moved off main cannot start until main is free — the exact condition this PR exists to survive. `beginPendingSession` already runs on main (`AppSwitcher.swift:532`), so take the snapshot there and pass it into the async block. Same at `CommandBarService.swift:2246`: `listWindowsForCommandBar()` is called from `DispatchQueue.global(qos: .userInitiated)` so the AX walk never touches main, and it now hops to main before doing anything. That call site needs the snapshot hoisted too, or every Command Bar open picks up a new main-thread block.

**`snapshot()` can trap where the old code couldn't.**

```swift
return (CGWindowID(window.windowNumber), title)
```

`CGWindowID(Int)` traps on a negative `windowNumber`, and `Dictionary(uniqueKeysWithValues:)` traps if two entries come back as the same key. `ownWindowTitle` used `first(where: { $0.windowNumber == Int(windowID) })`, which could do neither. I can't produce a window that passes `titled && canBecomeKey && (isVisible || isMiniaturized)` with `windowNumber <= 0`, so I'm not claiming it fires — but `guard window.windowNumber > 0` plus `uniquingKeysWith: { first, _ in first }` makes it impossible for the cost of one line.

Not verified: I have no macOS toolchain either, this is diff reading. Both checks are still unreported. The parts I'd watch there are `enumerationQueue.async { [weak self] in ... self.routeLock ... }` on a non-Sendable `ObservableObject` and `[SwitcherItem]` crossing back to main.

**Impact:** ⌘Tab can be swallowed with no panel when the frontmost app is missing from the snapshot; enumeration still has to wait on main to begin; Command Bar gains a main-thread hop it was written to avoid.
**Merge:** needs the discard path decided first.

MaximilianMauroner:
[GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I changed the foreground retry to apply only to the window-scoped shortcut. The all-app shortcut now opens through the existing sourceless-session path instead of discarding the keypress.

I also moved snapshot capture to the existing main-thread App Switcher and Command Bar call sites. Only the full WindowServer and Accessibility walk crosses to the worker queue. Own-window IDs now require an exact positive conversion, and the title map keeps the first value for duplicate IDs.

MaximilianMauroner:
@codex review

PathGao:
The three from last round are gone: the allApps discard, the main hop inside `snapshot()`, and both traps. Two left, and one fix covers both.

**`reportedFrontPID` moved from before the walk to after it.** On `main` it is read at the top of `beginPendingSession`, in the same instant as the enumeration. It is now read in `finishPendingSession`, up to `accessibilityBatchBudget` (5s) later. So the snapshot answers "what was there when the key was pressed" and the front PID answers "what is in front now". If anything takes focus during the walk — an app finishing launch, a dialog — the session is measured against that app instead: `sessionSourceItem` puts its window at index 0 and ⌘Tab toggles straight back to it. The body doesn't mention the move.

Read it on main next to the snapshot and pass it through:

```swift
let enumerationSnapshot = WindowEnumerator.snapshot()
guard let reportedFrontPID = NSWorkspace.shared.frontmostApplication?.processIdentifier else {
    discardPendingSessionStart(generation: generation)
    return
}
```

That deletes `snapshotRunningPIDs`, `retryCount` and the retry branch with them. A PID read at snapshot time is in the snapshot by construction, and the `scoped.isEmpty` discard below already covers "front app has no windows in the list" — which is what `main` does.

**Accessibility is still on main on the common path.** `needsFocusedWindowLookup` is true whenever the foreground app has two or more visible windows, which is most ⌘Tab presses, and `focusedWindowID(for:)` then makes two AX round trips on main out of `finishPendingSession`. Bounded at 0.35s each, but the body opens with "the main thread waited on WindowServer and Accessibility" and this is the last piece of that left at session start. With the PID read at snapshot time, `windows` and the focused-window lookup both fit inside the `enumerationQueue` block and only the publish needs main. Its rationale comment goes stale either way — `AppSwitcher.swift:925` says "The tap thread waits on the session start", and after this PR it doesn't.

Two small ones:

`WindowEnumerator.listWindows(groupByApp:preservingGroupedWindows:)` has no callers left. `AppSwitcher.swift:738` was the only one and this PR points it at `enumerateSwitcherWindows`; nothing in `Tests/` touches `WindowEnumerator`. Unless it is kept deliberately as the main-thread entry point, delete it.

`WindowEnumerator.swift:475` still says "This runs on the main thread (tap callback and activation warm-ups)" above the messaging timeout. The switcher path no longer does. The timeout still earns its place — the queue is serial and the budget is 5s — but not for the reason written there.

Not verified: no toolchain here either, this is the diff plus grep over `main`. I did check the new `dispatchPrecondition(.onQueue(.main))` in `snapshot()` against every path that reaches it — `listWindows(for:)` from DockPreview (4 sites), `WindowPreviewProvider`, `finishClosingWindow`, and `loadWindowsIfNeeded` are all on main, so nothing traps. What I can't check is the non-Sendable `self` capture in `enumerationQueue.async` and `[SwitcherItem]` crossing back; CI says Swift 6 compatibility is green, so it's a runtime question, not a compile one.

**Impact:** an app taking focus during the enumeration walk makes ⌘Tab open on that app's window and toggle back to it instead of the one the user pressed from; session start still spends up to two AX round trips on main.
**Merge:** needs the frontmost-PID read moved back before the walk.

MaximilianMauroner:
[GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I now capture the foreground PID on main beside the AppKit enumeration snapshot, before the WindowServer and Accessibility walk. That keeps the session measured against the app that owned the shortcut press.

The serial enumeration queue now checks the pending generation again after building the window list, resolves the focused source window there, and sends the PID, focused ID, and window list to main for publication. `finishPendingSession` no longer reads the current foreground app or performs AX calls. I also removed the unused list wrapper and updated the timeout comments for the queued path.

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

PathGao:
The PID read is back on main beside the snapshot and the AX lookup moved onto the queue. Two left, neither is a behavior bug.

**The batch still runs on main, and the comments that said why are gone.** `listWindows(for:)` calls `snapshot()` and then does the whole walk synchronously on the calling thread, and every call site is main: `DockPreviewService.swift:576`, `:750`, `:1398`, `:1452`, `WindowPreviewProvider.swift:429` (inside `main.asyncAfter`), `AppSwitcher.swift:1134`. A Dock hover over an app that isn't servicing its run loop still spends up to `accessibilityBatchBudget` on main — issue #189's shape, untouched. Both places that recorded it now point only at the queue:

> Ceiling on the whole batch, which the caller waits out … The serial session queue must remain bounded so a later shortcut can still make progress.

> Keep each operation short so the bounded batch can release its caller and the serial session queue.

The body reads the same way — "moves each current window enumeration to the Switcher serial queue". It moves the switcher's. The Command Bar's was already on `DispatchQueue.global` (`CommandBarService.swift:2245`), and `listWindows(for:)` didn't move at all. Those timeouts and the 5s ceiling are the only thing between a Dock hover and a frozen keyboard, and the next person weighing whether to relax them will read these comments and conclude nothing on main depends on them. Keep the #189 sentence, scoped to `listWindows(for:)`.

**The scope switch runs twice.** The queue builds `sessionWindows`, uses it for `needsFocusedWindowLookup`, drops it, and hands `allWindows` to main, which re-derives the same list from the same inputs. Send the scoped list instead:

```swift
DispatchQueue.main.async { [weak self] in
    self?.finishPendingSession(generation: generation,
                               reportedFrontPID: reportedFrontPID,
                               focusedSourceWindowID: focusedSourceWindowID,
                               windows: sessionWindows)
}
```

`finishPendingSession` then needs one `guard !windows.isEmpty else { discardPendingSessionStart(generation: generation); return }` and loses both the `scope:` parameter and the switch — `sessionScope` and the selection branch below read `pending.scope`, not the argument.

Not verified: no toolchain here, this is the diff plus grep over `main`. I did check every AppKit call left in `WindowEnumerator` on `main` — `NSWorkspace` (121), `Permissions.shared` (426), `NSScreen` (430, 633), `NSApp` (753) — and all five are covered by `Snapshot`, so nothing AppKit-only is left running on the queue. `WindowUseTracker.reconcile` was already reached off main by the Command Bar path and holds `stateLock`, so it isn't newly exposed. What I still can't check is the non-Sendable `self` capture in `enumerationQueue.async` and `[SwitcherItem]` crossing back; CI answers that, and neither check has reported yet.

**Impact:** none on behavior. A Dock hover can still block main for 5s and the comments no longer say so.
**Merge:** needs the #189 main-thread note kept where it still applies.

MaximilianMauroner:
[GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I restored the issue #189 rationale on both AX bounds. The comments now distinguish the queued Switcher session walk from the six synchronous Dock and preview callers that still run on main and depend on those bounds to avoid stalling the event taps.

I also pass the queues already-scoped window list to main. `finishPendingSession` now has one empty-list guard and no second scope switch.

MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
```diff
@@ -29,35 +46,70 @@ enum WindowEnumerator {
     /// tree within the process thread budget while collapsing it into a few
     /// bounded AX batches.
     private static let maximumConcurrentQueries = 24
-    /// Ceiling on the whole batch, which the main thread waits out. Generous
-    /// against the messaging timeouts above even with every query as slow as
-    /// they allow, so only a dispatch pool with no thread to give can outlast
-    /// it (issue #971) — and then no answer was ever coming. A stalled main
-    /// thread stalls the event taps with it (issue #189), so it must expire.
+    /// Ceiling on the whole batch. The Switcher waits on its serial session
```
MaximilianMauroner, line 55: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P1] **Cap Accessibility workers across timed-out batches**

Each enumeration creates a separate OperationQueue with up to 24 workers. When the five-second budget expires, cancellation removes queued work but an operation already walking a process does not observe cancellation. A later Switcher, Command Bar, Dock, or preview enumeration can therefore create another 24-worker pool while the prior pool is still draining, so the intended process-wide worker bound does not hold.

MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I replaced the per-enumeration queues with one shared Accessibility query queue capped at 24 workers across every caller. Each batch retains and cancels only its own operation handles when its budget expires. Running operations now check cancellation before starting and between window, attribute, and snapshot reads, so they stop after the current bounded AX call instead of continuing the full walk.

PathGao:
Both from last round are in. The #189 rationale is back on `accessibilityBatchBudget` and on the per-app messaging timeout, scoped to the six synchronous `listWindows(for:)` callers, and `finishPendingSession` takes the queue's already-scoped list with a single empty guard.

I reread the rest. Nothing left to block on. Two changes in the diff aren't in the body and should be:

**The AX worker pool went from per-batch to process-wide.** `accessibilityQueryQueue` is one shared static capped at 24 across Dock, preview, Command Bar and Switcher batches; before, each call built its own queue at `min(24, orderedPIDs.count)`. This PR is also what makes three batches concurrent — the Switcher's used to run on main, so it couldn't overlap a Dock hover. Sharing the cap is the right call when #971 was a "no thread to give" failure, but it changes a resource all six callers share and the body scopes the PR to the Switcher queue.

**Cancellation now interrupts running queries, not just queued ones.** `isCancelled` threads down through `isUserFacingWindow`, `accessibilityFrame` and `isFullscreenWindow`, so a PID query caught by the budget stops after its current bounded AX call instead of finishing the walk. That is what actually shortens a blown budget — `cancelAllOperations()` only ever dropped what hadn't started.

One dead counter, no behavior: an operation cancelled before it starts returns at `guard let operation, !operation.isCancelled` above the `defer`, so it never decrements `pendingQueries`. Only this batch cancels these handles and only after the deadline has already released the wait, so nothing reads the counter by then. Leave it.

Not verified: no toolchain here, this is the diff plus grep over `main`. Swift 6.0.3 compatibility and Build & selftest are both green, which answers the non-Sendable `self` capture and `[SwitcherItem]` crossing back that I flagged as unchecked the last three rounds. What I still can't check is real hardware — three concurrent AX batches sharing 24 slots is the one thing here I'd want a hung-app repro for, and I can't produce one. `needsFocusedWindowLookup` I did check against `sessionSourceItem` on `main`: with exactly one visible foreground window the nil-focusedID fallback (`candidates.first(where: { $0.isOnScreen && !$0.isMinimized })`) resolves to that same window, and under `groupByApp` the candidate set is the single group item either way.

**Impact:** none outstanding.
**Merge:** ship it.

MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -735,32 +738,77 @@ final class AppSwitcher: ObservableObject {
...
+        enumerationQueue.async { [weak self] in
+            guard let self,
+                  self.routeLock.withLock({
+                      SwitcherSupport.isCurrentSessionStart(
+                          generation: generation,
+                          pendingGeneration: self.routePendingSessionStart?.generation
+                      )
+                  })
+            else { return }
+            let allWindows = WindowEnumerator.enumerateSwitcherWindows(
```
MaximilianMauroner, line 772: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P1] **Space topology reads AppKit from a worker queue**

When this queued call handles an ordinary raw window, its hidden-Space checks reach `SpaceWindowBridge.topology()`. That method reads `NSScreen.screens` and each screen’s `deviceDescription` at `SpaceWindowBridge.swift:105-115`. Command Bar reaches the same enumerator from a global worker queue. These AppKit reads therefore run off main during normal Switcher and Command Bar enumeration, which violates AppKit thread confinement and can make session startup unsafe.

MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I moved the screen UUID-to-display-ID map into `WindowEnumerator.Snapshot`, where `snapshot()` captures it on main. The Switcher and Command Bar worker paths now pass that value to `SpaceWindowBridge.topology(displayIDsByUUID:)`, so those paths no longer read `NSScreen` or `deviceDescription`. The hidden-Space and fullscreen-Space checks still use the same CGS topology data.

MaximilianMauroner:
@codex review

MaximilianMauroner:
@codex review

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -744,32 +747,77 @@ final class AppSwitcher: ObservableObject {
...
```
MaximilianMauroner, line 772: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

[P1] **Obsolete sessions keep the enumeration queue busy**

The generation checks reject stale work only before and after this full window walk. If the event tap cancels or replaces `routePendingSessionStart` during Accessibility enumeration, this call continues for up to the five-second batch budget even though it can no longer publish. The obsolete task blocks the serial Switcher queue, so the current shortcut waits behind it. Its operations also occupy the shared 24-worker AX queue used by synchronous Dock Preview callers, which can delay those main-thread window lists.

MaximilianMauroner: [GPT-5.6-SOL] RESPONDING ON BEHALF OF MAX
-----

I passed the pending-generation check into the Switcher enumeration. The AX batch now checks that state every 10 ms while it waits, cancels only its own retained operations when the generation becomes stale, and returns so the serial Switcher queue can accept the current session. The default path remains non-cancellable for Command Bar and synchronous `listWindows(for:)` callers, and the shared 24-worker cap and five-second timeout remain unchanged.

MaximilianMauroner:
@codex review

Issue #1068, opened 2026-08-27
Configurable Margin in Window Layout


marcelharinck:
### Problem or Motivation

Snapped windows sit flush against each other and against the screen edge. On a
large display that makes it hard to see where one window ends and the next
begins, and it leaves no room for the window shadow. There is also no way to keep
a strip of wallpaper visible, or to avoid clipped corners on displays with
rounded corners.

The gap is fixed, so the only workaround is resizing by hand after every snap.

### Proposed Solution

Two configurable values in Window Layout settings:

- Window Gap: spacing between adjacent snapped windows
- Screen Gap: spacing between a window and the screen edge

Raycast uses a preset dropdown per value rather than a numeric field, which keeps
it simple: None, Tiny (8px), Small (16px), Medium (32px), Large (64px), Extra
Large (128px).

Two details for it to look right: the window gap should be the total distance
between windows, not applied twice, and the screen gap should be measured from
the visible frame so the menu bar and Dock stay excluded. Defaulting both to 0
keeps current behaviour unchanged.

### Feature Area

Windows and Dock

### Does another app already do this?

Raycast - Window Management

### Mockups / References

<img width="733" height="571" alt="Image" src="https://github.com/user-attachments/assets/5e03c865-4e11-4536-b3bf-535f2e2896ab" />

Raycast implementation

PR #1070, opened 2026-08-27
feat(window-layout): add configurable window and screen gaps

marcelharinck:
Refs #1068

Adds the two gap settings from the issue: a window gap between adjacent snapped windows and a screen gap between windows and the screen edge. Both are preset dropdowns (None / Tiny 8 / Small 16 / Medium 32 / Large 64 / Extra large 128 px, the same ladder Raycast uses) and both default to None, so nothing changes until you pick one.

All placement already goes through `WindowLayoutGeometry.rect()`, so the gaps are applied there and shortcuts, the panel grid and drag snapping pick them up from one place. The screen gap insets the visible frame before placement, which keeps the menu bar and Dock out of it. The window gap takes half the gap off each edge a placement shares with a neighbour, so two adjacent windows end up exactly one gap apart instead of two. Center, margin maximize, restore, the display moves and full screen are left alone. `anchoredRect` now anchors to the target rect rather than the visible frame; at gap 0 that is the same thing, since the target is derived from that frame.

One decision worth flagging: maximize respects the screen gap. Raycast does the same, and margin maximize is still there as the percentage version. If maximize should stay flush instead, it is a one-line change in `windowGapped`.

Settings uses menu pickers rather than a segmented control because of the Settings-width story in the other languages. The values are read per placement, so changing them takes effect immediately.

Build without warnings, selftest and `--test` all pass. There are new geometry checks in MetricsTests for the edge shaving, the adjacency distance, the frame insetting, the exempt actions and the small-frame floor, and I checked that they go red when the math is deliberately broken. I also ran it for real on a 3440×1440 display: two windows snapped left/right with gaps 64/16 measure exactly 64 px apart and 16 px from the screen edges, with the top gap starting below the menu bar.

M3 Pro MacBook Pro, macOS 26.6.2 (25G83), MacOSX26.sdk, local `--dev` build with the stable signing identity, app in English. Not tested: multiple displays, windows that refuse to resize, and the sixths/corners in real use (unit tests cover those paths). The strings exist in all thirteen languages but I have only seen English running.


PathGao:
Two things, one contradicts the body.

**Margin maximize eats the screen gap too, and the body says it doesn't.** `screenGapFrame` runs before `ungappedRect` for every action; the exempt switch lives only in `windowGapped`. So `marginMaximize` insets 5% of the *already inset* frame.

visibleFrame 1440×860 @ (0,40), screenGap 32:

- today: `(72, 83, 1296, 774)` — 72px margin
- with the gap: frame insets to `(32, 72, 1376, 796)`, then 5% → ~`(100, 111, 1238, 716)` — 101px margin

Body: "maximize respects the screen gap ... margin maximize is still there as the percentage version". It isn't — once a screen gap is set, the plain 5% margin is unreachable and the two margins compound. `.center` has the same shape: it clamps to `min(current.width, visibleFrame.width)`, so with a screen gap a wide window gets shrunk, and center is also listed as left alone.

The tests miss it because `marginMaximize` and `center` are pinned against `windowGap` only, never `screenGap`. `maximize` is the one exempt action that got both.

Laziest fix — give the screen gap the same exempt set `windowGapped` has, minus `.maximize`:

```swift
let frame: CGRect
switch action {
case .marginMaximize, .center, .restore, .previousDisplay, .nextDisplay, .fullScreen:
    frame = visibleFrame
default:
    frame = screenGapFrame(visibleFrame, screenGap: screenGap)
}
```

**The preset ladder is written twice.** `WindowLayoutGaps.presets` holds the values, `gapPresetTitle` holds a separate value→name switch whose `default` returns `gapNone`. Add 24 to the array and the picker labels it "None (24 px)". One-line fix:

```swift
case 0: return text.gapNone
default: return "\(value) px"
```

Two smaller notes:

- `screenGapFrame` floors at exactly 80pt and the comment justifies it with `accepts()`, but that guard is `actualRect.width > 80` — strict, so a frame at exactly 80 is rejected. Only reachable on a visible frame under ~340pt, so it's cosmetic, but the new test pins the off-by-one.
- I grepped `WindowLayoutGeometry.rect(` on main: three non-test call sites. Two got the gaps; the third is the maximize fallback in `continueSettle` (`WindowLayoutService.swift:478`), which should stay un-gapped — it's a scratch frame to coax stubborn apps into resizing, not a placement. Worth a comment so nobody "fixes" it later. Display moves go through `rectForDisplay`, untouched, so the body is right about those.

Not verified, no hardware: `anchoredRect` still clamps against the un-gapped `visibleFrame`, so an app that refuses to shrink into its gapped cell should end up flush against the screen edge with the screen gap lost. You listed refuse-to-resize as untested and I couldn't check it either.

**Impact:** anyone who sets a screen gap gets doubled margins on Maximize with Margin and no way back to the plain 5%. Nothing changes at the default of None.
**Merge:** needs the screen-gap exemption for margin maximize and center first.

ref #1068

marcelharinck:
Both fixed as you suggested, thanks — the exempt switch you sketched is in `rect()` verbatim, so the screen gap now only reaches the tiling actions and maximize, and margin maximize keeps its plain 5%. Both are pinned in MetricsTests against the screen gap now (the equality your counterexample broke is the assertion). The picker `default` labels an unknown value with its pixel count.

On the smaller notes: I dropped the `accepts()` justification from the 80pt comment rather than chase the off-by-one — the floor is doing its real job (not inverting the frame) and the comment now says only that. The maximize fallback in `continueSettle` has a comment marking it as an ungapped scratch frame.

The `anchoredRect` clamp is as you describe: a window that refuses to shrink ends up flush against the screen edge rather than at the gap. I left that deliberately — the window has to overflow its cell somewhere, and clamping to the full frame keeps it on-screen, where clamping to the gapped frame would push the overflow onto the neighbouring window instead. Happy to change it if you'd rather keep the gap and accept the overlap. Still untested on real refuse-to-resize windows here too.

PathGao:
Both fixes read correct in the diff: the exempt switch, the picker `default`, and the two comments. Screen gap now reaches only the tiling actions and maximize.

Keep `anchoredRect` clamping to the full `visibleFrame`. Clamping to the gapped frame wouldn't shrink the overflow, only shift it — the window keeps its size either way, so you'd hold the gap on one side and push the same excess further over the neighbour on the other. Losing the gap is the cheaper of the two.

One consequence of that choice, from `accepts()` rather than the diff: the anchor check runs against the gapped target with `anchorTolerance = 36` (`WindowLayoutService.swift:75`). The clamp only fires once the window is wider than `visibleFrame.width - screenGap`, and the anchor miss equals that excess, capped at the gap. So at 8/16/32 a clamped window still passes; at 64 and 128 a window within 36pt of full screen width fails `accepts`, takes the settle retries and then the maximize scratch frame — a flash and some latency, then it lands in the same place. Reachable only with a window that refuses to shrink below nearly full screen width, so it isn't worth code, but that's where the untested case will surface if someone reports it.

I read the geometry; I didn't build or run the tests, and there are no checks reported on the PR.

**Impact:** none at the default of None. At screen gap 64/128, a window that refuses to shrink below full screen width takes the retry path before settling flush against the edge.
**Merge:** ship it.

ref #1068

Issue #403, opened 2026-08-01
Mouse tuning: pointer speed / acceleration and scroll speed / acceleration, including a horizontal axis


hash00:
### Problem or Motivation

Vorssaint already owns mouse wheel events: it inverts the wheel without touching the trackpad, and it smooths wheel steps. What it does not expose is **magnitude** — how fast the pointer moves, how far a wheel step scrolls, and how either responds to speed.

macOS gives very little here, and what it does give is the wrong shape:

- **Pointer speed** — one coarse Tracking speed slider, system-wide.
- **Pointer acceleration** — can only be disabled with `defaults write -g com.apple.mouse.scaling -1`, which is global and so wrecks the trackpad too.
- **Vertical scroll speed** — one coarse Scrolling speed slider, also affects the trackpad.
- **Horizontal scroll speed** — nothing at all.
- **Scroll acceleration** — nothing at all.

The recurring issue is that every system control is global. On a laptop a mouse and a trackpad want opposite settings, which is exactly the problem `Invert mouse scrolling` already solves, and solves well — the description under that toggle says it outright: the trackpad is untouched. This request is the same principle applied to magnitude rather than direction.

### Proposed Solution

New controls in the existing **Mouse & Trackpad** tab.

**Pointer** — a new section, as the tab has no pointer concept today:

- **Tracking speed** — mouse only, trackpad unaffected.
- **Disable pointer acceleration** — a toggle. The single most valuable item here for me.

**Scrolling** — extending the existing section:

- **Scroll speed** — a multiplier on wheel step distance.
- **Scroll acceleration** — flat through to adaptive, based on wheel velocity.
- **Separate horizontal axis** — independent speed and acceleration. macOS offers no horizontal control whatsoever.

Why this should be a reasonable fit:

- The hard part, intercepting mouse wheel events per-device without affecting the trackpad, is already built and shipping.
- It is additive: new controls in an existing section, no change to current behaviour, consistent with every feature being optional.
- `Apps to leave alone` already covers the whole Mouse & Trackpad section, so these would inherit the existing escape hatch for 3D and design tools.

Alternatives:

- **The macOS Tracking speed and Scrolling speed sliders.** Too coarse, and both are system-wide — changing them for the mouse changes them for the trackpad.
- **`defaults write -g com.apple.mouse.scaling -1`.** Disables pointer acceleration globally, so the trackpad becomes unusable. Not viable on a laptop.
[...]

PR #1071, opened 2026-08-27
feat(mouse): add standalone mouse acceleration control

CrowKiller:
## Summary

Adds **Disable mouse acceleration** as a standalone Mouse & Keyboard feature. It belongs in Vorssaint because macOS exposes the per-device pointer setting through its own HID event-system APIs, without a helper, driver, Accessibility permission, or dependency on a particular mouse vendor.

When enabled, it uses Sonoma’s `HIDUseLinearScalingMouseAcceleration` property when available; otherwise it applies the established `-1` acceleration fallback. The service keeps the prior value for each mouse and restores it when the setting is disabled or Vorssaint exits. It refreshes the HID service list while running so a mouse connected after launch is covered too.

The feature is independent of Smooth scrolling and mouse button shortcuts. It is inactive when uninstalled in Feature Hub; while enabled, it retains a lightweight two-second refresh timer. The service deliberately accepts only HID devices with the `Mouse` usage and explicitly excludes the trackpad acceleration type, so built-in and external trackpads are unaffected. The PR adds a small system-library import for the macOS HID event-system headers, one service, Settings navigation, and localized user-facing copy for every supported language.

## Verification

- `swift build -c debug` completed successfully on a MacBookAir10,1 with macOS 15.6.1 (Apple Silicon).
- The test app built from that binary was ad-hoc signed and passed `codesign --verify --deep --strict`.
- New localization fields compiled for every supported language.
- Manual check: an external mouse used linear movement as expected; trackpad behavior was not observed to change.

Not run in this environment:

- `./build.sh` and `./build/Vorssaint --selftest`
- `./build.sh --test`
- a dedicated manual check of the trackpad’s acceleration before and after enabling the setting

## Anything else

No existing issue is referenced. This change intentionally does not alter Smooth scrolling or mouse button shortcuts.


PathGao:
Two things, one blocks.

**`syncWithPreferences` doesn't check `isAvailable`.** Every sibling mouse service ANDs it in — `SmoothScrollService.swift:59`, `MouseNavigationService.swift:44`, `ScrollInverter.swift:42`, `MiddleClickService.swift:77`. This one reads only the defaults key.

`FeatureRuntime.setAvailability` writes the availability key and *then* runs the binding (`FeatureRuntime.swift:122-124`), so uninstalling the feature from the hub while the toggle is on lands in `syncWithPreferences` → `start()`. The service re-arms, the 2s timer keeps re-applying, and the Settings section is behind `isAvailable` (`SettingsView.swift:767`) so it's no longer reachable to switch off. Quitting the app is the only way out. Body says "It is inactive when uninstalled in Feature Hub" — it isn't.

```swift
let wanted = AppFeature.mouseAcceleration.isAvailable
    && defaults.bool(forKey: DefaultsKey.mouseAccelerationDisabled)
if wanted { start() } else { stop() }
```

**`applyLinearMode` writes when it has nothing to restore.** When `currentSetting(for:)` returns nil the map stays empty, but the `else` branch still writes `-1`:

```swift
if originalValues[id] == nil, let setting = currentSetting(for: service) {
    originalValues[id] = setting
}
if originalValues[id]?.key == "HIDUseLinearScalingMouseAcceleration" {
```

`restoreOriginalValues` then has no entry for that device, so its acceleration stays at `-1` through toggle-off and through quit. Don't write what you can't undo:

```swift
guard let setting = originalValues[id] ?? currentSetting(for: service) else { continue }
originalValues[id] = setting
```

`SelfUninstall.suspendInputInterceptors` doesn't list the service, but that path ends in `NSApp.terminate` and `applicationWillTerminate` already stops it, so that one is fine.

Not verified: I didn't run `./build.sh` either, and `IOKit/hidsystem/IOHIDEventSystemClient.h` isn't in the public macOS SDK — whether it resolves under the pinned SDK the release build uses is open, and `swift build` on your SDK doesn't settle it. No CI results on this PR yet.

**Impact:** uninstalling the feature from Feature Hub with the toggle on leaves pointer acceleration off with no UI to undo it until the app quits. The second one is narrower — only a mouse whose acceleration property doesn't read back.
**Merge:** needs the `isAvailable` guard in `syncWithPreferences` first.

ref #403

CrowKiller:
Addressed in 1de5a32.

- `syncWithPreferences()` now requires both the feature to be available and its setting enabled, so uninstalling stops the timer and restores the saved device settings.
- `applyLinearMode()` now skips any HID service whose current acceleration setting cannot be read, ensuring every write has a restorable original value.

Verified with `./build.sh --test` (8798 checks). The local development build also compiles against the pinned SDK with the HID header included by `build.sh`.

PathGao:
Both blockers are closed — `syncWithPreferences` ANDs `isAvailable`, and `applyLinearMode` skips any service it can't read back. I also re-checked the uninstall path: `AppFeature.availabilityDefaults` is registered (`Defaults.swift:1251`), so `removePersistentDomain` in `SelfUninstall.removePreferences` leaves `isAvailable` reading `true` and `applicationWillTerminate` still restores. That one's fine.

One left.

**The energy badge claims an event tap; the service is a timer.** `FeaturePresets.swift:92` puts `.mouseAcceleration` in the `.mouse` group. That case is documented as "A mouse event tap (scrolls, clicks or pointer moves)" (`FeaturePresets.swift:77`) and renders in the hub as "Listens to the mouse". This service installs no tap — your own body says it "retains a lightweight two-second refresh timer", which is `.periodic`: "Samples or polls on an interval while active or visible". The badges are curated deliberately (`MetricsTests.swift:11839` — "energy badges tell the honest mechanism per feature"), so this row currently overstates its cost to anyone reading the hub for battery reasons. Move the case to the `.periodic` list with `clipboardHistory` and the monitors.

Smaller, same file you're already in — `applyLinearMode` recomputes the write key instead of using the one it just stored:

```swift
IOHIDServiceClientSetProperty(service, original.key as CFString,
                              (original.key == "HIDUseLinearScalingMouseAcceleration" ? 1 : -1) as CFNumber)
```

`accelerationKey(for:)` is exactly what `currentSetting` used to build `original.key`, so calling it again is a second copy of the same fact, and the only way the two can differ is by writing one property and restoring another.

Not verified: your `./build.sh --test` run and the pinned-SDK compile — I can't run either, and CI still reports no results on this PR.

**Impact:** the hub tells the wrong energy story for this row. No behavioral impact; the two paths from the last round are closed.
**Merge:** needs the `.periodic` energy profile first.

ref #403

CrowKiller:
Addressed in c02e2d6.

- The Feature Hub now reports `Disable mouse acceleration` as **Periodic**, matching its two-second device refresh timer.
- `applyLinearMode()` now writes via the saved `original.key`; it no longer recomputes the acceleration property between saving and applying.

Verified with `./build.sh --test` (8798 checks) and a development build against the pinned SDK.

PathGao:
`original.key` write and the `.periodic` badge are both in. I also re-walked every availability-flip path — `FeatureRuntime.setAvailable:76`, `replaceAvailable:95`, `setAllAvailable:119`, and `FeaturePresets.prepareFirstRunAvailability:19` (which runs before any binding exists) — all four run the binding, so the round-1 uninstall hole is closed on all of them, not just the hub row.

One left, and it's a hazard class this repo already has a written rule for.

**A missed restore becomes permanent, because the next launch re-captures the value this feature wrote.** `originalValues` lives only in memory. If the app dies without `applicationWillTerminate` — crash, Force Quit, `killall` — the device keeps `-1` (or linear `1`). That value survives the process: the property sits on the HID service, not here, which is the same reason the feature works at all. Next launch the toggle is still on, `originalValues` is empty, so `applyLinearMode` falls through to `currentSetting(for:)`, reads back what the last session wrote, and stores *that* as the original. Turning the setting off now restores `-1`. Quitting doesn't help either — `restoreOriginalValues` writes the same `-1` back. There is no move left inside Vorssaint.

`BrightnessService` names this exact case and solves it (`BrightnessService.swift:600-604`):

> If the app went away without putting it back, whether by a crash or by being forced to quit, the only way left was to unplug the screen. The intention is written down instead, and honoured on the next start.

`rememberDisplaySwitchedOff` / `restoreDisplaysLeftOff` (`:605`, `:626`) and `DefaultsKey.fanControlRecoveryNeeded` (`FanControlService.swift:209-212`) are the two existing instances. Three touch points here: persist `originalValues` when a capture is added, load it in `start()` before the first `applyLinearMode`, clear it in `restoreOriginalValues`. Nothing else moves.

Smaller, same map — `registryID(of:)` falls back to `0`, so two mice that don't report a registry ID collide on one key. The second gets the first's `original.key` written to it and the first's value restored. That's a sentinel used as an identity; skipping a service with no registry ID costs less than mis-restoring one.

Not verified: your `./build.sh --test` run and the pinned-SDK compile — I can't run either. CI still reports no results at all on this PR, so no workflow has checked it yet.

**Impact:** one abnormal exit with the toggle on leaves pointer acceleration off for the rest of the login session with no way back through Vorssaint, since switching the setting off writes the same `-1` again. The caption ships that promise in all 13 languages.
**Merge:** needs the captured originals persisted across launches first.

ref #403

Issue #1069, opened 2026-08-27
Allow the radial menu to trigger on mouse-button drag while preserving the button's normal click


KyrieHuang666:
### Problem or Motivation

I want to assign a mouse side button (Back or Forward) to the Radial Menu, but I also need a normal click on that button to keep navigating backward or forward.

At the moment, assigning the button to the Radial Menu makes the menu claim the press, so the button can no longer serve both purposes. This is especially limiting on mice with only two side buttons.

### Proposed Solution

Please add an optional **drag-to-activate** mode for Radial Menu mouse triggers:

- Press and release without exceeding the movement threshold: pass one complete ordinary click back to the foreground app, preserving Browser/Finder Back or Forward.
- Hold the same button and drag beyond the threshold: claim the gesture and open the Radial Menu around the pointer.
- Continue dragging to highlight an item, then release to execute it.
- A configurable movement threshold would be useful, or the feature could reuse the existing mouse-drag threshold.
- Existing per-app mouse exceptions should continue to take priority.

The open implementation in #1024 already uses the desired custody rule for a different action: it holds the press until the pointer moves far enough, replays an ordinary click below the threshold, and suppresses the click after a drag fires. It may be possible to reuse that arbitration for the Radial Menu while keeping normal Back/Forward clicks.

Example: **Button 4 click = Back; Button 4 + drag = Radial Menu.**

### Feature Area

Mouse and keyboard

### Does another app already do this?

Mac Mouse Fix demonstrates the same click-versus-drag distinction: a normal click keeps its original action, while holding and dragging runs a gesture. The requested destination here is Vorssaint's Radial Menu.

### Mockups / References

The attached screenshot shows the current Radial Menu settings in Vorssaint 3.3.3-beta.3. A new “Drag to activate” mouse-trigger option could be placed near “Opening method” or under the per-profile “Mouse button” setting.

<img width="1233" height="768" alt="Image" src="https://github.com/user-attachments/assets/10373964-d888-4cde-9de7-92da7e9a9988" />

Related implementation references: #1012 and #1024.

Issue #490, opened 2026-08-05
增加鼠标中键按住拖移功能


gyc-12:
### Problem or Motivation

<img width="646" height="626" alt="Image" src="https://github.com/user-attachments/assets/9950e4d3-0ded-4a96-b980-6ed5545b3ad1" /> 
可以参考mac mouse fix

### Proposed Solution

<img width="646" height="626" alt="Image" src="https://github.com/user-attachments/assets/9950e4d3-0ded-4a96-b980-6ed5545b3ad1" /> 

### Feature Area

Mouse and keyboard

PR #1073, opened 2026-08-27
feat(radial): add drag-to-activate mode for mouse triggers

Debrupbanik:
## Summary

Adds an optional **Drag to activate** mode for Radial Menu mouse triggers.

- When enabled, pressing and releasing a configured mouse side button (e.g. Back or Forward) below an 8pt movement threshold delivers an ordinary click to the foreground app, preserving browser and Finder back/forward navigation.
- Holding the button and dragging beyond the threshold claims the gesture, opening the Radial Menu centered around the gesture origin in hold-to-activate mode, where releasing on a slice triggers that action.
- When disabled, preserves the existing immediate click-capture behavior.
- Supported and localized across all 13 languages with full custody unit tests.

## Verification

- **Environment:** macOS (Apple Silicon), Xcode Command Line Tools.
- **Automated Tests:**
  ```sh
  ./build.sh --test
  # TESTS OK (8821 checks)
  ```
- **Build & Packaging:**
  ```sh
  ./build.sh --dev
  # Bundle ready: build/stage/Vorssaint (Developer).app
  ```
- Verified state transitions, custody arbitration, and backward-compatible profile encoding/decoding.

## Anything else

Refs #1069


PathGao:
Three things, one of them blocks.

**The `up` half is no longer swallowed when the toggle is off.** The old code ended with a single `return nil` for the matched button, and said why:

> The claimed button is the wheel's alone: both halves of every click are consumed, so the app under the pointer never sees half a gesture (the Settings caption promises exactly that).

The new `.otherMouseUp` branch only returns `nil` when `holdPhase && holdButton == button`. A non-drag session opens with `hold: false`, so `holdPhase` is false and the up falls through to `return Unmanaged.passUnretained(event)`. Every existing profile decodes `mouseDragToActivate` as `false`, so this is the default path: click the back side button, the wheel opens, and the app under the pointer gets an unpaired `otherMouseUp`. The body says "When disabled, preserves the existing immediate click-capture behavior" — it doesn't.

The same fall-through hits the new path when `activationMode == .press`: `startsHeld` returns false for `.press` regardless of `hasHeldButton`, so `holdPhase` is false, and a completed drag leaves a sticky wheel *and* leaks the up. The body's "opening the Radial Menu ... in hold-to-activate mode" only holds for `.pressOrHold` and `.hold`.

**`RadialMouseGestureSupport.decide` is never called.** `handleMouseTap` uses `exceedsThreshold` and hand-rolls the rest. All 20 new assertions test a table that governs nothing, and the table already disagrees with the code it describes: `decide(.idle, .buttonDown(isDragToActivate: false))` returns `.promote` unconditionally, while the implementation ends an already-open session instead; `.tapDisabled` splits on `buttonStillDown` while `handleMouseTap` flushes either way. Body says "Verified state transitions, custody arbitration ... with full custody unit tests". Route the tap through `decide` or drop the table — two copies of the custody rules is exactly what the tests were supposed to prevent.

**The `.navigation` exception gate is an unannounced behavior change, and it borrows another feature's list.** It sits above the profile match, so it applies to every radial mouse trigger, drag mode or not, and the body doesn't mention it. MetricsTests already asserts `MouseExceptionScope.navigation.feature == .mouseNavigation` — that list lives under the Mouse Navigation switch in Settings and hides with it. Turn Mouse Navigation off and an invisible list silently suppresses your radial menu button, with nothing in the UI to explain it. `.buttonShortcuts` is the closer neighbour; a `radialMenu` scope of its own is the honest one.

Smaller: `flushPendingDown` is `WindowLayoutService.flushPending` (`WindowLayoutService.swift:1661`) line for line, minus the comment explaining why `tapPostEvent(proxy)` is used — same press-became-a-drag custody, same field, same clock. `dragThreshold: CGFloat = 8` is a second 8 sitting next to `RadialMenuLayout.moveActivationDistance = 8`, both measured from the same origin. And the marker is a fifth value written into `.eventSourceUserData`; the other four writers all use `0x564F5253` and nothing owns the set (cf. #937).

Not verified: I read the diff and `main`, no hardware. Whether an unpaired `otherMouseUp` actually navigates in Safari or Finder, whether the replayed down reaches `MouseNavigationService`'s tap given both taps are `.cgSessionEventTap` at `.headInsertEventTap`, and whether swallowing `otherMouseDragged` while custody is pending still leaves the cursor free to cross the 8pt threshold — all three need a device.

**Impact:** with `Drag to activate` off (every existing profile), the app under the pointer receives a lone `otherMouseUp` on every radial trigger click; under `activationMode == .press` the new drag path leaks it too.
**Merge:** needs the button-up swallowed again when the session is not in `holdPhase` first.

ref #1069

Debrupbanik:
> Three things, one of them blocks.
> 
> **The `up` half is no longer swallowed when the toggle is off.** The old code ended with a single `return nil` for the matched button, and said why:
> 
> > The claimed button is the wheel's alone: both halves of every click are consumed, so the app under the pointer never sees half a gesture (the Settings caption promises exactly that).
> 
> The new `.otherMouseUp` branch only returns `nil` when `holdPhase && holdButton == button`. A non-drag session opens with `hold: false`, so `holdPhase` is false and the up falls through to `return Unmanaged.passUnretained(event)`. Every existing profile decodes `mouseDragToActivate` as `false`, so this is the default path: click the back side button, the wheel opens, and the app under the pointer gets an unpaired `otherMouseUp`. The body says "When disabled, preserves the existing immediate click-capture behavior" — it doesn't.
> 
> The same fall-through hits the new path when `activationMode == .press`: `startsHeld` returns false for `.press` regardless of `hasHeldButton`, so `holdPhase` is false, and a completed drag leaves a sticky wheel _and_ leaks the up. The body's "opening the Radial Menu ... in hold-to-activate mode" only holds for `.pressOrHold` and `.hold`.
> 
> **`RadialMouseGestureSupport.decide` is never called.** `handleMouseTap` uses `exceedsThreshold` and hand-rolls the rest. All 20 new assertions test a table that governs nothing, and the table already disagrees with the code it describes: `decide(.idle, .buttonDown(isDragToActivate: false))` returns `.promote` unconditionally, while the implementation ends an already-open session instead; `.tapDisabled` splits on `buttonStillDown` while `handleMouseTap` flushes either way. Body says "Verified state transitions, custody arbitration ... with full custody unit tests". Route the tap through `decide` or drop the table — two copies of the custody rules is exactly what the tests were supposed to prevent.
> 
> **The `.navigation` exception gate is an unannounced behavior change, and it borrows another feature's list.** It sits above the profile match, so it applies to every radial mouse trigger, drag mode or not, and the body doesn't mention it. MetricsTests already asserts `MouseExceptionScope.navigation.feature == .mouseNavigation` — that list lives under the Mouse Navigation switch in Settings and hides with it. Turn Mouse Navigation off and an invisible list silently suppresses your radial menu button, with nothing in the UI to explain it. `.buttonShortcuts` is the closer neighbour; a `radialMenu` scope of its own is the honest one.
> 
> Smaller: `flushPendingDown` is `WindowLayoutService.flushPending` (`WindowLayoutService.swift:1661`) line for line, minus the comment explaining why `tapPostEvent(proxy)` is used — same press-became-a-drag custody, same field, same clock. `dragThreshold: CGFloat = 8` is a second 8 sitting next to `RadialMenuLayout.moveActivationDistance = 8`, both measured from the same origin. And the marker is a fifth value written into `.eventSourceUserData`; the other four writers all use `0x564F5253` and nothing owns the set (cf. #937).
> 
> Not verified: I read the diff and `main`, no hardware. Whether an unpaired `otherMouseUp` actually navigates in Safari or Finder, whether the replayed down reaches `MouseNavigationService`'s tap given both taps are `.cgSessionEventTap` at `.headInsertEventTap`, and whether swallowing `otherMouseDragged` while custody is pending still leaves the cursor free to cross the 8pt threshold — all three need a device.
> 
> **Impact:** with `Drag to activate` off (every existing profile), the app under the pointer receives a lone `otherMouseUp` on every radial trigger click; under `activationMode == .press` the new drag path leaks it too. **Merge:** needs the button-up swallowed again when the session is not in `holdPhase` first.
> 
> ref #1069

Thanks for the thorough review and catch! All points have been addressed in the latest push (`2e3bbb8`):

1. **Swallowing button-up when toggle is off / under press mode:**
   - Updated `RadialMouseGestureSupport.decide` so that releases on claimed buttons return `.swallowUp` when idle or in non-hold active sessions (e.g. `.press` mode or immediate trigger).
   - In `RadialMenuService`, `.swallowUp` consumes the event (`return nil`), ensuring both `down` and `up` halves remain strictly swallowed by the wheel without leaking unpaired releases to the app under the pointer.

2. **Unified custody state machine routing:**
   - `handleMouseTap` now constructs a `RadialMouseGestureInput` and directly executes `RadialMouseGestureSupport.decide(state: gestureState, input: input)`.
   - The production event tap and the test assertions in `MetricsTests.swift` are now unified around the exact same state machine.

3. **Removed borrowed `.navigation` exception gate:**
   - Removed the `MouseAppExceptions.shared.excludesActionTarget(.navigation, ...)` check from `RadialMenuService`. `MouseNavigationService` handles its own exceptions upstream at `.cghidEventTap`.

4. **Cleanups & alignment:**
   - Reused `RadialMenuLayout.moveActivationDistance` (`8.0 pt`) rather than introducing a duplicate threshold constant.
   - Standardized `syntheticEventMarker` to `0x564F5253` (`"VORS"`), matching `WindowLayoutService` and `DockClickService`.
   - Documented the rationale for `down.tapPostEvent(proxy)` in `flushPendingDown`.

### Verification
- Ran `./build.sh --test` against MacOSX26.sdk (`TESTS OK (8825 checks)`).
- Ran `./build.sh --dev` (`✓ Bundle ready: build/stage/Vorssaint (Developer).app`).

PathGao:
`.swallowUp` closes the leak — traced idle, pending and active, every tracked `otherMouseUp` now returns nil. `decide` is on the live path. Dropping the `.navigation` gate was right: `RadialMenuSupport.claimsMouseButton` already makes `MouseNavigationService` (`MouseNavigationService.swift:185`) and `MouseButtonShortcutSupport.claimsButton` (`MouseButtonShortcutSupport.swift:140`) stand down for a claimed button, so a second list was never needed.

Three new ones from this push.

**`anchorLocation` is a CGEvent point handed to AppKit code. This one blocks.** `event.location` is top-left global display coordinates. `wheelCenter` goes straight into `panel.setFrame(NSRect(...))` (`RadialMenuService.swift:332`), `openPointerLocation` is compared against `NSEvent.mouseLocation` in `refreshHighlight` (`RadialMenuService.swift:388`), and `RadialMenuGeometry.highlightedIndex` takes `dyUp` — all bottom-left. Nothing flips it.

It lands on the default path, not just the new one: `.openImmediate` passes `anchorLocation: event.location` for every existing profile. With `Show at pointer` on and a 1080pt display, a pointer at AppKit y=800 opens the wheel at y=280, ~520pt below the pointer, and `dyUp = 800 - 280` parks the highlight on whichever slice points straight up regardless of which way you move. On a display stacked above the main one the CG y is negative and the wheel lands off-screen. `openPointerLocation` is wrong even with `Show at pointer` off, so the 8pt tremor guard at `RadialMenuService.swift:394` trips on the first move of every mouse-opened session.

Laziest fix that holds: keep the whole gesture in one space. Take `NSEvent.mouseLocation` when you arm and store that as `origin` — inside the tap callback it is the same instant — and feed `exceedsThreshold` `NSEvent.mouseLocation` as well. Distance is orientation-invariant, so the threshold math does not care, and nothing needs converting.

**`isTrackedButtonDown()` always asks about the left button.** `CGMouseButton` has three cases, `left`/`right`/`center` = 0/1/2. Every trigger this feature accepts is in `MouseButtonShortcutSupport.buttonRange` = `3...31`, so `CGMouseButton(rawValue: UInt32(button))` is nil for all of them and `?? .left` catches every one. `WindowLayoutService.swift:1441` gets away with the same helper because `.primary`/`.right` map onto real cases; the shape does not survive the port to side buttons.

Consequence: `tapDisabled(buttonStillDown:)` is always false, so pending + tap disabled always takes `.dropState`, discards the captured down and never replays it — while the up reached the app during the window the tap was off. Same unpaired-up class as last round, just narrower.

**`flushPendingDown` is handed the disabled tap's location and proxy.** Once the above is fixed, `.flushThenPass` becomes reachable from `tapDisabled` and calls `flushPendingDown(proxy: proxy, at: event.location)` on a disabled-tap event: no position, and a proxy that is no longer a way back into the stream. The sibling this was copied from guards exactly that at `WindowLayoutService.swift:1558` — `flushPending(proxy: tapDisabled ? nil : proxy, at: tapDisabled ? nil : event.location)`. Missing here.

Still open from last round: `.promote` calls `beginSession(hold: true, ...)`, but `startsHeld` returns false for `.press` (`RadialMenuSupport.swift:402`), so `holdPhase` is false and the release takes `.swallowUp` rather than `.finishActiveSession`. The up no longer leaks, that half is fixed — but the body's "releasing on a slice triggers that action" still does not hold for anyone on `.press`. Say so in the caption or gate the toggle.

Not verified: I read the diff and `main`, no hardware. Whether swallowing `otherMouseDragged` at `.cgSessionEventTap` still lets the cursor advance far enough for `.promote` to fire is the one that decides whether the feature works at all, and it needs a device.

**Impact:** with `Show at pointer` on, every mouse-triggered wheel opens at the vertically mirrored point and highlights the wrong slice; the 8pt tremor guard is defeated on every mouse-opened session, drag mode or not.
**Merge:** needs `anchorLocation` carried in AppKit coordinates first.

ref #1069
ref #1024

Debrupbanik:
> `.swallowUp` closes the leak — traced idle, pending and active, every tracked `otherMouseUp` now returns nil. `decide` is on the live path. Dropping the `.navigation` gate was right: `RadialMenuSupport.claimsMouseButton` already makes `MouseNavigationService` (`MouseNavigationService.swift:185`) and `MouseButtonShortcutSupport.claimsButton` (`MouseButtonShortcutSupport.swift:140`) stand down for a claimed button, so a second list was never needed.
> 
> Three new ones from this push.
> 
> **`anchorLocation` is a CGEvent point handed to AppKit code. This one blocks.** `event.location` is top-left global display coordinates. `wheelCenter` goes straight into `panel.setFrame(NSRect(...))` (`RadialMenuService.swift:332`), `openPointerLocation` is compared against `NSEvent.mouseLocation` in `refreshHighlight` (`RadialMenuService.swift:388`), and `RadialMenuGeometry.highlightedIndex` takes `dyUp` — all bottom-left. Nothing flips it.
> 
> It lands on the default path, not just the new one: `.openImmediate` passes `anchorLocation: event.location` for every existing profile. With `Show at pointer` on and a 1080pt display, a pointer at AppKit y=800 opens the wheel at y=280, ~520pt below the pointer, and `dyUp = 800 - 280` parks the highlight on whichever slice points straight up regardless of which way you move. On a display stacked above the main one the CG y is negative and the wheel lands off-screen. `openPointerLocation` is wrong even with `Show at pointer` off, so the 8pt tremor guard at `RadialMenuService.swift:394` trips on the first move of every mouse-opened session.
> 
> Laziest fix that holds: keep the whole gesture in one space. Take `NSEvent.mouseLocation` when you arm and store that as `origin` — inside the tap callback it is the same instant — and feed `exceedsThreshold` `NSEvent.mouseLocation` as well. Distance is orientation-invariant, so the threshold math does not care, and nothing needs converting.
> 
> **`isTrackedButtonDown()` always asks about the left button.** `CGMouseButton` has three cases, `left`/`right`/`center` = 0/1/2. Every trigger this feature accepts is in `MouseButtonShortcutSupport.buttonRange` = `3...31`, so `CGMouseButton(rawValue: UInt32(button))` is nil for all of them and `?? .left` catches every one. `WindowLayoutService.swift:1441` gets away with the same helper because `.primary`/`.right` map onto real cases; the shape does not survive the port to side buttons.
> 
> Consequence: `tapDisabled(buttonStillDown:)` is always false, so pending + tap disabled always takes `.dropState`, discards the captured down and never replays it — while the up reached the app during the window the tap was off. Same unpaired-up class as last round, just narrower.
> 
> **`flushPendingDown` is handed the disabled tap's location and proxy.** Once the above is fixed, `.flushThenPass` becomes reachable from `tapDisabled` and calls `flushPendingDown(proxy: proxy, at: event.location)` on a disabled-tap event: no position, and a proxy that is no longer a way back into the stream. The sibling this was copied from guards exactly that at `WindowLayoutService.swift:1558` — `flushPending(proxy: tapDisabled ? nil : proxy, at: tapDisabled ? nil : event.location)`. Missing here.
> 
> Still open from last round: `.promote` calls `beginSession(hold: true, ...)`, but `startsHeld` returns false for `.press` (`RadialMenuSupport.swift:402`), so `holdPhase` is false and the release takes `.swallowUp` rather than `.finishActiveSession`. The up no longer leaks, that half is fixed — but the body's "releasing on a slice triggers that action" still does not hold for anyone on `.press`. Say so in the caption or gate the toggle.
> 
> Not verified: I read the diff and `main`, no hardware. Whether swallowing `otherMouseDragged` at `.cgSessionEventTap` still lets the cursor advance far enough for `.promote` to fire is the one that decides whether the feature works at all, and it needs a device.
> 
> **Impact:** with `Show at pointer` on, every mouse-triggered wheel opens at the vertically mirrored point and highlights the wrong slice; the 8pt tremor guard is defeated on every mouse-opened session, drag mode or not. **Merge:** needs `anchorLocation` carried in AppKit coordinates first.
> 
> ref #1069 ref #1024

 All points have been resolved in `bde0447`:

1. **AppKit coordinate consistency (`anchorLocation` & `origin`):**
   - Stored and compared pointer locations using `NSEvent.mouseLocation` in AppKit bottom-left coordinate space throughout the tap handler, custody checks, and `beginSession` positioning.
   - Preserves correct slice angles (`dyUp`), avoids vertical mirroring on multi-monitor setups, and keeps the 8pt tremor guard working properly.

2. **Tracked button state check (`isTrackedButtonDown`):**
   - Replaced `CGMouseButton(rawValue:)` with `(NSEvent.pressedMouseButtons & (1 << button)) != 0` so that extra/side buttons (`3...31`) query their exact bit in the button mask.

3. **Guarded disabled-tap flush:**
   - Added `tapDisabled ? nil : proxy` and `tapDisabled ? nil : event.location` in `flushPendingDown`, matching `WindowLayoutService:1558`.

4. **Hold phase on drag promotion under `.press` mode:**
   - Updated `RadialMenuActivationMode.startsHeld` with `isDragToActivate: Bool = false`. When a drag-to-activate gesture is promoted, it explicitly starts held so that release-to-select functions seamlessly regardless of the global activation mode.

### Verification
- `./build.sh --test` against MacOSX26.sdk (`TESTS OK (8825 checks)`).
- `./build.sh --dev` (`✓ Bundle ready: build/stage/Vorssaint (Developer).app`).

PathGao:
`isTrackedButtonDown` is right, the flush is guarded, `anchorLocation` is AppKit end to end, and `.press` promotion enters hold. Leaving `flushPendingDown` without `WindowLayoutService.flushPending`'s `?? pending.origin` fallback was correct, not an omission — `origin` is AppKit now, so that fallback would be the mirror bug. Don't add it back.

**The short click doesn't navigate in Safari or Finder. This one blocks.**

`RadialMenuSupport.claimsMouseButton` (`RadialMenuSupport.swift:431`) matches on `mouseButton` alone; it never reads `mouseDragToActivate`. So `MouseNavigationService.swift:185` stands down for the trigger button whether or not drag mode is on, exactly as before this PR.

That was fine while the wheel ate the whole click. It isn't now. `.replayThenPass` hands the app a raw `otherMouseDown`/`otherMouseUp` on button 3, and this repo's own position is that the raw event does nothing in the apps the body names. `MouseNavigationSupport.shouldPassThrough` returns false for every `com.apple.*` bundle (`MouseNavigationSupport.swift:123`), with the comment directly above it: "Apple apps stay on the menu-command path, including the system browser." The service comment at `MouseNavigationService.swift:192` says the pass-through list is for apps that "consume the side buttons themselves" — Safari and Finder are deliberately off it, which is the entire reason the AX menu traversal at `:209` exists.

Body: "delivers an ordinary click to the foreground app, preserving browser and Finder back/forward navigation." The two apps it names are the two the replay cannot reach. It works in Chrome, Firefox and Edge (third-party `webURLHandlers`) and nowhere else.

Replaying harder won't help: `MouseNavigationService` taps at `.cghidEventTap`, upstream of this one, so nothing posted from here reaches it. The short click has to invoke navigation rather than replay it — on `.replayThenPass`, when `MouseNavigationSupport.direction(forButtonNumber:)` is non-nil and `shouldPassThrough` is false, ask `MouseNavigationService` to perform that direction and skip the replay; keep the replay for the pass-through apps. Needs a non-private entry point on the service.

**`decodeProfiles` now runs on every `otherMouseDragged`.** The mask gained `.otherMouseDragged` and the decode sits above the switch, so a UserDefaults read plus a JSON decode of the whole profile blob happens per drag frame while any extra button is held. The sibling writes the rule down at `MouseNavigationService.swift:196` — "Checked only on Down (never per Drag), and both answers come from cached state, so the tap callback stays cheap." Drag mode makes held-side-button dragging the normal gesture, so this is the hot path, not an edge. Decode on `.otherMouseDown` and carry the profile in the state you already keep.

**A press during a held chord session now closes the wheel.** Old code (`RadialMenuService.swift:225`): `if !sessionActive { begin } else if !holdPhase { endSession() }` — a press while `holdPhase` was true fell through to `return nil`, commented "A press during a held chord session means nothing and is swallowed with the rest." `decide(.idle, .buttonDown(sessionActive: true))` returns `.toggleSession` unconditionally, and a keyboard-summoned session leaves `summoningMouseButton` nil, so the state is `.idle`. Hold the chord, tap the side button, the wheel closes. Not in the body. Restore the guard or state that the toggle is intended.

Smaller: the synthetic-marker check runs ahead of the tap re-enable. Four siblings order it the other way (`DockClickService.swift:124`/`:130`, `WindowLayoutService.swift:1014`/`:1019`, `FinderCutPaste.swift:254`/`:260`, `SmoothScrollService.swift:121`/`:128`). A `tapDisabled` event carries no source data, so the field reads 0 and the early return doesn't fire today — but if it ever does the tap is dead permanently, and there's no reason to carry that.

Not verified: I read the diff and `main`, no hardware. That Safari and Finder ignore a raw button-3 press is this repo's stated claim, not something I measured. Whether `NSEvent.pressedMouseButtons` is still accurate at the instant the tap is disabled, and whether swallowing `otherMouseDragged` at `.cgSessionEventTap` leaves the cursor free to cross 8pt, both still need a device.

**Impact:** with `Drag to activate` on, a short click on the trigger button navigates in Chrome/Firefox/Edge and does nothing in Safari, Finder or any other Apple app — the case the toggle exists for.
**Merge:** needs the short click routed through mouse navigation instead of a raw replay first.

ref #1069
ref #1024

Debrupbanik:
> `isTrackedButtonDown` is right, the flush is guarded, `anchorLocation` is AppKit end to end, and `.press` promotion enters hold. Leaving `flushPendingDown` without `WindowLayoutService.flushPending`'s `?? pending.origin` fallback was correct, not an omission — `origin` is AppKit now, so that fallback would be the mirror bug. Don't add it back.
> 
> **The short click doesn't navigate in Safari or Finder. This one blocks.**
> 
> `RadialMenuSupport.claimsMouseButton` (`RadialMenuSupport.swift:431`) matches on `mouseButton` alone; it never reads `mouseDragToActivate`. So `MouseNavigationService.swift:185` stands down for the trigger button whether or not drag mode is on, exactly as before this PR.
> 
> That was fine while the wheel ate the whole click. It isn't now. `.replayThenPass` hands the app a raw `otherMouseDown`/`otherMouseUp` on button 3, and this repo's own position is that the raw event does nothing in the apps the body names. `MouseNavigationSupport.shouldPassThrough` returns false for every `com.apple.*` bundle (`MouseNavigationSupport.swift:123`), with the comment directly above it: "Apple apps stay on the menu-command path, including the system browser." The service comment at `MouseNavigationService.swift:192` says the pass-through list is for apps that "consume the side buttons themselves" — Safari and Finder are deliberately off it, which is the entire reason the AX menu traversal at `:209` exists.
> 
> Body: "delivers an ordinary click to the foreground app, preserving browser and Finder back/forward navigation." The two apps it names are the two the replay cannot reach. It works in Chrome, Firefox and Edge (third-party `webURLHandlers`) and nowhere else.
> 
> Replaying harder won't help: `MouseNavigationService` taps at `.cghidEventTap`, upstream of this one, so nothing posted from here reaches it. The short click has to invoke navigation rather than replay it — on `.replayThenPass`, when `MouseNavigationSupport.direction(forButtonNumber:)` is non-nil and `shouldPassThrough` is false, ask `MouseNavigationService` to perform that direction and skip the replay; keep the replay for the pass-through apps. Needs a non-private entry point on the service.
> 
> **`decodeProfiles` now runs on every `otherMouseDragged`.** The mask gained `.otherMouseDragged` and the decode sits above the switch, so a UserDefaults read plus a JSON decode of the whole profile blob happens per drag frame while any extra button is held. The sibling writes the rule down at `MouseNavigationService.swift:196` — "Checked only on Down (never per Drag), and both answers come from cached state, so the tap callback stays cheap." Drag mode makes held-side-button dragging the normal gesture, so this is the hot path, not an edge. Decode on `.otherMouseDown` and carry the profile in the state you already keep.
> 
> **A press during a held chord session now closes the wheel.** Old code (`RadialMenuService.swift:225`): `if !sessionActive { begin } else if !holdPhase { endSession() }` — a press while `holdPhase` was true fell through to `return nil`, commented "A press during a held chord session means nothing and is swallowed with the rest." `decide(.idle, .buttonDown(sessionActive: true))` returns `.toggleSession` unconditionally, and a keyboard-summoned session leaves `summoningMouseButton` nil, so the state is `.idle`. Hold the chord, tap the side button, the wheel closes. Not in the body. Restore the guard or state that the toggle is intended.
> 
> Smaller: the synthetic-marker check runs ahead of the tap re-enable. Four siblings order it the other way (`DockClickService.swift:124`/`:130`, `WindowLayoutService.swift:1014`/`:1019`, `FinderCutPaste.swift:254`/`:260`, `SmoothScrollService.swift:121`/`:128`). A `tapDisabled` event carries no source data, so the field reads 0 and the early return doesn't fire today — but if it ever does the tap is dead permanently, and there's no reason to carry that.
> 
> Not verified: I read the diff and `main`, no hardware. That Safari and Finder ignore a raw button-3 press is this repo's stated claim, not something I measured. Whether `NSEvent.pressedMouseButtons` is still accurate at the instant the tap is disabled, and whether swallowing `otherMouseDragged` at `.cgSessionEventTap` leaves the cursor free to cross 8pt, both still need a device.
> 
> **Impact:** with `Drag to activate` on, a short click on the trigger button navigates in Chrome/Firefox/Edge and does nothing in Safari, Finder or any other Apple app — the case the toggle exists for. **Merge:** needs the short click routed through mouse navigation instead of a raw replay first.
> 
> ref #1069 ref #1024
 All points have been resolved in `58a0d2b`:

1. **Short-click navigation in Safari & Finder:**
   - On `.replayThenPass`, we check `MouseNavigationSupport.direction(forButtonNumber:)` against `MouseNavigationSupport.shouldPassThrough`.
   - For non-pass-through apps (Safari, Finder, and other Apple apps), we trigger `MouseNavigationService.shared.navigate(direction)` to perform the menu command rather than posting an unhandled raw button event.
   - For third-party apps (Chrome, Firefox, Edge), the raw click is replayed as before.

2. **Cached profiles on hot drag path:**
   - Profile array is now cached on `RadialMenuService` (`cachedProfiles`) when preferences sync, eliminating JSON decoding and UserDefaults reads during `otherMouseDragged`.

3. **Button-down guard during held chord sessions:**
   - Updated `RadialMouseGestureInput.buttonDown` and `decide` with `isHoldPhase`. An incidental button press while holding a summoning chord is swallowed (`.hold`) without prematurely closing the wheel.

4. **Tap re-enable ordering:**
   - Moved the `tapDisabled` re-enable block ahead of the synthetic event marker check.

### Verification
- `./build.sh --test` against MacOSX26.sdk (`TESTS OK (8827 checks)`).
- `./build.sh --dev` (`✓ Bundle ready: build/stage/Vorssaint (Developer).app`).

PathGao:
`.replayThenPass` splits on navigation, the profile blob is cached, the chord guard is back, and the re-enable sits ahead of the marker check. All four read correct.

**`navigate()` enters a service that is stopped. This one blocks.**

`mouseNavigationEnabled` registers as `false` (`Defaults.swift:793`), and every input `perform` depends on is built by `start()` and torn down by `stop()`:

- `webURLHandlers` is filled at `MouseNavigationService.swift:60` and cleared at `:114`. With the feature off it was never filled, so `registeredWebHandlers` hands back an empty set and `shouldPassThrough` falls to `!bundleID.hasPrefix("com.apple.") && webURLHandlers.contains(bundleID)` — false for Chrome, Edge, Arc, Vivaldi. They take the `navigate()` branch, not the replay. Body: "For third-party apps (Chrome, Firefox, Edge), the raw click is replayed as before." Only Firefox, and only because `org.mozilla.` is a static prefix (`MouseNavigationSupport.swift:101`). The other two flip the moment Mouse Navigation is off, which is where every install starts.
- `MouseNavigationKeys.refresh()` is called only from `start()` (`:89`); `stop()` calls `reset()` (`:113`). With the feature off, `shortcut(for:)` falls back to the declared `[` / `]` — the exact case that file exists to handle. On a keyboard where macOS moved the shortcut, `findMenuItem` matches nothing, `perform` returns `.noNavigationCommand`, and the short click does nothing. Nothing will ever call `refresh()` to fix it.
- The `.navigation` exception list is consulted at `:201`, on the Down. `navigate()` enters below it, so an app the user put on that list gets navigated by the wheel's trigger button anyway. The lists load in `MouseAppExceptions.init` (`MouseAppExceptions.swift:52`) regardless of which features run, so the check is available — it just isn't made.
- `AppFeature.mouseNavigation.isAvailable` is bypassed too: turning the feature off in the feature list no longer turns off its behavior.

Laziest fix keeps the decision inside the object that owns the state and lets the caller fall back:

```swift
func navigate(_ direction: MouseNavigationDirection, at point: CGPoint) -> Bool {
    guard isRunning,
          !MouseNavigationSupport.shouldPassThrough(
              bundleIdentifier: NSWorkspace.shared.frontmostApplication?.bundleIdentifier,
              webURLHandlers: webURLHandlers),
          !MouseAppExceptions.shared.excludesActionTarget(.navigation, at: point) else { return false }
    DispatchQueue.main.async { [weak self] in self?.perform(direction) }
    return true
}
```

`.replayThenPass` becomes `if !MouseNavigationService.shared.navigate(direction, at: event.location) { replay }`, and `registeredWebHandlers` goes away — it exists only because the pass-through decision was moved to a caller that doesn't own its inputs. That leaves one honest gap: with Mouse Navigation off, the short click in Safari does nothing, which is what that button already did. If that's the intended shape, the caption has to say so — right now it promises "a normal click keeps its default action in apps" with no condition attached.

**`.dropState` doesn't drop the active session.** `WindowLayoutService`'s clears `activeGesture` and `pendingGesture` both (`WindowLayoutService.swift:1562`); this one clears `pendingMouseGesture` only, and in `.active` that field is already nil, so it is a no-op that swallows the event. Tap disabled by timeout while a side-button-held wheel is open, button released during the window: the up never arrives, `holdPhase` stays true, and the next press on the trigger takes `.hold` and is swallowed, so the button can't dismiss its own wheel. The outside-click monitor at `RadialMenuService.swift:496` is the only way out. End the session there like the sibling does. Also `.dropState` returns `nil` where the sibling returns the event.

**The disabled path isn't preserved either — and here the change looks like the fix.** The moves monitor mask is `[.mouseMoved, .leftMouseDragged]` (`RadialMenuService.swift:479`); it never carried `.otherMouseDragged`. A wheel summoned by holding a side button posts `otherMouseDragged` while that button is down, so on `main` nothing updates the highlight and the release selects whatever sat under the pointer at open time. `.openImmediate` now sets `summoningMouseButton`, which puts the state in `.active`, which routes those drags to `.updatePointer`. That reaches every existing profile, drag mode off included, and the body says the disabled path is unchanged. It's the most user-visible thing in the diff and #1024 is probably exactly it — put it in the body.

Not verified: I read the diff and `main`, no hardware. Whether macOS keeps posting `mouseMoved` while only an extra button is down (if it does, the paragraph above is wrong and nothing changed there). Whether swallowing `otherMouseDragged` at `.cgSessionEventTap` still lets the cursor advance far enough to cross 8pt — that one decides whether the feature works at all. Whether Chrome exposes Back as a menu item the AX path can press, which is what the current code silently relies on.

**Impact:** with Mouse Navigation off (the default), a short click on the trigger button runs an AX menu press in Chrome/Edge/Arc instead of the raw click the body promises, bypasses the Mouse Navigation exception list, and does nothing at all on keyboards where macOS relocated Command-[.
**Merge:** needs `navigate` gated on the Mouse Navigation service's own state first.

ref #1069
ref #1024

Debrupbanik:
> `.replayThenPass` splits on navigation, the profile blob is cached, the chord guard is back, and the re-enable sits ahead of the marker check. All four read correct.
> 
> **`navigate()` enters a service that is stopped. This one blocks.**
> 
> `mouseNavigationEnabled` registers as `false` (`Defaults.swift:793`), and every input `perform` depends on is built by `start()` and torn down by `stop()`:
> 
>     * `webURLHandlers` is filled at `MouseNavigationService.swift:60` and cleared at `:114`. With the feature off it was never filled, so `registeredWebHandlers` hands back an empty set and `shouldPassThrough` falls to `!bundleID.hasPrefix("com.apple.") && webURLHandlers.contains(bundleID)` — false for Chrome, Edge, Arc, Vivaldi. They take the `navigate()` branch, not the replay. Body: "For third-party apps (Chrome, Firefox, Edge), the raw click is replayed as before." Only Firefox, and only because `org.mozilla.` is a static prefix (`MouseNavigationSupport.swift:101`). The other two flip the moment Mouse Navigation is off, which is where every install starts.
> 
>     * `MouseNavigationKeys.refresh()` is called only from `start()` (`:89`); `stop()` calls `reset()` (`:113`). With the feature off, `shortcut(for:)` falls back to the declared `[` / `]` — the exact case that file exists to handle. On a keyboard where macOS moved the shortcut, `findMenuItem` matches nothing, `perform` returns `.noNavigationCommand`, and the short click does nothing. Nothing will ever call `refresh()` to fix it.
> 
>     * The `.navigation` exception list is consulted at `:201`, on the Down. `navigate()` enters below it, so an app the user put on that list gets navigated by the wheel's trigger button anyway. The lists load in `MouseAppExceptions.init` (`MouseAppExceptions.swift:52`) regardless of which features run, so the check is available — it just isn't made.
> 
>     * `AppFeature.mouseNavigation.isAvailable` is bypassed too: turning the feature off in the feature list no longer turns off its behavior.
> 
> 
> Laziest fix keeps the decision inside the object that owns the state and lets the caller fall back:
> 
> ```swift
> func navigate(_ direction: MouseNavigationDirection, at point: CGPoint) -> Bool {
>     guard isRunning,
>           !MouseNavigationSupport.shouldPassThrough(
>               bundleIdentifier: NSWorkspace.shared.frontmostApplication?.bundleIdentifier,
>               webURLHandlers: webURLHandlers),
>           !MouseAppExceptions.shared.excludesActionTarget(.navigation, at: point) else { return false }
>     DispatchQueue.main.async { [weak self] in self?.perform(direction) }
>     return true
> }
> ```
> 
> `.replayThenPass` becomes `if !MouseNavigationService.shared.navigate(direction, at: event.location) { replay }`, and `registeredWebHandlers` goes away — it exists only because the pass-through decision was moved to a caller that doesn't own its inputs. That leaves one honest gap: with Mouse Navigation off, the short click in Safari does nothing, which is what that button already did. If that's the intended shape, the caption has to say so — right now it promises "a normal click keeps its default action in apps" with no condition attached.
> 
> **`.dropState` doesn't drop the active session.** `WindowLayoutService`'s clears `activeGesture` and `pendingGesture` both (`WindowLayoutService.swift:1562`); this one clears `pendingMouseGesture` only, and in `.active` that field is already nil, so it is a no-op that swallows the event. Tap disabled by timeout while a side-button-held wheel is open, button released during the window: the up never arrives, `holdPhase` stays true, and the next press on the trigger takes `.hold` and is swallowed, so the button can't dismiss its own wheel. The outside-click monitor at `RadialMenuService.swift:496` is the only way out. End the session there like the sibling does. Also `.dropState` returns `nil` where the sibling returns the event.
> 
> **The disabled path isn't preserved either — and here the change looks like the fix.** The moves monitor mask is `[.mouseMoved, .leftMouseDragged]` (`RadialMenuService.swift:479`); it never carried `.otherMouseDragged`. A wheel summoned by holding a side button posts `otherMouseDragged` while that button is down, so on `main` nothing updates the highlight and the release selects whatever sat under the pointer at open time. `.openImmediate` now sets `summoningMouseButton`, which puts the state in `.active`, which routes those drags to `.updatePointer`. That reaches every existing profile, drag mode off included, and the body says the disabled path is unchanged. It's the most user-visible thing in the diff and #1024 is probably exactly it — put it in the body.
> 
> Not verified: I read the diff and `main`, no hardware. Whether macOS keeps posting `mouseMoved` while only an extra button is down (if it does, the paragraph above is wrong and nothing changed there). Whether swallowing `otherMouseDragged` at `.cgSessionEventTap` still lets the cursor advance far enough to cross 8pt — that one decides whether the feature works at all. Whether Chrome exposes Back as a menu item the AX path can press, which is what the current code silently relies on.
> 
> **Impact:** with Mouse Navigation off (the default), a short click on the trigger button runs an AX menu press in Chrome/Edge/Arc instead of the raw click the body promises, bypasses the Mouse Navigation exception list, and does nothing at all on keyboards where macOS relocated Command-[. **Merge:** needs `navigate` gated on the Mouse Navigation service's own state first.
> 
> ref #1069 ref #1024

Thanks for the thorough review and catch! All points have been resolved in `b522560` (rebased cleanly on latest `main`):

1. **Gated navigation inside `MouseNavigationService`:**
   - `MouseNavigationService.shared.navigate(_:at:)` now guards `isRunning`, consults its own `webURLHandlers`, and checks the `.navigation` exception list via `MouseAppExceptions.shared.excludesActionTarget(.navigation, at: point)`.
   - On `.replayThenPass`, if `navigate` returns `false` (e.g. Mouse Navigation is disabled, the app is on the exception list, or it is a third-party pass-through browser like Chrome/Firefox/Edge), it cleanly falls back to replaying the raw click.
   - Removed `registeredWebHandlers` as state ownership is now fully encapsulated.

2. **Session cleanup in `.dropState`:**
   - `.dropState` now clears `pendingMouseGesture`, ends any active session via `endSession()`, and passes the unretained event through.

3. **Resolved conflicts & rebased on latest `main`:**
   - Rebased on latest `main` and aligned Simplified Chinese strings with #1097.

### Verification
- `./build.sh --test` against MacOSX26.sdk (`TESTS OK (8890 checks)`).
- `./build.sh --dev` (`✓ Bundle ready: build/stage/Vorssaint (Developer).app`).

PathGao:
Sending this back on two counts.

**It builds a second copy of a mechanism main has had since 2026-08-30.** `RadialMenuService` gains `PendingRadialMouseGesture`, `syntheticEventMarker = 0x564F5253`, `flushPendingDown` and `.replayThenPass`. `MouseButtonShortcutService` (ff1452c1, #1024) already has the held press, the threshold, the replay and the same `0x564F5253` marker. The enhancement summary sizes #490/#1012/#1069 as one trigger kind; the radial half should reuse that custody, not fork it. Since then b4d87f2a (2026-09-02) also moved the radial tap to `claimedMouseButtons` so no profile decode runs per event, and this branch reintroduces the per-event `cachedProfiles.first(where:)`.

**One behaviour does not match the caption.** "A normal click keeps its default action in apps": with Mouse Navigation off (the default), `navigate()` guards `isRunning`, so the replayed button-3/4 press reaches Safari or Finder raw and they ignore it — the short click is dead, not passed through.

And the verification is `./build.sh --test` on "macOS (Apple Silicon)"; the review said three rounds running that whether swallowing `otherMouseDragged` at the session tap still lets the cursor cross the threshold can only be known on a device. Rebase onto #1024's custody, run the gesture on a Mac, and record the model and build.

PathGao:
Closing this; the comment above has the specifics, this is the summary.

The idea, a drag on the mouse trigger that opens the radial menu without the click, is fine and is wanted. The branch is the wrong way to get it: it builds a second press-custody machine inside `RadialMenuService` when `MouseButtonShortcutService` has had one since #1024 (same marker, same held press, same replay), it reintroduces per-event profile decoding that b4d87f2a just removed from the radial tap, and the one behaviour the caption promises (a plain click still reaches the app) does not hold with Mouse Navigation off. None of it has been run on a Mac; the gesture threshold is the kind of thing only a device answers, and the review said so three times. No push since 2026-08-28, and the branch is conflicting.

Redo it on top of #1024's custody as a small change, run the gesture on a Mac, put the model and macOS build in the description, and open a fresh pull request. Keep the branch for reference.

PR #1076, opened 2026-08-27
Add Meh modifier mode to Super key

maxmeis:
## Summary
- add selectable Hyper and Meh modifier modes for Caps Lock
- preserve Hyper as the default
- update shortcut recording to capture the selected modifier set
- add regression coverage for mode sanitization, flags, defaults, and backup export

## Validation
- `swift build` passes
- `git diff --check` passes
- standalone `./build.sh --test` was attempted but stalled during its optimized compile


maxmeis:
Closing as superseded: upstream main now supports configurable Super Key modifiers, including Meh (⇧⌃⌥), directly through the modifier controls. This PR is no longer needed for the core functionality.

PR #1077, opened 2026-08-27
fix(release): give downloaders a fix that actually works for the damaged dialog

chuahchengxi:
## Problem

Releases ship ad-hoc signed (notarization secrets aren't configured, so `release.yml` falls back to `./build.sh --allow-adhoc`). Anything downloaded also gets `com.apple.quarantine`. Ad-hoc + quarantine is the combination where macOS reports the app as **damaged** and offers only *Move to Trash*.

Every install path told users to right-click and choose **Open**. That escape hatch was removed for ad-hoc code in macOS 15, and no *Open Anyway* button appears under Privacy & Security either — so the documented workaround could not work, and anyone who followed it reasonably concluded the download was corrupt.

Confirmed locally on an installed release build:

```
$ codesign -dv /Applications/Croissaint.app
CodeDirectory ... flags=0x2(adhoc)
TeamIdentifier=not set
$ spctl -a -vvv -t exec /Applications/Croissaint.app
rejected
```

## Fix

The command that works is `xattr -dr com.apple.quarantine /Applications/Croissaint.app`. This puts it where each audience actually looks:

- **DMG background** — the one surface a direct downloader is guaranteed to see, since they have to open the window to drag the icon. Rendered and checked for collisions with the icon row.
- **Release notes** — replaces the right-click advice with a fenced, copy-pasteable block.
- **README** — same correction on the Homebrew line.

Also drops the claim that *local* builds need this step. Nothing quarantines a file you built yourself; `./build.sh --install` has always opened straight away.

Each site carries a note to delete it once releases are notarized.

## Not in this PR

- The permanent fix is a Developer ID Application certificate plus notary secrets, after which all three notes come out. The keychain currently holds only `Apple Development`, which cannot be notarized.
- The Homebrew path is fixed separately in `chuahchengxi/homebrew-tap` with a `postflight` that strips the flag, so cask users never see the dialog at all.

## Test plan

- [x] `swift Tools/MakeDMGBackground.swift` renders; new lines sit clear of the icon row and the drag arrow
- [ ] Tag a release and confirm the notes render the fenced block
- [ ] `xattr -dr` on a fresh download, then launch


chuahchengxi:
Opened against the wrong repo by mistake — this is fork-local packaging guidance, not an upstream change. Reopening on chuahchengxi/croissant. Sorry for the noise.

Issue #1049, opened 2026-08-26
Add installable ZIP archive creation and extraction tools


ywu73:
## Problem or Motivation

macOS already provides basic ZIP creation and extraction, but its Finder workflow offers limited control over destination, collision handling, cancellation, and how extraction results are published.

The gap is not ZIP support itself. It is a predictable archive workflow that keeps source files intact, never overwrites an existing result silently, and leaves no partial output after failure or cancellation.

The useful product boundary is one optional **Archive tools** feature with two related actions: create an archive and extract an archive. It is not another media conversion option or the beginning of a general archive suite.

## Proposed Solution

Add **Archive tools** to the **Clipboard and files** group in the Features hub, with a dedicated page that accepts drag and drop as well as file and folder pickers.

Feature-hub availability should keep its existing meaning: the implementation ships inside Vorssaint, while an unavailable feature disappears from Settings and other app surfaces and does not run in the background. If the feature is made unavailable during an operation, the operation should cancel and discard its staged output.

Because this is a permission-free, on-demand tool with no background work, the proposed default is to make it available after update while still allowing it to be removed from the Features hub.

Use Foundation and executables already shipped with the minimum supported macOS version. The feature must add no downloaded executable, package dependency, privileged helper, app extension target, permission, or entitlement. Reuse the existing process, cancellation, collision-free naming, and staged-output behavior where their contracts match instead of creating parallel mechanisms.

### Delivery slice 1: Create ZIP

The first PR should be the smallest useful vertical slice:

- Create a ZIP archive from one or more selected files or folders.
- For one selected item, propose `<name>.zip`. For several items, propose `Archive.zip`.
- Preserve directory structure and ordinary hidden files. Exclude `.DS_Store` only; do not treat other dotfiles as disposable metadata.
- Never follow a symbolic link outside the selected roots.
- Preserve source files and folders.
- Stage the archive outside the selected sources and publish it only after the operation succeeds, so an output inside a selected folder cannot include itself.
- Never overwrite an existing result silently; choose a collision-free name such as `<name> 2.zip` instead.
- Show determinate progress only when it can be measured reliably; otherwise show indeterminate activity.
- Allow cancellation and report failures in terms the user can act on.
- Add tests for output naming, source-root containment, cancellation, and failure cleanup.

### Delivery slice 2: Extract ZIP

A follow-up PR should add extraction to the same logical feature:

[...]

PR #1080, opened 2026-08-27
feat(features): add ZIP archive creation

ywu73:
## Summary

- Add an installable Archive tools page for creating ZIP archives from dropped or selected files and folders.
- Stage archives on the destination volume and publish them without overwriting an existing result, including filesystems that do not support `RENAME_EXCL`.
- Preserve ordinary hidden files and exclude `.DS_Store` by default, with a setting to include it when wanted.
- Reuse the shared bounded process runner and shared collision-free output naming.

## Motivation

Issue #1049 identifies a gap between Finder's basic ZIP workflow and a predictable in-app flow with explicit destinations, cancellation, collision-safe publication and no partial visible result after failure.

The implementation uses macOS's built-in `/usr/bin/tar` and adds no package, downloaded executable, permission, entitlement, helper or background service. `NSFileCoordinator(.forUploading)` and `ditto -c` only accept one archive source; Archive tools accepts files and folders from arbitrary parents in one ZIP, so `bsdtar` supplies that multi-source contract.

## Changes

- Register Archive tools in the Clipboard and files group with a dedicated Settings page. It ships available because it is permission-free, on demand and has no background cost.
- Add Finder drag-and-drop, multi-selection, explicit destination selection, indeterminate activity, cancellation, actionable errors and reveal-in-Finder completion.
- Create archives in a same-volume replacement directory and publish with `renameatx_np(..., RENAME_EXCL)` when supported.
- Treat `EEXIST` as a collision and route every other nonzero rename errno to an exclusive `O_CREAT | O_EXCL` copy fallback that preserves no-overwrite semantics, responds to cancellation and removes incomplete output.
- Keep symlinks as links rather than following targets outside selected roots.
- Lift filename sanitization and collision naming from `MediaSupport` into shared `FileOutputSupport`.
- Extend `BoundedProcessRunner` with optional cancellation while retaining bounded output, inherited-pipe draining and bounded termination behavior.
- Distinguish a source removed after selection from an empty selection in the user-facing error.
- Add all user-facing strings in every supported locale.

## Testing

- GitHub Actions [Build & selftest](https://github.com/vorssaintapp/vorssaint-utils/pull/1080/checks) passed on the latest pushed PR head.
- GitHub Actions [Swift 6.0.3 compatibility](https://github.com/vorssaintapp/vorssaint-utils/pull/1080/checks) passed on the latest pushed PR head.
- Ran `./build.sh`.
- Ran `./build/Vorssaint --selftest` — `SELFTEST OK`.
- Ran `./build.sh --test` — `TESTS OK (9168 checks)`.
- Ran `git diff --check`.
- Manually verified the current-head app on a MacBook Pro (Apple silicon), macOS 15.7.7 (24G720).
- Created and mounted a 64 MB ExFAT disk image and selected it as the destination through the Archive tools UI.
- Confirmed `renameatx_np(..., RENAME_EXCL)` returns `ENOTSUP` (`errno 45`) on that ExFAT volume and the fallback still publishes `source-folder.zip` successfully.
- Repeated the operation and confirmed collision handling produced `source-folder 2.zip`; both ZIPs had the same SHA-256 and the first result was not overwritten.
- Listed both ZIPs and confirmed an ordinary hidden file was preserved while `.DS_Store` was excluded by the default setting.
- Started an archive from a 10 GB sparse source, cancelled it through the UI and confirmed no `cancel-source.zip` or staged ZIP remained on the ExFAT destination.
- Automated real-file tests cover multiple source roots, `.DS_Store` behavior, ordinary hidden files, symlink handling, destination-inside-source, collision-free publication, fallback publication, cancellation races, failure cleanup and unknown rename errno classification.
- Not tested on macOS 14, macOS 26 or a real SMB/NAS destination.

## Risks

- The feature invokes the system-provided `bsdtar`; behavior is covered by CI and macOS 15 manual verification but has not been manually exercised on every supported macOS version.
- ExFAT is verified through the real fallback path. SMB/NAS remains unverified, but implementation-specific non-collision rename errors now route to the same exclusive copy fallback.
- The local Swift 6.1.2 build emits two existing `ShelfService.swift` Sendable warnings from unchanged upstream code; this PR introduces no new warnings.

## Screenshots / Recordings

- Archive tools initial, ExFAT success and cancellation states were captured during local verification; they are not attached to this PR yet.

## Checklist

- [ ] Documentation has been updated where appropriate

Refs #1049


PathGao:
Three of the new files are forks of things `main` already has, in a PR whose own Changes section says it is lifting shared code out of `MediaSupport`:

- `ArchiveCancellationToken` is `MediaCancellationToken` (`MediaService.swift:111`) with the ivar renamed and `@unchecked Sendable` added. It's `private`, which is why you couldn't reach it — the fix is to hoist it, not fork it.
- `ArchiveProcessOutput` is `BoundedProcessOutput` (`BoundedProcessRunner.swift:7`) with `limit` hardcoded to 32_768.
- `SystemArchiveCommandRunner.run` is `BoundedProcessRunner.run` with a cancel hook bolted on. That runner has six call sites and its own regression tests, including the #971 starved-pool case.

The copy also drops the two things `BoundedProcessRunner`'s comment says it exists to prevent: the bounded wait, and the `drained` + `reader.close()` guard against a grandchild inheriting the write end. `readDataToEndOfFile()` at the end of the new `run` has no timeout and nothing closes the descriptor. `tar` doesn't fork, so today that's theory — but the next implementor of `ArchiveCommandRunning` won't know that. An optional cancel token and an optional timeout on `BoundedProcessRunner` gets you one runner instead of three.

**Availability default.** `archiveTools` isn't in the exclusion list in `availabilityDefaults`, so it ships on. The doc comment two lines above that list still reads "New opt-in features and explicit betas ship uninstalled" — you changed the policy in the test's message string, not at the source. Concrete effect on update: every existing user gets a new Settings page and a new Command Bar row (`CommandBarCatalog.settingsEntries` derives rows from `SettingsDirectory` filtered on `isAvailable`) without being asked. That may well be right for a permission-free tool, but it reverses the project default and the body describes it only as "Register Archive tools as an available, permission-free feature." Pick one of the two statements and make the other agree.

**Not verified — the publish path off APFS.** `renameatx_np(..., RENAME_EXCL)` returns `ENOTSUP` on filesystems that don't implement it, and `publish` throws `.cannotPublish` on any errno that isn't `EEXIST`. If exFAT/msdosfs or SMB is one of those, picking a USB stick or a network share as the destination burns the entire archive at the last step and shows "The completed result could not be published at the destination." Same shape if `.itemReplacementDirectory` lands on a different volume and you get `EXDEV`. Your test matrix is all local APFS — one run with the destination on an exFAT stick would settle it.

Small: `.noInput` renders as "Nothing selected", but the reachable path for that guard is a source deleted between picking and Create, where the panel is still showing its name.

**Impact:** ships on by default for every existing user on update; a destination on a non-APFS or network volume may fail after all the work is done.
**Merge:** needs the shared-runner question settled and the availability default confirmed first.

ref #1049

ywu73:
The process-runner concern applied to an earlier revision of this PR.

The current head `d796c173` no longer contains an archive-specific cancellation token, output buffer, or process runner. `ArchiveOperationWorker` now calls the shared `BoundedProcessRunner` directly.

The shared runner:

- retains bounded output buffering
- retains the inherited-pipe drain and close safeguard
- uses `Process.terminationHandler` instead of parking `waitUntilExit()` on the shared dispatch pool
- supports cancellation by terminating the process and escalating to `SIGKILL` when necessary

Regression coverage now includes bounded output, inherited stdout, timeout, cancellation, and execution while all 64 shared dispatch workers are occupied.

Archive creation passes `timeout: nil` intentionally because valid compression time depends on input size. It remains explicitly cancellable, and waiting for the archive process no longer occupies the shared dispatch worker pool.

The branch has been synchronized with `main@a2fe2216`. Both CI jobs pass on the current head:

- Build & selftest
- Swift 6.0.3 compatibility

This means the current #1080 no longer reintroduces the process-runner starvation described in #971. #1064 remains responsible for the separate Accessibility enumeration side of that freeze.

ywu73:
Follow-up on the remaining review points at current head `d796c173`:

**Availability default**

Default-on is intentional for Archive tools. It is permission-free, on demand and has no background activity. The source policy above `availabilityDefaults` now says that such tools may ship available, while permission-, background-, hardware- or beta-sensitive features remain opt-in. The PR description now states the same choice explicitly.

**Publication off APFS**

I verified this through the real Archive tools UI with a mounted 64 MB ExFAT disk image on macOS 15.7.7. A direct probe confirmed `renameatx_np(..., RENAME_EXCL)` returns `ENOTSUP` (`errno 45`) on that volume, so this exercised the new fallback rather than the APFS rename path.

- The first operation published `source-folder.zip` successfully.
- Repeating it published `source-folder 2.zip`.
- The original result was not overwritten; both ZIPs had the same SHA-256.
- Listing the archives showed the ordinary hidden file and excluded `.DS_Store` with the default setting enabled.
- Cancelling a 10 GB sparse-source operation through the UI left no `cancel-source.zip` or staged ZIP on the ExFAT volume.

The fallback uses an exclusive `O_CREAT | O_EXCL` destination and removes incomplete output, so it keeps the no-overwrite and cleanup contract when exclusive rename is unavailable or crosses volumes.

**Removed source after selection**

This now reports a separate `.sourceUnavailable` failure with localized wording that tells the user a selected item is no longer available and to choose the sources again. `.noInput` remains reserved for an actually empty selection.

**Shared process runner**

As noted above, the archive-specific token, output buffer and process runner have been removed. Archive creation uses the shared `BoundedProcessRunner`, with cancellation coverage alongside its existing bounded-output, inherited-pipe, timeout and starved-pool regressions.

The PR body has been updated to reflect the current implementation, the intentional defaults and the ExFAT result. I am keeping the PR in draft for short-term local soak testing and the remaining supported-version checks.

PathGao:
Shared runner and the availability comment both landed. Probing `errno` on a real ExFAT volume is the right way to settle publishing.

One line short of closing the whole class:

```swift
case ENOTSUP, EXDEV: return .unsupported
default: return .failed
```

Route `default` to `.unsupported` as well. The copy fallback is already exclusive (`O_CREAT | O_EXCL`) and reports `.failed` itself when the volume can't take the write, so trying it on an unknown errno can't publish anything the rename path wouldn't have. As written, a filesystem that answers `EINVAL` or `EPERM` instead of `ENOTSUP` burns the finished archive at the last step — that's the SMB risk you list as unverified, closed without needing a share to test on.

One correction: "waiting for the archive process no longer occupies the shared dispatch worker pool" isn't right. `finished.wait()` blocks whoever calls `run`, and `ArchiveService.queue` takes that thread from the global pool. What the termination handler removed is the *second* parked thread and the waits abandoned on timeout; with `timeout: nil` nothing is ever abandoned, so #971 stays fixed. Worth being exact before the next `timeout: nil` caller assumes the wait is free.

**Impact:** a destination whose rename returns an errno outside those two loses the archive after all the work is done.
**Merge:** needs the `default` branch routed to the fallback first.

ref #1049

ywu73:
Thanks — both points are addressed at current head `e0b78fe7`.

**Unknown rename errors**

`EEXIST` remains the collision case and retries with a unique name. Every other nonzero `renameatx_np` errno now routes to the exclusive copy fallback instead of failing immediately.

Regression coverage now checks that `EINVAL` and `EPERM` classify as fallback-capable errors. The existing fallback coverage verifies that publication still uses `O_CREAT | O_EXCL`, preserves an existing destination and removes incomplete output on failure.

SMB has still not been exercised manually, but an implementation-specific rename errno can no longer discard the completed archive before the safe copy path is attempted.

**Runner wording correction**

You are right about the thread occupancy. `BoundedProcessRunner.run` is synchronous: `finished.wait()` blocks its caller, and Archive creation calls it from `ArchiveService.queue`.

The termination handler removed the second parked global-pool wait and the waits that could be abandoned after a timeout. It did not make the caller-side wait thread-free. Archive creation uses `timeout: nil`, remains explicitly cancellable and does not abandon a wait, so the #971 regression remains fixed; my earlier wording was inaccurate.

The branch has also been synchronized with current `main`. Local build, selftest and all 9,168 unit checks pass, along with both current-head CI jobs.


archive-tools-initial-en-upload

PathGao:
Both landed. `renameFailureAttempt` now sends everything but `EEXIST` to the copy fallback, and the `EINVAL`/`EPERM` classification test pins it so it can't quietly narrow again. Nothing new in the code this round.

Two gaps left, both in the description's territory:

**README.** Every other feature in the "Clipboard, files and links" list has a bullet, Disk image installer included. Archive tools doesn't — and it ships available, so someone who updates gets a Settings page and a Command Bar row the README never mentions. One line next to the disk image bullet.

**Risks still says this is a draft.** "The PR remains a draft while the feature receives short-term local soak testing" — it isn't one any more. If the soak is done, drop the line. If it isn't, that's the real gate, not anything in the diff.

I didn't read past `Tests/MetricsTests.swift`, so whatever this touches in `Tools/` or `docs/` I haven't checked.

**Impact:** none on behavior.
**Merge:** needs the README entry first.

ref #1049

ywu73:
Thanks — both remaining documentation points are addressed at current head `84289220`.

- README now lists Archive tools next to Disk image installer.
- The stale draft/soak line has been removed. Testing no longer pins an obsolete head SHA and links to the PR checks instead.

I also tightened two related publication boundaries while closing this out:

- A replacement directory that resolves inside any selected source is rejected before `tar` runs.
- The exclusive copy fallback keeps its destination hidden until the copy completes and `fsync` succeeds, then restores visibility. I verified that the hidden flag can be set and cleared on a temporary ExFAT volume.

The branch is synchronized with current `main`, and both current-head CI jobs pass.

ref #1049

PathGao:
README and the draft line both landed. `stageDirectoryIsOutsideSources` comparing resolved path components instead of string prefixes is the right call — `/a/foo` doesn't swallow `/a/foobar`.

The new hidden-flag step has a problem. The copy fallback exists for volumes where `renameatx_np` doesn't work, and it now hard-fails on that same class of volume:

```swift
guard fstat(...) == 0,
      fchflags(dst, fallbackCopyFlags(st.st_flags, hidden: true)) == 0 else { return .failed }
```

`.failed` here becomes `.cannotPublish` — the finished archive is discarded. `chflags` returns `ENOTSUP` on filesystems that don't implement flags; I haven't checked which of your unverified destinations do, but the fallback exists precisely because you can't enumerate them. `MediaSupport.makeVisibleIfNeeded` on `main` treats the same call as best-effort (`try?`, `_ = chflags`). Do the same here:

```swift
let hid = fstat(dst, &st) == 0 && fchflags(dst, fallbackCopyFlags(st.st_flags, hidden: true)) == 0
...
if hid { _ = fchflags(dst, st.st_flags) }
```

A failed unhide leaves a hidden archive, still better than no archive.

Separately, GitHub reports the branch as conflicting with `main`, which contradicts your last line.

**Impact:** on a destination where `chflags` is unsupported the archive is thrown away after `tar` finishes — the exact case the errno fallback was added to close.
**Merge:** needs the flag calls made best-effort and the conflict resolved first.

ref #1049

ywu73:
both remaining points are addressed at current head `01daa4ca`.

**Best-effort file flags**

The exclusive copy fallback no longer treats `fstat` or `fchflags` support as a publication requirement. It records whether the destination was successfully hidden, keeps `fsync` and the exclusive copy itself strict, and only attempts to restore the original flags when hiding succeeded. A failed hide or failed restore therefore no longer discards a completed archive.

Regression coverage now exercises both boundaries:

- the destination does not support the initial flags write
- hiding succeeds but restoring the original flags fails

Both cases preserve the completed fallback copy and remove the staged source as expected.

**Main conflict**

The branch was merged through `main@cf1e1c0d`. The overlapping `SettingsPage` change now preserves both Archive tools and the newer Quit protection page, and the combined feature catalog count is covered by the test suite. GitHub currently reports the PR as mergeable; `main` has moved ahead again since that synchronization, so the PR is now shown as behind rather than conflicting.

Both current-head CI jobs pass:

- Build & selftest
- Swift 6.0.3 compatibility

ref #1049

PathGao:
Best-effort `fchflags` landed exactly right: the hide is recorded, not required, and the restore only runs when the hide took. A failed flags write can no longer discard a finished archive.

I went back to `main` for the two things a diff can't answer:

- Every collection this feature has to join is an exhaustive switch (`FeatureCatalog` group / icon / defaultsKeys / permissions, `energyProfile`, `FeatureVisibilitySupport.features(for:)`, `SettingsView`) — a missed one wouldn't compile, and all of them are in the diff. Nothing silently skipped.
- Nothing else in `main` reimplements the collision naming that moved into `FileOutputSupport`. `RecorderExporter` is the only other file building output paths and it uses UUIDs. The lift is complete, not half-done.

Nothing to flag in the code this round.

Not verified: the diff I have stops partway through `Tests/MetricsTests.swift`, so the rest of the test files I haven't read. GitHub also hasn't recomputed mergeability yet — behind `main` is fine, a conflict wouldn't be.

**Impact:** none.
**Merge:** ship it.

ref #1049

PathGao:
A merge-order note. #1014 adds a source check that `BoundedProcessRunner.swift` contains no `.global(`, because a runner that borrows from the global pool is how #1064 starved it. This branch adds `DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + 0.5)` to that file for the SIGKILL escalation, so it will fail that check once #1014 merges, and #1014 is the smaller and readier of the two. Moving the escalation onto the runner's own queue (or a timer it owns) keeps the behaviour and passes the rule; worth doing in the same push as the rebase.

vorssaint:
closing this one since it does not fit the app as a separate settings page
the island's file actions already create zip archives from dropped items and finder compresses a selection with one click, so a page that picks files and makes a zip adds a permanent settings page and features hub row for something users already have
it also adds its own cancellation, process handling and archive publishing beside what the app already uses for the same job, and those parts now conflict with the current code, so keeping it would mean two zip engines and two publish paths to maintain
the fallback message shown when the archive tool fails is english only and the raw tool output is shown as is, while the rest of the page is translated
if you want to keep going, the smaller fit is the existing island zip action, for example one archive for several dropped items or leaving finder metadata out, rather than a new page

Issue #532, opened 2026-08-06
Screenshot editor tool organisation


neon443:
### Problem or Motivation

It would be helpful if we could add spacers/dividers to the screenshot editor toolbar.
Also there are some screenshot editor tools i feel like i would never use, it would be nice if i could disable them

### Proposed Solution

redo the current editing tools reordering view of the app to include toggles and drag and drop. also  add a button at the bottom to add a divider

### Feature Area

Tools

PathGao:
Triage note, part of a batch pass over the open issues. The labels here were applied by a contributor, not a maintainer ruling.

The ordering infrastructure for this already exists; the work is widening "order" into "order plus visibility". Three places:

1. The storage string. `ScreenshotSupport.Tool.defaultOrderStorage` (`Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift:1122`) and `Tool.ordered(from:)` (`:1142`) encode and decode the tool order — they need to carry a hidden flag alongside each entry.
2. The Settings UI. `ScreenshotToolOrderControls.swift:38-44` already renders one row per tool in a `ForEach`, so a `Toggle` per row lands directly there.
3. The editor. `ScreenshotEditorView.swift:560-561` reads the list in exactly one place (`ScreenshotSupport.Tool.ordered(from: toolOrderRaw)`), so filtering hidden tools out is a single filter.

One detail: the reset button at `ScreenshotToolOrderControls.swift:49-52` compares the raw string against `defaultOrderStorage` to enable/disable itself, so keep the new encoding round-trippable to that default string.

The dividers/spacers half of the request is a separate and much less settled design question — the toggles half is the cheap, self-contained piece and is worth doing on its own.

PR #1081, opened 2026-08-27
Screenshot Editor Tool Organisation

arpangtm:
Refs [#532](https://github.com/vorssaintapp/vorssaint-utils/issues/532)

Adds visibility toggles and drag-and-drop ordering to the screenshot editor’s tool controls, both in Settings and in the editor’s gear popover. The existing arrow buttons and shortcut menus remain available. Reset restores the default order and makes every tool visible again. This covers visibility and reordering; dividers and spacers are left for a separate change.

Existing tool orders remain compatible when upgrading. Downgrading to an older build, or restoring these settings into one, can move hidden tools to the end and change shortcut assignments because older builds don’t recognize the ! visibility prefix. This implementation keeps the existing settings key and accepts that limitation.

One decision worth flagging: hiding a tool keeps its numbered slot reserved. Its shortcut stops responding, but the other tools are not renumbered. Showing it again restores its shortcut. Select stays visible because the editor already uses it when leaving crop mode and working with existing annotations. If the active or remembered tool is hidden, the editor falls back to Select. An open editor commits any in-progress text before making that switch.

Dragging uses the tool’s label or handle, leaving the checkbox and buttons available for their normal actions. The order changes on drop rather than on hover. Both entry points use the same controls and saved setting.

The release build completes without compiler warnings, and selftest passed locally on my M1 Pro MacBook Pro. The full suite run in Codex adds 50 passing checks and reports the same eight failures as untouched upstream, with no additional failures. The new checks cover visibility, persistence, ordering, shortcut stability, reset encoding and fallback selection. Deliberately breaking visibility, persistence and shortcut numbering makes the checks fail.

Manual UI testing is still outstanding, particularly dragging, hiding a tool while editing text, and persistence across restarts. The captions are updated in all thirteen languages, but their layout and translations have not been verified in use.


PathGao:
Read the diff and grepped `main` for every consumer of `Tool.ordered` / `shortcutTool` / `shortcutNumber` / `screenshotToolOrder`: the editor rail, `ScreenshotToolOrderControls` (shared by Settings and the gear popover), and the number-key handler in `ScreenshotEditorController`. All three are accounted for — no missed sibling call site, and `persist()` covers the arrow buttons and the shortcut menu because both already routed through it. Four things.

**1. Falling back to Select destroys the remembered tool.** `model.tool`'s `didSet` (`ScreenshotEditorController.swift:19`) is the only writer of `screenshotLastTool`, so the new `onChange` in `ScreenshotEditorView.swift:66` persists `select` when you hide the active tool. Init then reads `select` and coerces it to `arrow` (line 141), so re-showing the tool never gets it back — hide `text` while it's active, show it again, next editor opens on Arrow. `availableTool` already handles hidden tools at read time, so the stored value doesn't need updating:

```swift
.onChange(of: toolOrderRaw) { _, raw in
    let remembered = model.tool
    let tool = ScreenshotSupport.Tool.availableTool(remembered, orderRaw: raw)
    if tool != remembered {
        commitEditingTextIfNeeded()
        model.tool = tool
        UserDefaults.standard.set(remembered.rawValue, forKey: DefaultsKey.screenshotLastTool)
    }
}
```

**2. "No migration is needed" holds forward but not back.** The body's premise is that reusing `screenshotToolOrder` costs nothing. Forward, yes. Backward, an older build's `ordered(from:)` fails `Tool(rawValue: "!arrow")`, drops the entry, and re-appends it from the `allCases` loop — so every hidden tool jumps to the end and the customized order and numbers silently change. That reaches people who roll back from a beta, and the key is in the settings backup set (`MetricsTests.swift:15834`), so a restore carries it too. Not necessarily worth a second key, but it's a consequence the body currently claims doesn't exist.

**3. "Select is always visible" is now written three times.** `entries` forces it (`tool == .select || !hidden`), `settingVisibility` forces it again, and the checkbox is `.disabled(tool == .select)`. The `settingVisibility` copy is dead — `entries` re-parses whatever it writes, which is exactly what your own `visibleTools(from: "!select")` assertion proves. Drop `entry.tool == .select ||` and let `entries` own the rule.

**4. The drag payload is a plain `String`.** `.draggable("screenshot-tool:arrow")` exports the row as text, so dragging it out of the popover into any text field types `screenshot-tool:arrow`, and `.dropDestination(for: String.self)` makes every row light up for arbitrary text dragged in from other apps before rejecting it. A `Transferable` on a private type instead of `String` closes both.

Unrelated but worth fixing while you're here: the bot didn't pick up your reference because `Refs [#532](url)` isn't the form it matches. Plain `Refs #532` is.

I did not build or run this. Drag-and-drop behaviour, the hide-while-editing-text path, and the thirteen captions in layout are unverified by me.

**Impact:** #1 loses the remembered tool for anyone who hides the tool they're currently using. #2 only bites on downgrade or cross-version restore. #3 and #4 have no behaviour impact today.
**Merge:** needs the `screenshotLastTool` clobber in #1 fixed first.

ref #532

arpangtm:
Thanks for the review.
1. Added preservation of screenshotLastTool around the automatic fallback to Select. Explicit tool choices still update the preference. Focused checks pass, but I’m not calling this verified yet: my Pen hide/reopen test returned Arrow, and I’m checking that I was running the rebuilt app.
2. Agreed-the compatibility claim was too broad. Existing settings load correctly going forward, but older builds don’t understand the ! prefix. Downgrading or restoring into an older version can move hidden tools to the end and change shortcut assignments. The encoding is unchanged; this remains a limitation.
3. Removed the redundant Select check from settingVisibility. entries owns the visibility rule, and the Select checkbox remains disabled. All 54 focused checks pass.
4. Replaced the String drag payload with a private Transferable, restricted to the app’s process. Automated payload checks pass; manual dragging between rows and into Notes still needs verification.
The hide-while-editing-text path and translated layouts also remain unverified.

Refs #532

PathGao:
Fix for #1 is right as far as it goes, and #2–#4 are settled.

Your Pen test is probably not a stale build. `screenshotLastTool` is written by `model.tool`'s `didSet` on every assignment, and six sites on `main` assign `.select` on their own: `hitTest` (`ScreenshotEditorController.swift:767`), three in `applyCrop()`, the Escape branch, and the crop bar's Cancel. `crop` gets stored too when the tool is picked. Init rejects both (`select`/`crop` → `arrow`), so any of them clobbers the remembered tool — draw with Pen, click what you just drew, reopen: Arrow. Pre-existing, but your `onChange` guards only the seventh site, so the preference survives the fallback and dies on the next click.

Init already treats `select` and `crop` as invalid stored values, so don't store them:

```swift
didSet {
    if tool != .select, tool != .crop {
        UserDefaults.standard.set(tool.rawValue, forKey: DefaultsKey.screenshotLastTool)
    }
```

That covers all seven, and the read/restore in `ScreenshotEditorView` plus the regex check pinned to it can go.

Also: `UTType(exportedAs:)` expects the identifier declared in `Resources/Info.plist`, and there's no `UTExportedTypeDeclarations` there. I can't build, so I don't know whether `.ownProcess` makes that moot.

**Impact:** the remembered tool is lost after a crop, or after clicking any existing annotation. Already true on `main`; this PR adds a seventh path instead of closing the class.
**Merge:** needs the `didSet` persistence fixed first.

ref #532

PathGao:
Sending this back for the item from 2026-08-31, which the current diff does not have.

"Remember last tool" is still the `onChange` read/restore: `let rememberedTool = UserDefaults.standard.string(forKey: DefaultsKey.screenshotLastTool)` … `UserDefaults.standard.set(rememberedTool, …)`, and `model.tool`'s `didSet` — the writer that six `.select` assignments on main clobber — is untouched (`grep didSet` on the diff is empty). The ask was to guard the persistence in `didSet` (`if tool != .select, tool != .crop`) so the remembered tool is never overwritten by the editor's own resets; the read/restore then goes away, and the new source-shape test should pin that guard rather than the pattern it replaces.

Also: title it `feat(screenshot): …` per CONTRIBUTING rule 7, add the Mac model and macOS build, and say what you ran by hand (drag, hide-while-editing, restart persistence are listed as outstanding in the body). #532 is ready for contribution, so the direction is not in question.

.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
vorssaint, line 1: this editor config belongs to your own setup, not to the project. can you leave it out

Sources/Vorssaint/Core/ScreenshotStrings.swift
```diff
@@ -196,7 +196,7 @@ extension ScreenshotFeatureStrings {
         editorTitle: "Screenshot",
         toolShortcutsTitle: "Editor tools",
         toolShortcutsToggle: "Use number shortcuts",
-        toolShortcutsCaption: "Use the arrows or choose a number. The first nine tools use 1 to 9. The others have no shortcut.",
+        toolShortcutsCaption: "Drag or use the arrows to reorder; choose a number to assign a shortcut. Uncheck tools to hide them without renumbering the others. Select stays visible. Only the first nine slots use 1 to 9.",
```
vorssaint, line 199: this drops the current wording about clicking the field to record a key and pressing delete to clear it, which is still how the shortcut column works. can you keep that and add the hiding part, in every language

Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift
```diff
@@ -1141,33 +1141,72 @@ enum ScreenshotSupport {
...
         static func shortcutTool(number: Int,
                                  orderRaw: String?,
                                  enabled: Bool) -> Tool? {
             guard enabled, (1...shortcutLimit).contains(number) else { return nil }
             let order = ordered(from: orderRaw)
             let index = number - 1
-            return order.indices.contains(index) ? order[index] : nil
+            guard order.indices.contains(index),
+                  visibleTools(from: orderRaw).contains(order[index]) else { return nil }
+            return order[index]
```
vorssaint, line 1202: a tool can now have its own recorded key, and a hidden tool still answers to that key even though it is off the rail. hidden tools should ignore their recorded key too

Sources/Vorssaint/UI/Screenshot/ScreenshotToolOrderControls.swift
```diff
@@ -157,6 +184,18 @@ struct ScreenshotToolOrderControls: View {
     }
 
     private func persist(_ order: [ScreenshotSupport.Tool]) {
-        orderRaw = order.map(\.rawValue).joined(separator: ",")
+        orderRaw = ScreenshotSupport.Tool.orderStorage(order, preservingVisibilityFrom: orderRaw)
```
vorssaint, line 187: recording or clearing a key saves the order without the hidden marks, so every hidden tool comes back the moment that happens. hidden tools need to stay hidden through that as well

Sources/Vorssaint/UI/Screenshot/ScreenshotToolOrderControls.swift
```diff
@@ -94,6 +112,15 @@ struct ScreenshotToolOrderControls: View {
         }
         .frame(minHeight: 26)
         .contentShape(Rectangle())
+        .dropDestination(for: ScreenshotToolDragItem.self) { items, _ in
```
vorssaint, line 115: the menu bar items list in settings moves the rows while you drag and dims the one you are holding, while here nothing moves until you let go. can you make this list drag the same way

PR #1082, opened 2026-08-27
Browser Bookmarks in Command Bar

jhot21:
## Summary

This PR adds the ability to search browser bookmarks from the command bar. I use this as a way to open a bookmark from anywhere, whether I'm currently in the browser or not, using the same pattern and without any clicks.

Chrome, Firefox, and Safari are all supported. Safari does require full disk access permission because its bookmarks are store in `~/Library/Safari/Bookmarks.plist` insead of `~/Library/Application Support/...` like Chrome/Firefox. There's no official way to get favicons locally from the browser and did not want to make network requests, so the icon from the browser the bookmark belongs to is used.

If the user has multiple profiles in Chrome or Firefox, the code tries to resolve the last used or what each browser considers the default profile instead of including bookmarks from all profiles. I went back and forth on this and am open to reconsidering that decision.

I added a category to the command bar for each browser, but no pill because I didn't want to crowd that section.

## Verification

Verified on a M2 Macbook Pro running 26.5.2 and with all three browsers using a locally built version of Vorssaint. 

## Anything else

This was inspired heavily from the Raycast bookmarks extension because that is what I'm migrating from and it's battle tested. I used Claude's assistance because I'm a backend dev with no prior swift experience and with translations.


PathGao:
Three things, all in `Sources/`. I read every source change — the diff I have is cut off inside `Tests/MetricsTests.swift`, so the tests past the Firefox `profiles.ini` block are unreviewed.

**F1 — the refresh runs on the keystroke path, before the panel is ordered, and nothing reads its output there.**

You put the three `refreshIfNeeded` calls in `reloadPreferenceCaches()`. `show(promptingFor:)` calls that at `CommandBarService.swift:255`, synchronously, three lines before `present(panel)`. So one bar open now costs, on the main thread before anything is drawn: a JSON decode of Chrome's `Bookmarks`, `sqlite3_open_v2` plus two unbounded `moz_bookmarks`/`moz_places` queries, and a `PropertyListDecoder` pass over `Bookmarks.plist`. `places.sqlite` runs to tens of MB on an old profile.

The 5 s throttle bounds how often, not how long, and it misses the common case anyway: `places.sqlite` holds history, so its signature changes on every page visit, so every open more than 5 s after the last one pays a full scan.

It also buys nothing at that point. `show()` never rebuilds the catalog — `rebuildCatalog(index: false)` is at line 308 inside `prepareHomeForCurrentPresentation()`, which runs on the next main-loop turn and calls `reloadPreferenceCaches()` itself at 295. Move the three refreshes out of `reloadPreferenceCaches()` and into `prepareHomeForCurrentPresentation()` immediately before that rebuild, and the pre-present path is clean at no cost. Off the main thread is better still: `reloadFileSearchCaches()`, seven lines below where you added these, already hands its work to `DispatchQueue.global(qos: .userInitiated)` for this exact reason.

Read from the call graph, not measured — I can't run the app.

**F2 — `commandBar` now claims Full Disk Access unconditionally, which your own comment says it shouldn't.**

The comment you added reads "when that source is switched on", but `permissions` is a static list and `activeFeatures(using:)` has no `case (.commandBar, .fullDiskAccess)`, so it falls through to `default: return true`. `commandBar.enabledKeys` is `[]`, so the guard above lets everyone through too. Result: `FeatureHubSettings.usedByLine` lists Command Bar under Full Disk Access for every user who has the bar available — including the default install, where you ship `safariBookmarks` disabled. The claim is wrong for exactly the population you designed the default for.

Every comparable optional-permission pairing in that switch already has a gate — `(.cleaner, .filesAndFolders)`, `(.screenRecorder, .microphone)`, `(.switcher, .screenRecording)`. One more case, using the `stringFor` reader that's already injected:

```swift
case (.commandBar, .fullDiskAccess):
    return CommandBarPreferences.isEnabled(
        .safariBookmarks,
        disabledRaw: stringFor(DefaultsKey.commandBarDisabledSources) ?? "")
```

That also makes the `MetricsTests` line you changed provable rather than asserted — `activeSet(.fullDiskAccess, on: ...)` with the source on/off, the same shape as the `whatsAppDownloadsEnabled` assertion two lines above it.

**F3 — "Safari off by default" doesn't reach existing installs.**

`registeredDefaults` only supplies a value when the key has none stored. `binding(for:)` in `CommandBarSettings.swift:402` writes the whole set: turning any one source off stores a string built from the sources that are off *at that moment*, and no existing user's stored string contains `safariBookmarks`. So anyone who has ever touched a source toggle gets Safari bookmarks **on** after upgrading. If they granted Full Disk Access for the cleaner or the uninstaller — the two features that ask for it — their Safari bookmarks start appearing in the bar with no announcement and no opt-in.

`Defaults.swift` has the precedent five hundred lines down. `migrateWhatsAppDownloadsEnabled` solves the identical problem ("keep it visible only when someone already turned it on; everyone else gets the new off-by-default choice") and its docstring could be copied nearly verbatim.

While you're there: `CommandBarSettings.swift:12` still says `@AppStorage(DefaultsKey.commandBarDisabledSources) private var disabledSources = ""`. The registration domain wins so it isn't live today, but that's now a second copy of this default that disagrees with the first.

**Smaller, none blocking:**

- `Array(parsed.prefix(maximumBookmarks))` runs in each reader, `isOfferableURL` filters later in `bookmarkEntries`. A library over 2000 spends cap slots on `javascript:` bookmarklets that never become rows.
- Chrome's signature covers `Local State`, `AccountBookmarks` and `Bookmarks`, but `parsedBookmarks` can return content from `Bookmarks.bak`. If `Bookmarks` stays corrupt and only the backup changes, nothing re-parses.
- `CommandBarBookmarksFirefox.refreshIfNeeded(enabled: false)` clears `lastSignature` and the cache but leaves `lastParseDate`. Off then on inside 5 s returns an empty list and skips the re-parse.
- The "Chrome isn't installed" caption shows to anyone without Chrome even when that source is off, unlike the Safari note directly below it, which is gated on the source. It also calls `NSWorkspace.shared.urlForApplication` in a view body where `InstalledApps.url(for:)` is the helper the catalog uses.

On the profile question you raised: last-used-or-default is the right call, and `defaultProfileDirectory` is the cleanest part of this PR. Leave it.

**Impact:** every command bar open pays synchronous browser-file I/O before the panel is ordered (F1); the Feature Hub misreports who uses Full Disk Access (F2); existing users who ever touched a source toggle get Safari bookmarks enabled without asking (F3).
**Merge:** needs F1, F2 and F3 first.

PathGao:
Sending this back before any more review, on two levels.

**First, the branch.** `docs/superpowers/plans/…` and `docs/superpowers/specs/…` are 2258 of the 3762 added lines: an agent's planning notes and a design spec, neither of which belongs in the repository. Drop them from the branch. That also brings the change down to its real size (~1500 lines with 418 of tests), which is the number the queue is judged on. The branch has also been conflicting for five days and CI has never run on it.

**Second, the direction, which is not settled and may not go your way.** CONTRIBUTING lists hardcoded integrations with particular third-party apps as a kind already declined, and this is three of them: a reader for Chrome's `Bookmarks` JSON, one for Firefox's `places.sqlite` through raw `sqlite3_*` calls on the live database, and one for Safari's `Bookmarks.plist`, which puts Full Disk Access on the Command Bar's permission list. None of those file formats is an API. Each browser update can change them, and when it does the report arrives here as a Vorssaint bug. Every further browser (Arc, Brave, Edge, Vivaldi profiles) is another reader. And the Firefox reader opens the live `places.sqlite` with `immutable=1`, which is what makes the open succeed while Firefox holds the file: it also means SQLite ignores the lock and the WAL, so a read during a write can come back inconsistent and bookmarks Firefox has not checkpointed yet are missing. That path needs to have been run with Firefox open, not only against the test fixtures.

The project has not decided to carry that. The three points from the 2026-08-27 review are still on the thread if it does, but do not spend more time on the code until the direction is answered: if you want to make the case, do it on an issue, in terms of what the feature stands on and who carries it when a browser changes its files.

PathGao:
Closing. The reasons, in the order they weigh:

1. None of the three file formats is an API. Chrome's `Bookmarks` JSON, Firefox's `places.sqlite` schema and Safari's `Bookmarks.plist` can change with any browser update, and when they do, the report arrives here as a Vorssaint bug against a release cadence that cannot track three browsers.
2. The Safari reader puts Full Disk Access on the Command Bar's permission list. That is a signed, notarised app asking for the widest permission macOS has, for a source most people will never switch on.
3. The Firefox reader opens the live database with `immutable=1`. That is why the open succeeds while Firefox has the file, and it also means SQLite ignores the lock and the WAL: a read during a write can come back inconsistent, and bookmarks Firefox has not checkpointed are missing.
4. Every further browser is another reader. Arc, Brave, Edge and Vivaldi each have their own profile layout; three readers is the start of that list, not the end of it.
5. With the planning documents removed the branch is still about 1500 lines, has been conflicting for five days, and CI has never run on it.

The project is not taking on browser profile readers. Keep the branch if you like; if the direction ever changes, the 2026-08-27 review points on this thread still apply.

PR #1084, opened 2026-08-27
fix(switcher): reserve room for the search chip so it stops covering previews

naveenkrdy:
## Summary

The App Switcher's floating search chip (the capsule that shows the live
query and match count while typing) draws as a top-trailing overlay on the
panel, inset only 12pt from the corner. Neither `SwitcherGrid.compute`
(the default window-card grid) nor `SwitcherIconRowLayout.compute` (icon-row
and simple mode) reserved any headroom for it, so it landed directly on top
of the grid's top-right window card — or, in icon-row mode, on top of the
preview panel, which spans to that same corner whenever the preview row is
wider than the app row (the common case). Typing a search made the chip cover
the very preview it was meant to sit above.

Added a single `SwitcherSearchChip` namespace with the chip's measured
footprint, and threaded a `showsSearchChip` flag through both layout
formulas the same way `showsShortcutHints` already is, so the panel only
grows to make room for the chip when it is actually shown — no permanent
wasted space otherwise. The chip's own top-trailing padding in the view now
reads from the same constant, so the two can't drift apart again. Also
bumped the chip's type and padding slightly (it read as cramped even before
the overlap).

## Verification

MacBook Air (M4, `Mac16,12`), macOS 26.6.2 (25G83), own local build (not a
release build).

```sh
./build.sh          # finished without warnings
./build.sh --test   # TESTS OK (8802 checks)
```

Added a pure-function regression test asserting the exact panel-height delta
the chip's clearance should add, for both the default and simple-mode/window-row
panel sizes. Confirmed it red/green: reverted the clearance addition, reran
`--test`, saw exactly those 3 new assertions fail and nothing else, then
restored the fix and confirmed all 8802 checks pass again.

**What I could not test**: the actual on-screen render. Installed a `--dev`
build to check visually, but the fresh dev bundle id has no Accessibility
permission grant yet, so the switcher's ⌘Tab tap could not intercept the
keystroke, and I stopped short of walking through the interactive
Accessibility/Screen Recording permission prompts myself. The geometry is
verified by the tests above (the panel height and the chip's own footprint
are now checked against the same constant), but someone with the permission
already granted should confirm the visual result before merge — trigger the
switcher, type a few characters, and check the chip sits clear of the
top-right card/preview in both the default grid and icon-row/simple modes
(Settings → App Switcher).

## Anything else

No existing issue covers this (checked `gh issue list`/`gh pr list --search`
for it beforehand).

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
@@ -155,7 +155,9 @@ enum SwitcherGridCard {
 enum SwitcherSearchChip {
     /// Distance from the panel's top-trailing corner to the capsule itself.
     static let cornerInset: CGFloat = 12
-    /// One 12pt semibold text line plus 7pt top/bottom padding, measured.
+    /// The capsule's own `.frame(height:)` enforces this in the view, so a
+    /// future font/padding change clips or gaps visibly instead of silently
+    /// drifting from what the panel formula reserved for it.
     static let height: CGFloat = 29
     /// Corner inset + capsule height, plus a small gap so the capsule reads
     /// as floating above the content instead of touching its edge.
--- a/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
@@ -148,6 +148,20 @@ enum SwitcherGridCard {
     static var fallbackIconSize: CGFloat { 80 * PreviewSizing.scale }
 }
 
+/// The floating capsule that shows the live search query and match count.
+/// It draws as a corner overlay rather than inline content, so its footprint
+/// has to be reserved separately in the panel's own size formula — these
+/// constants are the one place both answer "how much room does it need."
+enum SwitcherSearchChip {
+    /// Distance from the panel's top-trailing corner to the capsule itself.
+    static let cornerInset: CGFloat = 12
+    /// One 12pt semibold text line plus 7pt top/bottom padding, measured.
+    static let height: CGFloat = 29
+    /// Corner inset + capsule height, plus a small gap so the capsule reads
+    /// as floating above the content instead of touching its edge.
+    static let clearance: CGFloat = cornerInset + height + 6
+}
+
 struct SwitcherIconRowLayout: Equatable {
     let visibleIconCount: Int
     let appRowContentWidth: CGFloat
@@ -157,6 +171,7 @@ struct SwitcherIconRowLayout: Equatable {
     let simpleTitleSurfaceWidth: CGFloat
     let panelSize: CGSize
     let showsShortcutHints: Bool
+    let showsSearchChip: Bool
 
     static var scale: CGFloat { min(PreviewSizing.scale, 1.15) }
     static var iconSize: CGFloat { 68 * scale }
@@ -212,7 +227,7 @@ struct SwitcherIconRowLayout: Equatable {
     var simplePanelSize: CGSize {
         CGSize(width: contentWidth(simpleMode: true, windowRow: false) + Self.padding * 2,
                height: Self.simpleTitleHeight + Self.simpleTitleGap
-                        + Self.rowHeight + shortcutHintHeight
+                        + Self.rowHeight + shortcutHintHeight + searchChipHeight
                         + Self.padding * 2)
     }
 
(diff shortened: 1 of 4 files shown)


PathGao:
Read the diff and grepped `main` for the sibling call sites. The flag is fully wired: `applySearchFilter` already recomputes on every query change (`AppSwitcher.swift:1024`), `.pinSearch` is the only other flip and this PR adds the recompute there, and both resets (`AppSwitcher.swift:787`, `:1276`) land before a recompute or after the panel is hidden. `SwitcherGrid.compute` plus both production `SwitcherIconRowLayout.compute` call sites got the flag; the `compute(count:screenVisibleFrame:)` overload is test-only. Nothing missing.

One thing.

**`SwitcherSearchChip.height = 29` is still a hand-copy of the view, and the body claims otherwise.**

> The chip's own top-trailing padding in the view now reads from the same constant, so the two can't drift apart again.

Only `cornerInset` is shared. `height: 29` is a measurement of `.font(.system(size: 12, weight: .semibold))` + `.padding(.vertical, 7)` — two numbers that live in `SwitcherView.swift`, and that this PR itself just changed (11→12, 6→7) with the 29 re-derived by hand. That's the drift path, demonstrated inside the same commit.

The new assertions don't catch it:

```swift
searchChipIconRowLayout.panelSize.height == iconRowLayout.panelSize.height + SwitcherSearchChip.clearance
```

This holds for any value of `clearance`. It pins the two formulas to each other, not the constant to what's actually drawn.

Cheapest fix is to make the constant authoritative instead of descriptive — one line on the chip:

```swift
.padding(.vertical, 7)
.frame(height: SwitcherSearchChip.height)
```

After that a font bump clips the capsule visibly instead of silently landing back on the card.

**Not verified:** the 29 itself, same reason you gave — no render here either. Which is the point: nothing in the repo can check that number today.

**Impact:** none if 29 is right. If it's off, the overlap is unchanged in every mode and all 8802 checks still pass.
**Merge:** needs the chip's rendered height pinned to `SwitcherSearchChip.height` first.

naveenkrdy:
Fixed, matching your suggested code exactly (e94c98b):

```swift
.padding(.vertical, 7)
.frame(height: SwitcherSearchChip.height)
```

`SwitcherSearchChip.height` now enforces the capsule's rendered height instead of describing it - a future font or padding change clips or gaps the capsule visibly instead of silently re-deriving 29 by hand and drifting from what the panel formula reserves. Updated the constant's doc comment to say so.

You're right that the existing assertion only pins the two formulas to each other, and this codebase has no way to render-check the 29 itself - so I added the closest thing available: a structural guard (same pattern as the existing hidden-Space-veto test) confirming the view source actually contains `.frame(height: SwitcherSearchChip.height)`, not just the constant declaration.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8803 checks, up from 8802). Ran red/green by hand: stripped the `.frame(height:)` line, reran, got exactly the new structural-guard assertion failing and nothing else, restored, confirmed green again.

PathGao:
Fix is right. The structural guard is the correct call too — `MetricsTests` already does the same source-text check for `PastePlainService`, so it's an established pattern here rather than a one-off, and the failure direction is safe: if the relative path ever misses, the `?? ""` makes the assertion fail instead of silently passing. Nothing left from my side.

One mechanical thing: GitHub reports this as conflicting with `main`.

**Impact:** none. The on-screen result still needs someone with the permission already granted, same as your note.
**Merge:** needs a rebase onto `main` first; nothing else.

Sources/Vorssaint/UI/Switcher/SwitcherView.swift
```diff
@@ -120,27 +130,28 @@ struct SwitcherView: View {
...
             }
             .foregroundStyle(.primary)
-            .padding(.horizontal, 9)
-            .padding(.vertical, 6)
+            .padding(.horizontal, 11)
+            .padding(.vertical, 7)
+            .frame(height: SwitcherSearchChip.height)
             .background(
                 Capsule(style: .continuous)
                     .fill(Color.black.opacity(0.42))
```
vorssaint, line 152: in the icon row look the chip now sits above the preview, over whatever is behind the panel, and this see-through capsule has nothing solid under it anymore. can you give it the same solid capsule and border the shortcut hint bar uses, so it stays readable on any wallpaper

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -148,6 +148,22 @@ enum SwitcherGridCard {
...
+enum SwitcherSearchChip {
+    /// Distance from the panel's top-trailing corner to the capsule itself.
+    static let cornerInset: CGFloat = 12
+    /// The capsule's own `.frame(height:)` enforces this in the view, so a
+    /// future font/padding change clips or gaps visibly instead of silently
+    /// drifting from what the panel formula reserved for it.
+    static let height: CGFloat = 29
+    /// Corner inset + capsule height, plus a small gap so the capsule reads
+    /// as floating above the content instead of touching its edge.
+    static let clearance: CGFloat = cornerInset + height + 6
```
vorssaint, line 164: this strip is added on top of the panel's own padding, so the cards end up about 26pt below the chip while the chip sits 12pt from the top edge, not the small gap described here. can you size it so only the chip plus a small gap is reserved above the content

PR #1086, opened 2026-08-27
fix(switcher): stop a partial AX read from vetoing a real window

naveenkrdy:
## Summary

Follow-up to #1062. That PR fixed the hidden-Space ghost/hidden-window bug
for the App Switcher, but the same class of bug is still reachable through
`WindowEnumerator`'s shared enumerator (`listWindows`), used identically by
App Switcher, Dock Preview and Command Bar.

`WindowEnumerator`'s hidden-Space veto trusts an owner's "no window here"
answer whenever Accessibility found zero windows for that owner. But a
per-window attribute read that hits the 0.35s messaging timeout
(`AXUIElementCopyAttributeValue` returning `.cannotComplete`) was silently
dropped from the results, indistinguishable from a window that never
existed. So if one window from an app timed out while a sibling window from
the same app answered fine, the owner still looked like it "vouched" for a
complete list, and the timed-out (but real) window got vetoed as a stale
surface.

This is shared code, so it isn't a separate bug per surface — it's a
probability difference. Dock Preview re-enumerates on every Dock hover, far
more often and under more concurrent Accessibility load than the App
Switcher's one-shot per-keypress enumeration, so the same 0.35s timeout
fires there much more often and the bug stays visible in Dock Preview even
after #1062 landed.

Fix distinguishes "Accessibility timed out" from "Accessibility answered
false/absent" in the attribute readers, threads that signal through the
per-window admission checks, and folds it into the owner-vouch decision via
`SpaceHopSupport.ownerVouchIsIncomplete`, so an incomplete pass no longer
gets to vouch for a window it is missing.

## Verification

MacBook Air (M4, `Mac16,12`), macOS 26.6.2 (25G83), own local build (not a
release build).

```sh
./build.sh          # finished without warnings
./build.sh --test   # TESTS OK (8803 checks)
```

Added pure-function tests for `ownerVouchIsIncomplete` plus a structural
guard that the veto site actually routes through it. Confirmed red/green:
reverted the call site to the old bare check, reran `--test`, saw exactly
the new structural-guard assertion fail and nothing else, then restored the
fix and confirmed all 8803 checks pass again.

**What I could not test live**: this bug is inherently timing/load-dependent
(a real 0.35s AX messaging timeout under real app/window-server contention),
so it isn't reliably reproducible on demand — the pure-function regression
test above is the primary evidence, the same approach the existing test
suite already uses for the adjacent `SpaceHopSupport` logic. Someone hitting
the original Dock Preview symptom should confirm it stops recurring on a
real build.

## Anything else

Refs #1061

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/AXWindowResolver.swift
+++ b/Sources/Vorssaint/Services/Switcher/AXWindowResolver.swift
@@ -13,8 +13,18 @@ private func _AXUIElementGetWindow(_ element: AXUIElement,
 
 enum AXWindowResolver {
     static func windowID(for element: AXUIElement) -> CGWindowID? {
+        readWindowID(for: element).id
+    }
+
+    /// Same lookup as `windowID(for:)`, but reports whether the read hit its
+    /// messaging timeout instead of collapsing that into the same `nil` as
+    /// "this element genuinely has no window id" — a caller deciding whether
+    /// a missing id proves a window does not exist needs to tell those two
+    /// apart.
+    static func readWindowID(for element: AXUIElement) -> (id: CGWindowID?, timedOut: Bool) {
         var id: CGWindowID = 0
-        guard _AXUIElementGetWindow(element, &id) == .success, id != 0 else { return nil }
-        return id
+        let error = _AXUIElementGetWindow(element, &id)
+        guard error == .success, id != 0 else { return (nil, error == .cannotComplete) }
+        return (id, false)
     }
 }
--- a/Sources/Vorssaint/Services/Switcher/SpaceHopSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SpaceHopSupport.swift
@@ -45,6 +45,19 @@ enum SpaceHopSupport {
         windowSpaces.contains { fullscreenSpaces.contains($0) }
     }
 
+    /// Whether an owner's "no window here" answer for a hidden-Space surface
+    /// can actually be trusted. `hasNoWindowsForOwner` being true only proves
+    /// this owner truly has nothing else when the pass that produced it was
+    /// otherwise reliable — a per-window Accessibility read that failed
+    /// partway through (the 0.35s messaging timeout can drop one real window
+    /// out of the results while its siblings still answer) means some
+    /// window's true answer was simply never obtained, not that it does not
+    /// exist, so an incomplete pass does not get to vouch for a window it is
+    /// missing.
+    static func ownerVouchIsIncomplete(hasNoWindowsForOwner: Bool, hadAttributeReadFailure: Bool) -> Bool {
+        hasNoWindowsForOwner || hadAttributeReadFailure
+    }
+
     /// The window server marks surfaces that must stay out of app window
     /// cycling. This remains meaningful when Accessibility cannot inspect a
     /// window because it lives on another Space.
--- a/Sources/Vorssaint/Services/Switcher/SpaceHopSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SpaceHopSupport.swift
@@ -45,16 +45,15 @@ enum SpaceHopSupport {
         windowSpaces.contains { fullscreenSpaces.contains($0) }
     }
 
-    /// Whether an owner's "no window here" answer for a hidden-Space surface
-    /// can actually be trusted. `hasNoWindowsForOwner` being true only proves
-    /// this owner truly has nothing else when the pass that produced it was
-    /// otherwise reliable — a per-window Accessibility read that failed
-    /// partway through (the 0.35s messaging timeout can drop one real window
-    /// out of the results while its siblings still answer) means some
-    /// window's true answer was simply never obtained, not that it does not
-    /// exist, so an incomplete pass does not get to vouch for a window it is
-    /// missing.
-    static func ownerVouchIsIncomplete(hasNoWindowsForOwner: Bool, hadAttributeReadFailure: Bool) -> Bool {
+    /// Whether the hidden-Space stale-surface veto should be skipped for this
+    /// owner: either it genuinely reported zero windows (the original #339
+    /// forgiveness case, and the most complete a vouch can be), or a
+    /// per-window Accessibility read failed partway through this pass (the
+    /// 0.35s messaging timeout can drop one real window out of the results
+    /// while its siblings still answer) — that window's true answer was
+    /// simply never obtained, not proven absent, so this incomplete pass
+    /// does not get to veto a window it is missing.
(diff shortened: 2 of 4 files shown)


PathGao:
Read the diff and the surrounding code on main. The direction is right, but the timeout signal doesn't cover the drop point that matters most.

**The window-id resolve is the biggest unflagged drop, and it's in the same loop.**

```swift
// AXWindowResolver.swift:17
guard _AXUIElementGetWindow(element, &id) == .success, id != 0 else { return nil }
```

Every `AXError` collapses to `nil`, and in the snapshot loop `if let id = AXWindowResolver.windowID(for: window)` then drops that window out of both `byID` and `ordered` with `hadAttributeReadFailure` still false. That is the exact scenario in your body: two windows from one owner, one resolves fine, the other doesn't answer — `ordered` is non-empty so the snapshot is returned, the vouch reads as complete, and the missing window is still vetoed. The bool/string readers you threaded run *before* this call; if contention lands here instead, the fix never fires.

I did not verify whether `_AXUIElementGetWindow` honors the per-element messaging timeout you set two lines earlier — that it swallows all `AXError` into `nil` is plain from the code, whether `.cannotComplete` is reachable there is not something I can check without a real build under load.

**Second uncovered reader, same class:** `accessibilityFrame(for:)` — both `AXUIElementCopyAttributeValue` calls sit behind one `guard ... == .success` that returns `nil`. In `isUserFacingWindow` a timeout there makes `fillsScreen` false, and `isSwitchableNonstandardWindow` can then reject a real `AXUnknown` playback surface, so the window never reaches the snapshot loop at all and no flag is set.

**The predicate reads backwards from what it does.**

```swift
ownerVouchIsIncomplete(hasNoWindowsForOwner: true, hadAttributeReadFailure: false) == true
```

An owner that cleanly answered "zero windows" is the most complete vouch there is, not an incomplete one. The doc comment says an incomplete pass "does not get to vouch", but the body is `a || b`, which only ever widens trust. What it actually computes is "skip the stale-surface veto". Name it that. A `||` behind an inverted name is how the next caller gets it wrong.

**One consequence the body doesn't state:** the flag is per-owner, not per-window, so a single timeout makes *every* unmatched hidden-Space surface of that owner witnessed, including the stale dialog the original comment at that site was written to catch. Fine trade — showing a ghost beats hiding a real window — but it's a behavior change and should be in the body, since the body frames this as pure recovery.

**Minor:** after this patch `stringAttribute` and `boolAttribute` in `WindowEnumerator` have no callers left in the file. Delete them; keeping both shapes is how a new call site picks the lossy one.

**Impact:** partial. A timeout on the window-id resolve or on the frame read still produces the original symptom — a real window vetoed off a hidden Space — with no visible difference from before this PR.
**Merge:** needs the `AXWindowResolver.windowID(for:)` drop point covered by the same flag first (or a stated reason it cannot time out).

ref #1061

naveenkrdy:
All addressed (be4d4c9):

**Window-id resolve drop point.** You're right, and it's the same class of gap the bool/string readers already had before this PR. Added `AXWindowResolver.readWindowID(for:)`, a timeout-aware sibling of `windowID(for:)` (kept as a thin wrapper so the other ~15 call sites across the tree are untouched), and the snapshot loop now sets `hadAttributeReadFailure` on a timed-out id resolve before it drops the window from `ordered`/`byID`.

**`accessibilityFrame(for:)` drop point.** Same fix, same shape: it now returns `(frame: CGRect?, timedOut: Bool)`, and both call sites (the snapshot loop, and `isUserFacingWindow`'s `fillsScreen` check) thread the flag through.

I didn't verify `_AXUIElementGetWindow`'s timeout behavior under load either — same caveat as before, the fix costs nothing whether or not `.cannotComplete` is actually reachable there.

**Predicate name.** Renamed `ownerVouchIsIncomplete` to `shouldSkipStaleSurfaceVeto`, matching your read exactly - it's a `||` that only ever widens trust toward skipping the veto, and a clean "zero windows" answer is the most complete vouch there is, not an incomplete one. Doc comment rewritten to say what it computes.

**Per-owner not per-window.** Correct, and worth stating explicitly rather than leaving it implied: this flag makes *every* unmatched hidden-Space surface from that owner witnessed once one window's read fails, not just the one that timed out. Showing an occasional stale surface beats hiding a real window, but it is a behavior change from pre-#1062, not pure recovery of what #1062 intended.

**Dead code.** Deleted `stringAttribute`/`boolAttribute` - no callers left in the file.

Verified with `./build.sh` (no warnings), `./build/Vorssaint --selftest` (`SELFTEST OK`), `./build.sh --test` (`TESTS OK`, 8806 checks). Added three structural guard tests (same source-grep pattern the existing `ownerVouchIsIncomplete` test used) confirming both new drop points route through the flag; ran red/green by hand - stripped the three new `hadAttributeReadFailure`/`hadReadFailure` assignments, reran, got exactly those three new assertions failing and nothing else, restored, confirmed green again.

ref #1061

PathGao:
Both drop points are covered now, and `shouldSkipStaleSurfaceVeto` reads the way it computes. The red/green on the three new assertions is the part I'd have asked for.

**`main` moved, and the rebase brings a fourth drop point.** `isUserFacingWindow` gained `normalLevelWindowIDs`, and its `hasNormalWindowLevel` calls `AXWindowResolver.windowID(for:)` bare. A timeout there reads as "not normal-level", so `isSwitchableNonstandardWindow` rejects a real borderless window and it never reaches the snapshot loop — no flag set. Same fix, `readWindowID`. (`isCancelled:` also landed on both functions you rewrote.)

**Two new AX stalls on the hover path.** `accessibilityFrame` now issues the size read even after position returned `.cannotComplete`, so 0.35s per window becomes 0.70s — against the rule stated at line 616 of the same file, "skip the remaining calls, each would block again."

```swift
guard positionResult == .success else { return (nil, positionResult == .cannotComplete) }
```

It also left the `&&`, so it runs for every non-standard subrole, not just `canBePlaybackSurface`. Two extra round-trips per dialog on every Dock hover, and a timeout on a frame you then discard still widens trust owner-wide. Move the read back inside the branch.

**Impact:** partial. A timed-out id resolve in `hasNormalWindowLevel` still vetoes a real borderless window with nothing flagged.
**Merge:** needs the `hasNormalWindowLevel` resolve covered, after the rebase.

ref #1061

vorssaint:
closing this one
the current version already asks the window server, window by window, whether a window that accessibility left out is still open, so a real window parked on another desktop stays listed even when a sibling on the current desktop answered
one slow read here would bring back every dismissed dialog of that app as a listed window, which is the ghost this set out to avoid, and it would happen most in the dock preview where slow reads are common
the frame read now runs for every dialog and still asks for the size after the position already timed out, so a slow app costs more time on each hover instead of less
the code this touches was rewritten in the meantime, so it cannot be brought up to date without redoing it
if a real window still goes missing on the current version, a fresh report with the app and desktop layout is the useful next step

PR #1096, opened 2026-08-28
fix(settings): keep machine-local paths out of settings backups

iltonandrew:
Refs #1009.

Follow-up to #1023, found by @PathGao reading merged `main` rather than the branch.

## What was wrong

Before #1023 a mouse exception list held only bundle identifiers, which name the same app on any Mac, so exporting those five keys whole was safe. Since #1023 the same arrays also hold the resolved path of a program that has no identifier to be named by, and a path is authority on one Mac only.

The five keys are registered defaults (`Defaults.swift`), none is in `machineStateKeys` (`SettingsBackupSupport.swift`), and `portableMediaSettings` only touches the media keys — so `exportKeys()` carries them verbatim. A backup written after someone excepts a bundled Java runtime therefore contains something like:

```
/Users/<name>/Library/Application Support/PrismLauncher/java/jre-legacy/zulu-8.jre/Contents/Home/bin/java
```

Two consequences, both quiet. The file now carries the person's short account name, and restoring it on another Mac leaves a row pointing at a file that is not there. `valueLooksRight` cannot drop it on the way back in either — it checks `[String]` shape, and `["com.apple.Safari", "/Users/…/java"]` is a perfectly good `[String]`.

## The change

`machineStateKeys` already refuses exactly this class, with the reasoning written out on `commandBarFileScopes`, `mediaImageWatermarkLogoPath` and `brightnessDDCWriteOnlyPaths`: authority on one Mac, not portable configuration. Bundle identifiers travel, resolved paths do not.

It cannot be a `machineStateKeys` entry here, because the two kinds now share one array — dropping the key would throw away the portable half with it. So the refusal is per value instead, using `MouseAppExceptionSupport.isExecutablePathIdentity`, the predicate that already decides this question at every other site.

Two things worth naming about the shape:

- It is driven from `MouseExceptionScope.allCases` rather than a list of keys spelled out in the backup file. A scope that renames its key, or two scopes that come to share one, stay covered without this file being edited. That is not hypothetical — #1040 merges `smoothScrollExceptions` and `scrollInverterExceptions` into one new `mouseScrollingExceptions` key, and this filter follows that on its own.
- It runs on the way out **and** on the way back in, the same shape `portableMediaSettings` already uses, so a hand-edited file cannot put a path back either.

An emptied list keeps its key rather than being removed: an empty exception list still means "no exceptions".

## Verification

Three behaviour checks, not source-shape ones — `SettingsBackupSupport` is compiled into the test binary, so the real functions are called:

- every scope's list exports its bundle ids and drops the machine-local path
- a hand-edited backup cannot restore a path into any of them
- a list of only bundle ids survives a backup **unchanged**

That third one is the complement, and it is the one that matters most here: a filter that drops too much is invisible in the file and only shows up as someone's exception lists coming back empty. Mutation-proved both ways rather than asserted:

| Mutation | Result |
| --- | --- |
| remove the export-side call | 1 → 2 failures; the export check fires, the other two correctly do not |
| make the filter drop everything | 1 → 4 failures; all three fire, including the bundle-ids-survive complement |

Gate: `./build.sh` 0 warnings. `./build.sh --test` — 8821 checks, **1 failure**: the `dispatch pool starvation` precondition #1014 exists to skip, red on this 15-core M5 Pro and green on CI's smaller runners, and identical on unrelated branches gated in the same sitting. Reporting it rather than writing `TESTS OK`. `./build/Vorssaint --selftest` — `SELFTEST OK`.

## Environment

MacBook Pro (Mac17,9, Apple M5 Pro), macOS 26.6.2 (25G83), built-in Liquid Retina XDR, 3024x1964 native / 1512x982 points.

Not verified: I have not written and re-imported a backup file through the Settings UI by hand. The reasoning and the tests both go through the real `payload` and `sanitizedSettings` functions, but nobody has watched it happen in the app.

### DIFF
--- a/Sources/Vorssaint/Core/SettingsBackupSupport.swift
+++ b/Sources/Vorssaint/Core/SettingsBackupSupport.swift
@@ -136,6 +136,7 @@ enum SettingsBackupSupport {
             }
         }
         settings = portableMediaSettings(settings)
+        settings = portableMouseExceptions(settings)
         return [
             formatVersionKey: formatVersion,
             appVersionKey: appVersion,
@@ -153,7 +154,7 @@ enum SettingsBackupSupport {
         else { return nil }
         let allowed = exportKeys()
         let filtered = settings.filter { allowed.contains($0.key) && valueLooksRight($0.key, $0.value) }
-        return portableMediaSettings(filtered)
+        return portableMouseExceptions(portableMediaSettings(filtered))
     }
 
     static func formatVersion(from payload: [String: Any]) -> Int? {
@@ -168,6 +169,36 @@ enum SettingsBackupSupport {
             return intValue
         }
         return nil
+    }
+
+    /// A mouse exception list holds two kinds of identity at once: a bundle
+    /// identifier, which names the same app on any Mac, and the resolved path
+    /// of a program that has no identifier to be named by (issue #1009). The
+    /// never-exported list above already refuses the second kind wherever it
+    /// has a key to itself -- `commandBarFileScopes`,
+    /// `mediaImageWatermarkLogoPath`, `brightnessDDCWriteOnlyPaths`, all for
+    /// the same stated reason: authority on one Mac, not portable
+    /// configuration. Here the two kinds share one array, so the refusal has
+    /// to be per value rather than per key. Left in, a backup would carry the
+    /// short username inside the path, and restoring it on another Mac would
+    /// leave a row pointing at a file that is not there -- which
+    /// `valueLooksRight` cannot catch, since the value is a perfectly good
+    /// `[String]`.
+    ///
+    /// Driven from `MouseExceptionScope.allCases` rather than a list of keys
+    /// spelled here, so a scope that renames its key, or two scopes that come
+    /// to share one, are covered without this file being edited.
+    private static func portableMouseExceptions(_ source: [String: Any]) -> [String: Any] {
+        var settings = source
+        for key in Set(MouseExceptionScope.allCases.map(\.defaultsKey)) {
+            guard let stored = settings[key] as? [String] else { continue }
+            let portable = stored.filter { !MouseAppExceptionSupport.isExecutablePathIdentity($0) }
+            guard portable.count != stored.count else { continue }
+            // An emptied list still means "no exceptions", so the key stays
+            // rather than falling back to whatever a missing key would do.
+            settings[key] = portable
+        }
+        return settings
     }
 
     private static func portableMediaSettings(_ source: [String: Any]) -> [String: Any] {
--- a/Sources/Vorssaint/Core/SettingsBackupSupport.swift
+++ b/Sources/Vorssaint/Core/SettingsBackupSupport.swift
@@ -188,6 +188,23 @@ enum SettingsBackupSupport {
     /// Driven from `MouseExceptionScope.allCases` rather than a list of keys
     /// spelled here, so a scope that renames its key, or two scopes that come
     /// to share one, are covered without this file being edited.
+    /// The half of an exception list a backup file never carries, because
+    /// `portableMouseExceptions` filters it out on the way out.
+    static func pathIdentities(in list: [String]) -> [String] {
+        list.filter(MouseAppExceptionSupport.isExecutablePathIdentity)
+    }
+
+    /// What an exception list should hold once a backup has been applied.
+    /// Restoring clears every exported key before writing the file's values,
+    /// and the file only has the portable half -- so without carrying the
+    /// paths across, applying a backup would delete the entries for programs
+    /// that are not apps, including on the Mac the backup was written on. The
+    /// restored order wins and a carried path already present is not doubled,
+    /// so applying the same backup twice lands on the same list.
+    static func restoredExceptionList(restored: [String], carried: [String]) -> [String] {
+        restored + carried.filter { !restored.contains($0) }
+    }
+
     private static func portableMouseExceptions(_ source: [String: Any]) -> [String: Any] {
         var settings = source
         for key in Set(MouseExceptionScope.allCases.map(\.defaultsKey)) {
--- a/Sources/Vorssaint/Core/SettingsBackupSupport.swift
+++ b/Sources/Vorssaint/Core/SettingsBackupSupport.swift
@@ -171,6 +171,27 @@ enum SettingsBackupSupport {
         return nil
     }
 
+    /// The half of an exception list a backup file never carries, because
+    /// `portableMouseExceptions` filters it out on the way out.
+    static func pathIdentities(in list: [String]) -> [String] {
+        list.filter(MouseAppExceptionSupport.isExecutablePathIdentity)
+    }
+
+    /// What an exception list should hold once a backup has been applied.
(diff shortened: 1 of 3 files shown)


iltonandrew:
@PathGao — you were right, and it was worse than the framing you gave it. Fixed in `be967a6`, answering here rather than on #1009 so @joshrad-dev's thread stays about his bug.

**Confirmed from the code, not your word.** `SettingsBackup.swift:78-83` is `removeObject(forKey:)` across all of `exportKeys()`, then `set` for each of the file's values. The five keys are registered defaults and are not in `machineStateKeys`, so they are in that loop.

The part I had not seen: this made my own change a regression. Before the filter, restoring a backup written on the same Mac put the paths back, because the file still carried them. After it, the clear deletes them and the file has nothing to write back — so @joshrad-dev's `java` entry disappears on any restore, on his own Mac, with nothing said. I shipped a data-loss path while closing a privacy leak.

Your distinction is the thing I had wrong. A `machineStateKeys` entry gets "never travels" **and** "never restored" from one line, because it is neither cleared nor written. A per-value export filter only buys the first, and I treated the two as the same guarantee.

**The fix is your shape.** Path identities read before the clear, merged back after the file's values, driven from `MouseExceptionScope.allCases` for the same reason the filter is. Applying the same backup twice does not double an entry, and the restored order wins.

Both halves are pure functions in `SettingsBackupSupport` — `pathIdentities(in:)` and `restoredExceptionList(restored:carried:)` — because `SettingsBackup.swift` is not in `build.sh`'s test source list, and I would rather assert the decisions than describe the service as text.

**On testing it.** You are right that neither of my earlier mutation directions could reach this: they are on `payload` and `sanitizedSettings`, and the clear-then-write lives in the service where neither pure function can see it. The load-bearing property here is an **order**, not a presence, so the pin is an anchored scan asserting the read precedes the clear and the write-back follows the file's values. Mutation-proved by moving the capture below the clear with every line still present: 1 → 2 failures, naming *a settings restore reads the machine-local entries before clearing and writes them back after the file's values*. A presence check would have stayed green through that.

Worth saying plainly: my own PR body flagged "not verified: nobody has written and re-imported a backup by hand" — and that is exactly the check that would have caught this. Naming a gap is not the same as closing it, and I left it named.

**Gate:** `./build.sh` 0 warnings. `--test` 8825 checks, **1 failure** — the `dispatch pool starvation` precondition #1014 exists to skip, red on this 15-core M5 Pro and green on CI. `SELFTEST OK`.

Still not verified, unchanged and now more pointed: nobody has exported and re-imported a real backup through Settings. The round trip is asserted through the real functions on the export side and through the pure halves on the restore side, but the `UserDefaults.standard` loop in `applyAndRelaunch` itself is pinned by shape, not exercised.

ref #1009

PathGao:
Checked `be967a6` against `main`, not just the diff.

**The sibling sweep.** `AppBundleList` has four call sites — window preview exclusions, clipboard ignored apps, switcher app rules, mouse exceptions — and only `MouseExceptionsList.swift:25` passes `acceptsExecutables: true`. The picker drops a path for the other three, so `MouseExceptionScope.allCases` really is the complete set of exported keys that can hold one. And the backup file is the only way out: nothing in `Sources/` reads our own domain wholesale — the two `dictionaryRepresentation()` / `persistentDomain` sites are `com.apple.dock` and a foreign domain in `Defaults.swift:1581`. Coverage is closed, not just plausible.

**Your `build.sh` claim holds.** `Core/SettingsBackupSupport.swift` is on the test source list at `build.sh:188`; `Services/SettingsBackup.swift` is not. And reading a source file by relative path is already 35 sites deep in `MetricsTests.swift`, so the anchored scan adds no new working-directory dependency — it fails loudly rather than passing vacuously if the read misses, since every `firstIndex` goes nil.

**The old-file case works, for a reason the comment does not give.** All five keys register `[String]()` (`Defaults.swift:800-804`), so after the clear `defaults.stringArray(forKey:)` reads `[]`, not the pre-restore list. A backup written before these keys existed lands on `[] + carried` instead of doubling — the carry-back is safe on a file that omits the key, not only on one that carries an empty array.

**One thing to fix.** The doc block above `pathIdentities(in:)` is two comments run together. Everything from "A mouse exception list holds two kinds of identity" through "without this file being edited" describes `portableMouseExceptions`, which now has no doc at all; only the last two lines are about `pathIdentities`. As it stands the comment tells the next reader that `pathIdentities` is driven from `MouseExceptionScope.allCases`, and it takes a `[String]`. Move the block down to the private function.

Not verified: I have not built or run this, and the round trip through Settings is still unexercised — unchanged from what you said. Both checks read as blank state in what I can see, so I am not calling CI green and not adding `ready to merge`.

**Impact:** anyone who has excepted a non-app executable. Exporting used to put their short username in the file; before `be967a6`, restoring deleted the entry on their own Mac. Both are closed. Nobody without a path entry sees any change.
**Merge:** ship it once the two checks report green. The doc move is not worth a re-review.

ref #1009

iltonandrew:
Doc block moved in `3a20dfc`, and your reason for the old-file case is now in the comment rather than only in this thread.

**The cause was mechanical and worth naming.** I inserted the two restore helpers anchored on `private static func portableMouseExceptions`, which is the signature line — so they landed between that function and the comment written for it. The block describing the per-value export filter ended up above `pathIdentities`, telling the next reader that a function taking a `[String]` is driven from `MouseExceptionScope.allCases`, while the function it was written for had nothing. Anchoring an insertion on a signature silently steals whatever doc sits above it.

Each of the three now carries its own.

**Your third point closed a gap in my comment, so I took the reasoning too.** I had written that a carried path already present is not doubled, which is true but not the interesting half. The reason the omitted-key case is safe is that all five register `[String]()`, so after the clear the read is `[]` rather than the pre-restore list — a backup written before these keys existed lands on `[] + carried`. That is now in the doc for `restoredExceptionList`, credited.

**On your sweep.** The `AppBundleList` call-site check is the half I could not have proved from my own diff, and it is the one that makes `MouseExceptionScope.allCases` a complete set rather than a plausible one. Same for the `dictionaryRepresentation()` / `persistentDomain` sites being foreign domains — I had reasoned that the backup file is the only way out, but I had not enumerated the ways out. Recording that here so the next reader gets the closure and not just the claim.

**Gate after the move:** 0 warnings, 8825 checks with the one `dispatch pool starvation` failure #1014 exists to skip, `SELFTEST OK`. Doc-only, so unchanged from `be967a6` — I rebuilt rather than assuming a comment cannot break a build.

Agreed on the checks: both read as blank state from here too, and I am not calling them green either. Nothing else outstanding on my side.

ref #1009

PathGao:
Doc move is right — three functions, three docs, each about the one under it. Nothing else in the diff moved.

Both checks read green now, which was the only thing I was holding on, so `ready to merge` goes on. Unchanged from last round: I have not built this, and nobody has exported and re-imported a backup through Settings.

**Impact:** anyone who has excepted a non-app executable. Their export stops carrying their short account name, and restoring on another Mac no longer plants a row pointing at a file that is not there. On their own Mac the entry survives a restore, same as today. Nobody without a path entry sees a change.
**Merge:** ship it.

ref #1009

PR #1097, opened 2026-08-28
chore(i18n): Unify Simplified Chinese terminology

PathGao:
## Summary

Simplified Chinese carried two words for several actions, so the same action
read differently depending on which feature produced it. Copying to the
clipboard said 已复制 in the clipboard history, 已拷贝 in the scratchpad and
截屏已拷贝 after a capture; saving said 已保存 in the image converter and
已存储 after a recording. 应用 stood for both an app and the verb Apply, next
to each other in the same file — `applyCurve` reads 应用风扇曲线 while
`safetyCaption` says 应用断开, which is the app disconnecting.

This settles one word per action, following what macOS itself shows:

- **Copy is 拷贝**, the word in Finder's own Edit menu for ⌘C. That leaves 复制
  for Duplicate, exactly as the system menu spends it (`Duplicate Items = 复制项目`
  in Finder's `zh_CN.lproj`). `duplicateProfileButton` already read 复制配置 and
  is now the only string with that word.
- **Save splits by what it produces**: 存储 when a file lands on disk, 保存 when a
  setting or a history entry is kept. Screenshots and recordings already used
  存储; the image converter, the OCR TXT export and the backup export said 保存
  while writing real files, and the radial-menu, snippet and scratchpad Save
  buttons said 存储 while writing only settings.
- **An app is App**, leaving 应用 to the Applications folder alone, and Apply
  becomes 启用 for a mode you can turn back off, 使用套装 for a feature bundle,
  生效 for a change taking hold and 确定 for a typed value. The bundle button
  stays distinct from `installButton`, which installs a single feature — the
  bundle also uninstalls everything else, so it should not read 安装 on the
  same page.
- **An app name carried into a sentence or a row is quoted.** Chinese has no
  word spacing, so 要退出Safari吗 and 重新启动Safari ran together where the
  process list already read 要结束“Safari”吗. All four command-bar app-name
  formats and the Homebrew package prompts now quote.
- **The seven prompts macOS shows inside its own authorization dialog address
  the reader as 您** throughout. One of them already did and the others did not,
  which read as a different speaker at the moment the system asks for a
  password. `adminPromptClamshellOn` and `adminPromptClamshellOff` are the same
  feature on and off, in the same dialog.

Nothing outside `zhHans` values changed: no English, no other language, no
behaviour, no test. 97 strings across 17 files.

## Verification

MacBook Pro (Mac17,3, Apple M5), macOS 26.6.2 build 25G83, release build from
`./build.sh` with the legacy self-signed identity.

```sh
./build.sh --test              # TESTS OK (8818 checks)
```

The strings were changed by a script scoped to `zhHans` initialisers rather
than by search-and-replace, because several field names repeat across catalogs
with the same value — `saveButton` is 存储 in four structs and only two of them
should become 保存. Every value was then re-extracted from the tree and compared
against the intended result: 97 of 97 match, no unintended value moved, and no
English or other-language string changed.

**What I could not test:** I did not walk the app in Simplified Chinese and look
at these strings in place, so wrapping and truncation in narrow controls are
unverified. The growth is small — 存储 → 存储空间 and 安装 → 使用套装 are the
only strings that got longer by more than nothing, and 应用 → App is shorter or
equal — but the bundle button sits on a narrow card and is the one worth a look
on screen. The macOS authorization wording was read from
`AdminShell.appleScriptSource`, not observed on screen; I did not trigger the
seven admin prompts. Traditional Chinese and the other eleven languages are
untouched and unexamined.

Nothing here is covered by a new test. The existing localization checks still
apply — every string non-empty, no em-dash, positional specifiers preserved —
but they do not know about terminology, so a later string can reintroduce 复制
for Copy without turning anything red.

## Anything else

No issue covers this, so there is no `Refs` line — say the word if one should
be opened first and I will move the discussion there.

The word choices are deliberate and each has an alternative worth naming, in
case any of them is wrong for this project:

- **Copy → 拷贝** follows macOS. The alternative, 复制, is what most non-Apple
  Chinese software uses and reads more naturally to someone new to the Mac —
  but it is the word the system reserves for Duplicate, so taking it would have
  forced an invented term for `duplicateProfileButton`.
- **Save → 存储/保存** splits one English word into two Chinese ones. Keeping a
  single word would be simpler but would say 存储 for a clipboard history entry.
- **应用 → App** leaves English in the middle of Chinese sentences. That already
  matched the majority of the catalog (133 App against 40 应用 before this).

CHANGELOG is untouched.


PathGao:
Three gaps, then the guard's reach.

**1. `adminPromptClamshellOn` still says 你.** The body says the six authorization prompts now use 您 throughout. Five moved, the first one didn't.

```swift
// Strings+ChineseSimplified.swift:96
adminPromptClamshellOn:  "Vorssaint 需要你的密码才能在合盖后继续运行。",   // unchanged
adminPromptClamshellOff: "Vorssaint 需要您的密码才能恢复 Mac 的正常睡眠。", // this PR
```

Same feature, on and off, same system dialog, now two speakers. The new guard asserts on `adminPromptClamshellOff` and `adminPromptRecover` and skips `ClamshellOn`, so nothing turns red. One character, plus one term in the expect.

**2. Two of the four app-name formats in `CommandBarFeatureStrings.zhHans` stayed unquoted.**

```swift
restartAppFormat:   "重新启动%@",      // 1838, unchanged
forceQuitAppFormat: "强制退出“%@”…",   // 1839, this PR
uninstallAppFormat: "卸载%@…",         // 1841, unchanged
```

`forceQuitAppFormat` isn't a confirmation either, so the rule as applied is wider than "a confirmation carrying an app name" — and under the wider rule these two are the same shape and got missed. The guard only checks the two `…ConfirmFormat` strings.

**3. `presetApplyButton` and `installButton` now render the same word in the same view.** `FeatureHubStrings.swift:1336` `installButton: "安装"` installs one feature; `:1351` `presetApplyButton` becomes 安装 and installs a bundle while uninstalling everything else. One word, two actions, one page — the inverse of what this PR is fixing. The confirm sheet does spell out the uninstall, so nothing goes silent, but 应用 was carrying that distinction. 使用套装 keeps it without bringing 应用 back.

**Guard reach:** the mirror list is 17 catalogs; `static let zhHans` appears 42 times under `Sources/`. `SwitcherAppRulesStrings`, `MouseExceptionStrings`, `WindowPreviewExclusionStrings`, `AppUpdateStrings` and 21 others are app-facing and are where 应用 comes back first. "spends 复制 on Duplicate alone" reads repo-wide and holds over 40% of the catalogs.

The sweep itself checks out. I grepped `main` for 复制 and 应用: every zhHans occurrence outside 复制配置 and 应用程序 is in this diff, and the 存储/保存 split lands where the body says it does.

**What I didn't verify:** I didn't build or run the tests. The guard's assertions were checked by reading the values in `main`, not by compiling. No on-screen check of wrapping or truncation.

**Impact:** none on behaviour. A zhHans reader gets 你 in one authorization prompt and 您 in its sibling, bare app names in two command-bar rows, and one label doing double duty on the Features page.
**Merge:** needs `adminPromptClamshellOn` moved to 您, and covered by the guard, first.

PathGao:
All three fixed, and the guard is gone rather than widened.

**1. `adminPromptClamshellOn`** now reads 您, so the same feature on and off
speaks with one voice in the same dialog. Seven prompts, not six — the body
said six because I had built that set from the `AdminShell.run(prompt:)` call
sites and `ClamshellOn` has none, which was the wrong boundary to draw for a
string a reader can still meet.

**2. `restartAppFormat` and `uninstallAppFormat`** now quote. You are right that
the applied rule was already wider than the one the body stated: `quitFormat`
and `forceQuitAppFormat` are rows, not confirmations, so the rule is "an app
name carried into a sentence or a row", and under that rule these two were the
same shape. All four app-name formats in `CommandBarFeatureStrings` match now.

**3. `presetApplyButton` and `presetConfirmApply`** are 使用套装, your wording.
安装 next to `installButton` was the exact fault this change exists to remove —
one word, two actions, one page — and it was mine, introduced here rather than
inherited.

**On the guard:** removed, not extended. Your reach note sent me to measure what
this repo actually guards, and the answer decided it. The mechanism is house
style — `Mirror(reflecting:)` appears 41 times in `MetricsTests` on `main`, and
`!$0.contains("—")` runs 39 times across 24 catalogs in every language. But
every content rule there is language-agnostic and structural: non-empty, no
em-dash, positional specifiers kept. The per-language assertions that exist are
four value-equality checks and the `featureTitles` table — "this key equals this
string", never "this word must not appear". A Simplified Chinese terminology
rule would have been the first of its kind in the suite, and putting one
language's style opinions into the shared test target is not something this
project asked for. So this is now a strings-only change with nothing new that
can turn red, and the body says so plainly: a later string can reintroduce 复制
for Copy without failing anything.

Rebuilt after all three: 0 warnings, `SELFTEST OK`, `TESTS OK (8818 checks)` —
seven fewer than the run in the previous body, which is exactly the seven
assertions removed.

Still unverified, unchanged from before: I have not looked at these strings on
screen in Simplified Chinese. 使用套装 is four characters on a narrow preset
card and is the one most likely to wrap.

Issue #853, opened 2026-08-23
Add volume slider for microphone


GolamRafi27oo:
### Problem or Motivation

The microphone currently does not provide a volume control, making it difficult for users to adjust the input level. This can result in audio that is too quiet, too loud, or distorted, depending on the microphone and environment.

<img width="305" height="224" alt="Image" src="https://github.com/user-attachments/assets/3d0a8dca-0dad-448f-aae3-852feed5fa94" />

### Proposed Solution

Add a volume slider to the microphone settings so users can adjust the microphone input level in real time. The slider should clearly display the current volume level, provide immediate feedback, and save the selected setting for future sessions.

### Feature Area

Sound

### Mockups / References

Reference image

<img width="354" height="101" alt="Image" src="https://github.com/user-attachments/assets/e78f2d45-b741-4504-b24f-1df72736530b" />

You can check out this project for better understanding: https://github.com/GolamRafi27oo/AudioControlBar

PR #1100, opened 2026-08-28
feat(mixer): add microphone volume control

GolamRafi27oo:
## Summary

Adds a hardware input-volume control beneath the microphone selector in the mixer panel.

The slider follows the existing output-volume control, shows an editable percentage, and appears only when the effective microphone exposes writable CoreAudio gain controls. It supports virtual-main, main scalar, and channel-only input gain layouts. For channel-only devices, it displays the mean channel gain and updates every writable channel.

Per-device CoreAudio volume changes are handled by a dedicated, coalesced refresh path that reads only the active input device. A generation guard prevents stale hardware reads from moving the slider backward during rapid adjustments. Read-only gain properties are not observed or presented as editable controls.

Volume writes prefer the virtual-main or main scalar control and stop after the first successful master write, preserving the device’s existing channel balance. If no writable master control is available, the implementation falls back to input channels 1 and 2 and updates every writable channel.

This PR also gives the output, system sounds, and microphone icons a consistent fixed-width, leading-aligned frame. This prevents their different intrinsic widths from causing adjacent labels to start at inconsistent positions. The output and input slider-row icons also use matching size and alignment.

The implementation uses the existing CoreAudio and mixer infrastructure, adds no dependencies or background subsystem, and removes or migrates device listeners whenever the selected microphone changes or the mixer stops.

## Verification

Tested on:
  - MacBook Air (Mac17,3), Apple M5, 16 GB memory
  - macOS 26.6.2 (25G83)
  - Locally built, ad-hoc-signed release installed in /Applications
  - Confirmed that the app bundle builds, installs, and launches
  - Confirmed that input-volume adjustments remain synchronized with CoreAudio
  - Confirmed consistent icon and label alignment across the mixer rows

Commands run:
  swift build
  git diff --check
  ./build.sh --install
  ./build.sh --test

The test harness passed all 8,818 checks.

./build/Vorssaint --selftest was also run but could not complete successfully in the restricted development environment. It reported AppleSMC and mounted-volume access warnings and failed its swap-memory and uptime readings.

I could not test multiple external microphone models or independently verify every driver-specific CoreAudio gain layout, including channel-only gain controls.

### DIFF
--- a/Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
+++ b/Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
@@ -329,9 +329,12 @@ final class AudioInputDeviceManager: ObservableObject {
         }
         let effectiveDeviceID = snapshot.resolution.effectiveUID
             .flatMap { uid in snapshot.devices.first(where: { $0.uid == uid })?.audioObjectID }
+        // Rewiring listeners invalidates pending control reads. Capture the
+        // sweep's validity first so that initial wiring cannot invalidate the
+        // volume value discovered by this same sweep.
+        let sweepVolumeIsCurrent = volumeRefreshGeneration == snapshot.volumeGeneration
         updateVolumeListeners(for: effectiveDeviceID)
-        if volumeRefreshGeneration == snapshot.volumeGeneration,
-           inputVolume != snapshot.inputVolume {
+        if sweepVolumeIsCurrent, inputVolume != snapshot.inputVolume {
             inputVolume = snapshot.inputVolume
         }
 
--- a/Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
+++ b/Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
@@ -23,6 +23,7 @@ final class AudioInputDeviceManager: ObservableObject {
     @Published private(set) var preferredInputDeviceUID: String?
     @Published private(set) var currentInputDeviceUID: String?
     @Published private(set) var effectiveInputDeviceUID: String?
+    @Published private(set) var inputVolume: Double?
     @Published private(set) var preferredUnavailable = false
     @Published private(set) var lastError: String?
 
@@ -30,6 +31,8 @@ final class AudioInputDeviceManager: ObservableObject {
     /// Stored so stop() can remove the HAL listeners when the mixer leaves
     /// the hub.
     private var globalListeners: [AudioObjectPropertySelector] = []
+    private var volumeListenerDeviceID: AudioObjectID?
+    private var volumeListenerAddresses: [AudioObjectPropertyAddress] = []
     private var applyingPreferred = false
     private var refreshPending = false
     private var lastListenerRefreshAt: CFAbsoluteTime = 0
@@ -79,6 +82,7 @@ final class AudioInputDeviceManager: ObservableObject {
         // A sweep already reading the HAL must not publish into a manager that
         // has stopped watching.
         refresh.discardInFlight()
+        removeVolumeListeners()
         for selector in globalListeners {
             var address = AudioObjectPropertyAddress(mSelector: selector,
                                                      mScope: kAudioObjectPropertyScopeGlobal,
@@ -92,6 +96,7 @@ final class AudioInputDeviceManager: ObservableObject {
         if !inputDevices.isEmpty { inputDevices = [] }
         if preferredUnavailable { preferredUnavailable = false }
         if lastError != nil { lastError = nil }
+        if inputVolume != nil { inputVolume = nil }
     }
 
     func setPreferredInputDeviceUID(_ uid: String?) {
@@ -107,6 +112,23 @@ final class AudioInputDeviceManager: ObservableObject {
         // previous preference and would publish it back for an instant.
         refresh.discardInFlight()
         refreshAndApply()
+    }
+
+    func setInputVolume(_ volume: Double) {
+        guard let uid = effectiveInputDeviceUID,
+              let device = inputDevices.first(where: { $0.uid == uid }) else { return }
+        let clamped = min(max(volume, 0), 1)
+        halQueue.async { [weak self] in
+            let applied = Self.setInputVolume(Float32(clamped), for: device.audioObjectID)
+            DispatchQueue.main.async {
+                guard let self, self.effectiveInputDeviceUID == uid else { return }
+                if applied {
+                    self.inputVolume = clamped
+                } else {
+                    self.refreshAndApply()
+                }
+            }
+        }
     }
 
     /// The smallest possible answer to a change: the system decides which
@@ -176,6 +198,7 @@ final class AudioInputDeviceManager: ObservableObject {
         let currentUID: String?
         let devices: [MixerInputDevice]
         let resolution: MixerInputRouteResolution
+        let inputVolume: Double?
         /// Non-nil when this sweep actually pointed the system input at the
         /// preferred device; the write has already happened on the HAL.
         let applied: AppliedInput?
@@ -224,10 +247,14 @@ final class AudioInputDeviceManager: ObservableObject {
               request.mayApplyPreferred,
               let savedUID,
               let device = devices.first(where: { $0.uid == savedUID }) else {
+            let volume = resolution.effectiveUID
+                .flatMap { uid in devices.first(where: { $0.uid == uid }) }
+                .flatMap { inputVolume(for: $0.audioObjectID) }
             return RefreshSnapshot(savedUID: savedUID,
                                    currentUID: currentUID,
                                    devices: devices,
                                    resolution: resolution,
+                                   inputVolume: volume.map(Double.init),
                                    applied: nil)
         }
 
@@ -239,6 +266,7 @@ final class AudioInputDeviceManager: ObservableObject {
                                currentUID: currentUID,
                                devices: devices,
                                resolution: resolution,
+                               inputVolume: inputVolume(for: device.audioObjectID).map(Double.init),
                                applied: AppliedInput(device: device,
                                                      status: status,
                                                      deviceBeforeOverride: before))
@@ -281,6 +309,12 @@ final class AudioInputDeviceManager: ObservableObject {
         }
         if inputDevices != snapshot.devices {
             inputDevices = snapshot.devices
+        }
+        let effectiveDeviceID = snapshot.resolution.effectiveUID
+            .flatMap { uid in snapshot.devices.first(where: { $0.uid == uid })?.audioObjectID }
+        updateVolumeListeners(for: effectiveDeviceID)
+        if inputVolume != snapshot.inputVolume {
+            inputVolume = snapshot.inputVolume
         }
 
         guard let applied = snapshot.applied else { return }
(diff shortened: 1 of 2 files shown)


PathGao:
Two of these are places where the output side on `main` already solved the same problem a different way.

**1. The volume listeners are wired to the global refresh callback.**

`updateVolumeListeners` registers `Self.listenerCallback`, so every mic gain change lands in `scheduleListenerRefresh()` → `refreshAndApply()` → `readSnapshot`, which enumerates every audio device, reads ~7 properties on each, and can re-issue `setDefaultInputDevice`. A full device sweep per volume tick, throttled to 200ms.

`AppVolumeMixer` handles output with a dedicated callback for exactly this reason — `outputControlListenerCallback` (`AppVolumeMixer.swift:269`) → `scheduleOutputControlRefresh(for:)` (`AppVolumeMixer.swift:345`), which reads two properties on one device, coalesces the burst at 30ms, and carries a generation guard. Its comment names the failure you'll now get on input:

> A drag and the volume keys can emit several properties for each step. Read once after the burst so an older callback cannot pull the slider back while a newer value is already on screen.

`setInputVolume` publishes `clamped` optimistically, then the 200ms sweep publishes whatever the HAL read. Mid-drag that read is stale and pulls the slider backwards. Copy the `scheduleOutputControlRefresh` shape instead of reusing the global callback.

Same function, smaller: registration gates on `AudioObjectHasProperty` while `inputVolume(for:)` also requires settable, so a device with read-only input gain gets a listener and drives the sweep with no slider on screen.

**2. `setInputVolume` writes every settable element instead of stopping at the first that takes.**

`setOutputVolume` (`AppVolumeMixer.swift:1649`) returns `true` on the first successful write. The new one keeps going through Main, 1, 2. On a device where master and per-channel are both settable that's a master write followed by two per-channel writes at the same scalar, which flattens whatever L/R balance the driver had. `return true` after the first success.

While you're there: `outputVolumeSelectors` tries `kAudioHardwareServiceDeviceProperty_VirtualMainVolume` before `VolumeScalar`, which is what covers per-channel-only devices without hardcoding elements 1 and 2, and it accepts `kAudioDevicePropertyScopeInput`. Unless that was already ruled out for input, it's fewer moving parts than the three-element loop.

**3. Undeclared UI change.** The body says the PR adds a control beneath the microphone selector. The diff also puts `.frame(width: 16, alignment: .leading)` on the output, alert, and mic *header* icons — three rows the body doesn't mention. It's also half a sweep: the existing output slider-row icon at `MixerSection.swift:167` still has plain `.frame(width: 16)`, centered, so the new input slider icon won't line up with the output one above it.

No hardware here, so nothing about real driver gain layouts is tested — all of the above is read off `main` and the diff.

**Impact:** any microphone gain change, from this app or anything else, triggers a full CoreAudio device enumeration; the slider can jump backwards mid-drag.
**Merge:** needs the volume listener moved off the global refresh callback first.

ref #853

GolamRafi27oo:
@PathGao The header icon changes are intentional. Without `.frame(width: 16, alignment: .leading)`, the icons’ different intrinsic widths cause the header text to start at inconsistent positions, as shown in the screenshot. I’ll update the PR body to document this alignment fix and apply the same leading alignment to the existing output slider-row icon so both slider rows remain consistent.

Screenshot 2026-08-28 at 5 12 31 PM

GolamRafi27oo:
@PathGao Thanks for the detailed review, both issues are fixed.

Input-volume changes now use a dedicated 30 ms coalesced refresh with a generation guard, reading only the active input device instead of triggering a full device sweep. Listener registration also uses the same settable check as the slider.

inputSetVolume now tries virtual-main volume before scalar volume and returns after the first successful write, avoiding unnecessary per-channel writes and preserving channel balance.

PathGao:
Both are addressed. The new coalesced refresh has one ordering bug.

**`updateVolumeListeners` runs before the generation check, and it bumps the generation.**

In `apply`:

```swift
updateVolumeListeners(for: effectiveDeviceID)
if volumeRefreshGeneration == snapshot.volumeGeneration,
   inputVolume != snapshot.inputVolume {
    inputVolume = snapshot.inputVolume
}
```

`updateVolumeListeners` calls `removeVolumeListeners()` whenever `volumeListenerDeviceID != deviceID`, and that starts with `volumeRefreshGeneration &+= 1`. So any sweep that wires listeners invalidates its own read.

Two ways it shows:

- **First sweep after `start()`.** `volumeListenerDeviceID` is nil, so it always rewires, so the sweep that discovered the device never publishes its volume. `start()` on an already-installed listener is a single `refreshAndApply()` (`AudioInputDeviceManager.swift:64`) — nothing schedules a second one. The slider is absent until some unrelated HAL notification happens to fire another sweep.
- **Device with no settable input gain.** Registration appends nothing, so `volumeListenerDeviceID` stays nil and *every* sweep rewires and bumps. Switch from a mic with gain to one without and `inputVolume` never gets overwritten with the nil — the old device's slider stays on screen for good. Dragging it does nothing either: `Self.setInputVolume` returns false, and the `scheduleVolumeRefresh` fallback bails on `volumeListenerDeviceID == deviceID` with nil on the left.

Hoist the comparison:

```swift
let sweepIsCurrent = volumeRefreshGeneration == snapshot.volumeGeneration
updateVolumeListeners(for: effectiveDeviceID)
if sweepIsCurrent, inputVolume != snapshot.inputVolume {
    inputVolume = snapshot.inputVolume
}
```

The output side dodges this by not guarding the sweep publish at all (`AppVolumeMixer.swift:853`) — `subscribeToOutputControls` bumps its own generation too, but nothing on that path reads it.

**The output slider-row icon still isn't aligned.** The body now says the output and input slider-row icons use the same alignment, but the diff only touches the three header icons. `MixerSection.swift:167` on main is still plain `.frame(width: 16)`, centered, and it isn't in this diff. It's also `size: 10` against the new row's `10.5`, if those are meant to match.

Still no hardware here, so nothing about real driver gain layouts is tested — the above is read off main and the diff.

**Impact:** the mic slider doesn't appear on the sweep that finds the device, and it stays stuck showing the previous device's gain after switching to a mic with no settable gain.
**Merge:** needs the generation comparison read before `updateVolumeListeners` first.

ref #853

GolamRafi27oo:
@PathGao The sweep-generation comparison is captured before updateVolumeListeners(for:), so listener rewiring no longer invalidates the volume read from the same sweep. This allows the slider to appear during initial discovery and correctly clears it when switching to a device without writable gain controls. I also updated the output slider-row icon to match the input row’s 10.5 pt size and leading-aligned 16 pt frame.

PathGao:
Both fixes hold up.

`sweepVolumeIsCurrent` is read before `updateVolumeListeners`, so the rewire's `volumeRefreshGeneration &+= 1` can no longer invalidate the sweep that produced the value. First sweep after `start()` publishes; switching to a mic with no settable gain publishes the nil and clears the slider instead of leaving the previous device's value on screen. `setInputVolume` now walks `mainInputVolumeAddresses()` (virtual-main, then main scalar) and returns on the first successful write, so a device exposing both master and per-channel keeps its balance; the element 1/2 loop only runs when no master write landed.

The icon sweep is complete, not half. `speaker.wave.2.fill` (`MixerSection.swift:130`), `bell.fill` (`:206`) and `mic.fill` (`:442`) are the only three `Label`-icon slots in that file, and the output slider-row icon at `:167` is now 10.5/leading like the new input row. The other `Image` sites in there are chevrons, the boost badge, and the per-app reset/mute buttons — trailing controls, not a leading icon column, so they're correctly out of scope.

Not a defect, just noting it: `setInputVolume` assigns `inputVolume = clamped` unguarded, while every other publish in this file goes through an inequality check for the reason spelled out at `apply` — "assigning a @Published property signals SwiftUI even when the value is identical". `setCurrentOutputVolume` guards its equivalent. Cost is extra panel invalidations, never a wrong value.

Still no hardware here. Real driver gain layouts, and whether a device that accepts a write actually fires the property notification the 30 ms read-back depends on, are untested by me — the above is read off main and the diff.

**Impact:** none.
**Merge:** ship it.

ref #853

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -109,6 +116,25 @@ final class AudioInputDeviceManager: ObservableObject {
...
+        let clamped = min(max(volume, 0), 1)
+        // Keep the control under the pointer while the HAL write catches up,
+        // and invalidate any older control or device-sweep read.
+        volumeRefreshGeneration &+= 1
+        inputVolume = clamped
+        halQueue.async { [weak self] in
+            let applied = Self.setInputVolume(Float32(clamped), for: device.audioObjectID)
+            DispatchQueue.main.async {
+                guard let self, self.effectiveInputDeviceUID == uid else { return }
+                if !applied {
```
vorssaint: a device that answers the write with success but keeps its own level leaves the slider showing a value the microphone never took, until some unrelated refresh pulls it back. read the level back after every write, not only after a failed one

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -282,6 +327,16 @@ final class AudioInputDeviceManager: ObservableObject {
...
             inputDevices = snapshot.devices
         }
+        let effectiveDeviceID = snapshot.resolution.effectiveUID
+            .flatMap { uid in snapshot.devices.first(where: { $0.uid == uid })?.audioObjectID }
+        // Rewiring listeners invalidates pending control reads. Capture the
+        // sweep's validity first so that initial wiring cannot invalidate the
+        // volume value discovered by this same sweep.
+        let sweepVolumeIsCurrent = volumeRefreshGeneration == snapshot.volumeGeneration
+        updateVolumeListeners(for: effectiveDeviceID)
+        if sweepVolumeIsCurrent, inputVolume != snapshot.inputVolume {
```
vorssaint: if the microphone changes while the slider is being dragged, the level this pass read for the new microphone is dropped as stale and the old value stays on screen until something else refreshes it. a microphone change should still show the level of the new device

Sources/Vorssaint/UI/MenuPanel/MixerSection.swift
```diff
@@ -162,9 +163,9 @@ struct MixerSection: View {
                     Image(systemName: mixer.systemOutputMuted == true || volume <= 0.001
                           ? "speaker.slash.fill"
                           : "speaker.wave.2.fill")
-                        .font(.system(size: 10, weight: .semibold))
+                        .font(.system(size: 10.5, weight: .semibold))
                         .foregroundStyle(.secondary)
-                        .frame(width: 16)
+                        .frame(width: 16, alignment: .leading)
```
vorssaint: with icons this different in width, left aligning them leaves the narrow mic icon with a visible gap before its slider, and this row was centered before. keep the fixed width and center the icons on all four rows

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
GolamRafi27oo: Fixed. Volume writes now always trigger a read-back, even when CoreAudio reports success, so devices that retain or clamp their own level update the slider correctly. The existing device/UID guards remain in place.

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
GolamRafi27oo: Fixed. Volume reads for a newly selected microphone are now applied even if the read is stale relative to an in-progress slider drag. Same-device stale reads are still ignored.

GolamRafi27oo: Fixed. Volume reads for a newly selected microphone are now applied even if the read is stale relative to an in-progress slider drag. Same-device stale reads are still ignored.

Sources/Vorssaint/UI/MenuPanel/MixerSection.swift
GolamRafi27oo: Fixed. The 16-point icon frames are now centered consistently across all mixer rows, while keeping the fixed width so the sliders stay aligned.

Issue #1102, opened 2026-08-28
Command Bar: Cmd+A, Cmd+C, Cmd+X and Cmd+V do nothing in the search field


alexis-morain:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

None of the four editing shortcuts work in the Command Bar's field. Typed text cannot be selected, copied, cut or replaced. Esc clears the query and Cmd+Delete deletes to the start of the line, because macOS binds that one itself, but that is all there is.

Two things are in the way, and the second one only shows up once the first is out of the way.

**1. The Cmd combinations are matched on key position, not on the printed letter.** `installMonitors` in `CommandBarService.swift` reads `Int(event.keyCode)` and compares it against `kVK_ANSI_Q, kVK_ANSI_W, kVK_ANSI_M, kVK_ANSI_H` to swallow the combinations the app's own menu owns. Virtual key codes are positional, so on a layout that moves the letters this aims at the wrong keys. On AZERTY the key printed A sits where Q lives on ANSI, so Cmd+A is read as a quit and swallowed on the spot; the field never sees it. The key printed `,` sits on M, so Cmd+, is swallowed as well and Settings does not open from the bar on that layout. QWERTZ and Dvorak have their own version of this.

**2. Even when the event reaches the field, nothing turns it into an action.** The bar is a `.nonactivatingPanel` and `present(_:)` only calls `orderFrontRegardless()` and `makeKey()`. That is deliberate, since the app the bar floats over has to stay active for the menu, window and selection rows to mean anything, but it leaves Vorssaint key without being active. The Edit menu that `installMainMenu()` puts up never gets to fire its key equivalents, and macOS binds none of these four on its own, unlike Ctrl+A, Ctrl+K or Cmd+Delete, which do work in the bar today. The comment on `installMainMenu()` already scopes that menu to the Settings window, where the app is active.

So the four shortcuts are dead in the bar on any layout, and on AZERTY Cmd+A is eaten before it gets anywhere.

### Steps to Reproduce

1. Open the Command Bar with its shortcut
2. Type a few characters
3. Press Cmd+A: nothing is selected
4. Same for Cmd+C, Cmd+X and Cmd+V

On an AZERTY keyboard, press Cmd+, from the bar as well: Settings does not open.

### Feature Area

Tools

[...]

Issue #1047, opened 2026-08-26
Voirssant does not respect (custom) keyboard layout for shortcuts


lacinoire:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

I am using a different keyboard layout than my computer was built for (QuertZ instead of QuertY)
When configuring and using the shortcuts within vorssaint, it is detecting they keyscodes from my keyboard, instead of the remapped ones picked up by all other applications.

The same appears when using my custom keyboard layout made with ukelele. I did not run into any other problems with the keyboard layout and the MacOS golden gate beta yet.

### Steps to Reproduce

1. Use computer delivered with querty keyboard layout
2. Switch to keyboard layout to quertz 
3. Open Vorssaint
4. Select screen capture setting
5. make global shortcut 
6. click keys that have printed on the keyboard: shift + cmd + y (due to the Quertz, the key printed with "y" should produce a "z").

When adhering to the keyboard layout, the shortcut should be set to shift + cmd + z.
However, it is set to shift.+ cmd + y

### Feature Area

Other

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

### Language

English

### Mac and displays

MacBook Pro 14-inch Nov 2023

### Installation Method

dmg

lacinoire:
So the core issue seems to be fixed now, I can set the keyboard shortcuts as expected.

*However*: The display of the selected shortcut can behave weirdly!

In the "screenshot" menu, 
when I set a keyboard shortcut to (1), 
reset it again (by clicking on the button containing the keyboard shortcut), 
and then set an "warning-triggering" shortcut (2) (e.g. my typical paste)

The button showing the shortcut combination shows (1) and not as expected (2).

In my testing this only happened when (2) was a shortcut that triggered a warning text such as "This shotrcut is already used by..."


Sometimes, the text also shows absolutely absurd values. like I click shift, cmd, 4 and the button shows alt, shift, cmd 2.
The setting it to another shortcut, then setting it agin to shift, cmd, 4 works

lacinoire:
just a bit of context looking into the merged code:

I am using a custom keyboard layout (so it for example also switches around the numbers). When I saw the same issue in Quertz/Querty I reported it like this as a broader issue.
[...]

PR #1103, opened 2026-08-28
fix(command-bar): give the field back Cmd+A, C, X and V

alexis-morain:
## Summary

The Command Bar's field answers none of the four editing shortcuts. Two things were in the way.

The Cmd combinations were matched on the key's position, so a layout that moves the letters aimed them at the wrong commands. On AZERTY the key printed A sits where Q lives on ANSI, so Cmd+A was read as a quit and swallowed before the field could see it; the key printed `,` lands on M and took Cmd+, away from Settings the same way. Matching the character the person sees printed follows every layout that prints a Latin letter, and every letter the bar claims goes through one switch, Cmd+K and Cmd+P included, so no two shortcuts in the same field disagree about what a letter is.

A layout that prints no Latin letter has none to answer with. Cyrillic and Greek go by the key's position instead, which is where macOS resolves those layouts' command shortcuts: without that fallback Cmd+Q is not recognised in the bar, the event goes back to the system, and the app's own menu quits Vorssaint behind it. That is measured, not reasoned — see below. The App Switcher already reads its letters this way, so the reading moved into `KeyboardLetters` and both sides share one definition; `SwitcherSupport.latinLetter` and its `USKeyPosition` are gone. Punctuation the keyboard prints outright is answered before the position, or AZERTY's comma, which sits on the US M key, would resolve to `m` and be swallowed as a minimise. Digits stay positional on purpose: the top row needs Shift to reach a number on AZERTY, so Cmd+1…9 has to keep reading key codes.

Even reaching the field, none of the four would have worked. The bar is a non-activating panel on purpose, so the app is key without ever being active and the Edit menu never gets to fire its equivalents; macOS binds none of these four on its own. They are handed to the responder chain directly, and an event nobody takes is returned to the field untouched, so a combination carrying another modifier still means what it meant before.

One side effect that belongs in writing: `charactersIgnoringModifiers` reports the character with Shift applied. Where the comma is unshifted, as on ANSI and AZERTY, Cmd+Shift+, no longer opens Settings the way the key code did; on a layout that puts the comma behind Shift, that combination now does open it. Same for q/w/m/h/k/p. None of them were meant to carry Shift, so it reads as a tightening rather than a loss, but it is a change.

No new setting and no new string. One new file, `Core/KeyboardLetters.swift`, added to the test harness's source list in `build.sh`.

## Verification

MacBook Air, Apple M5, macOS 26.5.2, single built-in display, French AZERTY layout with Russian added for the run below. Release build installed over the official one, signed with the self-signed identity from `Tools/setup-signing.sh`.

```sh
./build.sh                                                        # no warnings
./build.sh --test                                                 # TESTS OK (8935 checks)
/Applications/Vorssaint.app/Contents/MacOS/Vorssaint --selftest   # SELFTEST OK
```

What a key reports under Command, read off `NSEvent` in a throwaway window on the Russian layout:

```
⌘ + Й     keyCode=12  characters=q  charactersIgnoringModifiers=й
⌘ + Ф     keyCode=0   characters=a  charactersIgnoringModifiers=ф
Й alone   keyCode=12  characters=й  charactersIgnoringModifiers=й
```

macOS does translate to Latin under Command, but into `characters`; `charactersIgnoringModifiers`, which is what this monitor reads, keeps the Cyrillic. That is the whole failure: the bar recognises nothing, hands the event back, and the main menu — which matches on the translated character — quits the app.

In the bar, French AZERTY: Cmd+A selects the query, Cmd+X, Cmd+C and Delete then act on the selection, Cmd+V pastes. Cmd+, opens Settings from the bar. Cmd+K, Cmd+P, Cmd+Return and Cmd+1…9 still do what they did.

In the bar, Russian: Cmd+Й, on the US Q position, does nothing and Vorssaint stays up; Cmd+Ф selects the query; Cmd+Б, on the US comma position, opens Settings. On the same machine before the fallback, with the bar open, Cmd+Й quit Vorssaint. Cmd+Ф appeared to work there, which is the same leak with a harmless end: the bar recognised nothing and something downstream resolved the translated `a`.

The position table could be `characters` instead, since that is the system's own answer under Command. I kept the table: it is one definition shared with the App Switcher, and the suite can exercise it without a layout installed. They agree on the layout measured here. Worth a word if you would rather read the event.

What I could not test: QWERTZ, Turkish, Greek, and Dvorak. A Dvorak-QWERTY-Cmd layout swaps the layout while Cmd is held, and the measurement above says `charactersIgnoringModifiers` would report the layout's own letter there, so that case follows the key position rather than the overlay — untested either way, and I have no machine set up for it. Nothing was tested on Intel, on macOS 14 or 15, or with an external keyboard whose layout differs from the built-in one.

The Quick Launcher and the Scratchpad read key codes the same way and may well have the same problem, but I have neither read them closely nor tested them, so they are left alone here. Three more of the same shape turned up while answering review, all outside this PR: `ClipboardHistoryService.swift:1188` and `ScreenshotEditorController.swift:1088` pin by key code, and `paste(_:)` in this file synthesises its Cmd+V from `kVK_ANSI_V`, which prints `.` on Dvorak. The first two are switch-shaped like this one; the third wants `MouseNavigationKeys.keyStroke(for:)` and is closer to #1047.

The Ctrl branch in the same monitor, Ctrl+N and Ctrl+P, keeps matching the printed character with no positional fallback. No menu sits behind those two waiting to quit the app, and macOS translates for Command, not for Control, so a fallback there would add behaviour on Cyrillic that nobody has asked for.

## Anything else

Refs #1102

### DIFF
--- a/Sources/Vorssaint/Core/KeyboardLetters.swift
+++ b/Sources/Vorssaint/Core/KeyboardLetters.swift
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Carbon.HIToolbox
+import Foundation
+
+/// What a keystroke means to a shortcut, on every layout.
+///
+/// A keyboard answers by the letter it types, so a shortcut stays on the key
+/// it is printed on even where the layout moves it: AZERTY puts A where ANSI
+/// has Q, Turkish moves both. Layouts that type no Latin letter at all, like
+/// Cyrillic and Greek, have no letter to answer with, so the key's position
+/// stands in — which is where macOS resolves those layouts' own command
+/// shortcuts anyway.
+enum KeyboardLetters {
+    /// The plain letter a keystroke typed, when it typed one. Accents fold
+    /// away, so a letter of a Latin alphabet is never mistaken for one of the
+    /// keys a panel claims.
+    static func latinLetter(in text: String?) -> Character? {
+        guard let folded = text?.folding(options: [.diacriticInsensitive, .caseInsensitive],
+                                         locale: .current),
+              folded.count == 1,
+              let letter = folded.first,
+              letter.isASCII,
+              letter.isLetter
+        else { return nil }
+        return letter
+    }
+
+    /// The character a key types on a US keyboard: the letters, plus the
+    /// comma the Command Bar claims for Settings.
+    static func usPositionCharacter(for keyCode: Int) -> Character? {
+        switch keyCode {
+        case kVK_ANSI_A: return "a"
+        case kVK_ANSI_B: return "b"
+        case kVK_ANSI_C: return "c"
+        case kVK_ANSI_D: return "d"
+        case kVK_ANSI_E: return "e"
+        case kVK_ANSI_F: return "f"
+        case kVK_ANSI_G: return "g"
+        case kVK_ANSI_H: return "h"
+        case kVK_ANSI_I: return "i"
+        case kVK_ANSI_J: return "j"
+        case kVK_ANSI_K: return "k"
+        case kVK_ANSI_L: return "l"
+        case kVK_ANSI_M: return "m"
+        case kVK_ANSI_N: return "n"
+        case kVK_ANSI_O: return "o"
+        case kVK_ANSI_P: return "p"
+        case kVK_ANSI_Q: return "q"
+        case kVK_ANSI_R: return "r"
+        case kVK_ANSI_S: return "s"
+        case kVK_ANSI_T: return "t"
+        case kVK_ANSI_U: return "u"
+        case kVK_ANSI_V: return "v"
+        case kVK_ANSI_W: return "w"
+        case kVK_ANSI_X: return "x"
+        case kVK_ANSI_Y: return "y"
+        case kVK_ANSI_Z: return "z"
+        case kVK_ANSI_Comma: return ","
+        default: return nil
+        }
+    }
+
+    /// The character a shortcut should be matched against: the letter the key
+    /// prints, the character itself when it prints punctuation the keyboard
+    /// owns outright, and the key's US position when it prints neither.
+    ///
+    /// The middle rule is what keeps AZERTY's comma, which sits on the ANSI M
+    /// key, from being read as a ⌘M the app's menu would swallow.
+    static func shortcutCharacter(typedCharacter: String?, keyCode: Int) -> Character? {
+        if let letter = latinLetter(in: typedCharacter) { return letter }
+        if let typed = typedCharacter, typed.count == 1, let character = typed.first,
+           character.isASCII, !character.isLetter, !character.isWhitespace {
+            return character
+        }
+        return usPositionCharacter(for: keyCode)
+    }
+}
--- a/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
+++ b/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
@@ -2416,6 +2416,21 @@ final class CommandBarService: ObservableObject {
 
     // MARK: - Monitors
 
+    /// Sends one editing shortcut down the responder chain, to the field
+    /// editor of whichever panel is key. Answers whether something took it,
+    /// so a combination nobody wanted is handed back to the field untouched.
+    private func sendEditingAction(for event: NSEvent) -> Bool {
+        let action: Selector
+        switch event.charactersIgnoringModifiers?.lowercased() {
+        case "a": action = #selector(NSText.selectAll(_:))
+        case "c": action = #selector(NSText.copy(_:))
+        case "x": action = #selector(NSText.cut(_:))
+        case "v": action = #selector(NSText.paste(_:))
+        default: return false
+        }
+        return NSApp.sendAction(action, to: nil, from: nil)
+    }
+
     private func installMonitors(for panel: NSPanel) {
         removeMonitors()
         keyMonitor = NSEvent.addLocalMonitorForEvents(matching: .keyDown) { [weak self, weak panel] event in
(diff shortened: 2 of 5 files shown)


PathGao:
`kVK_ANSI_P` at `CommandBarService.swift:2501` is the same shape and didn't move. ⌘K went onto the printed character, ⌘P stayed positional, five lines below it in the same monitor:

```swift
if event.modifierFlags.contains(.command), Int(event.keyCode) == kVK_ANSI_P {
```

AZERTY keeps P where ANSI has it, which is why your run didn't catch it. Dvorak and Colemak both move it, and Dvorak-QWERTY-Cmd is the layout the body says should now follow the letters the overlay shows. After this PR ⌘K follows the print and ⌘P follows the position, so two shortcuts in the same bar disagree about what a letter means. It belongs in the character switch you just built.

The rest holds up. Digits staying on key codes is the right call and the reason is stated. `onlyCommand` guarding only the four editing keys and not q/w/m/h/`,` matches what the key-code version already did, so nothing moved there. One undeclared side effect: `charactersIgnoringModifiers` honours Shift, so on a layout where the comma sits behind Shift, ⌘⇧+that key no longer opens Settings where the key-code version did. Harmless, but it isn't in the body.

What I did not verify: whether `NSApp.sendAction(_:to:from:)` actually reaches the field editor of a non-activating panel. That needs a run and I only read the code. Every layout claim above is reasoning, not testing.

**Impact:** ⌘P in the bar keeps pinning by key position — wrong key on Dvorak and Colemak, unchanged on ANSI and AZERTY.
**Merge:** needs `kVK_ANSI_P` moved onto the same character match first.

ref #1102

alexis-morain:
Moved, in f0407a2. The pin is now `case "p"` in the same switch as ⌘K, and no letter in the file is matched by position any more. AZERTY had nothing to show me there, since P sits where ANSI puts it, so it took a reader to catch it.

Two things in your note to correct, neither of which changes the conclusion. `kVK_ANSI_P` was at line 2532, not 2501 (2501 is the `openSettingsWindow()` call inside the `,` case), and it sat about thirty lines below ⌘K with the ⌘Return block in between.

The Shift example is the wrong way round. `charactersIgnoringModifiers` reports the character with Shift applied, so on a layout that puts the comma behind Shift, ⌘⇧+that key now *does* open Settings: the event reports `,`. What loses something is the opposite case, ANSI and AZERTY, where the comma is unshifted and ⌘⇧, used to reach Settings through the key code. Same for q/w/m/h/k/p. It is a real side effect and it wasn't in the body; it is now, stated that way round.

On `NSApp.sendAction(_:to:from:)` reaching the field editor of a non-activating panel: that half is run, not read. French AZERTY, macOS 26.5.2, release build installed over the official one. ⌘A selects the query, ⌘C and ⌘X act on the selection, ⌘V pastes, and an event nobody takes comes back to the field untouched. What stays unverified is every layout other than this one, ⌘P first of all, since this machine cannot tell the key code from the letter there.

Suite is green after the move: `./build.sh --test`, TESTS OK (8929 checks), no warnings.

Two more of the same shape, since you pointed me at #1047. `ClipboardHistoryService.swift:1175` matches ⌘C and ⌘A on the character with the reason written out, then pins on `kVK_ANSI_P` at 1188; `ScreenshotEditorController.swift:1088` pins by position in a switch that reads key codes throughout. The one I would look at first is in this very file, though: `paste(_:)` synthesises the ⌘V it presses in the target app from `CGKeyCode(kVK_ANSI_V)` (2007 and 2010). AZERTY and QWERTZ keep V where ANSI has it, so this is invisible here, but on Dvorak that key code prints `.` and the target app is handed a ⌘. it has no reason to read as a paste. That one needs the reverse lookup rather than a switch, so it doesn't belong in this PR, and it is reasoning: I have no Dvorak machine to run it on.

PathGao:
Both corrections stand. `charactersIgnoringModifiers` applies Shift, so the loss is on ANSI/AZERTY — I had it backwards. The ⌘P move is right.

One thing left. `SwitcherSupport.letterAction` on main is this same shape — Q quits, W closes, layouts that move the letters — and it falls back to the key position when the typed character isn't a Latin letter, naming Cyrillic and Greek as measured. The new switch has no fallback. If that measurement holds with Command held, then on Russian or Greek none of q/w/m/h match, ⌘Q stops being swallowed, and the bar quits the app. The body says character matching "fixes every layout at once".

I did not verify what a Cmd-held Cyrillic key reports; that parenthetical reads both ways. One run settles it and it needs no hardware — System Settings → Keyboard → Input Sources, add Russian. Same route gets you Dvorak for ⌘P.

The reverse lookup `paste(_:)` needs already exists: `MouseNavigationKeys.keyStroke(for:)`, character → (keyCode, needsShift), one caller today.

**Impact:** on layouts that type no Latin letter, ⌘Q in the bar may quit Vorssaint where the key code swallowed it. ANSI, AZERTY, QWERTZ unaffected.
**Merge:** needs one run on a Cyrillic layout first.

ref #1102

alexis-morain:
You were right, and it reproduces. Fixed in 952f49e; the run you asked for is below.

I read the parenthetical the way you did not, so I measured it. A throwaway window reading `NSEvent` on the Russian layout:

```
⌘ + Й     keyCode=12  characters=q  charactersIgnoringModifiers=й
⌘ + Ф     keyCode=0   characters=a  charactersIgnoringModifiers=ф
Й alone   keyCode=12  characters=й  charactersIgnoringModifiers=й
```

macOS does translate to Latin under Command, but into `characters`. `charactersIgnoringModifiers`, which is what the monitor reads, keeps the Cyrillic. So the bar recognised nothing, handed the event back, and the main menu — which matches on the translated character — quit the app. With the bar open, on this machine, ⌘Й quit Vorssaint before the fix and does nothing after it. ⌘Ф looked like it worked before the fix, which threw me for a while: same leak, harmless end, something downstream resolved the translated `a`.

The fallback goes through `KeyboardLetters.shortcutCharacter(typedCharacter:keyCode:)`: the letter the key prints, then the punctuation it prints outright, then the US position. `SwitcherSupport.latinLetter` and its `USKeyPosition` are gone and `letterAction` reads the same helper, so there is one definition of what letter a key typed instead of two.

Three things it needed beyond q/w/m/h. ⌘A/C/X/V fall in the same hole, so `sendEditingAction` reads the same resolver — otherwise the four keys this PR exists for stay dead on the layouts that also lose ⌘Q. ⌘, too, since Russian prints `б` on the comma key. And punctuation has to be answered before the position, or AZERTY's comma, which sits on the US M key, resolves to `m` and is swallowed as a minimise; that one has a test of its own.

Your call on one thing. `characters` is the system's own answer under Command, so the resolver could read that instead of a position table. I kept the table — one definition shared with the App Switcher, and the suite exercises it without a layout installed — and they agree on the layout I measured. Say the word if you would rather it read the event.

Runs: `./build.sh --test`, TESTS OK (8935 checks), no warnings; `--selftest` OK on the installed build. In the bar on Russian: ⌘Й does nothing, ⌘Ф selects the query, ⌘Б opens Settings. On French AZERTY the earlier pass is unchanged. Untested: QWERTZ, Turkish, Greek, Dvorak. The measurement above also says a Dvorak-QWERTY-Cmd layout would follow the key position rather than the overlay, so I took that claim back out of the body rather than leave it standing.

The Ctrl branch keeps matching the printed character with no fallback, on purpose: macOS translates for Command, not for Control, and no menu sits behind Ctrl+N or Ctrl+P waiting to quit the app.

`paste(_:)` and `keyStroke(for:)`: separate issue once this lands.

PathGao:
The Cyrillic measurement settles it, and the fallback is the right shape: one resolver, both panels, punctuation answered before position.

The digit rule doesn't survive it, though. `shortcutCharacter` runs before `digitIndex(for:)`, and on French AZERTY the unshifted 9 key prints `ç`, which `latinLetter` folds to `c`:

```swift
case "a", "c", "x", "v":
    guard onlyCommand else { break }
    return self.sendEditingAction(for: event) ? nil : event
```

`copy:` is always taken, so ⌘9 returns nil and row 9 never runs — the exact gesture the body keeps digits positional for, on the layout you run. The 0 key folds `à` to `a` the same way; nothing claims ⌘0, so that one only costs a stray select-all. Moving the existing `digitIndex` block above the character switch covers it without new code.

I did not run this. AZERTY's 9 printing `ç` is the layout, not a measurement — one press on your machine confirms it or kills it.

On your question: keep the table. `characters` would read better on Dvorak-QWERTY-Cmd, but it trades a definition the suite exercises for one that needs a layout installed to test, and buys you the single layout you can't check either way.

**Impact:** ⌘9 in the bar copies instead of running row 9 on AZERTY. Other digits and other Latin layouts unaffected.
**Merge:** needs the digit check moved ahead of the character switch first.

ref #1102

vorssaint:
closing this. the same fix is already on main through #1320, went out in 3.3.3, and on #1102 you confirmed the shortcuts work again on your mac
what is left here is the shared letter helper, and on its own it changes nothing a person would notice, so it would not earn its place
the release notes for that version already credit you for the report and the diagnosis

Issue #611, opened 2026-08-12
Upload Recorded Video to Cloud


risalfajar:
### Problem or Motivation

Frequently I want to share a quick video for my client through Whatsapp. But sending a video via Whatsapp destroys the video quality and take spaces on my client's device. I prefer to share via link.

### Proposed Solution

[Snapzy](https://github.com/duongductrong/Snapzy) already does this, it supports Cloudflare, AWS, and Google Drive as providers.

### Feature Area

Tools

### Mockups / References

<img width="1520" height="1276" alt="Image" src="https://github.com/user-attachments/assets/8627309c-6618-4780-a698-804e96a23279" />

PathGao:
Most of this already shipped. Vorssaint 3.3.1 (2026-08-09, three days before this issue) added temporary links for recordings: finish a recording, and the editor can create a link that lives for 1 or 6 hours. The video is compressed locally to under 100 MB first, then uploaded, and the local compressed copy is deleted right after. You can delete the link from the app at any time before it expires, and the whole feature can be turned off under Settings -> Screen recording.

That should cover the WhatsApp problem with no setup at all, so it is worth trying on 3.3.1 or later before going further.

On the provider approach: supporting Cloudflare, AWS or Google Drive means the app has to hold your cloud credentials. That is a new kind of secret to keep safe, added for a case the built-in link already handles, so it is a poor trade.

6 hours is the current ceiling because the links live on the maintainer's own server. Storage is limited, and there is no login on a link, so anyone holding the URL can watch it. A shorter life is the main protection.

risalfajar:
Sadly for my use case the 6 hours limit is not enough.

I think the long-term solution is for the users to provide their own cloud. Especially since the number of users will keep growing.

I agree though that this isn't high priority, users can upload the video themselves or use other recording tool.

khairulimran-97:
Adding a use case for user-supplied storage, and a shape that might sidestep the credential concern.

The 1/6-hour links cover "send it now" well. What they don't cover is anything a client comes back to — a walkthrough referenced during a week-long review, or a link in a ticket someone opens on Monday. Re-recording and re-sending on every expiry is what pushes people back to WhatsApp.

On credentials: I don't think this needs Cloudflare/AWS/Google Drive integrations at all. One generic destination would do:

- **Upload target** — a URL the user supplies
- **Method** — HTTP `PUT` (or multipart `POST`)
[...]

PR #1104, opened 2026-08-28
feat(quick-tools): add selected-text translation

zhanghk2001:
## Summary

- add installable/uninstallable `划词翻译` feature to the Features hub
- migrate the selected-text workflow from Tom into Vorssaint AppKit/SwiftUI style
- read selected text through Accessibility with a serial pasteboard fallback
- stream OpenAI-compatible chat-completion translations in a non-activating floating panel
- store provider API keys in Keychain and avoid persisting source/translation content
- add configurable provider, model, target language, prompts, shortcut, and connection test
- localize provider configuration and Keychain error messages through the active app language
- add `划词翻译` to the menu bar Utilities section with a manual blank-draft entry

Default shortcut: `⌃⌥⌘Y` (chosen to avoid the existing screen OCR shortcut collision).

## Follow-up review fixes

- Preserved the legacy `Command-V` timeout fallthrough while making selection `Command-C` fail safely when modifiers remain held.
- Missing selection-translation defaults now fall back to the single `SelectionTranslationSettingsStore.default*` source of truth instead of crashing when registration is unavailable.
- Clipboard-history capture is deferred before synthetic `Command-C`, then resumed after guarded restoration so temporary selection data is not recorded while newer user copies remain eligible.
- Re-merged the latest `upstream/main` through `f4d7966` without rebasing or force-pushing; the merge commit is `f4d7966`.
- Restored the latest `issue-fix-status.yml` behavior, including the `half` issue release-sweep handling.
- Nested capture deferrals now ignore each valid synthetic change count before the final deferral resumes history capture.
- Selection fallback deferrals have an idempotent five-second watchdog so a missing callback cannot disable clipboard history for the session.

## Scope note

This PR provides selected-text translation. The screenshot OCR and screenshot-translation request in issue #489 remains tracked separately.

## Verification

- `./build.sh --test` — passed: **9540 checks**
- `./build.sh` — passed; Release Stage bundle assembled and ad-hoc signed (no Developer ID identity installed on this machine)
- `build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest` — `SELFTEST OK`
- `codesign --verify --deep --strict build/stage/Vorssaint.app` — passed
- `git diff --check` — passed
- `git merge-base --is-ancestor upstream/main HEAD` — passed
- The workflow file matches the latest `upstream/main` exactly.
- Regression coverage includes timeout policy decisions, isolated missing-default fallback, nested clipboard capture deferral, one-shot deferral lifecycle, localized configuration errors, Qwen/generic request routing, clipboard preservation, and the Utilities menu selection-translation entry.

The six original review items remain addressed. The latest follow-up fixes are in commit `f4d7966`.

## Anything else

This PR intentionally does not change issue #489 status or labels.

## Source reference

Sorry, I forgot to include the source link for Tom.

Tom is an open-source, locally running text-selection translation app that I developed myself. It stores API credentials in the macOS Keychain and supports user-configured OpenAI-compatible translation providers.

[Tom source repository](https://github.com/zhanghk2001/Tom)

PR #1104 migrates Tom’s original selected-text translation workflow into Vorssaint. Tom was used as the functional reference for the initial migration.

The subsequent interaction, UI, and reliability improvements were implemented and tested locally in Vorssaint after the migration. They are based on testing the Vorssaint implementation, not on later changes from Tom. The current code follows Vorssaint’s own architecture and visual style.


PathGao:
Reviewed as a draft. Six things, worst first.

**1. The pasteboard fallback destroys anything on the clipboard that isn't plain text.** `SelectionTranslationSelectionReader.read()` snapshots one flavor:

```swift
let previous = board.string(forType: .string)
...
if let previous { board.clearContents(); board.setString(previous, forType: .string) }
```

Image, files, rich text: the synthesized ⌘C overwrites them, `previous` is nil, nothing is restored, it's gone. `TransientPaste.snapshot(of:)` already does the all-flavors version and returns nil (fails open rather than clearing incomplete user data) when a flavor can't be preserved. Reuse it.

**2. The ⌘C is posted while the trigger's own modifiers are still held.** Default shortcut is ⌃⌥⌘Y. `postCopy()` posts down/up back to back with `flags = .maskCommand` while ⌃ and ⌥ are still physically down, so the front app receives ⌃⌥⌘C. `TransientPaste.postPasteWhenModifiersReleased` polls `CGEventSource.flagsState` until they clear, then waits 0.06s and spaces down/up by 0.04s. Same problem, already solved in this repo. This is call site twelve for #937.

**3. The panel is presented before the read, and it is not a non-activating panel.** Body says "non-activating floating panel"; the styleMask is `[.titled, .closable, .resizable]`. Every other floating surface in the repo carries `.nonactivatingPanel` (`rg nonactivatingPanel Sources` — ~30 hits, no exceptions). `trigger()` posts `.selectionTranslationPresentPanel` and only then starts the read, and `CommandBarSelectionReader.readSelectedText()` returns `""` when the frontmost app is us — so if the panel ever takes focus, the AX path bails and the fallback ⌘C goes into our own panel. Read first, present after; add `.nonactivatingPanel` either way.

**4. Token accounting is dead weight.** `usage`, `startedAt`, `SelectionTranslationTokenEstimator`, `TokenUsage.adding`, `Language.qwenCode`, `LanguageSelection.swap()`, `ClientError.cancelled` — nothing in the panel or the settings page reads any of them. ~80 lines plus a test asserting the estimator. Delete unless a later commit puts usage on screen. Same for the notification hop: `SelectionTranslationPanelController.shared.present()` called from the service removes the `Notification.Name`, the observer, and the `_ = SelectionTranslationPanelController.shared` line in `AppDelegate`.

**5. Every default is written twice and the copies disagree.** `Defaults.registeredDefaults` sets `selectionTranslationBaseURL = "https://api.openai.com/v1"`, then `snapshot()` does `?? ""`. Model: `"gpt-4o-mini"` vs `""`. Provider name and `"zh-Hans"` also appear in both. `register(defaults:)` runs at launch so the `??` arms are unreachable, which is worse than a bug — a reader can't tell which copy is the truth. Drop the second one.

**6. Smaller:**
- `httpStatus(code, message)` always receives `let message = "Request failed"`. A bad key shows `HTTP 401: Request failed` instead of what the provider said. In a feature whose failure modes are all key/model/URL, the response body is the only diagnostic there is.
- `renderUserPrompt` branches on `sourceLanguage != SelectionTranslationLanguage.automatic.displayName` — a display string as a control-flow predicate, the exact trap this repo keeps hitting. It's also unreachable: no source-language picker exists and `snapshot()` always leaves it `.automatic`. Compare the enum, or drop the branch along with the rest of the source-language leftovers.
- `SelectionTranslationFeatureStrings` covers 4 of the 13 `AppLanguage` cases, which is a fine start, but `SelectionTranslationSettings` hardcodes "Saved", "Connected", "Permissions", "Open Accessibility Settings", "{{text}} is replaced…" and the panel hardcodes "Stop" — a zh-Hans user gets a half-translated page.
- `QuickToolHotkey(id: 31)` sits inside the range `ScreenCaptureService` allocates automatically ("Ids continue past the hand-assigned quick tool range, which ends at 24" — it starts at 25, four tools today). A fifth capture tool takes 29, a seventh takes 31, and `instances[id] = self` silently overwrites the loser. 21 is free.
- The panel labels the source box "Source language" and the result box "Target language". Those strings name settings, not the contents.

Not verified: anything needing real hardware — whether the AX read actually loses the race against the panel, whether ⌃⌥⌘C breaks copy in common apps, whether per-token repaint stutters.

**Impact:** every trigger that falls through to the ⌘C fallback (any app that doesn't expose an AX selection) destroys non-text clipboard content, and may send ⌃⌥⌘C instead of ⌘C.
**Merge:** needs 1 and 2 first — both have a working implementation in `TransientPaste` to copy.

ref #489

zhanghk2001:
Addressed the six draft-review items in [`504114b`](https://github.com/vorssaintapp/vorssaint-utils/commit/504114b79ef20ff94e70905f4c00f6d765dc5cd2) as follows.

## 1. Preserve all clipboard content types

**File:** [`Sources/Vorssaint/Services/TransientPaste.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/TransientPaste.swift#L108-L123)

`TransientPaste.snapshot(of:)` now owns the all-flavor snapshot implementation. It copies every advertised type from every pasteboard item. If any flavor cannot be materialized, it returns `nil`, so the selection fallback aborts without replacing the existing clipboard.

```swift
static func snapshot(of pasteboard: NSPasteboard) -> [NSPasteboardItem]? {
    guard let items = pasteboard.pasteboardItems else {
        return pasteboard.types?.isEmpty == false ? nil : []
    }

    var snapshot: [NSPasteboardItem] = []
    for item in items {
        let copy = NSPasteboardItem()
        for type in item.types {
            guard let data = item.data(forType: type) else { return nil }
            copy.setData(data, forType: type)
        }
        snapshot.append(copy)
    }
    return snapshot
}
```

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift#L34-L89)

The reader uses that shared snapshot. It restores the clipboard only when the current change count still belongs to its synthesized Command-C. A newer user copy is left untouched. Both the temporary copy and restoration are excluded from clipboard history.

Tests for preservation and guarded restoration are in [`Tests/SelectionTranslationTests.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Tests/SelectionTranslationTests.swift#L132-L151).

## 2. Wait for trigger modifiers before posting Command-C

**File:** [`Sources/Vorssaint/Services/TransientPaste.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/TransientPaste.swift#L128-L170)

The selection fallback now reuses `postKeyWhenModifiersReleased(...)`. It waits for Command, Option, Shift, and Control to be physically released, applies the existing 60 ms settling delay, spaces key-down/key-up by 40 ms, and fails when Secure Input is active or the retry limit is reached.

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift#L43-L54)

```swift
TransientPaste.postKeyWhenModifiersReleased(
    keyCode: SelectionTranslationPasteboardSupport.copyKeyCode,
    flags: .maskCommand
) { succeeded in
    // Continue the pasteboard read only after Command-C was posted.
}
```

## 3. Read the selection before presenting a non-activating panel

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationService.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationService.swift#L60-L102)

`trigger()` now records the pointer location and enters `reading` without presenting the panel. After the read finishes:

- A non-empty selection presents the panel and starts translation.
- An empty selection presents an editable blank draft without making a request.
- Cancellation and generation checks prevent an older read from reopening or updating the panel.

```swift
task = Task { [weak self] in
    let text = await SelectionTranslationSelectionReader.read()
    guard let self, !Task.isCancelled,
          self.generation == current else { return }

    self.draft.source = text

    if text.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
        self.phase = .ready
        self.requiresSubmission = true
        SelectionTranslationPanelController.shared.present(
            anchor: self.panelAnchor,
            focusSourceEditor: true
        )
    } else {
        self.submittedDraft = self.draft
        SelectionTranslationPanelController.shared.present(
            anchor: self.panelAnchor,
            focusSourceEditor: false
        )
        self.beginTranslation(self.draft, generation: current)
    }
}
```

**File:** [`Sources/Vorssaint/UI/SelectionTranslationPanel.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/UI/SelectionTranslationPanel.swift#L72-L89)

```swift
styleMask: [
    .titled,
    .closable,
    .miniaturizable,
    .resizable,
    .nonactivatingPanel
]
```

The old notification-based panel presentation path and AppDelegate bootstrap reference were removed. The remaining notification is limited to focusing the source editor after an empty selection.

## 4. Remove or surface unused code

**Files:**

- [`SelectionTranslationModels.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationModels.swift)
- [`SelectionTranslationService.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationService.swift)
- [`SelectionTranslationPanel.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/UI/SelectionTranslationPanel.swift)

Removed:

- `SelectionTranslationClientError.cancelled`
- `SelectionTranslationTokenUsage.adding(_:)`
- The mutating language `swap()`
- The unused `SelectionTranslationResult.text`
- The unused `shouldSubmit` parameter
- The immediate-presentation test helper
- The old panel-presentation notification hop
- Unused selection-translation string fields

Retained and surfaced:

- Token usage and frozen/running timing are displayed in the [panel footer](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/UI/SelectionTranslationPanel.swift#L342-L389).
- Source and target language selections are used by the panel and request builder.
- Language exchange uses the pure [`swapped()`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationModels.swift#L110-L116) implementation.

## 5. Use one source of truth for defaults

**File:** [`Sources/Vorssaint/Core/Defaults.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Core/Defaults.swift#L1155-L1162)

Every registered selection-translation default now refers to `SelectionTranslationSettingsStore.default*`. There is no second set of literal fallback values.

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSettingsStore.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSettingsStore.swift#L69-L96)

`snapshot(defaults:apiKey:)` reads only the supplied Vorssaint `UserDefaults` and API key. The injectable parameters also let tests avoid the real preferences and Keychain.

```swift
static func snapshot(
    defaults: UserDefaults = .standard,
    apiKey: String? = nil
) -> SelectionTranslationSettingsSnapshot {
    func string(_ key: String) -> String {
        guard let value = defaults.string(forKey: key) else {
            preconditionFailure("Missing registered default for \(key)")
        }
        return value
    }

    let system = string(DefaultsKey.selectionTranslationSystemPrompt)
    let user = string(DefaultsKey.selectionTranslationUserPrompt)
    let prompts = (
        try? SelectionTranslationPromptTemplates(
            systemPrompt: system,
            userPrompt: user
        )
    ) ?? .default

    return SelectionTranslationSettingsSnapshot(
        providerName: string(DefaultsKey.selectionTranslationProviderName),
        baseURL: string(DefaultsKey.selectionTranslationBaseURL),
        model: string(DefaultsKey.selectionTranslationModel),
        apiKey: apiKey ?? SelectionTranslationKeychain.read(),
        languages: SelectionTranslationLanguageSelection(
            target: .matching(
                string(DefaultsKey.selectionTranslationTargetLanguage)
            )
        ),
        prompts: prompts
    )
}
```

The old Tom UserDefaults suite and Keychain fallback paths were removed. No legacy data is deleted. Isolated-default tests are in [`Tests/SelectionTranslationTests.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Tests/SelectionTranslationTests.swift#L177-L195).

## 6. Address the smaller items

### Provider response diagnostics

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationClient.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationClient.swift#L37-L129)

Non-2xx responses now read `error.message` from the provider JSON. If that shape is unavailable, the trimmed response body is shown instead.

### Prompt rendering

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationModels.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationModels.swift#L230-L240)

Prompt rendering now performs deterministic `{{from}}`, `{{to}}`, and `{{text}}` replacement. It no longer uses a localized display string as a control-flow predicate.

### Localized UI text

**File:** [`Sources/Vorssaint/Core/SelectionTranslationStrings.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Core/SelectionTranslationStrings.swift#L6-L102)

Settings and panel text, including saved/connected states, permission actions, translation controls, labels, and footer formats, is provided through `SelectionTranslationFeatureStrings`.

### Hotkey identifier

**File:** [`Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationPasteboardSupport.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationPasteboardSupport.swift#L20-L22)

Selection Translation now uses the free Quick Tool hotkey ID `21`.

### Source and translation labels

**File:** [`Sources/Vorssaint/UI/SelectionTranslationPanel.swift`](https://github.com/vorssaintapp/vorssaint-utils/blob/504114b79ef20ff94e70905f4c00f6d765dc5cd2/Sources/Vorssaint/UI/SelectionTranslationPanel.swift#L207-L318)

The cards now use “Source text” and “Translation” rather than the source/target language setting labels.

## Verification

- `./build.sh` — passed without Swift compiler warnings.
- `./build/Vorssaint --selftest` — `SELFTEST OK`.
- `./build.sh --test` — `TESTS OK (8969 checks)`.
- The Release Stage bundle passed `codesign --verify --deep --strict`.
- `git diff --check` passed.
- Searches for the removed symbols and legacy Tom configuration fallbacks returned no matches.

Manual TextEdit verification of real AX selection and held-modifier behavior is still pending before this draft is marked ready for review.

PathGao:
Six in, and the shape is right: one `snapshot(of:)`, read before present, defaults with one home.

**1. The retry limit changed meaning, and it isn't only yours.** `main`: `if held.isEmpty || attempt >= 100 { …post… }` — exhausting it posts anyway. Now: `if attempt >= 100 { completion(false); return }`. `PastePlainService` passes no `didFail`, so holding the chord past ~1.5s turns paste-as-plain-text into a silent no-op where it used to paste; snippet expansion via `postExpansion` shares the path. Neither is in the body. Keep the old fallthrough for V, or say why V should change too.

**2. `snapshot()` traded a duplicate default for a crash.** `preconditionFailure` now runs on every shortcut press, and `Sources/` has exactly three `fatalError`s, all of them `init(coder:)`. `?? SelectionTranslationSettingsStore.defaultModel` is still one source of truth — the constant is the truth.

**3. The history guard is armed after the copy.** Every other caller pre-arms: write, `ignoreNextChange(upTo:)`, post. The reader can't know the count ahead, so it polls at 20ms and arms after; the history timer is 0.8s, so a tick inside that window captures the selection.

The panel and settings files are past where my diff cuts off — the styleMask and the footer are your word, not mine. Hotkey 21 I did check: free on `main`.

**Impact:** paste-as-plain-text and snippet paste expansion silently do nothing when a modifier stays down past ~1.5s. Both ship today.
**Merge:** needs 1 first.

ref #489

zhanghk2001:
@PathGao Follow-up review items are addressed in `dd7e749989af91105dc41ad2b9eaf2d8a839b117`:

1. `TransientPaste` keeps the legacy `Command-V` post-on-timeout behavior, while selection `Command-C` now fails safely if modifiers stay held. The decision is covered by `releaseDecision` and the separate timeout behavior at [TransientPaste.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/dd7e749989af91105dc41ad2b9eaf2d8a839b117/Sources/Vorssaint/Services/TransientPaste.swift#L139-L195).
2. Missing selection-translation defaults now fall back to the `SelectionTranslationSettingsStore.default*` constants instead of crashing at startup. See [SelectionTranslationSettingsStore.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/dd7e749989af91105dc41ad2b9eaf2d8a839b117/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSettingsStore.swift#L77-L99).
3. Clipboard-history deferral is armed before synthetic `Command-C` and released after guarded restoration, so temporary selection data is not recorded while newer user copies remain eligible. See [SelectionTranslationSelectionReader.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/dd7e749989af91105dc41ad2b9eaf2d8a839b117/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift#L49-L106) and [ClipboardHistoryService.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/dd7e749989af91105dc41ad2b9eaf2d8a839b117/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift#L102-L122).

Verification: `./build.sh --test` passed 9342 checks; Release build, Stage selftest, codesign verification, diff check, and latest-main ancestry check passed. The PR remains Ready for review.

PathGao:
Three landed and the shape is right: `releaseDecision` splits the V and C policies, the missing default falls back to the constant, the deferral arms in `willPost`.

**1. The merge reverted `issue-fix-status.yml`.** `main` has `SKIP_LABELS: '["summary"]'` plus `HALF_LABEL: half` and the tailored release-sweep comment; this diff deletes both and puts `half` back in the skip set, so a `half` issue loses its release comment and its two-week timer. The body mentions only the merge. GitHub reports the PR conflicting now, so it needs redoing anyway — leave the workflow out of it.

**2. `endCaptureDeferral` drops the ignore while another deferral is open.** `guard captureDeferralState.end(token) else { return }` returns before `ignoreNextChange(upTo:)`, and `end` is false for every non-final token. Under the nesting the type exists for, the synthetic change count is never ignored and the outer release captures it. Ignore first, guard on the result after.

**3. Nothing releases the deferral except the completion.** One path that calls `willPost` and never completes leaves history capture off for the rest of the session. `postKeyWhenModifiersReleased` is past where my diff cuts off, as are the panel files — that part is your word, not mine.

**Impact:** `half` issues lose their release sweep; clipboard history mis-records under a nested deferral.
**Merge:** needs the re-merge without the workflow revert first.

ref #489

zhanghk2001:
@PathGao Follow-up fixes are now pushed in `b44b99b5be691621aec350d660b0569246d89960`.

- Re-merged the latest `main` (`16216d5`) and kept `.github/workflows/issue-fix-status.yml` unchanged from `main`, including `HALF_LABEL` and the tailored release sweep.
- `ClipboardHistoryCaptureDeferralState.end` now distinguishes invalid, nested, and final tokens. Every valid token applies its ignore count before only the final token resumes capture: [ClipboardHistorySupport.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/b44b99b5be691621aec350d660b0569246d89960/Sources/Vorssaint/Services/Clipboard/ClipboardHistorySupport.swift#L20-L40) and [ClipboardHistoryService.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/b44b99b5be691621aec350d660b0569246d89960/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift#L113-L127).
- Selection fallback deferral is now idempotent and has a five-second watchdog; late callbacks cannot release or resume twice: [SelectionTranslationSelectionReader.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/b44b99b5be691621aec350d660b0569246d89960/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift#L20-L42) and [SelectionTranslationSelectionReader.swift](https://github.com/vorssaintapp/vorssaint-utils/blob/b44b99b5be691621aec350d660b0569246d89960/Sources/Vorssaint/Services/SelectionTranslation/SelectionTranslationSelectionReader.swift#L69-L91).
- PR description no longer uses `Refs #489`; selected-text translation remains separate from the screenshot OCR/translation scope in issue #489.

Verification: `./build.sh --test` passed 9522 checks; Release build, Stage selftest, codesign verification, diff check, workflow comparison, and latest-main ancestry check passed. The PR is mergeable again.

zhanghk2001:
@PathGao Follow-up: the branch now also includes the latest `upstream/main` through merge commit `f4d796672549f1c4e3b81d700eed133aa84ab0f1`. The issue-status workflow remains identical to `main`; the nested deferral and one-shot watchdog fixes from `b44b99b` are unchanged. Final verification after this merge: `./build.sh --test` passed 9540 checks, Release build passed, Stage selftest returned `SELFTEST OK`, codesign and diff checks passed. The PR description now explicitly keeps issue #489 separate.

zhanghk2001:
CI status report:

- GitHub currently reports `MERGEABLE` with no merge conflict.
- The PR merge state is `BLOCKED` because the required `Build & selftest` and `Swift 6.0.3 compatibility` checks have not reported a status for commit `f4d796672549f1c4e3b81d700eed133aa84ab0f1`.
- The commit status API is still `pending` with zero reported statuses; the only workflow check currently present is the reference-hint job, which passed.
- Local verification is complete: `./build.sh --test` passed 9540 checks, Release build/selftest/signature checks passed.

This is a CI/check-reporting gate, not a source merge conflict.

PathGao:
Workflow revert is gone — no `.github/` hunk in the diff at all — and `end` now separates nested from final.

**1. `guard captureDeferralBox.token != nil else { return }` eats every failure that lands before `willPost`.** `completion(false)` fires without ever reaching `willPost` on three paths: `.fail` after ~1.5s of held modifiers (the behavior this call site asked for), Secure Input, and CGEvent creation. Token is still nil, the guard returns, and nothing resumes the continuation until the 5s watchdog — with `phase` stuck at `.reading`, so `trigger()` drops a second press too. Hold ⌃⌥⌘Y a beat too long and the shortcut looks dead for five seconds. `finish()` is already the idempotency gate; guard on it instead.

**2. The watchdog is the only release path that can't ignore the synthetic change count.** `finish()` with no argument → `endCaptureDeferral(token, ignoringUpTo: nil)` → `.releasedLast` → straight to `captureIfChanged()`. If it fires after the ⌘C went out — the pasteboard lane can sit behind a password prompt for seconds, per its own comment — the selection lands in clipboard history, the thing the deferral exists to prevent. Release the deferral from the reader block, not from whoever resumes first.

`paste()` has no hunk, so the V shim is below my 80k cut — post-on-timeout is your word, as are the panel and settings files.

**Impact:** the shortcut looks dead for 5s when a modifier stays down or Secure Input is on; a slow pasteboard lane records the selection into clipboard history.
**Merge:** needs 1 first.

ref #489

zhanghk2001:
Follow-up implementation update (commit `751f215`):

- Added press/release handling for the selection-translation shortcut. The panel appears immediately in a locked non-activating state.
- Added the 1.5-second hold limit: when Accessibility (AX) has already returned selected text, translation starts at the limit without waiting for key release.
- When AX cannot provide text by the limit, the hold timer stops and the panel explains that the shortcut must be released before the safe Command-C fallback runs. This is an AX limitation and should be rare; it prevents sending Command-C while the trigger modifiers are still held.
- Captured the original target process so switching apps during the hold cannot copy text from the new app.
- AX and pasteboard paths are separated, with explicit failure handling and idempotent generation/cancellation checks.

This also preserves the reviewer-requested fixes already in the branch: legacy Command-V timeout behavior, safe selection Command-C timeout behavior, single-source default fallback, pre-armed/nested clipboard-history deferral cleanup, unchanged `issue-fix-status.yml`, and no automatic association with issue #489.

Verification on the pushed commit:
- `./build.sh --test` — `TESTS OK (9552 checks)`
- Release build — passed
- Stage selftest — `SELFTEST OK`
- `codesign --verify --deep --strict build/stage/Vorssaint.app` — passed
- `git diff --check` — passed

The PR remains open and Ready for review; no formal App or production configuration was changed.

PathGao:
Three landed: pre-`willPost` failures resume, the watchdog can't end the deferral. The press/release split reads clean.

**1. The unlock is conditioned on a phase that may have moved on.** `shortcutReleased` unlocks only for `.translating`/`.streaming`; every other phase falls to `handleShortcutAction(shortcutFlow.shortcutReleased())`, which is `.none` once `actionClaimed` is set — so nothing unlocks. Hold ⌃⌥⌘Y with no API key: at 1.5s `.translate` fires, the provider init throws `.emptyAPIKey` in a millisecond, phase is `.failed` before you let go, and the panel sits locked on "Enter an API key…" with its own settings button unclickable. Unlock unconditionally on release. (`setInteractionLocked` is past my 80k cut — I'm reading the service side.)

**2. The release poll has no cap.** `while !Task.isCancelled { flagsState … 20ms }` never gives up; `postPasteWhenModifiersReleased` caps at 100 × 15ms and proceeds anyway. A latched modifier keeps `shortcutIsHeld` true, and nothing else can leave `.waitingForShortcutRelease`.

**3.** `"The target application changed. Press the shortcut again."` is the one user-facing string not in `SelectionTranslationFeatureStrings`.

Dropping the `ref #489` I kept adding — screenshot OCR is a different feature, you're right.

**Impact:** a first run with no key configured leaves the panel locked and unusable; a latched modifier hangs the feature.
**Merge:** needs 1 first.

zhanghk2001:
Follow-up fixes pushed in `df034fd`:

1. Panel unlocking is now unconditional after the shortcut release is confirmed. This covers failure states such as a missing API key; the settings and retry actions remain usable. The pasteboard fallback re-locks the panel during its protected read and unlocks it when the read finishes.
2. Release-state polling is bounded to 100 attempts at 15 ms (about 1.5 seconds). A stuck modifier no longer leaves the feature waiting forever; it exits safely without synthesizing `Command-C`.
3. Target-application-change and release-timeout errors now use `SelectionTranslationFeatureStrings` instead of hard-coded English.

The existing reviewer fixes remain intact: legacy Command-V timeout behavior, safe selection Command-C timeout behavior, one-source defaults, nested/idempotent clipboard-history deferral cleanup, unchanged `issue-fix-status.yml`, and no `#489` automatic association.

Verification on `df034fd`:
- `./build.sh --test` — `TESTS OK (9557 checks)`
- Release build — passed
- Stage selftest — `SELFTEST OK`
- `codesign --verify --deep --strict build/stage/Vorssaint.app` — passed
- `git diff --check` — passed

The PR remains open and Ready for review.

zhanghk2001:
@PathGao Sorry, I forgot to include the source link for Tom.

Tom is an open-source, locally running text-selection translation app that I developed myself. It stores API credentials in the macOS Keychain and supports user-configured OpenAI-compatible translation providers.

[Tom source repository](https://github.com/zhanghk2001/Tom)

PR #1104 migrates Tom’s original selected-text translation workflow into Vorssaint. Tom was used as the functional reference for the initial migration.

The subsequent interaction, UI, and reliability improvements were implemented and tested locally in Vorssaint after the migration. They are based on testing the Vorssaint implementation, not on later changes from Tom. The current code follows Vorssaint’s own architecture and visual style.

PathGao:
Three from last round landed. Tom being yours settles the provenance — and it shows in the diff: two of these three are Tom's defaults surviving the port.

**1. The AX read falls back to whoever is frontmost once the captured target is gone.** In `readSelectedText`: `processIdentifier.flatMap { NSRunningApplication(processIdentifier: $0) } ?? NSWorkspace.shared.frontmostApplication`. The target quits or crashes mid-hold, the lookup returns nil, and it reads app B's selection instead. The pasteboard path guards on `frontmost == target`; the `??` undoes that guard on the AX path. Return `""` when a pid was passed and doesn't resolve.

**2. The footer's token count is always the estimator's guess.** `SelectionTranslationRequestBodyBuilder.body` never sets `stream_options.include_usage`, so on `api.openai.com` — the registered default base URL — no chunk carries `usage` and `translate()` always falls through to `SelectionTranslationTokenEstimator`.

**3. `{{from}}` renders as `自动检测`.** `makeRequest` reads `displayName` and the default source is `.automatic`, so an en-US user's prompt says "from 自动检测 into English". `qwenTranslationName` already holds the English names.

Panel and settings files are past my 80k cut — the styleMask and the footer are still your word.

**Impact:** after the target app quits mid-hold, the AX path can translate a different app's selection.
**Merge:** needs 1 first.

PathGao:
Closing this on direction, and I want to separate that from the review above, which is about code and which you have answered every round within hours.

The feature is translation through a user-supplied OpenAI-compatible endpoint: `defaultBaseURL = "https://api.openai.com/v1"`, a Bearer key in the Keychain, `chat/completions` as the only path. There is no on-device half. That is the "client that has to exist on the other end of the wire" docs/AI-CONTRIBUTIONS.md names, and the project decided this week not to take on features whose value depends on a third-party service and the user's account with it (#1201 was closed on the same ground; #611, user-supplied cloud credentials, was an explicit no earlier). The maintainer runs and supports every feature he ships, and a translation that stops working when a provider changes its API or a key expires is a support load that lands here.

Thanks for the provenance note on Tom. It stands well on its own as a separate app, and porting it here carefully was real work; the pasteboard handling is now correct and the review rounds were tight. macOS has a first-party Translation framework (macOS 15+); if a selected-text translation that stands on that is worth having, an issue settling that comes first, and #773 already uses it for a different surface. Keep the branch.

PathGao:
For the record, a few more things from reading the branch and the Tom repository, so they are on the thread if this comes back in another form.

**Provenance.** Tom was created on 2026-08-28, the day this PR opened, with one push, a single contributor and no users. It is the same code in a second wrapper, not a project with history behind it, so "migrated from a tested app" does not add evidence the PR does not already carry. Its `docs/superpowers/` plans and specs are the same generated planning artifacts #1082 was asked to remove. None of this is a licence problem (MIT, sole author, GPL contribution is fine); it is that the description claims maturity the code has not had.

**Two copies of one codebase.** Tom keeps evolving under you; the copy here would be the maintainer's to carry. Fixes made in one do not reach the other.

**Shared machinery.** A feature PR that changes `ClipboardHistoryService` (+30, a new capture-deferral token mechanism), `ClipboardHistorySupport` (+37), `TransientPaste` (+81, new public entry points), `GlobalShortcut` and `QuickToolHotkey` has a regression surface well beyond translation, and none of it has run on a maintainer's Mac or in CI on this head.

**Smaller.** `SelfUninstall.swift` removes the Keychain item but `Tools/uninstall.sh` does not, so the script route leaves the key behind. `SelectionTranslationStrings.swift` is 125 lines, which cannot hold thirteen catalogs; if it is English plus 简体中文 with a fallback, the build rule on translations applies. And #773 already carries a translation feature on Apple's framework; two translation features with two settings pages is a product question before it is a code one.

Where the endpoint and key are the user's own configuration, that part is their responsibility and not a mark against the branch.

Issue #1083, opened 2026-08-27
Expand Shelf Support for All Objects (Files and Emails from Outlook)


hoorrus:
### Problem or Motivation

Hi, I see your Shelf Support works a little bit more reliably, but very often I have to attach an email or an attachment in Outlook driectly to another place. 

### Proposed Solution

Expand the Shelf Support to handle .eml files from Outlook and attachments directly from Outlook

### Feature Area

Clipboard and files

### Does another app already do this?

Dropover does this.

PR #1106, opened 2026-08-28
feat(Shelf): Support Outlook emails and attachments in Shelf

muaz978:
## Summary

- Add Shelf support for AppKit file promises exposed by Outlook and Mail.
- Materialize promised `.eml` messages and attachments asynchronously, then copy them into Shelf's owner-only file store as normal file items.
- Preserve original filenames as Shelf tile titles and support dropping promised files onto existing Shelf piles.

## Review follow-up

- All `NSFilePromiseReceiver` instances from one drag use AppKit's required shared staging destination. The fallback coordinator tracks the advertised callback count, handles callbacks that arrive during registration, and keeps direct representations pending until the promise batch reaches a terminal outcome.
- AppKit receiver errors and Shelf copy errors now have different accounting. An AppKit error closes the affected receiver wholesale so an unavailable receiver cannot wedge the batch. A `storePromisedFile` failure credits only that callback, allowing Outlook's later attachments from the same receiver to arrive normally. This prevents a message-subject fallback tile from appearing beside attachments that did succeed.
- A late materialized file remains accepted after a wholesale receiver failure or after direct fallback was selected. Direct fallback also marks the failure handled, so that late success cannot re-arm the failure HUD.
- The Shelf panel and docked pill now use the shared AppKit `WindowMoveHandle` drop destination. Both preserve the original dragging pasteboard and reach `accept(pasteboard:)`, which constructs `NSFilePromiseReceiver`. The former SwiftUI `NSItemProvider.loadFileRepresentation` route, including its placeholder fallback behavior, has been removed so Outlook's subject text cannot be accepted as a substitute tile on those surfaces.
- Promise-backed drops treat promised files as authoritative. Auxiliary text and URL flavors are held as a fallback only, and are discarded when any promised file succeeds. Ordinary file, image, URL, text, GIF, and pile-merge drops continue to use the same AppKit pasteboard path.
- Promise-copy cleanup remains limited to regular files inside the staging directory. Error callbacks never remove an AppKit-supplied directory, absent URL, or partial path.
- Removed the obsolete provider-batch helper and its tests with the retired SwiftUI route, leaving one canonical promise implementation.

## Verification

- Rebased onto current `main` at `ed882cd` and verified GitHub reports `behind_by: 0`.
- `./build.sh` passed after the rebase, including protected fan-helper compilation, app bundle assembly, and ad-hoc signing.
- `./build.sh --test` compiled and ran 9,498 checks. The new promise-coordinator and drop-route regressions passed. Eight existing host-dependent checks remain unavailable in this environment: WhatsApp bucketing, non-PNG image conversion, in-process AppleScript, JPEG/movie tool routing, and external-media staging and byte-preservation tests.
- `git diff --check` passed. The final diff contains no credentials or generated artifacts.

Live Outlook or Mail drags were not available in this environment. A maintainer should perform one drag with multiple attachments onto both the expanded Shelf panel and docked pill before merge. That provider-specific manual check is distinct from the successful local build and regression coverage above.

Refs #1083

### DIFF
--- a/Sources/Vorssaint/Core/Localization.swift
+++ b/Sources/Vorssaint/Core/Localization.swift
@@ -1079,6 +1079,7 @@ struct Strings {
     let switcherCurrentSpaceOnly: String
     let switcherCurrentSpaceOnlyCaption: String
     let shelfFileMissing: String
+    let shelfDropFailed: String
     let previewSizeSmall: String
     let mixerSoundEffectsOutputTitle: String
     let mixerSoundEffectsOutputTooltip: String
@@ -2064,6 +2065,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Mostrar só a Mesa atual",
         switcherCurrentSpaceOnlyCaption: "Mostra no alternador apenas as janelas da Mesa em que você está. Escolher uma janela nunca leva você para outra Mesa.",
         shelfFileMissing: "O arquivo não existe mais",
+        shelfDropFailed: "Não foi possível adicionar o item arrastado à área temporária.",
         previewSizeSmall: "Pequeno",
         mixerSoundEffectsOutputTitle: "Sons do sistema",
         mixerSoundEffectsOutputTooltip: "Escolher onde alertas e efeitos sonoros tocam",
@@ -3050,6 +3052,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Show only the current desktop",
         switcherCurrentSpaceOnlyCaption: "Lists only windows from the desktop you are on. Picking a window never moves you to another desktop.",
         shelfFileMissing: "The file no longer exists",
+        shelfDropFailed: "Could not add the dropped item to the shelf.",
         previewSizeSmall: "Small",
         mixerSoundEffectsOutputTitle: "System sounds",
         mixerSoundEffectsOutputTooltip: "Choose where alerts and sound effects play",
--- a/Sources/Vorssaint/Core/Localizations/Strings+ChineseSimplified.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+ChineseSimplified.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "仅显示当前桌面",
         switcherCurrentSpaceOnlyCaption: "切换器中只显示你所在桌面的窗口。选择窗口不会把你带到其他桌面。",
         shelfFileMissing: "文件已不存在",
+        shelfDropFailed: "无法将拖放的项目添加到暂存架。",
         previewSizeSmall: "小",
         mixerSoundEffectsOutputTitle: "系统声音",
         mixerSoundEffectsOutputTooltip: "选择提醒和音效的播放设备",
--- a/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalHK.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalHK.swift
@@ -936,6 +936,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "僅顯示目前的桌面",
         switcherCurrentSpaceOnlyCaption: "切換器中只顯示你所在桌面的視窗。選擇視窗不會帶你前往其他桌面。",
         shelfFileMissing: "檔案已不存在",
+        shelfDropFailed: "無法將拖放的項目加入暫存架。",
         previewSizeSmall: "小",
         mixerSoundEffectsOutputTitle: "系統聲音",
         mixerSoundEffectsOutputTooltip: "選擇提示音和音效的播放裝置",
--- a/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalTW.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalTW.swift
@@ -936,6 +936,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "僅顯示目前的桌面",
         switcherCurrentSpaceOnlyCaption: "切換器中只顯示你所在桌面的視窗。選擇視窗不會把你帶到其他桌面。",
         shelfFileMissing: "檔案已不存在",
+        shelfDropFailed: "無法將拖放的項目加入暫存架。",
         previewSizeSmall: "小",
         mixerSoundEffectsOutputTitle: "系統聲音",
         mixerSoundEffectsOutputTooltip: "選擇提示音與音效的播放裝置",
--- a/Sources/Vorssaint/Core/Localizations/Strings+French.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+French.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Afficher uniquement le bureau actuel",
         switcherCurrentSpaceOnlyCaption: "N'affiche dans le sélecteur que les fenêtres du bureau où vous êtes. Choisir une fenêtre ne vous emmène jamais sur un autre bureau.",
         shelfFileMissing: "Le fichier n'existe plus",
+        shelfDropFailed: "Impossible d’ajouter l’élément déposé à l’étagère.",
         previewSizeSmall: "Petite",
         mixerSoundEffectsOutputTitle: "Sons du système",
         mixerSoundEffectsOutputTooltip: "Choisir où sont diffusés les alertes et effets sonores",
--- a/Sources/Vorssaint/Core/Localizations/Strings+German.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+German.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Nur den aktuellen Schreibtisch anzeigen",
         switcherCurrentSpaceOnlyCaption: "Zeigt im Umschalter nur Fenster des Schreibtischs, auf dem du gerade bist. Die Auswahl eines Fensters wechselt nie zu einem anderen Schreibtisch.",
         shelfFileMissing: "Die Datei existiert nicht mehr",
+        shelfDropFailed: "Das abgelegte Element konnte nicht zur Ablage hinzugefügt werden.",
         previewSizeSmall: "Klein",
         mixerSoundEffectsOutputTitle: "Systemtöne",
         mixerSoundEffectsOutputTooltip: "Auswählen, wo Hinweise und Toneffekte wiedergegeben werden",
--- a/Sources/Vorssaint/Core/Localizations/Strings+Italian.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+Italian.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Mostra solo la scrivania attuale",
         switcherCurrentSpaceOnlyCaption: "Mostra nel commutatore solo le finestre della scrivania in cui ti trovi. Scegliere una finestra non ti porta mai su un'altra scrivania.",
         shelfFileMissing: "Il file non esiste più",
+        shelfDropFailed: "Impossibile aggiungere l’elemento trascinato al ripiano.",
         previewSizeSmall: "Piccola",
         mixerSoundEffectsOutputTitle: "Suoni di sistema",
         mixerSoundEffectsOutputTooltip: "Scegli dove riprodurre avvisi ed effetti sonori",
--- a/Sources/Vorssaint/Core/Localizations/Strings+Japanese.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+Japanese.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "現在のデスクトップのみ表示",
         switcherCurrentSpaceOnlyCaption: "今いるデスクトップのウインドウだけをスイッチャーに表示します。ウインドウを選んでも別のデスクトップには移動しません。",
         shelfFileMissing: "ファイルはもう存在しません",
+        shelfDropFailed: "ドロップした項目をシェルフに追加できませんでした。",
         previewSizeSmall: "小",
         mixerSoundEffectsOutputTitle: "システムサウンド",
         mixerSoundEffectsOutputTooltip: "通知音とサウンドエフェクトの出力先を選択",
--- a/Sources/Vorssaint/Core/Localizations/Strings+Korean.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+Korean.swift
@@ -936,6 +936,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "현재 데스크탑만 표시",
         switcherCurrentSpaceOnlyCaption: "지금 있는 데스크탑의 윈도우만 전환기에 표시합니다. 윈도우를 선택해도 다른 데스크탑으로 이동하지 않습니다.",
         shelfFileMissing: "파일이 더 이상 존재하지 않습니다",
+        shelfDropFailed: "드롭한 항목을 선반에 추가할 수 없습니다.",
         previewSizeSmall: "작게",
         mixerSoundEffectsOutputTitle: "시스템 사운드",
         mixerSoundEffectsOutputTooltip: "알림 및 사운드 효과를 재생할 출력 선택",
--- a/Sources/Vorssaint/Core/Localizations/Strings+Russian.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+Russian.swift
@@ -936,6 +936,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Показывать только текущий рабочий стол",
         switcherCurrentSpaceOnlyCaption: "В переключателе видны только окна рабочего стола, на котором вы находитесь. Выбор окна никогда не переносит вас на другой рабочий стол.",
         shelfFileMissing: "Файл больше не существует",
+        shelfDropFailed: "Не удалось добавить перетащенный элемент на полку.",
         previewSizeSmall: "Маленький",
         mixerSoundEffectsOutputTitle: "Системные звуки",
         mixerSoundEffectsOutputTooltip: "Выбрать устройство для оповещений и звуковых эффектов",
--- a/Sources/Vorssaint/Core/Localizations/Strings+Spanish.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+Spanish.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Mostrar solo el escritorio actual",
         switcherCurrentSpaceOnlyCaption: "Muestra en el selector solo las ventanas del escritorio en el que estás. Elegir una ventana nunca te lleva a otro escritorio.",
         shelfFileMissing: "El archivo ya no existe",
+        shelfDropFailed: "No se pudo añadir el elemento arrastrado al estante.",
         previewSizeSmall: "Pequeño",
         mixerSoundEffectsOutputTitle: "Sonidos del sistema",
         mixerSoundEffectsOutputTooltip: "Elegir dónde se reproducen los avisos y efectos de sonido",
--- a/Sources/Vorssaint/Core/Localizations/Strings+Turkish.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+Turkish.swift
@@ -935,6 +935,7 @@ extension Strings {
         switcherCurrentSpaceOnly: "Yalnızca geçerli masaüstünü göster",
         switcherCurrentSpaceOnlyCaption: "Değiştiricide yalnızca bulunduğunuz masaüstündeki pencereler listelenir. Bir pencere seçmek sizi asla başka bir masaüstüne taşımaz.",
         shelfFileMissing: "Dosya artık yok",
+        shelfDropFailed: "Bırakılan öğe rafa eklenemedi.",
         previewSizeSmall: "Küçük",
         mixerSoundEffectsOutputTitle: "Sistem sesleri",
         mixerSoundEffectsOutputTooltip: "Uyarıların ve ses efektlerinin çalacağı çıkışı seç",
--- a/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+
+/// Coordinates direct pasteboard fallback content while AppKit fulfills a
+/// batch of file promises. A receiver's reader callback is per promised file,
+/// not per receiver, so completion must be counted against the file names
+/// reported after each receiver is activated.
+final class ShelfPromiseFallbackCoordinator<Value> {
+    struct CallbackOutcome {
+        let discardFallback: [Value]?
+        let fallback: [Value]?
+        let reportFailure: Bool
+    }
+
+    let items: [Value]
+    private let lock = NSLock()
+    private let receiverCount: Int
+    private var registeredReceiverCount = 0
+    private var expectedFileCount = 0
+    private var completedFileCount = 0
+    private var hasSuccess = false
+    private var hasFailure = false
+    private var fallbackConsumed = false
+    private var failureReported = false
+    private var registrationFinished = false
+
+    init(items: [Value], receiverCount: Int) {
+        self.items = items
+        self.receiverCount = receiverCount
+    }
+
+    func registerReceiver(expectedFileCount: Int) {
+        lock.lock()
+        registeredReceiverCount += 1
+        // Keep one defensive slot for an unusual receiver that does not
+        // publish names, so a failed receiver cannot leave the batch waiting
+        // forever. Normal receivers use their exact fileNames.count.
+        self.expectedFileCount += max(expectedFileCount, 1)
+        lock.unlock()
+    }
+
+    func finishRegistration() -> CallbackOutcome? {
+        lock.lock()
+        defer { lock.unlock() }
+        registrationFinished = true
+        return completedOutcomeIfReady()
+    }
+
+    func recordSuccess() -> CallbackOutcome {
+        lock.lock()
+        defer { lock.unlock() }
+        completedFileCount += 1
+        hasSuccess = true
+        let discardFallback = consumeFallbackIfNeeded()
+        return CallbackOutcome(discardFallback: discardFallback,
+                               fallback: nil,
+                               reportFailure: reportFailureIfReady())
+    }
+
+    func recordFailure() -> CallbackOutcome {
+        lock.lock()
+        defer { lock.unlock() }
+        completedFileCount += 1
+        hasFailure = true
+        guard let outcome = completedOutcomeIfReady() else {
+            return CallbackOutcome(discardFallback: nil,
+                                   fallback: nil,
+                                   reportFailure: false)
+        }
+        return outcome
+    }
+
+    private func consumeFallbackIfNeeded() -> [Value]? {
+        guard !fallbackConsumed else { return nil }
+        fallbackConsumed = true
+        return items.isEmpty ? nil : items
+    }
+
+    private func completedOutcomeIfReady() -> CallbackOutcome? {
+        guard registrationFinished,
+              registeredReceiverCount == receiverCount,
+              completedFileCount >= expectedFileCount else { return nil }
+        if hasSuccess {
+            return CallbackOutcome(discardFallback: nil,
+                                   fallback: nil,
+                                   reportFailure: reportFailureIfReady())
+        }
+        guard !fallbackConsumed else {
+            return CallbackOutcome(discardFallback: nil,
+                                   fallback: nil,
+                                   reportFailure: reportFailureIfReady())
+        }
+        fallbackConsumed = true
+        if items.isEmpty {
+            return CallbackOutcome(discardFallback: nil,
+                                   fallback: nil,
+                                   reportFailure: reportFailureIfReady())
+        }
+        return CallbackOutcome(discardFallback: nil,
+                               fallback: items,
+                               reportFailure: false)
+    }
+
+    private func reportFailureIfReady() -> Bool {
+        guard registrationFinished,
+              registeredReceiverCount == receiverCount,
+              completedFileCount >= expectedFileCount,
+              hasFailure,
+              !failureReported else { return false }
+        failureReported = true
+        return true
+    }
+}
--- a/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
@@ -100,7 +100,9 @@ final class ShelfPromiseFallbackCoordinator<Value> {
                                reportFailure: reportFailureIfReady())
     }
 
-    func recordFailure(for receiverID: Int) -> CallbackOutcome {
+    /// Records an AppKit error that terminated a receiver before it could
+    /// deliver its remaining promised files.
+    func recordReceiverFailure(for receiverID: Int) -> CallbackOutcome {
         lock.lock()
         defer { lock.unlock() }
         guard var receiver = receivers[receiverID], !receiver.isWholesaleFailure else {
@@ -122,6 +124,41 @@ final class ShelfPromiseFallbackCoordinator<Value> {
         receiver.completedFileCount = receiver.expectedFileCount
         receiver.isTerminal = true
         receiver.isWholesaleFailure = true
+        receivers[receiverID] = receiver
+        hasFailure = true
+        guard let outcome = completedOutcomeIfReady() else {
+            return CallbackOutcome(acceptFile: false,
+                                   discardFallback: nil,
+                                   fallback: nil,
+                                   reportFailure: false)
+        }
+        return CallbackOutcome(acceptFile: false,
+                               discardFallback: outcome.discardFallback,
+                               fallback: outcome.fallback,
+                               reportFailure: outcome.reportFailure)
+    }
+
+    /// Records one file that arrived from an otherwise healthy receiver but
+    /// could not be copied into the Shelf store. This must not terminate the
+    /// receiver: AppKit can still deliver its later attachments.
+    func recordFileFailure(for receiverID: Int) -> CallbackOutcome {
+        lock.lock()
+        defer { lock.unlock() }
+        guard var receiver = receivers[receiverID], !receiver.isWholesaleFailure else {
+            return CallbackOutcome(acceptFile: false,
+                                   discardFallback: nil,
+                                   fallback: nil,
+                                   reportFailure: false)
+        }
+        if receiver.isTerminal {
+            receiver.expectedFileCount = receiver.completedFileCount + 1
+            receiver.isTerminal = false
+        }
+        receiver.completedFileCount += 1
+        if receiver.isExpectedFileCountFinal,
+           receiver.completedFileCount >= receiver.expectedFileCount {
+            receiver.isTerminal = true
+        }
         receivers[receiverID] = receiver
         hasFailure = true
         guard let outcome = completedOutcomeIfReady() else {
--- a/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
@@ -69,16 +69,16 @@ final class ShelfPromiseFallbackCoordinator<Value> {
     func recordSuccess(for receiverID: Int) -> CallbackOutcome {
         lock.lock()
         defer { lock.unlock() }
-        guard var receiver = receivers[receiverID], !receiver.isWholesaleFailure else {
-            // A prior error closes a receiver wholesale. Do not let a late
-            // success create a promised tile after the direct fallback was
-            // selected, but still let the caller clean a regular file safely.
+        guard var receiver = receivers[receiverID] else {
             return CallbackOutcome(acceptFile: false,
                                    discardFallback: nil,
                                    fallback: nil,
                                    reportFailure: false)
         }
-        if receiver.isTerminal {
+        // A wholesale error may be reported before the receiver's later
+        // promised files materialize. Keep those files instead of deleting
+        // them, even if direct fallback has already been queued.
+        if !receiver.isWholesaleFailure, receiver.isTerminal {
             // `fileNames` is normally authoritative, but the defensive
             // minimum of one cannot cap a provider that omits its names and
             // then delivers more than one file. Treat the extra callback as
@@ -87,7 +87,8 @@ final class ShelfPromiseFallbackCoordinator<Value> {
             receiver.isTerminal = false
         }
         receiver.completedFileCount += 1
-        if receiver.isExpectedFileCountFinal,
+        if !receiver.isWholesaleFailure,
+           receiver.isExpectedFileCountFinal,
            receiver.completedFileCount >= receiver.expectedFileCount {
             receiver.isTerminal = true
         }
--- a/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
@@ -69,7 +69,7 @@ final class ShelfPromiseFallbackCoordinator<Value> {
     func recordSuccess(for receiverID: Int) -> CallbackOutcome {
         lock.lock()
         defer { lock.unlock() }
-        guard var receiver = receivers[receiverID], !receiver.isTerminal else {
+        guard var receiver = receivers[receiverID], !receiver.isWholesaleFailure else {
             // A prior error closes a receiver wholesale. Do not let a late
             // success create a promised tile after the direct fallback was
             // selected, but still let the caller clean a regular file safely.
@@ -77,6 +77,14 @@ final class ShelfPromiseFallbackCoordinator<Value> {
                                    discardFallback: nil,
                                    fallback: nil,
                                    reportFailure: false)
+        }
+        if receiver.isTerminal {
+            // `fileNames` is normally authoritative, but the defensive
+            // minimum of one cannot cap a provider that omits its names and
+            // then delivers more than one file. Treat the extra callback as
+            // evidence that the expected count needs to grow.
+            receiver.expectedFileCount = receiver.completedFileCount + 1
+            receiver.isTerminal = false
         }
         receiver.completedFileCount += 1
         if receiver.isExpectedFileCountFinal,
@@ -94,11 +102,18 @@ final class ShelfPromiseFallbackCoordinator<Value> {
     func recordFailure(for receiverID: Int) -> CallbackOutcome {
         lock.lock()
         defer { lock.unlock() }
-        guard var receiver = receivers[receiverID], !receiver.isTerminal else {
+        guard var receiver = receivers[receiverID], !receiver.isWholesaleFailure else {
             return CallbackOutcome(acceptFile: false,
                                    discardFallback: nil,
                                    fallback: nil,
                                    reportFailure: false)
+        }
+        if receiver.isTerminal {
+            // Apply the same overflow rule to a late failure. An omitted or
+            // stale file count must not turn a real failed file into silent
+            // loss merely because an earlier callback reached the floor.
+            receiver.expectedFileCount = receiver.completedFileCount + 1
+            receiver.isTerminal = false
         }
         // An error callback may be the receiver's wholesale failure rather
         // than one failed file. Close the receiver and credit every remaining
--- a/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
@@ -165,6 +165,10 @@ final class ShelfPromiseFallbackCoordinator<Value> {
                                    fallback: nil,
                                    reportFailure: reportFailureIfReady())
         }
+        // Direct fallback is the user-visible recovery for a completely
+        // failed promise batch. Mark that failure handled so a later file
+        // from the same receiver cannot re-arm the suppressed notification.
+        failureReported = true
         return CallbackOutcome(acceptFile: false,
                                discardFallback: nil,
                                fallback: items,
--- a/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfPromiseFallbackCoordinator.swift
@@ -4,22 +4,30 @@
 import Foundation
 
 /// Coordinates direct pasteboard fallback content while AppKit fulfills a
-/// batch of file promises. A receiver's reader callback is per promised file,
-/// not per receiver, so completion must be counted against the file names
-/// reported after each receiver is activated.
+/// batch of file promises. A receiver can represent several files, and its
+/// reader callback is normally per file. A receiver error is also allowed to
+/// terminate the receiver wholesale, so the remaining advertised files are
+/// credited as failed instead of leaving the batch waiting forever.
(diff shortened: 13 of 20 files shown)


PathGao:
Four things, then the one I couldn't check.

**1. All promises share one destination directory, and the first completion deletes it.**

`receivePromisedFiles` creates one `Promise-<uuid>` dir and hands the same one to every receiver, then each callback does `removeItem(at: destination)`. The handler fires once per promised file, so with two attachments the first callback deletes the directory — and whatever is still in it — while the second is mid-materialization. That file's copy fails and the item is dropped with no sign. An Outlook mail with two attachments is the common case here, not the edge one.

Move the destination inside the loop, one per receiver. While you're there, drop the `if storedURL != url` guard: `storedURL` is always a fresh UUID path in a different directory, so it's never equal.

**2. The two drop paths disagree on what a promise drop contains.**

`mergePasteboard` returns early on promises and never calls `items(from:)`. `accept` appends `items(from:)` *and* the promises. Same pasteboard, dropped on a pile vs. on the panel, gives one tile or two. Outlook puts a text/URL flavor on the board alongside the promise, so dropping one mail on empty shelf area lands the subject tile plus the .eml tile. The body describes promised files becoming "normal file items" and says nothing about a second tile.

Pick one. If the direct item is meant as a fallback for a promise that fails to materialize, it has to be added from the failure branch, not unconditionally up front.

**3. Every promise path reports success before anything lands.**

`receivePromisedFiles` returns `true` as soon as it dispatches, and `accept` / `mergePasteboard` hand that back to callers who treat it as done. `CommandBarCatalog.keepOnShelf` shows the shelf HUD on `true` and beeps on `false`; `ShelfTilesView.performDragOperation` returns it, which is what stops the drag image flying back to the source. So a promise that never materializes, a failed `storePromisedFile`, or a full shelf all produce: HUD says added, drag sticks, shelf is empty. The callback's `else { return }` branches say nothing to anyone.

There's no honest synchronous answer once materialization is async, and accepting optimistically is defensible. But the failure branches need to surface something — it's the same `false` those two callers already know how to render, just arriving later.

**4. `pasteboardCanCreateItem` now answers only from the promise check.**

```swift
if pasteboardHasFilePromises(pasteboard) {
    return !filePromises(from: pasteboard).isEmpty
}
```

`pasteboardHasFilePromises` matches the full `readableDraggedTypes` set, including `com.apple.NSFilePromiseItemMetaData` — the identifier `promisedFileTypeIdentifier` deliberately excludes because it isn't a file. A board that declares a promise flavor but yields no `NSFilePromiseReceiver` returns `false` here, and the fileURL / image / text checks below never run, so it gets refused where the old code accepted it. #873's title says Chromium re-declares promised flavors on image copies, so non-Outlook sources do put these types on the board. Fall through to the existing checks instead of early-returning.

**5. Not verified: the SwiftUI branch.**

`resolveItem` fulfills the promise with `loadFileRepresentation(forTypeIdentifier: "com.apple.pasteboard.promised-file-content-type")`. There is no `NSFilePromiseReceiver` on that path, and an `NSItemProvider` carrying that identifier holds the promised *type*, not the file — I'd expect a temp file containing a UTI string rather than the .eml. I can't settle it without a live Outlook drag onto the pill or the SwiftUI panel, which is the testing the body says wasn't available. Confirm that branch yields a real file before shipping, because the AppKit branch works differently and would mask it in casual testing.

I didn't run any of this — the above is `main` plus the diff.

**Impact:** multi-attachment Outlook drops silently lose all but the first file; a single mail dropped on empty shelf area can produce two tiles; a promise that fails to materialize still reports success to the HUD and the drag; boards that declare promise flavors without a readable receiver get refused outright.
**Merge:** needs the shared-destination fix (1) first.

ref #1083

muaz978:
Hi @PathGao, thank you for the careful review and for calling out the shared-destination, mixed-representation, asynchronous-error, and promise-detection issues. Your feedback exposed real edge cases in the first implementation, so I reworked the flow and pushed the fixes in commit `0d8f8c0`.

Here is what has been improved:

1. **Independent staging for every promise receiver**

   Each `NSFilePromiseReceiver` now gets its own UUID-backed staging directory. This prevents one receiver's callback from deleting another receiver's destination. I also stopped deleting the receiver directory from the first callback. A receiver can deliver more than one file, so removing the directory immediately could still race with a later attachment. Individual materialized files are removed after they are copied into Shelf's owner-only store, and the existing temporary-file sweep removes the now-unused `Promise-*` directories safely.

2. **One authoritative item for Outlook promise drops**

   Promise-backed pasteboards can advertise auxiliary text and URL flavors alongside the actual promised file. The implementation now consistently treats the promised file as authoritative for both empty Shelf drops and pile merges. It no longer adds the direct subject or URL representation up front, which prevents the extra subject tile plus `.eml` duplication you identified. The behavior is consistent across the AppKit pasteboard path and the mixed provider path.

3. **Visible reporting for asynchronous failures**

   A promise drop is accepted synchronously before Outlook finishes materializing the file, so a later failure cannot be returned through AppKit's immediate Boolean result. Promise materialization, storage, capacity, append, and pile-merge failures now use the existing beep convention and a throttled localized HUD message. Multiple failures from one multi-file drop are throttled so the user is not flooded with repeated alerts. This also covers a full Shelf and storage-copy failures instead of allowing an apparently successful drop to leave the Shelf empty.

4. **Fallback when promise metadata is present but unreadable**

   `pasteboardCanCreateItem` now only prioritizes file promises when it can read at least one `NSFilePromiseReceiver`. If the pasteboard merely declares a promise type without a readable receiver, detection falls through to the existing file, image, URL, and text checks. This preserves valid non-promise drops from providers that expose incomplete promise metadata.

5. **SwiftUI provider path clarified and hardened**

   The SwiftUI path still uses `NSItemProvider.loadFileRepresentation`, which provides a temporary file URL for the selected representation. The file is copied immediately into Shelf storage before the provider callback returns, and the provider's suggested filename is used when available. Failures now go through the same user-visible feedback path as AppKit promise failures.

I also added the new failure message to every supported localization so the error feedback does not regress into an English-only message.

Verification completed locally:

- `./build.sh --test` passed all 8,937 checks.
- `./build.sh` completed successfully.
- A file-backed `NSItemProvider` probe passed with `NSITEMPROVIDER FILE REPRESENTATION OK`.
- `git diff --check` passed.

The branch is updated and PR #1106 now contains the fixes and the expanded verification notes. I could not perform a live Outlook drag in this environment, so the remaining provider-specific manual check is still best done by a maintainer with Outlook and a real message or attachment. The upstream workflow may also remain pending until a maintainer approves it.

Thank you again for the detailed review. It materially improved the reliability of the implementation.

PathGao:
1–4 hold up on the new diff: per-receiver staging directories, promise-authoritative on both drop paths, failures routed through the beep + HUD, and `pasteboardCanCreateItem` falling through. Two things left.

**The SwiftUI promise branch is untested and now runs first, with no fallback.**

`resolveItem` takes the promise branch whenever `promisedFileTypeIdentifier` matches, and on failure it calls `reportPromiseFailure()` and `completion(nil)` — the fileURL / gif / image / URL / text branches below never run. Same shape as the `pasteboardCanCreateItem` bug you just fixed, minus the fallthrough.

That matters because the identifier it loads is `com.apple.pasteboard.promised-file-content-type`, which is where a promise provider writes the promised file's *UTI*, not its bytes. Any source registering a promise alongside real data — an ordinary AppKit image drag does — now takes the promise branch on the pill and the panel instead of the image branch that works today. If that branch doesn't yield a real file, this isn't "Outlook doesn't work in SwiftUI", it's a regression on drags that work on `main`.

The probe doesn't settle it. `NSITEMPROVIDER FILE REPRESENTATION OK` uses a file-backed provider, which was never the case in doubt. The open question is what `loadFileRepresentation(forTypeIdentifier: "com.apple.pasteboard.promised-file-content-type")` hands back when there is no `NSFilePromiseReceiver` on that path to invoke the sender's promise delegate at all.

One hint from the diff alone: `let title = provider.suggestedName ?? url.lastPathComponent`. For a promise-typed provider `suggestedName` is normally nil, so the title becomes whatever Foundation named the temp file, not the message subject. The body says original filenames become tile titles — that holds on the AppKit path, and I don't think it holds here.

Either exit works: fall through to the remaining branches on failure, or drag one Outlook message onto the pill and report what lands.

**Two smaller ones.**

`pasteboardHasDroppableContent` returns true on a declared promise type alone, while `pasteboardCanCreateItem` now requires a readable receiver. A board that declares the flavor without one makes the shelf slide out mid-drag and then refuse the drop. Use the same check in both, or say the looseness is deliberate.

The merge branch resolves `targetID` seconds after the drop was already accepted. If the pile is gone by then, `merge` returns false and the message is discarded with one beep. Appending it at top level instead of discarding loses nothing.

I didn't run any of this. Read the diff plus `ShelfService.swift` on `main`; the `readableDraggedTypes` contents and the SwiftUI branch's actual output are from documentation and reasoning, not from a machine.

**Impact:** on the Shelf pill and panel, any drag whose provider carries the promised-file-content-type flavor now goes through an unverified branch with no fallback, including image drags that work on `main`.
**Merge:** needs the `resolveItem` fallthrough first, or a live Outlook drag confirming that branch.

ref #1083

muaz978:
Hi @PathGao, thank you for the follow-up. I agree with the remaining concerns, especially the risk that the promise UTI could preempt an ordinary image drag that already works on `main`. I addressed both points by taking the fallback approach you suggested and pushed commit `bebb4a7`.

The remaining changes are:

1. **The SwiftUI promise branch now has a real fallback**

   `resolveItem` still attempts the promise representation first when a provider advertises the promised-file-content type, but it now accepts that result only when the returned URL is a readable regular file and can be copied into Shelf storage. If the load fails, returns no URL, returns something that is not a regular file, or cannot be stored, the code falls through to the existing standard loaders for file URLs, GIF data, images, URLs, and text.

   Promise failure feedback is delayed until that fallback also fails. This means a provider that carries the promise UTI alongside a working image or other direct representation will continue through the working path instead of producing an empty tile and an unnecessary error. A provider with neither a usable promise nor a usable direct representation still reports the failure through the beep plus localized HUD path.

   I did not treat the earlier file-backed provider probe as proof of Outlook behavior. It only verified the Foundation API contract for a provider that does return a real file. The implementation now remains safe for the unresolved provider case through the explicit fallback.

2. **Drag opening and drop validation now agree**

   `pasteboardHasDroppableContent` and `pasteboardCanCreateItem` now share `pasteboardHasReadableFilePromises`. A declared promise flavor alone will not make the Shelf slide out or count as a valid promise drop unless an `NSFilePromiseReceiver` can actually be read. Direct file, image, URL, and text types on the same pasteboard still fall through and remain eligible.

3. **A disappearing pile no longer loses a fulfilled file**

   If an asynchronous promise was dropped onto a pile and that pile is removed before fulfillment finishes, the completed file is now appended at top level. It is discarded only if the fallback append also fails, such as when the Shelf capacity is genuinely full, and that failure is reported to the user.

Verification after these changes:

- `./build.sh --test` reported `TESTS OK (8,937 checks)`.
- `./build.sh` completed successfully.
- `git diff --check` passed.

The branch is pushed and PR #1106 now includes the updated description documenting these final fixes. Live Outlook drag testing is still not available in this environment, so I am not claiming that provider-specific behavior has been proven here. The code now has the safe fallback required for the case you described, while preserving the intended Outlook promise path when it produces a valid file.

Thank you again for identifying this regression risk before merge.

PathGao:
The `resolveItem` fallback and the shared `pasteboardHasReadableFilePromises` both land as asked, and appending at top level when the pile is gone is right. Two left.

**The AppKit path never got the fallback.** `accept` and `mergePasteboard` return `receivePromisedFiles(...)` the moment a receiver reads, so `fileURLs(from:)` and `items(from:)` never run. Same shape you just fixed in `resolveItem`, minus the exit: a board carrying a promise alongside a usable image (#873's Chromium case) gets one beep where it lands a tile on `main`.

**`isReadableFile` doesn't answer the question it stands in for.** A temp file holding the promised UTI string is a readable regular file, so the SwiftUI branch stores it and builds a tile from it. The fallback only fires on a hard load failure, so the case I flagged walks straight through the guard. One live Outlook drag settles it; nothing in the diff does.

Small: `pasteboardHasDroppableContent` computes `hasReadablePromises` ahead of the cheap `directTypes` check, so every plain drag pays a `readObjects`.

I didn't run any of this — diff plus `main`.

**Impact:** any board with a readable promise plus ordinary flavors loses the ordinary path on the panel and on piles; the SwiftUI promise branch can still turn a non-file payload into a tile.
**Merge:** needs the `accept` / `mergePasteboard` fallback first.

ref #1083

muaz978:
Hi @PathGao, thank you for the additional review and for pointing me to the Chromium case in [issue #873](https://github.com/vorssaintapp/vorssaint-utils/issues/873) and its related [PR #874](https://github.com/vorssaintapp/vorssaint-utils/pull/874). I followed those references and made the AppKit and SwiftUI paths handle the promised-flavor edge case consistently.

I pushed the final changes in commits `4a610d3` and `62ff694`.

### AppKit fallback

`accept(pasteboard:)` and `mergePasteboard` now collect direct representations as a one-shot fallback before scheduling promise fulfillment. They do not add those representations up front, so Outlook's subject or URL flavor cannot create a duplicate tile beside the promised `.eml` or attachment.

If AppKit promise fulfillment fails, the direct representation is resolved on the main queue and added instead. If a promise succeeds, any owned fallback payload created while inspecting the pasteboard is discarded. The fallback state is lock-protected so several receivers cannot consume or add the same fallback more than once.

This preserves the Chromium behavior described in issue #873 and PR #874: when an image carries lazy promised flavors alongside a usable image representation, a failed promise no longer produces only a beep while the working image is lost.

### SwiftUI UTI placeholder protection

I reproduced the specific behavior you described with a focused Foundation probe. Registering `com.apple.pasteboard.promised-file-content-type` as a data representation caused `loadFileRepresentation` to return a readable, extensionless temporary file whose bytes were `public.jpeg\n`. That is a valid regular file from the filesystem's point of view, but it is only the promised UTI string, not the image content.

The SwiftUI path now rejects that exact shape: a short, extensionless file whose trimmed UTF-8 contents are a valid UTI identifier. It then falls through to the normal file, GIF, image, URL, or text loaders. A real promised file is still copied immediately into Shelf storage and keeps the provider's suggested name when available.

### Drag detection and disappearing targets

The drag auto-open path now checks cheap direct types first and only reads promise receivers when no direct representation was found. This removes unnecessary `readObjects` work for ordinary drags while still using the same readable-promise rule as drop validation.

The existing asynchronous pile-target fallback remains in place. If the target disappears before fulfillment finishes, the completed file is appended at top level instead of being discarded.

Verification after the final change:

- `./build.sh --test` reported `TESTS OK (8,937 checks)`.
- `./build.sh` completed successfully.
- `git diff --check` passed.
- The UTI-placeholder probe reproduced the extensionless `public.jpeg` temporary-file case used by the new guard.

The branch and PR description are updated. Live Outlook dragging is still not available in this environment, so I am not claiming provider-specific live confirmation. The implementation now handles the documented Foundation behavior and retains a direct fallback for providers such as Chromium that expose both promised and ordinary representations.

Thank you again for the precise review. The references helped identify a real regression path beyond Outlook itself.

muaz978:
One final update after the implementation review: upstream `main` advanced while this PR was being updated, so I rebased the branch onto the latest upstream commit and force-pushed it safely with lease.

The PR now points to `f541487` and is no longer behind the base branch. The rebased verification is also complete:

- `./build.sh --test` reported `TESTS OK (8,940 checks)`.
- `./build.sh` completed successfully.
- `git diff --check` passed.

The three additional checks come from the latest upstream `main`, so the PR description has been corrected to reflect the current count. The implementation and the detailed fixes described above are unchanged by the rebase. The only remaining merge blockers are the upstream workflow approval and code-owner review protections, plus the recommended live Outlook drag verification.

PathGao:
The probe is the right kind of evidence — you reproduced the placeholder shape instead of arguing about it. AppKit fallback, the shared readable-promise check, and cheap-types-first all land.

**One fallback state, every receiver races for it.** `receivePromisedFiles` builds a single `PromiseFallbackState`. Two attachments where the first fails and the second succeeds: the failure appends Outlook's subject item, then the success finds nothing left to discard and appends the `.eml`. Body says auxiliary flavors "are not added as duplicate tiles" — under partial failure they are. One state per receiver, or hold the fallback until every receiver has reported.

**The SwiftUI branch is safe now, still unproven.** The identifier it loads is the one your probe showed returning a UTI string, so that branch may reject every time and never deliver anything — in which case registering the promise types on the pill and panel buys nothing. Not a regression, just not evidence the feature works there. One live Outlook drag settles it.

I didn't run any of this — diff plus `main`.

**Impact:** a multi-attachment Outlook drop with one failed promise lands a stray subject tile beside the message; the SwiftUI surfaces may never take the promise path at all.
**Merge:** needs the per-receiver fallback fix first.

ref #1083

muaz978:
Hi @PathGao, you were right to call out the shared fallback race. My previous one-shot state prevented duplicate fallback consumption, but it still allowed the first failed receiver to add the direct subject before a second receiver had a chance to succeed.

I fixed that in commit `e97af88` and pushed it to the PR branch.

The fallback now coordinates all promise receivers as a batch:

- Each receiver is identified separately.
- A receiver that succeeds marks the batch as having a successful promise and suppresses the direct fallback. Any owned fallback payload is then discarded.
- A receiver that fails does not immediately add the direct item. It waits until every receiver has reported.
- The direct fallback is used only when all receivers fail.
- If at least one receiver succeeds and another fails, the successful promised files remain in the Shelf and the failed promise is reported without adding a stray subject or URL tile.
- The fallback remains lock-protected because receiver callbacks run asynchronously and may race on the promise operation queue.

This handles the two-attachment case you described: one failed receiver and one successful receiver can no longer produce a subject tile plus an `.eml` tile. The promised files remain authoritative, and the direct representation is only a last-resort recovery path when the complete promise batch fails.

The previous fixes remain in place as well:

- Per-receiver staging directories prevent attachment materialization from deleting another receiver's destination.
- Extensionless temporary files containing only a valid UTI such as `public.jpeg` are rejected instead of becoming Shelf files.
- SwiftUI resolution falls through to direct file, GIF, image, URL, or text representations when the promise representation is unusable.
- AppKit drag auto-opening uses the same readable-promise rule as drop validation and checks cheap direct types first.
- A fulfilled promise is appended at top level if its pile target disappears during asynchronous fulfillment.

Verification after this final change:

- `./build.sh --test` reported `TESTS OK (8,940 checks)`.
- `./build.sh` completed successfully.
- `git diff --check` passed.

The branch is pushed and the PR description now documents the receiver barrier. Live Outlook dragging is still not available here, so I am not claiming provider-specific live confirmation. The multi-receiver fallback race itself is addressed in the implementation and covered by the synchronized state design.

Thank you for catching this last partial-failure case. It was a real correctness issue in the previous fallback implementation.

muaz978:
Thank you, @PathGao. You identified a real remaining race in the previous fallback implementation, and the two-receiver failure/success example was especially useful.

I have fixed it and pushed the rebased update to this PR. The latest branch head is [`640f321`](https://github.com/muaz978/vorssaint-utils/commit/640f321475e0280c177177fc75cf50430ce75aa0), based on the current upstream `main`.

### Receiver-level fallback barrier

The previous implementation shared one fallback state across all promise receivers, but a failed receiver could consume the fallback before another receiver had reported success. That could produce exactly the stray Outlook subject tile described in your comment.

The new `PromiseFallbackState` coordinates the complete receiver batch:

- Each receiver is assigned its own identity and is recorded only once.
- A receiver failure is held as `.wait` until every receiver has reported.
- The first successful receiver marks the batch as successful and suppresses the direct fallback.
- The direct fallback is consumed only when every receiver has failed.
- A partial result therefore remains promise-authoritative: successful promised files are kept, failed receivers are reported, and no direct subject or URL tile is appended beside them.
- The state remains lock-protected because receiver callbacks can complete asynchronously and race with one another.

This closes the specific case of one failed attachment and one successful attachment. It also keeps the existing fallback behavior for a completely failed promise batch, including pile merges and owned-payload cleanup after success.

### SwiftUI and the UTI placeholder

The SwiftUI branch remains intentionally defensive. The focused probe reproduced the short, extensionless temporary file containing a UTI string such as `public.jpeg` for `com.apple.pasteboard.promised-file-content-type`. The implementation rejects that placeholder, while accepting a real readable file, and then falls through to the provider's ordinary file, GIF, image, URL, or text representations when the promised representation is not usable.

This follows the Foundation contract that `loadFileRepresentation(forTypeIdentifier:)` supplies a temporary file representation whose contents must be copied during the completion callback. See [Apple's `NSItemProvider` documentation](https://developer.apple.com/documentation/foundation/nsitemprovider/loadfilerepresentation%28fortypeidentifier%3Acompletionhandler%3A%29) and [Apple's `NSFilePromiseReceiver` documentation](https://developer.apple.com/documentation/appkit/nsfilepromisereceiver).

As you noted, this is still not a substitute for a live Outlook drag. I am not claiming that the SwiftUI promise branch has been provider-verified here. One real Outlook drop with a message, one attachment, and multiple attachments remains the final manual validation step.

### Related reference

I also reviewed [issue #873](https://github.com/vorssaintapp/vorssaint-utils/issues/873) and [PR #874](https://github.com/vorssaintapp/vorssaint-utils/pull/874). That prior fix confirms the same underlying principle for Chromium promised image drags: retain and consume the promised representation without re-declaring an already-promised flavor, otherwise clipboard history can receive duplicate entries. The current Shelf change preserves direct representations only as a coordinated last-resort fallback, so it does not reintroduce duplicate Outlook tiles.

### Verification

- `./build.sh --test` -> `TESTS OK (9,076 checks)` after rebasing onto current upstream `main`.
- `./build.sh` completed successfully.
- `./build/Vorssaint --selftest` -> `SELFTEST OK`.
- Focused Foundation probe reproduced the UTI-text placeholder shape.
- `git diff --check` passed.
- The branch comparison is now 6 commits ahead and 0 commits behind `main`.

Thank you again for catching this partial-failure case. The fallback now waits for the complete receiver batch, which removes the merge-blocking stray-tile behavior while preserving recovery when the entire promise operation fails.

PathGao:
The receiver-level barrier fixes the case I raised — one failed receiver plus one successful receiver no longer produce a stray tile.

**The barrier counts receivers, but the callback fires per file.** Your own staging comment says a receiver "can represent more than one file." If that holds, `resolvedReceivers` closes the batch on each receiver's *first* callback. One receiver, two files: file 1 fails, `resolvedReceivers.count == receiverCount`, `.fallback` appends Outlook's subject; file 2 then succeeds, `recordSuccess` sees the ID already inserted, returns nil, nothing is discarded, and the `.eml` lands beside it. Same stray tile, one level down.

Reverse order is worse. The success consumes the fallback, then the later failure returns `.wait` and is never re-evaluated — no beep, no HUD, for a file the user lost.

Either that staging comment is wrong or the barrier is. If it's the barrier, `NSFilePromiseReceiver.fileNames` gives the count up front, so the expected total is `promises.reduce(0) { $0 + $1.fileNames.count }` and the identity has to be per callback rather than per receiver.

I didn't run any of this — diff plus `main`, and `fileNames` is from documentation.

**Impact:** an Outlook mail whose attachments arrive through a single receiver can still land a stray subject tile, or drop a file with no feedback at all.
**Merge:** needs the barrier keyed per callback first.

ref #1083

muaz978:
Thank you, @PathGao. You found another genuine race, and you were right that the previous receiver-level barrier was still insufficient for a non-item-based receiver that reports multiple files.

I have corrected this in the latest rebased commit [`a3f13f1`](https://github.com/muaz978/vorssaint-utils/commit/a3f13f16e3517ec06dc805ec810c66973e2d3cce), now based on upstream `main` at `ed2d1a6`.

### Callback-level completion accounting

The previous coordinator keyed completion by receiver ID. That incorrectly treated the first callback from a multi-file receiver as the receiver's complete result. The coordinator is now a separate, testable `ShelfPromiseFallbackCoordinator` and tracks the actual reader callback count:

- After each receiver is activated, AppKit's `fileNames.count` contributes the expected number of file callbacks.
- The expected batch total is the sum of those per-receiver counts, with a defensive minimum of one for a receiver that does not publish names.
- Each reader callback increments the completed-file count independently, without deduplicating by receiver.
- A failure is held until the complete expected callback count is reached.
- A success suppresses and discards the direct fallback, but does not hide later failures.
- If any file fails and another succeeds, the promised files remain authoritative and one failure notification is emitted after the batch completes.
- The direct fallback is used only when every expected promised-file callback fails.

This covers both orderings from your comment:

1. File 1 fails, file 2 succeeds: no subject fallback is appended, and the failure is reported after file 2 completes.
2. File 1 succeeds, file 2 fails: the fallback remains suppressed, and the later failure is reported instead of disappearing.

The coordinator tests exercise both multi-file orderings and a multi-receiver all-failure batch. Those tests are compiled into `./build.sh --test` so the accounting cannot silently regress back to receiver-level counting.

### AppKit destination correction

While tracing this path, I also corrected the staging layout to follow AppKit's contract: all receivers belonging to one drag now receive the same staging destination. The implementation still copies every promised file into Shelf's owner-only persistent store immediately from the callback, and it keeps the generated staging directory outside the persistent store. See [Apple's `NSFilePromiseReceiver` documentation](https://developer.apple.com/documentation/appkit/nsfilepromisereceiver), including the multi-file non-item-based receiver behavior, and [the `receivePromisedFiles` documentation](https://developer.apple.com/documentation/appkit/nsfilepromisereceiver/receivepromisedfiles%28atdestination%3Aoptions%3Aoperationqueue%3Areader%3A).

The earlier safeguards remain in place: Outlook's promised file stays authoritative over auxiliary subject and URL flavors, invalid UTI-text placeholders are rejected, direct representations remain available as a coordinated fallback, disappearing pile targets keep successful files at top level, and storage or append failures produce the existing failure feedback.

### Verification

- The latest branch is rebased onto upstream `main` and compares as 7 commits ahead and 0 commits behind.
- `./build.sh` completed successfully, including the protected fan helper and application bundle assembly.
- `git diff --check` passed.
- `./build.sh --test` compiled and ran 9,096 checks. The new promise-coordinator checks passed. Eight existing checks remain unavailable in this host environment because they require AppleScript/media conversion or mounted-system resources: WhatsApp file buckets, image conversion, AppleScript execution, JPEG/video tooling, and editor staging.
- The application self-test was attempted separately. Its only failures were unavailable host resources here: AppleSMC, mounted disks, swap, and uptime. No Shelf promise assertion or compile error failed.

I am still not claiming live Outlook or Mail validation from this environment. One real Outlook drag with a message, one attachment, and multiple attachments remains the provider-specific manual check before merge. The callback-level race itself is now covered directly, and the protected workflow can run once a maintainer approves it.

Thank you again for reading the implementation closely and for pointing out the per-file callback semantics. This correction addresses the exact stray-tile and silent-loss cases in the latest comment.

PathGao:
Callback-level counting is the right fix, and putting the coordinator in its own file with tests means it can't silently regress. Two things.

**The error branch deletes `url` while every receiver now shares one destination.** The reader hands back a URL you didn't write, and on failure there's no guarantee it's a file rather than the destination directory itself. If it's the directory, `removeItem(at: url)` takes a sibling receiver's already-materialized attachment with it — round one's bug, back through the error path now that the destination is shared again. Nothing needs deleting there: the file was never written, and the sweep clears `Promise-*` at next launch anyway.

**The barrier assumes exactly `fileNames.count` reader calls.** If a receiver that fails wholesale reports once instead of once per file, `completedFileCount` stops short of `expectedFileCount` and the batch never completes — no fallback, no beep, no HUD, which is the silent loss the coordinator exists to prevent. There's no timeout. Crediting a failed receiver's remaining files on the first failure closes it. The new tests feed the counts by hand, so they pass either way.

I didn't run any of this — diff plus `main`. The reader's error-URL contract is from Apple's sample code, which never touches that URL.

**Impact:** on a multi-attachment Outlook drop, one failed attachment can delete a sibling's file, or wedge the batch with no feedback at all.
**Merge:** needs the error-path `removeItem` guard first.

ref #1083

muaz978:
Thank you for the careful follow-up and for checking the AppKit error-path contract. I audited both concerns against the current implementation and have pushed a further fix in `97d8f1a`.

The changes are:

- The error callback no longer removes the URL supplied by `NSFilePromiseReceiver`. That URL is not guaranteed to be a file written by the reader, and with the required shared destination it could be the destination directory itself. The error path now leaves it untouched for the existing temporary-file sweep.
- Cleanup after a successful promise, or after Shelf fails to store a returned promise file, is now guarded. Shelf only removes a regular file that is directly inside the shared staging directory. It will not remove a directory, an absent path, a partial path, or a sibling/out-of-destination item.
- Receiver state is now tracked independently. Each receiver starts with a provisional slot before activation, then adopts the receiver's advertised `fileNames.count` after activation returns. This also covers an immediately dispatched callback without treating it as an unknown receiver.
- A receiver error closes that receiver wholesale and credits all remaining advertised files as failed. This prevents a multi-file receiver that reports one error callback from leaving the batch waiting forever with no fallback, beep, or HUD feedback.
- A late success callback from a receiver already closed by an error is rejected by the coordinator. If AppKit has materialized a regular file for that late callback, the already-stored copy is removed safely, so it cannot create a duplicate tile after fallback has been selected.
- Existing partial-failure behavior is preserved: a successful receiver suppresses the direct representation fallback, while the completed batch still reports the failure once. If every receiver fails, the direct representation is used only after all receivers have reached a terminal state.

The coordinator tests now cover wholesale failure of a multi-file receiver, synchronous or early callbacks before the final file count is known, partial failure across receivers, late callbacks after wholesale failure, and all-receiver failure. The full application build also passed after rebasing the branch onto the latest `main` (`34121b4`).

Verification completed:

- `./build.sh --test` compiled and ran 9,161 checks. All new promise-coordinator checks passed.
- The same 8 unrelated host-dependent checks remain unavailable in this environment, covering AppleScript, WhatsApp bucket discovery, image/video conversion, and editor staging. They are not failures in the promise changes.
- `./build.sh` passed after the final rebase, including protected helper compilation and bundle assembly.
- `git diff --check` passed.
- The standalone app self-test reported only unavailable host resources here: AppleSMC, mounted disk volumes, swap, and uptime.

The implementation follows the `NSFilePromiseReceiver` behavior documented by Apple, especially the fact that one receiver can advertise multiple files, callbacks deliver individual promised files, and the same destination must be used for all receivers in one drag:

- [NSFilePromiseReceiver](https://developer.apple.com/documentation/appkit/nsfilepromisereceiver)
- [fileNames](https://developer.apple.com/documentation/appkit/nsfilepromisereceiver/filenames)
- [receivePromisedFiles](https://developer.apple.com/documentation/appkit/nsfilepromisereceiver/receivepromisedfiles%28atdestination%3Aoptions%3Aoperationqueue%3Areader%3A)

Live Outlook and Mail drag testing is still dependent on having those applications and their provider behavior available. The code paths and coordinator behavior are covered locally, and the branch has been rebased and pushed so the protected GitHub checks can run on the updated PR.

PathGao:
Both fixes land: the error path no longer touches AppKit's URL, and a wholesale failure credits the remaining files.

**`recordSuccess` also rejects a success from a receiver that went terminal by count.** The guard is `!receiver.isTerminal`, but the comment above it only describes wholesale failure. A receiver publishing no `fileNames` gets `max(0, 1) == 1`, so its second file is stored, rejected, then deleted — `acceptFile: false`, `reportFailure: false`, no beep. The defensive floor turns into a silent cap. Gate the rejection on `isWholesaleFailure`, and let an extra success raise `expectedFileCount`.

**The placeholder guard covers one identifier; the loader doesn't pick one.** `promisedFileTypeIdentifier` takes the first `registeredTypeIdentifiers` entry that is in `readableDraggedTypes` and isn't metadata — provider order decides, so it can be `promised-file-url` rather than the `promised-file-content-type` your probe covered. A temp file holding a URL string isn't a valid UTType, so `isUsablePromiseFile` passes it and it becomes a tile. Load the content-type identifier explicitly.

I didn't run any of this — diff plus `main`.

**Impact:** a receiver that advertises no filenames drops every file after the first with no feedback; on the pill and panel a non-content promise flavor can still become a tile.
**Merge:** needs the `recordSuccess` terminal-by-count fix first.

ref #1083

muaz978:
Thank you for checking the follow-up so closely. Both additional issues are valid, and I have fixed them in the latest commit [`0b5527b`](https://github.com/muaz978/vorssaint-utils/commit/0b5527b).

### Defensive receiver counts

You are right that `isTerminal` cannot by itself mean that no more callbacks are valid. The defensive `max(fileNames.count, 1)` is only a fallback for an omitted or incomplete advertised count, not a hard upper bound.

The coordinator now:

- Rejects callbacks only after `isWholesaleFailure` has closed a receiver because of an error.
- Accepts an additional success after a receiver reached its current count, expanding the expected count to include that callback.
- Applies the same expansion to an additional failure, so a late failed file cannot disappear without feedback either.
- Preserves the separate provisional-registration path for callbacks that arrive before activation returns.
- Keeps late callbacks from a receiver that genuinely failed wholesale rejected, preventing fallback duplication.

This means a receiver that reports no filenames but later delivers multiple files no longer silently drops its second file. The same behavior is covered for an extra failure following an early success, since that is the symmetric data-loss case.

### Deterministic SwiftUI promise loading

You are also right that selecting the first registered promise flavor lets provider ordering decide what Shelf loads. The SwiftUI path now explicitly requests only:

```text
com.apple.pasteboard.promised-file-content-type
```

It no longer selects `promised-file-url` or another promise flavor merely because it appears first in `registeredTypeIdentifiers`. A URL-string temporary file therefore cannot pass through the promised-file path as though it were a real attachment. If the content promise is not registered or cannot produce a usable file, the provider falls through to its ordinary file, GIF, image, URL, or text representation.

I also removed the now-unused metadata selector and added regression assertions that verify the explicit content-type selection and reject the old provider-order-dependent implementation.

### Verification

- `./build.sh --test` compiled and ran 9,189 checks.
- The new overflow-callback, wholesale-failure, late-callback, and explicit-promise-type checks passed.
- The same 8 unrelated host-dependent checks remain unavailable in this environment: WhatsApp bucket discovery, image/video conversion, AppleScript execution, and editor staging.
- `./build.sh` passed after the changes, including application compilation, protected helper compilation, bundle assembly, and diff validation.
- `git diff --check` passed.
- The branch remains rebased onto current upstream `main`; it is 10 commits ahead and 0 commits behind.

The branch is now pushed to the PR, and I updated the PR description with these additional safeguards. Live Outlook and Mail provider testing remains the only provider-specific validation that cannot be performed in this environment.

Thank you again for catching the distinction between a receiver reaching its current count and a receiver actually failing wholesale. That distinction is now explicit in both the coordinator state and the regression coverage.

PathGao:
Both fixes land: `recordSuccess` gates on `isWholesaleFailure`, and the SwiftUI loader names the content type instead of taking whatever came first.

**A first-file failure inside one receiver deletes every later file from it.** Outlook, one receiver, three attachments. Attachment 1 errors, `recordFailure` credits all three and closes the receiver, the batch appends the subject fallback. Attachments 2 and 3 then materialize, get copied into the store, and `recordSuccess` rejects them — `removePromisedFileIfRegular(storedURL)` deletes both, no beep. Reverse the arrival order and both survive, so the same drop lands differently depending on which callback wins. Body says the successful promised files remain in the Shelf; that holds across receivers, not inside one, and per-callback counting exists precisely because one receiver carries several files.

I called the stray subject tile a blocker earlier — it's the lesser of the two. If a late success can't be reconciled with an already-appended fallback, append it anyway rather than deleting it.

Small: `swiftUIDropTypes` registers every `readableDraggedTypes` entry, but only the content type can resolve one. A provider carrying just the promised-URL flavor targets the pill and then drops into nothing — no tile, no beep. That's from documentation, not run.

I didn't run any of this — diff plus `main`.

**Impact:** a multi-attachment Outlook drop whose first attachment fails silently deletes the ones that succeed.
**Merge:** needs the late-success path to keep the file first.

ref #1083

muaz978:
Thank you for catching this final same-receiver failure path. You are right that the previous change still deleted valid later files from a receiver after its first callback reported an error.

I have fixed this in the latest rebased commit [`9410500`](https://github.com/muaz978/vorssaint-utils/commit/9410500e29e1262a3fc27fcb8a47b0af18c894ec).

### Late files from a failed receiver are now preserved

The coordinator now distinguishes two different terminal states:

- A count-based terminal state means the receiver has reached its currently known count. Additional callbacks expand the expected count and are accepted.
- A wholesale-failure state means AppKit reported an error for the receiver. The receiver's remaining advertised files are credited as failed for barrier completion, but later materialized successes are still accepted rather than deleted.

This handles the exact three-attachment sequence from your comment:

1. Attachment 1 fails.
2. The receiver is closed for fallback accounting, so the direct subject representation may be queued if no other promise succeeds.
3. Attachments 2 and 3 materialize later.
4. Their stored files are accepted and appended, even if the subject fallback was already queued.

If the later success arrives before fallback has been consumed, the direct fallback is discarded as before. If fallback has already been returned or appended, the promised file is kept and appended rather than deleted. This follows your recommendation to preserve the user's files when the two asynchronous outcomes cannot be reconciled retroactively.

The same receiver-level rule now applies to extra failure callbacks too, so a defensive minimum cannot silently hide an additional failed file. Only callbacks from an unknown receiver are rejected by the coordinator. A receiver explicitly closed by wholesale failure no longer causes later valid successes to be rejected.

### SwiftUI drop registration now matches the loader

I also corrected the smaller SwiftUI issue. The SwiftUI pill and panel now advertise only `com.apple.pasteboard.promised-file-content-type`, which is the promise flavor the loader explicitly requests. They no longer register every AppKit-readable promise flavor while resolving only one of them.

A provider containing only the promised URL flavor therefore cannot target the SwiftUI drop surface and then disappear without producing a tile or failure signal. The ordinary direct representations remain available when the content promise is absent or unusable.

### Verification

- `./build.sh --test` compiled and ran 9,320 checks.
- The new same-receiver late-success, overflow-failure, wholesale-failure, and SwiftUI promise-registration checks passed.
- The same 8 unrelated host-dependent checks remain unavailable in this environment, involving WhatsApp bucket discovery, AppleScript, image/video conversion, and editor staging.
- `./build.sh` passed after rebasing onto the latest upstream `main`, including application compilation, protected helper compilation, bundle assembly, and signing.
- `git diff --check` passed.
- The branch is rebased onto the latest upstream `main` and compares as 11 commits ahead and 0 commits behind.

The branch has been pushed to this PR and the PR description has been updated. Live Outlook and Mail dragging remains the only provider-specific validation that is not available in this environment.

Thank you again for identifying the distinction between closing a receiver for fallback accounting and discarding files that may still materialize from that receiver. The late-success path now preserves those files instead of silently deleting them.

PathGao:
The late-success path keeps the file now, and splitting "closed for accounting" from "discard its files" is the right cut.

**A late success re-arms the beep the fallback branch deliberately suppressed.** One receiver, two files. File 1 errors, the receiver closes wholesale, `completedOutcomeIfReady` returns the subject fallback with `reportFailure: false` — and never sets `failureReported`. File 2 then materializes: `recordSuccess` calls `reportFailureIfReady()`, every condition holds, so the file is appended and the "could not add the dropped item" HUD fires in the same main-queue block. Set `failureReported` in that branch, or report it there; leaving it armed hands the beep to whichever callback lands next.

Still the only thing that shows the feature works: one live Outlook drag. Nothing in the diff substitutes for it, and no checks have run on this branch yet.

I didn't run any of this — diff plus `main`.

**Impact:** a multi-attachment Outlook drop with one failed attachment shows a drop-failed alert at the moment a later attachment lands successfully.
**Merge:** needs the fallback branch's `reportFailure` gate first.

ref #1083

muaz978:
Thank you for catching this notification-state race. You are right that the fallback branch intentionally suppresses the failure feedback, but it was leaving the failure state armed for a later success callback from the same receiver.

I fixed this in the latest rebased commit [`ca170f0`](https://github.com/muaz978/vorssaint-utils/commit/ca170f0309ea15ba04d0692548d1c9419199228d).

### Fallback failure state is now handled exactly once

When the coordinator determines that every currently known promise path has failed and selects the direct representation fallback, it now marks the failure as handled before returning that fallback outcome.

This preserves the intended behavior for the sequence you described:

1. The first attachment fails.
2. The receiver closes for fallback accounting.
3. The direct subject fallback is selected without a failure HUD because the user still has a usable recovery path.
4. Later attachments from the same receiver materialize and are preserved.
5. Those late successes are appended without re-arming the already-suppressed failure notification.

The change does not suppress real storage, append, merge, or fallback-insertion failures. Those paths still call the existing failure reporting behavior directly. It only prevents the coordinator from reporting the same receiver failure again when a later promised file arrives after fallback has already handled it.

The regression test now requires that a late success after selected fallback is accepted, does not request fallback disposal, and returns `reportFailure == false`.

### Verification and branch state

- `./build.sh --test` compiled and ran 9,413 checks.
- The new notification-state, late-success, overflow-callback, wholesale-failure, and SwiftUI promise checks passed.
- The same 8 unrelated host-dependent checks remain unavailable in this environment, covering WhatsApp bucket discovery, AppleScript execution, image/video conversion, and editor staging.
- `./build.sh` passed after the final rebase, including application compilation, protected helper compilation, bundle assembly, signing, and `git diff --check`.
- The branch was rebased onto the latest upstream `main` at `b0dde65` and pushed at `ca170f0`.
- The branch is currently 12 commits ahead and 0 commits behind that upstream base.

I also updated the PR description with this notification-state safeguard. Live Outlook and Mail dragging remains the only provider-specific validation that is not available in this environment.

Thank you again for identifying that fallback selection must close the notification lifecycle as well as the materialization lifecycle. The later-success path now preserves the files without producing a misleading drop-failed alert.

PathGao:
The notification lifecycle is closed now — fallback selection marks the failure handled, so a late file lands without an alert.

**A Shelf-side copy failure closes the receiver wholesale.** `storePromisedFile` returning nil routes to `recordFailure`, which sets `isWholesaleFailure` and credits every remaining advertised file. The body scopes that state to "if AppKit reports an error for a receiver" — a failed `copyItem` isn't one. Outlook, one receiver, three attachments: attachment 1 fails to copy, the batch reads as complete, the subject fallback is appended, then 2 and 3 land beside it. Credit one file, not the receiver.

**The pill and the panel never reach the coordinator.** `accept(providers:)` has no `NSFilePromiseReceiver` path — it calls `loadFileRepresentation` for the content type. If that returns the placeholder your own probe reproduced, `isUsablePromiseFile` rejects it, `resolveStandardItem` runs, and Outlook's subject becomes the tile: the duplicate the AppKit path exists to prevent. One live drag onto the pill settles it, and nothing else in the diff does.

I didn't run any of this — diff plus `main`.

**Impact:** one attachment failing to copy lands a stray subject tile beside the ones that worked; a mail dropped on the pill can become its subject text instead of the message.
**Merge:** needs the store-failure path to credit one file first.

ref #1083

muaz978:
Thank you, both findings were valid. I have updated the branch and rebased it onto the current `main`.

### 1. A Shelf-side copy failure now counts as one failed promised file

`storePromisedFile(...) == nil` no longer takes the same path as an AppKit receiver error.

- `recordReceiverFailure` is reserved for an error supplied by AppKit. It terminates that receiver wholesale and credits its remaining advertised callbacks, which prevents a genuinely failed receiver from keeping the batch open forever.
- `recordFileFailure` is used only after Shelf has received one callback URL but cannot copy that particular file into its persistent store. It advances the count by one and leaves the receiver open, so Outlook can still deliver attachments 2 and 3.

The new regression case uses one receiver advertising three files: the first local copy fails, the next two succeed, no subject fallback is produced, and the partial failure is reported once after the batch completes.

### 2. The expanded Shelf and docked pill now use the AppKit promise route

The SwiftUI `.onDrop` / `NSItemProvider.loadFileRepresentation` path has been removed from both surfaces. It was a separate representation loader and could not construct the `NSFilePromiseReceiver` that the pasteboard route uses.

Both surfaces now host the existing AppKit `WindowMoveHandle` destination above their material views and below their controls. It registers `ShelfService.tileDropTypes` and forwards the original `sender.draggingPasteboard` to `accept(pasteboard:)`. Promise-backed drops therefore enter the same canonical `NSFilePromiseReceiver` coordinator as the existing native Shelf tile/pasteboard paths. Outlook's auxiliary subject text is retained only as a failure fallback and cannot be appended as a second tile after promise materialization.

I also removed the provider-batch helper that became unused when the duplicate SwiftUI route was retired, so there is now one promise implementation rather than two competing ones.

### Verification

- Rebased onto `main` at `ed882cd`; GitHub now reports this branch is `0` commits behind the base.
- `./build.sh` passed after the rebase, including release compilation, protected fan-helper compilation, bundle assembly, and ad-hoc signing.
- `./build.sh --test` compiled and ran 9,498 checks. The new promise-accounting and native-drop-route regressions passed.
- Eight pre-existing host-dependent checks remain unavailable in this build environment, all outside Shelf promise handling: WhatsApp bucketing, image/media conversion, in-process AppleScript, and external-media staging checks.
- `git diff --check` passed.

I cannot perform a genuine Outlook or Mail drag from this environment, so I have documented that boundary rather than claiming it was tested. The remaining recommended manual check is one multi-attachment Outlook or Mail drag onto both the expanded Shelf panel and docked pill.

Thank you again for catching the difference between a callback-level storage failure and a receiver-level AppKit failure, and for calling out the split SwiftUI drop path.

PathGao:
Splitting `recordReceiverFailure` from `recordFileFailure` fixes the case I raised, and retiring the SwiftUI route leaves one promise implementation.

**The stray subject tile returns through the AppKit error path.** Body says the split "prevents a message-subject fallback tile from appearing beside attachments that did succeed" — that holds for `recordFileFailure` only. One receiver, three attachments, attachment 1 errors: `recordReceiverFailure` credits all three, the batch completes, the subject fallback is appended, then 2 and 3 materialize and are kept beside it. Reader errors are per file, so this is the ordinary one-bad-attachment case. Crediting the rest to unwedge the barrier is right; appending the fallback at that moment isn't — hold the fallback item's id and remove it when a promised file lands.

**`WindowMoveHandle` moved from bounded overlay to full-surface background.** Every use on `main` is an overlay framed to dodge the controls — `topMoveHandle` subtracts 58/96 for the pin and close buttons. The new ones cover the whole card and the whole pill, and `mouseDown` calls `window?.performDrag` unconditionally. The pill's only affordance is its tap gesture; whether SwiftUI gives that click to the gesture or to the NSView beneath isn't visible in the diff. One click settles it.

I didn't run any of this — diff plus `main`.

**Impact:** an Outlook drop where one attachment errors lands a subject tile beside the ones that worked; the pill may drag instead of open.
**Merge:** needs the wholesale-failure path to stop leaving a fallback tile beside late successes.

ref #1083

PathGao:
Separating receiver errors from copy errors is right, and retiring the SwiftUI route leaves one promise implementation instead of two.

**The pill's only affordance now sits over a move handle.** `ShelfPanelMoveView.mouseDown` runs `window?.performDrag(with:)`, and the new handle is in the pill's background under `.onTapGesture { shelf.expandDocked() }`. Every other `acceptsDrops: true` handle in the panel is an `.overlay` sized to dodge the controls — `topMoveHandle` is `panelWidth - 96` wide for exactly that reason. If it wins hit testing on the pill's empty area, clicking drags the docked window instead of opening the shelf.

**Half the drop-route assertions already pass on `main`.** `shelfViewSource.contains("WindowMoveHandle(acceptsDrops: true)")` matches `topMoveHandle` verbatim. The `.onDrop(` negatives pass on an empty string too, so an unresolved relative path turns the whole block green.

**A full shelf leaks the copy.** The reject branch calls `removePromisedFileIfRegular(storedURL)`; the `guard self.append(item) else` right below it doesn't, so the file sits in the store until the next launch sweep.

I didn't run any of this — diff plus `main`. The pill click and one live Outlook drag both need a real build.

**Impact:** clicking the docked pill may move the window instead of opening the shelf; a promised file dropped onto a full shelf leaves an orphan copy.
**Merge:** needs the pill's click path confirmed first.

ref #1083

PathGao:
Closing this. The mechanism is the right one (`NSFilePromiseReceiver.readableDraggedTypes`, nothing Outlook-specific), and the review rounds were answered quickly. What decides it is your own statement, twice: "Live Outlook or Mail drags were not available in this environment" and "I cannot perform a genuine Outlook or Mail drag from this environment". Thirteen pushes, none exercised by the drag the feature exists for, and no Mac model or macOS build anywhere in the description. Since 2026-08-25 a feature that its author cannot run on a real Mac is not accepted here, and the maintainer runs every branch locally before merging; a branch nobody has dragged into is a review he has to do from scratch.

Two things still in the diff for whoever has an Outlook to test against: the full-shelf branch (`guard self.append(item) else { reportPromiseFailure(); return }`) never calls `removePromisedFileIfRegular`, so a rejected promise leaks its temp file, unlike the `acceptFile == false` branch above it; and `recordReceiverFailure` sets `completedFileCount = expectedFileCount` and appends the subject tile while later files from the same receiver are still accepted, so one bad attachment lands the fallback tile beside the ones that worked, which the description says cannot happen.

Keep the branch. If you get access to a Mac with Outlook or Mail, reopen with the drag on record and this is a short review.

PR #1112, opened 2026-08-28
fix(clipboard): keep a long entry inside its row in the menu panel

andreisuslov:
## Summary

Clicking a long text entry in the menu panel's Clipboard utility painted its whole text over the entries below it, and the panel only recovered on reopen.

The row preview was a `Text` with `.textSelection(.enabled)` and `.lineLimit(3)`. A click on a selectable `Text` swaps in the selection renderer, which lays the whole string out and ignores the line limit, so the row grew under the rows after it. Reproduced in an isolated SwiftUI harness with the panel's row and nested-scroll shape: same overflow with selection on, correct truncation with it off.

The row no longer offers inline selection. The full, selectable text is in the clipboard window, which the panel already opens. The other clipboard previews (quick window rows, preview sidebar) were checked: the quick window rows never had selection, and the sidebar shows the full text without a line limit, so neither has this defect.

What a user notices: expanding the Clipboard utility and clicking or dragging across a long entry keeps the list intact.

## Verification

Checked by hand in the developer build: drag across a long entry in the panel, the list stays laid out.

### DIFF
--- a/Sources/Vorssaint/UI/MenuPanel/PanelClipboardView.swift
+++ b/Sources/Vorssaint/UI/MenuPanel/PanelClipboardView.swift
@@ -138,11 +138,15 @@ struct PanelClipboardView: View {
     private func entryPreview(_ entry: ClipboardHistoryEntry) -> some View {
         switch entry.kind {
         case .text:
+            // Deliberately not selectable: clicking a selectable Text swaps in
+            // the selection renderer, which lays the whole preview out and
+            // ignores the line limit, so a long entry paints over the rows
+            // below it. The history window shows the full, selectable text.
             Text(entry.preview)
                 .font(.system(size: 10.5))
                 .lineLimit(3)
                 .truncationMode(.tail)
-                .textSelection(.enabled)
+                .frame(maxWidth: .infinity, alignment: .leading)
         case .image:
             HStack(alignment: .center, spacing: 7) {
                 if let name = entry.imageFile,


PathGao:
The removal is right, and the clipboard-side survey in the body holds up against `main`: `ClipboardQuickPanelView` rows never set `textSelection`, and the sidebar's text preview (`ClipboardEntryPreviewSidebar.swift:98`) has no line limit, so neither can grow the way the panel row did.

The same shape survives one card over, `PanelURLCleanerView.swift:117`:

```swift
Text(output)
    .font(.system(size: 10.5, design: .monospaced))
    .lineLimit(3)
    .truncationMode(.middle)
    .textSelection(.enabled)
```

Same panel, same utilities slot in `MenuPanelView`, same three-line cap plus selection. A long URL clicked there should overflow its card for the same reason. It is the only other `lineLimit` + `textSelection` pair left under `Sources/Vorssaint/UI/MenuPanel` — the rest of the pairs in the repo are settings and editor windows, not panel cards. I have not reproduced that one; it is shape matching, not a repro, and it is outside this PR's scope.

`.frame(maxWidth: .infinity, alignment: .leading)` looks like a no-op: the button row's `Spacer()` already stretches the card, and the enclosing `VStack` is `.leading`. No harm, just nothing behind it.

I read the diff and `main`; I did not build or run it, so the selection-renderer behaviour is taken from your harness result, not verified here.

**Impact:** panel rows lose inline selection, which the body declares; the full selectable text stays in the history window. No other behaviour change.
**Merge:** ship it.

PathGao:
The clipboard coverage claim checks out: `ClipboardQuickPanelView.swift` has no `textSelection` at all, and `ClipboardEntryPreviewSidebar.swift:98` has selection with no line limit. Neither can hit this.

Outside the clipboard, `lineLimit` + `.textSelection(.enabled)` on the same `Text` is still on main in 8 places. One is in this same panel:

`PanelURLCleanerView.swift:117`
```swift
Text(output)
    .font(.system(size: 10.5, design: .monospaced))
    .lineLimit(3)
    .truncationMode(.middle)
    .textSelection(.enabled)
```

Same shape, same popover, and a cleaned URL is exactly the long unbroken string that runs past three lines once the selection renderer takes over. Whether it overlaps or just pushes the panel depends on what sits under the utility section — I didn't check that, and I didn't run it. `URLCleanerSettings.swift:134` is that same block copied into the settings window, so it is two sites, not one.

The other six are `lineLimit(1)` URL rows (`ScreenshotSettings.swift:465`, `ScreenRecorderSettings.swift:343`, `ScreenshotEditorView.swift:1221`, `RecorderEditorView.swift:491`, `ScreenshotQuickPreviewController.swift:474`) and `MediaWorkspaceView.swift:484` at `lineLimit(5/8)`. You already have the isolated harness that reproduced this one; it is the cheap way to say which of those are real.

None of that blocks this.

**Impact:** none beyond the fix — the panel row loses inline selection, the history window keeps it.
**Merge:** ship it.

andreisuslov:
Thanks. Agreed on `PanelURLCleanerView.swift:117` being the same shape in the same slot; I left it out of this PR to keep it one topic and can send it on its own once this lands. The `.frame(maxWidth:)` is indeed doing nothing here and is gone in the follow-ups that touch this row.

PR #1113, opened 2026-08-28
fix(panel): open the hosted utility's own Settings page

andreisuslov:
## Summary

The menu panel's Settings button always opened the general Settings page, even while the panel was showing one utility (Clipboard, URL cleaner, Uninstaller, App updates, Window layout). Now the utilities section derives the hosted tool's Settings page from the same flags that decide `isHostingUtility`, hands it to `PanelInteractionState` while that tool is on screen, and the footer button routes there. From the panel's own lists the button lands on the general page (before, it reopened whichever page the router last held).

All nine hosted tools are covered by the one list (Recent captures goes to the Screenshot page). The page is cleared only where the section leaves the screen, in its `onDisappear` next to the keep-open reset (another section or a metric replacing it). Closing the popover does not clear it, on purpose: the hosting view lives on and the tool is still on screen when the panel reopens.

What a user notices: Settings from inside the Clipboard utility lands on Clipboard settings.

## Verification

Checked by hand in the developer build for Clipboard and URL cleaner; the rest were compiled and unit-tested only.

### DIFF
--- a/.github/workflows/issue-fix-status.yml
+++ b/.github/workflows/issue-fix-status.yml
@@ -7,11 +7,15 @@ name: Issue fix status
 #   comment
 #
 #   the next stable release                         ->  every issue still
-#   carrying that label becomes `please confirm`, and its reporter is asked
-#   to check that release
+#   carrying that label becomes `please confirm close`, and its reporter is
+#   asked to check that release
 #
 #   two weeks of silence after that                 ->  the issue is closed,
 #   after a comment saying so and how to reopen it
+#
+# A label put on by hand joins that at the second step: the daily job asks the
+# question the release sweep would have asked, and the same two weeks run from
+# that comment. Nothing is closed on a silence nobody was asked to break.
 #
 #   a description that closes its issue with a    ->  one comment on the pull
 #   keyword, or mentions one and references none        request, before it merges
@@ -45,16 +49,28 @@ name: Issue fix status
 # nothing either: one word from anyone stops it, and it never reads what that
 # word says.
 #
-# `please confirm` carries no version. The version lives in the comment, which
-# is the only part of this the reporter ever sees -- labels are silent, they
-# notify nobody, and a label per release would grow the repository's label list
-# without ever telling anyone anything.
+# `please confirm close` carries no version. The version lives in the comment,
+# which is the only part of this the reporter ever sees -- labels are silent,
+# they notify nobody, and a label per release would grow the repository's label
+# list without ever telling anyone anything.
+#
+# It says `close` because that is what the label is for and the only thing it
+# can lead to: an answer, or a close two weeks later. A bare `confirm` is a
+# bucket -- confirm a repro, confirm a design, confirm a duplicate -- and a
+# label that can mean three things is read as none of them.
 #
 # A `summary` issue is never touched by any of this, wherever it turns up: a
 # long-lived index has no reporter waiting on a build, and a pull request that
 # fixes one line of a fifty-line audit has not fixed the audit. They are also
 # the issues most often referenced, so the number that reaches these jobs most
 # is the one that must never be labelled or closed by them.
+#
+# A `half` issue is tracked on the same release timeline, but with tailored
+# wording: two root causes were reported together, one of them was fixed, and
+# the release sweep tells the reporter which part landed and that the rest
+# remains open. The same two-week silence timer applies, closing the issue if
+# the landed part was all that was needed, while any reply reopens it for
+# further work.
 #
 # Both labels and both comments are addressed to the person who filed the
 # report, and most of them are not developers. Nothing in what they read says
@@ -78,8 +94,10 @@ on:
     workflows: [Release]
     types: [completed]
   schedule:
-    # Once a day is enough for a two-week clock.
-    - cron: '23 7 * * *'
+    # Once a day is enough for a two-week clock, and late is as good as
+    # on time for one: GitHub queues the top of the hour heavily and runs
+    # it when it can, which costs this job nothing.
+    - cron: '0 7 * * *'
   workflow_dispatch:
 
 permissions:
@@ -94,8 +112,15 @@ env:
   GH_TOKEN: ${{ github.token }}
   GH_REPO: ${{ github.repository }}
   PENDING_LABEL: fixed (pending release)
-  CONFIRM_LABEL: please confirm
-  SKIP_LABEL: summary
+  CONFIRM_LABEL: please confirm close
+  # Labels this bot keeps its hands off entirely. JSON so a name with a space
+  # stays one name. `half` is deliberately not one of them: a report whose
+  # landed part reaches a release is asked and timed like any other, because
+  # that part may be the one the reporter wanted, and silence leaves nothing to
+  # act on -- a remainder that still bites someone comes back as its own report.
+  # It only changes what the release comment says, below.
+  SKIP_LABELS: '["summary"]'
+  HALF_LABEL: half
   CONFIRM_DAYS: 14
 
 jobs:
@@ -138,8 +163,10 @@ jobs:
             }
             [ "$(jq 'has("pull_request")' <<<"$issue")" = false ] || continue
             [ "$(jq -r '.state' <<<"$issue")" = open ] || continue
-            if jq -e --arg l "$SKIP_LABEL" 'any(.labels[]; .name == $l)' <<<"$issue" >/dev/null; then
-              echo "#$number: $SKIP_LABEL, left alone"
+            skip=$(jq -r --argjson skip "$SKIP_LABELS" \
+                     'first(.labels[].name | select(IN($skip[]))) // empty' <<<"$issue")
+            if [ -n "$skip" ]; then
+              echo "#$number: $skip, left alone"
               continue
             fi
             jq -e --arg l "$PENDING_LABEL" 'any(.labels[]; .name == $l)' <<<"$issue" >/dev/null && continue
@@ -149,7 +176,7 @@ jobs:
             gh api "repos/$GH_REPO/issues/$number/labels" -f "labels[]=$PENDING_LABEL" --silent
 
             # One fix status at a time. An issue its reporter sent back is still
-            # carrying `please confirm` from the last release, and a new fix puts
+            # carrying `please confirm close` from the last release, and a fix puts
             # it behind the next one.
             if jq -e --arg l "$CONFIRM_LABEL" 'any(.labels[]; .name == $l)' <<<"$issue" >/dev/null; then
               gh issue edit "$number" --remove-label "$CONFIRM_LABEL"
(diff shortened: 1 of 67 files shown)


PathGao:
Body says "all five hosted views that already mark `viewKeepsPopoverOpen`". The section hosts nine, and `MenuPanelView.swift:636` already names them: Uninstaller, Cleaner, URLCleaner, Homebrew, Media, Clipboard, RecentCaptures, WindowLayout, AppUpdates. Four are untouched — Cleaner (`:575`), Homebrew (`:583`), Media (`:587`), Recent captures (`:597`). They mark `viewKeepsPopoverOpen` from inside their own view (`CleanerView.swift:908`, `MediaWorkspaceView.swift:21`, `PanelRecentCapturesView.swift:12`) or via the section-level `onChange` at `:625`, so "already marks it" didn't select the set you meant. Pages exist for all four: `.cleaner`, `.homebrew`, `.media`, and `.screenshot` for recent captures.

The membership test is already computed in one place. Deriving it there covers all nine and deletes the 10 lines added to the five view files:

```swift
private var hostedSettingsPage: SettingsPage? {
    if showUninstaller { return .uninstaller }
    if showCleanerPanel { return .cleaner }
    if showURLCleaner { return .urlCleaner }
    if showHomebrewPanel { return .homebrew }
    if showMediaPanel { return .media }
    if showClipboardPanel { return .clipboard }
    if showRecentCapturesPanel { return .screenshot }
    if showWindowLayoutPanel { return .windowLayout }
    if showAppUpdatesPanel { return .appUpdates }
    return nil
}
```

mirrored next to the keep-open one at `:625`:

```swift
.onChange(of: hostedSettingsPage) { _, page in
    PanelInteractionState.shared.hostedSettingsPage = page
}
```

That also retires the onAppear/onDisappear ordering question: with paired hooks, any hosted→hosted branch switch (something flips `showUninstaller` while Clipboard is up) can run the new view's `onAppear` before the old view's `onDisappear` and leave the page nil.

Second, "The panel's own lists still open the general page" — they don't, and didn't before this PR either. `SettingsRouter.page` is a singleton nothing resets; every assignment site I found writes a specific page (CommandBar, RadialMenu, SnippetLibrary, the panel rows at `:1571`/`:1601`/`:1625`/`:1653`, `PanelAppUpdatesView.swift:39`). Open Clipboard → Settings → close the window → back to the list → Settings, and you get Clipboard settings again. The `if let` makes that stale page easier to reach now that the footer writes to the router. One line instead:

```swift
SettingsRouter.shared.page = PanelInteractionState.shared.hostedSettingsPage ?? .general
```

I read main's source, didn't build or run it. `.screenshot` as the destination for Recent captures is my reading of `SettingsSectionAnchor.screenshot` — confirm that's the page you want.

I couldn't find the issue this was reported on among the open ones. Link it if it exists.

**Impact:** Settings from inside Cleaner, Homebrew, Media, or Recent captures still lands on whatever page the router last held, as does Settings from the panel's own list.
**Merge:** needs the other four hosted utilities covered first.

PathGao:
Three things.

**The five isn't the whole set.** Body says "all five hosted views that already mark `viewKeepsPopoverOpen`". Grepping `main`, eight panel-hosted wrappers mark it: the five here plus `PanelCleanerView` (`CleanerView.swift:883`), `PanelMediaView` (`MediaWorkspaceView.swift:16`), `PanelRecentCapturesView` (`PanelRecentCapturesView.swift:6`). All three sit in the same `isHostingUtility` switch at `MenuPanelView.swift:571-611`, so Settings from inside them still lands on the general page. Homebrew is correctly out — `hostedUtilityKeepsPopoverOpen` excludes it. Recent captures is the awkward one: `.screenshot` or `.screenRecorder`, the surface shows both. If that ambiguity is why it was skipped, put that in the body — as written the sentence claims coverage the diff doesn't have.

**`hostedSettingsPage` isn't cleared where its sibling flag is.** `viewKeepsPopoverOpen` is reset in five places besides the hosted views' `onDisappear`: the dismiss closures at `MenuPanelView.swift:589/594/599/604/609`, the section's `onChange`/`onDisappear` at `:626`/`:629`, and `popoverDidClose` at `AppDelegate.swift:1091`. The new field has only the `onDisappear`. That last reset exists because closing the popover isn't trusted to run SwiftUI's `onDisappear` — if that's the reason, closing the panel while Clipboard is hosted leaves `hostedSettingsPage = .clipboard`, and the next Settings click from the panel's own list goes to Clipboard settings. One line next to 1091. I can't check whether `onDisappear` fires on popover close without a device; the existing defensive reset is the evidence, not a repro.

**"The panel's own lists still open the general page" isn't what the code does.** `SettingsRouter.page` is a sticky `@Published` on a singleton and `openSettingsWindow()` (`AppDelegate.swift:1316`) never resets it. The footer assigns only when `hostedSettingsPage != nil`, so once anything has written a page — this PR, `PanelAppUpdatesView.swift:39`, `MenuPanelView.swift:1571/1601/1625/1653`, the Command Bar — the root footer button reopens that page. Dropping the `if let` makes the sentence true:

```swift
SettingsRouter.shared.page = PanelInteractionState.shared.hostedSettingsPage ?? .general
```

That is a behavior change for the root button, which today keeps whatever page you left on. Your call which of the two is the one that's wrong.

**Impact:** as it stands the panel root's Settings button starts landing on the last hosted utility's page instead of where the user left off, on default config. The three uncovered utilities keep the reported bug.
**Merge:** needs the `popoverDidClose` clear first; the coverage gap and the general-page sentence are your call.

andreisuslov:
Rewritten as suggested, in `b56a5a4`:

- `hostedSettingsPage` is now derived in the utilities section from the same `show*` flags as `isHostingUtility`, so all nine hosted tools are covered (Recent captures goes to `.screenshot`), and the ten lines in the five view files are gone. It reaches `PanelInteractionState` through an `onChange` next to the keep-open one, plus the section's `onDisappear`.
- Cleared in `popoverDidClose` next to the keep-open reset, for the reason that reset exists.
- The footer sets `SettingsRouter.shared.page = hostedSettingsPage ?? .general`, so the root button now lands on the general page, which is what the body claimed and what I meant. That is the behaviour change you named; I think landing on a page the user did not ask for is the wrong one of the two.

No issue exists for this that I could find; it came from using the Clipboard utility. Diff is now +33/−1 across `MenuPanelView`, `PanelInteractionState` and `AppDelegate`.

PathGao:
The derivation matches: same nine flags as `isHostingUtility`, same order as the branch chain at `MenuPanelView.swift:571`. `.screenshot` is right — `SettingsPage` has no `screenRecorder` case, so the ambiguity I raised isn't one.

The `popoverDidClose` clear I asked for is wrong now, and that one's mine. It exists because `onDisappear` isn't trusted on close — i.e. the content view survives. `setUpPopover` builds one `NSHostingController` for the app's life, and `MenuPanelView` re-syncs on `.menuPanelWillShow` because `onAppear` doesn't re-fire. So: close the panel while Clipboard is hosted, reopen. `showClipboardPanel` is still true, the panel still shows Clipboard, but `hostedSettingsPage` is nil and `onChange` won't fire on an unchanged value. Settings lands on general again.

Drop it. Flags and mirror are both long-lived, so `onChange` covers every transition, and nil is only reachable when nothing is hosted.

If that reopen path is real, `viewKeepsPopoverOpen` has had the same hole since before this PR (`AppDelegate.swift:1091`, restored only by the sibling `onChange`). I can't check either without a device.

**Impact:** panel closed and reopened with a utility still on screen — Settings goes to general.
**Merge:** needs the `popoverDidClose` clear dropped first.

andreisuslov:
Dropped in `679069e`, and the comment on `hostedSettingsPage` now says why nothing clears it: the hosting view lives as long as the app and a closed panel keeps its tool on screen, so the `onChange` mirror is the only writer. The section's `onDisappear` clear went with it for the same reason. You are right that `viewKeepsPopoverOpen` has had the matching hole; I left it alone here since it is not this PR's topic.

PathGao:
The derivation and the dropped `popoverDidClose` clear are right. The section's `onDisappear` clear wasn't — I only asked you to drop the `popoverDidClose` one, and the new comment justifies both with "the hosting view lives as long as the app." That's `MenuPanelView`, not `UtilitiesSection`.

`navigablePanel` renders one section at a time (`MenuPanelView.swift:177`), and the nav strip above it stays live while a tool is hosted (`:320`). Tapping another section unmounts `UtilitiesSection` and destroys the nine `@State` flags — that is why the sibling reset at `:628` exists. Open Clipboard, tap System, click Settings: `hostedSettingsPage` is still `.clipboard` with nothing hosted. Going back to Utilities doesn't fix it; the fresh state computes nil and `onChange` doesn't fire on an initial value. (Read from main, not run.)

Put it back next to the sibling, and only there:

```swift
.onDisappear {
    PanelInteractionState.shared.viewKeepsPopoverOpen = false
    PanelInteractionState.shared.hostedSettingsPage = nil
}
```

Panel close still won't run it, which is what you want — the tool is still on screen.

**Impact:** switch sections, or open a metric, while a utility is hosted and Settings lands on that utility's page instead of general. Default config, one click.
**Merge:** needs the section's `onDisappear` clear restored.

andreisuslov:
Restored in `f5cc57c`, next to the keep-open reset in the section's `onDisappear` and only there; the `popoverDidClose` clear stays out. The comment on `hostedSettingsPage` now says where it is cleared and why a closed panel is not one of those places. You are right that the view I was describing as long-lived is `MenuPanelView`, not the section.

andreisuslov:
@PathGao every point raised here is addressed and pushed, main is merged in, tests pass and GitHub shows the branch as mergeable. From my side this is complete and ready to merge.

PathGao:
Diff is right. The derivation matches the branch chain at `MenuPanelView.swift:571` flag for flag and in order, the section's `onDisappear` is the only clear, and the comment there says why a closed panel isn't one.

The body didn't follow the code. It still says "the page is cleared when the popover closes, next to the existing keep-open reset" — that's the `popoverDidClose` clear from `b56a5a4`, dropped in `679069e`, and it states the opposite of what the diff now does. Squash-merged that sentence becomes the commit message, and it's the one thing here someone could later read and "fix" back into place.

Read from main, not run. I didn't check the AppDelegate menu-bar Settings item, which sets no page at all — pre-existing and not this PR's topic.

**Impact:** none on behavior; the record says the reverse of the code.
**Merge:** needs that sentence in the body corrected first.

andreisuslov:
Body corrected: it now says the page is cleared only in the section's `onDisappear`, and that closing the popover does not clear it, on purpose. Nothing in the code changed.

andreisuslov:
@vorssaint ready to merge whenever you have a moment. Main is merged in (branch is current), `./build.sh` finishes without warnings, `--selftest` passes and `./build.sh --test` is green; @PathGao's review points here are all addressed and answered above. This one is independent.

PathGao:
Body and code agree now — the clear is described where it actually is, and the reason a closed panel isn't one of those places is in both the body and the comment on `hostedSettingsPage`.

Re-read the whole diff, not just the delta: nine flags in `hostedSettingsPage`, same set and same order as the branch chain at `MenuPanelView.swift:571`; one clear, in the section's `onDisappear` next to the keep-open reset; the footer assigns unconditionally with `?? .general`. Nothing left over from the earlier rounds.

Read from main, not run. Per the body, hands-on checks covered Clipboard and URL cleaner; the other seven hosted tools are compiled and unit-tested only — the derivation is one list, so that's a reasonable place to stop, but it isn't the same as having seen them.

**Impact:** none outstanding.
**Merge:** ship it.

andreisuslov:
Thanks. Nothing further from my side; ready for @vorssaint.

PR #1114, opened 2026-08-28
feat(clipboard): open the window with the preview pane and show entry metadata

andreisuslov:
## Summary

Two things, both about seeing what a clipboard entry is without pasting it.

- **Open with the preview pane** (Settings → Clipboard): the ⌥Space window opens with its preview pane already showing the selected entry. Today the pane remembers its last state instead; the remembered state becomes this setting's initial value on update, so nobody loses an open pane. Refs #884.
- **Entry metadata**, in the preview pane and on the rows: what the entry is (text, link, image, files, with a glyph and label), character/line/word counts or image dimensions or file kind, size, full and relative copy time, the app it was copied from (with its icon), and the file location or link host.

Two more Settings-backed row changes, from use: **Keep every row the same height** (`clipboardUniformRows`, on by default, so on update text rows go from two lines to one and image rows to a 30 px thumbnail; the switch in Settings → Clipboard restores the old rows) shows one line of text and a small thumbnail per row, leaving the whole item to the preview pane; and the trailing column has one fixed width in both its states, so the hover buttons no longer reflow the text under the pointer.

**Space no longer opens the preview pane.** On main, Space toggled the pane once the arrow keys were driving the list; in this diff it marks the row the arrow keys are on, the same batch selection ⌘Return and ⌘-click build. The predicate that decides when the list owns Space is unchanged, only renamed, `ClipboardHistoryPreview.handlesSpace` to `ClipboardHistorySpaceKey.handledByList`, since it no longer answers a question about the preview. That leaves the pane with no keyboard toggle at all: the toolbar button and the pane's own close button are what remain, both mouse-only. The change and the reasoning for it are #1142's; the commit is on this branch because it was made here first, so whichever of the two merges first is the one that brings it.

The source app needs new capture code: the pasteboard never says who wrote it, so `ClipboardSourceApp` keeps the apps that held the front between two pasteboard polls and names the one that held it longest. Copying is usually followed at once by switching to the app the text goes into, so "whoever is in front when the poll notices" would name the destination; the longest-held app is the better guess. This is a label only: `ClipboardIgnoredApps`, which decides whether a copy is kept at all, is untouched. Entries saved before this decode with no source.

Built on: `NSWorkspace.didActivateApplicationNotification` and the entry's existing fields. No new permission, dependency or helper. Surface: one setting, one stored optional field, strings in all thirteen languages.

**Where `clipboardUniformRows` shows up.** The setting is on by default and it is read in two places, not one. In the ⌥Space window it does what the paragraph above says. In the menu bar panel it touches only the thumbnails: `PanelClipboardView` drops an image or single-image-file preview from 110×40 to 22×22, while its text rows keep the `.lineLimit(3)` they have today. So panel rows do not become uniform there, the images alone get smaller, and that is a visible change on update for anyone who uses the panel. Say the word and I will scope the key to the window instead, which leaves the panel exactly as it is now.

## Verification

Checked by hand in the developer build: the toggle opens the window with the pane up; metadata shows for text, a link, an image and a Finder file copy; the source app resolves to Finder/TextEdit for copies made there. Not tested: an older history file with no `source` field on a real user's Mac beyond the unit test that decodes one.

Refs #884


PathGao:
Three things.

**1. `sourceAppFormat` is translated into thirteen languages and never called.** The Source row is built from `text.sourceLabel` + `source.name`; "From %@" has no call site, and `MetricsTests` now pins the clipboard string count at 65, so dropping it later costs a test edit too. Unless it's staged for something later in this stack, cut it and make that 64.

**2. The Type row shows a name where a type belongs, for any extensionless file.**

```swift
let kind = UTType(filenameExtension: (path as NSString).pathExtension)?.localizedDescription
let dimensions = ClipboardImageStore.imageDimensionsLabel(atPath: path)
let details = [kind, dimensions].compactMap { $0 }
return details.isEmpty ? label : details.joined(separator: " · ")
```

Copy a folder in Finder: `pathExtension` is `""`, `UTType(filenameExtension: "")` is nil, `imageDimensionsLabel` is nil, `details` is empty, so Type falls back to `label` — which for a one-file entry is the file name. "Type: Downloads". Same for `Makefile`, `LICENSE`, anything without an extension.

The filesystem already answers this, and the repo already asks it that way — `ShelfTilesView.resolvedFileKind(for:)`:

```swift
let kind = (try? URL(fileURLWithPath: path)
    .resourceValues(forKeys: [.localizedTypeDescriptionKey]))?.localizedTypeDescription
```

Gives "Folder" for a folder, the Get Info text for everything else, and takes the new `import UniformTypeIdentifiers` back out.

**3. `isLink` is the third copy of the web-URL predicate in the clipboard module.** `ClipboardHistoryService.isWebURL(_:)` and `ClipboardHistorySensitiveText.isWebURL(_:)` both already say `scheme == "http" || scheme == "https"` + `host != nil`, and the second one is in the same file as the new property. Whitespace guard aside, dropping `private` off it and calling it is the whole fix. (`QuickToolsSupport`, `CommandBarLinks`, `RadialMenuSupport` and `URLCleaning` carry their own — not this PR's job, but this PR is where the clipboard module goes two to three.)

What I checked and what I didn't: only three files under `Sources/Vorssaint/UI` mention `ClipboardHistoryEntry` and all three are in the diff, so the glyph/capsule rollout has no missed sibling; `setHistoryRunning` has exactly two call sites and both got the `ClipboardSourceApp` line. The source-app heuristic I only read — the longest-held pick and its tie-break hold up against the unit tests, but nothing here tells me what it actually names on a real Mac. I can't see #1112, so I can't confirm the `.textSelection(.enabled)` removal in `PanelClipboardView` is the hunk you're carrying rather than a change of its own; `imageURL(named:)`'s comment ("for dragging a copied image out as a file" — nothing here drags) reads like it arrived from somewhere else too.

**Impact:** the Type row reads as the file's name for any folder or extensionless file, in a pane this PR makes open by default. The other two are dead weight, not behavior.
**Merge:** needs the Type row fixed first.

ref #884

andreisuslov:
All three done, in `3054912` on this branch:

1. `sourceAppFormat` removed from the struct, the thirteen catalogs and the test; count is 64.
2. Type row asks the filesystem (`localizedTypeDescriptionKey`), as `ShelfTilesView.resolvedFileKind` does, so a folder reads "Folder" and `Makefile` gets its Get Info kind; `UniformTypeIdentifiers` import gone.
3. `isLink` now calls `ClipboardHistorySensitiveText.isWebURL`, made internal; the module is back to two definitions (the service's `URL`-typed one is the other).

Also from the #1115 review, since it is this branch's half: the pane state is decided at open time (`quickPreviewPresented = byDefault`, assigned before `position()`), the setter no longer writes to disk, and the sticky `clipboardHistoryQuickPreview` key is gone from `DefaultsKey` and the registered defaults, with the test that pinned it moved to the new key. `fileSizeString` goes through `fileSizeBytes`. The split comment in `Defaults.swift` is restored, and `imageURL(named:)`'s comment now says what it is for here.

On the source-app heuristic: on this Mac a copy made in TextEdit while the clipboard window was open resolved to TextEdit, and a Finder path copy to Finder. That is two data points, not a survey.

PathGao:
Type row asks the filesystem, `isLink` calls the one predicate, the format string is gone.

The rename has nothing carrying the old key over. On main the setter writes `clipboardHistoryQuickPreview` and the service reads it at init, so anyone with the pane open today has `true` stored, and after this they get it closed with the new setting reading off. `register()` already runs sixteen of these; this is `migrateScrollInverterAxes` verbatim, in the pre-register batch:

```swift
guard defaults.object(forKey: DefaultsKey.clipboardHistoryQuickPreviewByDefault) == nil else { return }
defaults.set(defaults.bool(forKey: DefaultsKey.clipboardHistoryQuickPreview),
             forKey: DefaultsKey.clipboardHistoryQuickPreviewByDefault)
```

Keep the old constant the way `switcherShowWindowlessFinder` is. Side note: the body's "has to be opened by hand every time" isn't what main does — it sticks.

Smaller: `ClipboardKindPresentation.label` returns the file name for a one-file entry, so the new trailing label and the capsule repeat the name the row already prints, in the slot meant to say what the item is. Fixing it costs a singular string ×13 ("1 files") — your call.

**Impact:** anyone with the pane open loses it silently on upgrade; the rest is cosmetic.
**Merge:** needs the defaults migration first.

ref #884

andreisuslov:
Migration added in `2dcf90d`, `migrateClipboardQuickPreview`, in the pre-register batch next to the scroll one, and the old constant is kept with the same note `switcherShowWindowlessFinder` carries. Two checks in the test suite: an open pane carries over, and a newer choice is preserved. The body now says the pane sticks today rather than that it has to be opened each time.

Two more changes on this branch since, from using it: the metadata block's disk and workspace lookups now run once per entry, off the main thread (`ClipboardEntryFacts`), because stepping through the list with the arrow keys paid for them on every redraw and a big file cost visibly more than a word; and the Location row is gone for files, since the pane already shows the path below, which now has a copy button. The one-file label repeating the name I left; a singular string in thirteen languages for a slot that is otherwise right did not seem worth it, but I can add it if you would rather.

PathGao:
Migration is right, and it is `migrateScrollInverterAxes` verbatim — old key kept, pre-register, newer choice wins, both directions tested.

`facts` is `@State` and never cleared, so `.task(id:)` leaves the previous entry's values on screen until the new gather returns — longest exactly where you moved the work off the main thread, on a file. Arrow down from a text entry onto a Finder copy and the Source row shows the new app's name beside the old app's icon, with the old size and type. One line at the top of the task:

```swift
.task(id: entry.id) {
    facts = nil
```

Type already falls back to `ClipboardKindPresentation.label`, and the Size row hides itself when nil.

Smaller: `copyPlainText`'s "The history records it as its own entry" is the opposite of what happens — `writeToPasteboard` raises `lastChangeCount` precisely so the poll skips the app's own write. The behaviour is the right one; the sentence isn't.

GitHub currently reports the branch as conflicting.

**Impact:** the metadata block shows the previous entry's icon, size and type for as long as the disk takes, while stepping through the list.
**Merge:** needs the stale `facts` cleared first.

ref #884

andreisuslov:
Both in `4e3e238`: `facts = nil` is the first line of the task, so the rows fall back to the entry's own label and the Size row hides until the gather for this entry returns; and the `copyPlainText` comment now says what happens, that the poll skips the app's own write. The branch has main merged in (no rebase), so the conflict GitHub showed should be gone.

andreisuslov:
@PathGao every point raised here is addressed and pushed, main is merged in, tests pass and GitHub shows the branch as mergeable. From my side this is complete and ready to merge. This is the base of the clipboard stack; #1115 and #1116 follow it.

PathGao:
Migration, `facts = nil`, the `copyPlainText` comment: all three landed, and merging main in cleared the conflict.

`.task(id: entry.id)` doesn't re-fire on an edit. `updateText` writes `entries[index].text` in place and keeps the id, so after saving in the editor the text below updates and the metadata above it doesn't: Type still carries the old character, line and word counts, Size the old byte count, until you step to another entry and back. The entry is `Equatable`, so keying on it is enough:

```swift
.task(id: entry) {
```

Smaller: `ClipboardKindBadge`'s comment says the look is shared by "the rows, the ⌘K card and the preview". Only three files under `Sources/Vorssaint/UI` mention `ClipboardHistoryEntry` and none of them is the command bar, so there is no ⌘K card to share it with.

**Impact:** edit an entry in the preview and its own metadata block goes on describing the text it had before the edit.
**Merge:** needs the metadata re-gathered after an edit first.

ref #884

andreisuslov:
@vorssaint ready to merge whenever you have a moment. Main is merged in (branch is current), `./build.sh` finishes without warnings, `--selftest` passes and `./build.sh --test` is green; @PathGao's review points here are all addressed and answered above. This one is the base of the clipboard stack.

PathGao:
The `.task(id:)` key in `ClipboardEntryPreviewSidebar.metadata` is still `entry.id` in the diff I'm reading, so the edit case from last round is open — and no commit is named for it, unlike the three before it.

`updateText` writes `entries[index].text` in place and keeps the id, so after saving in the editor the text below updates and the metadata above it doesn't: Type stays on the old character, line and word counts and Size on the old byte count until you step to another entry and back. The entry is `Equatable`, so keying on it is the whole fix:

```swift
.task(id: entry) {
```

Everything else from last round is in: the migration, `facts = nil`, the `copyPlainText` comment, and merging main cleared the conflict.

**Impact:** edit an entry in the preview and its own metadata block goes on describing the text it had before the edit.
**Merge:** needs `.task(id: entry)` first.

ref #884

andreisuslov:
In `f369b3f`: the metadata task is keyed on the entry itself, `.task(id: entry)`, so an edit that keeps the id and changes the text refreshes the counts and size with it. The same commit keeps the selection on the file when its path is copied from the pane (below, on #1115).

PathGao:
`.task(id: entry)` is in, and with it every point from the four rounds before this one.

Space no longer opens the pane, it adds the row to the pile. Deliberate in the key monitor, but nowhere in the body, and two comments in this PR still say the old thing: `quickPreviewPresented`'s own doc ("Space and the toolbar button change it for that one showing"), and `ClipboardHistoryEscape`'s, which cites `ClipboardHistoryPreview.handlesSpace` — a symbol this PR deletes. What is left is the toolbar button and the pane's close button, both mouse only; there is no keyboard path to the pane any more.

`clipboardUniformRows` is on by default and also reaches the menu bar panel, where `PanelClipboardView` keeps `.lineLimit(3)` and only the thumbnail drops 110×40 → 22×22. Rows there do not become uniform, the images just get small, and the body describes only the ⌥Space window.

**Impact:** on update everyone loses Space-to-preview and gets 22 px thumbnails in the menu bar panel, neither of which the release notes will carry.
**Merge:** needs both written into the body, and the two Space comments corrected.

ref #884

andreisuslov:
Declared rather than reverted, since #1142 argues the change on its own merits and this branch only happens to carry the commit.

The body now says it: Space marks the row the arrow keys are on, the pane has no keyboard toggle left, and the toolbar button and the pane's own close button are what remain. It also names #1142 as the owner of the change and notes that whichever of the two merges first is the one that brings it.

Both comments are corrected in fdc41ee:

- `quickPreviewPresented`'s doc no longer claims Space changes it, and points at `ClipboardHistorySpaceKey`.
- `ClipboardHistoryEscape` no longer cites `ClipboardHistoryPreview.handlesSpace`. `grep -rn ClipboardHistoryPreview Sources/ Tests/` is now empty, so the deleted symbol is not named anywhere.

`./build.sh --test` green on this branch (9219 checks after the merge up, 9212 here).

Still open from your comment, not in this push: `clipboardUniformRows` reaching `PanelClipboardView`, where `.lineLimit(3)` stays and only the thumbnail goes 110×40 → 22×22. The body still describes only the ⌥Space window. Say whether you want the panel left as it is with the body corrected, or the setting scoped to the window so the panel is untouched, and I will do that next.

PathGao:
Both Space comments read right now, and the deleted symbol is gone from the tree.

Scope it to the window. The caption is what sells the tradeoff — "the preview pane shows the whole item" — and the menu bar panel has no preview pane, so a 22 px thumbnail there gives up the picture with nothing taking it back. Dropping the `uniformRows` static in `PanelClipboardView` and putting 110×40 back is a smaller change than writing body prose for a setting that only half applies.

While you are in that file: the new comment says "Double-clicking the row opens the full, selectable text in the history window instead." Nothing in `entryRow` takes a tap or a double-click — the way out of the panel is the `arrow.up.forward.app` button in the header. The sentence it replaced was true.

**Impact:** as it stands, a setting whose caption promises a preview pane shrinks thumbnails in a view that has none.
**Merge:** needs the panel scoped out first.

ref #884

andreisuslov:
Both merge conditions from your last pass are done.

**Space**: declared in the body, with the consequence stated plainly, the pane has no keyboard toggle left, only the toolbar button and the pane's own close button, both mouse. #1142 is credited as the owner of the change, and the body notes that whichever of the two merges first brings it.

**Both comments**: corrected in fdc41ee. `quickPreviewPresented`'s doc no longer claims Space changes it, and `ClipboardHistoryEscape` no longer cites `ClipboardHistoryPreview.handlesSpace`. `grep -rn ClipboardHistoryPreview Sources/ Tests/` is empty, so the deleted symbol is named nowhere.

**`clipboardUniformRows` in the panel**: in the body now, with your numbers, since you were right on both halves. `PanelClipboardView.swift:155` keeps `.lineLimit(3)` whatever the setting says, and lines 165 and 181 are the only readers of it there, taking a thumbnail from 110×40 to 22×22. So the body says panel rows do not become uniform, the images alone get smaller, and that this is visible on update for panel users.

I wrote it down because that is what you asked for, but the alternative is one line: scope the key to the ⌥Space window and leave the panel untouched, which removes the change instead of disclosing it. Say which you prefer and it is the next commit.

Current with main (0 behind), `./build.sh` clean, `--selftest` passes, `./build.sh --test` green at 9369 checks.

andreisuslov:
@vorssaint every point @PathGao raised across rounds 1 to 6 is now answered on all four, and each branch is 0 behind main with `./build.sh` clean, `--selftest` passing and `./build.sh --test` green.

Round 6 in particular: the Space rebind is declared in every body that carries it, the two comments that still described the old behaviour are corrected, `ClipboardHistoryPreview` is named nowhere, #1115's duplicate activation observer is deleted, and the `clipboardUniformRows` effect on the menu bar panel is written into #1114's body.

Merge order, since three of them stack:

1. #1114
2. #1115
3. #1116

#1142 is independent of the three and needs no ordering, with one caveat recorded in its body: it shares its commit with #1114, so whichever of those two merges first is the one that brings the Space change.

One open question, on #1114, and it is yours to settle rather than mine: `clipboardUniformRows` also shrinks the menu bar panel's thumbnails from 110×40 to 22×22 while that view's text rows keep `.lineLimit(3)`. @PathGao asked for it written into the body and it is. If you would rather the panel were left alone, scoping the key to the ⌥Space window is a one-line change and I will push it on request. Happy to hold the merge for that answer or to have it land as it is.

PathGao:
The body's panel paragraph is accurate: `.lineLimit(3)` stays whatever the setting says, and the two thumbnail frames are the only readers. That is what the code does.

But last round wasn't a request to write it down, it was the answer to the question you asked: scope it to the window. Drop the `uniformRows` static in `PanelClipboardView` and put 110×40 back at both call sites. A setting whose caption sells the tradeoff as "the preview pane shows the whole item" should not reach a view that has no preview pane.

Also still open, and not mentioned in your last two comments: the new comment above `Text(entry.preview)` in that same file says "Double-clicking the row opens the full, selectable text in the history window instead." Nothing there takes a tap or a double-click — the way out of the panel is the `arrow.up.forward.app` button in the header. The sentence it replaced was true.

GitHub currently reports the branch as conflicting again, after the merge up.

**Impact:** as it stands, a setting whose caption promises a preview pane shrinks thumbnails in a view that has none.
**Merge:** needs the panel scoped out first.

ref #884

andreisuslov:
Scoped `clipboardUniformRows` to the clipboard window. `PanelClipboardView` no longer reads the setting, both menu-panel thumbnail paths stay at 110×40, and the text comment again names the real history-window path. Added regression checks for both conditions in `5f71ed7`, then merged current `main` in `9faa40a`. Local verification on that head: `./build.sh --test` (9572 checks), `./build.sh`, and `./build/Vorssaint --selftest` all pass.

PathGao:
The panel is out: `PanelClipboardView` no longer reads the key, both thumbnail paths are back at 110×40, and the comment names the header button again. The two checks in `5f71ed7` pin both, so a later change can't quietly undo them. That was the last condition, and everything from the five rounds before it is still in.

One thing left over from an earlier round, never answered: `ClipboardKindPresentation`'s doc says the look is shared by "the rows, the ⌘K card and the preview". The ⌘K card is `CommandBarCatalog.clipboardRow`, and it keeps its own switch over `entry.kind` — `doc.on.clipboard`, `clipboardImage`, `filePath` — with no link case at all. So it is the rows, the panel capsule and the preview; the command bar is a second mapping of the same fact. Not worth holding a merge for, but the comment as written tells the next person there is only one.

Source-app naming and the metadata gather I read but did not run; nothing in them needed a device to check.

**Impact:** none.
**Merge:** ship it.

ref #884

andreisuslov:
@PathGao All requested changes are addressed, and both required checks are green on head ea198a0:

- Build & selftest
- Swift 6.0.3 compatibility

The PR is mergeable and good to go from my side. Thanks!

PathGao:
Code side is settled: everything from the six rounds is in, and re-reading the diff I have nothing new to raise.

One thing before it can land, and it contradicts the "mergeable" in your last comment: GitHub reports the branch as CONFLICTING on `ea198a0`. Green checks don't cover that — they ran on the merge result, not on the mergeability. Same state as after the previous merge up, so another merge of `main` and a push should clear it. I can't tell from here whether it's a real content conflict or a stale computation; the API only says conflicting.

**Impact:** none on behavior — the branch just can't be merged as it stands.
**Merge:** needs the conflict cleared first.

ref #884

andreisuslov:
@PathGao the PR is fully mergeable now. If you merge anything else, it will become unmergeable again.

andreisuslov:
@PathGao Not sure why you marked the PR incomplete, you stated that the PR itself is complete (shippable), it just wasn't ready because of the conflict, but it doesn't render it "incomplete".

"Not ready" != "Incomplete"

andreisuslov:
@vorssaint Conflicts are resolved in df269a1, and the local test suite passes (10,689 checks). GitHub now reports the PR as mergeable; please merge #1114 once the queued CI checks complete.

andreisuslov:
@PathGao merged current `main` (`c4975e2`) into this branch and resolved the conflict in `0da4099`. The resolution keeps main's shared localization suite in place of the old fixed clipboard string-count checks; the shared suite covers the clipboard catalog, including this PR's added strings.

Verified locally: `./build.sh --test` passes (38,250 checks, plus preference cleanup), `./build.sh` passes, and `./build/Vorssaint --selftest` passes. GitHub currently reports this head as **MERGEABLE**; Build & selftest and Swift 6.0.3 compatibility are still running.

Please take another look and merge once the required checks pass.

vorssaint:
closing this one, the pieces bundled here go past what the app should take on.
space stops opening the preview and marks the row instead, which leaves the pane with no keyboard way in at all, and that is a change to argue on its own, not inside this one.
the pane already remembers whether it was open the last time, so a setting that decides it at every opening takes that memory away and adds a control the page does not need.
every row changes shape for everyone on update, one line of text and a tiny thumbnail, behind a switch that is on by default, and a change of look that big has to be asked for, not shipped as a default.
naming the app a copy came from keeps an app switch watcher running whenever the history is on and still guesses wrong when the switch to paste is quick, so the label costs something at rest and cannot be trusted.
the kind of an item is now shown twice on every row of the window, as a colored square and a colored label, and again as a capsule in the menu bar panel, which is more decoration than a native list carries.
a smaller version could still land, a metadata block in the preview pane with the type, the size and the copy time only, with the rows, the settings page and the space key left as they are

PR #1115, opened 2026-08-28
feat(clipboard): add a ⌘K action card and open panel entries by double-click

andreisuslov:
## Summary

- **An action card in the clipboard window** for the selected entry: paste, copy, edit (text), look up in Dictionary (a word or short phrase), open (link), show in Finder and copy path (files), pin/unpin, move up/down (when the list allows), preview, delete. ⌘K opens it, and so does → (a Settings switch, on by default; ← closes). ↑/↓ move, ↩ runs, Esc closes; a row that already has a key shows it (↩, ⇧↩, ⌥P, ⌥⌫). Preview shows none, because Space no longer opens the pane. Same shape and keys as the Command bar's action list, so there is nothing new to learn. The footer carries a ⌘K button that opens the same card by mouse.
- **Double-click a row in the menu panel's Clipboard utility** closes the panel and opens the clipboard window on that entry with the preview pane up. Opened that way, the app itself is frontmost, so Return had nowhere to paste and only closed the window; the service now remembers the last other app to come to the front and pastes there. The panel rows show three lines at most, and until now the whole entry had no way in from there.

Why it belongs: the clipboard window already has nearly all these operations, spread across a context menu, a hover button and unlabeled keys; the card is the one place that lists them, as the Command bar does for its rows. Dictionary look-up goes through the `dict:` URL macOS registers, so it adds no dependency. Built on the existing key monitor and services, no new permission. Surface: one setting of this PR's own, `clipboardArrowOpensActions` (→ opens the card, on by default). The other two keys in the diff, `clipboardHistoryQuickPreviewByDefault` and `clipboardUniformRows`, belong to #1114 and are declared there; `clipboardUniformRows` is on by default and changes text rows from two lines to one, which #1114's body says. Strings in all thirteen languages.

**Space no longer opens the preview pane** in this diff, as in #1114's: it marks the row the arrow keys are on instead, the selection ⌘Return already builds. So the pane has no keyboard toggle left and the card's Preview row carries no key hint, which is why the list above drops the Space it used to name. The change is #1142's, declared in #1114 and argued in #1142; it is not this PR's.

The last other app to come to the front, which the double-click path pastes into, is tracked by `ClipboardSourceApp` rather than by a second activation observer of this PR's own. That observer answered a question the source tracker already followed, was registered from `init` whether or not the history ran, was never removed, and filtered on `.regular` where the other did not, so the two could disagree about which app was last in front. There is one observer now, and it keeps the `.regular` filter for the paste target: a paste needs an app with a UI to receive the keystroke, while a copy can come from an accessory.

Stacked on #1114; the diff includes it until that merges. #1112 has merged, so its `.textSelection(.enabled)` removal is no longer part of this diff.

## Verification

Checked by hand in the developer build: ⌘K on text, link, image and file entries; every action; arrows, Return and Esc; double-click from the panel lands on the entry with the pane open and the panel closed. Not tested: a language other than English on screen (strings compile for all thirteen).


PathGao:
Reviewed the whole diff including the #1114/#1112 parts, since it ships as one thing. Flagging which finding is probably whose.

**1. Opening the window on an entry silently flips a persisted preference.**

`showHistoryWindow(focusing:)` reaches `setQuickPreviewPresented(true)`, and that setter writes to disk:

```swift
func setQuickPreviewPresented(_ presented: Bool) {
    guard presented != quickPreviewPresented else { return }
    quickPreviewPresented = presented
    UserDefaults.standard.set(presented, forKey: DefaultsKey.clipboardHistoryQuickPreview)
```

So one double-click on a panel row turns the preview pane on for every later ⌥Space, at the larger window size, until the user presses Space to turn it back off. `.edit` in the ⌘K card does the same. The body says the double-click "opens the clipboard window on that entry with the preview pane up" — not that it changes what every subsequent open does.

It also makes the new setting unobservable. `clipboardHistoryQuickPreview` (sticky last state) and `clipboardHistoryQuickPreviewByDefault` (the new toggle) both answer "is the pane up when the window opens", and only one of them is a preference: with the toggle off and the sticky key true the pane still comes up, and after a single double-click the sticky key is true for good. Decide the state at open time instead of carrying it across:

```swift
quickPreviewPresented = entry != nil
    || UserDefaults.standard.bool(forKey: DefaultsKey.clipboardHistoryQuickPreviewByDefault)
```

Assigned directly, not through the setter — `position(panel)` on the next line already reads the flag for the size, as your comment there says. Space and the toolbar button then change it for the session, and `clipboardHistoryQuickPreview` has no job left.

**2. The focused entry is selected but never shown as selected.**

`showHistoryWindow` calls `resetQuickSelection()`, which sets `quickSelectionIsVisible = false`, and then sets `quickSelectionIndex`. The list draws with `isSelected: history.quickSelectionIsVisible && …` and `scrollSelectedEntry` returns early on the same flag. Double-click row 30 in the panel and the window opens scrolled to the top, nothing highlighted, with the preview pane showing row 30. One line in the `if let entry` branch:

```swift
quickSelectionIsVisible = true
```

**3. The link row reads "Open in Finder" in Korean.**

`case .openLink` borrows `L10n.shared.s.diskOpenInFinder`, which belongs to the Disk section. Twelve languages have it as "Open"/"Öffnen"/"打开" and get away with it; `Strings+Korean.swift:704` is `"Finder에서 열기"`. This diff already adds fourteen clipboard strings — add a fifteenth rather than borrowing that one. (`symbolName: "safari"` also names one browser for whatever the default happens to be; smaller.)

**Same fact, more copies** — all three are probably #1114's half:

- `ClipboardHistoryEntry.isLink` is the third definition of the web-URL test. `ClipboardHistoryService.isWebURL(_ url:)` (ClipboardHistoryService.swift:733) and `ClipboardHistorySupport.swift:525`'s `isWebURL(_ text:)` are both on main, and the new one differs from them: no trimming, and it rejects any whitespace. One of those three decides whether a copy is treated as sensitive.
- `ClipboardImageStore.fileSizeString(atPath:)` still carries the exact guard `fileSizeBytes(atPath:)` was extracted from, in the same file, in the same diff: `fileSizeBytes(atPath: path).map { ByteCountFormatter.string(fromByteCount: Int64($0), countStyle: .file) }`.
- `ClipboardQuickActions.movedIndex` is `CommandBarService.moveActionSelection` (CommandBarService.swift:1775) written again. Fine to leave, worth naming since the card is deliberately modeled on that list.

**Not verified:**

- I can't separate #1112's hunks from #1114's from this PR's in a squashed diff, so the routing above is a guess.
- `sourceAppFormat` ("From %@") is added in all thirteen languages and `MetricsTests` asserts its placeholder, but nothing reads it — the source row uses `sourceLabel` plus `source.name`. Either a later PR in the stack consumes it, or it is dead in thirteen files.
- In `Defaults.swift` the new key replaced the `// Auto clear: …` heading, leaving "Deliberately outside the clipboardHistory family" hanging off `clipboardHistoryQuickPreviewByDefault`. Looks accidental.
- Nothing here is checked on hardware: the panel resize, the overlay's hit area, and `NSWorkspace.didActivateApplicationNotification` ordering are read, not run.

**Impact:** anyone who double-clicks a panel row, or uses ⌘K → Edit, gets the preview pane on for every later ⌥Space and can never see the new setting's off state; double-clicking anything past the first screenful opens the window with no row highlighted and no scroll; Korean users get "Open in Finder" on a link.
**Merge:** needs the preview-state and link-label fixes first, then holds for #1114 and #1112.

ref #884

andreisuslov:
Fixed in `faceb36` (this branch also carries #1114's fixes, merged, not rebased):

1. Opening on an entry no longer touches a preference: `showHistoryWindow(focusing:)` assigns `quickPreviewPresented = entry != nil || byDefault` directly before `position()`, the setter does not persist, and the sticky key is gone (that half lives in #1114). ⌘K → Edit still opens the pane, for that showing only.
2. The focused entry sets `quickSelectionIsVisible = true`, so it draws selected and `scrollSelectedEntry` brings it into view.
3. The link action has its own string, `openLink` ("Open link", in all thirteen), and the symbol is `arrow.up.right.square` rather than one browser's.

`movedIndex` I left as is. `sourceAppFormat` is gone as you guessed; nothing later in the stack read it.

PathGao:
Three fixes landed as described, and `sourceAppFormat` is gone.

**⌘K → Edit still won't start editing when the pane is closed.** The sidebar is inserted and `quickEditRequest` is set in the same turn, and `onChange` doesn't see the value a view appears with. Whether it fires rests on the blocking `setFrame(animate:)` in `setQuickPreviewPresented` flushing SwiftUI first — with Reduce Motion on, that resize doesn't block. A miss is sticky: `clearQuickEditRequest` runs only from that handler, so Edit on the same entry can never fire again (same value, no change).

```swift
.onChange(of: history.quickEditRequest, initial: true) { _, requested in
```

Smaller: `.showInFinder` leaves the panel floating over Finder, where `.openLink` and paste call `hideHistoryWindow()` first. And the panel row dropping `.textSelection(.enabled)` is worth a line in the body.

Not verified: read, not run.

**Impact:** the pane is closed by default now, so ⌘K → Edit is the common path — it opens the pane and stops, and that entry's Edit stays dead for the session.
**Merge:** needs the Edit fix, then holds for #1114 and #1112.

ref #884

andreisuslov:
Both in `6a98024`: `.onChange(of: history.quickEditRequest, initial: true)` so Edit starts when the pane and the request arrive together, and Show in Finder hides the window first like paste and Open link. The body notes that the `.textSelection(.enabled)` removal belongs to #1112.

Also added on this branch since: a Copy path action for file entries (`9eed48a`), one path per line, through the same lane as any copy.

andreisuslov:
@PathGao every point raised here is addressed and pushed, main is merged in, tests pass and GitHub shows the branch as mergeable. From my side this is complete and ready to merge. Follows #1114.

PathGao:
The Edit fix is right — `initial: true` catches the sidebar and the request arriving in the same turn — and Show in Finder now matches paste.

**"Surface: one setting" — the diff registers three.** `clipboardHistoryQuickPreviewByDefault`, `clipboardArrowOpensActions`, `clipboardUniformRows`, the last two on by default. `clipboardUniformRows` isn't in the body at all: main draws text rows at `.lineLimit(2)`, so on update everyone's rows drop to one line. If it is #1114's, that body should carry it.

**Copy path never reaches the history.** `copyPlainText` goes through `writeToPasteboard`, which raises `lastChangeCount` so the poll skips the app's own write. The comment on `.copyPath` says the opposite.

**The ⌘K dim is the `.overlay` of the list column**, so the 280pt sidebar stays lit and clickable under a card that otherwise takes every key.

Not verified: read, not run; I still can't tell this PR's hunks from #1114's.

**Impact:** on update every text row in the clipboard window loses its second line, unannounced; Copy path leaves no history entry.
**Merge:** needs the row-height change declared or defaulted off, then holds for #1114 and #1112.

ref #884

andreisuslov:
In `5edc9f3`: the dim and card now overlay the whole content row (list and pane), so the sidebar is neither lit nor clickable under the card; and the `.copyPath` comment says what happens, that the poll skips the app's own write and no history entry results (the description says the same).

On the settings: this PR registers one key of its own, `clipboardArrowOpensActions`. `clipboardHistoryQuickPreviewByDefault` and `clipboardUniformRows` are #1114's, and #1114's description now states that `clipboardUniformRows` is on by default and takes text rows from two lines to one, with the switch that restores them. This description names all three and who owns which.

andreisuslov:
@vorssaint ready to merge whenever you have a moment. Main is merged in (branch is current), `./build.sh` finishes without warnings, `--selftest` passes and `./build.sh --test` is green; @PathGao's review points here are all addressed and answered above. This one is on #1114.

PathGao:
Both landed — the dim now takes the whole content row, and the three keys are named with owners across the two bodies.

**Copy path: the code and the description disagree now.** `copyPlainText` calls `promote` in the write's completion, so a path copy does get its own history entry. The description says the poll skips the app's own write and no entry results. The code's behavior is the better one; the sentence is what needs changing.

Smaller: the copy button beside the path in the pane doesn't close the window, so the new entry lands above the file entry while the selection is an index — the pane switches to the path you just copied while you're looking at it.

Not verified: read, not run; I still can't separate this PR's hunks from #1114's.

**Impact:** the description understates Copy path by one entry; the pane jumps under the in-pane copy button.
**Merge:** hold for #1114 and #1112.

ref #884

andreisuslov:
Both taken. The description now says what the code does: the path is recorded as a text entry on purpose, because the poll would skip the app's own write. And the pane's copy button keeps the selection: `copyPlainText(_:keepingSelectionOn:)` re-selects the file entry by id after the new row goes in at the top, so the pane stays on the file you copied from (`f369b3f` on #1114, merged here).

PathGao:
The path copy keeps the selection now, and the description matches the code.

**Space no longer opens the preview pane.** On main `kVK_Space` calls `toggleQuickPreview()`; here it calls `toggleSelectedQuickEntryBatchSelection()`, which ⌘Return already does. Batch-select gains a second key, the preview pane loses its only one — the toolbar button and the ⌘K row are what's left. Neither body says so, and this one says the opposite:

> each row shows the key that already does the same thing (↩, ⇧↩, ⌥P, Space, ⌥⌫)

No row carries a Space hint; `.preview` is `keyHint: nil`. Same reason the new doc comment on `quickPreviewPresented` ("Space and the toolbar button change it for that one showing") is already stale.

Smaller: `observeFrontApp()` adds a second `didActivateApplicationNotification` observer for the question `ClipboardSourceApp` already answers, registered from `init` whether the history runs or not and never removed — and its `.regular` filter differs, so the two disagree on who was last in front.

Not verified: read, not run; I still can't tell this PR's hunks from #1114's.

**Impact:** anyone who learned Space as Quick Look gets a batch selection instead, and the preview pane has no key left.
**Merge:** needs the Space change declared or reverted, then holds for #1114 and #1112.

ref #884

andreisuslov:
Declared, and the sentence that said the opposite is gone.

You were right that no row carries a Space hint: only `.paste`, `.copy`, `.preview`'s neighbour `⌥P` and `.delete` pass a `keyHint`, and the preview action passes `nil`. The body now reads "a row that already has a key shows it (↩, ⇧↩, ⌥P, ⌥⌫). Preview shows none, because Space no longer opens the pane", and a paragraph above it states the Space change, credits it to #1142, and says the pane keeps no keyboard toggle.

The two stale comments you found live in #1114's diff and are fixed there (fdc41ee), merged up to here. `ClipboardHistoryPreview` is no longer named anywhere in the tree.

`./build.sh --test` green, 9219 checks.

Still open, not in this push: the second `didActivateApplicationNotification` observer in `observeFrontApp()`. You are right that `ClipboardSourceApp` already answers the question, that this one is registered from `init` whether or not the history runs, that it is never removed, and that the `.regular` filter differs so the two can disagree. I would rather delete it and read the front app from `ClipboardSourceApp` than keep two answers; confirm and it is the next commit.

PathGao:
Space is declared now, and the two stale comments are gone with #1114 merged up.

**Not as stated — `ClipboardSourceApp` can't answer this one.** `sourceSinceLastCheck()` resets its window on every poll tick (0.8s), and `currentFront` returns nil while we are frontmost, so a second after the menu panel opens the window is empty. That is exactly when `rememberPasteTarget()` runs on the double-click path.

The merge that does work: keep `lastOtherFrontApp`, but move it onto `ClipboardSourceApp`, which already carries the observer, and read it from `rememberPasteTarget`. One registration, and the `.regular` / `!isTerminated` filter stays at the read — the source label wants accessory apps, the paste target does not.

One thing I haven't checked: `ClipboardSourceApp` observes only while the history runs, while the current observer is registered from `init`. If the window is reachable with the history off (radial menu, Command bar, quick launcher all call `showHistoryWindow()`), that path would lose its paste target.

**Impact:** none new — the two observers as they stand cost a duplicate registration, not a wrong paste.
**Merge:** hold for #1114 and #1112.

ref #884

andreisuslov:
Both of your points are in.

**Space**: declared. You were right that no row carries a Space hint, `.preview` passes `keyHint: nil` and only paste, copy, ⌥P and delete pass one. The body now reads "a row that already has a key shows it (↩, ⇧↩, ⌥P, ⌥⌫). Preview shows none, because Space no longer opens the pane", and the paragraph above it states the change and credits #1142.

**The second observer**: deleted in d0432c0. Every part of your reading held up, so rather than keep two answers I folded the paste target into `ClipboardSourceApp`, which was already following activations. `observeFrontApp()`, `frontAppObserver` and `lastOtherFrontApp` are gone from `ClipboardHistoryService`; `rememberPasteTarget()` reads `ClipboardSourceApp.shared.lastRegularFrontApp`. The `.regular` filter moved with it and is now the only one, so the two cannot disagree, and it follows the history the way the source window does, so nothing is observed while the history is off.

The one remaining `didActivateApplicationNotification` in `ClipboardHistoryService` is the pre-existing dismiss-on-deactivate observer that hides the open window. It is on main, it answers a different question, and it is removed when the window closes.

#1112 has merged, so its hunk is out of this diff and the body no longer claims it. Only #1114 is ahead of this one now.

Current with main (0 behind), `./build.sh --test` green at 9369 checks.

PathGao:
The observer merge is right: one registration, one `.regular` filter at the read, and it follows the history the way the source window does.

**That last part is where it breaks.** `CommandBarClipboardAccess.canUseHistory` is `captureEnabled || hasSavedItems`, and MetricsTests pins it ("saved clipboard items stay available when capture is off"). The panel's Clipboard utility draws its list off `history.entries` with no capture gate, so the double-click row is live with capture off — and `setHistoryRunning(false)` has by then torn the observer down and nil'd `lastRegularFrontApp`. `frontmostApplication` is us at that moment, so Return closes the window with nowhere to paste: the case this PR exists to fix. Same from the radial menu, the Command bar row and the Settings button.

The paste target isn't a history question, it's a window question, and the window opens either way. Let the observer run for the lifetime of `ClipboardSourceApp` and keep `guard historyIsRunning` on `sourceSinceLastCheck` alone.

Also: GitHub reports this branch CONFLICTING, not the 0-behind state above.

Not verified: read, not run.

**Impact:** with capture off and saved entries, double-click then Return pastes nowhere.
**Merge:** needs the paste target on the capture-off path first, then holds for #1114.

ref #884

PathGao:
Sending this back for one thing before it can go further; it is the case the PR exists for.

`ClipboardSourceApp.setHistoryRunning(false)` removes the activation observer and sets `lastRegularFrontApp = nil`. `rememberPasteTarget()` falls back to that value. The clipboard window still opens with capture off when there are saved entries, so in that state double-click → Return resolves `pasteTargetApp = nil` and the paste goes nowhere. The paste target's lifetime is tied to capture, which is a different feature from paste; either keep tracking the front app independently of the capture switch, or read `NSWorkspace.shared.frontmostApplication` at the moment the window opens as the fallback.

This was named on 2026-08-31 and nothing has been pushed since; the branch is also conflicting now. Fix, merge up on #1114, and it is reviewable again.

PR #1116, opened 2026-08-28
feat(clipboard): drag entries out and shake to open the Shelf

andreisuslov:
## Summary

Two small additions that arrived on this branch from use: the action card's Preview is now **Quick Look**, the system's own preview panel (Finder's Space), so an image, a file or a long text gets a window of its own rather than the pane beside the list (text goes through a temporary file that is replaced on the next look and removed on quit); and the window opens with the **first row selected**, so Return pastes it and ↓ goes to the second.

Copied files and images can be dragged out of the clipboard window as real files, and shaking while dragging opens the Shelf. Text can join in with two Settings switches, both off by default: drag text entries out as text (to anything that takes a drop, not only the Shelf; shaking during the drag opens the Shelf when it is on), and offer "Add to Shelf" for text in the ⌘K card. Neither switch waits for the Shelf to be on. Files and images always get "Add to Shelf" while the Shelf is on.

How: an AppKit drag source (`ClipboardFileDragSource`) on file/image rows, because SwiftUI's `.onDrag` carries one item provider and never reports the pointer. Several files drag as several file URLs under their own icons, Finder-style; a copied image drags as its stored PNG. The Shelf's shake-to-open ran only on a global event monitor, which never receives events from the app's own drags, so the shake rule moved into `ShelfShakeDetector` (same thresholds, now unit-tested) and the drag source feeds it from `draggingSession(_:movedTo:)`. The global path uses the same detector, unchanged behaviour. A press that never becomes a drag is handed back as the row's click, and right-click goes up the responder chain so the context menu still opens.

Built on the Shelf's existing `accept(pasteboard:)` and drop handling. No new permission or dependency. Surface: two settings, strings in all thirteen languages.

**Space no longer opens the preview pane** in this diff either: it marks the row the arrow keys are on, so the pane keeps no keyboard toggle. That is #1142's change, carried up the stack from #1114 and declared in both; neither the code nor the reasoning is this PR's.

Stacked on #1115 (on #1114); the diff includes them until they merge. #1112 has merged and is no longer part of it.

## Verification

Checked by hand in the developer build: drag a Finder file copy and an image copy out to Finder and to the Shelf; shake during the drag opens the Shelf; with the text switches on, a text entry drags and "Add to Shelf" appears for it; with them off neither does. Not tested: a second display or a fullscreen Space during the drag.


PathGao:
Three things, one of them a real defect.

**A shelved copied image dies when its history entry ages out.**

`dragItems(for:)` hands an `.image` entry the URL of its PNG inside `ClipboardImageStore` (`ClipboardImages/`), and `addToShelf` passes the same URL to `add(fileURLs:)`. The Shelf keeps foreign file URLs by reference — `items(from:)` → `fileItem(for:)` → `.file(url)`, and `isShelfOwnedFile` only claims paths under `ShelfFiles/` or the shelf temp dir, so it neither copies nor deletes. The clipboard, meanwhile, ends every `persist()` with

```swift
ClipboardImageStore.cleanup(keeping: Set(self.entries.compactMap(\.imageFile)))
```

Shelve a copied image → its entry falls off `clipboardHistoryLimit` (or is deleted, or cleared) → the next save deletes the PNG → the Shelf tile points at nothing and stays on screen. `try? url.bookmarkData()` doesn't survive a delete, and the `fileItem` comment covers move and rename, not removal.

Finder is unaffected: a drop copies at drop time. It's the Shelf specifically, which is the flow this PR exists for.

The Shelf already owns this case for a pasteboard image with no file URL — `imageItem(for:)` → `storePayloadData` writes a copy under `ShelfFiles/`. But `fileURLs(from:)` is checked first in both `accept(pasteboard:)` and `items(from:)`, so while the drag carries the store URL that path is unreachable. Keeping Finder working probably means fixing it on the Shelf side (copy a dropped file that lives under `ClipboardImages/`) rather than changing what the drag writes; the ⌘K action on its own could just hand over the PNG data.

**Only one of the three shelf drag affordances got an internal path.**

`noteInternalContentDrag` covers shake. `handleDragForDock` (`shelfDropZoneEnabled`, default true) and `handleDragForEdge` also run only off the global monitor, which by this PR's own premise never sees a drag that starts in the app. On stock settings shake and the drop zone are both on, and only shake answers a drag out of the clipboard window. The body claims shake only, so this isn't a contradiction — but the drop zone is the on-by-default one and it just won't appear.

**Double-click on file and image rows — I could not verify this.** `ClipboardFileDragView` takes `mouseDown`/`mouseUp` and reports back through `onClick(modifiers)`, which drops `clickCount`. The row wiring is past the truncation point of the diff I can read, so I don't know what happens to the new "Double-click to see the whole item" path on exactly the rows that carry this view. If that is a SwiftUI gesture sitting under the drag view, a double-click becomes two `activate` calls. Worth a check.

Also: `ShelfService.lastSummon` is dead now — its only two uses were the lines `handleDrag` gave up, and Swift doesn't warn on an unused stored property.

Not verified: anything on real hardware, plus the Settings UI, the row wiring and the tests — all past the cut in the diff I can see.

**Impact:** a copied image added to the Shelf goes dead silently once its clipboard entry ages out, and the tile stays. The rest is scope.
**Merge:** needs the shelved-image lifetime settled first.

andreisuslov:
Thanks, the first one was real. Pushed a fix commit on this branch.

**Shelved copied image.** Fixed on the Shelf side, as suggested, so Finder drops keep working unchanged: `fileItem(for:)` now runs a fresh URL through `adoptedIfOnLoan`, which copies any file living under `ClipboardImageStore.directory` into the shelf's own store via `storePayloadData` before the tile is made. It applies only when no bookmark is passed, so restored tiles keep pointing at their own copy. One guard in the one function every entry path reaches (`accept(pasteboard:)`, `items(from:)`, the drop resolver) rather than one per caller. The copy lands under `ShelfFiles/`, so `isShelfOwnedFile` claims it and it is deleted with the tile. The ⌘K action and the drag both keep handing over the store URL, so there is one mechanism.

**Drop zone and edge peek.** `noteInternalContentDrag(at:)` now drives all three: shake, `handleDragForDock(contentKnown: true)` and `handleDragForEdge(timestamp:contentKnown:)`. `contentKnown` skips the drag-pasteboard baseline check that the global path needs, since a drag the app started is known to carry content before its first event. `endInternalContentDrag()` ends the docked drag and the edge peek the way a mouse-up on the global monitor does. Global behaviour is unchanged apart from the parameter.

**Double-click on file and image rows.** The "Double-click to see the whole item" path is on the menu panel's rows (`PanelClipboardView`), which do not carry the drag view; the drag view sits only on rows in the clipboard window (`ClipboardQuickPanelView`). In the window a row's gesture is a single-click `activate`, and the drag view's `onClick` replaces it on the covered area rather than sitting on top of it (the AppKit view takes the mouse-down, so SwiftUI's tap never fires). A double-click there produces what it did before: two activates, the first of which hides the window.

**`lastSummon`.** Gone; the detector carries its own cooldown.

Verified: `./build.sh --test` (8996 checks), `./build.sh --dev --install` builds and launches. Not verified by hand yet: the shelved image surviving a history clear, and the drop zone or edge peek appearing during a drag out of the window. I will confirm those on screen and say so here.

PathGao:
The adoption guard sits at the chokepoint — `addFileBatch` reaches `fileItem(for:)` too, so one line covers all three entry paths, and `bookmark == nil` keeps restores off it.

**`noteInternalContentDrag` gates on a stale source app.** `handleDragForDock`/`handleDragForEdge` guard on `automaticOpenAllowed` → `dragSourceBundleIdentifier`, which only `beginDragGesture` sets and only `stopDragMonitor` clears. An internal drag never sets it, so it still holds whatever app the mouse was last pressed in. With that app in the shelf's automatic exclusions, the drop zone and edge peek stay off for a drag out of the clipboard window. Shake skips the guard, so the three disagree inside one function. `dragSourceBundleIdentifier = nil` at the top of the function covers it.

**Adopted tiles are named by UUID.** `fileItem` falls back to `url.lastPathComponent` and the copy is `UUID().uuidString + ".png"`, so a shelved copied image reads `8F3A-….png` where a pasted one reads "Image". Pass `title:`.

Settings, the row wiring and the tests are still past the cut in the diff I can read; nothing verified on hardware.

**Impact:** with any exclusion configured, two of the three shelf affordances never appear for a drag out of the clipboard window. Otherwise cosmetic.
**Merge:** needs the stale `dragSourceBundleIdentifier` first.

andreisuslov:
Both in `283af4b`: `noteInternalContentDrag` clears `dragSourceBundleIdentifier` first, so an exclusion on whichever app the mouse was last pressed in no longer gates the drop zone and edge peek for a drag out of the window; and an adopted copy takes the title a pasted image gets, so the tile reads "Image" rather than a UUID.

Two more on the same branch: the preview pane's thumbnail or file icon drags out the same way the row does, and the drag image now starts centred on the pointer rather than on the row, which had it sliding in from the row's middle.

PathGao:
The `dragSourceBundleIdentifier = nil` placement holds up: `allowsAutomaticOpen` returns true on a nil source (`ShelfSupport.swift`), so an internal drag now clears the guard instead of inheriting whatever app the mouse was last pressed in, and the global path re-sets it on the next `beginDragGesture`. Both gaps closed as described — though I can't read `283af4b` itself, the diff I get is still cut at the same place.

One question on the new preview-pane drag: does the thumbnail/file icon go through `ClipboardFileDragSource`, or is it a separate `.onDrag`? Only the first feeds `noteInternalContentDrag`, and without it that drag gets neither shake nor the drop zone nor the edge peek — the same split this PR just closed for rows. If it shares the drag source, ignore this.

The two by-hand checks you named are the acceptance tests for both fixes; nothing I can read replaces them.

**Impact:** none confirmed. The preview-pane question is unverified, not a defect.
**Merge:** hold for #1115.

andreisuslov:
It shares the drag source. `PreviewDraggable` in `ClipboardEntryPreviewSidebar` is a modifier that overlays the same `ClipboardFileDragSource` the rows use (built from `dragItems(for:)` for an image entry, or the single file URL for a file entry), so the pane's drags report to `noteInternalContentDrag` like the rows' and get shake, drop zone and edge peek the same way. There is no `.onDrag` anywhere in the clipboard UI.

The two by-hand checks are still owed; I will post them here once done on this Mac.

andreisuslov:
Checked by hand on this Mac (developer build, Shelf on): a copied screenshot added to the Shelf through ⌘K keeps its tile, with the image, after Clear recent empties the history. The drop zone during a drag out of the window is still being checked.

andreisuslov:
Second check done by hand on the same Mac: dragging a file row out of the clipboard window toward the menu bar icon shows the Shelf's drop target under the icon, and shaking during that drag opens the Shelf. Both by-hand checks for this branch pass; nothing else is owed here.

andreisuslov:
@PathGao every point raised here is addressed and pushed, main is merged in, tests pass and GitHub shows the branch as mergeable. From my side this is complete and ready to merge. Follows #1115; both by-hand checks are recorded above.

PathGao:
Answered. On `main` the `.onDrag` sites are Screenshot, Recorder, Shelf tiles, Radial/Monitor settings and PanelLayout — none under the clipboard UI, so the preview pane going through `ClipboardFileDragSource` means rows and pane share one drag path and one `noteInternalContentDrag`. Nothing left from my side on the code I can read.

The two by-hand checks were the acceptance tests for both fixes and they're recorded above; that closes what I asked for.

The two additions from the last push — the preview-pane drag itself and the pointer-centred drag image — are past the cut in the diff I get, so I haven't read them.

Still stacked, though: your own comment says this follows #1115, and none of this branch is on `main` yet. That's the one thing between here and merge, not anything in the change.

**Impact:** none.
**Merge:** hold for #1115.

andreisuslov:
@vorssaint ready to merge whenever you have a moment. Main is merged in (branch is current), `./build.sh` finishes without warnings, `--selftest` passes and `./build.sh --test` is green; @PathGao's review points here are all addressed and answered above. This one is on #1115.

andreisuslov:
This branch carries the same Space change up the stack from #1114, so its body now declares it too: Space marks the row the arrow keys are on, and the preview pane keeps no keyboard toggle. The code and the reasoning belong to #1142; nothing about it is this PR's.

`./build.sh --test` green, 9224 checks.

PathGao:
The body now declares the Space change, and the ⌘K list reuses the Command bar's shape, so there's nothing new to learn in it.

One thing I can read but can't test: the panel's dismiss monitors are unchanged from `main`, and they don't know about Quick Look. `installDismissMonitors` hides the window on any mouse-down whose `event.window` isn't the clipboard panel and whose location is outside it — a click inside `QLPreviewPanel` is exactly that, and `hideHistoryWindow()` now calls `ClipboardQuickLook.shared.hide()`. So the first click in the Quick Look window should close it. The key monitor has the same shape: it stays installed while Quick Look is key, so ↑/↓ move the list underneath instead of paging the panel, and Space marks a row rather than closing it — Finder's Space does close it.

Smaller: the `.preview` action's comment still describes the pane as the one "which Space already toggles", which this PR just made false.

Settings, the row wiring and the tests are still past the cut in the diff I get; nothing verified on hardware.

**Impact:** clicking or arrowing inside Quick Look doesn't do what Quick Look does. Code-read, not confirmed on a Mac.
**Merge:** needs the Quick Look dismissal settled first; still stacked on #1115.

andreisuslov:
Nothing of this PR's own changed. The branch carries #1115's observer fix and #1114's comment corrections up the stack, is current with main, and #1112 has merged so its hunk is out of the diff and the body no longer claims it.

`./build.sh --test` green at 9369 checks.

PathGao:
Carrying #1115's observer fix and #1114's comment corrections up the stack without touching this PR's own change is the right call — nothing new to re-read here.

One correction: GitHub has this branch as CONFLICTING right now, not current with main. Something landed between your merge and now, so the merge has to be redone before the mergeable state means anything.

Still open from last time, neither addressed in this push: by code-read, the panel's dismiss monitors close Quick Look on the first click inside it and keep taking ↑/↓/Space while it is key; and the `.preview` action's comment still describes the pane as the one "which Space already toggles".

**Impact:** none new.
**Merge:** needs the conflict resolved and the Quick Look dismissal settled; still stacked on #1115.

PathGao:
Sending this back for the Quick Look dismissal, which is still open from 2026-08-31.

`hideHistoryWindow()` now calls `ClipboardQuickLook.shared.hide()`, but `installDismissMonitors` is unchanged: a click inside the `QLPreviewPanel` is `event.window !== panel` and outside the history window, so the monitor calls `hideHistoryWindow()` and the preview closes on the first click inside it. The monitor needs to treat the QL panel as inside, or Quick Look needs its own dismissal rule; either way ↑/↓/Space should keep driving the list while the panel is up.

Conflicting since main moved; this sits on #1115, which has its own open item, so the order is #1114 → #1115 → #1116.

PR #1120, opened 2026-08-29
fix: stop features setting each other's Accessibility timeout

PathGao:
## Summary

`AXUIElementSetMessagingTimeout` on the system-wide element is the default for
every Accessibility question this process asks, whoever asks it — the header
says so, and it is measurable: a new element inherits whichever global was
written last. Five features wrote it. Each reads locally as that feature's own
limit, and none of them was one. A sixth file, `DockPreviewService`, set no
timeout anywhere and simply inherited whatever the others had left. After this
the global has one writer: `AppDelegate`, once at launch, which is the only one
that meant it.

What that costs the person using the Mac: when some app stops answering, how
long Vorssaint waits before giving up depended on which unrelated feature they
had used last. Hover a Dock icon after using Finder cut/paste and it waits
0.15s; hover it after clicking a green maximize button and it waits 0.35s. Same
action, 2.3× apart, for a reason nothing on screen explains.

The sharpest case is the last row: focus-follows-mouse asks Accessibility on a
background queue, so how long it waits costs the user nothing — a slow answer
just defers a focus change to the next tick. It was setting the value for three
features that stall the main thread inside an event tap, where the same wait is
a frozen cursor and a click that looks dropped.

| | before | after |
|---|---|---|
| `WindowMaximizer` | writes the global 0.35, then hit-tests | writes nothing; the hit test inherits the launch default (also 0.35), everything it reaches capped 0.35 |
| `AutoQuitService` | writes the global 0.15, then hit-tests | writes nothing; **the hit test moves 0.15 → 0.35**, everything it reaches capped 0.15 |
| `WindowLayoutService` | writes the global 0.25, then hit-tests | writes nothing; that one read inherits the launch default |
| `FocusFollowsMouseService` | writes the global 0.25 at init | writes nothing; same |
| `FinderCutPaste` | writes the global 0.15, then asks system-wide for the focus | asks Finder's element, capped 0.15 |
| `DockPreviewService` | sets nothing anywhere; inherits | caps every element it obtains, 0.35 |
| `AppDelegate` | writes the global 0.35 once at launch | unchanged — the deliberate floor |

**No hit test is narrowed here.** A hit test can carry its own cap only when it
is asked of an application element, and that form answers for the app's own
hierarchy whatever is drawn over the point — measured with two helper windows on
one rect, one at layer 0 and one floating above: the system-wide call returns the
floating window, the layer-0 owner's application element returns its own. Each of
these paths has a caller that needs the answer to be what is actually on top:
`WindowServerWindowHitTest.candidate` skips windows that are not layer 0, so on
the traffic-light paths the `pid(of: window) == candidate.pid` comparison is what
rejects a click that landed on a panel over the window; `dockOwnsPoint` skips a
window of *ours* on the point rather than rejecting it, so a pinned preview panel
over the Dock strip reaches the hover hit test and the same comparison is what
stops it opening a preview for the icon beneath. Narrowing any of them would
remove the rejection, not the hit.

So every hit test stays on the system-wide element, writes no timeout, and
inherits the launch default for that one read. Everything reached through it
carries the value its feature chose.

The Finder guard is the one exception, and it has no hit test: it asks for the
focused element, and its caller has already established that Finder is frontmost,
so the app being asked is the same app either way.

**The Dock preview is the other half of the same rule.** #938 counts it as the
one file that queries Accessibility and caps nothing, and it is the worst place
to inherit: its reads run on the main thread on every mouse-move sample along
the Dock's edge, and the process they question is the Dock, which is the one
that stops answering in #971. It now caps every element it obtains, at the two
doors an element comes through — the hit test, and the return of
`elementAttribute`, the only place a parent element is produced — so a read
added later is capped without anyone remembering to. Its hit test goes through
the Dock's application element for the same reason as the two above.

**Three reads move, and they are the ones that cannot own a value.** A hit test
can carry its own cap only if it is asked of an application element, and that
form ignores occlusion — so on these paths the choice is: own the value, keep
the occlusion check, or write the global. Two of the three, and the global is
what this change removes. So the hit test in `AutoQuitService` goes from 0.15 to
0.35, `WindowLayoutService`'s from 0.25 to 0.35, and focus-follows-mouse's from
0.25 to 0.35 on a background queue where it stalls nothing. Against a hung app
that is one read per click or press, 2.3× longer in the auto-quit tap than
before. Restoring it means restoring a global write, or dropping an occlusion
check that a click on another app's panel over a window currently relies on.

One thing this does across the whole app, not only these files: an
Accessibility read that caps nothing has always taken the process-wide default,
and that default used to be whichever of 0.35, 0.25 and 0.15 ran most recently.
It is now the launch default, always. Swept for reads that would narrow to
widen: seven places produce an element without capping it, and only
`FinderRenameService.focusedRole` sat under code stating a tighter intent — it
caps its own now. One states 0.35 already, one is capped on the next line, and
four state nothing and were always on whatever the global held.

**Every other value is unchanged.** This is about which code owns a number, not
what the number should be. Holding that takes capping every element on a path, not only
the one it starts from: a fresh element takes the process-wide default rather
than the cap on the element it was read from — with the global at 0.5 and an
application element capped at 1.2, a window read off that element gives up at
0.505s, and at 1.205s once capped itself. So the two paths that wait 0.15s cap
what they derive — the focused element the Finder guard reads a role off, and
the window, parents and close button auto-quit walks to. Those reads were
getting 0.15 from the global before; without that write they would have started
taking `AppDelegate`'s 0.35, inside the taps that a click and every Finder
keystroke wait on. Whether a Dock hover and a ⌘-drag should wait the same length
of time is #938's question and wants an answer there, not a rider here.

**The Finder guard was the one worth being careful about.** It decides whether
⌘X cuts files or edits text, and it fails toward cutting: any Accessibility
error means "not editing", so a wrong answer there cuts files while someone is
renaming one. Asking Finder's element rather than the system-wide one had to be
the same question, not merely a similar one. Measured on 26.6.2, driving Finder
through both states — file list focused, Go to Folder field focused — the two
reads return the same role each time, and the caller has already established
that Finder is frontmost, so the app whose focus is being asked about is the
same app either way. The rename field itself was not exercised; it is the same
`AXTextField` the Go to Folder field is.

With every element on these paths carrying its own value, changing what any one
feature waits is now a change to that feature alone, which is what #938's first
question has to be settled before anyone can do. It is not yet one line: the
value is written wherever the feature produces an element, and collecting it
into a constant the way `DockPreviewService` does is a separate change that
moves no value either.

**One thing dropped rather than corrected.** Four comments justified their caps
against "the 6 second default". An element with no timeout set gives up after
1.5s — measured against a bundled, signed application holding its main thread,
three runs within 10ms of each other, a 0.4s control tracking exactly. The
figure is removed rather than replaced: Apple documents no default for
`AXUIElementSetMessagingTimeout`, so any number written here is an observation
about one OS version and rots the way this one did. What those comments need is
that their cap is shorter than the default, which is what they now say.

**The guardrail** is #938's fourth question, in two rules that a spelling cannot
retire. Which files may name `AXUIElementCreateSystemWide` at all: five, each
listed with why — `AppDelegate`, which writes it, and the ones that hold it for a
hit test and must not. A file that never names it cannot write the global
whatever it calls its variables, so that list is the whole surface. And what may
be done with it there: outside `AppDelegate` it may only be bound to a name,
never passed to another function, because a helper that caps what it is given
turns one argument into a global write with no `AXUIElementSetMessagingTimeout`
on the line to find — `bounded(AXUIElementCreateSystemWide())` is that shape, and
`bounded` is what this change recommends elsewhere. The write check reads the
call's first argument rather than the text of the line, so a receiver in front of
the name does not walk past it. Two more assertions keep the list honest: every
file on it still holds the element, and `AppDelegate` still writes the floor the
uncapped hit tests fall back to, and that the call reaching it still runs at
startup — a write nobody calls is not a floor.

## Verification

display.

```sh
./build.sh --test              # TESTS OK (8942 checks)
```

Every assertion verified in both directions, each red on its own and green once
put back:

- reverting the Dock preview change, and putting `AXUIElementCreateSystemWide`
  back into that file
- `bounded(AXUIElementCreateSystemWide())` in a file not allowed to hold the
  element, and in one that is
- a write on the held element spelled three ways: `systemElement`,
  `self.systemElement`, `Self.shared.systemElement`
- `AppDelegate`'s launch write removed, commented out rather than deleted, and
  the call that reaches it removed and commented out — a string inside a
  function nobody calls is not a floor — the assertions that keep the list honest read code, not file text
- a file allowed to hold the element commented out of holding it

Capping happens where elements are produced rather than at the call sites, in
all five files: an element read out of another one is a fresh element and takes
the process-wide default, not the cap on the one it came from. The producers
take the value from the caller with no default, so the compiler covers the call
sites — it found two in `gestureTarget` that reading the code had not.

The occlusion difference was measured rather than reasoned about: two helper
windows on one rect, one at layer 0 and one floating above it, hit-tested from a
third process. The system-wide call returns the floating window; the layer-0
owner's application element returns its own window at the same point. That is
why only the Dock preview, which settles occlusion in `dockOwnsPoint` before it
asks, hit-tests through an application element.

**What I could not test.** Any of this against a genuinely frozen app. The
timeout only has an effect when a process stops answering, which does not happen
on demand; the behaviour of the caps themselves was measured against a synthetic
Accessibility target holding its main thread, not against a real hang. Nothing
here changes a healthy path: a query that answers in 0.02 ms reaches no timeout
either way, and every hit test returns the element it did before —
checked for the equivalent substitution on the running Dock, where hit-testing
through the application element returns the system-wide call's element,
`CFEqual`, same frame, same pid. I have not exercised the traffic-light paths on
a Mac with several displays.

## Anything else

This was opened as two pull requests and folded into one: the Dock preview half
was #1110, whose review is worth reading for the arithmetic corrections it
carried. It is the same change as the rest and reviewing it apart from
them was the wrong split.

Refs #971
Refs #938

### DIFF
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -489,8 +489,7 @@ final class AutoQuitService: ObservableObject {
             }
         }
         for attribute in [kAXMainWindowAttribute, kAXFocusedWindowAttribute] {
-            if let window = Self.windowAttribute(appElement, attribute as String) {
-                AXUIElementSetMessagingTimeout(window, 0.35)
+            if let window = Self.windowAttribute(appElement, attribute as String, cappedAt: 0.35) {
                 if Self.isStandardWindow(window) { Self.appendUnique(window, to: &result) }
             }
         }
@@ -524,12 +523,21 @@ final class AutoQuitService: ObservableObject {
         windows.append(window)
     }
 
-    private static func windowAttribute(_ appElement: AXUIElement, _ attribute: String) -> AXUIElement? {
+    /// The cap travels with the caller because this is the door elements come
+    /// through on two paths that wait different lengths — the observer walk at
+    /// 0.35 and the click tap at 0.15. An element produced here is a fresh one:
+    /// it takes the process-wide default, not the cap on the element it was
+    /// read from, so leaving it uncapped puts it on whatever the global says.
+    private static func windowAttribute(_ appElement: AXUIElement,
+                                        _ attribute: String,
+                                        cappedAt timeout: Float) -> AXUIElement? {
         var value: CFTypeRef?
         guard AXUIElementCopyAttributeValue(appElement, attribute as CFString, &value) == .success,
               let value,
               CFGetTypeID(value) == AXUIElementGetTypeID() else { return nil }
-        return (value as! AXUIElement)
+        let element = value as! AXUIElement
+        AXUIElementSetMessagingTimeout(element, timeout)
+        return element
     }
 
     private static func boolAttribute(_ element: AXUIElement, _ attribute: String, default defaultValue: Bool = false) -> Bool {
@@ -765,7 +773,7 @@ final class AutoQuitService: ObservableObject {
     private func closeButtonPID(at point: CGPoint, candidate: TrafficLightCandidate) -> pid_t? {
         guard candidate.pid != getpid(), observers[candidate.pid] != nil else { return nil }
         guard let element = elementAt(point: point, in: candidate.pid),
-              let window = Self.topLevelWindow(from: element)
+              let window = Self.topLevelWindow(from: element, cappedAt: 0.15)
         else { return candidate.pid }
 
         var pid: pid_t = 0
@@ -774,7 +782,7 @@ final class AutoQuitService: ObservableObject {
         guard pid == candidate.pid else { return nil }
 
         guard Self.isStandardWindow(window),
-              let closeButton = Self.windowAttribute(window, kAXCloseButtonAttribute as String),
+              let closeButton = Self.windowAttribute(window, kAXCloseButtonAttribute as String, cappedAt: 0.15),
               Self.boolAttribute(closeButton, kAXEnabledAttribute as String, default: true),
               let buttonFrame = Self.frame(of: closeButton)
         else { return candidate.pid }
@@ -817,14 +825,15 @@ final class AutoQuitService: ObservableObject {
         return CFEqual(element, AXUIElementCreateApplication(pid))
     }
 
-    private static func topLevelWindow(from element: AXUIElement) -> AXUIElement? {
+    private static func topLevelWindow(from element: AXUIElement,
+                                       cappedAt timeout: Float) -> AXUIElement? {
         guard !isApplicationElement(element) else { return nil }
         if role(of: element) == (kAXWindowRole as String) { return element }
-        if let window = windowAttribute(element, kAXWindowAttribute as String),
+        if let window = windowAttribute(element, kAXWindowAttribute as String, cappedAt: timeout),
            role(of: window) == (kAXWindowRole as String) {
             return window
         }
-        if let window = windowAttribute(element, kAXTopLevelUIElementAttribute as String),
+        if let window = windowAttribute(element, kAXTopLevelUIElementAttribute as String, cappedAt: timeout),
            role(of: window) == (kAXWindowRole as String) {
             return window
         }
@@ -837,7 +846,7 @@ final class AutoQuitService: ObservableObject {
         // already ends with no window in that case.
         var current = element
         for _ in 0..<8 {
-            guard let parent = windowAttribute(current, kAXParentAttribute as String) else { return nil }
+            guard let parent = windowAttribute(current, kAXParentAttribute as String, cappedAt: timeout) else { return nil }
             if isApplicationElement(parent) { return nil }
             if role(of: parent) == (kAXWindowRole as String) { return parent }
             current = parent
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -764,7 +764,7 @@ final class AutoQuitService: ObservableObject {
 
     private func closeButtonPID(at point: CGPoint, candidate: TrafficLightCandidate) -> pid_t? {
         guard candidate.pid != getpid(), observers[candidate.pid] != nil else { return nil }
-        guard let element = elementAt(point: point),
+        guard let element = elementAt(point: point, in: candidate.pid),
               let window = Self.topLevelWindow(from: element)
         else { return candidate.pid }
 
@@ -782,15 +782,21 @@ final class AutoQuitService: ObservableObject {
         return buttonFrame.insetBy(dx: -4, dy: -4).contains(point) ? pid : nil
     }
 
-    private func elementAt(point: CGPoint) -> AXUIElement? {
-        let system = AXUIElementCreateSystemWide()
-        // This runs while a click is being handled, and the app being asked is
-        // the one that was just told to close, which is exactly when an app
-        // stops answering. A short limit here means a slow answer is dropped
-        // rather than holding this app, and every tap it runs, still.
-        AXUIElementSetMessagingTimeout(system, 0.15)
+    /// This runs while a click is being handled, and the app being asked is the
+    /// one that was just told to close, which is exactly when an app stops
+    /// answering. A short limit here means a slow answer is dropped rather than
+    /// holding this app, and every tap it runs, still.
+    ///
+    /// Hit-tested through that app's own element rather than the system-wide
+    /// one, so the limit belongs to this call; written on the system-wide
+    /// object it would be the process-wide default (#938). The caller has the
+    /// pid from the window list already and rejects any window that does not
+    /// match it.
+    private func elementAt(point: CGPoint, in pid: pid_t) -> AXUIElement? {
+        let app = AXUIElementCreateApplication(pid)
+        AXUIElementSetMessagingTimeout(app, 0.15)
         var element: AXUIElement?
-        guard AXUIElementCopyElementAtPosition(system, Float(point.x), Float(point.y), &element) == .success
+        guard AXUIElementCopyElementAtPosition(app, Float(point.x), Float(point.y), &element) == .success
         else { return nil }
         if let element { AXUIElementSetMessagingTimeout(element, 0.15) }
         return element
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -480,12 +480,9 @@ final class AutoQuitService: ObservableObject {
 
     private func standardWindows(of appElement: AXUIElement) -> [AXUIElement] {
         var result: [AXUIElement] = []
-        var value: CFTypeRef?
-        if AXUIElementCopyAttributeValue(appElement, kAXWindowsAttribute as CFString, &value) == .success,
-           let windows = value as? [AXUIElement] {
-            for window in windows {
-                AXUIElementSetMessagingTimeout(window, 0.35)
-                if Self.isStandardWindow(window) { Self.appendUnique(window, to: &result) }
+        if let windows = Self.windowsAttribute(appElement, cappedAt: 0.35) {
+            for window in windows where Self.isStandardWindow(window) {
+                Self.appendUnique(window, to: &result)
             }
         }
         for attribute in [kAXMainWindowAttribute, kAXFocusedWindowAttribute] {
@@ -528,6 +525,18 @@ final class AutoQuitService: ObservableObject {
     /// 0.35 and the click tap at 0.15. An element produced here is a fresh one:
     /// it takes the process-wide default, not the cap on the element it was
     /// read from, so leaving it uncapped puts it on whatever the global says.
+    /// Every element in the list, for the same reason as the one above: a
+    /// caller that picks one of them still asked the others something first.
+    private static func windowsAttribute(_ appElement: AXUIElement,
+                                         cappedAt timeout: Float) -> [AXUIElement]? {
+        var value: CFTypeRef?
+        guard AXUIElementCopyAttributeValue(appElement, kAXWindowsAttribute as CFString, &value) == .success,
+              let windows = value as? [AXUIElement]
+        else { return nil }
+        for window in windows { AXUIElementSetMessagingTimeout(window, timeout) }
+        return windows
+    }
+
     private static func windowAttribute(_ appElement: AXUIElement,
                                         _ attribute: String,
                                         cappedAt timeout: Float) -> AXUIElement? {
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -772,7 +772,7 @@ final class AutoQuitService: ObservableObject {
 
     private func closeButtonPID(at point: CGPoint, candidate: TrafficLightCandidate) -> pid_t? {
         guard candidate.pid != getpid(), observers[candidate.pid] != nil else { return nil }
-        guard let element = elementAt(point: point, in: candidate.pid),
+        guard let element = elementAt(point: point),
               let window = Self.topLevelWindow(from: element, cappedAt: 0.15)
         else { return candidate.pid }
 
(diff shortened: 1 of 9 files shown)


PathGao:
The split is right, and the grep backs it: on `main` those five plus `AppDelegate` are the complete set of system-wide timeout writers, and `DockPreviewService` is the only file holding one with no cap. `elementAttribute` really is the only other door an element comes through, so `bounded` at those two points covers it.

Two holes in the guardrail.

**A type annotation defeats the writer detection.** The name is taken as the last token before `=`, so `let system: AXUIElement = AXUIElementCreateSystemWide()` yields `AXUIElement` and the write on the next line never matches. That is an ordinary Swift spelling, and the body claims the check survives spelling. Split on `:` first:

```swift
.components(separatedBy: "=").first?.components(separatedBy: ":").first?
```

**The reported line numbers are wrong.** Comment lines are dropped before `enumerated()`, so `\(file):\(index + 1)` is an index into the filtered array, not a file line — off by every comment above the match, and these files are comment-heavy. The sibling assertion right above enumerates the unfiltered source.

Not verified: anything against a real hang, or the traffic-light paths on several displays.

**Impact:** none on shipped behaviour; the guardrail can be walked past by a declaration spelling it does not recognise.
**Merge:** needs the name extraction to survive a type annotation first.

ref #938
ref #971

PathGao:
@PathGao review again

PathGao:
Both holes are closed — the type annotation is split off before the name, and `enumerated()` now runs over the unfiltered lines.

**"Values are unchanged" does not hold for the two 0.15 features.** Only the first element on each path carries a cap. Everything derived from it is a fresh element, and by the body's own measurement a fresh element inherits the last global write, which is now `AppDelegate`'s 0.35.

`FinderCutPaste.isEditingText` caps Finder's element, then asks the focused element for `AXRole` with nothing set on it — 0.15 before, 0.35 now, inside the tap whose comment says the whole session's typing waits on it.

`AutoQuitService.closeButtonPID` is the worse one: `topLevelWindow` walks parents and `windowAttribute` hands back the close button, then role, enabled, position and size are read off elements that carry no cap. Half a dozen reads, each 0.35 instead of 0.15, in the click tap.

`DockPreviewService.bounded` is the shape that fixes both — the cap belongs on the element-producing helper, not on the entry point.

Not verified: anything against a real hang.

**Impact:** when the app under the cursor stops answering, Finder ⌘X and an auto-quit close click hold the event tap longer than they did before this change.
**Merge:** needs the derived elements on the two 0.15 paths capped first.

ref #938
ref #971

PathGao:
@PathGao review again

PathGao:
Both 0.15 paths are capped end to end now — `windowAttribute(cappedAt:)` has no default, so the compiler covers the call sites, and the guardrail survives a type annotation.

**The pid check was the occlusion check.** `WindowServerTrafficLightHitTest.candidate` `continue`s past windows that are not layer 0 or under 80×80, so it hands back the layer-0 window *under* a floating panel owned by another process. The system-wide hit test returned that panel's element, `pid(of: window) != candidate.pid`, and the click passed through. Hit-testing through `candidate.pid` returns that app's own zoom/close button whatever is on top, so the pid guard can no longer fail. The body licenses the narrowing with "both then reject any window whose pid does not match" — the narrowing removes the rejections, not the hits.

`DockPreviewService` is not the same shape: `dockOwnsPoint` returns false at the first non-self window on the point that is not the Dock strip, so occlusion is settled before the hit test. The traffic-light path has no counterpart to that gate.

Smaller: the guardrail's needle is `"…SetMessagingTimeout(\(name)"`, so `self.systemElement` walks past it. Strip `self.` from the line first.

Not verified: whether `AXUIElementCopyElementAtPosition` on an application element ignores occlusion, or anything against a real hang.

**Impact:** a click on another app's floating panel over a window's traffic lights gets swallowed — window maximized, or the app treated as close-clicked.
**Merge:** needs the occlusion the pid guard was carrying restored on the two traffic-light paths first.

ref #938
ref #971

PathGao:
The 0.15 paths check out end to end: `windowAttribute(cappedAt:)` is the only element-producing helper in `AutoQuitService`, so the walk, close button and frame all read off capped elements. Pulling the pid from the same `frontmostApplication` snapshot also closes a race the old Finder guard had.

**`WindowMaximizer.topLevelWindow` still walks parents on the global.** Its `elementAttribute` doesn't cap, so up to 8 parents and their role reads take `AppDelegate`'s value — the same walk this PR just fixed in `AutoQuitService`. Same number today, so nothing moves, but the table row claims everything it reaches is capped 0.35. A `cappedAt:` with no default covers the call sites and deletes two of the three literal 0.35s.

**The guardrail is blind to the shape this PR introduces.** Detection needs `let x = AXUIElementCreateSystemWide()` and `…SetMessagingTimeout(x` on one line. `bounded(AXUIElementCreateSystemWide())` binds no name and hits no needle — and `bounded` is the shape the body recommends. `Self.` receivers slip past too; only `self.` is stripped. The DockPreview assertion already has the rule that survives both: allow-list the files that may *hold* the symbol and drop the name tracing. (Body says "the two files that may are listed"; the list has one.)

Not verified: anything against a real hang, or the traffic-light paths on several displays.

**Impact:** none on shipped behaviour; the cap rule holds in the code but not in the test meant to keep it.
**Merge:** needs the parent walk capped and the guardrail closed against the helper form first.

ref #938
ref #971

PathGao:
The parent walk is capped and the guardrail now catches `bounded(AXUIElementCreateSystemWide())` and `Self.` receivers. Grep agrees on the allow-list: those five are exactly the files naming the symbol once DockPreview drops it.

**`WindowLayoutService` still runs a loop on the global.** In `resolveTarget`, `windowAttribute(axApp, …)` and `windowsAttribute(axApp)` hand back fresh elements, and `target(from:)` asks each one role, minimized, subrole, frame and two settable checks. None carries a cap, all on the main thread. `AXUIElementSetMessagingTimeout(axApp, 0.35)` two lines up reads as if it covered them, and the table row says "that one read". Same shape you just fixed twice.

**Both honesty assertions match comment text.** `holders` and `launchFloor` are whole-file `contains` while the violation loop drops comments — comment out `AppDelegate`'s write and CI is green with four tap hit tests on the ~1.5s default. You tested removed, not commented out.

**The floor's value is unpinned.** Four features now take their timeout from that literal, and DockPreview's `messagingTimeout = 0.35` is a second copy held equal only by a comment.

Not verified: anything against a real hang.

**Impact:** none on shipped behaviour; the rule holds in the code, the test pins the floor's existence rather than its number.
**Merge:** needs the MRU loop's windows capped first.

ref #938
ref #971

PathGao:
`bounded` on both doors checks out: the hit test and `elementAttribute` are the only two places `DockPreviewService` produces an element, so a read added there later is capped without anyone remembering to.

Nothing else has moved, and the MRU loop is still uncapped. One site I had not looked at: `makeEdgeSnapDrag` is a second producer in the same file — `windowsAttribute(axApp)` hands back the whole array, `AXWindowResolver.windowID` is asked of every window in the `first(where:)`, and only the survivor is capped on the next line, inside the press tap. Both sites go away if `windowAttribute` and `windowsAttribute` take `cappedAt:` with no default, the way `AutoQuitService` already does — then the compiler holds the line instead of the next reader.

**The `messagingTimeout` comment ships a number this PR falsifies.** "six features here already write that global with three different values" — after merge, one does. Same failure the four "6 second default" comments were deleted for, in the file the change is proudest of.

Not verified: anything against a real hang, or whether `_AXUIElementGetWindow` messages the app — if it does not, the edge-snap site is empty and only the MRU loop stands.

**Impact:** none on shipped behaviour; `WindowLayoutService` reads off elements the cap rule claims are covered.
**Merge:** needs the MRU loop's windows capped first.

ref #938
ref #971

PathGao:
The MRU loop and the edge-snap array are both capped now, and `cappedAt:` with no default on both helpers means the compiler holds the line instead of the next reader.

**"Values are unchanged" still isn't true — now it's the hit tests themselves.** `AutoQuitService.elementAt` wrote 0.15 immediately before its hit test; with that write gone the read runs on `AppDelegate`'s 0.35, inside the click tap. `gestureTarget` goes 0.25 → 0.35 on a press. The table row says "same, everything it reaches capped 0.15", which reads as nothing moved. One read against a hung app, 2.3× longer than before, in the path #971 is about.

**The guardrail closes the one-line hand-off, not the two-line one.** It only inspects lines naming `AXUIElementCreateSystemWide`, so `bounded(AXUIElementCreateSystemWide())` is caught but

```swift
let system = AXUIElementCreateSystemWide()
bounded(system)
```

is not — no needle on either line. All four holders use the name in exactly one place, so one rule covers both: a collected name may appear only in its binding and as the first argument of `AXUIElementCopyElementAtPosition`. That subsumes the write loop too.

Not verified: anything against a real hang.

**Impact:** with the app under the cursor hung, an auto-quit close click and a layout drag press hold the tap 0.35 instead of 0.15 and 0.25.
**Merge:** needs the auto-quit hit test's 0.15 settled — restored, or the row stopped claiming it.

ref #938
ref #971

PathGao:
Both gaps from last round are closed: the row states the 0.15 → 0.35 move, the binding rule catches the two-line hand-off, and `code()` closes the comment-out escape.

**"A one-line change in that feature" doesn't hold.** `WindowMaximizer` carries 0.35 in five places after this — hit-test result, two window reads, parent walk, button read. `AutoQuitService`'s click path carries 0.15 in three, its observer path 0.35 in five. The global write was one copy per feature; the produce sites are more. `DockPreviewService.messagingTimeout` is the shape that holds, and it makes #938's next step a one-line change for real.

**Two producers still cap at the call site.** `FocusFollowsMouseService.topLevelWindow` hands back a fresh element uncapped and `target(at:)` caps it the next line; `AutoQuitService.standardWindows` does the same for the `kAXWindowsAttribute` array, five lines above the `cappedAt:` you just added. Correct today, held by the reader not the compiler.

**The new test comment ships a number this PR falsifies** — "six other features already write with three different values". After merge, one does.

Not verified: anything against a real hang.

**Impact:** none on shipped behaviour.
**Merge:** needs the per-feature value collapsed to one constant, or the closing claim dropped — either one.

ref #938
ref #971

PathGao:
Both producers from last round are closed, and `cappedAt:` with no default means the compiler covers the call sites rather than the next reader.

**`FinderRenameService.focusedRole` is the shape you just fixed in `FinderCutPaste`, and this PR moves it.** It caps Finder's app element at 0.15, then reads `AXRole` off the focused element with nothing set on it. A fresh element takes the global, and in Finder the last global write was usually `isEditingText`'s 0.15 — written on every keystroke. With that write gone it is `AppDelegate`'s 0.35, on the rename shortcut's tap, asking the app this PR names as the one that stops answering. One line after `let element = focused as! AXUIElement`, the same one you added three lines up in the sibling file. Grep says it is the only other 0.15 site in the tree.

**Two comments still ship numbers this change falsifies**: the test's "six other features already write with three different values", and the body's "a one-line change in that feature" — `WindowMaximizer` carries 0.35 in five places after this.

Not verified: anything against a real hang.

**Impact:** against a hung Finder, the rename shortcut's focus check holds its tap 0.35 instead of 0.15.
**Merge:** needs the rename service's focused element capped first.

ref #938
ref #971

PathGao:
The rename service is capped end to end now, and the table matches the code.

**`dockOwnsPoint` never settles our own occlusion.** It `continue`s past a window whose pid is ours instead of returning false, so a Vorssaint window over the Dock strip reaches the hit test. The system-wide call returned that window and `pid(of: candidate) == dockPID` rejected it; through the Dock's application element the icon underneath comes back instead. The body says "narrowing the search to its hierarchy removes no case that reaches here" — it removes the rejection, same as the traffic-light paths. `currentZone` checks `activePanelFrame` first, but a pinned panel outlives its session and `handleMouseMove` asks with nothing in front of it.

**`DockClickService.handleMinimizeMenu` still has the shape.** It writes 1.0 on the app element for busy JVMs, then walks `elementAttribute`/`elementArray` three levels with nothing capped, so every read takes the global. Pre-existing and off the tap; the sweep only looked for tightenings.

Not verified: anything against a real hang, or a pinned panel over the strip.

**Impact:** a pinned preview, or any window of ours over the Dock strip, now opens a preview for the icon beneath it instead of counting as off the Dock.
**Merge:** needs the own-window case the pid loop was carrying settled on the Dock preview hit test first.

ref #938
ref #971

PathGao:
Nothing left uncapped: every producer in the five files caps what it hands back, no hit test is narrowed, and the guardrail catches both hand-off spellings.

**It conflicts, and the fix has a trap.** `WindowEnumerator`'s comment was rewritten on main — new wording, and it still says "the 6 second default timeout". That one line is this PR's whole footprint in that file, so taking main's side drops the correction with nothing left to notice it.

**The floor assertion pins the text, not the call.** Deleting `Self.boundAccessibilityWaits()` at `AppDelegate:49` leaves the string inside the now-unreferenced function, so the assertion stays green while five uncapped hit tests fall to the 1.5s default inside event taps. You tested the write removed and commented out, not the call site. A second `contains` closes it.

**"Four" where the list says five.** `mayHoldSystemWide` is `AppDelegate` plus five holders. The body, the floor comment and the verification all say four — `DockPreviewService` is the one missing.

Not verified: anything against a real hang.

**Impact:** none on shipped behaviour; the floor everything falls back to is guarded by a string that outlives its call.
**Merge:** needs the rebase, with the `WindowEnumerator` correction re-applied to main's wording.

ref #938
ref #971

PathGao:
The rebase landed with the `WindowEnumerator` correction applied to main's wording, and the floor assertion now checks the call as well as the string.

**The figure survives in `AppSwitcher.focusedWindowID`** — "the 6s default AX timeout", spelled `6s`, which is why a sweep for "6 second" missed it. Grep says the tree holds six of these and this drops five. Same reasoning applies: the measured default is 1.5s, and what the comment needs is that its cap is shorter than the default. The body says four.

**"Five" where the list is six.** `mayHoldSystemWide` is `AppDelegate` plus five holders — third round running this count is off by one.

**The guardrail paragraph's last sentence describes a check the diff doesn't have.** There is no write check now; the binding rule subsumed it, and a receiver is handled by `mentions`'s whole-identifier match — which also flags `AXUIElementCopyElementAtPosition(Self.shared.systemElement,` as a violation. Safe, but not what the sentence says.

Not verified: anything against a real hang.

**Impact:** none on shipped behaviour.
**Merge:** needs the sixth `6 second` comment dropped in `AppSwitcher.focusedWindowID` first.

ref #938
ref #971

PathGao:
Superseded by #1136, which is the same defect at 15 lines instead of 326.

What that one keeps: the five global writes deleted, and the wrong figure for the unset default removed. What it drops: element-scoped caps replacing the values those writes were carrying, and a guardrail against a sixth writer. Both were me answering #938's first and fourth questions on the way past — the caps preserve a difference between 0.15 and 0.35 that is a few hundred milliseconds on a path that only runs when an app is already frozen, and the rule is worth having only once the shape it enforces has been decided.

The branch stays if any of it is wanted later.

Issue #1123, opened 2026-08-29
Developer build never sees its Accessibility grant: setup-signing.sh dies silently on stock LibreSSL, leaving every rebuild ad-hoc


iltonandrew:
### What happens

On a self-built Developer build (`./build.sh --dev --install`), Accessibility-gated features stop working after any rebuild, and the state is a dead end:

- System Settings → Privacy & Security → Accessibility shows **Vorssaint (Developer) as granted**.
- The app's Permissions page shows Accessibility as **not granted** (`AXIsProcessTrusted()` is false).
- Clicking the grant button does nothing visible: `AXIsProcessTrustedWithOptions(prompt:)` never prompts because a TCC row for `com.vorssaint.utils.dev` already exists, and the guide overlay points at a toggle that is already on.

Since most features lean on Accessibility, the Developer build is effectively unusable.

### Why — two defects chained

**1. `Tools/setup-signing.sh` dies silently on the stock openssl.** The script runs `openssl pkcs12 -export -legacy …`. `-legacy` is an OpenSSL 3 flag; macOS ships LibreSSL at `/usr/bin/openssl`, which rejects it (`unknown option '-legacy'`). Both openssl calls have `2>/dev/null`, so under `set -e` the script exits 1 printing nothing at all. On any Mac without a Homebrew OpenSSL 3 in PATH, the documented cure for unstable local signing has never worked:

```
$ ./Tools/setup-signing.sh; echo $?
1
$ /usr/bin/openssl version
LibreSSL 3.3.6
```

**2. `build.sh` falls through to ad-hoc, whose grants cannot survive a rebuild.** With no identity, the bundle is ad-hoc signed and its designated requirement degenerates to the binary's hash:

```
$ codesign -d -r- "/Applications/Vorssaint (Developer).app"
# designated => cdhash H"b59de61e6f652f000fb44ce1ad8bf5eed446ac96"
```

macOS ties the Accessibility grant to that requirement. The next build has a different cdhash, so tccd stops trusting the app — while System Settings keeps rendering the stale row as granted and suppresses any new prompt. The only mid-build hint is one quiet line (`signing ad-hoc (no identity installed — run Tools/setup-signing.sh)`), advising a script that then fails silently (defect 1).

### The docs already disagree about this

[...]

PR #1132, opened 2026-08-29
fix(permissions): let the guide card start over on a stale entry and relaunch for Screen Recording

andreisuslov:
## Summary

The permission guide card can wait forever on a switch that is already on. The entry macOS keeps for Accessibility or Screen Recording is bound to the app's code signature, so a copy signed differently (a local build, or an update signed another way) finds its switch on in System Settings and the permission gone. The card says "Turn Vorssaint on in that list", the person finds it on, and nothing happens.

Two additions to the card:

- After 12 seconds of waiting it says why that can happen and offers **Start over**, which removes this app's own entry (`tccutil reset <service> <bundle id>`, the command Apple documents for it, no privilege needed for one's own entries) and asks again, so System Settings shows a fresh row that can be turned on.
- For Screen Recording, which applies only to a fresh process, the granted state keeps the card up with a **Relaunch to apply** button (the existing `relaunchApp`) instead of fading out and leaving the feature not working.

Built on `tccutil` and the app's own permission refresh; no new dependency or permission. Strings in all thirteen languages.

What a user notices: a card that is stuck on an already-on switch explains itself and has a button that fixes it.

## Verification

MacBook Pro 14" (MacBookPro18,3, M1 Pro), macOS 15.7.7 (24G720), one display, English (US). Own build: `./build.sh --dev --install`, `./build.sh --test`. Reproduced the stuck state by rebuilding an ad-hoc-signed developer build (the switch stays on, `CGPreflightScreenCaptureAccess` stays false) and cleared it with the same `tccutil reset` the button runs. Not tested by hand yet: the button itself on screen, and a localized card. I will confirm and say so here.

### DIFF
--- a/Sources/Vorssaint/Core/PermissionGuideStrings.swift
+++ b/Sources/Vorssaint/Core/PermissionGuideStrings.swift
@@ -15,6 +15,12 @@ struct PermissionGuideStrings {
     let waiting: String
     let granted: String
     let closeHelp: String
+    /// Shown once the wait has gone on a while: the usual cause is an entry
+    /// left by an earlier copy of the app, which macOS shows as on but no
+    /// longer honours.
+    let staleHint: String
+    let startOver: String
+    let relaunch: String
 }
 
 extension FeatureStrings {
@@ -45,7 +51,10 @@ extension PermissionGuideStrings {
         stepReturn: "여기로 돌아오세요. 이 카드가 자동으로 확인합니다.",
         waiting: "권한을 기다리는 중…",
         granted: "권한이 허용되었습니다!",
-        closeHelp: "닫기"
+        closeHelp: "닫기",
+        staleHint: "목록에서 이미 켜져 있나요? 그 항목은 이전 앱 사본의 것입니다. 다시 시작하여 교체하세요.",
+        startOver: "다시 시작",
+        relaunch: "적용하려면 다시 실행"
     )
 }
 
@@ -57,7 +66,10 @@ extension PermissionGuideStrings {
         stepReturn: "Come back. This card notices by itself.",
         waiting: "Waiting for the permission…",
         granted: "Permission granted!",
-        closeHelp: "Close"
+        closeHelp: "Close",
+        staleHint: "Already on in that list? That entry belongs to an earlier copy of the app. Start over to replace it.",
+        startOver: "Start over",
+        relaunch: "Relaunch to apply"
     )
 
     static let ptBR = PermissionGuideStrings(
@@ -67,7 +79,10 @@ extension PermissionGuideStrings {
         stepReturn: "Volte para cá. Este cartão percebe sozinho.",
         waiting: "Esperando a permissão…",
         granted: "Permissão concedida!",
-        closeHelp: "Fechar"
+        closeHelp: "Fechar",
+        staleHint: "Já está ativado nessa lista? Essa entrada pertence a uma cópia anterior do app. Recomece para substituí-la.",
+        startOver: "Recomeçar",
+        relaunch: "Reabrir para aplicar"
     )
 
     static let tr = PermissionGuideStrings(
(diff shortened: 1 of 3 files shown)


PathGao:
Right diagnosis — the stuck card really is a signature-bound entry that macOS still shows as on.

**`startOver` hand-rolls the subprocess.** `SelfUninstall.resetTCC()` already runs the identical command as `Shell.run("/usr/bin/tccutil", ...)`, which routes through `BoundedProcessRunner` — the runner whose comment names #971 as why a bare `waitUntilExit()` was removed. Yours is bare, and it's on the main thread (button action), so a stuck `tccutil` hangs the UI. `Shell.run` on a global queue, hop back to main for the refresh.

**Start over tears down its own card mid-click.** `startOver` → `requestAccessibility()` → `show()`, which does `panel = nil` while the button's action is still on the stack, then builds a fresh model — the hint and button vanish and the 12s wait restarts. `DispatchQueue.main.async` around the call.

**A granted Screen Recording card polls forever.** `visibleSurfaceCount > 0` pins `PermissionPollingSupport.interval` at 2.5s whatever the grant state, and this card no longer leaves. Drop the surface demand in the granted sink, keep the panel.

Not verified: whether `CGPreflightScreenCaptureAccess` flips before relaunch. If it doesn't, that Relaunch button never appears.

**Impact:** UI hang on a stuck `tccutil`; a 2.5s timer that never stops after a Screen Recording grant.
**Merge:** needs the `Shell.run` swap, off the main thread, first.

ref #1123

andreisuslov:
@PathGao every point raised here is addressed and pushed, main is merged in, tests pass and GitHub shows the branch as mergeable. From my side this is complete and ready to merge.

PathGao:
The diff I'm reading doesn't have those changes in it. `Permissions.startOver` still builds its own `Process` and calls `waitUntilExit()` inline; `onStartOver` still calls it straight from the button action, so that wait is on the main thread; and the granted sink still leaves `pollingDemandID` demanding a surface, which pins the interval at 2.5s after a Screen Recording grant.

The first one is one line — `SelfUninstall.resetTCC()` already makes the same call:

```swift
_ = Shell.run("/usr/bin/tccutil", ["reset", service, bundleID])
```

on a global queue, `refreshActivePermissions()` hopped back to main.

If you pushed after the snapshot I'm looking at, say so and I'll re-read. Otherwise check what actually landed on the branch — a merge of main wouldn't carry these.

**Impact:** unchanged from my last comment — UI hang on a stuck `tccutil`, and a 2.5s timer that never stops after a Screen Recording grant.
**Merge:** needs the `Shell.run` swap, off the main thread, first.

andreisuslov:
You were reading the branch as it was; the earlier reply described the diagnosis, not a push. All three are now in `de82182`:

- `startOver` runs `Shell.run("/usr/bin/tccutil", ["reset", service, bundleID])` on a global queue, the same call as `SelfUninstall.resetTCC`, with `refreshActivePermissions()` and the re-request hopped back to main. No bare `Process` or `waitUntilExit` left.
- That hop also means the button's click finishes before `show()` rebuilds the card, so the hint and button no longer vanish under the pointer and the 12 s wait does not restart mid-click.
- The granted sink releases the polling surface demand (`setActivePermissionSurface(pollingDemandID, visible: false)`) for both kinds; the Screen Recording card keeps its panel and Relaunch button without the 2.5 s timer behind it.

On whether `CGPreflightScreenCaptureAccess` flips before a relaunch: on this Mac (15.7.7) it did, after the switch was turned on for the running process, which is what let the Relaunch button appear. I have not checked other versions.

andreisuslov:
@vorssaint ready to merge whenever you have a moment. Main is merged in (branch is current), `./build.sh` finishes without warnings, `--selftest` passes and `./build.sh --test` is green; @PathGao's review points here are all addressed and answered above. This one is independent.

PathGao:
All three landed in `de82182` and I re-read them against `main`.

`startOver` goes through `Shell.run` now, which is bounded at 5s in `BoundedProcessRunner` and off the main thread, with the refresh and the re-request hopped back — that also takes care of the mid-click teardown for free. And releasing `pollingDemandID` in the granted sink drops `PermissionPollingSupport.interval` from 2.5s to 60s (or nil) for both kinds, so the Screen Recording card can stay put without a timer behind it.

Nothing else to flag.

Not verified: the two new buttons rendering in the non-activating borderless panel, and `CGPreflightScreenCaptureAccess` on anything other than 15.7.7. If it doesn't flip pre-relaunch on some version, the Relaunch button just never appears there and the card behaves as it does today — no worse.

**Impact:** none outstanding.
**Merge:** ship it.

andreisuslov:
Thanks. On the two unverified points: both buttons render and act in the borderless non-activating panel on this Mac (the card is the one I used to re-grant after the signing change), and 15.7.7 is the only version I have. Ready for @vorssaint.

Issue #215, opened 2026-07-12
[Bug]: App Switcher doesn't display certain applications


AmberFirefox:
### Description

App Switcher does not display certain applications, Such as Steam, and FL Studio.

<img width="1174" height="902" alt="Image" src="https://github.com/user-attachments/assets/f366a4b5-cd70-4837-9042-2f0d475aaedd" />
Both have a custom window frame so that might be part of it?

### Steps to Reproduce

1. Use the App Switcher
2. Using Steam, When alt-tabbing using the App Switcher Steam doesn't appear.

### Vorssaint Version

3.1.12

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Github Release

Namak1719:
Ditto, MS Word for me.

PR #1133, opened 2026-08-29
fix(switcher): list windows from apps that draw their own chrome

PathGao:
## Summary

Apps that draw their own title bar never appear in the App Switcher at all.

macOS gives a borderless window the `AXUnknown` subrole. The switcher read
that as "utility surface" and accepted it only when it filled a screen or the
process was a compatibility layer (`isSwitchableNonstandardWindow`). For an
app whose windows are all borderless, nothing passed the filter, and the
Accessibility cross-check that removes stale surfaces then vetoed every window
server surface the app owned — so the app disappeared from Cmd-Tab entirely
rather than losing one window. That veto has behaved this way since at least
v3.1.12, which is why the report predates the ghost-window fix in #826.

`AXUnknown` is the absence of a description, not a description of a utility
surface, so no bundle identifier can decide this correctly. The window server
already separates the two: an ordinary window sits at the normal window level
and a HUD, palette or overlay floats above it. An undescribed `AXWindow` at
the normal level is now a switch target whichever app shipped it. Subroles the
app did describe are untouched — a dialog, sheet or floating panel is filtered
exactly as before unless it fills the screen.

Measured on the shapes an AppKit app can produce (role, subrole, window
server layer):

| window | subrole | layer | before | after |
| --- | --- | --- | --- | --- |
| `.borderless` | `AXUnknown` | 0 | filtered | switchable |
| `.borderless` at `.floating` level | `AXUnknown` | 3 | filtered | filtered |
| `.borderless, .resizable` | `AXDialog` | 0 | filtered | filtered |
| `.titled` | `AXStandardWindow` | 0 | switchable | switchable |

Alternatives considered:

- **Add the bundle identifiers to `isSupportedMediaFloatingWindow`.** That
  list is the same defect handled per app; it has already been extended once
  and would need a line for every game client, DAW and custom-chrome app that
  ever gets reported.
- **Drop the screen-size requirement for every nonstandard subrole.** Panels
  and dialogs would become switcher entries; the measurements above show the
  window level draws the line where the subrole cannot.
- **Stop the veto instead — treat any window the app still lists as proof its
  surface is real.** This fixes the disappearance without deciding what is a
  switch target, but it puts every dialog and panel surface back into the
  list, which is a wider behaviour change than the report asks for.

The `AXUnknown` half of the media-app allowlist is now redundant for anything
at the normal window level, and the compatibility-layer exception only still
matters for surfaces that resolve to no window server id. Both are left in
place: removing them needs a check on Adobe and Crossover hardware, which I do
not have.

## Verification

Mac17,3 (Apple M5), macOS 26.6.2 build 25G83, two displays, one Space, my own
`./build.sh` release build.

```
./build.sh                  # no warnings
./build.sh --test           # TESTS OK (8974 checks)
```

Since I have neither FL Studio nor Steam, I verified the filter against live
Accessibility data instead: a probe built the same window shapes as the table
above, then ran both the old and the new rule over every window of every
regular app running on this Mac. One verdict changed — the normal-level
borderless window flipped from filtered to switchable. Eleven other live
windows kept their verdict and nothing became newly filtered.

**Not tested:** FL Studio and Steam themselves, so I cannot confirm they use
this window shape rather than another nonstandard one; other Spaces and
full-screen Spaces; Adobe and Crossover apps, whose existing exceptions this
change leaves in place but no longer needs at the normal window level.

## Anything else

#512 carries two reports and this covers only the second one, FL Studio never
appearing; the closed-window half was fixed by #826.

Refs #512
Refs #215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Vorssaint expands screen text recognition, clipboard, Super key, Dock Preview, A
 - Cleaner leftover scans now cover more preference panes and plugin folders while refusing nested app data, version folders, links and other ambiguous paths.
 
 ### Fixed
+- App Switcher now lists windows from apps that draw their own title bar or use borderless windows. Thanks to @PathGao.
 - Switching capture modes in the screen capture chooser now transitions smoothly without flickering or restarting the overlay session.
 - Clicking long text previews in the menu panel's clipboard section now keeps the preview confined to its row instead of overflowing across neighboring items. Thanks to @andreisuslov.
 - Navigating clipboard history with arrow keys now prevents hovering pointers from fighting keyboard selection, keeps the list responsive during scrolling and renders large text previews without layout pauses. Thanks to @andreisuslov.
--- a/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
@@ -515,17 +515,32 @@ enum SwitcherSupport {
             || lower.hasPrefix("com.adobe.characteranimator")
     }
 
-    /// Some full-screen playback surfaces keep a nonstandard Accessibility
-    /// subrole. A screen-sized AX window is still a real switch target, while
-    /// smaller utility surfaces remain filtered. Compatibility-hosted windows
-    /// retain their existing role-based exception at every size.
+    /// Whether a window whose Accessibility subrole is not a standard one is
+    /// still a switch target.
+    ///
+    /// `AXUnknown` is the absence of a description, not a description of a
+    /// utility surface: apps that draw their own title bar ship borderless
+    /// windows, and macOS reports those as an undescribed `AXWindow`. The
+    /// window server tells those apart from overlays for us — an ordinary
+    /// window sits at the normal window level, while a HUD or panel floats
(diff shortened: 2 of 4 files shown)


PathGao:
Window level is the right signal: it separates a borderless window from an overlay without naming apps, and the four-shape table makes it checkable.

One gap. Non-fullscreen `AXUnknown` windows now enter the AX snapshot, and `appendAccessibilityOnlyWindows` appends every snapshot entry the main loop did not mark `seen`. The main loop drops `alpha == 0` surfaces with a bare `continue` — no `seen.insert`. So a transparent borderless normal-level window ≥80×60 that the app lists over AX goes from filtered to an untitled switcher entry. Splash screens and popups an app left at the default level come through the same door. Read off the code, not seen on hardware.

Smaller: `hasNormalWindowLevel` already carries `subrole == "AXUnknown" && role == AXWindow`, and `isSwitchableNonstandardWindow` re-checks both. The new floating-panel test passes `hasNormalWindowLevel: true` with `AXFloatingWindow`, an input the caller cannot produce — it asserts the duplication, not behaviour. One side should own the check.

**Impact:** apps with invisible or splash borderless windows at the normal level gain phantom switcher entries.
**Merge:** needs the alpha-0 / AX-only path checked first.

ref #512

PR #1136, opened 2026-08-29
fix: give the Accessibility messaging timeout one writer

PathGao:
## Summary

`AXUIElementSetMessagingTimeout` on the system-wide element sets the default for
every Accessibility question this process asks, whoever asks it — the header says
so, and it is measurable: a fresh element takes whichever global was written
last, and a value written on a specific element survives a later write to the
global.

Six places wrote it. `AppDelegate` means it: once at launch, documented as the
floor for the paths that cap nothing of their own. The other five each read
locally as that feature's own limit, and none of them was one.

| | wrote | trigger |
|---|---|---|
| `AutoQuitService` | 0.15 | clicking a window's close button |
| `FinderCutPaste` | 0.15 | ⌘X/⌘C/⌘V in Finder |
| `WindowLayoutService` | 0.25 | a modifier-drag on a window |
| `FocusFollowsMouseService` | 0.25 | enabling the feature |
| `WindowMaximizer` | 0.35 | clicking a green button |

Last writer wins, so how long Vorssaint waited on a frozen app depended on which
unrelated feature you had used most recently. Hover a Dock icon after using
Finder cut/paste and it waited 0.15s; after clicking a green maximize button,
0.35s. Same action, 2.3× apart, for a reason nothing on screen explains.

Those five stop writing it, and the reads they make wait the launch default —
which is what an uncapped read was always meant to get. The ~30 timeouts written
on specific application and window elements are untouched: those are genuinely
local, and each feature keeps its values.

**What this costs.** Against an app that has stopped answering, the reads on
those five paths wait 0.35s where they used to wait 0.15s or 0.25s — a few
hundred milliseconds per click or keystroke, and only when something is already
wrong. Two of the deleted lines carried reasoning worth keeping even though the
line is gone: auto-quit asks the app it has just told to close, which is exactly
the app that stops answering; the Finder guard runs on a tap the whole session's
typing waits behind. If either deserves to wait less than the floor, the way to
give it that is a timeout on the elements it holds, which is #938's second
question and not this change.

**Also, a figure four comments cite for the unset default.** Measured on 26.6.2
against a bundled, signed application holding its main thread: an element with no
timeout set gives up at 1.5s — three runs within 10ms of each other, with a 0.4s
control tracking exactly. Not the 6s those comments claim. Removed rather than
corrected, since Apple documents no default and any figure written here rots the
way this one did.

## Verification

display.

```sh
./build.sh --test              # TESTS OK (9076 checks)
```

No test is added: the change is the absence of five calls, and the assertion that
would pin it — no file but `AppDelegate` may write the process-wide default — is
#938's fourth question, which wants an answer before a rule enforces one.

**What I could not test.** Any of this against a genuinely frozen app. The
timeout has an effect only when a process stops answering, which does not happen
on demand; the behaviour of the global was measured against a synthetic
Accessibility target holding its main thread. Nothing changes on a healthy path:
a query that answers in 0.02s reaches no timeout either way.

## Anything else

Refs #938

### DIFF
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -172,7 +172,7 @@ final class AutoQuitService: ObservableObject {
 
         let appElement = AXUIElementCreateApplication(pid)
         // A launching app that is busy (say, blocked on a Keychain prompt)
-        // would hold every synchronous AX call below for the 6 second default
+        // would hold every synchronous AX call below for the default
         // timeout apiece, freezing the main thread and every event tap with
         // it: typing dies system wide (issue #189). Give up fast and retry
         // with growing spacing until the app services its run loop again.
@@ -335,7 +335,7 @@ final class AutoQuitService: ObservableObject {
 
         let appElement = AXUIElementCreateApplication(pid)
         // Bounded AX: an unresponsive app must not hold the main thread
-        // (and with it every event tap) for the 6 second default timeout.
+        // (and with it every event tap) for the default timeout.
         AXUIElementSetMessagingTimeout(appElement, 0.35)
         let hasMinimizedWindow = hasKnownMinimizedWindow(pid: pid, appElement: appElement)
         let hasVisibleWindow = hasUserFacingWindow(pid: pid, appElement: appElement)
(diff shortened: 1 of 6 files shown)


PathGao:
Grepped `main`: seven files create a system-wide element, six wrote the global, and this removes exactly the five that shouldn't. `DockPreviewService.swift:1118` is the seventh — creates one, never writes, inherits whatever came last. That's the victim in your Dock-hover example.

One correction to the cost paragraph. The global governed every element with no pin of its own, not just the system element, so the multiplier lands per hop, not once:

- `FinderCutPaste.isEditingText` — `AXFocusedUIElement` on `system`, then `AXRole` on `focused`, which is unpinned. 0.30 → 0.70, inside the ⌘X/⌘C/⌘V tap callback.
- `AutoQuitService.closeButtonPID` — `elementAt` pins the hit element at 0.15, but `topLevelWindow`'s parent walk, `kAXCloseButtonAttribute`, `kAXEnabledAttribute` and `frame(of:)` are all unpinned. Each was 0.15, each is now 0.35.

Not a new worst case — those hops already got 0.35 whenever the maximizer wrote last. It's the body's "a few hundred milliseconds" that undercounts, not the change.

Not verified: anything against a frozen app, and the 1.5s figure — I took it as measured. Both build checks were still blank when I read this.

**Impact:** none on healthy paths; on a hung app the two tap paths hold longer than the body states, bounded by 0.35 per hop.
**Merge:** ship it.

ref #938

PR #1137, opened 2026-08-29
fix(shortcuts): resolve shortcut key labels against the active keyboard layout

egesecgin:
## Summary

Shortcut key labels (in Settings, the Menu Panel, Quick Launcher, App Switcher hints, and the Shortcut Recorder) were statically mapped from ANSI key codes (`kVK_ANSI_A`...`kVK_ANSI_Z`, numbers, symbols) directly to US QWERTY characters.

On non-US layouts (such as Turkish Q / German QWERTZ, French AZERTY, Dvorak, Colemak, or custom layouts created via Ukelele):
- On Turkish Q: pressing the physical Turkish keys `Ğ`, `Ü`, `Ş`, `İ`, `Ö`, `Ç`, `<` recorded their keycodes but displayed US punctuation `[`, `]`, `;`, `'`, `,`, `.`, ` ` ` instead.
- On German QWERTZ: pressing physical **Z** (key code 16, `kVK_ANSI_Y`) recorded key code 16 but displayed `⌘Y` instead of `⌘Z`.
- On French AZERTY: pressing **A** (key code 12, `kVK_ANSI_Q`) displayed `⌘Q` instead of `⌘A`.
- `matchesByCharacter` failed because it compared the typed character against the hardcoded ANSI label rather than the active layout character.

### Changes:
- Routed alphanumeric and symbol keys through `layoutKeyLabel(for:)` which derives the character using `UCKeyTranslate` on the active keyboard layout's `UCKeyboardLayout` data.
- Preserved dedicated symbols for non-character structural keys (`Tab`, `Space`, `Return`, `Esc`, arrows, delete, navigation, and function keys).
- Added `fallbackKeyLabel` to cleanly fall back to ANSI US characters when layout data is absent or when queried off the main thread before caching.
- Registered an observer for `kTISNotifySelectedKeyboardInputSourceChanged` so labels update dynamically when the user switches keyboard layouts.
- Added test coverage in `MetricsTests.swift` for alphanumeric keys, special navigation keys, and dynamic layout resolution.

## Verification

- Ran `./build.sh --test` (9,089 checks passed).
- Built full bundle via `./build.sh` and verified with `./build/Vorssaint --selftest` (passed cleanly).
- Verified on macOS (Apple Silicon, macOS 26) with Turkish-QWERTY and US layouts.

## Anything else

Refs #1047

image

### DIFF
--- a/Sources/Vorssaint/Core/GlobalShortcut.swift
+++ b/Sources/Vorssaint/Core/GlobalShortcut.swift
@@ -389,6 +389,75 @@ struct GlobalShortcut: Equatable, Hashable {
 
     private var keyLabel: String? {
         switch Int(keyCode) {
+        case kVK_Tab: return "Tab"
+        case kVK_Space: return "Space"
+        case kVK_Return: return "Return"
+        case kVK_Escape: return "Esc"
+        case kVK_LeftArrow: return "←"
+        case kVK_RightArrow: return "→"
+        case kVK_UpArrow: return "↑"
+        case kVK_DownArrow: return "↓"
+        // Editing and navigation keys. They print as the caps the keyboard
+        // itself carries, the same way the arrows above do: spelling them out
+        // ("Page Down") would overflow the shortcut field on a full keyboard
+        // combination, and these caps are what every menu on this system shows.
+        case kVK_Delete: return "⌫"
+        case kVK_ForwardDelete: return "⌦"
+        case kVK_Home: return "↖"
+        case kVK_End: return "↘"
+        case kVK_PageUp: return "⇞"
+        case kVK_PageDown: return "⇟"
+        case kVK_ANSI_KeypadEnter: return "⌤"
+        case kVK_F1: return "F1"
+        case kVK_F2: return "F2"
+        case kVK_F3: return "F3"
+        case kVK_F4: return "F4"
+        case kVK_F5: return "F5"
+        case kVK_F6: return "F6"
+        case kVK_F7: return "F7"
+        case kVK_F8: return "F8"
+        case kVK_F9: return "F9"
+        case kVK_F10: return "F10"
+        case kVK_F11: return "F11"
+        case kVK_F12: return "F12"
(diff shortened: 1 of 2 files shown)


PathGao:
The layout-first / ANSI-fallback split is the right shape, and keeping Tab/Return/arrows out of it is correct.

**Duplicate observer.** `AppSwitcher.startObservingKeyboardLayout` already registers the same `kTISNotifySelectedKeyboardInputSourceChanged` observer and calls `refreshLayoutLabels()`. Now both fire on every layout switch, and `stopObservingKeyboardLayout()` stops nothing — the new static observer is permanent. Drop the observer half in `AppSwitcher`, keep the priming call.

**The main thread never reads the cache.** `layoutKeyLabel` re-derives on every call: `TISCopyCurrentKeyboardLayoutInputSource` + `UCKeyTranslate` per label. That used to happen only for ISO/JIS extras; now it's every letter, digit and symbol, on every `displayString`. I haven't measured the cost — but with an observer keeping the cache fresh, the main-thread path can just read it.

**`derivedLayoutKeyLabel(for:layoutData:)` lost `private` and nothing calls it.** The body says the tests cover dynamic layout resolution; they don't — every new assertion passes on the ANSI fallback alone, and no test feeds a non-US layout blob. Feed one through that entry point, or put `private` back.

**Impact:** labels unaffected; double cache rebuild per layout switch, and the switcher's stop path leaves behind an observer it believes it removed.
**Merge:** needs the AppSwitcher observer removed first.

ref #1047

PathGao:
Closing as already on main. The same shape landed in three commits after this was opened: 7d6ecfea (2026-08-31, label resolution for non-QWERTY layouts), then #1226 and #1220 on 2026-09-01. `GlobalShortcut.swift` on main now resolves labels layout-first with the ANSI fallback and starts the keyboard-layout observer from `AppDelegate`, which is what this branch did.

#1047 is labelled fixed pending release, so the release comment will land there. Thanks for the branch; it matched the fix that shipped.

Issue #1138, opened 2026-08-29
Desktop widgets for the system monitor


BacterioBranches:
## Problem or Motivation

Everything the system monitor knows is one click away, and only one click away.
To see CPU, temperature or free space I have to open the menu panel, look, and
close it again. The menu bar carries a metric or two before it runs out of
room, and it cannot show a bar or a trend at all.

macOS already has the surface for exactly this — the desktop widgets people
keep next to Calendar and Batteries — and Vorssaint samples all of it already.
The gap is not the data, it is that there is nowhere to leave it visible.

## Proposed Solution

A WidgetKit extension with a small number of widgets built on the readings
`SystemMonitor` already produces:

- **Performance** — CPU, GPU, memory (tinted by pressure) and boot volume.
- **Temperature** — CPU and GPU, with fan RPM when Fan Control makes it
  available, and the row simply absent when it does not.
- **Quick actions** — up to four configurable buttons for the reversible quick
  toggles (appearance, hidden files, desktop icons, display off, lock screen,
  screen saver, mic mute).

I have a working prototype running on my Mac and would rather settle the
direction before polishing it, per CONTRIBUTING #6. Screenshots below.

### How it crosses the sandbox

A widget extension is sandboxed and runs in its own process, so it can read
neither the SMC nor any volume. The app samples and writes a snapshot; the
extension only renders it. The transport is a JSON file under a fixed absolute
path, because a sandbox temporary exception cannot express `~`, namespaced per
bundle id so an official install and a local Developer build do not overwrite
each other. Only usage percentages, RPM and free space travel through it.

Quick actions go the other way: the button records the request, the app runs
it. Only reversible actions are reachable, and the request is checked for
ownership and age and consumed exactly once — the shared directory is writable
by every account on the Mac, so emptying the trash and ejecting disks are
deliberately not reachable from a widget.

### The four things that decide it (CONTRIBUTING #6)

**What it is built on, and whether that will still be here.** WidgetKit and
App Intents, both first-party and both the only way macOS lets anything draw on
the desktop. Nothing private, no undocumented behaviour.

**What it exposes the project to.** One world-readable directory outside
`$HOME`. This is the honest cost of the design and the part most worth
challenging: it is forced by the sandbox refusing `~` in a path exception, and
it is mitigated by carrying nothing private, by the allow-list, and by the
ownership and age checks. The alternatives are an App Group, which needs a Team
ID and so breaks every contributor's local build, or an XPC service, which
[...]

PR #1139, opened 2026-08-29
feat(widgets): add a desktop performance widget

BacterioBranches:
Refs #1138

**Scope cut down after review.** This was three widgets; it is now one. Temperature and Quick actions are kept on a branch and will follow separately, once the direction here is settled. 1360 lines became 624, and with the buttons went the App Intents, the metadata build step, the command channel and the write access.

Still a draft while Q1 and Q2 in the issue are open — whether this belongs here at all, and whether the shared directory is acceptable.

## What a user sees

One widget to drop on the desktop: **CPU, GPU, memory** (tinted by pressure) **and the boot volume**, small or medium.

With no widget placed nothing changes: the monitor stays exactly as idle as it is today. Placing one is the switch, removing the last one turns it back off — no setting, no hub entry.

## How it works, and why

A widget extension is sandboxed in its own process and can read neither the SMC nor any volume, so the app samples and writes a snapshot and the extension only renders it — **read-only, with nothing to say back**. The path is absolute and outside `$HOME` because a sandbox temporary exception cannot express `~`, and it is namespaced per bundle id so an official install and a local Developer build do not overwrite each other's data. Only usage percentages and free space travel through it, and a stale snapshot is shown as stale instead of as current numbers.

The widget asks only for what it draws, so a widget on the desktop never wakes the SMC.

`build.sh` compiles and signs the extension next to the fan helper, with its own entitlements and before the app.

## On the three points raised in review

**Uninstall — already covered, please re-check.** Both paths carry it in this branch: `SelfUninstall.removePreferences()` removes `/Users/Shared/Vorssaint/<bundle id>` and then `rmdir`s the parent, which refuses a non-empty directory so a second install's data survives; `Tools/uninstall.sh` does the same. Worth naming a third route that neither of us mentioned: `--uninstall` on the binary (`Uninstaller.runAndExit`) only detaches the daemon and login item and deletes no files — the script does that part. What I have verified is that the code is on both paths, not that either runs clean end to end: when I removed my own Developer build I deleted the directory by hand.

**Q3 — taking the suggestion, and it is better than what I had.** Putting already-localized labels in the snapshot removes the problem instead of solving it: no second catalog, and a language change lands on the next sample. It applies to the Quick actions widget, so it will land with that one. This PR needs none of it — the widget has no user-facing text beyond its gallery entry.

**Q4 — done, Performance only.** As you say, that makes it cheap: no intents, so no localization work at all.

## Callers checked

- `SystemMonitor.currentPlan` / `shouldRun` — the new activation flag follows the shape of `menuBarActive` and `alertsActive` rather than adding a path of its own, and the existing `available(_:)` gates still veto every metric.
- `SelfUninstall.removePreferences` and `Tools/uninstall.sh` — both updated, so neither uninstall route leaves the directory behind.

## Verified on

MacBook Pro (Apple Silicon), macOS 26.6 build 25.5.0, Spanish, own Developer build signed with the local identity. `./build.sh` finishes without warnings, `--selftest` passes, and the unit tests pass except for the pre-existing dispatch pool starvation check, which fails the same way on a clean `main`.

Checked by hand: the widget appears in the gallery and renders live values; the extension loads from a locally signed, Team-ID-less bundle; a stale snapshot is presented as stale.

**Not verified:** a second Mac, an Intel Mac, a Developer ID signature, multiple displays or Spaces.

## Note

The extension leaves a sandbox container at `~/Library/Containers/<bundle id>.widgets` — new here, since the app itself is not sandboxed. It holds no user data and `containermanagerd` owns it, so no uninstall route can remove it without Full Disk Access; macOS reclaims it. Flagged rather than papered over with a removal that would silently fail.

### DIFF
--- a/Resources/VorssaintWidgets-Info.plist
+++ b/Resources/VorssaintWidgets-Info.plist
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>VorssaintWidgets</string>
+	<key>CFBundleIdentifier</key>
+	<string>com.vorssaint.utils.widgets</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>Vorssaint</string>
+	<key>CFBundleDisplayName</key>
+	<string>Vorssaint</string>
+	<key>CFBundlePackageType</key>
+	<string>XPC!</string>
+	<!-- Xcode stamps this on every extension it builds, and the widget host
+	     uses it to decide the bundle is loadable on this platform. Building
+	     with swiftc directly means writing it by hand. -->
+	<key>CFBundleSupportedPlatforms</key>
+	<array>
+		<string>MacOSX</string>
+	</array>
+	<key>DTPlatformName</key>
+	<string>macosx</string>
+	<key>DTSDKName</key>
+	<string>macosx14.0</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+	<key>LSMinimumSystemVersion</key>
+	<string>14.0</string>
+	<key>NSExtension</key>
+	<dict>
+		<key>NSExtensionPointIdentifier</key>
+		<string>com.apple.widgetkit-extension</string>
+	</dict>
+</dict>
+</plist>
--- a/Resources/VorssaintWidgets.entitlements
+++ b/Resources/VorssaintWidgets.entitlements
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<!-- Widget extensions must be sandboxed; that is what makes the snapshot
+	     file necessary in the first place, since this process can read neither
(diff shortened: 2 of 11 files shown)


BacterioBranches:
Two follow-ups after living with the build.

**The quick action buttons are confirmed working end to end.** Pressing one on the desktop runs the toggle in the app — the description above only claimed the channel was consumed and validated, since I had not yet pressed one myself at the time of writing. Same Mac and build as described.

**The extension leaves a sandbox container behind.** Uninstalling shows `~/Library/Containers/<bundle id>.widgets` surviving, which is new: the app itself is not sandboxed and so never had one. It holds no user data — everything the widget reads and writes lives in the shared directory the uninstall paths already remove — and `containermanagerd` owns it, so neither uninstall route can delete it without Full Disk Access; macOS reclaims it on its own. Flagging it rather than adding a removal that would silently fail in the ordinary case, but happy to add a best-effort one if you would rather it tried.

PathGao:
The command channel is the part most likely to be wrong and it isn't: allow-list on both write and read, owner check, consume-once, delete before dispatch, tests for each.

Three things.

**`keyboardLight` is allow-listed but no end runs it.** It's in `WidgetCommandStore.allowedActions`, so `offerable` picks it up and the Settings picker draws a row for it — but `QuickAction.named` has no case for it (no button ever appears) and `WidgetCommandWatcher` breaks on `.keyboardLight` under a comment saying the allow-list rejects it. It doesn't. Ticking that row does nothing, silently. Drop it from `allowedActions` or wire both ends.

**The tests write into the shipping install's channel.** With no bundle, `containerIdentifier` falls back to `com.vorssaint.utils` — the release id. `build.sh --test` on a Mac running the official app with a widget placed: the app consumes the test's `darkMode` and flips appearance, the test then races it, and the closing `removeItem(snapshotURL)` deletes the live snapshot. Give the harness its own suffix.

**A second account's widgets can't work.** `/Users/Shared/Vorssaint/<id>` is created 0755 by whichever user gets there first, so the other one's app can't rewrite the snapshot and their widget can't write a command. `consume()` already treats other accounts as a threat; they're also users.

I read the diff and grepped `main` — none of this is machine-verified.

**Impact:** a dead row in Settings; a test run can toggle appearance on a Mac running the release build.
**Merge:** needs the `keyboardLight` mismatch and the test container id fixed first.

ref #1138

Issue #250, opened 2026-07-16
[Feature]:  add a button like maccy.


sheiy:
### Problem or Motivation

like maccy

### Proposed Solution

<img width="469" height="538" alt="Image" src="https://github.com/user-attachments/assets/3641b955-9a3d-49ba-9701-fe4286ed00af" />

### Feature Area

Menu Bar

vorssaint:
Thanks for the suggestion. I prototyped a dedicated clipboard icon, but a second menu bar icon is currently very unstable on macOS 27 and it was crashing in my tests. Keeping the app stable for the version that is coming has to come first, so I am closing this for now. You can still open the clipboard history from the panel and with a shortcut, and I may revisit a dedicated icon when macOS behaves better.

sheiy:
Got it. Thanks for your work!

PR #1141, opened 2026-08-29
feat(clipboard): show latest copy in the menu bar

adnn-alc:
## Summary

Adds an opt-in "Show latest copy in the menu bar" item next to the main icon (Settings → Clipboard), previewing the most recent clipboard entry with a settable character limit — similar to Maccy's menu bar preview. Clicking it opens the existing clipboard history quick panel, the same one the shortcut opens; right-click opens the usual context menu, same as the metric items — or, with "toggle Keep Awake on right-click" turned on, toggles Keep Awake instead, again matching what right-click already does everywhere else in the menu bar. The main icon's own click behavior is untouched.

It reuses the existing `menuBarSeparateMetrics` mechanism — the same independent, separately clickable `NSStatusItem` already used for CPU/GPU/memory — rather than a new bespoke icon. That distinction matters: #250 asked for something similar and was closed not-planned because a dedicated clipboard icon prototype crashed on macOS 27. The multi-item mechanism this reuses has shipped since 3.1.2, well before that instability was found, and is in production today; #698 is still open and blocked on a macOS 27 menu bar backend, so I'd rather flag that precedent than assume it no longer applies. #884 asks for something adjacent — an always-on preview pane with metadata — but is a different feature than this menu bar item.

Testing this live surfaced several correctness gaps in tracking "what's actually on the clipboard right now," all fixed in this PR. `entries` is literally `pinned + recent`, so `entries.first` is always a pinned item when anything is pinned — never the actual latest copy. Reusing a history entry to paste it again (`copy(entry:)` → `touch()`) deliberately doesn't reorder `entries`, so nothing signaled the preview to update either. A new `latestPasteboardEntry`, tracked separately from `entries`' own order, fixes both: it's set from every path that actually writes to the pasteboard (a fresh copy, reusing a single entry) and falls back correctly for a multi-entry batch paste, which doesn't correspond to one entry's text. It's also cleared the moment `ClipboardAutoClearService` actually empties the real system pasteboard (which `entries` never reflects on its own), and reconciled against `entries`' own `didSet` so removing the referenced entry, Clear Recent, Clear All, or a future history-limit trim all stop the preview from showing stale content — the same reconciliation happens to pick up editing that entry's text, too. Last one: the status item itself now stays alive but blank when there's simply nothing to show, rather than being removed — macOS doesn't reliably restore a dragged position for a status item recreated later, so removing it reset a manually repositioned item back to the default spot the next time something was copied.

## Verification

MacBook Air (M2, 8GB), macOS 26.6.2 (build 25G83), two displays (built-in Retina plus an external 1080p). Installed via `./build.sh --install` and used live across several rounds: enabled the toggle, copied text, reused an old history entry, removed the referenced entry, Clear Recent, Clear All, repositioned the item and re-tested after clearing again.

```sh
./build.sh                     # finishes without warnings
./build.sh --test              # TESTS OK (8946 checks)
```

Not tested: other app languages (11 of the 13 locales are my own translations, not native-checked), a single-display setup, or living with it over several days per the usual feature soak.

## Anything else

No existing issue matches this proposal, so there's no `Refs #N` — #250 (closed, not planned) and #884 (open, different ask) are referenced above as context. Given the direction question around #250/#698, I'd suggest opening this as a draft rather than ready-for-review.

**Edit:** a reviewer caught that `nil` on `latestPasteboardEntry` was overloaded — it meant both "never captured anything yet" (fall back to `recentEntries.first`) and "know the pasteboard was actually cleared," and the fallback treated every `nil` as the first case. `ClipboardAutoClearService` only empties the real pasteboard, never the saved history, so that fallback landed right back on the entry that was just wiped — the clear was a no-op unless history itself was empty. Fixed by seeding `latestPasteboardEntry` from `recentEntries.first` once at load instead of leaving it `nil` until the first capture, then dropping the fallback entirely: `nil` now unambiguously means nothing is currently on the pasteboard. The same gap existed in the capture poll itself, which can detect a pasteboard change and then decide not to record it (an ignored app, a concealed/secret copy, an image with the images toggle off) — it now resets to `nil` as soon as a change is detected, and the recording path sets it back a few lines down only when something is actually captured.

Re-verified live: the auto-clear path now correctly clears the menu bar preview instead of leaving the just-wiped text showing.


PathGao:
`latestPasteboardEntry` is the right split — `entries.first` really is wrong whenever anything is pinned, and reconciling it in `entries`' `didSet` covers remove/clear/trim in one place.

One gap, and it's the one the body claims to close. `pasteboardWasCleared()` nils it, then `syncClipboardPreviewItem` reads `latestPasteboardEntry ?? history.recentEntries.first`. Auto clear only empties the system pasteboard — the history keeps every entry — so nil falls straight back to the copy that was just wiped. The clear is a no-op unless the history is empty.

`nil` is carrying both "cleared" and "never knew". Cheapest split: seed it from `recentEntries.first` in `load()`, drop the fallback in the sync.

Same shape in `captureIfChanged`: it returns early for an ignored app, a concealed copy, or an image with the images toggle off — the pasteboard changed, nothing was recorded, and the item still advertises the previous copy as your latest. `latestPasteboardEntry = nil` right after `self.lastChangeCount = changeCount` covers all three; the recording paths set it back a few lines down.

Not verified on hardware, and I read the two sinks' ordering rather than ran it.

**Impact:** with auto clear on, the menu bar keeps showing text that is no longer on the pasteboard. Same after a copy the history deliberately skips.
**Merge:** needs the nil-means-cleared split first.

adnn-alc:
@PathGao fixed and updated the description, thanks for the quick review!

PathGao:
Both gaps are closed, and seeding at `load()` is the right shape — `nil` means "nothing" everywhere the poll and auto clear touch it.

Except at launch. `baselinePasteboard()` sets `lastChangeCount` to the pasteboard's current change count, so the first poll captures nothing and the seed never self-corrects. Copy something while Vorssaint is quit, then launch it: the menu bar advertises the old entry until you copy again. Either drop the seed (blank until the first copy) or read the content in `baselinePasteboard` and match it with `matchesContent`.

`copy(selectedEntries:)` still sets `nil` for a batch, but the pasteboard is full there — with the fallback gone that renders blank, and the comment on that line still says it "falls back to the most recent entry". One of the two has to change.

Minor: the tooltip comment says `preview` is capped at `previewCharacters` "for every kind". Only `.text` is; `.files` is an uncapped `fileNames.joined(...)`. Copy 300 files, get a 300-name tooltip.

Not verified on hardware — I read the ordering.

**Impact:** after launch the menu bar can advertise a copy that isn't on the pasteboard, for the whole session.
**Merge:** needs the launch seed settled first.

adnn-alc:
@PathGao fixed and confirmed that after relaunch it starts in clean state with nothing showin in the menu bar until copied again.

PathGao:
`load()` staying nil is right — with `baselinePasteboard()` no seed could self-correct, and blank-until-first-copy is the honest state. Batch comment and tooltip cap both fixed.

Three left, none the shape of the last two.

`previewCharacters` is 2000. Capping `.files` with it was the point, but for `.text` it lets a 2000-character tooltip through — copy a page of prose, hover, and it's a wall. Use a tooltip-sized constant, not the list-row one.

`captureStatusClick()` in `onClipboardPreviewClick` is dead here: `position(_:)` centres the quick panel on the pointer's screen, nothing is popover-anchored. Its only reader is `panelAnchor`'s 0.5s fallback for the *main* item, so inside that window it can anchor the main panel at the preview item. Drop it.

Body says "right-click still opens the usual context menu". With `keepAwakeRightClickToggle` on, `onRightClick` toggles Keep Awake. The code matches the metric items — it's the sentence that's off.

Not verified on hardware.

**Impact:** none on pasteboard tracking; a long copy gets an oversized tooltip.
**Merge:** hold for the maintainer's call on the #250/#698 precedent the body raises — nothing has answered it.

adnn-alc:
Tightened the clipboard preview's hover tooltip to a real ~200-character cap instead of letting a full page of text through, and removed a leftover click-position recording that had no effect on this feature but could have nudged the main panel's anchor if it opened moments later.

PathGao:
Tooltip cap and the dropped `captureStatusClick()` are both in, and the comment left on the click handler explains why it's absent — that's the part that stops someone adding it back.

Two left.

`installClipboardPreviewStatusItem` never sets `button.image`. Both existing item kinds do: `updateIconAppearance` sets `emptyStatusImage` for the reason spelled out in its own comment ("lets macOS apply the inactive-display appearance to the title"), and `refreshMetricStatusItems` sets it plus `imagePosition = .noImage` on every pass. Without it the preview text stays full strength on an inactive display's menu bar while the metrics beside it dim.

```swift
button.image = Self.emptyStatusImage
button.imagePosition = .noImage
```

Body still says "right-click still opens the usual context menu". With `keepAwakeRightClickToggle` on, `onRightClick` toggles Keep Awake instead — the code matches the metric items, it's the sentence that's off.

Not verified on hardware.

**Impact:** the preview text doesn't dim with the rest of the menu bar on a second display.
**Merge:** hold for the maintainer's call on the #250/#698 precedent the body raises — still unanswered.

adnn-alc:
Fixed inactive disaplay issue and updated original description.

adnn-alc:
Rebased and fixed merge conflict.

PathGao:
Inactive-display image and the right-click sentence are both in.

I went looking for a sibling to the auto-clear hook and there isn't one. `clearContents()` has 22 call sites on `main`; every one except `ClipboardAutoClearService` either writes new content immediately after, or is `TransientPaste`'s rewrite-then-restore, which must not clear the preview. The one other bare wipe — `resetCutState` cancelling a Finder cut — doesn't call `ignoreNextChange`, so the poll sees it and the new reset in `captureIfChanged` covers it. Auto clear is the only one that needed its own call, and it has it.

Also checked that all three promote paths funnel through `insertPromoted`, and that the nil sits after the `ignoreNextChange` guard rather than before it. Nothing to flag.

Not verified on hardware.

**Impact:** none.
**Merge:** hold for the maintainer's call on the #250/#698 precedent the body raises — still unanswered.

Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
```diff
@@ -868,6 +912,14 @@ final class ClipboardHistoryService: ObservableObject {
...
         normalizeEntryOrder()
         trimToLimit()
+        // latestPasteboardEntry is deliberately left nil here rather than
+        // seeded from recentEntries.first: baselinePasteboard() takes
+        // whatever the pasteboard's change count already is as the starting
+        // point, so a copy made while Vorssaint was quit is never seen as a
+        // change once it launches — a seed here would have no way to
+        // self-correct and could advertise stale content for the rest of
+        // the session. The menu bar preview starts blank and fills in on
+        // the first real capture or reused entry instead.
```
vorssaint: the same stale case is still there when the history is switched off and on again, since copies made while it was off are never seen and the old entry comes back as the latest. and after every launch the item stays blank until the next copy even though the last copy is still on the clipboard. matching what is on the clipboard against the saved entries when the history starts watching again would settle both

Sources/Vorssaint/Services/Clipboard/ClipboardHistorySupport.swift
```diff
@@ -79,6 +79,13 @@ struct ClipboardHistoryEntry: Codable, Equatable, Identifiable {
         }
     }
 
+    /// `preview` collapsed further to a menu bar sized excerpt, for the
+    /// optional "show latest copy" status item.
+    func menuBarText(maxCharacters: Int) -> String {
+        guard preview.count > maxCharacters else { return preview }
+        return String(preview.prefix(maxCharacters)) + "…"
```
vorssaint: a copied image shows up here as bare numbers like 400×300, while everywhere else in the app the word image sits next to them. can the bar and the tooltip use that same label

Sources/Vorssaint/App/StatusItemController.swift
```diff
@@ -637,6 +659,110 @@ final class StatusItemController {
...
+        else {
+            removeClipboardPreviewStatusItem()
+            return
+        }
+
+        // The feature itself is on; keep the item alive even with nothing to
+        // show right now (e.g. right after Clear All) rather than removing
+        // it. macOS does not reliably restore a dragged position for an item
+        // recreated later — it would otherwise land back at the default spot
+        // the next time something is copied.
```
vorssaint: while there is nothing to preview, what stays in the bar, an empty gap? if so it should leave the bar like the metric items do once their reading is gone

Sources/Vorssaint/App/StatusItemController.swift
```diff
@@ -637,6 +659,110 @@ final class StatusItemController {
...
+        //
+        // No fallback to recentEntries here: latestPasteboardEntry is seeded
+        // at load and kept correct from then on (nil means the pasteboard was
```
vorssaint: this still says it is seeded at load, but the load now leaves it empty on purpose

adnn-alc:
@vorssaint Thank you for the feedback! All mentioned issues have been addressed. Please let me know if you need anything else in order to release this.

Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
```diff
@@ -573,13 +616,24 @@ final class ClipboardHistoryService: ObservableObject {
             if baseline {
                 self.lastChangeCount = max(self.lastChangeCount, result.changeCount)
                 self.captureState.didBaseline()
+                // A match means the current clipboard content is a real,
+                // previously captured entry; no match (nothing recorded it,
+                // or it was copied while history was off) leaves the preview
+                // blank rather than guessing.
+                self.latestPasteboardEntry = result.content.flatMap(self.matchingEntry)
```
vorssaint, line 641: the last shown copy is kept while the history is off, so switching it back on brings it back for a moment until the clipboard check answers. clearing it when the history stops would keep the bar blank until then

Tests/MetricsTests.swift
```diff
@@ -16088,10 +16116,23 @@ struct MetricsTests {
...
                          "\(language.rawValue) clipboard bulk-delete format")
         }
+        for language in AppLanguage.allCases {
+            let values = Mirror(reflecting: FeatureStrings.mouseButtons(language)).children
+                .compactMap { $0.value as? String }
+            expect(values.count == 32 && values.allSatisfy { !$0.isEmpty },
+                   "every mouse button string is set for \(language.rawValue)")
+            expect(values.allSatisfy { !$0.contains("—") },
+                   "no em-dash in visible mouse button strings (\(language.rawValue))")
+        }
```
vorssaint: main dropped these two count checks on purpose, since the localization test now covers every clipboard string in every language by itself, and the mouse button one has nothing to do with this change. can you take both out so they do not come back with the merge

Sources/Vorssaint/App/StatusItemController.swift
```diff
@@ -174,6 +182,16 @@ final class StatusItemController {
...
 
+        ClipboardHistoryService.shared.$latestPasteboardEntry
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in self?.syncClipboardPreviewItem() }
+            .store(in: &cancellables)
+
+        ClipboardHistoryService.shared.$isRunning
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in self?.syncClipboardPreviewItem() }
+            .store(in: &cancellables)
```
vorssaint: these two run at launch for everyone, so the clipboard history comes to life, reads its saved file and scans its image folder even when the feature is not installed. today nothing of it loads unless the feature is on, can these only hook up when it is

adnn-alc:
Fixed additional issues, `stop()` now clears `latestPasteboardEntry`.
Two redundant Mirror-based string-cont tests are dropped.
`ClipboardHistoryService.shared` no longer loads for everyone.

Tests/MetricsTests.swift
```diff
@@ -16088,11 +16116,6 @@ struct MetricsTests {
             expect(hub.activeCountFormat.contains("%1$d") && hub.activeCountFormat.contains("%2$d"),
                    "count format keeps positional specifiers (\(language.rawValue))")
         }
-        for language in AppLanguage.allCases {
-            let clipboard = FeatureStrings.clipboard(language)
-            expectFormat(clipboard.deleteSelectedFormat, ["d"],
-                         "\(language.rawValue) clipboard bulk-delete format")
-        }
```
vorssaint, line 16102: the two count checks are gone, but the check that the bulk delete text still carries its number in every language went out with them, and main keeps that one. can you leave it in

adnn-alc:
Apologies for extra removal, added it back.

Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
```diff
@@ -24,8 +24,25 @@ final class ClipboardHistoryService: ObservableObject {
...
-        didSet { entriesStamp &+= 1 }
+        didSet {
+            entriesStamp &+= 1
+            // Keeps latestPasteboardEntry from outliving the entry it points
+            // to: removing it, clearing recent/all, or trimming to a smaller
+            // limit must stop the preview from claiming stale content is
+            // still the latest copy. Re-reading it here (rather than nil-ing
+            // it) also picks up an edit to that same entry's text.
+            if let current = latestPasteboardEntry {
+                latestPasteboardEntry = entries.first(where: { $0.id == current.id })
```
vorssaint, line 35: pinning or unpinning the entry that is currently shown takes it out of the list for a moment, and this drops the preview right there, so the insert that follows never brings it back and the bar goes blank until the next copy. can the preview survive a pin change, for example by looking the entry up again once the move is done

adnn-alc:
Fixed issue where pinning an item removes it from the top bar, checked locally and works as expected.

PR #1142, opened 2026-08-30
feat(clipboard): let Space mark the row the arrow keys are on

andreisuslov:
## Summary

In the clipboard window, Space toggled the preview pane once the arrow keys were driving the list. Space now marks the row the arrow keys are on, the same selection ⌘Return and ⌘-click build, so a multiple selection can be made from the keyboard alone and pasted or copied together with the footer's buttons.

Why: the pane already has two ways in, the toolbar button and the Preview toggle, while the batch selection had no keyboard path that did not need a modifier. Space over a highlighted row reads as "take this one" in Finder, Mail and every list that has a checkbox column, which is the behavior people arrive with.

Space still belongs to the list only after arrow navigation has started, so a space typed into the search field stays a space, and a modified Space is left alone. The predicate that decides this is the same one, renamed from `ClipboardHistoryPreview.handlesSpace` to `ClipboardHistorySpaceKey.handledByList`, since it no longer answers a question about the preview.

What a user notices: ↓ ↓ Space ↓ Space builds a selection without touching the mouse or a modifier.

The same commit is also on #1114's branch, where it was made before this PR existed, and it is declared in the bodies of #1114, #1115 and #1116 for that reason. Whichever merges first brings the change; the other diffs lose it at that point.

## Verification

MacBook Pro 14" (MacBookPro18,3, M1 Pro), macOS 15.7.7 (24G720), one display, English (US). Own build: `./build.sh --dev --install`; `./build.sh --test` green. Checked by hand: arrow to a row, Space marks it and the footer switches to the batch buttons; Space in the search field before any arrow key types a space. Not tested: a keyboard layout where Space carries a modifier by default.


andreisuslov:
One thing worth recording here, since it decides the merge order: the same commit is also on #1114's branch, where it was made before this PR existed. Reviewing #1114 and #1115 you saw the change with no declaration anywhere, which is the fault being fixed now: #1114, #1115 and #1116 all state it in their bodies and credit it here, and the body above notes the duplication.

Whichever merges first brings the change and the other diffs lose it. No code change in this push.

PathGao:
The rename earns its keep: `handledByList` no longer claims to answer a question about the preview, and the tests moved with it.

The body says the pane has "two ways in, the toolbar button and the Preview toggle" — I find one. `toggleQuickPreview()` has a single call site, the toolbar button in `ClipboardQuickPanelView.toolbar`; the sidebar's `onClose` only calls `setQuickPreviewPresented(false)`, which is a way out. Nothing else writes `DefaultsKey.clipboardHistoryQuickPreview`, so there is no Settings toggle either. Space was the preview's only keyboard path, and this leaves the pane mouse-only. #884 is open asking for the preview to be more available, not less.

`handledByList` already declines modified Space, so that case is free:

```swift
if event.keyCode == UInt16(kVK_Space), modifiers == [.option] {
    self.toggleQuickPreview()
    return nil
}
```

Unless mouse-only is the intent — the diff can't tell me that, so put it in the body if so. Code read only, not run.

**Impact:** keyboard users lose the preview entirely.
**Merge:** needs a keyboard path back to the preview, or a body line declaring it mouse-only.

PathGao:
Closing this one, and nothing is lost by it.

The commit here (0f59bfee) is the same one #1114 carries, and the one question the review raised, whether the preview keeps a keyboard path or the body declares it mouse-only, was answered on #1114: its body now declares mouse-only, and that branch has a ship-it. The comment fix from f23f3f8c is also in #1114 (fdc41ee). #1114 is first in the stack, so it is the one that brings the Space change; once it merges this diff would be empty.

Keeping both open would leave the same decision to be argued twice. Thanks for splitting it out so it could be read on its own; that is where the mouse-only call got made.

Issue #1075, opened 2026-08-27
Scroll wheel dead after fast user switching until restart (scroll event taps not rebuilt on session change)


rafihajek:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

With Smooth Scroll and/or Scroll Inverter enabled, the scroll wheel stops working
entirely after switching macOS user accounts (fast user switching) and switching
back. It stays broken until Vorssaint is quit and relaunched. Toggling the scroll
features off/on also fixes it, which points at the event taps rather than the OS.

This happens every time when Vorssaint is running in more than one logged-in
user session at once.

### Steps to reproduce
1. Two macOS user accounts, both logged in (fast user switching), Vorssaint
   running in both (or at least in the affected account).
2. Enable Smooth Scroll and/or Invert Scrolling in the affected account.
3. Switch to the other user, use it for a bit, then switch back.
4. Try to scroll with a mouse wheel.

**Expected:** scrolling works.
**Actual:** scroll wheel input is dead (trackpad-style gestures may vary) until
Vorssaint is restarted or the scroll features are toggled off/on.

### Environment
- Vorssaint 3.3.2 (build 78), installed via Homebrew cask
- macOS 27.0 (build 26A5421a), Apple Silicon
- Two user accounts logged in simultaneously; one Vorssaint instance per session

### Diagnostics
`CGGetEventTapList` while both sessions were logged in shows both instances
holding **active filter taps** (`.defaultTap`, i.e. they can consume events)
with scrollWheel in the mask, and WindowServer has recorded ~4097 ms max
latency on all four scroll taps — right at the tap-timeout threshold,
consistent with WindowServer stalling on a tap owned by a process in the
inactive (suspended) session:

    id=1101513929 tapPID=<my instance>    enabled=Y ACTIVE(filter) scroll=Y maxLat=4097.39ms
    id=468703135  tapPID=<my instance>    enabled=Y ACTIVE(filter) scroll=Y maxLat=4097.10ms
    id=491705403  tapPID=<other session>  enabled=Y ACTIVE(filter) scroll=Y maxLat=4097.67ms
    id=709393584  tapPID=<other session>  enabled=Y ACTIVE(filter) scroll=Y maxLat=4097.01ms

[...]

PR #1145, opened 2026-08-30
fix(scroll): hand both scroll taps back across a fast user switch

iltonandrew:
Refs #1075

Both scroll taps are handed back while this login session is switched away, and built again when it comes back.

Your reading of the dump was the one this follows: the missing half is the resign, not the return. An instance left in a switched-away session keeps a live filter tap with `scrollWheel` in its mask, so the window server still routes every scroll through a process that cannot answer and waits out the timeout on each one — which is the ~4097 ms on all four taps, and why the account actually in use is the one that suffers.

**The re-arm.** You warned that a teardown which only disables the tap loses to the unconditional `tapEnable(true)` on `.tapDisabledByTimeout`. It is handled twice over. The teardown routes through the existing `stop()`, so the tap is gone rather than off, and the re-arm now asks whether this session is on screen before re-enabling — the second one covers the window where a timeout is processed before the resign notification is.

**Teardown joined the pattern the other tap owners already use.** `stop()` in both scroll services did `tapEnable(false)`, removed the run-loop source and dropped the reference, leaving the port to be released whenever the last reference went. Ten of the other tap owners — `TextSnippetService`, `SuperKeyService`, `FinderCutPaste`, `WindowLayoutService` and the rest — call `CFMachPortInvalidate` at that point. The scroll services never joined, and for this bug the difference is the whole point: "release rather than disable" has to be something the code does, not something reference counting gets around to. Both now invalidate, in the same order as the others.

**One predicate owns whether a tap should exist.** `ScrollWheelSupport.tapShouldRun(featureWanted:accessibilityGranted:sessionIsActive:)` is asked by both services' `syncWithPreferences()`, so the session condition cannot drift away from the preference and permission conditions that were already there.

**Starting in a session that is already switched away.** macOS announces that between `willFinishLaunching` and `didFinishLaunching`, which is before `FeatureRuntime.syncAtLaunch()` builds the services that would hear it, so the state is read from `CGSessionCopyCurrentDictionary()` at startup rather than assumed, and the watcher subscribes before it reads so a switch landing during startup is not dropped between the two.

That reading fails open, and the asymmetry is the reason. The call is documented as nullable and has been seen returning nothing in an ordinary GUI session, so an unreadable answer is not exotic. A mistaken off would be permanent — an already-active session is never told it became active, so both taps would stay down for the life of the process, on a Mac that may only ever have had one account. A mistaken on costs nothing, because the next resign notification corrects it. Only an explicit, valid false counts as off screen.

**Scope.** The two scroll services are the ones with evidence, and they are what this changes. The same shape reaches the other tap owners — a session switched away is a bad place for any filter tap — but nobody has reported it there and joining twenty services to a new watcher without a report is a much larger change than this one; `SessionActivity.onChange` is a line each whenever that is wanted.

**Tests.** `tapShouldRun` and the console-flag reading are behaviour tests on the pure helpers (the flag covers a missing dictionary, an absent key, and both the `Bool` and `NSNumber` spellings). The wiring is pinned as text, since the tap owners cannot be reached from the `--test` list: each service follows the session, guards its re-arm, and invalidates its port. Checked in both directions — green as committed, red with the guard removed, red with either subscription removed, red with the resign half dropped from the watcher.

**Verification.** macOS 26.6.2 (25G83), M5 Pro, own Developer build, single logged-in account. `./build.sh` finishes warning-free and `./build.sh --test` passes 9105 of 9106, the exception being the pre-existing `dispatch pool starvation` check that fails on this 15-core machine and passes in CI.

**Not verified, and it is the part that matters.** I have one account on this machine, so I have not reproduced the bug or its fix. What needs two logged-in sessions: that the resign arrives early enough to tear down before the session stops being serviced, that the old session's tap PIDs actually leave `CGGetEventTapList`, that coming back rebuilds exactly the intended taps and no duplicates, and that the dead wheel and the ~4097 ms latency are genuinely gone — with Smooth Scroll alone, the inverter alone, and both together. The API behaviour supports the chain; it does not prove it. If that is worth a draft rather than a merge until someone with two accounts confirms, say so and I will mark it.

### DIFF
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -238,6 +238,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         // Takes the Super key mapping back out before the process goes away.
         SuperKeyService.shared.suspend()
         MiddleClickService.shared.suspend()
+        ScrollInverter.shared.suspend()
         SmoothScrollService.shared.suspend()
         MouseNavigationService.shared.suspend()
         DockPreviewService.shared.stop()
--- a/Sources/Vorssaint/Services/ScrollInverter.swift
+++ b/Sources/Vorssaint/Services/ScrollInverter.swift
@@ -48,9 +48,9 @@ final class ScrollInverter: ObservableObject {
         let wanted = AppFeature.scrollInverter.isAvailable
             && (defaults.bool(forKey: DefaultsKey.scrollInverterEnabled)
                 || defaults.bool(forKey: DefaultsKey.scrollInverterHorizontalEnabled))
-        if ScrollWheelSupport.tapShouldRun(featureWanted: wanted,
-                                           accessibilityGranted: Permissions.shared.accessibility,
-                                           sessionIsActive: SessionActivity.shared.isActive) {
+        if SessionActivitySupport.tapShouldRun(featureWanted: wanted,
+                                               accessibilityGranted: Permissions.shared.accessibility,
+                                               sessionIsActive: SessionActivity.shared.isActive) {
             start()
         } else {
             stop()
@@ -64,6 +64,9 @@ final class ScrollInverter: ObservableObject {
 
     private func start() {
         guard tap == nil else {
+            if let tap {
+                CGEvent.tapEnable(tap: tap, enable: true)
+            }
             MouseAppExceptions.shared.setSourceTracking(true, for: .scrollDirection)
             isRunning = true
             return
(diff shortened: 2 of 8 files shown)


PathGao:
Merged already, so this is a follow-up, and it's against `main` — the tap-create retry picked up a one-shot guard and eight more services joined `SessionActivity` since.

That spread vindicates `tapShouldRun`. Every service that joined after this one — `MouseNavigationService`, `QuitProtectionService`, `MiddleClickService`, `MouseButtonShortcutService`, and `SmoothScrollService` itself — recomputes the full predicate in its `tapDisabledByTimeout` branch, and on a false result tears the port down and resyncs.

`ScrollInverter.handle` is the last holdout on the shape this PR shipped:

```swift
if SessionActivity.shared.isActive, let tap {
    CGEvent.tapEnable(tap: tap, enable: true)
}
```

It ignores `featureWanted` and the Accessibility grant, and when it declines to re-arm it leaves a disabled-but-owned port — the state your own `stop()` comment says a switched-away session must not be in.

Not a repro: the resign handler calls `stop()` synchronously, so `isActive == false` with a live tap looks unreachable on the main run loop. It's the divergence I'd close, not a failure I can trigger.

**Impact:** none I can reproduce; it is the one copy of the pattern that will read wrong to whoever extends it next.
**Merge:** already merged — a one-line follow-up on `ScrollInverter`, not a revert.

ref #1075

Issue #1098, opened 2026-08-28
Automatic window capture omits attached sheets and modal dialogs


youkq95:
### Before you file

- [x] Vorssaint can capture other application windows, so its Screen Recording permission is active.
- [x] This still happens on the currently installed version (3.3.2).

### Description

Vorssaint's **automatic window capture** omits sheets, alerts, modal dialogs, or other attached/child windows displayed on top of the selected application window.

When the screenshot tool automatically highlights/snaps to a window and that window is clicked, only the selected main window is captured. A visible dialog stacked on top of it is missing from the resulting image.

A true **manually dragged region capture works correctly** and includes everything visible. The built-in macOS window capture (`Shift-Command-4`, then `Space`) also includes the main window and its attached dialog correctly.

My original report described this as a System Settings-specific problem and incorrectly said that region capture also failed. After retesting with Mail.app, the behavior is more general and limited to automatic window capture.

**Expected behavior:** Capturing a window should include its visible attached sheets, alerts, and modal/child windows, matching the built-in macOS window capture.

**Actual behavior:** Only the selected main window is captured; windows attached or stacked above it are omitted.

### Steps to Reproduce

1. Open an application that displays a sheet, alert, or modal dialog over its main window (Mail.app is one example).
2. Invoke Vorssaint's screenshot tool.
3. Let the tool automatically snap/highlight the application window and click it.
4. Observe that the resulting image contains the main window but omits the visible dialog.
5. Repeat using a manually dragged rectangular region; the dialog is captured correctly.

### Feature Area

Tools (Screenshot / screen capture)

### Screenshots / Recordings

**Vorssaint automatic window capture — dialog omitted:**

![Vorssaint capture missing the Mail modal dialog](https://github.com/user-attachments/assets/ed5544c1-b04f-49a4-82dd-69f36a7f43e1)

**Built-in macOS window capture — dialog included:**

![Built-in macOS capture including the Mail modal dialog](https://github.com/user-attachments/assets/1cf18577-218e-4cd7-a73b-25b13dbbe639)

### Vorssaint Version

3.3.2 (build 78)

### macOS Version

macOS 26.6.2 (build 25G83)

### Language

Simplified Chinese (`zh-Hans-CN`)

### Mac and displays

MacBook Air (Mac17,3, Apple M5, 24 GB) · built-in 2560×1664 Liquid Retina display + external 3840×2160 display (main)

### Installation Method

Homebrew

PathGao:
[...]

PR #1146, opened 2026-08-30
fix(capture): draw what an app stacked on the window being captured

iltonandrew:
Refs #1098

Draft on purpose: the reasoning and the tests are here, the hardware run is not. I have no second display and have not put the reporter's Mail dialog in front of this build, and the last section says exactly what that leaves open.

A capture of a clicked window now draws what the app has stacked on it.

macOS gives a sheet a window of its own, so both single-window routes — `WindowPreviewProvider.captureViaWindowServer` and the `SCContentFilter(desktopIndependentWindow:)` fallback — return the window without it, which is why the dialog that is plainly on screen is missing from the file. `ScreenshotCapturePolicy.attachedCapturePlan(target:frontToBack:)` decides what belongs in the shot, and `SCContentFilter(display:including:)` draws them together. Nothing attached means `nil`, and the existing fast path answers exactly as before.

**The rule, and why it is this strict.** Same application, in front of the clicked window, and lying entirely within it. The relationship is not in the window list, and the Accessibility API that does know it would put a second permission behind a feature that needs only Screen Recording, so the shape a sheet always has stands in for it.

Containment is doing the work, and the case that set it is a Mail compose window over the message viewer: same app, in front, overlapping, smaller. An earlier version of this took anything smaller in both directions and would have drawn it into the shot. Putting a window in the picture that the person did not choose is worse than leaving a dialog out of it — the file then quietly holds something they never selected, which is a privacy question rather than a fidelity one — so anything reaching past the edge is left to the ordinary capture. Containment also takes a sheet exactly as wide as its parent, which the "smaller" rule would have thrown away, and it keeps the captured area the clicked window's own frame, so the preview placement and the stored capture metadata still describe the same rectangle.

**What this gives up.** A dialog hanging over its parent's edge is not contained, so it is left out and that capture behaves as it does today. It degrades to the current behaviour rather than risking the wrong picture.

**Scope.** The screenshot path only. The recorder reaches the same defect through `RecorderCaptureEngine.filter(for:in:excluding:)`, and it is filed separately (#1147) because a live stream has to keep answering the question as sheets come and go, which a filter fixed at start time cannot — likely a different mechanism, and not something to bolt onto this. The switcher's thumbnail route is left alone, as you said.

**Tests.** Nine behaviour cases on the pure decision: a sheet joins with the window drawn first; a full-width sheet joins; two stacked windows are ordered back to front; a window behind is not attached; another app's window is not; a same-app window reaching past the edge is not; a non-touching window is not; nothing attached returns `nil`; and a target missing from the list does not treat everything as being in front of it.

**Verification.** macOS 26.6.2 (25G83), M5 Pro, single 1× display, own Developer build. `./build.sh` finishes warning-free and `./build.sh --test` passes 9096 of 9097, the exception being the pre-existing `dispatch pool starvation` check that fails on this 15-core machine and passes in CI.

**Measured, on 26.6.2.** A harness put an `NSAlert` sheet and an app-modal dialog over a window of known frame and read the window list back. Both come back as their own window, same pid, **layer 0**, ordered **in front** of their parent, and **entirely inside** it with zero overhang on every edge:

```
COMPARE role=SHEET     samePID=true layer=0 contained=true inFront=true
        child={x=626,y=438,w=260,h=138} parent={x=356,y=207,w=800,h=601}
        overhang={left=0,right=0,top=0,bottom=0}
COMPARE role=APP_MODAL samePID=true layer=0 contained=true inFront=true
        child={x=526,y=377,w=460,h=261} parent={x=356,y=207,w=800,h=601}
        overhang={left=0,right=0,top=0,bottom=0}
```

Those were the two ways this could have been a no-op: the window list is filtered to layer zero, and the rule wants strict containment. Both hold for the two kinds of window the issue is about.

**Still not verified.** The reporter's own Mail dialog, since a system-provided dialog is not guaranteed to behave like an `NSAlert`; and whether `SCContentFilter(display:including:)` draws the second window as expected, which needs Screen Recording granted to something I can watch. Multi-display is no longer on this list — the source rectangle that depended on how a display origin is read is gone, and the shot is cropped with the helpers the region capture already uses.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotCaptureEngine.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotCaptureEngine.swift
@@ -186,9 +186,9 @@ enum ScreenshotCaptureEngine {
     }
 
     /// Draws the clicked window together with what the app stacked on it.
-    /// ScreenCaptureKit is the only route that takes more than one window, and
-    /// the source rectangle keeps the result to the area they cover instead of
-    /// the whole display. `nil` leaves the single-window routes to answer.
+    /// ScreenCaptureKit is the only route that takes more than one window. It
+    /// captures their composited display, then crops it to the area they cover.
+    /// `nil` leaves the single-window routes to answer.
     private static func captureAttached(
         _ plan: ScreenshotCapturePolicy.AttachedCapturePlan) async -> CGImage? {
         guard let content = try? await SCShareableContent.excludingDesktopWindows(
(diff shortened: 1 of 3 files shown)


iltonandrew:
Two of the three things I could not check are now measured on 26.6.2, and the third is gone rather than answered.

A harness put an `NSAlert` sheet and an app-modal dialog over a window of known frame and read the window list back. Both are their own window, same pid, layer 0, ordered in front, and entirely inside the parent with zero overhang on every edge — so the layer-zero filter does not drop them and strict containment does not exclude them. Those were the two ways this could have quietly been a no-op.

The multi-display worry is no longer a worry: the source rectangle that depended on how a display origin is read is gone, and the composite is cropped through the same helpers `captureDisplayRegion` already uses, which are covered by the existing tests. I would rather delete an assumption I cannot exercise on one display than ship it untested.

Leaving it a draft for two reasons, both about the picture rather than the decision: your Mail dialog specifically, since a system-provided dialog need not behave like an `NSAlert`, and whether `SCContentFilter(display:including:)` actually draws the second window — that needs Screen Recording granted to something I can watch pixels come out of. Mark it ready if you would rather test it on the branch.

PR #1149, opened 2026-08-30
feat(i18n): add Arabic, the first right-to-left language

Mulhamx4:
## Summary

Adds **العربية** as a supported interface language, and with it the layout
direction handling that a right-to-left language needs. Arabic is the first RTL
locale in the app, so the change is in two parts.

**Translation.** Every catalog gains an `ar` case: the 942-field `Strings`
struct in `Core/Localizations/Strings+Arabic.swift`, plus the 43 feature
catalogs under `Core/`. As CONTRIBUTING describes, the compiler is the
completeness check here, so nothing can silently fall out of sync.
`Resources/ar.lproj/InfoPlist.strings` carries the permission prompts and the
locale is registered in `Resources/Info.plist`.

**Layout direction.** macOS only mirrors an app when the *system* language is
right to left, and Vorssaint picks its language from its own setting, so
choosing Arabic on an English Mac flipped nothing. `AppLanguage.isRightToLeft`
marks the language and a new `appLayoutDirection()` modifier sets
`\.layoutDirection` at every point where SwiftUI meets AppKit — all 34 root
construction sites, plus the 3 places a host replaces its `rootView` in place.
Three of them name their root view's type for that reason (`BrightnessOSD`, the
cleaning overlay, the capture guide), so they carry `MirroredView<Content>`
instead. That part is +68 / -43 across 27 files; the rest of the diff is
translation.

Nothing else changes. No string was reworded, no view was restructured, and
every other language behaves exactly as before — the RTL branch is dead code
for all thirteen of them.

Two small things follow the pattern each existing language already has: the
Arabic OCR recognition languages in `MediaSupport.recognitionLanguages`, which
both call sites filter against `supportedRecognitionLanguages()` so it degrades
to English where the OS has no Arabic model, and the `ar` resource folder in
the Command Bar's System Settings lookup.

On direction: CONTRIBUTING documents how to add a language, so I took the
locale itself as settled. The RTL mechanism is the part worth a look — happy to
split it into its own pull request, or to drop it and ship the translation
alone, if you would rather take those separately.

## Verification

MacBook Pro 14" M3 Pro (Mac15,6), macOS 26.5.2 (25F84), Xcode Command Line
Tools, Swift 6.3.3, single built-in display, system language English. Release
build for the checks below; a `--dev` build for the on-screen pass.

```sh
./build.sh                     # clean release build, no warnings
./build.sh --test              # TESTS OK (9470 checks)
```

The bundle assembles with `ar.lproj` present and 14 entries in
`CFBundleLocalizations`. I also checked field parity mechanically: the Arabic
`Strings` catalog has the same 942 fields in the same order as `enUS`, matching
every other language file.

`Tests/MetricsTests.swift` picks Arabic up on its own where the suites iterate
`AppLanguage.allCases`, so the format-specifier and non-empty contracts already
cover it. I added Arabic to the four explicit per-language tables and one line
asserting Arabic is the only RTL language.

**What I could not test:** the mirrored layout on screen. Everything above is
compile- and test-verified, but I have not run the app in Arabic across all its
surfaces, so I cannot tell you how each panel *looks* mirrored. The places I
would check first are the ones where mirroring is arguably wrong rather than
merely different — the recorder editor's timeline, the screenshot editor's
tool rail and the radial menu. I also have not tested on a Mac whose system
language is Arabic, where AppKit's own direction and this setting agree rather
than being set independently, nor on an external display or multi-Space setup.
No Intel Mac, and no notarized build.

## Anything else

No issue exists for this — I searched open and closed issues for Arabic, RTL
and right-to-left and found nothing — so there is no `Refs` line to give. Happy
to open one first if you would rather settle direction there; I can also close
this and reopen it against an issue number.


PathGao:
Checked the two claims I could: both OCR call sites do filter against `supportedRecognitionLanguages()`, and `resourceFolder(for:)` is an exhaustive switch, so the compiler really is the check there.

One gap. `MenuPanelView.OverlayScrollView.makeNSView` builds a second root — `HeightReportingHostingView(rootView: content)`. Environment doesn't cross an `NSViewRepresentable` into a fresh `NSHostingView`, and `NSView.userInterfaceLayoutDirection` defaults to the app's, not the superview's. So `.appLayoutDirection()` on `MenuPanelView()` mirrors the header and section nav while the scrolled body — and the metric detail below it — stay LTR:

```swift
let host = HeightReportingHostingView(rootView: content.appLayoutDirection())
```

I count 34 `rootView:` construction sites on main, not 31. The ones I'd check are that one, `PermissionGuideOverlay` and the two in `QuickToolHUD`. My copy of the diff was truncated at 80 KB so I never saw those files — if they're already covered, say so. I can't see the mirrored layout either, same limit you hit.

**Impact:** Arabic users get a mirrored shell with LTR content in the menu bar panel, the app's main surface.
**Merge:** needs those three hosts confirmed first.

Mulhamx4:
Thanks for digging into this — and yes, all four are already covered. Your diff
truncation hid the exact lines, so here they are:

- `OverlayScrollView.makeNSView` — [MenuPanelView.swift#L2132](https://github.com/Mulhamx4/vorssaint-utils/blob/9a6e44ea3bb51cbfaec742441bf2858847008291/Sources/Vorssaint/UI/MenuPanel/MenuPanelView.swift#L2132)
- `OverlayScrollView.updateNSView` — [MenuPanelView.swift#L2150](https://github.com/Mulhamx4/vorssaint-utils/blob/9a6e44ea3bb51cbfaec742441bf2858847008291/Sources/Vorssaint/UI/MenuPanel/MenuPanelView.swift#L2150)
- `PermissionGuideOverlay` — [PermissionGuideOverlay.swift#L50](https://github.com/Mulhamx4/vorssaint-utils/blob/9a6e44ea3bb51cbfaec742441bf2858847008291/Sources/Vorssaint/UI/PermissionGuideOverlay.swift#L50)
- `QuickToolHUD` — [L123](https://github.com/Mulhamx4/vorssaint-utils/blob/9a6e44ea3bb51cbfaec742441bf2858847008291/Sources/Vorssaint/Services/QuickTools/QuickToolHUD.swift#L123) and [L159](https://github.com/Mulhamx4/vorssaint-utils/blob/9a6e44ea3bb51cbfaec742441bf2858847008291/Sources/Vorssaint/Services/QuickTools/QuickToolHUD.swift#L159)

The line you wrote is byte-identical to what is on the branch:

```swift
let host = HeightReportingHostingView(rootView: content.appLayoutDirection())
```

You diagnosed a real hazard — environment genuinely does not cross an
`NSViewRepresentable` into a fresh `NSHostingView` — it just happens to be the
one that made me change `installReporter` and `Coordinator` to
`HeightReportingHostingView<MirroredView<Content>>`, which is why that file shows
8 changed lines rather than 2. `updateNSView` needed it too, or the body would
have reverted to LTR on the first height re-measure.

**On the count: you are right and I was wrong.** There are 34 construction
sites, not 31. I had counted only the direct `NSHostingView` /
`NSHostingController` calls and left out the three subclass constructions
(`OverlayHostingView`, `HeightReportingHostingView`, `PassThroughHostingView`) —
all three are covered, but the number in the description was not counting them.
I have corrected the description to 34.

I then re-checked it mechanically instead of by eye: walking every
`*Hosting*(rootView:)` construction and every `.rootView =` assignment in
`Sources/`, taking the balanced expression at each, gives **34 construction
sites + 3 in-place reassignments = 37, and all 37 carry `appLayoutDirection()`**.
The three reassignments are `BrightnessOSD:33`, `MenuPanelView:2150` and
`ScreenshotSelectionController:755`; each one is a host that keeps its
controller and swaps the root view, so missing one would look exactly like the
bug you described, just triggered later.

One more in the same family, since your reasoning applies past hosting views:
`OverlayScrollView` is the only `NSViewRepresentable` here that hosts SwiftUI.
The other twenty host AppKit views, and of those the text ones — `PlainTextEditor`,
`ClipboardTextPreview`, `MarkdownPreview`, the Homebrew search field, the mixer
percent field — are deliberately untouched: `NSTextView` and `NSTextField`
default to `baseWritingDirection = .natural`, which resolves per paragraph from
the first strong character. So an Arabic note already lays out RTL and an English
one stays LTR, in every interface language. Forcing those to follow the interface
direction would be the regression, not the fix.

What I still cannot give you is the screenshot. The mirrored layout is compile-
and test-verified only, so the review you flagged as needed is exactly the one I
could not do myself.

Mulhamx4:
Ran it on screen, which closes the gap I flagged as untestable — and it turned
up one real bug, so the pass was worth it.

**MacBook Pro 14" M3 Pro (Mac15,6), macOS 26.5.2 (25F84)**, `--dev` build,
single built-in display, **system language English** with the app set to Arabic
— the harder of the two cases, since AppKit and the app setting disagree there.

**The one you asked about is fine.** The menu bar panel mirrors end to end:
header, section nav, and the scrolled body below it — mixer, fan control, quick
toggles, keep awake with its Options group expanded. `OverlayScrollView` was
covered, so the second root behaves.

Also checked: Settings (sidebar right, content left, section reordering list),
the recorder editor, the screenshot editor, the capture palette overlay, the
radial menu and the feedback window. No layout breakage, no clipping, no
stranded LTR block.

**Found and fixed: the chevrons never mirrored.** `chevron.right` and
`chevron.left` are fixed-direction SF Symbols — they ignore `\.layoutDirection`
entirely. So every disclosure and navigation arrow sat on the correctly mirrored
side of its row while still pointing the unmirrored way. Most visible on the ten
quick-toggle rows and the panel's Options rows.

`chevron.forward` / `chevron.backward` are the mirroring pair, and this codebase
already used them in five places, so it is your idiom rather than a new one. I
moved the remaining 22 over — all 22 are directional (disclosure indicators,
Dock Preview prev/next, back buttons); `chevron.down` is untouched. **No visual
change in the other thirteen languages**: SF Symbols draws `.forward` exactly as
`.right` in a left-to-right layout. That is the second commit.

**A limit I cannot fix, so I would rather document it.** The window's traffic
lights stay top-left while the content mirrors, because `NSWindow` chrome
follows `NSApp.userInterfaceLayoutDirection` — derived from the *system*
language — and there is no public per-window override. It only shows when the
system is LTR and the app is set to Arabic; on an Arabic-language Mac macOS
picks `ar.lproj`, now that it is registered, and the chrome mirrors on its own.
Say the word if you would rather I note it in the docs.

**Still not exercised:** the brightness OSD (the third in-place `rootView`
swap — the other two, `OverlayScrollView` and the capture guide, are confirmed
above), switching language live with a window already open, an external display
or multi-Space layout, and a Mac whose system language is actually Arabic.

`./build.sh` clean with no warnings, `SELFTEST OK`, `TESTS OK (9470 checks)`
after the chevron commit.

Mulhamx4:
Screenshots, since neither of us could see the layout. `--dev` build, macOS
26.5.2 (25F84), system language English with the app set to Arabic.

**The menu bar panel — the surface you flagged.** Header, section nav and the
scrolled body below them, all mirrored together:

image


**Settings**, sidebar on the right, content on the left:

image


**The chevron bug, before and after.** This is the one the on-screen pass
turned up and the compiler could never have: `chevron.right` kept pointing
right while its row mirrored around it.

Before:

image

After:

image

And the Keep Awake panel with its Options group expanded, where the Automation
row's chevron now points into the reading direction:

image


The traffic lights in the Settings shot are the `NSWindow` chrome limit I
mentioned: they follow the system language, not the app setting, so they stay
left here and would mirror on their own on an Arabic-language Mac.

PathGao:
The on-screen pass is what this needed — the chevron bug is exactly what compiling can't find. Count reconciles: I get 27 `chevron.right`/`.left`/`.forward`/`.backward` on main, so your 22 + 5 covers all of them.

One sibling the sweep couldn't reach, because it isn't a chevron. `SuperKeySettings.swift`, the `diagram` view:

```swift
Image(systemName: "arrow.right")   // → arrow.forward
```

It sits between the held key and the modifier caps it stands for. The HStack mirrors, the glyph doesn't, so in Arabic it points from the modifiers back at the source key. I checked the other `arrow.*` sites on main: window layout and the screenshot arrow tool are spatial and must not mirror, undo/redo already use the `.uturn.forward`/`.backward` pair. This is the only one I'd change. Reading main, not your branch — if the chevron commit already caught it, ignore.

Traffic lights: I'd keep it out of user docs. There is no fix, and it resolves itself on an Arabic-language Mac. One comment line next to `appLayoutDirection()` spares the next person the hunt.

**Impact:** one settings diagram points backwards in Arabic. Nothing else.
**Merge:** needs the `arrow.forward` swap first.

Mulhamx4:
Done, both of them — `7e4a837`.

`arrow.forward` in the super key diagram. You had it exactly right: the glyph
sits inside the diagram's `HStack`, the stack mirrors and the arrow did not, so
in Arabic it pointed from the modifier caps back at the held key. The chevron
sweep never reached it — that commit only moved `chevron.*`, so this was still
`arrow.right` on my branch too, not just on main.

I re-ran your audit rather than take it on trust, and it holds: that was the
only bare `arrow.right` / `arrow.left` in the app. The resize and window-layout
glyphs (`arrow.up.left.and.arrow.down.right`, `arrow.left.arrow.right`,
`arrow.up.arrow.down`) describe physical screen positions and would be wrong
mirrored, undo/redo are already on the `arrow.uturn.forward` / `.backward` pair,
and the ten `rectangle.lefthalf` / `righthalf` layout glyphs are spatial for the
same reason — a left half is a left half in any language.

Traffic lights: agreed, and out of the user docs. It is now a note on
`AppLayoutDirection` itself, saying what the limit is, that AppKit chrome
follows the system language with no per-window override, and that it agrees with
itself on an Arabic-language Mac. Next person hits the explanation where they
hit the code.

Clean release build with no warnings, `SELFTEST OK`, `TESTS OK (9470 checks)`.

Worth recording that both of these came out of the same blind spot rather than
two separate mistakes: fixed-direction SF Symbols are invisible to the compiler
and to every test here, and they only show up in a language that mirrors — which
is why nothing caught them in thirteen years of left-to-right. If you ever want
it enforced, a grep for `chevron.right|chevron.left|"arrow.right"|"arrow.left"`
in CI would hold the line cheaply now that the codebase is clean of them.

Three surfaces I still have not exercised, unchanged from before: the brightness
OSD, switching language live with a window already open, and a Mac whose system
language is actually Arabic.

PathGao:
The chevron sweep reconciles against main: 21 bare `chevron.right`/`chevron.left` plus the `chevron.right.circle.fill` in `ShelfTilesView` is your 22, and the super key `arrow.right` was the only bare one. That family is clean.

One sibling outside it. `ShelfTilesView` is an `NSViewRepresentable` that is neither SwiftUI nor text, so your enumeration didn't reach it. `ShelfTileLayout.tileFrame` puts column 0 at `inset` and walks right — absolute frames in a flipped document view, which AppKit will not mirror. Assuming `ShelfService`'s two roots carry `appLayoutDirection()` as you say, the shelf header and buttons mirror while the tile grid still fills from the left. Most obvious with fewer items than columns. It is the only representable here that orders a list by index.

```swift
let column = rtl ? safeColumns - 1 - index % safeColumns : index % safeColumns
```

The shelf wasn't in your on-screen pass, and I can't see it either — this is read off main, so if the branch already handles it, ignore.

**Impact:** shelf tiles read left to right inside a mirrored panel, Arabic only.
**Merge:** needs the tile grid mirrored, or a line saying it stays LTR on purpose.

Mulhamx4:
You were right, and the sweep missed it for exactly the reason you gave — fixed
in `5fe43c97`.

`ShelfTilesView` is an `NSViewRepresentable` hosting AppKit views rather than
SwiftUI, so it is not a hosting root and my enumeration only ever walked
`rootView:` sites. `tileFrame` then lays tiles out as absolute frames in a
flipped document view, which nothing mirrors. The column is now counted in from
the other edge when the interface is right-to-left; the row is untouched, so
only horizontal order changes. It is defaulted to `false`, so every existing
caller and all thirteen left-to-right languages keep the frames they had. Two
tests cover it, including the not-full row you pointed at.

That is one of two `NSViewRepresentable`s here that needed thought — the other
is `OverlayScrollView`, which hosts SwiftUI and was already covered. The
remaining nineteen host AppKit views where `.natural` writing direction is the
right answer, as before.

## Catching up with main

I had been away two weeks, so this is also rebased onto current main — 271
commits, and the repository moved to `vorssaint/` in the meantime.

Everything that landed in that window is translated: the Dynamic Island family,
quit protection, the click-debounce filter and the shelf promise strings, plus
106 new fields across fifteen existing catalogs and 36 in the base `Strings`.
Four fields that left main left the Arabic catalog with them.

Re-running both sweeps against the new main caught regressions worth knowing
about, since they are the same two failure modes and neither is visible to the
compiler:

- **Three new hosting roots** with no layout direction: the notch's own two in
  `NotchWindowHost` and the disk image installer's progress card.
- **Eight new fixed-direction symbols**: the notch calendar's month arrows, two
  notch back buttons, the power and brightness rows, and two in the recorder
  inspector.

Both are clean again — 40 `rootView` sites all carrying the modifier, zero bare
`chevron.right`/`.left`/`arrow.right`/`arrow.left`. This is the third time the
pair has drifted, which is the argument for the CI grep I mentioned; happy to
add it here if you want it.

## Three lint rules reached Arabic and it failed them

The new harness checks things the old one did not, and Arabic was the first
language to meet them. It had an em-dash in two captions, eleven labels ending
in three periods instead of the ellipsis character, and 114 guillemets.

The quotes are the one worth a sentence. I had used « » throughout, because they
are common in Arabic print — but the rule here is what the system ships, and
`Locale(identifier: "ar").quotationBeginDelimiter` reports `U+0022` on this Mac,
not the angled pair. So the test was right and my copy was wrong: Arabic now
uses the curly pair, like every other language here that is not French, Russian
or German. Say the word if you would rather it follow Arabic print convention
instead and the rule gain a third language.

`ar.lproj` also gains the calendar prompt the notch introduced.

## Verification

Same machine as before: MacBook Pro 14" M3 Pro (Mac15,6), macOS 26.5.2 (25F84),
system language English.

```sh
./build.sh                     # clean release build, no warnings
./build.sh --test              # TESTS OK (52299 checks)
```

One thing worth flagging rather than hiding: on my first runs after the rebase I
saw 16 to 22 failures in `SwitcherScrollTests`, varying run to run. Current main
was stable across four runs, my branch was stable in a clean worktree, and my own
checkout has since been stable across four more — the difference was a live
Vorssaint instance on this Mac while those window-server tests ran. I could not
reproduce it afterwards, so I am not claiming it is a real defect, only that
those tests appear sensitive to a running copy of the app.

**Not re-checked on screen since the rebase:** the two-week feature surface is
large, and my earlier visual pass predates all of it. The notch in particular is
a whole new mirrored surface nobody has looked at in Arabic, and the shelf grid
change above is tested but not seen. Also still unexercised, unchanged from
before: the brightness OSD, switching language live with a window open, and a
Mac whose system language is Arabic.

Mulhamx4:
On-screen pass done after the rebase, and it found a third one in the same
family — fixed in `a01568a3`.

**The notch level bars filled from the left.** Volume, brightness and playback
position are all one control: an `NSSlider` behind a custom cell. The
environment does cross into an `NSViewRepresentable`, but an AppKit control does
nothing with it, so the bar stayed anchored left inside a mirrored island. Only
visible by running it.

Handed the direction, `NSSliderCell` mirrors its own tracking — a probe here
reads 0.06 for a click near the left edge left-to-right and 0.94 mirrored — so
pointer input needed nothing beyond passing it through. The fill is drawn by
hand and did need mirroring, and that geometry moved to
`NotchLevelBar.fillRect` in `NotchSupport`, beside the other notch helpers, so
it is covered: a quarter-full bar starts at the trailing edge, a full or empty
one is identical either way, and a value that is not a number draws nothing.

That is three now — the shelf grid, this, and the switcher's own representable
was already fine — and all three are the same blind spot you first pointed at:
**my sweep only ever walked SwiftUI hosting roots, and an `NSViewRepresentable`
hosting AppKit is not one.** The two in this codebase that draw or place
anything horizontally are now both handled. The remaining nineteen host text or
media, where `.natural` writing direction is the right answer.

## Verified on screen

MacBook Pro 14" M3 Pro (Mac15,6), macOS 26.5.2 (25F84), `--dev` build, system
language English with the app in Arabic.

- **The shelf tile grid** — your merge condition. Checked with fewer items than
  columns, which is where it shows. Tiles start at the trailing edge.
- **The notch level bars** — after the fix above.
- **The power and brightness rows** in the panel, whose chevrons were two of the
  eight upstream reintroduced.
- **The brightness OSD** and **switching language with a window already open** —
  both were open from the first review and both are now closed. The OSD was the
  third in-place `rootView` swap; the other two were confirmed earlier.

## What I am not claiming

The notch is large and this was not an exhaustive audit of it. The volume bar
was found and fixed, and what was looked at afterwards was clean, but the
calendar, the music and lyrics views, notifications and the downloads list have
not each been walked through in Arabic. Given that three of these bugs have now
come from AppKit-backed views rather than SwiftUI, that is where I would look
first if anything else turns up.

Still unexercised, unchanged: an external display or multi-Space layout, and a
Mac whose system language is actually Arabic.

`./build.sh` clean with no warnings, `SELFTEST OK`, `TESTS OK (52302 checks)`.

Resources/ar.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,26 @@
+/* macOS Automation permission prompt. */
+"NSAppleEventsUsageDescription" = "يستخدم Vorssaint الأتمتة لنقل الملفات في Finder، وإرسال العناصر إلى المهملات، وفتح أوامر Terminal عندما تطلب إجراءات Homebrew.";
```
vorssaint, line 2: this one still describes an older version of the prompt. the english now says it moves files, sends items to the trash, opens requested commands and controls playback in your media apps, so can you bring it in line with that

Resources/ar.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,26 @@
...
+
+/* macOS Screen & System Audio Recording permission prompt. */
+"NSAudioCaptureUsageDescription" = "يصل Vorssaint إلى صوت كل تطبيق على حدة لضبط مستوى صوته في الخالط فقط. ولا يُسجَّل شيء ولا يغادر أي شيء الـ Mac الخاص بك.";
```
vorssaint, line 5: this still says nothing is recorded and only mentions the mixer, but the permission now also covers the mac's sound in the screen recordings you start. can you bring it in line with the english

README.md
```diff
@@ -65,7 +65,7 @@ First setup offers three one click bundles, Essentials, Windows, and Battery and
   <img src="docs/assets/readme/features-hub.png" width="720" alt="The Features hub in Settings, installing and uninstalling whole features">
 </p>
 
-The rest bends the same way: panel sections reorder and hide, the compact layout trades sections for tabs, settings export to a file and import on a new Mac, the app can stay light or dark apart from the Mac, and the whole app speaks more than a dozen languages.
+The rest bends the same way: panel sections reorder and hide, the compact layout trades sections for tabs, settings export to a file and import on a new Mac, the app can stay light or dark apart from the Mac, and the whole app speaks more than a dozen languages — including العربية, which mirrors the whole interface right to left.
```
vorssaint, line 68: the readme keeps this sentence generic on purpose and never names a language, so it does not go stale with each new one. can you leave this line as it was

Sources/Vorssaint/UI/MenuPanel/PanelLayout.swift
```diff
@@ -266,7 +266,7 @@ struct PanelSection<Content: View>: View {
     }
 
     private var collapseIcon: some View {
-        Image(systemName: "chevron.right")
+        Image(systemName: "chevron.forward")
             .font(.system(size: 9, weight: .semibold))
             .foregroundStyle(.tertiary)
             .rotationEffect(.degrees(collapsed ? 0 : 90))
```
vorssaint, line 272: in arabic this chevron already points left, so the 90 degree turn makes it point up while the section is open, and the same happens with every other chevron that is rotated to show an open state. can they all point down in both directions

Sources/Vorssaint/Services/Shelf/ShelfSupport.swift
```diff
@@ -80,13 +80,20 @@ enum ShelfTileLayout {
...
     static func tileFrame(index: Int,
                           columns: Int,
                           tileSize: CGSize,
                           spacing: CGFloat,
-                          inset: CGFloat) -> CGRect {
+                          inset: CGFloat,
+                          mirrored: Bool = false) -> CGRect {
         let safeColumns = max(1, columns)
-        let column = index % safeColumns
+        let column = mirrored ? safeColumns - 1 - index % safeColumns : index % safeColumns
```
vorssaint, line 96: reversing the column keeps the leftover width on the right, so with three columns in the shelf the tiles sit about 18 points from the right edge and only 4 from the left, the mirror of what a right to left grid should do. can the mirrored frame be measured from the right edge of the content instead

Resources/ar.lproj/InfoPlist.strings
Mulhamx4: Done. It now covers all four: moving files, sending items to the Trash, opening
the requested commands and controlling playback in your media apps. The old
wording also named Finder, Terminal and Homebrew, which the English has stopped
doing, so those are gone too.

Resources/ar.lproj/InfoPlist.strings
Mulhamx4: Done. It now says the audio is used both to set each app's level in the mixer
and to include the Mac's sound in the screen recordings you start, and it keeps
only the claim the English still makes — that nothing leaves your Mac — rather
than the older "nothing is recorded", which stopped being true here.

README.md
Mulhamx4: Reverted, byte for byte. Understood on the reasoning.

I left the `CONTRIBUTING` change, since the list of shipped locales there names
each one and would be wrong without Arabic — tell me if you would rather that
stayed untouched too.

Sources/Vorssaint/UI/MenuPanel/PanelLayout.swift
Mulhamx4: You are right, and it was all eight of them, not just this one.

`chevron.forward` already points the way the language reads, so turning it the
same way in both directions was one turn too many: mirrored, a quarter turn
clockwise takes it from pointing left to pointing up. The turn now follows the
glyph, negative when the interface is mirrored, and open reads as down either
way.

Rather than fix eight copies, they all go through one `disclosureRotation(open:)`
that reads `\.layoutDirection` from the environment. No header has to remember
the mirrored case, and there is one place to change if this is ever wrong again.

Sources/Vorssaint/Services/Shelf/ShelfSupport.swift
Mulhamx4: You are right — reversing the index mirrors which column a tile lands in, not
where the row begins, so the slack stayed on the wrong side.

It is measured in from the trailing edge of the content now:
`contentWidth - inset - tileWidth - column * stride`. The first tile of every
row sits one inset from the right and the leftover falls on the left, where a
reader of the language ends. That is why the parameter became
`mirroredIn contentWidth:` instead of a flag — there is no way to ask for
mirroring without supplying the width it is measured against, so this cannot be
got wrong the same way twice.

Your three-column example is now a test: with a 276pt content width the tiles
sit at 194, 106 and 18, an inset from the right and the slack on the left,
against 4, 92 and 180 unmirrored.

`sidewaysTileFrame`, the island's version of the same grid, had the same edge to
answer to and got the same treatment.

Mulhamx4:
All five addressed in `0f583e14`, with a reply on each thread.

Main had moved another 77 commits, including the `MetricsTests` split, so this
is rebased again and the new surface is translated: the port manager, general
and page settings, settings navigation, 38 new fields across nine catalogs and
29 in the base `Strings`. Three fields that left main left Arabic with them. My
two test groups moved into `ShelfFeatureTests` and `NotchAudioLevelTests` where
their subjects now live.

Two things from that catch-up worth naming:

**Alert sound names.** The file says Apple's own loctable keeps English names
for the languages it does not translate. I looked for an Arabic table on this
Mac and found none, so Arabic joins that group rather than inventing names.

**The installer prompts gained a second argument** for the destination folder,
and Arabic had been carrying one. The format-argument test caught it, which is
the test earning its keep.

**Upstream reintroduced six more fixed-direction symbols** while I was away —
that is the fourth time the pair has drifted. Four are the calendar's week and
month arrows and now mirror. The two in the notch mixer stay as they are: their
own labels say left and right, so the glyph should match the word rather than
the reading order, the same call you would make for the window-layout glyphs.
That is the fourth drift, so the CI grep I offered earlier would now have paid
for itself; say the word and I will add it.

Clean release build with no warnings, `SELFTEST OK`, `TESTS OK (59546 checks)`.

**Not re-checked on screen since this rebase.** The last visual pass was against
`a01568a3`, and everything above it is unseen in Arabic — the rotated chevrons
and the reworked shelf grid included, both of which are tested but not looked at.

Mulhamx4:
Rebased onto current main — the branch had gone into conflict while this sat.

Nothing needed from the review side; all five notes were answered in the
previous push and nothing new has come back. This is only about staying
mergeable, plus the Arabic the last 36 commits made necessary.

New since the last push: recorder export speed, the notch's full-screen and
menu-cover options, the port manager's read failure, the menu bar "allow in the
menu bar" explanation, the music block's unavailable state, and the keep-awake
end-time picker's hour and minute labels — the last one lives in a view rather
than a catalog, so the app build caught it where the test build does not look.

Two documentation points from the review, now that main has rewritten both
files. The README sentence no longer names any language at all upstream, so
there is nothing of mine left in it. `CONTRIBUTING` was rewritten too and no
longer lists locales; the one clause I kept is on the existing bullet about
adding a language, noting that a right-to-left one also sets
`AppLanguage.isRightToLeft`. Say the word if even that should go.

Both sweeps are clean against the new main: 41 `rootView` sites all carrying the
modifier, no bare `chevron.right`/`.left`/`arrow.right`/`arrow.left` outside the
notch mixer pair that names a screen direction in its own label, and no rotated
chevron bypassing `disclosureRotation(open:)`.

Clean release build with no warnings, `SELFTEST OK`, `TESTS OK (60648 checks)`.

Still not re-checked on screen: the visual pass predates the last two rebases,
so the newer surface — the rotated chevrons, the reworked shelf grid, and
everything main has added since — is tested but unseen in Arabic.

Resources/ar.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,26 @@
+/* macOS Automation permission prompt. */
+"NSAppleEventsUsageDescription" = "يستخدم Vorssaint الأتمتة لنقل الملفات، وإرسال العناصر إلى المهملات، وفتح الأوامر المطلوبة، والتحكم في التشغيل في تطبيقات الوسائط لديك.";
+
+/* macOS Screen & System Audio Recording permission prompt. */
+"NSAudioCaptureUsageDescription" = "يستخدم Vorssaint صوت كل تطبيق لضبط مستوى صوته في الخالط ولتضمين صوت الـ Mac في تسجيلات الشاشة التي تبدأها. ولا يغادر أي شيء الـ Mac الخاص بك.";
```
vorssaint: the current english audio prompt also explains that this permission drives the dynamic island music bars. the arabic prompt still mentions only the mixer and screen recordings. can you include that third use as well

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
```diff
@@ -64,7 +64,7 @@ struct BrightnessSection: View {
                 optionsExpanded.toggle()
             } label: {
                 HStack(spacing: 7) {
-                    Image(systemName: "chevron.right")
+                    Image(systemName: "chevron.forward")
```
vorssaint, line 67: the brightness options chevron still uses a fixed positive quarter-turn, and the mixer options header does the same. both therefore point up when expanded in arabic. can these two use the direction-aware disclosure rotation too

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
```diff
@@ -164,10 +164,18 @@ struct ShelfTilesView: NSViewRepresentable {
...
+        // direction has to be carried in by hand rather than inherited.
+        let mirroredWidth = L10n.shared.language.isRightToLeft ? contentWidth : nil
         let frame: (Int) -> CGRect = { index in
-            sideways
-                ? ShelfTileLayout.sidewaysTileFrame(index: index, rows: rows, tileSize: tile, spacing: spacing, inset: inset)
-                : ShelfTileLayout.tileFrame(index: index, columns: columns, tileSize: tile, spacing: spacing, inset: inset)
+            if sideways {
+                return ShelfTileLayout.sidewaysTileFrame(index: index, rows: rows, tileSize: tile,
+                                                         spacing: spacing, inset: inset,
+                                                         mirroredIn: mirroredWidth)
```
vorssaint: the earlier feedback missed this layout case. the sideways grid is mirrored around the viewport width, while its document is widened afterwards. once a row needs more columns than fit on screen, later tiles get negative x coordinates outside the scrollable document. can the sideways frames use the full document width and keep the first column reachable at the right edge

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
```diff
@@ -164,10 +164,18 @@ struct ShelfTilesView: NSViewRepresentable {
             ? ShelfTileLayout.rowCount(contentHeight: scroll.contentSize.height, tileHeight: tile.height,
                                        spacing: spacing, inset: inset)
             : max(1, Int(ceil(Double(items.count) / Double(columns))))
+        // The grid is AppKit, laid out in absolute frames, so the layout
+        // direction has to be carried in by hand rather than inherited.
+        let mirroredWidth = L10n.shared.language.isRightToLeft ? contentWidth : nil
```
vorssaint: changing the language does not invalidate the tile rebuild cache. with the same items, selection and content size, the early return below leaves an open shelf laid out in the previous direction. can the cached layout include the direction so switching into or out of arabic rebuilds the frames

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -1945,7 +1945,7 @@ final class AppSwitcher: ObservableObject {
         panel.hidesOnDeactivate = false
         panel.isReleasedWhenClosed = false
         panel.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary, .transient, .ignoresCycle]
-        panel.contentViewController = NSHostingController(rootView: SwitcherView().environmentObject(self))
+        panel.contentViewController = NSHostingController(rootView: SwitcherView().environmentObject(self).appLayoutDirection())
```
vorssaint, line 1950: mirroring the switcher reverses the displayed grid and icon rows, but the right arrow still advances the index and the left arrow decrements it. in arabic each arrow therefore selects the neighbor in the opposite screen direction. can horizontal arrow navigation follow the displayed order while keeping shortcut cycling unchanged

Sources/Vorssaint/UI/Notch/NotchMixerView.swift
```diff
@@ -537,6 +537,9 @@ private struct NotchAppFader: View {
             Button(action: togglePin) {
                 Label(isPinned ? strings.unpin : strings.pinFirst, systemImage: isPinned ? "pin.slash" : "pin")
             }
+            // Left and right here are the screen, not the reading order: the
+            // label says which way the tile moves, so the glyph matches the word
+            // in every language.
             Button { moveBack?() } label: { Label(strings.moveLeft, systemImage: "arrow.left") }
```
vorssaint, line 549: the mirrored app rail makes the previous item the neighbor on the right, so these move-left and move-right actions now move in the opposite screen direction, including the accessibility actions. the horizontal drag target also still treats the right half as after, while the insertion marker mirrors to the left. can the callbacks and drop-side calculation follow the mirrored order

Sources/Vorssaint/Services/Recorder/RecorderEditorController.swift
```diff
@@ -1406,7 +1406,7 @@ final class RecorderEditorController: NSObject, NSWindowDelegate {
 
     func show() {
         let content = RecorderEditorView(model: model, controller: self)
-        let host = NSHostingController(rootView: content)
+        let host = NSHostingController(rootView: content.appLayoutDirection())
```
vorssaint, line 1409: this also mirrors the filmstrip's thumbnail stack, but the native zoom, text, image and blur lanes still draw earlier times on the left. the same time therefore appears on opposite sides of the filmstrip and those lanes in arabic. can the timeline keep a consistent time direction across its swiftui and native parts while the editor controls mirror

Resources/ar.lproj/InfoPlist.strings
Mulhamx4: Done — it now names all three: setting each app's level in the mixer, moving the
Dynamic Island's music bars, and including the Mac's sound in the screen
recordings you start.

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
Mulhamx4: Both fixed. These two were the last holdouts — my earlier sweep looked for a
`rotationEffect` within a few lines of a `chevron`, and in these two the two
sit further apart, so the grep missed them rather than the reasoning.

They go through `disclosureRotation(open:)` now, which is every rotated chevron
in the app. The grep I used to confirm is `rotationEffect(.degrees(` with
`disclosureRotation` excluded: the only survivors are the shelf's decorative
-8° tilt and the HUD's -90°, neither of which is a disclosure.

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
Mulhamx4: Good catch — that is a real out-of-bounds, not just a cosmetic mirror.

The width was the viewport, and the document is widened to `flowWidth` further
down, so every column past the fold mirrored to a negative x and fell outside
the scrollable document entirely.

`flowWidth` and the final `documentWidth` are computed before the frames now,
and the sideways grid mirrors around `documentWidth`. The first column sits one
inset from the right edge of the document and the flow runs left from there, so
everything stays inside and reachable. The vertical grid keeps mirroring around
the content width, which is what its document is.

`document.frame` reuses the same `documentWidth` rather than recomputing it, so
the two cannot drift apart again.

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
Mulhamx4: Right — nothing else in that comparison moves when the language does, and the
frames are absolute, so an open shelf kept whatever direction it was built in.

The direction is part of the cache key now: `lastRebuiltRightToLeft` on the
coordinator, checked alongside the items, selection, expanded batches and
content size. Switching into or out of Arabic invalidates and the tiles are
laid out again.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
Mulhamx4: Fixed, and the distinction you drew is the one that matters: the arrows name a
direction on screen, the shortcut names next and previous.

Both arrow branches go through `SwitcherSupport.horizontalArrowDelta`, which
reverses the step when the interface is mirrored. Shortcut cycling still calls
`advanceSelection` with its own delta and is untouched, so Tab keeps its order
in every language. Up and down are unchanged too — rows do not mirror.

Covered in `SwitcherScrollTests`, including the case that would regress if
someone later routed the shortcut through the same helper.

Sources/Vorssaint/UI/Notch/NotchMixerView.swift
Mulhamx4: Both halves fixed.

The move actions — the menu buttons and the accessibility actions, which share
the same two callbacks — now take their offset from
`MixerReorderDirection.offset`, so "move left" moves left on screen whichever
way the rail is drawn. The labels and glyphs stay as they are, since they name
a screen direction and now agree with what actually happens.

The drop side was the more interesting half: the insertion marker is placed
with `.trailing`, which mirrors on its own, while `point.x > bounds.midX` did
not — so they pointed at opposite gaps. `dropsAfter` mirrors the side for a
sideways rail only, which is the notch; the panel mixer is vertical and keeps
the comparison it had.

Both are in `MixerFeatureTests`, including the vertical and unmirrored cases so
the fix cannot leak into them.

Sources/Vorssaint/Services/Recorder/RecorderEditorController.swift
Mulhamx4: Agreed, and I think the right answer here is that the timeline should not
mirror at all.

Time is not a reading order, and half of that row is AppKit drawing absolute
positions that will not mirror whatever the environment says — so mirroring the
SwiftUI filmstrip could only ever put the same moment on opposite sides of the
same row.

Every lane is pinned to one direction inside `timelineRow`, which is the one
place they all pass through: the filmstrip, the ruler, the audio tracks and the
native zoom, text, image and blur lanes. A moment sits at the same x across all
of them again. The row labels and everything around the timeline still follow
the language, so the editor chrome mirrors and only the time axis holds still.

Happy to flip this the other way if you would rather the timeline mirror and
the native lanes be taught to mirror with it — it is a bigger change and I did
not want to assume it.

Mulhamx4:
All seven addressed, with a reply on each thread.

Five of them turned out to be one mistake wearing different clothes: mirroring
**where** something is drawn without mirroring **what it then means**. The
layout flipped, and the index arithmetic, the drop-side comparison, the move
callbacks and the cache key all carried on as before. That is a sharper way to
put the blind spot than the one I had — it is not "AppKit does not inherit the
environment", it is "reversing a view reverses the meaning of every direction
expressed against it".

Two of them were not cosmetic at all:

- The sideways shelf grid was mirroring around the viewport while its document
  is widened afterwards, so any column past the fold landed at negative x,
  outside the scrollable document. Tiles were unreachable, not merely on the
  wrong side.
- The rebuild cache had no idea the language could change, so an open shelf
  kept the direction it was built in until something else invalidated it.

And the recorder timeline is the one that should not mirror. Its native lanes
draw absolute positions and cannot, so mirroring the SwiftUI half could only
put the same moment on opposite sides of one row. Every lane is pinned to a
single time direction now, with the labels and controls around it still
following the language. Say the word if you would rather it mirror and the
native lanes learn to mirror with it — that is a bigger change and I did not
want to assume it.

Three pure helpers came out of this and are covered by tests, because each one
encodes a decision rather than a calculation: `SwitcherSupport.horizontalArrowDelta`,
`MixerReorderDirection.offset` and `.dropsAfter`. The switcher tests include the
case that would regress if shortcut cycling were ever routed through the arrow
helper, since keeping those two apart is the whole point.

Clean release build with no warnings, `SELFTEST OK`, `TESTS OK (60653 checks)`.

**Not checked on screen.** None of this has been run in Arabic — the shelf
overflow, the arrow keys, the mixer drag and the timeline are all reasoned and
tested, not seen. The shelf overflow in particular wants a rail with more tiles
than fit, which is exactly the case I have not put in front of anyone.

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -805,9 +805,11 @@ final class AppSwitcher: ObservableObject {
                 advanceWindowInSelectedApp(by: delta)
             }
         case KeyCode.rightArrow:
-            advanceSelection(by: 1)
+            advanceSelection(by: SwitcherSupport.horizontalArrowDelta(
```
vorssaint: this and most of the notes that follow were already in the earlier push and should have come with the earlier requests. the switcher now follows the displayed order, but the same left and right arrows still step the index in other grids that mirror in arabic, the command bar category chips, the quick launcher grid and its island rail, and the island explore gallery. there each arrow moves the highlight the opposite way on screen, and at a row edge the launcher and the gallery block the arrow that still has a neighbor on screen. can those follow the displayed order as well, with up and down left as they are

Sources/Vorssaint/UI/Recorder/RecorderEditorView.swift
```diff
@@ -372,7 +372,13 @@ struct RecorderEditorView: View {
                 .foregroundStyle(.secondary)
                 .frame(width: 64, alignment: .trailing)
                 .lineLimit(1)
+            // A time axis is not a reading order, and half of this row is
+            // AppKit drawing absolute positions that mirror for nobody. Pinning
+            // every lane to one direction is what keeps a moment at the same
+            // place across the filmstrip, the ruler and the native lanes; the
+            // labels and the controls around them still follow the language.
             content()
+                .environment(\.layoutDirection, .leftToRight)
```
vorssaint: keeping this time axis left to right matches the platform guidance, which keeps video controls, timeline indicators, graphs and images in one orientation in right to left languages. the island music block still mirrors though. its position bar fills from the right, the elapsed and remaining labels swap sides, and the transport row puts next on the left while its media glyphs keep pointing the same way. the sparklines and the fan curve editor are graphs, and the recorder picture overlays and the screenshot canvas place marks in image coordinates, all inside mirrored roots, so whether they flip or come apart there depends on how swiftui treats paths, positions and gestures in a mirrored layout. can those get the same left to right pin, while the volume and brightness bars keep mirroring

Sources/Vorssaint/UI/Theme.swift
```diff
@@ -271,3 +271,51 @@ struct BrandBadge: View {
...
+/// Mirrors the interface for right-to-left languages. macOS only flips an app on
+/// its own when the *system* language is right to left, and Vorssaint picks its
+/// language from its own setting, so every window and panel root applies this.
+///
+/// This reaches SwiftUI content only. A window's own chrome — the traffic lights
+/// above all — follows `NSApp.userInterfaceLayoutDirection`, which comes from the
+/// system language and has no per-window override, so on an English Mac showing
+/// the Arabic interface they stay on the left while everything below them
+/// mirrors. On an Arabic-language Mac AppKit picks `ar.lproj` and moves them
+/// itself, so the two agree and there is nothing to reconcile.
```
vorssaint: on a mac whose system language is arabic, the new ar localization makes appkit run the app in arabic, since the language picked in vorssaint never reaches the system language list the bundle is resolved from. its menus, alerts and window buttons then turn right to left even for someone who picked english or another left to right language in vorssaint, while this modifier keeps their content left to right, and those users had a fully left to right app before. can the chrome follow the chosen language in that case, or can this note at least describe it instead of saying the two agree on an arabic mac

Sources/Vorssaint/Core/Localizations/Strings+Arabic.swift
```diff
@@ -0,0 +1,1013 @@
...
+        shelfTooltipImagePlural: "%d صورة",
+        shelfTooltipFileSingular: "%d ملف",
+        shelfTooltipFileFew: "%d ملفات",
+        shelfTooltipFilePlural: "%d ملف",
+        shelfTooltipNoteSingular: "%d ملاحظة",
+        shelfTooltipNoteFew: "%d ملاحظات",
+        shelfTooltipNotePlural: "%d ملاحظة",
+        shelfTooltipLinkSingular: "%d رابط",
+        shelfTooltipLinkFew: "%d روابط",
+        shelfTooltipLinkPlural: "%d رابط",
```
vorssaint: arabic never reaches these few strings, because uses few count form is only true for russian, so every count from two up takes the plural string, which holds the singular noun arabic uses from eleven up and reads wrong from two to ten. the russian rule would not fit either, since arabic wants the few form for 3 to 10 of every hundred and the other one for 11 to 99. can arabic get its own count rule here, or can these strings use wording that reads right for any count

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
```diff
@@ -207,11 +229,9 @@ struct ShelfTilesView: NSViewRepresentable {
             document.addSubview(view)
         }
         if sideways {
-            let tileColumns = max(1, Int(ceil(Double(items.count) / Double(rows))))
-            let flowWidth = inset * 2 + CGFloat(tileColumns) * tile.width + CGFloat(max(0, tileColumns - 1)) * spacing
             scroll.hasHorizontalScroller = flowWidth > scroll.contentSize.width + 1
             document.frame = NSRect(x: 0, y: 0,
-                                    width: max(flowWidth, scroll.contentSize.width),
+                                    width: documentWidth,
```
vorssaint: with more columns than fit, a sideways shelf in arabic still opens with its clip view at the left end of the document, so it shows the last columns first and the first column waits past the right edge until the user scrolls. can a newly built sideways shelf start scrolled to its right end for a right to left interface

Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
```diff
@@ -899,7 +899,7 @@ final class RadialMenuService: ObservableObject {
         panel.hasShadow = false
         panel.acceptsMouseMovedEvents = true
         panel.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary, .transient, .ignoresCycle]
-        panel.contentViewController = NSHostingController(rootView: RadialMenuView())
+        panel.contentViewController = NSHostingController(rootView: RadialMenuView().appLayoutDirection())
```
vorssaint: the wheel under this root places its chips with offsets and its highlight with a custom wedge shape, while the slice is picked from the pointer's real direction on screen. if swiftui mirrors those in arabic, as the timeline pin assumes for the filmstrip, the highlight and the chip under it land opposite the pointer. does the highlight still sit under the pointer in arabic? keeping the wheel left to right would avoid the question, since its slices are screen directions

vorssaint:
main has moved on, and the rebase needs arabic for what landed there, the new agents catalog, the island editor summaries and related fields, two music extras fields, the quarter and third names in window layout, and five base strings for the disk display picker and the scroll coast. the disk picker test on main also expects arabic to translate all four disk strings, keep the three options distinct and word the label apart from the usage display label. the alert sound names can come along too, since the system table they follow does translate them for arabic, unlike japanese, korean and chinese. can the next push rebase and carry them

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
Mulhamx4: Done, and you are right that it should have come with the earlier round.

The category chips, the launcher grid, the island rail it shares and the explore gallery all read the key before the walk now, through one helper beside the walk itself: a side arrow maps to the opposite step once the grid is mirrored, and up and down come back untouched.

Reading it there rather than inside the walk is also what puts the row edge right. The walk stops at index 0 and at the last column of a row, and mirrored those are the right and left edges on screen, so the arrow that still has a neighbour in front of the reader is no longer the one that blocks.

Sources/Vorssaint/UI/Recorder/RecorderEditorView.swift
Mulhamx4: All of them, and I measured the question underneath rather than assuming it.

In a mirrored subtree SwiftUI flips more than the order of rows: `offset`, `position` and the path a `Shape` draws all mirror. A view offset by +40 in a 200-wide frame lands at midX 140 left to right and at 60 right to left, and a `Path` filling `rect.minX` through `minX + 20` renders against the right edge.

So every surface you listed does come apart, and they are pinned now: the island's playback position with its elapsed and remaining labels, the transport row, the sparklines, the fan curve editor, the recorder's picture and the screenshot canvas. The fan curve was the worst of them, since the drag compared a location against a drawing that had moved, so a handle followed the pointer the wrong way.

The volume and brightness bars still mirror. A magnitude does run along the reading order, so those are the ones that should.

Sources/Vorssaint/UI/Theme.swift
Mulhamx4: You are right, and the note was wrong. The bundle resolves `ar.lproj` on an Arabic Mac now, so AppKit runs the whole app right to left there for everyone, including someone who picked English here and had a left-to-right app before Arabic existed.

The note says that instead of claiming the two agree.

I did not make the chrome follow the chosen language. The only way I know to do it is to hand AppKit that language at launch, in the app's own `AppleLanguages`, and that decides the question for all fourteen rather than this one: an English Mac set to German here would start getting German menus and alerts too, from the next launch onwards. That reads like your call, not mine. Say the word and I will add it.

Sources/Vorssaint/Core/Localizations/Strings+Arabic.swift
Mulhamx4: Arabic has its own rule now. `usesFewCountForm` became `countRule`, with three cases: one and many, Russian's, and Arabic's.

Arabic asks for the middle form from three through ten of every hundred, so 3, 7, 103 and 110 take the plural, and 11 through 99 come back to the singular noun after the numeral. Two goes with eleven rather than with three, since the dual is a word of its own and is not what follows a numeral.

Nothing else moved. Every other language still reaches one and many through the same call, Russian's rule is unchanged, and the three slots stayed as they are, so no catalog gained a field.

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
Mulhamx4: Yes. A sideways shelf whose flow is wider than its clip view now opens at the document's right end in a right-to-left interface, so it starts on the first column instead of the last.

Only when the end itself moved, which is when the direction flips or the viewport resizes. An ordinary rebuild, a tile added or renamed, leaves the shelf where the reader left it, and a newly revealed tile still wins over both.

Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
Mulhamx4: It does not, and the measurement says so rather than my guess: in a mirrored subtree SwiftUI flips `offset`, `position` and a `Shape`'s path, so the chips and the highlight wedge both move while the slice is still picked from the pointer's real direction on screen.

The wheel is left to right now. Its slices are directions on screen, as you say, and the labels inside it still read in the chosen language.

Mulhamx4:
Rebased and carried, and the six notes are in.

**Arabic for what landed on main.** The AI agents catalog, the island editor's section summaries and the strings around them, the two music extras fields, the thirds and quarters that joined window layout, and the five base strings for the disk display picker and the scroll coast. The disk picker's contract asks for more than a translation, and عرض القرص against عرض الاستخدام is what keeps its label apart from the usage display one while the three options stay distinct.

The alert sound names came along. You were right that Apple's own table translates them for Arabic: I read `AlertSounds.loctable` on this Mac, and `ar` has وسطي, نسيم, حصى and the rest, while `ja`, `ko` and `zh` keep the English names, which is why Arabic had been grouped with them.

**The five notes were one question asked of five surfaces**, and I measured the answer instead of assuming it. In a mirrored subtree SwiftUI flips `offset`, `position` and the path a `Shape` draws, not only the order of rows: a view offset by +40 in a 200-wide frame lands at midX 140 left to right and 60 right to left, and a `Path` filling `rect.minX` through `minX + 20` renders against the right edge.

So the surfaces whose horizontal axis is a reading order now follow the displayed order: the category chips, the launcher grid and its island rail, and the explore gallery. Reading the key before the walk rather than inside it also puts the row edges back where the eye sees them.

And the surfaces whose axis is not a reading order keep one orientation: the island's playback position and transport row, the sparklines, the fan curve editor, the recorder's picture, the screenshot canvas, and the radial wheel, whose highlight would otherwise have landed opposite the pointer. The volume and brightness bars still mirror, since a magnitude does run along the reading order.

**Arabic counts by its own rule.** Three through ten of every hundred take the plural, eleven up returns to the singular noun after the numeral, and two goes with eleven rather than with three. No other language changed and no catalog gained a field.

**The sideways shelf** opens at its right end in a right-to-left interface when its flow is wider than the viewport, but only when that end moved.

**The layout note no longer claims the chrome and the content agree on an Arabic Mac.** They do not, and the note says what actually happens. Making the chrome follow the chosen language means handing AppKit that language at launch, which decides it for all fourteen, so I left that to you.

Two symbols had drifted since the last sweep and are fixed: a Homebrew disclosure chevron and the mouse Back and Forward icon.

Clean release build with no warnings, `SELFTEST OK`, and `TESTS OK (62492 checks)` with new coverage for the mirrored arrow walk and for Arabic's count rule at 0, 1, 2, 3, 10, 11, 25, 99, 100, 101, 103, 110 and 111.

**Not checked visually.** The radial wheel and the fan curve editor in particular are worth a look, since both are gesture surfaces and neither has a test that can see where a mark lands.

Sources/Vorssaint/Core/Localizations/Strings+Arabic.swift
```diff
@@ -0,0 +1,1022 @@
...
+        shelfActionOpen: "فتح",
+        shelfActionOpenWith: "فتح باستخدام",
+        shelfActionShare: "مشاركة",
```
vorssaint: main moved again and the arabic initializers now miss fields it added, so the merged tree stops compiling. uninstallerconfirmationexpired, shelfactionpin and shelfactionunpin belong here, exportfailed in the backup strings, hidemenubaricon and hidemenubariconhint in the island editor, keyboardtaprefused in the super key strings, and trackpadtaplabel and trackpadtapconflict in the radial menu. the english also reworded two existing fields, shelfremoveafterdropcaption now says pinned items always stay and mousetriggerrequirement now points trackpad users to the four finger tap. can the next rebase carry all of them

Sources/Vorssaint/UI/Shelf/ShelfTilesView.swift
```diff
@@ -192,10 +213,15 @@ struct ShelfTilesView: NSViewRepresentable {
             }
             return
         }
+        // Where the flow ends moves when the direction flips or the viewport
+        // resizes, and nothing else here moves it.
+        let endMoved = coordinator.map { $0.lastRebuiltRightToLeft != rightToLeft
```
vorssaint: a merge onto a tile rebuilds the grid from concludedragoperation without a coordinator, and this fallback reads that as a moved end. after such a drop the island strip jumps back to its first column in every language, even when the tile that took the drop sits further along and the user had scrolled to it. a left to right strip also goes back to its start on the first rebuild after its viewport changes size, which it never did before. a rebuild without a coordinator should leave the scroll alone, and a left to right strip should only move when the direction flips

Sources/Vorssaint/Services/Notch/NotchWindowHost.swift
```diff
@@ -566,7 +566,7 @@ private final class NotchQuickAccessContainer: NSView {
         self.canvas = canvas
         let motion = NotchQuickAccessMotion()
         self.motion = motion
-        quickView = NSHostingView(rootView: content(motion))
+        quickView = NSHostingView(rootView: AnyView(content(motion).appLayoutDirection()))
```
vorssaint: this root was already mirrored in the earlier push and belonged with the last round. the quick access bubbles are placed with position around the notch and joined to it by a drop shape, while notchquickaccesscontainer hit tests clicks and tracks hover against the same unmirrored placements. with position and shape paths flipped in a right to left root, as measured for the wheel, the side bubbles draw on the opposite side of the notch in arabic, and when the two sides hold different numbers of bubbles the extra ones land where neither a click nor hover reaches them. the layout editor in settings places its bubbles with position too, so a button set to the left shows on the right there. these are screen sides like the wheel's slices, so can this root and the editor's canvas keep one orientation, with the island preview inside the editor still following the language

Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
```diff
@@ -899,7 +899,14 @@ final class RadialMenuService: ObservableObject {
...
         panel.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary, .transient, .ignoresCycle]
-        panel.contentViewController = NSHostingController(rootView: RadialMenuView())
+        // The wheel is picked by the pointer's real direction from the
+        // centre, and it places its chips and its highlight wedge with
+        // offsets and a path — all of which a mirrored root flips, measured
+        // on this Mac. Its slices are directions on screen, not a reading
+        // order, so the wheel keeps one orientation and the highlight stays
+        // under the pointer. The labels inside it still read in the chosen
+        // language.
+        panel.contentViewController = NSHostingController(rootView: RadialMenuView().unmirroredLayout())
```
vorssaint: pinning the live wheel leaves radialmenuvisualcanvas in settings behind. it draws this same wheel with the same offsets and wedge shape and picks the drop slot from the pointer's direction, but it still sits under the mirrored settings root, which flips those offsets and that shape. someone using arabic would now put an action on one side in settings and find it on the other side when the wheel opens. can the canvas area take the same pin, with its header still following the language

Sources/Vorssaint/UI/MenuPanel/Sparkline.swift
```diff
@@ -48,6 +48,10 @@ struct Sparkline: View {
                 }
             }
         }
+        // Oldest sample at the left edge, newest at the right, in every
+        // language: this axis is time, and a mirrored root would flip the
+        // path along with everything else drawn in it.
+        .unmirroredLayout()
```
vorssaint: the ai agents section that arrived with this rebase has two more time axes, the trend bars with their axis labels and the activity heatmap, and both still mirror, so in arabic the latest bar and the latest week sit on the left while these sparklines keep the oldest sample there. following the rule contributing now spells out, can they take the same pin, leaving the figures beside the heatmap in the reading order

Sources/Vorssaint/UI/Theme.swift
```diff
@@ -271,3 +271,73 @@ struct BrandBadge: View {
...
+    /// Mirroring a view mirrors everything drawn inside it, not only the order
+    /// of its rows: measured on this Mac, SwiftUI flips `offset`, `position`
+    /// and the path a `Shape` draws. That is what a reading order wants, and
+    /// it is wrong wherever the horizontal axis means something other than
+    /// reading — time, an image's own coordinates, or a direction on screen
+    /// that a pointer is compared against. Apple's own guidance keeps video
+    /// controls, timeline indicators, graphs and images in one orientation in
+    /// right-to-left languages for the same reason. The labels and controls
+    /// around a pinned surface still follow the language; a level or progress
+    /// bar is not pinned, since its axis is the reading order.
```
vorssaint: one of these bars was already in the earlier push and got missed. with liquid glass on under macos 26 the panel mixer uses liquidglassmixerslider, which places its fill with a leading alignment and its knob with an offset, while updatevalue turns location.x into a value counted from the left edge. if the drag location stays unmirrored, as found for the fan curve, the knob would move away from the pointer in arabic and a click near the right end would send it to the far left. does that knob follow the pointer in arabic, and if it does not, can the drag count from the trailing edge so the bar keeps mirroring

Sources/Vorssaint/Services/Notch/NotchWindowHost.swift
```diff
@@ -656,11 +656,11 @@ private final class NotchCanvas: NSView {
 
     init(content: AnyView, background: (NotchBackdropPresentation) -> AnyView, size: CGSize) {
         contentSize = size
-        backdrop = NotchHostingView(rootView: background(backdropPresentation))
+        backdrop = NotchHostingView(rootView: AnyView(background(backdropPresentation).appLayoutDirection()))
         backdrop.sizingOptions = []
         backdrop.wantsLayer = true
         backdrop.autoresizingMask = []
-        host = NotchHostingView(rootView: content)
+        host = NotchHostingView(rootView: AnyView(content.appLayoutDirection()))
```
vorssaint: dragging media onto the island offers two drop targets side by side, files and then optimize media, and this root mirrors that row, so in arabic files sits on the right. updatefiledrop still takes the right half as the media target through notchfiletoolssupport.mediadroparea, so the highlight lands on the tile opposite the pointer and a file dropped on files opens the media tools instead of joining the shelf. can that area follow the layout direction

PR #1154, opened 2026-08-30
Add Hebrew localization

Y-PLONI:
## Summary

Adds Hebrew as an application language, including the complete UI catalog, feature-specific strings, macOS permission prompts, and bundle metadata. It also updates localization-contract tests so Hebrew is checked alongside every supported language.

The app is AppKit-hosted, so every `NSHostingController`/`NSHostingView` starts a fresh SwiftUI environment and nothing mirrors a Hebrew layout on its own. Each hosting root applies `localizedLayoutDirection()`, and a `MetricsTests` assertion walks every `.swift` under `Sources/` so a hosting root added later cannot quietly skip it.

## Known limitation

`CFBundleLocalizations` gaining `he` makes AppKit itself right-to-left on a Hebrew-system Mac, because `NSApp.userInterfaceLayoutDirection` follows the bundle while the in-app language is a separate preference. The two can therefore disagree: a Hebrew system with English picked in-app pins content left-to-right inside mirrored chrome — traffic lights, scrollers, menu bar. The converse mismatch already exists in the other direction, on an English system with Hebrew picked. Aligning chrome with the in-app pick would mean making the language a relaunch-scoped setting, which is a change to the language model rather than a Hebrew fix, so it is left alone here; some form of this is unavoidable once any RTL language ships.

## What mirroring does not reach

The layout modifier covers what SwiftUI lays out. Three kinds of thing need more, and this branch handles each:

- **Fixed-direction symbols.** `chevron.right`, `chevron.left` and `arrow.right` never flip. 29 move to `.forward` / `.backward`, the same set #1149 moves for Arabic, plus `sidebar.trailing`, `arrowshape.turn.up.backward` and the log-out `rectangle.portrait.and.arrow.forward`. Disclosure rotations are unchanged: SwiftUI mirrors `rotationEffect` with the layout, so the mirrored chevron still turns down.
- **AppKit.** The shelf grid (absolute frames) is reflected across the document width, the island's level bars (a hand-drawn `NSSliderCell` fill) fill from the trailing edge, and the shelf tile's `NSImage` chevron is picked by language. These follow #1149's fixes.
- **Drags on mirrored drawing.** A gesture reports the physical pointer while the view under it is drawn mirrored. Things that stand for a physical place stay left to right: the radial wheel (it picks a slice from the physical pointer), the recorder's picture and timeline lanes, the island's floating buttons, the caption and watermark anchor grids, and the fan curve graph. The Liquid Glass mixer slider mirrors instead, measuring its drag from the trailing edge.

## Verification

- Rebased onto current `main` (303 commits since the last base). Conflicts in `CommandBarStrings`, `FeatureStrings` (Mixer), `MetricsTests`, `AppDelegate`, `Localization`, `ScreenshotQuickPreviewController` and `ScreenshotSelectionController`, each resolved to keep both sides; main's 448 new strings are translated.
- `./build.sh --test`: **TESTS OK (57686 checks)**, no failures.
- `plutil -lint Resources/Info.plist Resources/he.lproj/InfoPlist.strings` and `git diff --check`: clean.
- Direction behaviour measured on hosted SwiftUI views, both directions: `rotationEffect` mirrors with the layout, and a `DragGesture` location does not.
- Negative control on the hosting-root assertion: removing the modifier from one root reports `ScratchpadService.swift hosts a SwiftUI root that never mirrors layout direction`.
- Not yet run as the full app on screen in Hebrew.

## Anything else

No related issue.


PathGao:
Solid mechanical work: 43 catalogs, every `AppLanguage` switch, and `he.lproj` carries the same key set as the other twelve. The compiler and `MetricsTests` already cover completeness, so I looked at what they can't see.

**RTL.** Hebrew is the first right-to-left language here, and nothing in `Sources/` sets `layoutDirection` or `semanticContentAttribute` — I grepped. `AppLanguage` is decoupled from the system locale, so on an English Mac someone who picks עברית gets Hebrew text in an unmirrored layout: labels, chevrons, sliders all on the wrong side. `CFBundleLocalizations` only flips when the *system* language is Hebrew. Laziest fix, at each window root:

```swift
.environment(\.layoutDirection, l10n.language == .he ? .rightToLeft : .leftToRight)
```

I haven't run it, so I can't say how bad it looks — but the decoupling is certain.

**Undeclared change**, `AppLanguage.systemDefault`: `preferred.hasPrefix` → `p.hasPrefix`. Only line in the PR that touches existing behavior, and the body doesn't mention it. No-op in practice (subtags come back lowercase), but it belongs in the description.

`CommandBarSystemSettingsSupport.resourceFolder(for:)` is past the diff truncation — I didn't see its `.he` value. It needs `"he"`, else pane names silently fall back to English.

**Impact:** Hebrew UI renders LTR for anyone whose system language isn't Hebrew. Other twelve languages unaffected.
**Merge:** needs the layout-direction decision first.

Y-PLONI:
Thanks for the review.

Regarding `CommandBarSystemSettingsSupport.resourceFolder(for:)`: that one was already handled, the `.he: "he"` case was already there — you just couldn't see it past the diff truncation.

The rest of your feedback was spot on, and I've addressed it:
- Added a `localizedLayoutDirection()` view modifier that reads the app's language reactively and applies `.environment(\.layoutDirection, ...)` accordingly, and wired it into all 31 `NSHostingController`/`NSHostingView` roots across the app (menu panel, settings, onboarding, update flows, shelf, switcher, command bar, radial menu, clipboard, every QuickTools panel, dock preview, permission overlay). This fixes the RTL/LTR mismatch for anyone who picks Hebrew from within the app while their system language stays non-Hebrew.
- Left the `preferred.hasPrefix` → `p.hasPrefix` change as-is (confirmed no-op, as you noted), just flagging here for the record since it wasn't called out in the PR description.

And thank you for the app itself! I've been waiting a long time for something like this, and it just got covered by a major tech site (in Hebrew)!

English isn't my native language, so I wrote this comment with AI assistance.

Thanks again!

Y-PLONI:
I didn't check what he did - but he also needs RTL...
https://github.com/vorssaintapp/vorssaint-utils/pull/1149

PathGao:
Counted the direct roots on `main`: 31 `(rootView:)` construction sites, so your wiring hits every one.

Three more roots come from subclasses, so they fall outside that count:

- `OverlayScrollView` → `HeightReportingHostingView` in `MenuPanelView`. That one is the panel body. A nested `NSHostingView` starts a fresh SwiftUI environment, so the popover root's `\.layoutDirection` never reaches it — mirrored header sitting over LTR content. Cheapest fix is at the two call sites, which keeps the generic inferred:

```swift
OverlayScrollView(measuredHeight: $navigableContentHeight) {
    VStack(alignment: .leading, spacing: 12) { ... }
        .localizedLayoutDirection()
}
```

- `OverlayHostingView` in `CleaningModeManager` and `PassThroughHostingView` in `ScreenshotSelectionController` — same shape, smaller surfaces.

Read only, not run.

Nothing stops the next root from forgetting the call. `MetricsTests` already asserts against source text, so checking that every `(rootView:` line carries `.localizedLayoutDirection()` is in-idiom here and kills the class. Same rule answers #1149: it needs the modifier only if it adds a hosting root.

**Impact:** Hebrew menu panel renders mixed — mirrored chrome, LTR content.
**Merge:** needs the three remaining roots first.

Y-PLONI:
Good catch — all three fixed.

You were right: `HeightReportingHostingView` (behind `OverlayScrollView` in `MenuPanelView`), `OverlayHostingView` (`CleaningOverlayView`), and `PassThroughHostingView` (`CaptureGuideView`) each start their own SwiftUI environment, so `localizedLayoutDirection()` at the outer root didn't reach them. Fixed by applying it directly at each of the three nested roots — at the two `OverlayScrollView` call sites in `MenuPanelView.swift` since that one stays generic, and inside the `body` of `CleaningOverlayView`/`CaptureGuideView` themselves since those hosting subclasses are concrete/pre-declared and can't take the modifier at the call site without breaking their type.

Then had two independent QA passes go over the whole thing from scratch (not just re-checking the three you flagged): every direct and nested `NSHostingController`/`NSHostingView` root, every `.rootView =` reassignment, popovers/sheets, alerts, menu items — nothing else turned up. Build's green, existing test suite unaffected.

Thanks again for the thorough review!

PathGao:
Worth confirming your sweep is backed by more than inspection: both `.rootView =` reassignment sites are safe by construction. `NSHostingController<T>.rootView` and `HeightReportingHostingView<Content>` are typed, so once the construction site is wrapped the assignment has to match or it doesn't compile — `MenuPanelView` and the `guideHost` in `ScreenshotSelectionController` can't silently drop the modifier.

What the compiler doesn't cover is a *new* hosting root added later. That's still the source-text assertion in `MetricsTests`, and you didn't mention it in the two passes. Existing tests there read one named file at a time, so this one needs to enumerate `Sources/` — slightly past the idiom, but it retires the whole class instead of the three instances.

GitHub reports the branch as conflicting with `main`, so it needs a rebase regardless.

The three nested roots are your report, not my verification — the head diff is past truncation for me.

**Impact:** none new.
**Merge:** needs a rebase first.

Y-PLONI:
Both done.

**Rebased** onto current `main`. One conflict, in `CaptureGuideView` — `main` added `loupeEnabled` to `UnifiedCaptureGuideContent` while this branch wrapped the same `if/else` in a `Group`; the resolution keeps both. The rebase also brought in strings the branch had never seen, so Hebrew is now filled in for quit/close protection, the magnifier toggle, the permission guide's stale-entry hints, clipboard multi-select delete, the Spaces drag gesture, the extra-click filter, focus-follows-mouse exceptions, and the cleaning and pointer-acceleration settings. While there, three Hebrew formats that took `%d` where the argument is a string are fixed, along with a category title and a Scratchpad name the suite was already flagging on this branch.

**The source-text assertion** is in `MetricsTests`. You were right that it belongs there and that it has to enumerate `Sources/` rather than name a file — the gap reopens in a file the test cannot know the name of yet. It walks every `.swift` under `Sources/`, reads each `(rootView:` as a balanced expression rather than a line (several roots span five or six), and accepts the modifier in exactly four places: on the construction expression, on the local it hosts, inside the hosted view's own `body` (where a pre-declared hosting subclass needs it), or — for `HeightReportingHostingView`, which is generic and forwards what it is handed — at every `OverlayScrollView` call site that wraps it. `init(rootView:)` and its `super` call are skipped as the subclass plumbing they are. Anything else fails with the file and the expression.

Checked it fails for the right reason: dropping the modifier from one root reports `ScratchpadService.swift hosts a SwiftUI root that never mirrors layout direction`. Suite is otherwise green.

Thanks — this is a much better state than where the PR started.

PathGao:
The `MetricsTests` walk is the right shape, and `(rootView:` holds up as the enumeration token — I checked it against `main`: 39 `rootView` matches across 24 files, of which 34 are construction expressions (both hosting classes and all three subclasses inherit the label), 3 are `.rootView =`, and the last two are the `init(rootView:)`/`super` pair in `HeightReportingHostingView` — exactly what your skip rule drops. Nothing can be added later that escapes it.

Two notes, neither blocking.

`.rootView =` has three sites, not two: `BrightnessOSD.show` reassigns onto a `static var host: NSHostingController<BrightnessOSDView>?`. Your conclusion still holds — that generic is spelled out twice, so the compiler pins it — but the count in your sweep was short one.

`CFBundleLocalizations` gaining `he` makes AppKit itself RTL on a Hebrew-system Mac; `NSApp.userInterfaceLayoutDirection` follows the bundle, not `AppLanguage`. So the decoupling now cuts the other way: Hebrew system, English picked in-app, and the modifier pins content LTR inside mirrored chrome — traffic lights, scrollers, menu bar. Much narrower than the bug it fixed, and unavoidable once any RTL language ships. Worth a line in the body rather than a code change. Read only, not run.

The test itself is past diff truncation for me — that part is your report.

**Impact:** none new.
**Merge:** ship it.

Y-PLONI:
Both noted, thanks.

You are right about the count — `BrightnessOSD.show` is a third `.rootView =` site, on `static var host: NSHostingController<BrightnessOSDView>?`. I missed it; the generic is spelled out at the declaration and again at construction, so it is pinned the same way the other two are, but the sweep I reported was short one.

The chrome caveat is now a paragraph in the body rather than a code change, and I widened it slightly: the mismatch runs in both directions. An English system with Hebrew picked in-app already gets mirrored content inside left-to-right chrome, which is the more common case for this PR. Aligning either one means making the language relaunch-scoped, which is a change to the language model rather than a Hebrew fix, so I left it out of this branch.

While editing the body I also refreshed the Verification section, which still said the test run had not returned.

PathGao:
Two things worth knowing before you spend more time on this branch.

**Arrows and chevrons.** #1149 (Arabic) ran its build on screen and found that flipping the layout direction is not enough: `chevron.right`/`chevron.left` and the Super Key diagram arrow are fixed glyphs, so in a mirrored layout they point the wrong way. It fixed that in 64192d9d and 7e4a837e, 23 glyphs switched to the `.forward`/`.backward` variants that follow direction. This branch has none of those changes and, per its Verification, has not been run on screen, so the same thing will show up here.

**Two right-to-left scaffolds.** #1149 and this PR each add a layout-direction modifier to the same ~30 UI files under different names (`appLayoutDirection` there, `localizedLayoutDirection` here). Only one can land, and the other language then rebuilds its strings on top of it. Which one is the maintainer's call, and he has not made it; a rebase before that answer would be wasted work. Both branches also miss the four string sets main gained on 2026-09-01, so neither compiles against current main as it stands.

Sources/Vorssaint/Core/FeatureStrings.swift
```diff
@@ -1152,6 +1167,63 @@ struct ClipboardFeatureStrings {
...
+        search: "חפש טקסט שהועתק",
+        copy: "העתק",
+        copied: "הועתק",
+        delete: "מחק פריט",
+        selectMultiple: "הוסף למחסנית",
+        unselectMultiple: "הסר מהמחסנית",
+        selectShortcutAction: "בחר",
+        pasteSelectedFormat: "הדבק %d",
+        copySelectedFormat: "העתק %d",
+        deleteSelectedFormat: "מחק %d",
```
vorssaint, line 1203: this argument has to come last, the way it does in the english block, or the file does not build

Sources/Vorssaint/Core/ScreenshotStrings.swift
```diff
@@ -1998,4 +1999,145 @@ extension ScreenshotFeatureStrings {
...
+        backdropSavePreset: "שמור רקע",
+        backdropDeletePreset: "הסר",
+        backdropCustomLabel: "מותאם אישית",
+        openEditorToggle: "פתח את העורך מיד לאחר הצילום",
+        openEditorCaption: "הצילום מדלג על התצוגה המקדימה הצפה ונפתח מוכן להערות.",
+        autoCopyToggle: "העתק אוטומטית ללוח",
+        autoCopyCaption: "כל צילום עובר ללוח מיד כשהוא נלקח, מוכן להדבקה. שמירה כקובץ נשארת בחירה נפרדת.",
+        hintLoupe: "Z מפעיל/מכבה את הזכוכית המגדלת",
+        lastRegionToggle: "הצג את קווי המתאר של הצילום האחרון",
+        loupeStartsOnToggle: "להתחיל את הבחירה עם זכוכית מגדלת פעילה",
```
vorssaint, line 2091: this line belongs at the very end of the list, like in the other languages, otherwise the build fails

Sources/Vorssaint/Core/QuitProtectionStrings.swift
```diff
@@ -475,4 +476,33 @@ extension QuitProtectionStrings {
...
+        exceptions: "חריגים",
+        noExceptions: "לא נבחרו יישומים",
+        addApp: "הוספת יישום…",
+        feedback: "להציג חיווי חזותי",
+        accessibilityCaption: "ההגנה משתמשת בנגישות כדי לעקוב גלובלית אחר ⌘Q ו-⌘W בלבד.",
+        holdHUDFormat: "לחצו לחיצה ממושכת על %@ ליציאה/סגירה",
+        doubleHUDFormat: "לחצו %@ שוב ליציאה/סגירה",
+        extraHUDFormat: "השתמשו ב-%@ ליציאה/סגירה",
+        cancelHint: "Esc מבטל",
+        releaseHint: "שחררו לאישור"
```
vorssaint, line 506: the names of the shift, option and control keys that the other languages end with are missing here, so this block does not build

Sources/Vorssaint/Core/QuitProtectionStrings.swift
```diff
@@ -475,4 +476,33 @@ extension QuitProtectionStrings {
...
+        selectedOnly: "יישומים נבחרים בלבד",
+        allExceptSelected: "כל היישומים פרט לנבחרים",
+        exceptions: "חריגים",
+        noExceptions: "לא נבחרו יישומים",
+        addApp: "הוספת יישום…",
+        feedback: "להציג חיווי חזותי",
+        accessibilityCaption: "ההגנה משתמשת בנגישות כדי לעקוב גלובלית אחר ⌘Q ו-⌘W בלבד.",
+        holdHUDFormat: "לחצו לחיצה ממושכת על %@ ליציאה/סגירה",
+        doubleHUDFormat: "לחצו %@ שוב ליציאה/סגירה",
+        extraHUDFormat: "השתמשו ב-%@ ליציאה/סגירה",
```
vorssaint, line 504: these lines and the release hint below talk to the user in the plural, while the rest of the app speaks in the singular. can you keep one form

Sources/Vorssaint/Core/Localizations/Strings+Hebrew.swift
```diff
@@ -0,0 +1,988 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+
+extension Strings {
+    static let he = Strings(
```
vorssaint, line 7: since this branch started the app gained the notch island with its calendar, files, notifications and music screens, the shelf delivery notes, a calendar permission prompt and a few dozen smaller strings, and none of that speaks hebrew yet. can you fill those in so hebrew covers everything the other languages do

Tests/MetricsTests.swift
```diff
@@ -21185,6 +21188,103 @@ struct MetricsTests {
...
+                        wrapped += 1
+                        if balancedSlice(text, from: site.upperBound, open: "{", close: "}")
+                            .contains(layoutModifier) { mirrored += 1 }
+                    }
+                    expect(wrapped > 0 && wrapped == mirrored,
+                           "every OverlayScrollView call site mirrors the content it hosts "
+                           + "(\(mirrored) of \(wrapped))")
+                    continue
+                }
+                expect(false, "\(path) hosts a SwiftUI root that never mirrors layout direction: "
```
vorssaint, line 21282: the island window and the disk image installer now host views of their own, and this sweep will point at them. they need the same mirroring as the rest

Sources/Vorssaint/Core/Localization.swift
```diff
@@ -56,13 +61,32 @@ enum AppLanguage: String, CaseIterable, Identifiable {
...
+/// Mirrors the chosen `AppLanguage` into SwiftUI's layout direction. This app
+/// is AppKit-hosted (`NSHostingController`/`NSHostingView`), so nothing gives
+/// a Hebrew UI a mirrored layout on its own: `CFBundleLocalizations` only
+/// follows the *system* language, and `AppLanguage` is a user override that's
+/// decoupled from it. Every hosting root applies this explicitly instead.
+private struct LocalizedLayoutDirection: ViewModifier {
+    @ObservedObject private var l10n = L10n.shared
+
+    func body(content: Content) -> some View {
+        content.environment(\.layoutDirection, l10n.language.isRTL ? .rightToLeft : .leftToRight)
```
vorssaint, line 80: flipping the layout does not turn the chevrons and arrows around, so in hebrew the back button in the panel and the disclosure chevrons still point the wrong way. can you switch those to the forward and backward symbols so they follow the reading direction

Resources/he.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,24 @@
+/* macOS Automation permission prompt. */
+"NSAppleEventsUsageDescription" = "Vorssaint משתמש באוטומציה כדי להעביר קבצים ב-Finder, לשלוח פריטים לאשפה ולפתוח פקודות בטרמינל כשאתה מבקש פעולות Homebrew.";
+
+/* macOS Screen & System Audio Recording permission prompt. */
+"NSAudioCaptureUsageDescription" = "Vorssaint ניגש רק לאודיו של אפליקציות בודדות כדי להתאים את עוצמת הקול שלהן במיקסר. שום דבר לא מוקלט ושום דבר לא עוזב את ה-Mac שלך.";
```
vorssaint: this still says nothing is recorded, but the permission also covers the mac's sound in screen recordings and the music bars. can you update it to match those uses

Resources/he.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,24 @@
+/* macOS Automation permission prompt. */
+"NSAppleEventsUsageDescription" = "Vorssaint משתמש באוטומציה כדי להעביר קבצים ב-Finder, לשלוח פריטים לאשפה ולפתוח פקודות בטרמינל כשאתה מבקש פעולות Homebrew.";
```
vorssaint: this still describes file actions and terminal commands, but the same permission now controls music playback too. can you include that use

Sources/Vorssaint/Core/Localizations/Strings+Hebrew.swift
```diff
@@ -0,0 +1,1039 @@
...
```
vorssaint, line 7: please bring the hebrew catalogs up to current main, including recorder export, mixer actions, compact-calendar labels, island visibility, network addresses, menu-bar recovery and port errors. these added required fields and the new language switch otherwise prevent the combined source from compiling; quit protection also replaced the three shared hud formats with six quit/close-specific ones. please update the changed music-protection, port-manager and empty-folder descriptions too

Sources/Vorssaint/UI/Notch/NotchCalendarMonthView.swift
```diff
@@ -27,8 +27,8 @@ struct NotchCalendarMonthView: View {
                         .font(.system(size: 11)).foregroundStyle(.white.opacity(0.45))
                 }
                 .frame(maxWidth: .infinity, alignment: .leading)
-                NotchIconButton(symbol: "chevron.left", title: text.previousMonth) { move(-1) }
-                NotchIconButton(symbol: "chevron.right", title: text.nextMonth) { move(1) }
+                NotchIconButton(symbol: "chevron.backward", title: text.previousMonth) { move(-1) }
```
vorssaint, line 30: please apply the same directional symbols to the new compact week strip and month grid when updating this branch. current main added those below this view with fixed left/right chevrons, so smaller islands still show previous and next arrows pointing against the hebrew reading direction

Sources/Vorssaint/Services/Notch/NotchWindowHost.swift
```diff
@@ -438,7 +438,7 @@ private final class NotchCanvas: NSView {
 
     init(content: AnyView, size: CGSize) {
         contentSize = size
-        host = NotchHostingView(rootView: content)
+        host = NotchHostingView(rootView: AnyView(content.localizedLayoutDirection()))
```
vorssaint, line 441: please adapt the new horizontal island mixer to this mirrored layout. its rail reverses the visible app order, but move left still decrements the array index and the native drop target still treats the physical right half as after. in hebrew the action moves an app right, and command-drag inserts on the opposite side of the hovered column. make the move actions and drop-side calculation follow the layout direction, preserving the vertical panel list

Issue #1164, opened 2026-08-30
Add a safe read/write speed test for selected disks


Mahmutakin99:
## Problem or motivation

The Disk panel shows current I/O activity, but it cannot measure the sequential throughput of the internal disk or a selected external volume. Checking a new SSD, enclosure, cable, or file-system change therefore requires a separate app.

## Proposed solution

Add an optional test inside the existing Disk tools area:

- Quick (1 GB) and Standard (4 GB) sequential write/read modes.
- Run only on the selected physical disk, with read-only and free-space checks.
- Use an uncached, immediately unlinked temporary file; include `fsync` in write timing and support cancellation.
- Keep only the latest successful result per volume.
- Disable eject and disk switching while a test is active.

This would use public macOS/POSIX APIs, add no dependency, helper, permission, or background monitoring, and remain opt-in. Results would be practical indicators rather than laboratory-grade storage benchmarks.

A draft implementation is ready for evaluation, but the direction still needs maintainer confirmation before it is considered merge-ready.

## Feature area

System monitor / Tools

## Existing apps

Blackmagic Disk Speed Test and AmorphousDiskMark provide dedicated versions of this workflow.

PathGao:
Tagged `question` because the only open item is the direction call, and that's the maintainer's to make, not mine. Nothing else moves until it's answered.

Two things worth having in front of that decision:

The landing point already exists. The disk panel tracks a selected volume and owns the eject action, so "disable eject while a test runs" is a constraint that can live where the state already is, not new plumbing.

The support cost is the underestimated part. Once a throughput number is on screen it gets compared against Blackmagic's, and every gap arrives as a bug report. "Indicators, not benchmarks" is the right framing, but it has to be visible in the UI, not just in this issue.

I'd hold the draft PR until the direction is confirmed. Opening it now doesn't speed the decision up, and an unreviewed draft that sits is worse than no PR.

PR #1165, opened 2026-08-30
feat(disks): add selected-disk speed benchmark

Mahmutakin99:
## Summary

Add an opt-in sequential read/write benchmark to the existing Disk tools area. It offers 1 GB Quick and 4 GB Standard modes, reports decimal MB/s, remembers the latest successful result per volume, and prevents disk switching or eject while a run is active.

The feature stays within Vorssaint’s existing system-monitor surface and uses public macOS/POSIX APIs. It adds no runtime dependency, helper, permission, or background monitoring. The scratch file is created on the selected physical disk, unlinked before I/O, opened with `F_NOCACHE`, flushed with `fsync`, and guarded by identity, read-only, cancellation, and free-space checks.

User-facing text is supplied for all 13 supported locales.

This is a draft because the direction still needs maintainer confirmation and the full-size benchmark needs real-device use before merge readiness.

## Verification

Mac mini (Mac16,10), Apple M4, 16 GB, macOS 26.6.2 (25G83).

- `swift build`
- `./build.sh` (release bundle, no warnings)
- bundled application `--selftest` (`SELFTEST OK`)
- `./build.sh --test` (`TESTS OK (9150 checks)`)
- Developer bundle installed, signature verified, launched separately as `com.vorssaint.utils.dev`
- The test suite runs an 8 MiB uncached POSIX write/read and covers persistence, cancellation, APFS identity matching, capacity/read-only failures, and locale completeness

Not yet tested: an actual 1 GB or 4 GB run, external/removable storage, unplugging a disk during I/O, or several days of normal use.

## Anything else

Refs #1164


Mahmutakin99:
Withdrawing this draft while local UI and full-size disk benchmark verification are completed.

PR #1167, opened 2026-08-30
chore: ignore local agent and secret files

RuanMD:
## Summary

Ignore local AI-agent and planning workspaces so contributor-specific BMAD, Claude, Serena, and agent files cannot be added accidentally. Also ignore common local environment, credential, service-account, and secret-file names while keeping `.env.example` and `.env.sample` available as safe templates.

## Verification

Ran on Apple Silicon macOS with an ad-hoc release build:

```sh
./build.sh
./build/Vorssaint --selftest
./build.sh --test
gitleaks git . --redact --no-banner
gitleaks dir . --redact --no-banner
```

Results: `SELFTEST OK`, `TESTS OK (9130 checks)`, and no leaks found across 769 commits or the current working tree.

No runtime behavior changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,24 @@ AGENTS.md
 CLAUDE.md
 E2E/
 
+# Local AI agent and planning workspaces
+.agents/
+.claude/
+.serena/
+_bmad/
+_bmad-output/
+
+# Local environment and secret files
+.env
+.env.*
+!.env.example
+!.env.sample
+*.secret
+*.secrets
+secrets/
+credentials.json
+service-account*.json
+
 # macOS noise
 .DS_Store
 


PathGao:
The agent-workspace block is right, and it matches what's already in the file — `.cursor/` and `.gemini/` sit at the bottom for the same reason.

**Fold the new agent dirs into that block.** `.agents/ .claude/ .serena/` are the same fact as `.cursor/ .gemini/`, now written in two sections; whoever adds `.windsurf/` next has two places to pick from. `_bmad*/` is one contributor's tool — `.git/info/exclude` or a global `core.excludesFile` covers that without the repo carrying it.

**The secret block guards nothing here.** No `.env`, `credentials.json` or `service-account*.json` anywhere in the tree. `notarize.sh` and `ci-setup-signing.sh` read credentials from CI env vars and write the decoded `.p8` into `mktemp -d`, outside the working tree. The existing signing lines (`*.p12 *.pem *.p8 *.key`) already cover the shapes that can appear here — your clean gitleaks run over 769 commits says the same. And `.env.*` plus two `!` negations is four lines of subtlety for template files that don't exist. Unless non-Swift tooling is planned, cut it to `.env` or drop it.

I checked main's tree, not history. Nothing tracked matches any new pattern.

**Impact:** none — `.gitignore` doesn't untrack anything, and no file in the tree matches the new patterns.
**Merge:** needs the secret block cut to `.env` (or dropped) first.

PathGao:
this should be done in your locally

Issue #1166, opened 2026-08-30
Adaptive app icons flicker between light and dark appearances in the app switcher


EugeneCarldotme:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When macOS is using dark mode,  ethe Postman and Slack icons briefly display their light-mode appearance while cycling through apps in the Vorssaint app switcher. The icons then return to their dark-mode appearance, causing a visible flicker.

### Steps to Reproduce

1. Set macOS Appearance to Dark.
2. Open Postman, Slack, and several other apps.
3. Invoke the Vorssaint app switcher.
4. Cycle through the apps, moving the selection to and away from Postman and Slack.
5. Observe both icons as the selection changes.

### Feature Area

Other

### Screenshots / Recordings

Expected
https://github.com/user-attachments/assets/9a905dc6-9ad2-46e2-b28b-2d1a0aabfd1c
Issue
https://github.com/user-attachments/assets/c6f7327e-2536-4306-9977-0f4aa8c89ab7

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

### Language

English

Issue #801, opened 2026-08-23
[Bug]: App switcher shows the default app icon instead of the user-selected alternate icon


EugeneCarldotme:
### Description

Some macOS apps ship multiple app icons and let the user choose one from within the app's own settings. When a non-default icon is selected, the Dock and Finder update to the new icon, but the app switcher continues to display the app's default (bundled) icon.
Example: ChatGPT for macOS lets you choose between the OpenAI/ChatGPT mark and the Codex icon. With the Codex icon selected, the Dock shows Codex, but the app switcher still shows the ChatGPT logo.

### Steps to Reproduce

1. Install an app that supports alternate icons (e.g. ChatGPT for macOS).
2. Open the app's own settings and switch the app icon from the default to an alternate one (e.g. ChatGPT → Codex).
3. Confirm the Dock and Finder now show the newly selected icon.
4. Invoke the app switcher.
5. Observe the icon shown for that app.

### Screenshots / Recordings

Expected: The switcher shows the currently selected icon (Codex)
<img width="1568" height="297" alt="Image" src="https://github.com/user-attachments/assets/34cbc98f-6905-4234-99c8-7e6a25ab218b" />
Actual: The switcher shows the default bundled icon (ChatGPT logo) 
<img width="1120" height="236" alt="Image" src="https://github.com/user-attachments/assets/2414f592-458f-438a-b20e-df471b9070f8" />

### Vorssaint Version

v3.3.2

### macOS Version

macOS 27 (Golden Gate)

PR #1170, opened 2026-08-30
Keep adaptive app icons stable in the switcher

EugeneCarldotme:
### What was wrong

**Before:** apps with appearance-aware icons, including Postman and Slack, can flicker between their light and dark artwork while the switcher selection moves.

The switcher gets the current icon through NSWorkspace.icon(forFile:). That returns an appearance-aware NSImage. SwiftUI redraws the image while animating the selected tile, so the same icon can resolve another appearance during the transition.
The change

**After:** adaptive icons keep the appearance Vorssaint resolved when the switcher drew them.

The existing fresh bundle lookup from https://github.com/vorssaintapp/vorssaint-utils/pull/811 remains unchanged in behavior. The switcher resolves that image under the effective appearance reported by Vorssaint and wraps its CGImage pixels in a non-adaptive NSImage before SwiftUI animates it. No cache or view-specific workaround was added. One file, +7/-2.
Verification

./build.sh completes successfully and ./build/Vorssaint --selftest prints SELFTEST OK.

A deterministic AppKit probe confirmed the failure and fix: an adaptive NSImage changed artwork between Aqua and Dark Aqua redraws, while the CGImage snapshot kept the originally resolved artwork under both appearances.


PathGao:
Right call putting the freeze on the model rather than in a view — `DockPreviewPanelView` reads the same `appIcon` and gets it for free. I grepped the other `icon(forFile:)` call sites: the command bar already freezes into an `NSBitmapImageRep`, so there's no sibling site left flickering.

The snapshot collapses every representation into one. `cgImage(forProposedRect: nil, context: nil, hints: nil)` resolves against a generic 1x context, so the bitmap comes back at roughly `icon.size` pixels — and `NSWorkspace.icon(forFile:)` hands back a 32pt image. `SwitcherView` draws these at 72pt and at `SwitcherGridCard.fallbackIconSize`, so on a Retina display that's 32px stretched to 144px. Your probe compared artwork across appearances, not pixel scale, so it wouldn't catch this. I can't run AppKit here — print `icon.size` and `pixels.width` to settle it.

A drawing handler pins the appearance without dropping the reps:

```swift
return NSImage(size: icon.size, flipped: false) { rect in
    NSApplication.shared.effectiveAppearance.performAsCurrentDrawingAppearance { icon.draw(in: rect) }
    return true
}
```

**Impact:** every switcher icon soft on Retina, if `icon.size` is what I think it is.
**Merge:** needs the snapshot's pixel size checked first.

ref #1166

EugeneCarldotme:
I replaced the flattened CGImage snapshot with the suggested drawing-handler wrapper. This keeps the icon pinned to Vorssaint's effective appearance while allowing AppKit to select the correct representation for the destination scale.
./build.sh passes and ./build/Vorssaint --selftest returns SELFTEST OK.

PathGao:
The wrapper is resolution-independent, which settles the pixel-scale question: `NSImage(size:flipped:)` builds an `NSCustomImageRep` with no fixed pixel dimensions, so `icon.draw(in:)` runs against the destination CTM and picks the large rep at 144pt.

One thing left — `NSApplication.shared` is read inside the block. The handler runs whenever the image is drawn, on whatever thread draws it, and it re-resolves the appearance each time instead of capturing it, which is the opposite of freezing. Every other drawing handler here (`StatusItemController.tintedImage`, `composedImage`) captures its inputs outside and only draws inside:

```swift
let appearance = NSApplication.shared.effectiveAppearance
return NSImage(size: icon.size, flipped: false) { rect in
    appearance.performAsCurrentDrawingAppearance { icon.draw(in: rect) }
    return true
}
```

My snippet had it inside; that one's on me.

Also worth a line in the `appIcon` doc comment on why the wrapper exists — otherwise the next reader unwraps it.

**Impact:** appearance re-resolved per draw rather than pinned, and `NSApp` touched off the main thread if AppKit ever draws the image there. I can't run AppKit, so this is read off the code, not observed.
**Merge:** needs the appearance read hoisted out of the block first.

ref #1166

EugeneCarldotme:
I moved the effectiveAppearance lookup outside the drawing handler, so each wrapper now captures its appearance when appIcon is resolved. The handler no longer accesses NSApplication.shared from the drawing thread or re-resolves the appearance during later draws.
The resolution-independent wrapper remains in place, so icon.draw(in:) can still select the appropriate representation for the destination scale.
./build.sh passes and ./build/Vorssaint --selftest returns SELFTEST OK.

PathGao:
The hoist is right and it's in the cheap place: `appIcon` is only read from view bodies (`SwitcherView`, `DockPreviewPanelView`), so `NSApplication.shared` is now touched on main, and each read re-freezes against the current appearance instead of pinning one forever.

I went looking for a follow-on problem and didn't find one. The wrapper allocates a fresh `NSImage` on every body evaluation, which normally I'd flag — but that's the same property #801's fix depends on: a cached instance is exactly what kept the switcher showing the stale bundled icon. So no cache here, deliberately.

Only leftover is the doc comment above `appIcon`, which still explains just the NSWorkspace lookup. Not blocking.

I can't run AppKit, so the pinning itself rests on your probe and on reading the code, not on anything I observed.

**Impact:** none.
**Merge:** ship it.

ref #1166

PR #1175, opened 2026-08-30
feat(panel): add complete keyboard navigation

mayaanhafeez:
## Summary
- add shared keyboard focus, row registration, and layout-derived navigation for the menu panel
- support tab, arrow, Return/Space, value adjustment, nested utility panels, and Escape-back behavior across panel sections and hosted tools
- provide focus indicators and accessibility labels while keeping text editors and native controls in control of their own keys

## Verification
- `swift test --filter PanelKeyboardNavigatorTests`


PathGao:
The `keyboardSection: PanelSectionID?` injection is the right shape — the host decides, so Settings and the launcher stay untouched.

`MediaWorkspaceView` breaks it. It hardcodes `compact ? PanelRowID(.utilities, …)`, but `compact` is also true in the quick launcher (`QuickLauncherView:88` hosts `PanelMediaView`, same as `MenuPanelView:591`). `PanelCleanerView` got a `keyboardSection` param for exactly this case; `PanelMediaView` didn't. So opening media from the launcher registers every row into the panel's `.utilities` section.

Row IDs are written twice — once at `.panelKeyboardRow(…)`, once as a literal in the scroll-follow lookup (`focusedItemID`, `focusedResultID`, `focusedContentAnchor`). Rename one and scrolling silently stops following; media's `default:` case scrolls to the wrong card instead. One constant per row.

I read 80KB of the 296KB diff. `PanelKeyboardNavigator` itself, and everything under `MenuPanel/` after `DiskSection.swift`, I have not seen. No hardware check either.

**Impact:** launcher-hosted media rows land in the panel navigator, which doesn't own that window.
**Merge:** needs the launcher path fixed first; conflicts also need resolving.

mayaanhafeez:
Ready for review. Since the draft review, this now scopes Media keyboard navigation to the panel host and centralizes its row IDs, merges current `main`, limits key handling to the panel popover, passes through keys the navigator cannot act on, protects native controls and Recent Captures confirmation, clears hosted Settings routing on close, preserves lazy panel lists, and runs the package tests through `./build.sh --test`. I also verified the latest Developer build locally.

PathGao:
Both draft-review points are fixed — `MediaKeyboardID` plus the `keyboardSection` passthrough on `PanelMediaView` closes the launcher hole.

Your comment says this "preserves lazy panel lists", but `AppUpdatesListView.list` goes `LazyVStack` → `VStack`. If that swap is there because a lazy container never registers offscreen rows, then `CleanerView`'s results `List` has the same defect and can't be fixed the same way — `List` is always lazy. Which is it?

Row IDs are still written twice outside media: `focusedItemID` re-spells `"appUpdates-\(item.id)-select"`, `focusedResultID` re-spells `"cleaner-group-\(…)-include"`. Media got constants, these two didn't.

`main` hosts 9 `Panel*View`s at `MenuPanelView.swift:575-611` against 7 launcher call sites. I can see `keyboardSection` threaded on 2 of them (Cleaner, Media); the other 7 are past my diff truncation — unverified, not wrong.

`PanelKeyboardNavigator` itself I still haven't read. No hardware check.

**Impact:** app-updates rows render eagerly now; if the swap was needed, offscreen cleaner rows never register at all.
**Merge:** needs the lazy-list answer first.

PathGao:
Sending this back on shape; the goal (keyboard navigation of the menu panel) is fine and nothing on main does it.

**Build system.** `Package.swift` gains `.testTarget(name: "PanelKeyboardNavigatorTests", …)` and `build.sh --test` gains `swift test`, which CI runs. This project builds with a plain `swiftc` call (CONTRIBUTING: `Package.swift` is there so SwiftPM-aware editors can index the code; `build.sh`: "no XCTest needed"). Every CI run would compile the app a second time through SwiftPM with XCTest. Tests here go in `Tests/MetricsTests.swift` on the existing list; drop the target and the `swift test` line.

**Size.** +2777 across 38 files: about 630 lines of navigator and `MenuPanelView` core, and 228 `.panelKeyboardRow(...)` tags across 30 section views. There is no use of SwiftUI's own focus system anywhere in the diff — `@FocusState`, `focusSection`, `onMoveCommand`, `.focusable`. A version that makes the panel key and rides the native focus order would plausibly be a fifth of this and would not need every row tagged by hand. Worth trying before this shape is reviewed further.

**Verification.** `swift test --filter …` is the only line; the 09-01 comment says a Developer build was tried but gives no Mac model, macOS build or what was not tested. Add those.

mayaanhafeez:
@PathGao 

Spiked the native-focus version before replying — standalone accessory app, same setup as ours (`NSPopover`, `.applicationDefined`, `makeKey`). 60 rows, 200pt viewport, 180 `selectNextKeyView` steps, Full Keyboard Access off:

| container | focusable views in key-view loop | rows reachable |
|---|---|---|
| eager `VStack` | 60 | 60 |
| `LazyVStack` | 10 | 10 |
| `List` | 9 | 1 |

Native traversal only reaches rows the container has already realized, and traversing never realizes more — scroll offset stayed 0 in all three. `List` is NSTableView-backed, so its rows never join the key-view loop at all.

That's `CleanerView.swift:736` (`List`) and `AppUpdatesListView.swift:174` (`LazyVStack`) — the two you flagged earlier. Nine panel-hosted views use one or the other. Focus also doesn't scroll to follow, so the scroll-follow code is needed either way. The per-row registration is what makes offscreen rows reachable; that's the reason for it, not a focus ring reinvented by hand.

Build system: you're right, and it's done. `.testTarget` and the `swift test` line are gone, and all 13 navigator cases now run in the `MetricsTests` harness — `./build.sh --test` reports TESTS OK (9575 checks). Reaching that meant splitting the SwiftUI modifiers out of `PanelKeyboardNavigator.swift` and `PanelSectionID` out of `PanelLayout.swift`, so the navigator compiles without the view layer; the `MetricsTests` note about the Command Line Tools not being able to run `swift test` was the deciding point.

Verification: MacBook Air (Mac16,12), M4, 16 GB, macOS 26.6.2 (25G83), app language en-CA. Two displays, separate Spaces, no mirroring: built-in Liquid Retina 2560x1664 plus an external 2560x1440 @ 144Hz as main. Developer build via `./build.sh --dev`, plus `./build.sh --test`.

Walked every panel section by keyboard on that setup, arrows in both directions. Not tested: a non-English app language, or any macOS version other than 26.6.2.

Do you want me to continue with this approach or experiment more with SwiftUI's focus system?

PathGao:
The spike answers it — carry on with this shape. `List` reaching 1 of 60 and never realizing more under traversal is what I was missing, and it kills the native-focus alternative. Build-system point is resolved too.

`AppUpdatesListView.rows` is still `LazyVStack` on this head, so the 09-01 inconsistency is gone: both lazy containers declare their order through `panelKeyboardRowList`. Nothing left there.

One thing your table implies but doesn't cover: the order is declared, the actions are not. `.panelKeyboardRow(_:actions:)` sits on the row view, so a row that has never been realized is in the arrow order with nothing attached to it. Hold Down to run to the end of a long cleaner result list and press Return before `scrollTo` has built that row — does `handleKeyDown` no-op, or does it defer? `PanelKeyboardNavigator.swift` is past my diff truncation, so I'm asking, not claiming.

**Impact:** if activating an unrealized row is a silent no-op, Return does nothing on rows reached faster than the list builds them.
**Merge:** needs that answer first.

mayaanhafeez:
No-op, not deferred: activateFocused looks up the focused row’s registered actions and returns false when none exist, so the monitor hands Return back to AppKit. Nothing is queued (MetricsTests: “return on a row with no action is handed back”).

In normal keyboard traversal, moving focus also scrolls the target row into view. Realizing that row registers its actions during layout, normally before the next key event is processed. I hardware-tested this by holding Down through a long cleaner result list and pressing Return on the final row; it activated correctly.

The reachable missing-action case is mouse-scrolling the focused row out of a lazy container. That unregisters its actions while preserving focus, so Return is handed back. The next Up/Down scrolls to and realizes its target row again. That behavior is deliberate: activating a row that is no longer visible would be worse.

So there is no deferred activation. I also couldn’t reproduce a silent no-op during keyboard traversal.

PathGao:
That closes it. Handing Return back when the focused row has no registered actions is right, and unregistering on mouse-scroll-out instead of activating an off-screen row is the right side of that trade.

I also checked the new Escape guard against main: both `isPresentingPopoverModal` sites are a real `.confirmationDialog` / `.alert`, so they cancel themselves and the guard holds.

One leftover from the launcher hole. `QuickLauncherView:86-98` hosts seven `Panel*View`s; Cleaner and Media now take `keyboardSection` defaulting to nil, so those two are clean. Homebrew, URLCleaner, Uninstaller, WindowLayout and QuickToggles are past my diff truncation — if any of them spells a literal `PanelRowID(.section, …)` the way `BrightnessSection` does, launcher rows register into the panel navigator again. Grep on your side; I can't see those files.

**Impact:** none from the activation question. The launcher case is unverified, not claimed broken.
**Merge:** needs the conflict with main resolved first.

mayaanhafeez:
Grepped. Homebrew, URLCleaner, Uninstaller and WindowLayout all spell `PanelRowID(.utilities, …)` and QuickToggles spells `PanelRowID(.toggles, …)`, so the literal IDs are there. They don't reach the navigator from the launcher though: `PanelKeyboardRowModifier` (and the row-list and chrome variants alongside it in `PanelKeyboardRegistration.swift`) only register when `isMenuPanelHost` is true, and that environment flag is set once via `markMenuPanelHost()` at the root of `MenuPanelView`. The launcher tree never sets it, so those rows render as plain content there. The `keyboardSection` parameter on Cleaner and Media is a second layer on top of that gate, not the only one.

PathGao:
The gate answers it. `isMenuPanelHost` set once at the `MenuPanelView` root and never in the launcher tree covers all nine hosted utilities, so the literal `PanelRowID(.utilities, …)` in the other five is safe. Conflict with main is resolved too.

That gate is also why `keyboardSection` worries me — it fails the other way round. A literal-tagged view is reachable anywhere under the panel root. A param-tagged view is dead until every call site passes it, and nothing catches a miss.

`HomebrewOperationStatusView` has five call sites on main; three are panel-reachable — `AppUpdatesListView`, `PanelUninstallerView`, `PanelHomebrewView`. Only the first is inside my diff slice, and it got the param. If the other two didn't, cancel / clear log / open Terminal / copy register no rows in exactly the two panels where a Homebrew operation actually runs. Grep on your side; both files are past my truncation, so this is a question, not a claim.

`PanelKeyboardNavigator.swift` I still haven't read. No hardware check.

**Impact:** a missed `HomebrewOperationStatusView` call site leaves its buttons unregistered and the arrows skip past them.
**Merge:** needs those two call sites confirmed first.

mayaanhafeez:
Confirmed by grep that all three panel-reachable HomebrewOperationStatusView call sites are wired correctly:

- PanelUninstallerView passes keyboardSection: .utilities
- PanelHomebrewView passes keyboardSection: .utilities
- PanelAppUpdatesView passes .utilities into AppUpdatesListView, which forwards it to HomebrewOperationStatusView

Therefore cancel, clear log, open Terminal, copy command, details, and copy log all register as keyboard rows in the relevant panel views. The Settings call sites keep the default nil, as intended.

That resolves the call-site concern. I’ll also complete a physical-keyboard smoke test of PanelKeyboardNavigator before merging, covering initial focus, arrow and Tab traversal, activation with Return/Space, drill-in Escape behavior, scrolling, and the Homebrew operation controls above.

mayaanhafeez:
@PathGao pushed 5883838, a fix for the arrows skipping around dynamic lists (uninstaller, CPU/RAM/GPU breakdowns, and the other declared lists). Two causes, and two of the changes reach outside the navigator itself, so flagging them:

**Navigator.** `PanelKeyboardRowOrderModifier` was sorting row ids one at a time. Every id from a `panelKeyboardRowList` shares the list's frame and entry index, so the tiebreak never separated them and they came out in `Dictionary` iteration order. Sorting now happens on whole layout entries in `PanelKeyboardNavigator.orderedRows(from:)`, which is pure and covered in `MetricsTests` (declared list order, same-line ties, duplicated ids).

**Outside the navigator:**

1. `ProcessUsageRow` now registers every row as a keyboard stop, not just processes whose app can be activated. Previously kernel_task, WindowServer etc. had no row, so Down skipped over them. A non-activatable row takes focus and hands Return back (no action registered). If you'd rather those rows stay unfocusable, that's a one-line revert.
2. `.panelKeyboardRowList` moved from the inner `LazyVStack` onto the `ScrollView` in `AppPickerView`, `PanelHomebrewView`, `PanelRecentCapturesView` and `AppUpdatesListView`. The stack's frame in the panel coordinate space shifts with the inner scroll offset, so a scrolled list could sort above the rows physically above it. Same mechanism, not the reported bug. `CleanerView` already had it on the `List`.

`./build.sh --test` reports TESTS OK (9837 checks); `./build.sh --dev` builds. Hardware pass on the same MacBook Air setup as before is next.

PathGao:
Sorting whole layout entries instead of single ids is the right level — a declared list is one entry, so its members keep declaration order.

Keep the `ProcessUsageRow` change. A rendered row the arrows silently jump over is worse than one that takes focus and hands Return back, and that's the rule you already set for rows with no registered action.

The four moved views leave a fifth with the same shape. `PanelClipboardView` on main is `ScrollView { LazyVStack }` (110-113), hosted at `MenuPanelView:596`, one line above the `PanelRecentCapturesView:601` you did move. Does it declare a row list? If it does and the modifier is still on the inner stack, clipboard entries scrolled inside the panel sort against the wrong frame. `PanelWindowLayoutView:203` and `DiskSection:147` are `LazyVGrid` with no inner `ScrollView`, so those are fine.

`PanelKeyboardNavigator.swift` and the four moved files are past my diff truncation — I checked shapes on main, not the patch. No hardware check.

**Impact:** if clipboard declares rows on the inner stack, arrows there jump around the way the uninstaller did.
**Merge:** needs the clipboard answer first.

mayaanhafeez:
Clipboard declared no list at all, which turned out to be worse than a list on the inner stack. Each card's five buttons registered individually through the row modifier, inside a plain VStack in a ScrollView capped at 260pt. Two consequences: Up/Down walked sideways across a card's buttons before reaching the next card, and once the inner list scrolled, off-screen cards measured above the enable/clear/show-window controls in the panel space, so the arrows hopped into the header and back. Same failure as the uninstaller, different route.

Fixed in 9dbb295 the same way as the other four: panelKeyboardRowList on the ScrollView, not the stack, plus a panelKeyboardRowGroup per card so Up/Down is one stop per card and Left/Right walks its buttons. The group only lists buttons the card actually registers, since the up/down arrows are conditional and a missing neighbour would strand Left/Right.
Three follow-ups on top, all clipboard-only:
- e388bb7: cards report their frames in the list's own coordinate space. If a manual scroll pushes the focused card fully out of the viewport, focus moves to the nearest visible card, same button where it exists. Return no longer acts on a button you can't see. The outer panel keeps its existing rule of leaving scroll alone while focus stays put.
- 5eb749b: focus changes and frame updates arrive in either order, so both go through one reconcile step that checks whether focus itself moved before treating a hidden focused card as a manual scroll.
- 5238194: the list-height preference reduced with value = nextValue(), and the scroll content's default 0 could land after the background's measurement. Height read as 0, every card counted as visible, nothing scrolled. Now max.

Verified on hardware this time by driving the dev build with synthetic key and scroll-wheel events: eleven Downs from a fresh panel walk header, controls, then card by card with the hidden seventh card scrolling into view; a partial scroll keeps focus; a scroll that fully hides the focused card moves focus to the bottom-most visible one.

PanelWindowLayoutView and DiskSection unchanged, agreed they're fine as grids with no inner scroll.

Also merged main in 8ee6fe9. The only conflict was KeepAwakeIconPicker, where your line.diagonal fix landed inside the tintGlyph helper this branch split out. Applied as-is.

vorssaint:
closing this. the panel now also lives inside the dynamic island, which has its own keyboard handling, and this design assumes the panel only ever lives in the popover window, so it would need to be redesigned rather than just updated
every row in the panel has to be tagged by hand and every shared control grows a parameter for it, so any row added later is silently unreachable from the keyboard, and that is more upkeep than the panel can carry
a button that turns disabled after you press it, like select all, upgrade all or reset volume, drops out of the keyboard order and the focus ring jumps back to the section tabs
a small pr that only lets escape back out of a metric's detail page or a tool opened inside the panel, and moves between the section tabs with tab and the arrows using the framework's own focus handling, could still come in

PR #1176, opened 2026-08-30
feat: add safe disk read/write speed test

Mahmutakin99:
## Summary

Adds a selectable, verified sequential disk read/write benchmark to the Disks panel. It writes a temporary unlinked file on the selected mounted volume, performs uncached sequential I/O where supported, flushes it, reads it back and verifies the data.

Quick runs one 1 GB pass. Standard runs three 4 GB passes and reports the median, which reduces cache and background-activity variance. The implementation validates writable volume identity and available capacity, persists the last result per volume, supports cancellation/cleanup, and blocks every user-facing disk-ejection path while a benchmark is running.

Progress uses one overflow-checked total for both write and read work across every pass, so it remains monotonic and finishes exactly at 100%. Unsupported `F_NOCACHE` errors fall back to normal I/O; genuine I/O failures are still reported.

The interface is localized in every currently supported language.

## Verification

- `swift build` — passed.
- `./build.sh --test` — passed: 9259 checks.
- `git diff --check upstream/main...HEAD` — passed.
- `./build.sh` — passed; release bundle was assembled and its code signature verified.
- `build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest` — passed.

## Anything else

Refs #1164


PathGao:
The unlinked-scratch-file design is right: the fd holds the space, and per-pass read-back verification means a cache can't fake the numbers.

**Eject guard covers one of four call sites.** `DiskSection` disables its three buttons on `benchmark.isRunning`, but `QuickTogglesService.ejectAllDisks()` never consults the service and is reachable from `QuickTogglesSection`, `CommandBarCatalog` and `RadialMenuService`. Unmounting a volume with a 4 GB unlinked file open fails, so the user gets a silent eject failure — the thing the DiskSection guard exists to prevent. The guard belongs in `ejectAllDisks()`, not the view.

**Progress rewinds once per pass.** `runPass` publishes `.reading` at `completedBytes: passOffset` after the write phase already reached `passOffset + byteCount`. Standard jumps backward six times. The total should be `2 × byteCount × passCount`, and it's computed twice — `DiskBenchmarkService.start` and `DiskBenchmarkRunner.run` — with different overflow behaviour.

**"Uncached sequential I/O where supported"** doesn't match the code: it throws `.cacheControl` when `F_NOCACHE` fails, while the sibling `F_FULLFSYNC` path falls back to `fsync`. Fall back, or drop "where supported". I didn't confirm which volumes reject `F_NOCACHE`.

Diff was truncated at 80 KB; the tail of `Tests/MetricsTests.swift` and anything after it I didn't read.

**Impact:** eject fails silently from three UI paths while a benchmark runs; progress jumps backward each pass.
**Merge:** needs the eject guard inside `ejectAllDisks()` first.

ref #1164

Mahmutakin99:
Addressed in 891244f after rebasing onto the latest `main`:

- Added the benchmark guard inside `QuickTogglesService.ejectAllDisks()`, covering Quick Toggles, Command Bar, and Radial Menu. `DiskProtectionService` now also guards its direct single/all eject entry points.
- Centralized the overflow-checked progress total as `2 × byteCount × passCount`. Write/read offsets now stay monotonic across passes and finish exactly at the total.
- Unsupported `F_NOCACHE` errors (`ENOTSUP`, `EINVAL`, `ENOTTY`) now fall back to normal I/O. Genuine errors such as `EIO` still return `.cacheControl`.
- Added regression coverage for the eject guard, exact/monotonic progress, overflow, supported fallback errors, and genuine cache-control failures.

Verification: `swift build`, 9,259 unit checks, release bundle build, packaged self-test, code-signature verification, and `git diff --check` all pass.

Mahmutakin99:
Closing this for local review before any future submission.

PR #1179, opened 2026-08-31
fix: make dictation activation and delivery reliable

RuanMD:
## Summary

- make a second short press stop an active hands-free Hybrid dictation session, matching the existing VoiceInk interaction model;
- resolve the focused accessibility target when transcription completes, so users can switch apps while the provider is processing and still paste into the current field;
- retain safe clipboard fallback when Accessibility or a focused field is unavailable;
- add regression coverage for Hybrid second-tap behavior and current-target insertion decisions.

## Validation

- `./build.sh --test` (9,236 checks)
- `./build.sh`
- `./build/Vorssaint --selftest`
- `gitleaks detect --no-banner --redact`

This PR is intentionally incremental: further dictation resources will be added as separate commits and validated within the same branch.


PathGao:
Permission strings updated across all 12 `.lproj` plus the base plist, in step with what the feature actually sends.

**1. The modifier tap starts deaf.** `DictationModifierShortcutTap.sync` sets `registeredKeys = keys`, then calls `stop()`, which does `registeredKeys.removeAll()`. On the first sync (`tap == nil`) the tap gets created with an empty key set, so `handle` drops every flagsChanged event. It only works from the second sync on, which takes the early return. Move the assignment below `stop()`.

**2. `DictationModifierKey` duplicates `SuperKeySource`.** main already has the same modifier→`kVK_*` map in `SuperKeySupport.swift`, and it names keys with `symbol`/`systemImage` — your `displayName` returns Portuguese in all 13 languages. Extending `SuperKeySource` with the left variants and fn kills both the copy and the localization gap.

**3.** `sanitizedDays` allows 0, and `isExpired` then returns true for every entry — if the settings UI offers 0 as "keep forever", the first sweep wipes history. Unverified: the UI and `DictationService` sit past the 80 KB diff cutoff, so nothing in them is reviewed here.

**Impact:** single-modifier activation does nothing until a later sync; the key picker reads Portuguese in every language.
**Merge:** needs the `sync()` ordering fix first.

PathGao:
Closing in favour of #1201. It carries all fifteen commits from this branch plus twelve more, and you named it as the canonical dictation PR, so this keeps the review in one place. Nothing here is lost.

PR #1184, opened 2026-08-31
fix(capture): let a straddling window fall back to the single-window routes

iltonandrew:
Refs #1098 — the first of your three post-merge notes.

`captureAttached` chose its display by intersection, so a window straddling two displays came back as the slice on whichever display matched first — and this route runs ahead of the `captureCoversWindow` guard that exists to reject exactly that shape, so the clipped composite beat the correct fallback. The display is now taken only when it contains the whole captured area; a straddling window finds no display, the composite declines, and the single-window routes answer with their existing off-screen-slice handling, as before #1146.

A source-shape check pins the predicate (contains present, intersects absent in `captureAttached`), verified green as committed and red with the predicate reverted.

Verification: `./build.sh` warning-free; `./build.sh --test` passes with only the known `dispatch pool starvation` machine-specific failure. Not verified live: an actual two-display straddle — no second display here; the change restores the pre-#1146 behavior for that geometry rather than introducing a new one.


PathGao:
The source-shape check is the right guard here — `captureAttached` is private and needs live `SCShareableContent`, so pinning the predicate is the only thing that can hold it. I also checked the other display-picking sites: `RecorderCaptureEngine` selects by `displayID`, so there's no sibling with this shape.

The predicate reaches further than the body claims. A two-display straddle isn't the only thing with no containing display — a window dragged off the left, right or bottom edge of a **single** display doesn't fit either. That's the geometry `captureWindow`'s own doc comment names ("a window that runs off the screen"), so it's reachable with one display, and on it #1098's symptom comes back: the shot loses the sheet.

A narrower predicate keeps both:

```swift
let hits = content.displays.filter { $0.frame.intersects(plan.bounds) }
guard windows.count == plan.windowIDs.count, hits.count == 1, let display = hits.first,
```

Only the straddle is rejected; the off-edge single-display window still composites and is still clamped by `ScreenshotSupport.clamp`.

Read from the source, not observed — no second display here, and I can't run live capture.

**Impact:** a window hanging off a screen edge with a sheet attached loses the sheet again. One display, no straddle needed.
**Merge:** needs the single-display off-edge case kept on the composite route first.

iltonandrew:
Taken as specified in 4f4120e: the composite now declines only when more than one display intersects the captured area, so the off-edge single-display window keeps its sheet and the crop clamps what is on screen. The shape needle pins the counting predicate and rejects the containment form, checked both directions. Same caveat as yours — read, not run, no second display here.

PathGao:
4f4120e is what I asked for, and the needle holds up both directions — `hits.count == 1` present, `.contains(plan.bounds)` absent, and the split terminates on `windowBounds` so it stays scoped to `captureAttached` rather than swallowing the rest of the file.

Re-grepped `main` for the display-picking shape after the change: every other site selects by `displayID` — `RecorderCaptureEngine` and the `captureDisplay` / `captureDisplayRegion` / `captureAllDisplays` entries — so no sibling is left on the intersect form.

Nothing to flag. Same caveat as yours: read, not run. The two-display straddle itself is still unverified here.

**Impact:** none beyond the fix — an off-edge single-display window keeps the composite route, a straddle falls through to the single-window routes.
**Merge:** ship it.

ref #1098

PR #1185, opened 2026-08-31
fix(capture): confirm attached windows through Accessibility when it is granted

iltonandrew:
Refs #1098 — the second of your three post-merge notes: geometry cannot tell a compose window inside a maximized viewer from a sheet, and the window list carries no marker.

When Accessibility happens to be granted, the capture asks the real question instead of guessing. The target application's window map is resolved through the same `AXWindowResolver` bridge the switcher already uses, and the rule is single-sided: only a candidate whose resolved AX element positively reads as a standard window (`kAXStandardWindowSubrole`, or the fullscreen subrole the auto-quit and enumeration paths already treat as standard) is filtered out — the maximized-parent compose case. Everything Accessibility cannot answer for — an unresolved element, an unreadable subrole, a sheet, a dialog — stays in the shot, so an AX blind spot degrades to the geometric answer rather than losing the dialog. Without the permission the geometric rule stands alone; the feature still requires only Screen Recording. No per-element messaging timeouts: the process-wide bound set at launch owns that number.

The filtering decision is pure — `ScreenshotCapturePolicy.confirmedAttachment(_:confirmedIDs:)` — and behaviour-tested: a nil answer passes the plan through, a partial confirmation keeps order and drops the rest, an empty intersection drops the plan. A source-shape check pins that the AX path stays behind `Permissions.shared.accessibility`.

Verification: `./build.sh` warning-free; `./build.sh --test` passes with only the known starvation failure. Needs hardware: whether a real Mail compose window reports `AXStandardWindow` and how a real sheet's AX element resolves to its WindowServer id — the two assumptions the filter rests on. After review, that claim is enforced rather than asserted: only a window Accessibility positively identifies as standard is filtered out — unresolved elements, unreadable subroles, sheets and dialogs all stay, so an AX blind spot degrades to the geometric answer instead of losing the dialog. The per-element timeouts are gone too; the process-wide bound set at launch owns that number.


PathGao:
Splitting the filter into a pure `confirmedAttachment` with nil / partial / empty tests is the right seam: the AX side stays untestable, the decision doesn't.

One thing contradicts the body. You write that if the hardware assumptions disagree, the filter "fails toward the geometric answer, not toward dropping a dialog." It only does that when the whole window map is missing. A candidate whose element never turns up in `elementsByID` hits `continue`, stays out of `confirmed`, and if it was the only candidate `confirmedAttachment` returns nil — the dialog is dropped. That is exactly the second assumption you flagged: a sheet whose AX element doesn't resolve to its WindowServer id. It regresses against main for anyone who has Accessibility granted.

Unresolvable is not the same as rejected. A compose window resolves and reports `AXStandardWindow`, so keeping unknowns still filters it out:

```swift
guard let element = elementsByID[candidateID] else {
    confirmed.insert(candidateID)   // AX had no answer for this one
    continue
}
```

Minor: four more literal `0.35` timeouts, which is the pile #938 is about.

Not verified: the real-hardware roles either way — I can't reproduce a Mail sheet.

**Impact:** with Accessibility granted, apps whose sheets AX can't resolve lose the dialog from window screenshots, worse than today's geometry-only behaviour.
**Merge:** needs the unresolved-candidate branch first.

iltonandrew:
You were right, and the body's claim was backwards until b56309f: an unresolved candidate was dropped, which is exactly failing toward losing the dialog. The function is now built around your rule — absence of an answer confirms; only a window Accessibility positively identifies as standard (`kAXStandardWindowSubrole`, resolved and read) filters out. That made the sheet/dialog identification branches dead weight, so they are gone rather than kept as decoration. The 0.35 literals went with them — `boundAccessibilityWaits()` at launch already owns that number process-wide, so the copies were the #938 shape again. PR body corrected to say what the code now enforces.

PathGao:
The inversion landed right: absence of an answer confirms, only a positively-read `AXStandardWindow` filters out. That version can't regress against main.

Two leftovers from the version that's gone.

The children pass can't change the output any more. A sheet child isn't a standard window, so it's confirmed; a candidate the pass never adds is confirmed by the absence branch. Anything that could read `AXStandardWindow` is already in `kAXWindowsAttribute`. So it's an AX round trip per child of the target, on the capture path, for a decision it can't reach.

The body's second paragraph still describes the deleted code — `kAXSheetRole`, `kAXDialogSubrole`, standard-window-is-filtered as the positive rule — while the last paragraph says the opposite. On a squash merge that contradiction becomes the commit message.

Minor, not a bug: `AutoQuitService.isStandardWindow` and `WindowEnumerator` both read the standard set as `AXStandardWindow` + `AXFullScreenWindow`. Containment likely makes the second unreachable here, so it's a consistency note only.

**Impact:** none on capture output — dead AX calls plus a body that misdescribes the diff.
**Merge:** needs the children pass dropped and the second paragraph fixed first.

ref #1098

iltonandrew:
Both taken in f306469. The children pass is gone — your dead-code reading was exact: a sheet child confirms as not-standard, an unadded candidate confirms through the absence branch, so it was an AX round trip per child for a decision it could not reach. The target lookup stays only as the map-names-the-target existence gate. The consistency note taken too: the filter now reads the same standard set as `AutoQuitService.isStandardWindow` and `WindowEnumerator` — `AXStandardWindow` and `AXFullScreenWindow` both count. The body's second paragraph was rewritten when you flagged it, so the squash message now describes the code that exists.

PathGao:
Both closed. The children pass is gone, the standard set now reads the same as `AutoQuitService.isStandardWindow` and `WindowEnumerator`, and the body's positive rule matches what the code enforces.

Read the rest of the diff, nothing that should hold it. Two notes, neither blocking:

`AXStandardWindow || AXFullScreenWindow` is now spelled out in three places — `AutoQuitService.isStandardWindow`, `WindowEnumerator`, and this filter. Making them agree was the right move inside this PR; collapsing them into one predicate is its own change.

The source-shape test matches the two branches' exact indentation, so a reformat breaks it with a message about AX semantics rather than about formatting. Cheap to live with, worth knowing when it fires.

Still unverified from here: the real subroles on hardware — I can't run a Mail sheet either, so both assumptions in the body stay assumptions until someone does.

**Impact:** none.
**Merge:** ship it.

ref #1098

Issue #1147, opened 2026-08-30
Window recording omits attached sheets and modal dialogs, like #1098 did for screenshots


iltonandrew:
Splitting this out of #1098 at @PathGao's suggestion, so the screenshot fix can land on its own.

### What happens

Start a screen recording by clicking a window, then have the app open a sheet, alert or modal dialog on top of it. The dialog is not in the video, exactly as it was not in the screenshot in #1098: macOS gives a sheet a window of its own, and a recording scoped to one `CGWindowID` never reaches it.

### Why it is separate

`RecorderCaptureEngine.filter(for:in:excluding:)` builds `SCContentFilter(desktopIndependentWindow:)` when `RecorderSupport.Region.windowID != nil`, which is set only when a window was clicked. That is the same root cause as #1098 and the third of the three call sites on `main` (the fourth, `WindowPreviewProvider` for switcher thumbnails, is fine as it is — a thumbnail of the window itself is what that wants).

It is not the same fix, though, which is why it is not in the screenshot PR. A screenshot resolves the attached windows once, at the instant of capture. A recording is a live stream: sheets appear and disappear while it runs, so whatever answers this has to keep answering it for the duration, and a filter fixed at start time cannot. That may push the recorder toward a bounds-limited display capture rather than resolving children, which is a different mechanism from the one #1098 needs.

### Where the screenshot fix landed, for reference

PR #1146 adds `ScreenshotCapturePolicy.attachedCapturePlan(target:frontToBack:)`, a pure decision that treats a window as attached when it belongs to the same application, sits in front of the clicked window, overlaps it, and is smaller in both directions — then draws them together with `SCContentFilter(display:including:)`. The decision is reusable; the capture around it is not, since it produces one image.

Deliberately not using the Accessibility API there, which does know the true sheet-parent relationship: screenshots need only Screen Recording today, and an AX lookup would put a second permission behind the feature. The same reasoning would apply here.

### Not verified

I have not reproduced this — it follows from the shared call site and #1098's confirmed behaviour rather than from a recording I made. Worth a check with Mail or System Settings before anyone builds on it: record a window, trigger a modal dialog over it, and see whether the dialog is in the resulting file.

PathGao:
Right call to split, and the mechanism holds: the recorder's window branch is one of exactly three `desktopIndependentWindow` sites on `main`.

[...]

PR #1186, opened 2026-08-31
fix(recorder): keep a clicked window's sheets and dialogs in the recording

iltonandrew:
Refs #1147

A recording started by clicking a window was scoped to that one CGWindowID, so a sheet or dialog the app stacked on it mid-recording was never in the video — same root as #1098, but a live stream cannot resolve children once at start and be done.

This takes the bounds-limited display capture you leaned toward: when a clicked window's recording begins, the filter is the display excluding Vorssaint's own capture UI, with the stream's source rectangle fixed to the window's frame. Whatever the app stacks inside that rectangle is in the video by construction, with nothing to track while sheets come and go. Every recording builds that one filter — there is no second path left.

Deliberate behavior changes, stated plainly — four, after review: the recorded rectangle stays fixed if the window is moved mid-recording (which the pointer track and zoom already assume, so the old following filter was the thing out of step); another window dragged over the rectangle is visible in the recording (it is what the screen showed); the rectangle records the current Space, so switching Spaces mid-recording records what replaces the window there, where the old filter followed the window across; and a window spanning two displays records the part on the display it was picked from, where main scaled the whole window into those dimensions.

That fourth one is what dropping the straddle fallback costs, and the selection path had already made the choice: `anchorRect` comes from `snappedPixelRect(raw, in: displayPixels)`, which intersects with the display and clamps to its edges, so no `Region` escapes the display it was picked from and the fallback's predicate could not be false. `pixelRect` has been that clipped rectangle all along, and the pointer track, the zoom and the dimensions badge all read it, so the on-display slice is the reading consistent with everything downstream.

The truth table went with the function it tested. In its place the invariant that makes the fallback unnecessary is tested where it lives — `snappedPixelRect` clipping a straddling rect to its display, and leaving a contained one alone — with shape checks that no filter mode, clicked window or scaling survives in the engine, and that every recording shows its boundary guide.

Verification: `./build.sh` warning-free; `./build.sh --test` passes with only the known starvation failure. Needs live recording hardware: sheets appearing/disappearing mid-recording landing in the file, the fixed-rectangle behavior when the window moves, exclusion of the recording indicator, and mixed-DPI dimensions. No second display here, so the straddle geometry is read off the selection path rather than run.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotSelectionController.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotSelectionController.swift
@@ -346,8 +346,7 @@ final class ScreenshotSelectionController {
     /// anchored to it. The point rectangle is derived from the SNAPPED pixels,
     /// so what gets recorded and what the person was shown never disagree.
     private func region(fromView viewRect: CGRect,
-                        on panel: ScreenshotOverlayPanel,
-                        windowID: CGWindowID?) -> RecorderSupport.Region {
+                        on panel: ScreenshotOverlayPanel) -> RecorderSupport.Region {
         let displayPixels = CGRect(origin: .zero, size: panel.pixelSize)
         let raw = ScreenshotSupport.imagePixelRect(fromView: viewRect,
                                                    viewSize: panel.screenFrame.size,
@@ -360,7 +359,6 @@ final class ScreenshotSelectionController {
                                  height: snapped.height / scale)
         return RecorderSupport.Region(
             displayID: panel.displayID,
-            windowID: windowID,
             pixelRect: snapped,
             anchorRect: ScreenshotSupport.cocoaRect(fromFlippedView: snappedView,
                                                     screenFrame: panel.screenFrame),
@@ -373,11 +371,11 @@ final class ScreenshotSelectionController {
         markCapturePending()
         Self.lastRegion = (panel.displayID, viewRect)
         if activeMode == .geometry {
-            finish(.region(region(fromView: viewRect, on: panel, windowID: nil)))
+            finish(.region(region(fromView: viewRect, on: panel)))
             return
         }
         if scrollingCaptureEnabled {
-            finish(.scrollingRegion(region(fromView: viewRect, on: panel, windowID: nil)))
+            finish(.scrollingRegion(region(fromView: viewRect, on: panel)))
             return
         }
         let pixelRect = ScreenshotSupport.imagePixelRect(
@@ -411,11 +409,11 @@ final class ScreenshotSelectionController {
         guard activeMode != .color else { return }
         markCapturePending()
         if activeMode == .geometry {
-            finish(.region(region(fromView: frame, on: panel, windowID: windowID)))
+            finish(.region(region(fromView: frame, on: panel)))
             return
         }
         if scrollingCaptureEnabled {
-            finish(.scrollingRegion(region(fromView: frame, on: panel, windowID: windowID)))
+            finish(.scrollingRegion(region(fromView: frame, on: panel)))
             return
         }
         Task { @MainActor [weak self] in
@@ -440,7 +438,7 @@ final class ScreenshotSelectionController {
         markCapturePending()
         if activeMode == .geometry {
             let whole = CGRect(origin: .zero, size: panel.screenFrame.size)
-            finish(.region(region(fromView: whole, on: panel, windowID: nil)))
+            finish(.region(region(fromView: whole, on: panel)))
             return
         }
         if let frozen = panel.frozenImage {
--- a/Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
@@ -217,17 +217,12 @@ final class RecorderCaptureEngine: NSObject {
                                   mode: mode)
         case .displayRegion:
             guard let display else { return nil }
+            // The fixed display rectangle records the current Space, so a Space
+            // switch records what replaces the window there, matching the
+            // fixed-rectangle contract assumed by the pointer track.
             let excluded = Set(windowNumbers.map { CGWindowID($0) })
             let ownPID = NSRunningApplication.current.processIdentifier
             let ownWindows = content.windows.filter { $0.owningApplication?.processID == ownPID }
-            if window != nil {
-                let protectedWindows = ownWindows.filter { excluded.contains($0.windowID) }
-                return PreparedFilter(
-                    content: SCContentFilter(display: display,
-                                             excludingWindows: protectedWindows),
-                    mode: mode)
-            }
-
             guard let ownApplication = content.applications.first(where: { $0.processID == ownPID })
                     ?? ownWindows.compactMap(\.owningApplication).first else { return nil }
             let exceptedIDs = RecorderSupport.exceptedOwnWindowIDs(
--- a/Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
@@ -82,11 +82,11 @@ final class RecorderCaptureEngine: NSObject {
         // Keep the call and optional binding as separate type-checking targets.
         // Older Swift compilers otherwise crash in their constraint walker when
         // this expression sits inside the larger async start routine.
-        let preparedFilter: PreparedFilter? = Self.filter(
+        let preparedFilter: SCContentFilter? = Self.filter(
             for: region,
             in: content,
             excluding: excludedWindowNumbers)
-        guard let preparedFilter else { return .noContent }
+        guard let filter = preparedFilter else { return .noContent }
 
         let configuration = SCStreamConfiguration()
         configuration.width = Int(region.pixelRect.width)
@@ -95,9 +95,9 @@ final class RecorderCaptureEngine: NSObject {
                                                     timescale: CMTimeScale(frameRate))
         configuration.pixelFormat = kCVPixelFormatType_32BGRA
         configuration.colorSpaceName = CGColorSpace.sRGB
-        // The straddling-window fallback keeps the output dimensions chosen
-        // when recording starts as that window's buffer changes size.
-        configuration.scalesToFit = preparedFilter.mode == .independentWindow
+        // The recorded rectangle is fixed and the source rect matches the
+        // output size, so there is nothing to scale into the frame.
+        configuration.scalesToFit = false
         configuration.preservesAspectRatio = true
         configuration.captureResolution = .best
         // The pointer is drawn by us afterwards, from the track the sampler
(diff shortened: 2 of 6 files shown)


PathGao:
The mode split is right, and the fixed rectangle isn't a cost: `Region`'s doc already promises the recorded area never moves, "which is what the pointer track and the zoom assume". The old filter was the thing disagreeing with the pointer sampler. Keep it.

**Own-window exclusion is weaker on the new path.** `excludingWindows: protectedWindows` is a fixed ID list from start time. The existing branch excludes the application with exceptions, so chrome appearing *mid*-recording (the mic-unavailable `QuickToolHUD` in `ScreenRecorderService.onMicrophoneUnavailable`) stays out; on the new path it lands in the video. That construction works as-is on a display filter with a `sourceRect` — drop the branch, keep one path.

**A third deliberate change is missing.** `SCContentFilter(display:)` follows the current Space; `desktopIndependentWindow` followed the window. Switch Spaces mid-recording and the video becomes the other Space; same if the window is minimized. Name it in the body or gate on it.

Not verified: no recording hardware here, and the Space claim is the API contract, not a run.

**Impact:** clicked-window recordings — app chrome shown after start can be baked in, and a Space switch replaces the content.
**Merge:** needs the own-window exclusion collapsed onto the existing `excludingApplications:exceptingWindows:` path first.

ref #1147

iltonandrew:
Both taken in cfdb868. The bounded path now falls through to the same application-with-exceptions exclusion the display path uses — one construction, so the microphone-unavailable card and any other mid-recording chrome stays out of the video identically on both. The Space contract is named where the filter is chosen and in the body as the third deliberate change: the rectangle records the current Space, which is consistent with the fixed-rectangle contract the pointer track already assumes. Glad the fixed rectangle turned out to be the correction rather than the cost.

PathGao:
Both collapsed cleanly, and the Space contract is named where the filter is chosen rather than only in the body.

`RecorderIndicator.showRegionGuide` still guards on `region.windowID == nil`, so a clicked-window recording gets no boundary panel. That was right when the capture followed the window; it's wrong now that the rectangle is fixed and, as the body says, another window dragged over it lands in the video — nothing on screen tells the person where that rectangle is. `ScreenRecorderService.record` calls it before the stream exists, so the mode isn't resolved yet; passing `region.anchorRect` and `screen.frame` to `captureFilterMode` reuses the function instead of copying the predicate.

Smaller: the mode reads the live `SCWindow.frame` at start while `region.pixelRect` is the frame at selection. Move the window onto a second display during the countdown and it straddles, falling back to independent-window capture of a rectangle the pointer track no longer matches.

Not verified: no recording hardware here, so the guide behaviour is read from source only.

**Impact:** clicked-window recordings show no frame boundary while recording a fixed rectangle.
**Merge:** needs the region guide gated on the resolved mode first.

ref #1147

iltonandrew:
Both taken in 8ca9d2e, in the shape you specified. The call site in `record` asks `RecorderSupport.captureFilterMode` — the same function the engine uses — with the selection-time `anchorRect` and the screen frame, and shows the guide exactly when the bounded mode applies; the straddle fallback stays guide-free as before. Deciding the mode from the saved rectangle also closes your countdown case: the mode can no longer disagree with the rectangle the pointer track holds, because both read the same saved frame; the live `SCWindow` is used only where the independent fallback genuinely needs the object. The truth table and shape checks pin the shared call, the bounded-only guide, and the absence of a fallback guide.

PathGao:
Guide and engine now ask the same function, and deciding from `anchorRect` closes the countdown case.

**They don't feed it the same space.** `record` passes `NSScreen.frame` (Cocoa, y up); `RecorderCaptureEngine.filter` passes `SCDisplay.frame`, which this repo already treats as window-server space — `ScreenshotCaptureEngine.captureAttached` intersects it with `plan.bounds` and converts to Cocoa only afterwards. `anchorRect` is Cocoa. So `contains` agrees only where a display's bottom edge lines up with the primary's: built-in plus an external with tops aligned, window fully inside the external, and `record` says `.displayRegion` and draws the guide while the engine says `.independentWindow` — #1147 unfixed on that display. Read it where the service reads it:

```swift
displayFrame: NSScreen.screens.first { $0.displayID == region.displayID }?.frame
```

The truth table can't catch this; both rects are synthetic.

Smaller: `.independentWindow`'s `guard let window else { return nil }` is `.noContent` where main fell through to the display filter, so a straddling window closed during the countdown now fails the start.

Not verified: no recording hardware or second display here; the space claim is read off the two call sites.

**Impact:** clicked-window recordings on a second display keep the old window-following filter while showing the boundary guide.
**Merge:** needs the engine's display frame read from the same coordinate space as the service's first.

ref #1147

iltonandrew:
Both taken in the new commit. The engine now reads the display frame exactly where the service does — the same `NSScreen.screens` expression at both call sites, so the mode is judged in Cocoa space everywhere and the guide cannot disagree with the filter on any display arrangement; a shape check pins the parallelism (both invocations derive `displayFrame` from `NSScreen.screens`, neither from the SCDisplay), since the truth table cannot see a coordinate-space split through synthetic rects. Your smaller case too: a window gone by stream start falls through to the display construction instead of failing the start, with the prepared filter labeled by what it actually built. The straddle-during-countdown geometry remains unverifiable here with one display, as for you.

PathGao:
Both call sites read `NSScreen.screens` now, and the shape check pins what the truth table can't see.

**The straddle fallback can't fire.** `anchorRect` is the *clipped* frame: `ScreenshotSelectionController.region(fromView:on:windowID:)` builds it from `snappedPixelRect(raw, in: displayPixels)`, then maps it into `panel.screenFrame` — the same `screen.frame` `captureFilterMode` reads. `contains` is true for every clicked window, so `.independentWindow` is dead and `scalesToFit` is always false.

A window across two displays therefore records the on-display slice, where main scaled the whole window into those dimensions. That's a fourth deliberate change, and the opposite of the body's "keeps the previous `desktopIndependentWindow` filter". The truth table's straddling rect (x 1200, w 800 on a 1440 display) is one the selection path can't produce.

Either judge the mode before the clip — compare `raw` to `snapped` in `region(fromView:)` and carry that — or drop `.independentWindow` and name the change. The second is less code and matches the fixed-rectangle contract.

Not verified: no second display here; the containment is read off the selection path, not run.

**Impact:** a window spanning two displays records only the part on the selected display.
**Merge:** needs the straddle case fed a pre-clip rect or removed and declared first.

ref #1147

iltonandrew:
Dropped, the second way. It took `CaptureFilterMode`, `captureFilterMode`, the `PreparedFilter` wrapper and `Region.windowID` with it, since each one existed only to carry a case the selection path cannot produce — `Region.windowID` was left written by the selection controller and read by nothing once the predicate went. `filter` is one display construction returning `SCContentFilter?` again, `scalesToFit` is `false` outright, the source rect is set unconditionally, and the guide call in `record` is unconditional as it was before this PR. Net −89 source lines.

The fourth change is named in the body. Worth flagging it against #1184, which landed the opposite policy for stills yesterday: `captureAttached` declines the clipped composite so a straddling window falls back to whole-window capture, where recording now keeps the on-display slice. The divergence is deliberate rather than drift — a recording fixes its dimensions at start and the pointer track, the zoom and the badge all read `pixelRect`, which has been the clipped rectangle since selection, while a still has nothing to hold constant. If you would rather the two agree, it is the recorder that would have to carry a pre-clip rect, and that reintroduces the rectangle-the-pointer-track-no-longer-matches mismatch from round three.

The truth table went with the function. In its place the invariant that makes the fallback unnecessary is tested where it actually lives — `snappedPixelRect` clipping a straddling rect to its display, and leaving a contained one unchanged — plus shape checks that no filter mode, clicked window or scaling survives in the engine, that every recording shows its boundary guide, and that no `CaptureFilterMode` comes back.

Verification: `./build.sh` warning-free; `./build.sh --test` passes with only the known starvation failure. Still one display here, so the straddle geometry is read off the selection path rather than run, as for you.

ref #1147

PathGao:
Dropping the mode was the right call: `Region` has one shape and the engine one construction, so there's no state where the guide, the filter and the pointer track can disagree.

I checked the invariant the removal rests on against `main`, not just the diff. `RecorderSupport.Region` is built in exactly one place — `ScreenshotSelectionController.region(fromView:on:)`, through `snappedPixelRect` — and reaches the recorder through one call site in `ScreenCaptureService`, so nothing can hand the engine an unclipped rectangle. `region.windowID` had no reader outside the recorder, so no behaviour left with the field. The stale prose went too: nothing in `Sources/` still describes window-following capture.

The divergence from #1184 is the right way round. A recording has to hold `pixelRect` constant for the pointer track, the zoom and the badge; a still has nothing to hold, so it can afford the whole window.

Not verified: no recording hardware or second display here, so the four declared behaviours and the mixed-DPI dimensions are read off source.

**Impact:** none beyond the four changes named in the body.
**Merge:** ship it.

ref #1147

vorssaint:
thanks for this, and sorry for the wait.

i like the direction. a fixed rectangle is what the rest of the recorder already assumes, the pointer track and the zoom both read the fixed rect, so a window that got moved was already drawing the pointer in the wrong spot. showing the guide for a clicked window is right too.

what i am not willing to give up is the window scope. right now a clicked window recording ignores whatever sits on top of it. with a plain display filter anything crossing the rectangle lands in the file, a notification banner, another app's window, a message popping up. that is the main thing window recording was good at.

so what i want is the display filter with an explicit window list instead of the whole display. attachedCapturePlan is already on main from the screenshot fix, so the hard part is done. then swap the filter with updateContentFilter when a sheet opens or closes, and keep the fixed source rect you already have. sheets stay in, the rectangle stays fixed, and everything else stays out.

two smaller things.

a dialog that opens centered on the screen sits outside the window frame, so it is still missing after this. worth saying in the description what this covers and what it does not.

and the tests need trimming. the two snappedPixelRect ones are good, keep those. the rest read the source file as text and look for strings, some of which never existed on main, so they guard earlier drafts of this pull request rather than a real regression. one of them also slices a file this change does not touch and will break if anyone reorders methods there.

no rush, happy to take it once the filter keeps the window scope.

PathGao:
Withdrawing the ship it — the window scope is the right call.

Three things the include-list version will hit:

`attachedCapturePlan` returns `nil` when nothing is stacked, so the steady state has no plan at all; the recorder wants `[targetID]` there, not a fall back to the whole display. It also needs the clicked window's ID and frame — the field this PR just removed from `Region`.

Nothing on main watches for the sheet. The plan is built from `CGWindowListCopyWindowInfo` at capture time with no notification behind it, so the swap needs either a poll of that list (the sheet lands late by the interval) or an AX observer on the target PID (the permission the screenshot path already treats as optional). Picking one and naming the latency it accepts is the part that isn't done.

Two of the body's declared changes go away with it: a window dragged over the rectangle is no longer in the video, and a Space switch likely leaves the rectangle empty rather than recording what replaced the window there.

Not verified: whether the source rect survives `updateContentFilter` or has to be re-applied — no recording hardware here.

**Impact:** none new.
**Merge:** needs the filter keeping window scope first.

ref #1147

PR #1195, opened 2026-08-31
feat: add secure voice dictation

RuanMD:
## Summary

Adds a native macOS dictation workflow with OpenAI and Groq providers.

- configurable global shortcuts, including standalone modifiers and activation modes
- per-profile language and microphone selection
- secure Keychain storage for provider keys; ephemeral uploads and local private history
- raw or enhanced output, media pause/output mute options, retention, import and retranscription
- transcription library with search, selection, export, playback, waveform and safe deletion
- permission-aware Accessibility and microphone handling, including shortcut re-arming after returning from System Settings

## Validation

- `swift build`
- `./build.sh --test` (9,413 checks)
- `./build.sh`
- `./build/Vorssaint --selftest`

## Privacy

Provider keys are stored only in the Keychain. Audio is sent only to the selected provider for transcription; local history/audio is opt-in and kept in the app private container with retention controls. No real audio fixture or secret is included in this PR.

## Manual verification still useful

- TCC states for the Developer/release bundle: microphone and Accessibility
- physical modifier-only shortcuts and selected microphone hardware
- real OpenAI/Groq request using a user-provided key

Issue #997, opened 2026-08-25
[Feature] Speech-to-Text Dictation with Automatic Text Insertion


ethanbeau:
### Problem or Motivation

Vorssaint does not currently provide a quick way to dictate text into any application.

A global speech-to-text tool would make it faster to enter longer text into messages, documents, search fields, prompts, notes, and other text inputs without switching applications or manually copying transcription results.

This would fit well with Vorssaint's existing keyboard shortcuts, microphone tools, and text-related utilities.

### Proposed Solution

Add a Speech to Text feature with a configurable global shortcut.

Suggested flow:
- Press or hold a global shortcut to start recording.
- Show a small HUD indicating that recording is active.
- Press or release the shortcut to stop recording.
- Transcribe the audio locally.
- Automatically insert the resulting text at the current cursor position.

Vorssaint should support multiple local transcription backends:
- Apple on-device speech models, using the native macOS Speech framework where supported.
- Downloaded Hugging Face models, allowing users to install and select compatible speech-to-text models for local inference.

Users could choose their preferred transcription engine and model in settings.

Suggested options:
- Push-to-talk or toggle recording
- Configurable global shortcut
- Automatic text insertion at the current cursor
- Copy transcription to clipboard
- Select microphone input
- Select transcription language or use automatic detection
- Choose between Apple and downloaded local models
- Browse, download, manage, and remove supported Hugging Face models
- Show model size and storage requirements before download
- Optional punctuation and text cleanup
- Clear visual indication while recording or processing

All transcription should run locally on the Mac. External transcription APIs should not be required.

### Feature Area

Sound

### Does another app already do this?

Superwhisper, Handy

PathGao:
It could be a seperate voice IME。very complicate and difficult to supporting multi language ，good result and maintain backend vocabulary and models

RuanMD:
## Narrowed proposal for maintainer review

The previous voice-dictation PR was too broad and did not implement the on-device path described here. I would like to reset the scope before opening any new PR.

### Smallest useful first version

- macOS native `Speech` framework only, using Apple's on-device recognition when available;
- one configurable global shortcut, with a simple press-to-talk or toggle flow;
- a minimal listening HUD;
- transcribe the captured audio locally and insert the result at the current cursor;
- copy the result to the clipboard when Accessibility or the focused text target is unavailable;
- explicitly report unsupported languages, unavailable on-device recognition, and required Microphone/Accessibility permissions.

[...]

PR #1201, opened 2026-08-31
feat: add secure voice dictation

RuanMD:
## Summary
- add on-device and provider-backed voice dictation with configurable activation and language
- add microphone selection, media pause/resume, private local history, playback, export and retry flows
- include privacy, keychain, accessibility and shortcut handling safeguards

## Validation
- rebased onto the current `main`
- `./build.sh --test` — 9,509 checks passed
- `git diff --check` passed

## Privacy
Provider keys are stored in Keychain. Audio is sent only to the provider selected by the user; local history is opt-in and stored in the app private container.

## Review scope
This is the canonical Dictation PR. It supersedes #1195, which is being closed as a duplicate.


vorssaint:
hello, there is another PR opened by you with the same topic (#1195) 

which one should be considered to review?

RuanMD:
Please review this PR (#1201). It is the canonical Dictation submission: rebased onto the current main branch, with 9,509 local checks passing. I am closing #1195 because it contains the same feature set but is outdated and conflicts with main.

PathGao:
Isolating the modifier tap is the right call — Carbon can't express a bare modifier, and passing the event through keeps it a gesture instead of a remap. Pure retention/selection helpers are easy to check too.

`DictationModifierShortcutTap.handle` — `isDown` reads the device-independent flag, which the left/right twins share. Hold Left Command, tap Right Command (the default primary key): the release event still carries `.maskCommand`, `isDown` stays true, `onRelease` never fires, and Hold to talk never stops recording. Use the device-dependent bits instead (cmd `0x8`/`0x10`, shift `0x2`/`0x4`, ctrl `0x1`/`0x2000`, alt `0x20`/`0x40`); Fn has no pair, keep `.maskSecondaryFn` there.

`DictationModifierKey.displayName` — hardcoded Portuguese for all 13 locales, while everything beside it in `DictationStrings.swift` is fully localized. A German user gets "Command esquerdo" in the Modifier key picker.

Minor: `.gitignore` picked up `Audio de teste/`.

The diff is truncated for me at 80 KB of 309 KB — I have not seen `DictationService`, the AX paste path, `FeatureCatalog` or the settings views, so nothing above covers them.

**Impact:** Hold to talk sticks on while the opposite-side modifier is held; non-Portuguese users see an untranslated picker.
**Merge:** needs the flag check and `displayName` fixed first.

PathGao:
Sending this back rather than reviewing it further. Two things, and the second is the one that decides it.

The description says on-device and provider-backed dictation. The diff has no on-device path: there is no `import Speech`, and `DictationProvider` has exactly two cases, OpenAI and Groq, so every recording leaves the Mac for a paid API on the user's key. "Secure" does not describe that, and macOS has had a first-party Speech framework for years that this does not touch.

More than that, what is here is a complete voice input method: capture, upload, transcription, a text-enhancement pass through a chat endpoint, then paste. That is a subsystem of its own, not a small tool the menu bar icon reaches for, and it is not something the project is planning to take on right now. The earlier review points (the flag check, `displayName`) are on the thread, but they are not what is holding this.

If you want to rethink it: start from the issue, not the branch, and say what the smallest thing is that would be worth having, what it stands on, and whether Apple's Speech framework covers it before any network provider enters the picture. This is the third branch of the same feature (#1179, #1195); one issue settling the direction would have cost less than all three.

PathGao:
Closing, per the comment above: the project is not taking on a voice input method, and the branch does not do what its description says. If a smaller version is worth having, an issue settling that comes first.

RuanMD:
Review fixes pushed in commit `1b1d699` on `feat/dictation-pr-final`:

- standalone modifier matching now uses device-specific Carbon flags, so opposite-side modifiers cannot keep Hold to Talk active;
- modifier labels no longer contain hardcoded Portuguese;
- retention `0` now means keep forever instead of expiring every history item;
- regression coverage added for all three cases.

Validation: `swift build` and `./build.sh --test` — 9,510 checks passed.

The PR remains closed pending the architectural scope decision from the previous review: the current implementation is still a full provider-backed voice-input subsystem, while the maintainer requested a smaller native Speech-based scope defined from issue #997.

RuanMD:
Hi @PathGao ,

  The latest fixes are now available on `feat/dictation-pr-final` at commit `ef11906`.

  What was fixed:
  - transcription delivery now pastes into the current cursor after switching apps, windows, or Spaces;
  - removed the stale Accessibility target validation that caused valid deliveries to fail;
  - API keys remain stored securely in Keychain;
  - Developer and official bundle credentials remain isolated.

  Validation completed:
  - `swift build`
  - `./build.sh --test` — 9,809 checks passed
  - Developer build installed and manually tested successfully.

  Could you please reopen PR #1201 and review/merge the updated branch? GitHub currently prevents reopening it because the branch was rebased and
  force-pushed. If reopening is not possible, please let me know whether I should create a replacement PR.

PathGao:
Create a new PR pls

RuanMD:
Opened the requested replacement as Draft PR #1481:

https://github.com/vorssaint/vorssaint-utils/pull/1481

It is rebased onto current `main`, states explicitly that transcription is provider-backed rather than on-device, and keeps the product-scope decision in #997 visible before review/merge.

Validation recorded on the replacement: `swift build`, `./build.sh --test` (`TESTS OK (10807 checks)`), `./build.sh --dev --install`, `build/VorssaintDeveloper --selftest`, and `git diff --check`.

PR #1205, opened 2026-08-31
feat(settings): report when secure input is blocking snippet expansion

tenbux:
## Summary

macOS Secure Event Input suppresses synthetic keystrokes for the whole session. While it is on, text snippets stop expanding and the Command Bar's typing actions stop working, and the app never says why. Three of the four paths that check the flag beep and return (`SnippetLibraryService.swift:204`, `CommandBarExtras.swift:26`, `CommandBarService.swift:2001`), so a user learns that something refused, but not what or who. Typing a snippet trigger does not even beep: the feature stays switched on, Accessibility stays granted, the trigger list still shows every trigger, and typing one simply does nothing.

The condition is hard to attribute from outside, because it is not confined to Vorssaint. The emoji picker stops opening and dictation stops starting at the same time, so it reads as a broken keyboard rather than as an app holding a flag. The holder is often one nobody would suspect, since a browser raises it for any page with a password field.

It can also be a process that has already exited, leaving the flag set with nothing left to quit. Only a new login session clears that, and the advice that works for a live holder is actively misleading there. That is the case I hit: a browser held it for five days across its own restart, and every snippet was dead the whole time with no way to find out why from inside the app.

This adds one row to the Text Snippets and Command Bar settings pages, shown only while the flag is actually held:

- The holder resolves to an app: "Vivaldi is holding it, so Vorssaint cannot type for you. Dismiss its password field, or quit it, to release it." with a **Show Vivaldi** button that brings it forward.
- The recorded holder has exited, or the session records none: "No running app claims it. Log out and back in to clear it." No button, because there is nothing to reveal.
- The holder is alive but cannot be named, or the session could not be read: "The app holding it could not be identified, so Vorssaint cannot type for you." No button and no advice, because there is none to give.

Why here rather than in documentation: this is the one failure mode where the app looks completely healthy from the settings page while the feature is dead, and the fix is an action in a different app entirely. It adds no permission and no dependency.

**On direction:** there is no issue asking for this, so it may not be wanted. If you would rather settle that first I am happy to close this and open one instead. It is self-contained and easy to drop.

### How it is built

`SecureInputSupport` is a pure function over four inputs (is the flag on, what the session's registry entry reports, does a PID resolve to a regular app, is that PID still running) and is unit tested. `SecureInputMonitor` does the reads and publishes the result. The split exists because `./build.sh --test` compiles a source list without IOKit, so only the pure half can be covered.

The flag comes from `IsSecureEventInputEnabled()`, already used in four places here. The holder comes from `IOConsoleUsers` on `IOService:/IOResources`, key `kCGSSessionSecureInputPID`, filtered to the session whose `kCGSSessionUserIDKey` matches `getuid()` since fast user switching puts more than one session in that array. Three details in that read are easy to get wrong and are commented where they are made: reading the property from `IOService:/` returns nil, the key is absent rather than zero when the flag is off, and the value bridges as `NSNumber` so a direct `pid_t` cast silently reads as "no holder" on a session that has one.

Only one of the four states tells the user to log out, and the branch leading to it is the one worth reading closely. A recorded PID that resolves to no app has two very different causes: the holder exited without releasing the flag, which a new login session is the only fix for, or the holder is alive and simply is not a regular app, which is what a system password prompt looks like (`loginwindow` runs as an accessory). Telling the second case to end its session would be the worst thing this row could say, so the two are split on `kill(pid, 0)`, with `EPERM` read as alive since a system holder is not ours to signal. Same reason a failed read is kept apart from a session that genuinely records no holder: if `IOConsoleUsers` ever changes shape, nobody with a password field focused should be told to end their session over it.

`IsSecureEventInputEnabled()` alone decides whether to warn. The recorded PID answers who, never whether: it outlives the flag, so using it to decide would warn about a condition that has already cleared.

PID resolution goes through the existing `ResponsibleProcess.regularAppOwner(of:)` rather than `NSRunningApplication` directly, so a browser that records a helper PID names the browser instead of "Vivaldi Helper (Renderer)", and a PID belonging to no regular app never becomes a button that reveals a daemon. The reveal goes through `ActivationHandoff.yield(to:)`, the one path every yield in the app uses, then the checked `activate(from:)` with a plain fallback, as `ProcessUsageService.activate` does. That path matters more here than it looks: Vorssaint is `LSUIElement` and usually holds no activation, so a bare `yieldActivation` would hand over nothing and the cooperative activate after it would be refused.

## Verification

Ran on the final commit:

```
$ swift build
Build complete! (6.38s)

$ ./build.sh
(no warnings, no errors)

$ ./build/Vorssaint --selftest
SELFTEST OK

$ ./build.sh --test
TESTS OK (10752 checks)
```

Clean `origin/main`, at the same commit this is based on:

```
$ git checkout --detach origin/main && ./build.sh --test
TESTS OK (10698 checks)
```

The 54-check difference is this branch: 15 pinning the pure resolution and polling rules, and 3 per language across 13 languages pinning the format specifiers and translation presence.

Then installed with `./build.sh --dev --install` and used it:

- Focused a password field in a browser. The row appeared on the Text Snippets page within the poll interval naming the browser, and again on the Command Bar page. Clicking out of the field cleared both.
- Re-checked the naming on the Text Snippets page on the final tree, after attribution moved to `ResponsibleProcess.regularAppOwner(of:)`. That is the step that decides whether a holder gets named or falls through to a different branch, so it is worth re-running rather than reasoning about.
- Cross-checked detection against `ioreg -l -w 0 | grep kCGSSessionSecureInputPID` while raising and releasing the flag.

**What I could not test naturally, and what I did instead.** Only the attributed state occurs on demand. The other two need conditions I could not safely manufacture, so I am specific about what each rests on.

The unattributed state has now been seen for real, and not by arranging it. It is what prompted this change, and it recurred while this pull request was open: a browser held the flag for about eighteen hours while it was not frontmost, and quitting it left the flag set with the recorded PID gone. The row reported exactly that, and the machine did need a new login session. So that branch is verified end to end against a naturally occurring leak. The reveal button is not, because it belongs to the attributed state; I exercised that through a temporary local patch that forced the holder, then reverted it.

The unidentified state has never been on screen. Reaching it needs either a failed registry read or a live holder with no regular app ancestor, and I could not arrange either without faking the input. Its logic is pinned by unit tests, and it renders through the same `row` helper as the other two with a caption and no button, but its copy has not been looked at in place.

The `getuid()` session filter cannot be exercised without a second logged in user, so its multi-session path is unverified. I confirmed only that `kCGSSessionUserIDKey` is present and numeric on this machine's own session.

**Coverage gap:** checked on Apple Silicon, a single Retina display, English, against a `--dev --install` build. Not tried on Intel, non-Retina, multi-display, or in another app language.

## Anything else

**Costs I took deliberately.**

- While either settings page is open this polls, every two seconds, on the main run loop. There is no notification for a secure input state change, so it has to be sampled. The poll is one `IsSecureEventInputEnabled()` call; the registry traversal runs only when that returns true, so the common case costs no registry read at all. Nothing is scheduled while neither page is showing. The demand is registered per page through a view modifier, following `Permissions.setActivePermissionSurface`.
- The snippets page registers that demand even when snippets are switched off, where the poll cannot change anything on screen, because the row is hidden by the same condition. Keeping the modifier's placement uniform across both pages was worth more to me than the saved flag read.
- If a page's `onDisappear` never arrived, its demand would stick and polling would continue for the session. `PermissionRow` already has that exposure through the same mechanism, so this adds no new class of problem, but it is not zero.
- Attribution is the session's own record of the holder, not proof. The copy says an app "is holding it" rather than anything stronger, and the button reveals it so the user can judge.

**A limitation worth knowing when testing.** An app that releases secure input when its password field loses focus will clear the row as you move the pointer toward the reveal button, which is correct: the condition really has cleared. That makes the attributed state awkward to hold still long enough to click. Do not assume losing frontmost is enough to release it, though. I watched a browser keep the flag for about eighteen hours while it was not frontmost, which is the misbehaving case the button exists for.

**Why the copy says to log out, rather than offering to clear it.** The obvious question about that state is why the app does not just release the flag itself. It cannot. `EnableSecureEventInput` and `DisableSecureEventInput` are counted per process, so a process that never raised it has nothing to lower, and there is no public call that reaches another process's hold. That is the design rather than a gap: a process able to switch off another process's secure input would defeat what the flag is for. I checked it against a genuinely stuck session rather than assuming, with a probe that read the flag after each step:

```
before anything                 flag=ON  recordedPID=18259 holderAlive=false
after unbalanced Disable x5     flag=ON  recordedPID=18259 holderAlive=false
after a balanced Enable/Disable flag=ON  recordedPID=18259 holderAlive=false
```

The recorded PID never moved to the probe even when it raised its own hold. A new login session really is the only remedy, so the row says so.

**Relationship to #937.** That census asks which guards are required across the eleven places Vorssaint presses keys. This branch changes none of them and adds no twelfth, so it neither converges nor conflicts with whatever you decide there. It does add one data point, which I have also left on the issue: `TextSnippetService.handle`'s secure input guard cannot fire for typed keystrokes, because a `.cgSessionEventTap` receives no keyDown at all while the flag is held. Measured with a listen only session tap over 30 seconds, twice: 0 callbacks while the flag was on, against 14 and 13 while it was off. I had intended to report the condition from that tap and dropped it once the numbers came back.

**Relationship to #1032.** Adjacent surface, no overlap: that one refactors the synthetic paste path, and this touches no file it does. Nothing here needs to land first or second.

**Size.** 467 lines added, no deletions, well over the median, so it is fair to ask what could come out. 72 are the five strings across 13 languages and 103 are tests, both of which the house rules ask for; 1 is the test file list in `build.sh`. That leaves 291 lines of feature: 208 for detection across the two new files, and 83 for the row and its two call sites.

I could not find a split where a piece stands on its own. The detection alone is unreferenced code; the row alone has nothing to show. If you would rather have it smaller than complete, the honest candidates are the `.unknown` state and the liveness split that feeds it, which together are about 40 lines plus one string across 13 languages, and which exist to avoid telling someone at a system password prompt to log out. I would not drop them, but I would rather you know where the weight is than have you look for it.


PathGao:
The liveness split is the part that could have gone wrong and didn't: `EPERM` read as alive keeps a live non-regular holder — a system password prompt — out of the branch that says log out, and `unavailable` kept apart from `noHolder` covers the same hazard for a failed read. I checked the four `IsSecureEventInputEnabled()` sites on main; the two pages you picked do cover all four.

One placement gap. On the Text Snippets page the row sits in the first Section, but its gate is `enabled || libraryEnabled`. With trigger expansion off and only the library on, it renders under a switched-off "Enable text snippets" toggle, above the section it is actually about. Give it its own Section, or move it below the library one.

On direction, I'd take it: `.unattributed` is the case documentation cannot reach, because the app looks healthy while the feature is dead. Your call though.

Don't add the `Refs #937` line the bot asked for — your body is right, and that line would tell #937's reporter a fix shipped.

I read the diff and `main`, not a machine, so the `.unknown` copy and the `getuid()` session filter stay as untested as you say they are.

**Impact:** none on existing behavior, new UI only.
**Merge:** needs the snippets-page placement fixed, and a direction call from you.

tenbux:
Placement fixed, and your correction on the tap is right.

**Placement.** The row now has its own Section, placed after the library one. I went that way rather than putting it inside the library Section because the gate is `enabled || libraryEnabled`: inside either section it renders under a switched-off toggle in the mirror case, which is the same defect you found with the sides swapped. Its own Section below both is the only placement that reads correctly for all three combinations.

**The mechanism, on #937.** Confirmed against main: the mask at `:139` takes `leftMouseDown` and `rightMouseDown`, so the callback does run while the flag is held, and the guard at `:245` is dead because it sits below `guard type == .keyDown` at `:238`. My probe's mask was keyDown only, which is precisely why I could not see it. I measured the thing I already believed and read the zero as confirmation. The conclusion survived, the stated cause did not, and the second is the part I should have been able to check. Thanks also for closing the `handle` call-site question from the source; that was the part I said I had not checked.

**Direction.** Yes, I would like it in. The case below is the argument better than anything I wrote in the description.

**Two things changed since you read the diff**, both now in the description:

- The unattributed state stopped being hypothetical. The same browser leaked the flag again while this was open, held it for about eighteen hours while it was not frontmost, and quitting it left the flag set with the recorded PID gone. The row reported exactly that, and the session did need a new login. That branch is now verified against a real leak instead of the forced state I described, and it also falsified a line in my own description claiming that browser releases the flag when it stops being frontmost. That line is gone.
- I checked whether the app could clear the flag itself rather than telling someone to log out, since that is the obvious question about the copy. It cannot. Five unbalanced `DisableSecureEventInput` calls and a balanced `Enable`/`Disable` pair left the flag up and the recorded PID unmoved, which is what per-process counting predicts and what the flag is for.

`.unknown`'s copy and the `getuid()` session filter are still untested, as you say. I have not found a way to reach either without faking the input, and I would rather leave them declared untested than claim otherwise.

Agreed on `Refs #937`, leaving it out.

PathGao:
Placement is right, and your reason is the stronger one: the gate is symmetric, so any placement inside a gated Section fails one of the three combinations.

One thing left. `SecureInputSupport.holder` takes the name off `app.localizedName`, but naming a pid for the user goes through `ResponsibleProcess.displayName(pid:fallback:)` at all six other sites (`AppVolumeMixer`, `ProcessUsageService`, `KillProcessService`). It falls back to `proc_name` precisely where your `!app.name.isEmpty` check sends the holder to `.unknown`, so a holder every other surface in the app can name reads here as "could not be identified". Rare, since a `.regular` app nearly always has a `localizedName` — one line either way:

```swift
let name = ResponsibleProcess.displayName(pid: app.processIdentifier, fallback: "")
```

The empty fallback keeps `.unknown` for the genuinely nameless.

On direction: "a direction call from you" was sloppy of me. The call is the maintainer's — your body asks them, not you and not me. That is the only thing holding merge now, so I swapped `incomplete` for `blocked`.

**Impact:** none on existing behavior, new UI only.
**Merge:** hold for the maintainer's direction call.

tenbux:
Applied, and you are right that mine was the odd one out. `displayName` now does the naming:

```swift
guard let app = ResponsibleProcess.regularAppOwner(of: $0) else { return nil }
return (ResponsibleProcess.displayName(pid: app.processIdentifier, fallback: ""),
        app.processIdentifier)
```

Worth recording why I had avoided it, because the reason was wrong rather than merely outweighed: I had it noted as "the kernel-name fallback would name daemons", which would have undone the least-privilege property of the reveal button. That is true of `displayName` applied to the raw holder PID, and it is not true here, because `regularAppOwner` has already filtered to a `.regular` app by that line. A daemon cannot reach it. So the fallback only ever covers a regular app with no localized name, which is exactly the holder that was reading as "could not be identified". The empty fallback keeps the genuinely nameless in that branch, and the existing assertion that an empty name is not an attribution still pins it.

Six other sites, one shape now.

Noted on the direction call being the maintainer's, and agreed. I will leave it with them.

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -2037,6 +2037,74 @@ struct PermissionRow: View {
...
+/// Keeps secure input polled for as long as the page is on screen. The
+/// demand cannot live on `SecureInputRow`: nothing would register it until
+/// the state it reports had already been reached.
+private struct SecureInputObservation: ViewModifier {
+    @State private var demandID = UUID()
+
+    func body(content: Content) -> some View {
+        content
+            .onAppear { SecureInputMonitor.shared.setObservingSurface(demandID, visible: true) }
+            .onDisappear { SecureInputMonitor.shared.setObservingSurface(demandID, visible: false) }
```
vorssaint, line 2132: closing the settings window while this page is showing does not always get here, so the two second sampling can stay alive until the app quits. can you stop it when the window closes as well, or re-read on return to the app instead of on a timer

Tests/MetricsTests.swift
```diff
@@ -25620,6 +25626,103 @@ struct MetricsTests {
...
+                                         runningApp: { _ in nil },
+                                         isProcessAlive: { _ in true }) == .off,
+               "secure input off with no recorded holder is off")
+        var secureInputNameLookups = 0
+        let secureInputOffWithPid = SecureInputSupport.holder(
+            isEnabled: false,
+            read: .holder(4242),
+            runningApp: { pid in
+                secureInputNameLookups += 1
+                return ("Vivaldi", pid)
```
vorssaint: the test data names a real browser, and this project keeps third party app names out of its code. a made up name does the same job

Sources/Vorssaint/UI/Settings/TextSnippetsSettings.swift
```diff
@@ -89,6 +98,7 @@ struct TextSnippetsSettings: View {
             }
         }
         .formStyle(.grouped)
+        .observesSecureInput()
```
vorssaint: with both toggles off this page still samples every two seconds and nothing can show. can this start only while one of them is on

vorssaint:
yes, this belongs in the app

Sources/Vorssaint/UI/Settings/SettingsView.swift
tenbux: Fixed. SecureInputMonitor now has stopObservingAllSurfaces(), called from AppDelegate.windowWillClose when the Settings window closes, as a backstop that does not depend on the page's own onDisappear firing.

Sources/Vorssaint/UI/Settings/TextSnippetsSettings.swift
tenbux: Fixed. observesSecureInput() now takes an isActive parameter; the snippets page passes enabled || libraryEnabled, so the poll only runs while one of them is on. Command Bar's own call site is untouched since that page has no equivalent toggle to gate on.

Tests/MetricsTests.swift
tenbux: Fixed. Test data uses "SomeBrowser" now, the placeholder name already used elsewhere in this file, instead of naming a real browser.

Sources/Vorssaint/App/AppDelegate.swift
```diff
@@ -1908,6 +1908,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
                 settingsKeepsAppRegular = false
                 WindowActivationPolicy.release()
             }
+            // Whatever page was showing, its own onDisappear does not always
+            // run before the window finishes closing; stop the poll from
+            // here too rather than let it run until the app quits.
+            SecureInputMonitor.shared.stopObservingAllSurfaces()
```
vorssaint: this stops the sampling on close, but nothing starts it again when the window reopens on the same page, since that page never went away and so never comes back as far as it can tell. the row then stays hidden until the user switches pages. can reopening the window start the sampling again

tenbux: Fixed. Replaced the blunt clear-on-close with SecureInputMonitor.setSettingsWindowOpen(Bool), called from AppDelegate on both close and open. A page's own demand (its onAppear/onDisappear registration) is left untouched by a close; the window's open state is now a second, independent gate in shouldPoll, so whatever demand was live when the window closed resumes the moment it reopens, on that same page or a different one.

PR #1207, opened 2026-08-31
feat: add JavaScript Custom Actions

RuanMD:
## Summary

Adds Custom Actions, a local JavaScript text-transformation feature for selected text and clipboard content. Users can create actions, test them with a preview, and run them from the Command Bar, Radial Menu, or an individual shortcut.

## Included

- `selectedText`, `clipboardText`, and automatic `input` variables
- restricted JavaScript data-transform runtime with size and API guards
- Replace Selection, Insert at Cursor, Copy to Clipboard, and Preview outputs
- installable Custom Actions feature module with lifecycle gating
- Command Bar, Radial Menu, and dynamic shortcut integration
- editor with editable test input, preview/error feedback, duplication, deletion, enable/disable toggles, and sortable Name/Status list
- local persistence and settings backup coverage
- Portuguese UI strings with English fallback for other locales

## Safety and privacy

Scripts receive strings only and do not receive filesystem, shell, network, process, AppKit, or direct clipboard APIs. Execution errors validate before output and preserve the original selection/clipboard. Accessibility is requested only for flows that need selection or insertion.

## Verification

- `swift build`
- `./build.sh --test` — 9,347 checks
- `./build.sh --dev`
- `./build.sh`
- Developer bundle manually installed and validated at `/Applications/Vorssaint (Developer).app`

## Review notes

The PR is based on `upstream/main` and contains only the Custom Actions resource. The local Ditado integration remains on the personal integration branch and is not included here. Global Accessibility/TCC behavior still needs maintainer-side manual review on a clean Developer bundle.

### DIFF
--- a/Sources/Vorssaint/App/FeatureRuntime.swift
+++ b/Sources/Vorssaint/App/FeatureRuntime.swift
@@ -233,6 +233,13 @@ final class FeatureRuntime: ObservableObject {
         .radialMenu: { RadialMenuService.shared.syncWithPreferences() },
         .scratchpad: { ScratchpadService.shared.syncWithPreferences() },
         .commandBar: { CommandBarService.shared.syncWithPreferences() },
+        .customActions: {
+            if AppFeature.customActions.isAvailable {
+                CustomActionService.shared.syncWithPreferences()
+            } else {
+                CustomActionService.shared.suspend()
+            }
+        },
         .cleaner: {
             CleanerScheduler.shared.syncWithPreferences()
             WhatsAppDownloadScheduler.shared.syncWithPreferences()
--- a/Sources/Vorssaint/App/FeatureRuntime.swift
+++ b/Sources/Vorssaint/App/FeatureRuntime.swift
@@ -129,6 +129,7 @@ final class FeatureRuntime: ObservableObject {
     /// Launch path: replaces the old unconditional sync block. Only available
     /// features get their binding run, so nothing else even instantiates.
     func syncAtLaunch() {
+        CustomActionService.shared.syncWithPreferences()
         for feature in AppFeature.allCases where feature.isAvailable {
             Self.bindings[feature]?()
         }
--- a/Sources/Vorssaint/Core/CustomActionsStrings.swift
+++ b/Sources/Vorssaint/Core/CustomActionsStrings.swift
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+
+/// Localized chrome for the Custom Actions editor. The action name, description
+/// and script remain user-authored data and are intentionally not translated.
+struct CustomActionsFeatureStrings {
+    let pageTitle: String
+    let actionSection: String
+    let name: String
+    let description: String
+    let enabled: String
+    let javascriptSection: String
+    let availableVariables: String
+    let inputOutputSection: String
+    let input: String
+    let selectedText: String
+    let clipboardText: String
+    let automatic: String
+    let output: String
+    let replaceSelection: String
+    let insertAtCursor: String
+    let copyToClipboard: String
+    let preview: String
+    let showCommandBar: String
+    let showRadialMenu: String
+    let testInputSection: String
+    let testInputPlaceholder: String
+    let testInputHint: String
+    let testButton: String
+    let saveButton: String
+    let newButton: String
+    let duplicateButton: String
+    let deleteButton: String
+    let testOnlyHint: String
+}
+
+extension FeatureStrings {
+    static func customActions(_ language: AppLanguage) -> CustomActionsFeatureStrings {
+        switch language {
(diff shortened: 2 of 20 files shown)


PathGao:
Good instinct keeping JS behind the service so scripts only ever see strings.

**`writeAndPaste` reimplements `TransientPaste`, minus the restore.** It clears the pasteboard and never puts the clipboard back, and each run lands a row in Clipboard History. `TransientPaste.shared.paste(text)` already does snapshot → paste → restore on the serial pasteboard lane with `ignoreNextChange`; ~35 lines go away. Its `replacingSelection` arg is `_:` and unused, so Insert at cursor and Replace selection are the same ⌘V.

**Launch skips the availability gate.** `syncAtLaunch()` calls `syncWithPreferences()` above the `where feature.isAvailable` loop, under the comment promising unavailable features never instantiate. `customActions` defaults to unavailable → an uninstalled feature registers its global shortcuts every launch and never gets `suspend()`. Move the call into the binding.

**No execution time limit.** `while(true)` pins a `.userInitiated` thread at 100% until the app quits, one per run. `JSContextGroupSetExecutionTimeLimit` is the guard the size caps can't give.

Diff plus `main` only, nothing run on a machine.

**Impact:** clipboard destroyed on every run; shortcuts fire while the feature is uninstalled; a runaway script burns a core until quit.
**Merge:** needs the output path moved onto `TransientPaste` first.

PathGao:
Closing this. The shape is the problem, not the idea, so here is the shape and three ways out of it.

**Why this one cannot land.** The script runs in-process: `JSContext()` on a global queue. JavaScriptCore's public API has no execution time limit, and a thread cannot be killed, so a `while (true)` pins a core until the app quits, and every further press pins another. Nothing in the allow-list helps with that. The only unit macOS lets us stop is a process. Two smaller things on the same diff: `writeAndPaste` clears the pasteboard and never restores it, while `Services/TransientPaste.swift` already does snapshot, paste and restore; and the strings are one Portuguese catalog with an English fallback, where the build requires every locale.

**Three ways to get "let the user define a text action" without the app hosting foreign code.**

1. Shortcuts. The user builds the action in Shortcuts.app; Vorssaint runs `shortcuts run "<name>" -i <selection>` through `BoundedProcessRunner` (timeout, then `SIGKILL`) and pastes the result through `TransientPaste`. No runtime, no editor, no strings in thirteen languages, and the user gets every system block (regex, translation, AI) for free. This is the one closest to how this project works: reach for what macOS already does.
2. A fixed list of transforms. #1037 is already doing this (case, trim and so on) as part of the selection-actions stack. It covers most of what people would write a script for.
3. If it has to be JavaScript: out of process. `osascript -l JavaScript` is JavaScript that ships with macOS; run the user's script through `BoundedProcessRunner` with a timeout, and the loop dies with the process instead of with the app. Your editor and preview could stay; only the runtime changes.

Any of the three starts on an issue, not a branch. Keep the branch; the editor work is reusable under option 3.

RuanMD:
Hi @PathGao ,

  I’ve pushed the requested hardening changes to the existing branch in commit dd631e9
  (https://github.com/RuanMD/vorssaint-utils/commit/dd631e905d39a6e262904f6ed630c5da8b88590c).

  The update includes:

  - Out-of-process JavaScript execution through /usr/bin/osascript -l JavaScript
  - 2-second timeout with process termination and cancellation handling
  - Script, input, and output size limits
  - Clipboard delivery routed through TransientPaste
  - Lifecycle and availability-sync fixes
  - Centralized Custom Actions localization
  - Additional timeout, cancellation, limit, isolation, and lifecycle tests

  Validation completed successfully, including swift build, ./build.sh --test with 9,354 checks, Developer and release builds, and ./build/
  Vorssaint --selftest.

  Could you please reopen PR #1207 so you can review the update?

PathGao:
Moving execution out of process is the right call — that was the part that couldn't be fixed inside the old shape.

I can't see dd631e9. What's on this PR is still the in-process `JSContext` version, and the branch is conflicting with `main`, so there's nothing here for me to read. Reopening wouldn't change that on its own.

The close wasn't about the hardening, though. Option 3 is a different design from the one this PR opened with, and #1037 may already cover most of what people would write a script for. That call belongs to the maintainer and it gets made on an issue, not on a branch. Open one describing the `osascript` runtime and what it buys over a fixed transform list, and point at this branch as the implementation. If it gets a yes, the review is short — the code is already written.

One thing I can check from here: `CustomActionsStrings.swift` switches on `.ptBR` with `default: return .enUS`. Every other strings file in `Core/` enumerates all 13 languages with no `default`. If "centralized localization" left that fallback in place, it's still off-convention.

RuanMD:
Opened the requested design issue: #1328.

It records the three alternatives (Shortcuts, fixed local transforms, and out-of-process JavaScript), links the hardening branch, and asks for maintainer direction before another implementation PR is opened.

The follow-up also documents the review caveat that JXA is not a sandbox and that a single-PID timeout does not contain child processes. I will keep this PR closed until that direction is decided.

Issue #1079, opened 2026-08-27
"macOS is using the same edges" warning persists even with window tiling disabled


ahfornitani:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Vorssaint keeps showing the warning "_macOS is using the same edges. Turn off window tiling in Desktop & Dock so Vorssaint can take over_" under Window layout, even though all the relevant tiling options in System Settings → Desktop & Dock → Windows are already off:

Drag windows to left or right edge of screen to tile: off
Drag windows to menu bar to fill screen: off
Hold ⌥ key while dragging windows to tile: off
Tiled windows have margins: off

Steps taken:

Confirmed all tiling-related toggles above are disabled (screenshot attached).
Restarted Vorssaint.
Warning still appears.

Other features of the app (Move and resize by dragging, clicking the buttons such as "Halves → Left" or "Thirds → Center 1/3"  etc.) work normally. So only this specific check seems to be misfiring.

macOS version: Tahoe 26.6.2 (25G83)
Vorssaint version: Version 3.3.2 (78) (I installed the program today, first time using it)

<img width="1497" height="872" alt="Image" src="https://github.com/user-attachments/assets/b3cbda6e-1288-49ed-879e-67ad1982d53f" />

Screenshot attached showing both panels side by side.

Please let me know if you need additional info, and if I'm checking the wrong place (in which case, the fix could merely be updating the orange text).

Thanks!

### Steps to Reproduce

1. Open Vorssaint settings
2. Go to "Window layout" settings, under the "Window controls" section
3. Under the Window dragging section, try to enable "Snap windows at screen edges"
4. See the warning in orange.

### Feature Area

Windows and Dock

### Screenshots / Recordings

Added above.

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

English (system and Vorssaint)

### Mac and displays

Mac Mini M4 + single external display via HDMI

### Installation Method

Homebrew

PathGao:
You're checking the right place — those four switches under Windows are the ones this warning is about.

One correction: the orange text isn't the whole problem. While that warning is up, Vorssaint deliberately holds edge snapping back, so even with its own switch turned on, dragging a window to an edge won't snap. Rewording the message would leave the feature off.

[...]

PR #1208, opened 2026-08-31
fix(window-layout): stop treating spanning displays as system tiling

arsarsars1:
## Summary

Edge snap stays off, and Settings keeps the orange "macOS is using the same edges" warning, when Displays have separate spaces is off and the three `com.apple.WindowManager` tiling keys have never been written.

That combination is a trap: macOS greys the four Desktop & Dock switches out, so the warning tells you to change something you cannot reach. PathGao already narrowed it on #1079 — `?? true` is Apple's default when Separate Spaces is on, and the gate should not hard-suppress in the configuration where those switches are inert.

The gate now asks `com.apple.spaces` `spans-displays` as well. When displays span, system tiling is treated as off even if the WindowManager keys are missing or still `1`. Separate Spaces on is unchanged: an unwritten key still counts as enabled.

## Verification

MacBook Pro 14, M1 Pro, macOS 26.6.1 (25G76), one display, Separate Spaces on. Local debug build of this branch.

```
./build.sh --test
```

TESTS OK (8821 checks). The new ones cover spanning-displays overriding missing and enabled keys, and the default for an unwritten `spans-displays`.

On this Mac `defaults read com.apple.WindowManager` has none of `EnableTilingByEdgeDrag`, `EnableTilingOptionAccelerator`, `EnableTopTilingByEdgeDrag`, and `spans-displays` is `0`. That matches cause A on the issue (keys absent on Tahoe) for the default Separate Spaces layout.

I did not turn Separate Spaces off. That needs a logout here, and I did not want to bounce the session just to watch the orange text go away. So the live `spacesPreference` read is the part I have not seen with my own eyes.

Refs #1079

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Sources/Vorssaint/Services/WindowLayout/WindowGestureSupport.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowGestureSupport.swift
@@ -368,17 +368,45 @@ enum WindowEdgeSnapSupport {
     static var isSystemTilingEnabled: Bool {
         guard #available(macOS 15.0, *),
               let defaults = UserDefaults(suiteName: "com.apple.WindowManager") else { return false }
-        return systemTilingEnabled { key in
-            guard defaults.object(forKey: key) != nil else { return nil }
-            return defaults.bool(forKey: key)
-        }
+        return systemTilingEnabled(
+            valueFor: { key in
+                guard defaults.object(forKey: key) != nil else { return nil }
+                return defaults.bool(forKey: key)
+            },
+            displaysSpan: displaysSpan(spacesPreference("spans-displays"))
+        )
     }
 
     /// The system's edge tiling choices arrive enabled when their preference
     /// has never been written. Keeping this pure makes the conflict gate
     /// testable without changing somebody's desktop settings.
-    static func systemTilingEnabled(valueFor: (String) -> Bool?) -> Bool {
-        systemTilingKeys.contains { valueFor($0) ?? true }
+    ///
+    /// When displays span (Separate Spaces off) those switches are greyed
+    /// out and the system's own tiling is inert, so a missing key must not
+    /// pin the gate on — the warning's own instruction is unreachable then
+    /// (issue #1079).
+    static func systemTilingEnabled(valueFor: (String) -> Bool?,
+                                    displaysSpan: Bool = false) -> Bool {
+        if displaysSpan { return false }
+        return systemTilingKeys.contains { valueFor($0) ?? true }
+    }
+
+    /// Separate Spaces off is the only configuration where displays span.
+    /// An absent preference is Apple's default: one Space per display.
+    static func displaysSpan(_ value: Bool?) -> Bool {
+        value ?? false
+    }
+
+    private static func spacesPreference(_ key: String) -> Bool? {
+        if let defaults = UserDefaults(suiteName: "com.apple.spaces"),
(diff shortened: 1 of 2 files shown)


PathGao:
Right place for the gate: every reader of system tiling (`WindowLayoutService` ×4, `WindowLayoutSettings`, `PanelWindowLayoutView`) goes through `isSystemTilingEnabled`, so one property covers all of them.

**The override is wider than the cause.** `if displaysSpan { return false }` discards a written `EnableTilingByEdgeDrag = 1`, and the new test pins that in. #1079's cause A is *keys absent*, so only the absent-key default needs to move:

```swift
systemTilingKeys.contains { valueFor($0) ?? !displaysSpan }
```

Same fix for the trap, and a user who deliberately turned macOS tiling on keeps the gate. As written that user loses it and both edge previews fire.

That matters because the premise underneath is the part neither of us has seen — you didn't turn Separate Spaces off, and I can't check whether those switches really go inert on a spanning setup, or on a single display with `spans-displays = 1`. Deriving from the written value makes the premise cheaper to be wrong about.

Minor: the `persistentDomain` fallback in `spacesPreference` is unreachable. `UserDefaults(suiteName:)` reads the same domain, so a key missing from one is missing from the other.

**Impact:** Separate Spaces off + macOS tiling explicitly on → gate silently off, competing previews.
**Merge:** needs the `?? !displaysSpan` narrowing first.

ref #1079

Sources/Vorssaint/Services/WindowLayout/WindowGestureSupport.swift
```diff
@@ -368,17 +368,45 @@ enum WindowEdgeSnapSupport {
...
-    static func systemTilingEnabled(valueFor: (String) -> Bool?) -> Bool {
-        systemTilingKeys.contains { valueFor($0) ?? true }
+    ///
+    /// When displays span (Separate Spaces off) those switches are greyed
+    /// out and the system's own tiling is inert, so a missing key must not
+    /// pin the gate on — the warning's own instruction is unreachable then
+    /// (issue #1079).
+    static func systemTilingEnabled(valueFor: (String) -> Bool?,
+                                    displaysSpan: Bool = false) -> Bool {
+        if displaysSpan { return false }
```
vorssaint, line 390: keep this as it is. a later report on the issue has a tiling key written on while the switches stay greyed out and macos says tiling needs separate spaces, so narrowing this to the missing key case would leave that mac stuck with the warning

Sources/Vorssaint/Services/WindowLayout/WindowGestureSupport.swift
```diff
@@ -368,17 +368,45 @@ enum WindowEdgeSnapSupport {
...
+    private static func spacesPreference(_ key: String) -> Bool? {
+        if let defaults = UserDefaults(suiteName: "com.apple.spaces"),
+           defaults.object(forKey: key) != nil {
+            return defaults.bool(forKey: key)
+        }
+        guard let domain = UserDefaults.standard.persistentDomain(forName: "com.apple.spaces"),
+              let raw = domain[key] else { return nil }
+        if let bool = raw as? Bool { return bool }
+        if let number = raw as? NSNumber { return number.boolValue }
+        return nil
```
vorssaint, line 409: this second read never finds anything the first one missed, they look at the same domain. can you drop it and keep just the first read, like the dock check below

PR #1212, opened 2026-08-31
refactor: share the window server's on-screen list and its parsing

PathGao:
## Summary

This changes no behaviour and saves nothing at runtime. The window server is
asked exactly as often as before — the number of `CGWindowListCopyWindowInfo`
calls per click is identical, and nothing is cached, snapshotted or reused
across calls. It is a maintainability change only: four places had copied out
the same expression, and a copied expression is either wrong in all four
copies from the start or gets fixed in only one of them.

What was duplicated:

- `WindowServerWindowHitTest` and `WindowServerTrafficLightHitTest` each had
  their own `CGWindowListCopyWindowInfo([.optionOnScreenOnly,
  .excludeDesktopElements], kCGNullWindowID)` call and their own near-identical
  `bounds(from:)` dictionary parsing.
- `DockClickService.onScreenWindows()` and `MouseAppExceptions.onScreenWindows()`
  were byte-for-byte the same function.

The new `Sources/Vorssaint/Services/WindowServerSupport.swift` holds one copy
of each: the on-screen list, the same list reduced to
`MouseAppExceptionSupport.Window`, the bounds parsing, and the two scans over
that list. The two hit-test enums keep their public signatures and are now one
call each.

It is **+23 net source lines, not a deletion**, which is worth explaining for a
change described as removing duplication. Two things add lines: the extracted
scans need explicit parameters (the window list, the own process id, the
eligibility test) they used to read from their enclosing scope, and about
twenty lines of new comment exist to record why the two scans deliberately
disagree. What went down is the count of copies: four of the list call to one,
two of the bounds parsing to one.

### The two scans are not interchangeable

They look like near-duplicates and are not, so they stay as two functions with
their bodies kept byte-for-byte. They differ in two ways, and a later cleanup
that unifies them would be a silent behaviour change:

1. **The far edge of a window.** `windowCandidate` compares the edges by hand
   and is inclusive on the max edge; `trafficLightCandidate` uses
   `bounds.contains(point)`, which is exclusive there. A point sitting exactly
   on a window's right or bottom edge is inside for one and outside for the
   other.
2. **A window that matches geometrically but fails a later check.**
   `windowCandidate` puts every condition in one guard and `continue`s, so it
   keeps looking behind that window. `trafficLightCandidate`'s third guard
   returns `nil` instead. That stops a click that landed on a foreground
   window from being attributed to a window behind it.

Both differences now have a test that goes red if they are erased. Each was
verified by breaking it on purpose and confirming the named check failed:

| mutation applied on purpose | check that went red |
|---|---|
| traffic-light scan switched to the inclusive edge comparisons | `a traffic light on a window's far edge belongs to the window that owns the pixel` |
| click lookup switched to `bounds.contains(point)` | `a click on a window's far edge still belongs to that window` |
| traffic-light third guard `return nil` → `continue` | `a traffic light click stops at the window in front, never reaching one behind it` |
| click lookup `else { continue }` → `else { return nil }` | `the click lookup carries on behind a window it was told to leave alone` |

### Left for later

Two more copies of the same list call exist and are untouched here:
`DockClickService.onScreenWindowIDs(pid:)` and
`DockPreviewService.onScreenWindows()`. Both can call
`WindowServerSupport.onScreenWindowInfo()` unchanged; they are follow-up work,
not omissions.

This is one of four independent changes from the same review. They touch
disjoint service files and were verified to build and test together.

## Verification

MacBook Pro (Mac17,3, Apple M5), macOS 26.6.2 build 25G83, single display,
release build from `./build.sh` (no flags), arm64.

```sh
./build.sh                   # compiles clean, no warnings
./build/Vorssaint --selftest # SELFTEST OK
./build.sh --test            # 9404 checks, 9399 on main; the 5 new ones pass
```

The five new checks live in `Tests/MetricsTests.swift` and call
`WindowServerSupport` directly, so they need no window server and no
permissions.

Two other assertions fail on this Mac, `nil layout falls back to ANSI
semicolon` and `App Switcher icon-row hints describe default app and window
shortcuts`. Both read the machine's live keyboard layout through
`TISCopyCurrentKeyboardLayoutInputSource()`, and both reproduce on unmodified
`main` here, where the input source is a Chinese input method; they are
unrelated to this change and are noted only so a clean run is not implied.

**What I could not test: the app was never launched.** `./build.sh` cannot
complete code signing in this environment (`errSecInternalComponent` on the
fan-control helper), so no `.app` bundle was produced and no real click was
ever routed through the changed hit tests on a live system.

That gap is narrower here than for a behavioural change, and it is worth
saying why. Both scans were kept byte-for-byte, and the four red-then-green
mutations above exercise exactly what a runtime check would be looking at:
which window a point on a far edge is attributed to, and whether a scan stops
at the front window or carries on behind it. It is not a complete substitute:
the tests cover the scan logic, not the call sites that feed it, so a reviewer
should still confirm on a real Mac that a click on a traffic light and an
edge-snap window drag behave as they did before.

## Anything else

`CHANGELOG.md` is untouched. No user-facing strings, so no translations are
involved.


PathGao:
The mutation table is the right way to defend a refactor claiming zero behaviour change: four deliberate breaks, each naming the check that caught it, both scans kept byte-for-byte.

One fix, and it's the body, not the code. "Two more copies of the same list call exist and are untouched here" undercounts. Grepping `main` for `[.optionOnScreenOnly, .excludeDesktopElements], kCGNullWindowID`: ten sites, four collapsed here, six left — not two. Beyond the two you name, also:

- `WindowLayoutService.onScreenWindowIDs()`
- `ScreenshotCaptureEngine.onScreenWindows()`
- `ScreenshotCaptureEngine.pickableWindows(...)`
- `WindowUseTracker.frontToBack()`

All four can call `WindowServerSupport.onScreenWindowInfo()` unchanged. The follow-up gets sized off that line, and "two left" reads as nearly done.

The app build globs `Sources/Vorssaint/**/*.swift` (build.sh:385), so the test-list entry is the only build.sh edit needed. I did not build or run anything — CI green is the only build evidence, and no click was routed through either hit test.

**Impact:** none. Both scans byte-for-byte, call count per click unchanged.
**Merge:** ship it. The "Left for later" list needs four more entries; that's a body edit, not a code change.

PR #1213, opened 2026-08-31
perf(dock-preview): drop the Combine read from the mouse-move tap

PathGao:
## Summary

The Dock Preview event tap is installed on `.mouseMoved`, `.leftMouseDown`,
`.rightMouseDown` and `.otherMouseDown`, so its callback runs on the main thread
for every mouse move on the whole system, up to about 1000 a second on a
high-poll-rate mouse. The first thing it did for a move was read `isRunning`,
which is `@Published`. A `@Published` read is a Combine subscript that performs
a `swift_dynamicCast` into a protocol conformance lookup each time, so the
hottest input path in the app paid for one on every event and did nothing with
it.

Both reads of `isRunning` on that path are gone, by different routes, because
the two methods are not the same case:

- **`discardFarMouseMove` now checks nothing.** Its only caller is `handle()`,
  the tap callback itself, which cannot run unless the tap is installed and
  enabled — and the `.tapDisabledByTimeout` / `.tapDisabledByUserInput` types
  return before reaching it. The liveness term was invariantly true there, so
  the right end state is not a cheaper read but no read. This is the one on the
  1000-a-second path.
- **`handleOnMain` reads `tap != nil`.** Its two callers are deferred
  main-queue blocks — the `DispatchQueue.main.async` in `handle()` and the
  sampling `DispatchWorkItem` in `admitMouseMove` — so the tap can be torn down
  between enqueue and execution. That guard is a real staleness check and has to
  stay; it just reads a stored `CFMachPort?`, a pointer compare, instead of a
  Combine subscript. Both methods now carry a line saying which case they are,
  so the asymmetry does not read as an oversight.

`isRunning` itself is untouched.

One correction to an earlier draft of this description, which said the wrapper
stays because the settings UI observes `isRunning`. It does not: the Settings
and menu-panel views hold the service as an `@ObservedObject` but read only
`blockedReason`. After this change the sole remaining read of `isRunning` in the
codebase is the 10-second settings timer, which resyncs when the tap is down.
The wrapper still earns its keep, because writing it is what publishes
`objectWillChange` to those two views, but no view reads the value.

Be clear about the size of this: roughly 0.28 µs per event, about 0.03% of a
core at 1000 events a second. Nobody will feel it and it will not show up as
battery life. What makes it worth a net deletion is that it was the largest
avoidable slice of a callback on the hottest path in the app, and that the slice
was Combine overhead doing no work at all.

Two measurements, on an M5 MacBook Air (Mac17,3), macOS 26.6.2 (25G83):

- A 30-second `sample(1)` of the shipping build: of the 8 main-thread samples
  inside `handle(type:event:)`, 3 were in `Published.subscript.getter` ->
  `swift_dynamicCast` -> `_conformsToProtocol`.
- A 500k-iteration `swiftc -O` benchmark on the same machine: reading a
  `@Published private(set) var Bool` on an `ObservableObject` costs 0.2818 µs,
  and reading a plain stored `var Bool` on the same class is below the timer's
  resolution.

For `handleOnMain`, where a liveness check does still belong, the reviewable
question is whether `tap != nil` really equals `isRunning`. There are six
assignments to `isRunning`, and at every one of them `tap` already carries the
answer:

| site | `tap` there |
|---|---|
| `syncWithPreferences()`, `canRun` bail | `stopTap()` on the line above nils it |
| `syncWithPreferences()`, `dockUnavailable` bail | same |
| `stop()` | same |
| `startTap()` early return | `guard tap == nil else` — non-nil by construction |
| `startTap()`, `tapCreate` returned nil | never assigned, and `startTap()` returns early if it were non-nil |
| `startTap()` success | `self.tap = tap` on the line above |

`tap` itself is written in exactly two places in the file — `self.tap = tap` in
`startTap()` and `tap = nil` in `stopTap()` — so there is no second copy of the
fact to keep in step and nothing for a seventh assignment to drift from. Both
the writes and the tap callback run on the main run loop, so no callout
interleaves with a stop. `isRunning` is not meant to diverge from tap liveness
later; if that ever changes, this is the place to come back to.

Two earlier revisions of this branch are worth recording, because each was
larger than the one that replaced it. The first added a stored `tapIsRunning`
mirror, a `didSet` on `isRunning` to write it, and a `MetricsTests` source-shape
assertion whose only job was to keep the two in step; review pointed out that
`tap` already held the fact, so all three went. The second kept a `tap != nil`
term in `discardFarMouseMove`; review pointed out that the term is invariantly
true at that one call site, so it went too. The change that started as +10 lines
and a test is now a net deletion.

Deliberately not changed:

- `isVisible` stays the computed `panel?.isVisible == true`. Measured at 7-16 ns
  against the 0.28 µs above, and AppKit owns that value through `orderOut`,
  `close` and screen changes that no assignment here brackets, so a shadow would
  trade nanoseconds for an unbounded correctness risk.
- `isNearDock`'s `NSScreen.screens.first { ... }` stays. Measured at 0.21 µs but
  AppKit already caches the array, so there is nothing to win.
- `menuBarScreenTopY` runs a second `NSScreen.screens` scan on the same path and
  is now the largest remaining cost in `discardFarMouseMove`. It is left for a
  follow-up because it carries its own screen-reconfiguration question and does
  not belong in this diff.
- The 10-second settings timer reads `isRunning` too, but it runs off the tap
  path once every ten seconds, so the published read costs nothing there and
  changing it would only add noise to this diff.

No test accompanies this. The one assertion this branch ever had was the
source-shape check guarding the mirror, and it was deleted with the mirror as
over-guarding. Asserting that `discardFarMouseMove` does not read `tap`, or that
it keeps exactly one call site, would re-add that same shape for a smaller fact.
The invariant that makes the deletion safe is the single call site, and it is
stated in the code at the point where it matters.

This is one of four independent performance fixes from the same profiling pass,
each in a different service file.

## Verification

MacBook Air (Mac17,3), Apple M5, macOS 26.6.2 build 25G83, single built-in
display, one Space. arm64.

```sh
./build.sh                     # zero warnings; ✓ Bundle ready: build/stage/Vorssaint.app
./build.sh --test              # TESTS FAILED (2 of 9399) — the input-source pair below
```

`./build.sh --test` is worth a note, because on its own it does not cover this
diff: its whitelist compiles `DockPreviewSupport.swift`, not
`DockPreviewService.swift`, so a green `--test` says nothing about whether the
changed file builds. The full `./build.sh` above is the check that covers it,
and it was run on this head rather than on an earlier revision.

The two failures are `nil layout falls back to ANSI semicolon` (`got "⌃⌥⌘ ；",
expected "⌃⌥⌘ ;"`) and `App Switcher icon-row hints describe default app and
window shortcuts`. Both read the live keyboard layout, and both track the
machine rather than the branch: they fail while the active input source is
`com.apple.keylayout.PinyinKeyboard`, whose semicolon key is a full-width `；`.
This branch was run twice with nothing changed in between except that source,
and it went `TESTS OK (9399 checks)` under `ABC` and `TESTS FAILED (2 of 9399)`
under Pinyin — so the pair is environment-dependent by demonstration here, not
just by argument. The check count is 9399 either way, the same as `main`.

**What I could not test: the app was never launched.** The codesigning note in
an earlier draft of this section is out of date — signing no longer fails here,
and the build gets as far as `✓ Bundle ready: build/stage/Vorssaint.app` using
the legacy self-signed identity. A bundle exists; it was simply never started.
The Dock Preview tap has not been installed on a live system and no mouse move
has gone through the changed guards.

That leaves the runtime half unverified, and it is worth being precise about
what a live run would add now that the term is gone rather than cheap. The
argument for deleting it is a reachability claim — that `discardFarMouseMove`
has one caller and that caller only runs while the tap is live. That is read off
the source, and a live run would catch it being wrong as previews reacting to
the pointer after Dock Preview is switched off. The paths worth a minute at
review are the ones that write `tap`: toggling Dock Preview off and on in
Settings, and revoking and re-granting Accessibility. I have also not tried a
second Mac, an external display, or the Dock on the left or right edge.

## Anything else

No `Refs` line. This came out of a profiling pass rather than a report, and I
searched the open and closed issues for Dock Preview lag, stutter, CPU and
battery on mouse movement before saying so. The nearest thing on file is the
report of the app freezing after a few hours and needing `kill -9`, and that is
a different mechanism — the main thread blocking on a dispatch pool filled with
stranded `waitUntilExit` calls, not the cost of a guard — so referencing it here
would be a false trail. Naming it without a link on purpose, so nothing gets put
on a fix track by accident.

`CHANGELOG.md` is untouched.


PathGao:
The six-assignment count is right, and `didSet` is the correct place for the mirror — writing it at each call site is exactly how it would rot on the seventh.

Two things.

`handleOnMain` still reads the published `isRunning`, and it is on this same tap path: every move that survives `discardFarMouseMove`, plus the three mouse-down types (the mask is four types, not just `.mouseMoved` as the body says). Far cheaper than 1000/s, but it is not in your "deliberately not changed" list, so I cannot tell whether it was a decision.

The mirror may not need to exist. `tap != nil` equals `isRunning` at all six sites — `stopTap()` nils it before each `isRunning = false`, the `tapCreate` failure path leaves it nil, and both true-assignments have a live tap. `tap` is written in exactly two places in the file, so the two cannot diverge. Reading a stored `CFMachPort?` is a pointer compare. That deletes the new property, the `didSet`, and the `MetricsTests` assertion that exists only to guard them — no second copy of the fact, nothing to keep in step. Unless `isRunning` is meant to diverge from tap liveness later, in which case the mirror is the safer shape and that belongs in the comment.

I did not run this; the reasoning above is from `main` source, not a live tap.

**Impact:** none. The change is correct as written.
**Merge:** needs an answer on `tap != nil` first — it decides whether this is +1 line or +10 lines plus a test.

PathGao:
@PathGao You were right that the mirror does not need to exist. Deleted it in `e9534db3`; the PR is now **−13 lines** instead of +10 plus a test.

**On `tap != nil`.** I checked all six sites rather than take it on the argument, and the equivalence holds:

| site | `tap` |
|---|---|
| `syncWithPreferences` `canRun` bail | `stopTap()` on the line above nils it |
| `syncWithPreferences` `dockUnavailable` bail | same |
| `stop()` | same |
| `startTap` early return | `guard tap == nil else` — non-nil by construction |
| `startTap` `tapCreate == nil` | never assigned, and `startTap` returns early if it were non-nil |
| `startTap` success | assigned on the line above |

`tap` is written in two places. So there is no second copy of the fact to keep in step, the `didSet` goes, and the `MetricsTests` assertion goes with it — it existed only to guard a property that no longer exists. Reading a stored `CFMachPort?` is a pointer compare, so it is also cheaper than the mirror was.

No, `isRunning` is not meant to diverge from tap liveness later. If it ever is, the mirror is the shape to come back to, and the comment now says which fact is being read and why.

**On `handleOnMain`.** It was not a decision, it was an omission — you are right that it is on the same path and that it should have been in the list either way. It now reads `tap != nil` too, so the question disappears rather than being answered.

**On the mask.** The description was wrong: the tap takes `.mouseMoved`, `.leftMouseDown`, `.rightMouseDown` and `.otherMouseDown`, not moves alone. Corrected.

`TESTS FAILED (2 of 9399)` — the count drops by one because the removed assertion was mine, and the two are the pre-existing input-source-dependent pair. Still not run against a live tap; toggling Dock Preview in Settings and revoking/re-granting Accessibility are the paths worth a minute, since those are what write `isRunning` and now also what write `tap`.

Worth saying plainly: this started as a measured micro-optimisation and your reading turned it into a deletion. The original diff added a property, an observer and a test to protect a fact the file already had.

PathGao:
Checked the table against `main` instead of taking it: `tap` is written in exactly two places, `stopTap()` nils it before every `isRunning = false`, and the timer that calls `syncWithPreferences` is on `RunLoop.main`, so no callout can interleave with a stop. The equivalence holds. −13 with the test gone is the right end state.

One thing left. `discardFarMouseMove` has exactly one call site — `handle()`, the tap callback itself — and `handle()` cannot run unless `tap` is non-nil and enabled (the `.tapDisabledBy*` branch returns before reaching it). So `tap != nil` there is invariantly true. Delete the term rather than making it cheap:

```swift
guard !isVisible, pendingHover == nil else { return false }
```

That takes the read the PR is about to zero, and the comment collapses to one line saying why nothing checks running on this path. `handleOnMain` is the opposite case — both its callers are deferred main-queue blocks, so its guard is a staleness check and has to stay.

Still not run against a live tap; none of the above depends on that.

**Impact:** none. The change is correct as written.
**Merge:** needs the `tap != nil` term out of `discardFarMouseMove` first — one token, and it is the token this PR exists for.

PathGao:
@PathGao Verification pass over the deletion in `e9534db3`. No code change was needed — every claim in the previous comment holds against the source. What did not hold is that the body had been corrected.

Checked against the file rather than the argument: there is no `tapIsRunning` and no `didSet`/`willSet` anywhere in the repo, `Tests/` is byte-identical to `main` so the source-shape assertion is gone with the property it guarded, `handleOnMain` reads `tap != nil`, and `tap` is still written in exactly two places — `self.tap = tap` in `startTap()` and `tap = nil` in `stopTap()`. I walked all six `isRunning` assignments again and the equivalence table is right at each one. The mask is the four types.

Two things in the body were still describing the first revision. The mask fix was only ever made in that comment — the body still opened with a `.mouseMoved`-only tap, and still described the mirror, the `didSet` and the `MetricsTests` assertion as if they were in the diff. I have rewritten it to match the head, and moved the equivalence table into it so the argument sits with the change instead of in a thread.

The second is a claim neither of us checked: the body said `isRunning` keeps its wrapper because the settings UI observes it. It does not. `SettingsView` and `MenuPanelView` hold the service as an `@ObservedObject` but read only `blockedReason`; after this change the sole read of `isRunning` left in the codebase is the 10-second settings timer in `startSettingsTimer()`. The wrapper still earns its keep, because writing it is what publishes `objectWillChange` to those two views, but for a different reason than the body gave. The body now says that.

One note on verification that matters more than it looks. `./build.sh --test` does not cover this diff: the whitelist compiles `DockPreviewSupport.swift`, not `DockPreviewService.swift`. `TESTS OK (9399 checks)` on this head, but that number is silent about whether the changed file builds, and the clean full `./build.sh` quoted in the body was from the revision that added the mirror. A full build of the current head is the one thing still outstanding. Also worth knowing: the two assertions previously reported as failing (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints…`) passed here — they track the machine's active input source, which was `ABC` for this run and `PinyinKeyboard` for the earlier one.

No `Refs`. I searched open and closed issues for Dock Preview lag, stutter, CPU and battery on mouse movement; #971 is the only Dock Preview performance report and it is the dispatch-pool deadlock, a different mechanism, so linking it would be a false trail. Still not run against a live tap.

PathGao:
@PathGao Done in `3d43b25c`. The term is out of `discardFarMouseMove`:

```swift
// Nothing checks whether the tap is running: the only caller is the tap
// callback, which cannot run unless it is.
guard !isVisible, pendingHover == nil else { return false }
```

I checked the reachability claim rather than taking it. `discardFarMouseMove` has exactly one call site in the tree — line 364, inside `handle()` — and `handle()` has exactly one, the `tapCreate` callback. The `.tapDisabledByTimeout` / `.tapDisabledByUserInput` branch returns above it. So the term was invariantly true and this is a deletion, not a cheaper read.

`handleOnMain` keeps its guard, and I added a line saying why, because the asymmetry is the part a later reader will get wrong. Its two callers are the `DispatchQueue.main.async` in `handle()` and the sampling `DispatchWorkItem` in `admitMouseMove`, both deferred, so the tap can go away between enqueue and execution. Without that line the next pass at this file deletes the one guard that is doing work.

No test. The only assertion this branch ever had was the source-shape check on the mirror, and it went as over-guarding; asserting that `discardFarMouseMove` does not read `tap`, or that it keeps one call site, is that same shape again for a smaller fact. The invariant is stated in the code where it matters instead. Say the word if you want the call-site count pinned in `MetricsTests` and I will add it.

Two corrections to my previous comment. It said the full `./build.sh` was outstanding — it has now run on this head: zero warnings, `✓ Bundle ready: build/stage/Vorssaint.app`. And it said no code change was needed, which was written against a stale read of this thread and is superseded by the above. `./build.sh --test` is `TESTS FAILED (2 of 9399)`, the input-source pair; I happened to run this branch under both `ABC` and Pinyin with nothing else changing, and it goes `TESTS OK (9399 checks)` under the former, so that pair is now environment-dependent by demonstration rather than by argument. Worth knowing that `--test` does not compile `DockPreviewService.swift` at all — the whitelist has `DockPreviewSupport.swift` — so on this file only the full build is evidence.

Still not run against a live tap. The reachability argument is read off the source, and what a live run would catch is previews reacting to the pointer after Dock Preview is switched off.

On the bot's prompt above: the mention was only a mention. The freeze report is the dispatch-pool deadlock, a different mechanism from a guard's cost, so there is no `Refs` line. I have taken the `#` off it in the description so nothing gets put on a fix track by accident.

PathGao:
@PathGao The term is out of `discardFarMouseMove`, and the line explaining why `handleOnMain` keeps its guard is what stops the next pass deleting the one guard that does work. Verified against `main`: this file reads `isRunning` in three places, so the settings timer is all that is left.

One sibling, worth checking before the other three in this set land. `SmoothScrollService.handle` reads `ScrollInverter.shared.isRunning` — same `@Published` on a tap callback, once per wheel event past the exception guards. Its `startTap` already says

```swift
// Its isRunning is read from the tap callback, so it is built off the event path.
_ = ScrollInverter.shared
```

so the path was known and only the singleton construction was moved off it. It spans two services, so a per-file split drops it. `ScrollInverter.tap` is private and `SettingsView` and `MenuPanelView` both hold the inverter as `@ObservedObject`; I have not checked whether either reads the value, so I am not prescribing a shape. Ignore if one of the other three covers it.

Not run against a live tap.

**Impact:** none.
**Merge:** ship it.

I only get this PR's diff here, so 1218/1221/1222/1224/1236/1242 need their own runs.

PathGao:
Checked the sibling. `SettingsView` does read it — `if scrollDirectionEnabled, inverter.isRunning` gates the green "active now" row in the mouse pane. `MenuPanelView` holds the inverter as an `@ObservedObject` but never reads the value. So the inverter is not this file's case: the wrapper has a live reader and stays published.

That makes it an accessor rather than a deletion. `ScrollInverter.tap` is private and the reader is another type, so the shape is:

```swift
/// Tap liveness for the smooth-scroll callback, without the Combine subscript.
var isTapInstalled: Bool { tap != nil }
```

Computed off `tap`, so still no second copy of the fact. The equivalence holds at all four `isRunning` writes in that file: `start()` early return (non-nil by the `guard tap == nil`), `tapCreate` failure (nil — the same guard returned early otherwise), success (assigned the line above), `stop()` (nil'd the line above). `tap` is written in exactly two places, 107 and 132.

The size is where it diverges from this PR. That read is per wheel tick while the wheel is turning, not 1000/s continuously. Same shape, much smaller slice — its own PR, and it should not hold these four.

Not run against a live tap.

**Impact:** none.
**Merge:** ship it.

PR #1214, opened 2026-08-31
perf(autoquit): stop re-reading every window on every focus change

PathGao:
## Summary

AutoQuit rebuilds an app's window list through the Accessibility API on the main thread, and it does that far more often, and more expensively, than it needs to.

`handleAX` ran a full `refreshWindows` inline for every notification in `windowRefreshNotifications`, which includes `applicationActivated`, `mainWindowChanged` and `focusedWindowChanged`. A single app switch fires all three back to back, and the workspace activation handler adds a fourth refresh, so switching apps paid for four identical rounds with no coalescing. Each round is synchronous cross-process traffic: `AXWindows` on the app element, then two to four attribute reads per window to decide whether it is a standard window, then the main and focused window fetched and put through that same predicate again although they are almost always already in the list, then three `AXObserverAddNotification` calls for every window on every refresh — including pairs already registered, which still cost a full round trip and differ only in their return code.

Three changes, all in `AutoQuitService`:

**Deduplicate before the predicate.** `appendUnique` became `appendIfStandard`, which does the `CFEqual` match against what has already been collected before asking whether the element is a standard window. The main and focused window are normally duplicates of entries in the `AXWindows` list, so they cost nothing now instead of two to four round trips each.

**Do not re-register notifications the observer already holds.** A per-app set remembers the windows already watched and `watch` skips them. It is dropped in `detach(pid:)` and `stop()` alongside the other per-app state, the individual entry is dropped when `kAXUIElementDestroyedNotification` arrives, and each refresh intersects the set with the windows it actually saw. A skipped window still counts toward the registered total `AutoQuitSupport.needsWindowWatchRetry` reads; leaving it out would restart the retry round on every refresh for as long as the app runs.

A window is memoized only once **all three** of `windowNotifications` registered, while the retry count still turns on the destroyed notification alone. The two conditions look mergeable and are not:

| destroy | miniaturize pair | counts as watched | memoized | next refresh |
|---|---|---|---|---|
| ok | ok | yes | yes | skipped |
| ok | refused | yes | no | asks all three again |
| refused | — | no | no | `needsWindowWatchRetry` fires |

Row two is why the gate exists. An app that stops answering partway through `watch` returns `.cannotComplete` for the miniaturize pair; memoizing on the strength of the destroy registration alone would mean nothing ever asks for the other two again. Without `AXWindowDeminiaturizedNotification` the window is never handed to `clearMinimizedWindow`, so the id `recordMinimizedWindows` put in `minimizedWindows[pid]` stays there, `hasKnownMinimizedWindow` keeps answering true, and `shouldQuitAfterWindowCheck` refuses for the life of the app. It also has to keep counting as watched, or the retry never settles.

`isWindowNotificationRegistered`'s doc comment argued from "every refresh registers the windows it is already watching again", which is the premise this change removes. Rewritten to the one that now holds.

**Coalesce refreshes per app.** `handleAX` and the activation handler now schedule at most one refresh per app onto the next main run-loop turn, so a burst collapses into one. A single turn rather than a timer is deliberate: the refresh is what marks an app as having had a window, and the close that quits it can follow immediately. A pending refresh is dropped when the app's observer is torn down.

Two properties of that deferral the rest of this description does not otherwise claim, both checked by reading the code. It is safe across a detach and a recycled pid: the queued block re-tests `pendingRefreshes.remove(pid) != nil`, which `detach` has already cleared, and it reads `observers[pid]` at run time rather than capturing an observer. And deferring the deminiaturize refresh moves `recordMinimizedWindows`'s minimized read one turn after the `clearMinimizedWindow` that precedes it, which fails in the safe direction — by then the window reads un-minimized, so its id is not put back.

### Why the memo is keyed on the element rather than the window id

The obvious key is the window id, but `AXWindowResolver.windowID(for:)` calls `_AXUIElementGetWindow`, which is a MIG round trip into the target app rather than a local read. An id-keyed memo therefore spends two round trips per window to save three, and on an app's first refresh, when nothing is remembered yet, it spends them to save none. Keyed on the `AXUIElement`, `CFEqual` and `CFHash` answer locally and the memo costs nothing to consult. Elements for the same window come back from separate lists as distinct objects that compare equal, which is the property the window collection in this file has always deduplicated on.

### Effect

Round trips per refresh, for an app with N Accessibility-listed standard windows:

| | before | first refresh | steady state |
|---|---|---|---|
| formula | 9 + 7N | 3 + 7N | 3 + 4N |
| N = 1 | 16 | 10 | 7 |
| N = 3 | 30 | 24 | 15 |

The coalescing then takes one app switch from four refreshes to one: 64 round trips down to 7 for a single-window app in steady state, 120 down to 15 at three windows.

**These counts are derived by reading the code, not measured on a running build.** See Verification for why nothing was run.

### Deliberately not changed

`isStandardWindow` results are still not cached across calls. A window's subrole genuinely changes with its state — a minimized window reports `AXDialog` and reports `AXStandardWindow` again once restored — so a cache surviving a state change would be wrong. The deduplication above is within a single collection pass only.

The close-button, Command-W, attach, watch-retry and Space re-arm paths still refresh inline. They are rare or user-initiated, and the retry paths' timing is bound to their own origin.

`WindowEnumerator.accessibilityWindows(for:)` has the same predicate-before-deduplication ordering with a more expensive predicate. It is left alone to keep this to one file; #1217 makes that edit in the file that owns it and points back here for `standardWindows(of:)`. That cross-reference is accurate as of `30490260`: this branch does turn `appendUnique` into `appendIfStandard`, and the `CFEqual` match runs ahead of `isStandardWindow` in both the `kAXWindows` loop and the main/focused loop. If this PR is dropped, `standardWindows(of:)` still needs the same treatment.

### One behaviour change worth knowing about

Re-registering on every refresh used to quietly self-heal a registration lost without a destroy notification, and the memo removes that. It fails in the safe direction: a lost destroy registration means an app is not quit, never that an app with a window is quit.

## Verification

Mac17,3 (Apple M5), macOS 26.6.2 build 25G83, single display, arm64.

```
./build.sh                    # compiled with no warnings; build/Vorssaint produced, 41.5 MB arm64
./build.sh --test             # TESTS OK (9402 checks) — three more than this branch's base
```

`./build.sh` and `--selftest` are from the original submission and were not re-run for this revision; the commit is unchanged. `./build.sh --test` was re-run twice for it, because a source-shape assertion is worth no more than its red state. `AutoQuitService.swift` is not in the `--test` compile list, so what guards the gate reads the file as source, and an assertion of that kind that has never been seen red is decoration. Reverting the gate to `if watched` gives:

```
TESTS FAILED (3 of 9402):
  - nil layout falls back to ANSI semicolon: got "⌃⌥⌘ ；", expected "⌃⌥⌘ ;"
  - AutoQuit registers each window once and coalesces refreshes, both dropped when the
    app goes: missing ["if allRegistered { watchedWindowElements[pid, default: []].insert(watchedWindow) }"]
  - App Switcher icon-row hints describe default app and window shortcuts
```

Restoring it gives `TESTS OK (9402 checks)`. The other two in that list are the input-source-dependent pair described below; they failed on the first run and passed on the second because the machine's active input source changed between them, not because anything in this branch did.

**What I could not test: the app was never launched, so none of the changed paths ran.** `./build.sh` cannot finish codesigning in the environment I built in, so no `.app` bundle exists to grant Accessibility to and there was no way to watch a real app switch. Compilation, the self-test and the unit suite are all that stand behind this, and none of them executes `refreshWindows`, the watch memo or the coalescer — the suite covers `AutoQuitSupport`'s pure helpers plus source-shape assertions, since `AutoQuitService` itself is not in the test binary.

That gap falls exactly on what this change alters: the memo changes when `AXObserverAddNotification` is called, the coalescer changes when a refresh happens relative to a close, and the note above means a lost registration is no longer re-registered. Worth exercising on real hardware:

- switching repeatedly between apps that have several windows, checking each still quits on its last close
- closing the last window of an app, including leaving full screen with the green button and after minimizing
- an app that creates its window asynchronously on launch, the case 9adbaf9c added the watch retry for
- quitting an app while a refresh is pending for it
- a window destroyed while its app is in the background

The `CFEqual`/`CFHash` property the memo depends on was checked separately against the live windows of four running apps, each read through two independent `AXWindows` calls: distinct objects every time, equal under `CFEqual`, equal hashes, and set membership holding across the two lists.

Two assertions in the suite pass or fail with the machine's active input source rather than with this branch — `nil layout falls back to ANSI semicolon` and `App Switcher icon-row hints describe default app and window shortcuts`. Under a Pinyin input source they fail here, on unmodified `main` as well; under an ANSI one the whole suite is green. Both runs above are on the same checkout, minutes apart, and the difference between them on those two is the input source alone.

## Anything else

No `Refs` line, after searching rather than assuming. Every issue in the repository was scanned by label (`bug`) and by keyword — `quit on close`, `auto-quit`, `minimi[sz]e`, `stays open` — and the #836 bug summary was read for the same. The only open report of quit-on-close failing to fire is #1008, whose cause is the one #1022 fixed, Accessibility listing no window at all; it is already labelled `fixed (pending release)` for that PR and nothing here brings it closer. No open issue describes auto-quit working until an app is minimized and then stopping, which is the shape the gate above prevents — and that defect was introduced by this branch's own first commit, so it has never been in a release for anyone to report.

#938 stays out as well. `appendIfStandard` now carries the `AXUIElementSetMessagingTimeout(window, 0.35)` write that used to sit inline in `standardWindows(of:)`, so the write moved, but the number of writers and the three values they disagree on are unchanged and nothing here takes ownership of the setting.

One other place in this file records a multi-step registration as complete: `attach` discards all seven return codes from the `appNotifications` loop and stores `observers[pid]` regardless, and it returns early on a non-nil entry, so a partial app-level registration is permanent for the life of the process. It is left alone, and not only to keep this to one change. It is not the same defect — every notification in that loop does nothing but schedule a refresh, and `NSWorkspace.didActivateApplicationNotification` reaches `scheduleRefresh` without going through Accessibility at all, so the worst case is a refresh delayed to the next app switch rather than a state nothing clears. And gating that store on all seven would fail in the worse direction: an app that permanently refuses one of them would never be attached, which turns auto-quit off for it entirely instead of delaying one refresh. Pre-existing on `main`, unchanged here.

One of four independent performance fixes from the same review, each in a different service file.


PathGao:
The `CFEqual`/`CFHash` key choice is right, and checking that property against four live apps before relying on it is what makes the memo trustworthy.

One gap.

**A partial registration is memoized as a complete one.** `watch` sets `watched` from `kAXUIElementDestroyedNotification` alone — the first of the three in `windowNotifications` — then memoizes on it. If the app stops answering after that first call, the two miniaturize notifications come back `.cannotComplete`, the window is memoized anyway, and no refresh asks again. Every refresh used to retry all three.

Cost: with no `AXWindowDeminiaturizedNotification`, `clearMinimizedWindow` never runs for that window, so the id `recordMinimizedWindows` put in `minimizedWindows[pid]` stays, `hasKnownMinimizedWindow` keeps returning true, and `shouldQuitAfterWindowCheck` refuses for the life of the app.

Keep `watched` as it is, gate only the memo:

```swift
if !AutoQuitSupport.isWindowNotificationRegistered(result) { allRegistered = false }
...
if allRegistered { watchedWindowElements[pid, default: []].insert(watchedWindow) }
```

The body's disclosure covers the destroy notification; this is the miniaturize pair. Also stale now: the doc comment on `isWindowNotificationRegistered` still says every refresh re-registers what it already watches.

I read this, did not run it — the failure above is inferred from call order, not observed.

**Impact:** an app that refuses one miniaturize registration never auto-quits again until relaunch. Needs a busy app mid-refresh plus one minimize/restore.
**Merge:** needs the memo gated on all three registrations first.

PathGao:
@PathGao Confirmed and fixed in `30490260`. You were right on both counts.

**The partial registration.** `watched` was read from the destroy notification alone and the memo was written from `watched`, so a window whose miniaturize pair was refused got remembered as fully watched and nothing asked again. I followed the chain you named and it holds: no `AXWindowDeminiaturizedNotification` means `clearMinimizedWindow` never runs for that window, the id `recordMinimizedWindows` put in `minimizedWindows[pid]` stays, `hasKnownMinimizedWindow` keeps answering true, and the app never quits again until relaunch.

Gated the memo exactly as you suggested, and kept `watched` as it was:

```swift
let registered = AutoQuitSupport.isWindowNotificationRegistered(result)
if !registered { allRegistered = false }
if notification == kAXUIElementDestroyedNotification { watched = registered }
...
if allRegistered { watchedWindowElements[pid, default: []].insert(watchedWindow) }
```

The split matters and is now written down in the test, because merging the two conditions is the tempting simplification:

| destroy | miniaturize pair | counts as watched | memoized | next refresh |
|---|---|---|---|---|
| ok | ok | yes | yes | skipped |
| ok | refused | **yes** | **no** | retries all three |
| refused | — | no | no | `needsWindowWatchRetry` fires |

Row two is the fix. It has to keep counting as watched or the retry never settles, and it must not be memoized or the missing pair is never requested again. `needsWindowWatchRetry` reads the count, which is unchanged, so the retry behaviour is untouched.

**The stale comment.** Also right, and it was mine to fix — `isWindowNotificationRegistered`'s doc still argued from "every refresh registers the windows it is already watching again", which is the premise this PR removed. Rewritten to the premise that now holds.

**Test.** The two source-shape assertions in this PR went red on my own change, which is what they are for. Rather than relax them I re-pinned them to the stricter invariant, then verified: reverting the gate to `if watched` turns the memo assertion red by name and restores green when put back. `TESTS FAILED (2 of 9402)` before and after, the two being the pre-existing input-source-dependent pair.

Still not run on a live tap — same gap as the original description. Row two of that table needs a busy app mid-refresh plus a minimize and restore, which is the case worth a minute on real hardware.

On the `Refs #938` bot hint: leaving it out on purpose. #938 is about messaging-timeout ownership; this PR keeps the existing 0.35 s value and does not touch it. Referencing it would put that report on the fix track for something this does not fix.

PathGao:
The split is right, and the table pins it better than prose would — row two is exactly the state that has to count as watched without being memoized.

Re-read the whole diff against `main`, not just the changed lines. Two things the body does not claim, both of which hold:

- `scheduleRefresh` survives detach and a recycled pid. The queued block's `pendingRefreshes.remove(pid) != nil` fails once `detach` cleared it, and it re-reads `observers[pid]` at run time rather than capturing one.
- Deferring the deminiaturize refresh puts `recordMinimizedWindows`'s minimized read one turn after the `clearMinimizedWindow` that precedes it. Safe direction — the window reads un-minimized by then, so the id is not re-added.

Nothing else to flag.

On `Refs #938`: agreed. The 0.35 s write moved into `appendIfStandard`, but the writer count and the values are unchanged, so nothing in that report is closer to fixed.

Row two still has not run on a live tap, and I cannot check that either. If auto-quit stops firing for someone after this ships, that is the first place to look — not a gate.

**Impact:** none beyond what the body declares.
**Merge:** ship it.

PathGao:
@PathGao Description-only revision on `30490260`. No code changed and nothing was pushed — the gate you asked for is already the commit you reviewed, and the diff is the one that got `ship it`.

What the body was missing was the gate itself. It still described the memo as it stood before `30490260`: three changes, the middle one written from `watched`, no mention of the all-three condition. Anyone reading the description rather than the diff would have concluded the defect you named is still in the branch. The memo section now carries the condition, the three-row table, and the chain it prevents — no `AXWindowDeminiaturizedNotification`, so `clearMinimizedWindow` never runs, so the id stays in `minimizedWindows[pid]`, so `hasKnownMinimizedWindow` answers true for the life of the app. The two properties from your re-read are in there too, since neither was claimed anywhere: `scheduleRefresh` re-testing `pendingRefreshes.remove(pid)` and re-reading `observers[pid]` at run time rather than capturing one, and the deferred deminiaturize refresh landing after `clearMinimizedWindow` in the safe direction.

I did not take the assertion on trust. `AutoQuitService.swift` is not in the `--test` compile list, so what guards the gate is a source-shape assertion, and one that has never been seen red is decoration. Reverting the gate to `if watched` gives `TESTS FAILED (3 of 9402)`, with `AutoQuit registers each window once and coalesces refreshes...` naming the missing line; restoring it gives `TESTS OK (9402 checks)`.

One thing worth flagging about that pair of numbers: the other two failures in the red run are the input-source assertions, and they passed on the green run. Same checkout, minutes apart — the machine's active input source changed between the two, from Pinyin to an ANSI layout. So a fully green `--test` here is not evidence they were fixed, and the earlier `2 of 9402` in this thread is not evidence of a regression. Those two report the keyboard, not the branch.

**The other memo of the same shape.** There is one, and it is `attach`: the `appNotifications` loop discards all seven return codes, `observers[pid] = observer` is stored regardless, and `attach` returns early on a non-nil entry, so a partial app-level registration is permanent for the process. I left it, and the body says why rather than only that it was left. It is not the same defect — every notification in that loop does nothing but schedule a refresh, and `NSWorkspace.didActivateApplicationNotification` reaches `scheduleRefresh` without touching Accessibility, so the worst case is a refresh delayed to the next app switch, not a state nothing clears. Gating that store on all seven would also fail the wrong way: an app that permanently refuses one notification would never attach at all, which is auto-quit off for that app rather than one late refresh. Pre-existing on `main`, untouched here.

**#1217.** Its description now says this PR fixes `AutoQuitService.standardWindows(of:)` by turning `appendUnique` into `appendIfStandard` so the `CFEqual` match runs ahead of `isStandardWindow`, in both the `kAXWindows` loop and the main/focused loop. I checked that against the branch rather than against the description: it holds exactly as written, both loops call `Self.appendIfStandard`, and the dedup guard is the first statement in it. Nothing to correct over there. Said the same here so the two descriptions do not drift apart.

**Refs.** Still none, but the body now shows the search instead of only the conclusion. Every issue scanned by the `bug` label and by keyword, plus #836. #1008 is the only open report of quit-on-close not firing and its cause is the one #1022 fixed; it already carries that fix pending release. Nothing describes auto-quit stopping after a minimize, which is unsurprising — that defect was introduced by this branch's own first commit and has never shipped. On #938 I kept your wording about the 0.35 s write having moved into `appendIfStandard` while the writer count and the values did not.

Row two is still unrun on a live tap. I did not run it either — no launch, no GUI testing from me.

PathGao:
Showing the source-shape assertion red by name is the part worth keeping — an assertion never seen red is decoration.

One correction to the body, not the code. `attach`'s stated reason doesn't hold: not all seven only schedule a refresh. `kAXApplicationHiddenNotification` and `kAXApplicationDeactivatedNotification` are absent from `windowRefreshNotifications`, and `appHidden` is the only app-level notification `shouldScheduleWindowCheck` ever answers true for (`appDeactivated` reaches nothing at all).

The conclusion holds, the reason doesn't. A refused `appHidden` registration is not a late refresh — it drops the quit check for an app that hides instead of destroying its last window. Bounded, because `markCloseButtonRequest` already schedules a round at 0.35/1.0/2.2 s, so what is lost is an app that hides later than that inside the 5 s grace. Worth saying that instead, since the body is now the record of why the memo was left.

Read on main, not run.

**Impact:** none from this PR; the `attach` gap is pre-existing.
**Merge:** ship it — already merged, this is about the body.

PR #1216, opened 2026-08-31
fix(panels): let a composing input method keep Return, the arrows and Esc

PathGao:
## Summary

The snippet library and the clipboard quick panel take Return, the arrows and
Esc away from an input method that is still composing, so their search fields
are unusable in Chinese, Japanese and Korean. Typing pinyin into either search
box brings up the candidate window, and then the arrows walk the list instead
of the candidates, Return inserts a snippet or copies an entry instead of
accepting the candidate, and Esc closes the whole panel instead of dropping
the composition. Dead keys for accents lose the same way.

Two panels already get this right and say why. `CommandBarService` guards its
monitor with `fieldIsComposing(in:)` — `panel.firstResponder as? NSTextView`,
then `hasMarkedText()` — and the scratchpad guards Esc the same way. The
snippet library had no composition check at all. The clipboard panel had one
that yields to a focused text view, but only when it is editable and *not* a
field editor, which is exactly the shape a `TextField`'s editing session has,
so the search box was excluded by construction. Its comment ("The search box
uses a field editor, so its existing list shortcuts stay intact") described a
deliberate choice that predated the composition case; it is rewritten rather
than left standing next to the new behaviour.

- Snippet library: the same two-line guard the Command Bar uses, ahead of the
  key switch. The panel's only text input is the search field, so outside
  composition nothing changes.
- Clipboard quick panel: the responder test moves into
  `ClipboardHistoryFocus.textViewOwnsKeys(isComposing:isFieldEditor:isEditable:)`,
  which yields on composition first and otherwise keeps the previous answer.
  The multiline editor still owns its editing keys; the search field and the
  read-only preview still leave the list's shortcuts alone.

Two things I did not do. The composition question is now answered in four
places, and folding all four into one shared predicate is the obvious next
step — but the other two panels are correct today, and pulling them in would
put three more features in a fix that has to be reviewed against an IME. If
that consolidation is wanted I would rather do it as its own change. I also
left the clipboard predicate under a `ClipboardHistory*` name rather than a
neutral shared one, because the snippet panel's version takes one input and
this one takes three: they are the same rule over different panel contents,
not the same function.

Extracting the clipboard predicate is what makes any of this checkable:
`build.sh --test` compiles a file whitelist, `ClipboardHistorySupport.swift`
is already on it, and the four new assertions pin the truth table including
the case that was wrong. The snippet guard has no executable assertion — it is
a literal copy of the Command Bar's guard, and the service file is not on the
test whitelist. I did not add it there.

## Verification

MacBook Air (Mac17,3), Apple M5, macOS 26.6.2 (25G83), Xcode 26.6, my own
build off `origin/main` @ `b0dde656`.

`./build.sh` finished with no warnings; `./build/Vorssaint --selftest` OK;
`./build.sh --test` `TESTS OK (9403 checks)`. Clean `origin/main` on the same
machine gives `TESTS OK (9399 checks)`, so the four new assertions are the
whole difference and nothing else moved.

The new assertions were checked against their own subject: reverting
`textViewOwnsKeys` to the old `isEditable && !isFieldEditor` turns "a
composing search field keeps Return, the arrows and Esc" red and nothing
else, so the guard has a check that fails when it is removed.

What I could not test: I have not exercised this on a real IME. Nobody has
typed pinyin, kana or hangul into either search field on this branch and
watched the candidate window take the arrows, Return and Esc. The change is
argued from the existing `CommandBarService` and `SettingsView` guards, which
use the same `firstResponder as? NSTextView` + `hasMarkedText()` path against
a SwiftUI `TextField`, and from the truth table now under test. Someone with a
CJK input source should confirm the candidate window behaves in both panels,
and that Esc with no composition still closes them.


PathGao:
Extracting `textViewOwnsKeys` is the right call: the truth table now has a check that fails when the guard is removed, which neither existing guard ever got.

Same shape, one site the body doesn't count. `AppDelegate.handlePopoverKeyDown` takes Esc before its own `isTextEditingActive` check runs:

```swift
if popover.isShown, event.keyCode == UInt16(kVK_Escape) {
    closePopover()
    return nil
}
```

The menu bar popover hosts the clipboard search and the Homebrew search field named in the comment right below that. Composing pinyin there and hitting Esc closes the popover instead of dropping the composition. Return and Space already yield; only Esc doesn't. So it's five places, and the fifth is still wrong.

Not asking you to widen this PR — separate change. It does make the consolidation case concrete: the site was missed because the rule lives in each monitor.

Not verified: no IME here either, and the popover path is outside the diff, read off `main`.

**Impact:** none beyond the two panels this touches.
**Merge:** ship it.

ref #1217

Issue #1158, opened 2026-08-30
[Feature]: Add protection against accidental ⌘Q and ⌘W


RuanMD:
## Feature request

It is easy to accidentally close a window with Command + W or quit an application with Command + Q.

This can interrupt work, close important windows, stop recordings, interrupt downloads or cause the user to lose context.

It would be useful to have an optional protection layer for these shortcuts.

### Proposed Solution

Please add configurable protection for Command + Q and Command + W.

Suggested options:

- Hold the shortcut for a configurable amount of time
- Press the key twice to confirm
- Require an additional modifier
- Show a small confirmation HUD
- Disable protection for selected applications
- Enable protection only for selected applications

Example:

Command + Q
→ hold for 800 ms
→ application quits

or:

Command + Q
→ press Q twice
→ application quits

The same system could be available for Command + W.

This should be configurable globally and per application.

### Feature Area

Other

solid-pixel:
Might be a duplicate of https://github.com/vorssaintapp/vorssaint-utils/issues/892

PathGao:
Same request as #892 — that's the main thread, follow it there.

The ⌘W half is new though. Add a note on #892 saying you want the same protection when closing a window, not just when quitting, otherwise it'll likely get built for ⌘Q only. The options you listed (hold, double-press, per-app on/off) are worth putting there too.

PathGao:
You don't need to add that note on #892 after all. The fix that just landed covers closing a window, not only quitting — both are protected.

The rest of the message above still holds: nothing on your Mac has changed yet, and you'll get a comment here with the version to install once it ships.

Issue #892, opened 2026-08-24
Add a confirmation gate for ⌘Q to prevent accidental app quits


MrPike:
### Problem or Motivation

It is easy to hit ⌘Q by mistake and immediately close the current app, losing its state.

### Proposed Solution

Introduce functionality that intercepts ⌘Q and requires a more deliberate action before the quit actually fires. This may include:
- ⇧⌘Q (Shift as an additional modifier)
- Double-tapping Q while holding ⌘
- Pressing and holding ⌘Q for a short duration

### Feature Area

Mouse and keyboard

solid-pixel:
+1 for this.

Supercharge has a really nice implementation of this: pressing ⌘Q doesn’t quit immediately, you have to keep holding the shortcut for a short moment before the app actually quits. That small delay is enough to prevent accidental quits without adding a confirmation dialog or changing the normal shortcut.

A similar “hold ⌘Q to quit” option in Vorssaint would be ideal, especially if the hold duration could be configurable and there were a small visual progress indicator while holding.

MrPike:
> The fix for this is out now, in **3.3.3**.
> 
> [@MrPike](https://github.com/MrPike), could you update to that version and say whether it fixed things on your Mac? We cannot tell from here — yours is the setup that found this.
> 
> Please say **how much of it is fixed**, not only whether it is. If part of what you reported still happens, or something related turned up, say so here and the issue goes back for another look — this is not a last call.
> 
> If nothing arrives in two weeks the issue will be closed as it stands, and a comment reopens it at any point after that.

Having updated to the latest version - I can confirm that the functionality appears to work as described - thank you. There are no obvious issues or further enhancements I can suggest regarding this feature.

solid-pixel:
For me the Hold to Quit feature does not work as expected.

The "New in this update" window shows this:

<img width="500" height="auto" alt="Image" src="https://github.com/user-attachments/assets/4e10b960-9747-4210-b299-a6e0a9d3cc70" />

whereas this is what I see at the bottom of my screen (also, ideally, it would be make more sense for it to be at the top) while holding `CMD+Q`. It looks different and shows no visual feedback:

<img width="500" height="auto" alt="Image" src="https://github.com/user-attachments/assets/c3eab9b7-7971-4227-9c80-039888355bf7" />

And I did enable the feature:

[...]

PR #1218, opened 2026-09-01
fix(quit-protection): follow the keyboard layout and bound the confirmation swallow

PathGao:
## Summary

Three fixes to the Command-Q / Command-W protection from #1200.

**Protection never engaged on a non-Latin keyboard layout.** Matching compared `charactersIgnoringModifiers` with "q" and "w". That character is the one the key types *without* Command, so a Russian layout reports "й" and a Greek one ";" for the very key that quits — no match, the event passed through untouched, and Command-Q quit the app with protection switched on and no feedback. The positional `fallbackKeyCode` never covered it: the character is non-nil for these presses, so that branch was unreachable.

Keyboard layouts carry a separate Command table, and it is the one the system resolves key equivalents through. `UCKeyTranslate` with `cmdKey` returns "q" for key code 12 on Russian and Greek, 7 on Dvorak — where Command-Q genuinely moves — and 12 on "Dvorak - QWERTY ⌘", which reverts to QWERTY exactly while Command is held. Matching now compares key codes and asks that question of the keycap cache in `GlobalShortcut`, which since #1226 holds both of the layout's tables for every key code, refreshes on the same `kTISNotifySelectedKeyboardInputSourceChanged` notification, and since #1220 already reads the ASCII-capable layout under an input method such as Pinyin. The cache only ever answered key code → label; the reverse question it now also answers is `GlobalShortcut.layoutKeyCode(for:usesCommand:)`, and `QuitProtectionShortcut.commandKeyCode` is that call. The typed character is left as the answer only where no layout has been read at all.

Rejected alternatives: falling back to the ANSI position (12/13) whenever the character misses fixes Cyrillic and Greek but claims ⌘' on Dvorak, where key code 12 is not Command-Q and confirming would terminate the app; gating that fallback on "the character is not ASCII" breaks on Greek, whose Q key types ";". Reading the layout needs no such guess.

**The post-confirmation swallow could latch forever.** `swallowShortcut` was set on confirmation and cleared only by a matching key up or by the feature being switched off. A key up can be lost: the tap is re-enabled after being disabled on timeout, a tap ahead of ours takes the event, the session switches user. From then on every Command-Q and Command-W was swallowed, and since the same branch also swallows `isRepeat`, so was every key's autorepeat — holding any letter produced a single character. Nothing the user could do restored either.

The swallow now ends at two more points: the tap being re-enabled after it was disabled (precisely the window where events go missing), and a two-second ceiling checked on the next key event.

Command coming up is deliberately not a release point, so `handleFlagsChanged` keeps `main`'s behaviour of ending only a pending confirmation there. Key ups are independent, so the keys the swallow covers can still be down when Command comes up, and Command first is the common order to let go. Clearing the swallow there would leave Q or W held with no swallow: the next autorepeat finds nothing to eat, `dropsAutorepeat` declines because Command is up, and the repeat passes through as a stray q or w into whatever app came forward. On `main` the swallow holds until the letter's key up and eats it, and that stays so here. The two-second ceiling already makes a lost key up recoverable, which was the only reason to release earlier.

**Holding the Q or W key with no Command produced one character and no repeat.** This is the second `isRepeat` in `handleKeyDown`, the one outside the swallow branch. It sat below the enabled and scope guards but above every modifier read, so with protection on it dropped the autorepeat of any key `matchingShortcut` resolved — and matching answers on the bare key, since the layout's Command table names a key code, not a chord. Following the layout carried that onto й and ц on Russian and ; and ς on Greek. The drop is now gated on Command (`QuitProtectionSupport.dropsAutorepeat`). Nothing below it acts without Command — both `isBaseShortcut` and `isExtraShortcut` require it — so a held Command-Q still cannot restart a hold or count as the second of a double press, which is what the drop exists for. Git history gives it no other purpose: it arrived unconditional in 4320c2e8 with the feature and was untouched by the audit pass in f25d1aa8, and nothing else in the tree reads it.

Deliberately unchanged: the `|| isRepeat` inside the swallow branch keeps its width — it is what stops the keys held through a confirmation from repeating into the app — and it is bounded by the two-second ceiling above rather than narrowed. The HUD and the settings surface are untouched.

## Verification

MacBook Air (Mac17,3), Apple M5, macOS 26.6.2 (25G83), Xcode 26.6 / Swift
6.3.3, self-built from this branch.

- `./build.sh --test` — `TESTS OK (9543 checks)` against the 9528 on `main`, so
  the 15 new assertions are the whole difference. The pair that fails whenever
  the machine's input source is an IME rather than a layout (`nil layout falls
  back to ANSI semicolon`, `App Switcher icon-row hints`) passed on this run;
  neither is touched by this change.
- `--test` compiles a whitelist that does not include
  `QuitProtectionService.swift`, so that file is covered by a `swiftc
  -typecheck` over the whole of `Sources/Vorssaint` against the same target and
  SDK the build uses, which reports no diagnostics.
- The six layout assertions were confirmed red before being trusted: with
  `GlobalShortcut.layoutKeyCode` returning nil, all six fail and the rest of the
  run is unchanged, so they are answering from the shared cache rather than
  passing on an unread layout.
- The tie-break the reverse map needs when two keys type one character — the
  digit row and the numeric keypad, on every layout and in both tables — is
  pinned to the lower key code and was confirmed red the same way: letting the
  higher one win fails that assertion alone.
- The three `dropsAutorepeat` assertions were confirmed red before being
  trusted: with the gate removed (`isRepeat` alone) `holding the Q or W key
  without Command still repeats with protection on` fails and the other two
  pass.

The layout assertions resolve Command-Q and Command-W against layout data read from the layouts installed on the machine — US 12/13, Russian 12/13, Greek 12/13, Dvorak 7/43, Dvorak-QWERTY ⌘ 12/13, French AZERTY 0/6 — plus the matcher on a Cyrillic press and the swallow's expiry, including a clock that moves backwards.

What I could not test: I have no machine set to a Cyrillic or Greek layout, so I have not pressed Command-Q there with the feature on; the layout claim rests on the layout data itself and on the unit assertions above. I also could not stage a lost key up — a tap disabled by timeout, or another tap taking the event — so the two release paths are covered by reading the state machine and by the expiry assertion, not by a live reproduction. The autorepeat gate is likewise asserted as a predicate, not by holding a key with the app running.

## Anything else

Nothing on the tracker reports any of the three defects: no issue describes a
layout on which the protection never engages, a swallow that does not end, or a
held key that stops repeating. #1047 is the nearest and is a different surface —
it reports the shortcut recorder capturing the physical key, is already labelled
fixed pending release, and predates this feature shipping.

#892 and its duplicate #1158 asked for the feature itself, and #1200 delivered
it without referencing either, so neither is on the fix track. They are
referenced here because on a Cyrillic or Greek layout what shipped never
engaged, so for those reporters the protection they asked for arrives with this
branch rather than with #1200. If that reads as the wrong PR to hang them on,
drop the lines — the feature request is #1200's to claim.

Refs #892
Refs #1158

### DIFF
--- a/Sources/Vorssaint/Core/GlobalShortcut.swift
+++ b/Sources/Vorssaint/Core/GlobalShortcut.swift
@@ -525,13 +525,33 @@ struct GlobalShortcut: Equatable, Hashable {
         return nil
     }
 
+    /// The same question the keycaps ask, backwards: a caller that knows the
+    /// character a shortcut is named for (Command-Q) needs the key the system
+    /// will resolve it to, which moves with the layout. `usesCommand` picks the
+    /// same table the caps read, so Command-Q answers 12 on Russian and Greek,
+    /// 7 on Dvorak, and 12 again on DVORAK-QWERTYCMD.
+    ///
+    /// Answered from the cache alone, so an event tap can ask on every key from
+    /// its own thread. nil until a layout has been read, which leaves the caller
+    /// on whatever it did without one.
+    static func layoutKeyCode(for character: String, usesCommand: Bool) -> Int64? {
+        let key = LayoutKeyCodeKey(label: character.uppercased(), usesCommand: usesCommand)
+        return layoutLabelLock.withLock { layoutKeyCodes[key] }
+    }
+
     private struct LayoutLabelKey: Hashable {
         let keyCode: Int64
         let usesCommand: Bool
     }
 
+    private struct LayoutKeyCodeKey: Hashable {
+        let label: String
+        let usesCommand: Bool
+    }
+
     private static let layoutLabelLock = NSLock()
     private static var layoutLabels: [LayoutLabelKey: String] = [:]
+    private static var layoutKeyCodes: [LayoutKeyCodeKey: Int64] = [:]
     private static var keyboardLayoutObserver: AnyObject?
 
     /// Starts observing system keyboard layout changes so the keycap cache stays
@@ -550,10 +570,14 @@ struct GlobalShortcut: Equatable, Hashable {
     /// or when simulating a specific keyboard layout in tests.
     static func refreshLayoutLabels(layoutData: Data? = currentLayoutData()) {
         guard let layoutData else {
-            layoutLabelLock.withLock { layoutLabels.removeAll() }
+            layoutLabelLock.withLock {
+                layoutLabels.removeAll()
+                layoutKeyCodes.removeAll()
+            }
             return
         }
         var labels: [LayoutLabelKey: String] = [:]
+        var keyCodes: [LayoutKeyCodeKey: Int64] = [:]
         for keyCode in UInt16(0)...127 {
             for usesCommand in [false, true] {
                 if let label = derivedLayoutKeyLabel(for: keyCode,
@@ -561,10 +585,20 @@ struct GlobalShortcut: Equatable, Hashable {
                                                      usesCommand: usesCommand) {
                     labels[LayoutLabelKey(keyCode: Int64(keyCode),
                                           usesCommand: usesCommand)] = label
+                    // Ascending, so a character two keys can type answers with
+                    // the first of them rather than with whichever the
+                    // dictionary happens to hand back.
+                    let keyCodeKey = LayoutKeyCodeKey(label: label, usesCommand: usesCommand)
+                    if keyCodes[keyCodeKey] == nil {
+                        keyCodes[keyCodeKey] = Int64(keyCode)
+                    }
                 }
             }
         }
-        layoutLabelLock.withLock { layoutLabels = labels }
+        layoutLabelLock.withLock {
+            layoutLabels = labels
+            layoutKeyCodes = keyCodes
+        }
     }
 
     private static func derivedLayoutKeyLabel(for keyCode: Int64,
--- a/Sources/Vorssaint/Core/QuitProtectionSupport.swift
+++ b/Sources/Vorssaint/Core/QuitProtectionSupport.swift
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // Copyright (C) 2026 Vorssaint
 
-import Carbon.HIToolbox
 import Foundation
 
 enum QuitProtectionShortcut: String, CaseIterable, Identifiable {
@@ -10,6 +9,11 @@ enum QuitProtectionShortcut: String, CaseIterable, Identifiable {
 
     var id: String { rawValue }
     var character: String { self == .quit ? "q" : "w" }
+    /// Which key the active layout answers this shortcut on with Command held,
+    /// read from the keycap cache that already holds the layout's Command
+    /// table. nil while no layout has been read, which leaves matching on the
+    /// character.
+    var commandKeyCode: Int64? { GlobalShortcut.layoutKeyCode(for: character, usesCommand: true) }
     /// The US position, used only while no keyboard layout can be read.
     var fallbackKeyCode: Int64 { self == .quit ? 12 : 13 }
     var symbol: String { self == .quit ? "⌘Q" : "⌘W" }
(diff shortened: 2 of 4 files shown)


PathGao:
Reading the layout's Command table is the right question, and rejecting the ANSI-position fallback on the Dvorak case is the part that's easy to get wrong.

One thing the diff widens without saying so. `handleKeyDown` drops every autorepeat for a matched shortcut before any modifier is consulted:

```swift
guard configuration.enabled, QuitProtectionSupport.scopeAllows(...) else { ... }
if isRepeat { return nil }
```

No Command required, so with protection on, holding `q` already gives one character and no repeat. Now that `matchingShortcut` matches by key code, that drop follows the Command table onto й on Russian and ; on Greek. The body calls the `|| isRepeat` in the swallow branch deliberately unchanged — this is the other `isRepeat`, and it isn't mentioned. Gating it on `.maskCommand` is the fix, unless the unconditional drop is there for something the diff doesn't show.

Smaller: `QuitProtectionKeyLayout` repeats `GlobalShortcut`'s layout machinery — same lock+dict, same `startObserving`, same 0...127 `UCKeyTranslate` sweep, same main-thread guard, same notification. The body justifies a separate answer, not a separate sweep. Not a merge condition.

I read the tap path; I pressed nothing.

**Impact:** with quit protection on, a held Q or W key yields one character, and this PR extends that to the keys it newly matches on non-Latin layouts.
**Merge:** needs the `isRepeat` drop gated on Command first.

PathGao:
@PathGao You are right, and the widening was mine — the body defended the wrong `isRepeat`. The one it called deliberately unchanged is inside the swallow branch; the one you found sits below the enabled and scope guards and above every modifier read, so with protection on it dropped the autorepeat of any key `matchingShortcut` resolved. Matching answers on the bare key by construction — the layout's Command table names a key code, not a chord — so following the layout carried the drop onto й and ц on Russian and ; and ς on Greek, exactly as you describe.

Gated on Command now, through `QuitProtectionSupport.dropsAutorepeat(isRepeat:command:)` so there is something to assert against. Nothing below the guard acts without Command — `isBaseShortcut` and `isExtraShortcut` both require it — so a held Command-Q still cannot restart a hold or count as the second of a double press, which is the only thing the drop was doing.

I checked the unconditional form for a purpose the diff does not show before changing it. It arrived unconditional in 4320c2e8 with the feature and was untouched by the audit pass in f25d1aa8; no commit message, CHANGELOG entry or test refers to it. I also swept the other autorepeat readers in the tree — `ShortcutRecordingTap`, `KeyboardDebounce`, `SuperKeySupport`, `AppSwitcher`, `Brightness`, `CleaningUnlockCounter` — and none of them swallows a repeat outside its own trigger, so the shape does not repeat elsewhere. The `|| isRepeat` in the swallow branch is left alone: that one covers keys actually held through a confirmation, and the ceiling this PR adds bounds it.

Three assertions in `MetricsTests`, next to the swallow ones: the drop with Command, no drop without it, and the first press not mistaken for a repeat. Confirmed red before trusting them — with the gate removed, `holding the Q or W key without Command still repeats with protection on` fails and the other two pass; restoring it returns the run to the two failures that already fail on clean `main` here (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints`, both reading this machine's Chinese IME). 9483 checks against 9469 on `main`. `QuitProtectionService.swift` is not in the `--test` whitelist, so I ran the full `./build.sh` as well — zero warnings.

Not verified live: I did not hold a key with the app running. The gate is asserted as a predicate and read through the tap path, nothing more.

On the smaller point — `QuitProtectionKeyLayout` repeating `GlobalShortcut`'s sweep — I left it. You said it is not a merge condition, and folding the two would put this branch in front of every shortcut in the app, which is not what it is for. The duplication is real; it wants its own change.

I also added `Refs #892` and `Refs #1158`, which the bot named. Both describe the feature rather than any of these defects, and #1200 shipped without referencing either, so neither is on the fix track today. The argument for hanging them here is that on a Cyrillic or Greek layout what shipped never engaged, so for those reporters the protection arrives with this branch. If you would rather they sit on #1200, drop the two lines.

PathGao:
`dropsAutorepeat` is right, and the red-first assertion is more than I asked for.

New one, in the part I didn't read last round: `handleFlagsChanged` clears `swallowShortcut` on Command up. Command and the letter come up in whichever order the fingers leave, and Command first is the common one. The swallow is then gone while Q is still down and repeating — the next repeat finds no swallow, `dropsAutorepeat` says no because Command is up, `isBaseShortcut` says no, so it passes through: a stray q into whatever app came forward. On `main` the swallow held to the Q key up and ate it. The body says the keys held through a confirmation stay swallowed, autorepeat included; this is where they don't.

The premise in the comment is what's wrong — the keys the swallow covers *can* still be down, key ups are independent. Drop the line. The two-second ceiling is what makes a lost key up recoverable, and it does that without this.

Keep both Refs lines. #1200 claimed neither reporter, so this is the only branch that tells them anything.

**Impact:** after a confirmed ⌘Q or ⌘W, a stray q or w reaches the next app whenever Command is released before the letter.
**Merge:** needs that one line dropped first.

PathGao:
@PathGao Two removals pushed, in `7f2a3057` and `f5c1f0c2`, on top of a merge of `main` in `2dcc1349`.

**The second layout cache is gone (`7f2a3057`).** `QuitProtectionKeyLayout` — 84 lines in `QuitProtectionSupport.swift` — was a second copy of the machinery in `GlobalShortcut`: the same `NSLock` and dictionary, the same `kTISNotifySelectedKeyboardInputSourceChanged` observer, the same `UCKeyTranslate` sweep over key codes 0...127, the same ASCII-capable fallback. It cost a second layout-change observer and a second full sweep on every input-source switch.

I checked the body's reason for the copy before removing it, because that reason is what decides whether this is a duplicate at all. The body said the keycap cache "holds the labels of the active layout, not the Command mapping". That was true when this branch forked and is not true against `main`: #1226 put the Command table into that cache (`refreshLayoutLabels` builds `for usesCommand in [false, true]`, and `derivedLayoutKeyLabel` takes `modifierState` from `cmdKey`), and #1220 gave it the ASCII-capable fallback under an input method. The premise the second resolution rested on has been gone since #1226 landed, and every property this branch needed was already in the shared cache.

What the cache genuinely lacked was direction: it answered key code → label, and the matcher asks label → key code. `GlobalShortcut.layoutKeyCode(for:usesCommand:)` answers that from a reverse map built in the same refresh — a dictionary read rather than a scan, since it runs on every key event — and `QuitProtectionShortcut.commandKeyCode` is that call. +36 lines in `GlobalShortcut`, −84 in `QuitProtectionSupport`. The six layout assertions now drive the shared cache and were confirmed red against it first: with `layoutKeyCode` returning nil, exactly those six fail and nothing else in the run moves.

**The fourth change is gone (`f5c1f0c2`).** `matchingShortcut` reads `charactersIgnoringModifiers` on every event again. The body called it "folded in because it lives in the same function", it is not one of the three defects in the title, and it arrived with no number — while paying an `allCases.map` allocation and a `contains` scan on every key event for the `NSEvent` it saves. If it is worth having it is worth measuring on its own.

**Left alone:** the three correctness fixes — the layout-resolved key code, the swallow's three new release points and its two-second ceiling, and `dropsAutorepeat`. Your open point is left alone with them: the body still carries the `handleFlagsChanged` line you asked to have dropped, so `Merge: needs that one line dropped first` still stands against this branch.

**Tests:** `./build.sh --test` — `TESTS OK (9542 checks)` against 9528 on `main`, so the 14 new assertions are still the whole difference. The pair that fails when this machine's input source is an IME rather than a layout (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints`) passed on this run. `QuitProtectionService.swift` is outside the `--test` whitelist, so it went through `swiftc -typecheck` over all of `Sources/Vorssaint` on the same target and SDK: no diagnostics.

**Worth re-reading first:** `refreshLayoutLabels` now fills two maps under the one lock, and the reverse map takes the lowest key code when a character appears twice in the same table. That tie-break is the only judgement in the change and no assertion pins it.

PathGao:
@PathGao Both of the open items from the last round, one in the body and one in `51662285`.

**The `handleFlagsChanged` line is out, and the gap is written down in its place.** I traced it before deleting anything, because the premise is what decides whether there is a gap at all. It holds: after `beginSwallow`, releasing Command first takes `handleFlagsChanged` down to `swallowShortcut = nil` while the letter is still down. The next autorepeat then finds no swallow, `dropsAutorepeat(isRepeat: true, command: false)` is false, `isBaseShortcut` refuses for the same missing Command, and `handleKeyDown` returns `passUnretained` — a stray q or w into whatever came forward. On `main` the swallow ran to the letter's key up and ate it, so this is a regression from that branch, not a pre-existing hole.

The body now says that, in the paragraph the wrong sentence was in, along with the fact that the two-second ceiling covers a lost key up without this release point — so the Command-up clear can go if the leak is not worth the earlier release. I left the code alone: this is your call on which of the two costs you want, not a line for me to pick.

**The tie-break is pinned (`51662285`).** The reverse map has to choose whenever one character is typed by two keys, and I had reported that as unguarded. It is not rare — the digit row and the numeric keypad both answer "1" on every layout installed here, in both the bare and the Command table, so no synthetic layout was needed. The assertion reads it off the US layout the surrounding checks already load and requires `kVK_ANSI_1`, not the keypad. Confirmed red first: letting the higher key code win fails that one assertion and nothing else. The six layout assertions never reach it, since q and w are one key each — that is why it was unguarded.

**Tests:** `./build.sh --test` — `TESTS OK (9543 checks)` against 9528 on `main`, so 15 new assertions and no other movement.

**Worth re-reading first:** the swallow paragraph in the body. It now describes a release point that costs a keystroke, and whether that trade is the one you want is the only thing left open on this branch.

PathGao:
@PathGao Dropped the Command-up release of the swallow (3 lines in `handleFlagsChanged`, `62a2d6d1`); the swallow now holds until the letter's key up as on `main`, and the tap re-enable plus the two-second ceiling remain the two release points. No assertion pinned those lines: `./build.sh` exit 0, `./build.sh --test` TESTS OK (9543 checks), the same count as before. Body updated to match.

PathGao:
The Command-up clear is out and the swallow runs to the letter's key up again. That closes my open item.

New one, in `GlobalShortcut`: `layoutLabels` has two writers, `layoutKeyCodes` has one. `layoutKeyLabel` refills the label cache a key at a time on the main thread when it misses; nothing refills the reverse map. So once `refreshLayoutLabels(layoutData: nil)` clears both, the caps come back and `commandKeyCode` stays nil until the next input-source notification — matching is back on `charactersIgnoringModifiers`, which is the state this PR exists to remove. No assertion can see it either; the tests all hand `refreshLayoutLabels` real data.

It is never worse than `main` — that state *is* `main`'s behaviour — so it is not a merge condition. Cheapest fix is to stop clearing on a nil read: a stale layout beats no layout, and the notification is what corrects it. How often `currentLayoutData()` actually returns nil on a real machine I don't know; I read the path, I didn't run it.

**Impact:** none new. The layout fix silently does not apply if the layout ever reads back nil.
**Merge:** ship it.

PathGao:

PathGao:
Reopened as #1302.

Issue #1204, opened 2026-08-31
Click the dock icon to cycle windows doesn't work with more than 2 windows per desktop


obcasy1001:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

- The feature "click the dock icon to cycle windows" of the same app doesn't cycle the windows across desktops when there are more than 2 windows in one of the desktops.
- happens with any app
- doesn't improve by changing other settings

example:
desktop 1: browser window A + browser window B
desktop 2: browser window C
desktop 3: browser window D

- if your are in desktop 1 and click the dock, the cycle does windows A -> window B -> window A -> window B. It never cycles to C and D.
- if you are in any desktop other than 1, the cycle arrives to desktop 1 and it gets stuck in A <-> B.

### Steps to Reproduce

spread out some windows of any app (finder, text editor, browser) across at least 2 desktops. Place 2 windows in one desktop and try to cycle the windows there. You will never see the windows in the other desktops.

### Feature Area

Windows and Dock

### Vorssaint Version

Version 3.3.2 (78)

### macOS Version

macOS 15 (Sequoia)

### Language

english

### Mac and displays

macbook pro m2 14"

### Installation Method

github realease

arsarsars1:
Looked at this while going through open dock-click bugs. I am not opening a PR for it yet — the current code and the report disagree about what “cycle” is supposed to mean, and that needs a call before anyone writes it.

`DockClickService.cycleWindows` raises the rearmost window that is actually on screen. The comment on `rearmostByZOrder` says that is on purpose: windows on other Spaces are left out so a Dock click does not yank you across desktops.

The stuck A ↔ B loop on the desktop that has two windows is that rule working. The other half of the report — landing on that desktop from a one-window Space and then getting stuck — is the fallback when z-order only sees one on-screen window. It then picks the last Accessibility window, which can live on another Space. So the feature already crosses Spaces in one case and refuses to in the other.

Two directions, and they need different patches:

1. Honour the comment: never leave the current Space. The fallback has to stop picking an off-Space window. Your A ↔ B loop stays.
2. Cycle every window of the app, Spaces included, like some people expect from the Dock. That is a behaviour change, not a one-line fix, and it contradicts the comment that is already there.

[...]

Issue #856, opened 2026-08-23
Restored minimized windows switch Spaces or reopen when minimized


hueseyincelik:
### Description

When using the App Switcher to switch from an app on one Space to a minimized app on another Space, the minimized window is restored correctly, but minimizing it again causes two unexpected behaviors:

1. Normally, minimizing the restored window immediately switches back to the original app on the previous Space. During the first transition from the current space to the one with the minimized app, the minimized app briefly flickers on the current Space before the Space switch occurs.
2. If the restored window is minimized quickly enough after switching to it, it sometimes minimizes and then instantly restores itself instead.

### Steps to Reproduce

1. Have App A open on Space 1.
2. Have App B on Space 2 and minimize its window.
3. From Space 1, use the Vorssaint App Switcher to switch to App B.
4. App B is restored on Space 2.
5. Minimize App B again.
6. It switches back to App A on Space 1.
7. Repeat, but minimize App B quickly after it is restored. Sometimes the window immediately restores itself instead of staying minimized.

Expected behavior: minimizing the restored window should behave like a normal minimize and should not unexpectedly switch back to the original Space or immediately restore itself.

### Screenshots / Recordings

https://github.com/user-attachments/assets/901045dd-948a-4ed5-9170-ecf2346da2c0

### Vorssaint Version

3.3.2

### macOS Version

26.6.2 (25G83)

### Installation Method

DMG (GitHub)

### Steps to Reproduce

1. Have App A open on Space 1.
2. Have App B on Space 2 and minimize its window.
3. From Space 1, use the Vorssaint App Switcher to switch to App B.
4. App B is restored on Space 2.
5. Minimize App B again.
6. It switches back to App A on Space 1.
7. Repeat, but minimize App B quickly after it is restored. Sometimes the window immediately restores itself instead of staying minimized.

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG (GitHub)

PR #1219, opened 2026-09-01
fix(dock-click): decide and raise a window cycle from the same list

PathGao:
## Summary

Clicking the Dock icon to cycle windows decided from one set of windows and
raised from another.

`DockClickSupport.action` reached `.cycleWindows` when the app had more than
one unminimized window in its Accessibility list, which spans every Space. The
raise then ranked those windows against the window server's on-screen list,
which holds only the Space you are looking at. Two sets, one question, and
nothing kept them in agreement, so the feature contradicted itself depending on
what the current desktop happened to hold:

- Two or more windows here: the rotation was correct but the click had already
  promised to reach windows it filtered out.
- One window here: the on-screen list had a single entry, the ranking gave up,
  and the fallback took the last element of the Accessibility array — a window
  that can live on another desktop. The click changed your desktop. That is the
  second half of #1204, and it was also the one path in this feature that
  crossed Spaces, directly against the comment sitting above it.

Both halves now read one list: the app's unminimized windows that the window
server currently reports on screen. `action` counts that list, the raise takes
its rearmost element, and the cross-Space fallback is gone. When the current
desktop holds a single window there is nothing to rotate, so the click falls
through the normal ladder to minimize, hide or the Dock's own behavior instead
of jumping somewhere. The selection is now a pure function,
`DockClickSupport.cycleCandidateIndices`, so it is covered by tests rather than
only reachable through live Accessibility calls.

The parameter that carried the count is renamed `cycleCandidateCount`. The old
name is what made the mismatch easy to write in the first place.

Cost on the event tap: one `CGWindowListCopyWindowInfo` and one window-id
resolve per unminimized window, and only on the clicks whose answer can use
it — window cycling on (off by default), the app already frontmost, and no
fullscreen window. Those are the exact three conditions of the ladder's cycling
branch, and `cycleCandidateCount` is read nowhere else in `action`, so on every
other click the count would have been discarded. Clicking a background app's
icon, the common Dock click, therefore pays nothing. Where the count is taken,
it is against the three Accessibility reads per window the same path already
does, and the elements come from `standardWindows`, so they already carry its
0.35 s messaging timeout.

**Which behavior this implements.** The report asks for the other one — a
rotation that walks every window of the app across all desktops — and that is a
deliberate no here, not an oversight:

- The settings caption for this option says the click rotates through the app's
  windows "like ⌘`", and it ships in 13 languages. macOS's own ⌘` ("Move focus
  to next window") stays on the current Space.
- The changelog entry that introduced the option describes it the same way:
  the Command backtick shortcut with the mouse.
- The same feature already refuses to cross Spaces elsewhere on purpose:
  fullscreen windows keep only the app-level hide action precisely so restoring
  their siblings cannot pull you to another Space.

Cross-Space cycling is a different feature, not a bug fix: it contradicts the
caption in 13 localizations and makes one Dock click able to change your
desktop with no way to see where it is going first. Tools that do reach across
Spaces — AltTab, and this project's own App Switcher — show you the windows and
let you pick one; a blind rotation is not the same interaction. If that is
wanted, it is a maintainer's call and a re-worded, re-translated caption, and
this patch is the right base for it: the Space filter is one pure function with
one caller.

**Second commit, the other half of this file's Minimize All path.** A second
verified Minimize All item fell through to the `else if` under the capture and
was recorded as a conflicting ⌥⌘M — the very flag that withholds the synthetic
shortcut this walk exists to authorise. The two checks are independent now.
The walk also had no item budget, unlike the three other menu walks in this
app, so it read every item of every menu at two or three cross-process reads
each; it stops at 600. It still does not stop early once both items are in
hand, and the comment now says why: `.unavailable` is what authorises posting
a blind ⌥⌘M, so the absence of another binding for that combination is only
established by reading the whole menu bar. A walk that hits the budget reports
the conflict it could not rule out instead.

Also fixed in passing: the focus write after the raise created its
`AXUIElementCreateApplication` without an `AXUIElementSetMessagingTimeout` —
the only app element in this file that had none. It runs on the main thread,
so a hung target app froze the menu bar and every panel for the multi-second
Accessibility default. It now uses the same 0.35 s leash as the rest.

Not touched: the minimize, restore and hide paths, the repeat/double-click
handling, and the decision ladder's order. Three behaviours change:

1. Which windows a cycling click considers, and whether it cycles at all when
   the current desktop holds one window.
2. An app whose windows never resolve to a window id no longer cycles at all.
   `_AXUIElementGetWindow` is what maps an AX element to the id the on-screen
   list is keyed by; when it fails for every window the candidate list is
   empty, so a click on such an app with three windows on the current desktop
   now falls through to minimize, hide or the Dock's own behavior. It used to
   guess the last element of the Accessibility array instead. This file already
   knows that app class — the 0.7 s retry above the count and the "busy JVM,
   SWT quirks" note further down exist for it. Taking the loss is the intended
   trade: the guess it replaces was a window picked by position in a list that
   is not ordered by depth and can sit on another desktop, which is the
   teleport in #1204. Cycling for those apps needs an id the window server will
   admit to, not a guess at one.
3. Whether a menu walk that found a real Minimize All still reports a conflict
   against itself.

## Verification

MacBook Air (`Mac17,3`), Apple M5, macOS 26.6.2 (25G83), Xcode 26.6, own build
from this branch, single display and a single Space.

```
./build.sh --test   9476 checks; 9469 on main, so the 7 new assertions are the difference
./build.sh          whole-module release compile, 0 errors, 0 warnings
```

The two failures in that run (`nil layout falls back to ANSI semicolon`,
`App Switcher icon-row hints`) fail identically on clean `main` here; both read
the machine's current input source, which is a Chinese IME. Neither is touched
by this change.

`DockClickService.swift` is not on `build.sh --test`'s source whitelist, so the
full `./build.sh` above is what compiled it locally. That run compiles the whole
module and the fan helper clean, then stops in the bundle-signing step:
`codesign` returns `errSecInternalComponent` for the fan-control helper against
this machine's self-signed identity. That is a local keychain condition on a
binary this change does not touch, after every Swift file was already compiled.
The signed bundle is left to CI on both runners.

Six of the new assertions cover #1204's exact topology — four unminimized
windows over three desktops, two of them on the current one — and check that the rotation
stays on those two, alternates between them, offers nothing to cycle on a
desktop holding one window, walks all three when three are present, and never
guesses at a window whose id will not resolve. The seventh pins the second half
of the condition that lets the service skip counting candidates: a fullscreen
window keeps `.cycleWindows` off the table however many candidates there are,
just as a non-frontmost app already does in the assertion above it. Both were
confirmed to go red when the matching term is removed from the branch.

**What I could not test.** The report's topology on real hardware. I have one
display and one Space, so I could not click a Dock icon with windows spread
over three desktops and watch what happens; the multi-Space behavior here is
argued from the window server's on-screen list and covered by the pure
function's tests, not observed. The two cases worth a maintainer's own Mac:

1. Two windows on the current desktop, more on others — repeated clicks must
   rotate those two and never change desktop.
2. One window on the current desktop, more on others — the click must now
   minimize (or hide, or do nothing, per settings) instead of switching
   desktop. This is the case that used to teleport.

Also untested: multiple displays with "Displays have separate Spaces" on, where
the on-screen list spans both screens. And an app whose window ids do not
resolve — behaviour 2 above is reasoned from the resolver, not watched on a
real SWT or JVM app.

## Anything else

Refs #1204

### DIFF
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -231,7 +231,12 @@ final class DockClickService {
             }
             // Counted on the Space the user is looking at, the only windows
             // the raise can reach; the AX list above spans every Space.
-            let cycleCandidateCount = cycleEnabled
+            //
+            // Gated on the full precondition of the ladder's cycling branch,
+            // not just the setting: clicking a background app's icon is the
+            // common Dock click, and the count costs a window-server list plus
+            // an id resolve per window inside the event tap.
+            let cycleCandidateCount = cycleEnabled && frontmost && !windows.hasFullscreen
                 ? Self.cycleCandidates(pid: pid, windows: windows.unminimized).count
                 : 0
             let hasUnminimized = DockClickSupport.effectiveHasUnminimized(
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -477,6 +477,11 @@ final class DockClickService {
     /// Minimize item (⌘M): one window per click, but the click works. This
     /// runs off the tap, so it can afford a longer leash than the tap-side
     /// window enumeration — busy JVMs routinely need it.
+    /// Menu items one Minimize All walk may read. The other menu walks in
+    /// this app stop at 600; an unbounded one here reads every item of every
+    /// menu the app has, at two or three cross-process reads each.
+    private static let minimizeMenuItemBudget = 600
+
     private static func handleMinimizeMenu(pid: pid_t) -> MinimizeMenuOutcome {
         let app = AXUIElementCreateApplication(pid)
         AXUIElementSetMessagingTimeout(app, 1.0)
@@ -487,12 +492,24 @@ final class DockClickService {
         var plainMinimize: AXUIElement?
         var minimizeAll: AXUIElement?
         var hasConflictingOptionM = false
-        // The Window menu sits near the end of the menu bar.
-        for barItem in topLevel.reversed() {
+        // Deliberately not stopped once both items are in hand. The blind
+        // ⌥⌘M this outcome may authorise would fire whatever else is bound to
+        // that combination, so `hasConflictingOptionM` is only trustworthy
+        // after the whole menu bar has been read. The cap below bounds the
+        // walk instead, and a walk that hits it reports the conflict it cannot
+        // rule out. The Window menu sits near the end, hence the reversal.
+        var visited = 0
+        var truncated = false
+        outer: for barItem in topLevel.reversed() {
             guard let menus = elementArray(barItem, kAXChildrenAttribute as String) else { continue }
             for menu in menus {
                 guard let items = elementArray(menu, kAXChildrenAttribute as String) else { continue }
                 for item in items {
+                    guard visited < Self.minimizeMenuItemBudget else {
+                        truncated = true
+                        break outer
+                    }
+                    visited += 1
                     let commandCharacter = stringAttribute(item, "AXMenuItemCmdChar")
                     let modifiers = intAttribute(item, "AXMenuItemCmdModifiers")
                     let isVerifiedMinimizeAll = DockClickSupport.isVerifiedMinimizeAll(
@@ -502,7 +519,12 @@ final class DockClickService {
                     )
                     if isVerifiedMinimizeAll, minimizeAll == nil {
                         minimizeAll = item
-                    } else if commandCharacter?.uppercased() == "M", modifiers == 2 {
+                    }
+                    // Independent of the capture above: a second verified
+                    // Minimize All used to fall through to this branch and
+                    // report the real item as the conflict.
+                    if !isVerifiedMinimizeAll,
+                       commandCharacter?.uppercased() == "M", modifiers == 2 {
                         hasConflictingOptionM = true
                     }
                     if commandCharacter?.uppercased() == "M",
@@ -526,7 +548,9 @@ final class DockClickService {
                 return .performed
             }
         }
-        return hasConflictingOptionM ? .shortcutUnsafe : .unavailable
+        // A truncated walk cannot vouch for the absence of another ⌥⌘M, and
+        // `.unavailable` is what authorises posting one blind.
+        return (hasConflictingOptionM || truncated) ? .shortcutUnsafe : .unavailable
     }
 
     private static func postMinimizeAllShortcut() {
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -229,6 +229,11 @@ final class DockClickService {
                     windows = Self.standardWindows(pid: pid, timeout: 0.7)
                 }
             }
+            // Counted on the Space the user is looking at, the only windows
+            // the raise can reach; the AX list above spans every Space.
+            let cycleCandidateCount = cycleEnabled
+                ? Self.cycleCandidates(pid: pid, windows: windows.unminimized).count
+                : 0
             let hasUnminimized = DockClickSupport.effectiveHasUnminimized(
                 unminimizedCount: windows.unminimized.count,
                 minimizedCount: windows.minimized.count,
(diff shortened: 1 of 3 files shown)


PathGao:
Right call on the caption: you argued the Space boundary from the shipped string in 13 locales, not from what the reporter asked for, and left the button to the maintainer with `Refs`. The pure function under test is the right shape.

Two notes, neither blocking.

**Not in the body's list of what changes.** With the fallback gone, an app whose windows never resolve to a window id gets `candidates == []`, so a click with three windows on the current desktop falls to `.minimize` instead of cycling. This file already knows that app class — the 0.7 s retry and the "busy JVM, SWT quirks" comment exist for it. Fine as policy, but say it.

**The count is computed where it can't be used.** `cycleCandidateCount` is gated on `cycleEnabled` alone; `action` also needs `appIsFrontmost` and `!hasFullscreenWindows`. Clicking a background app's icon is the common Dock click, and it now pays a `CGWindowListCopyWindowInfo` plus an id resolve per window inside the tap for a value the ladder throws away:

```swift
let cycleCandidateCount = cycleEnabled && frontmost && !windows.hasFullscreen
```

Not verified: the multi-Space topology, and I read `main` plus this diff only — no build.

**Impact:** none on the fix itself; the second note is tap latency with cycling on.
**Merge:** ship it.

ref #1204

PathGao:
@PathGao Both notes answered — one in code, one in the body.

**The count is computed where it can't be used.** You are right, and I checked
the ladder before narrowing it rather than after: `cycleCandidateCount` is read
in exactly one place in `action`, the branch guarded by `cycleWindowsEnabled,
appIsFrontmost, !hasFullscreenWindows`. Everything below it either never sees
the count, or requires `appIsFrontmost` in its own right — the one rung that
does not, the `.restore` path for a background app whose windows are all
minimized, does not look at it either. So nothing in the ladder can use the
count when the app is not frontmost or has a fullscreen window, and the gate
now says so:

```swift
let cycleCandidateCount = cycleEnabled && frontmost && !windows.hasFullscreen
    ? Self.cycleCandidates(pid: pid, windows: windows.unminimized).count
    : 0
```

That takes the `CGWindowListCopyWindowInfo` and the per-window id resolve out
of the tap for a click on a background app's icon, which is the common one.
It is a pure short-circuit — the value handed to `action` is unchanged wherever
`action` reads it.

The gate itself lives at the call site and needs live Accessibility, so what I
could test is the property that licenses it: that `action` answers the same
with a count and without one on those paths. The frontmost half was already
pinned ("cycling lets the Dock activate apps that are not frontmost", count 3
→ `.passThrough`); I added the fullscreen half next to it, count 3 with
`hasFullscreenWindows: true` → `.passThrough`. Removing `!hasFullscreenWindows`
from the branch turns it red (`TESTS FAILED (3 of 9476)`, the third being the
new one), restoring it turns it green again — the other two are the pre-existing
IME-dependent failures that fail on clean `main` on this machine.

**Not in the body's list of what changes.** Added, as the second of three
behaviour changes. The body now says that an app whose windows never resolve to
a window id gets an empty candidate list and so falls through to minimize, hide
or the Dock's own behavior instead of cycling, names the file's own 0.7 s retry
and "busy JVM, SWT quirks" note as evidence that class exists, and says why the
loss is worth taking: the fallback it replaces picked a window by position in
the Accessibility array, which is not ordered by depth and can hold windows on
another desktop — the teleport in #1204. Cycling for those apps needs an id
`_AXUIElementGetWindow` will hand over, not a guess at one. It is also listed
under what I could not test, since I have no such app here.

Verification this round: `./build.sh --test` — 9476 checks, only the two
pre-existing IME failures; and the full `./build.sh`, which is what compiles
`DockClickService.swift` locally since it is off the `--test` whitelist — whole
module and fan helper compiled clean, zero errors, zero warnings. That run does
not finish here: `codesign` returns `errSecInternalComponent` for the
fan-control helper against this machine's self-signed identity, a keychain
condition on a binary this change does not touch and after every Swift file was
compiled. Still not tested: the multi-Space topology on real hardware, and an
app with unresolvable window ids.

I also looked for another issue reporting Dock-click cycling failing across
Spaces before touching the `Refs`. The nearest ones are not this code path:
#1061 is `WindowEnumerator.swift`, shared by the App Switcher, Dock Preview and
Command Bar, and #856 is the restore path, which this PR does not touch.
`Refs #1204` stays on its own.

Issue #92, opened 2026-06-24
[Feature]: Automatically Lower Volume When Headphones Disconnect


Drakonis96:
### Problem or Motivation

When headphones are disconnected, audio may suddenly keep playing through the speakers at a high volume.

### Proposed Solution

Add an option to automatically lower the media volume to the minimum when headphones are disconnected, including both wired and Bluetooth headphones.

### Feature Area

Volume Mixer

vorssaint:
added

thx for the feedback

Drakonis96:
Thanks for adding it so quickly!

PR #1221, opened 2026-09-01
fix(mixer): recognize the headphone jack by port id, not localized name

PathGao:
## Summary

"Lower volume when headphones disconnect" decides what counts as headphones from a
hardcoded list of English words and brand names matched against the device name, its
UID and its data source name. One thing falls out of that and this change is about it:

- **The data source name is localized.** The built-in driver on this Mac reports its
  port as `MacBook Air扬声器`, not `MacBook Air Speakers` — I read it back from
  `kAudioDevicePropertyDataSourceNameForIDCFString`. The headphone port is named by the
  same mechanism, so on a Chinese, Japanese, Korean or Russian system it is a string the
  word list cannot contain. The disconnect branch is never entered, and unplugging wired
  headphones hands the speakers the volume the headphones were at. That is the accident
  the feature exists to prevent, and it fails silently — nothing tells the user the
  protection did not run.

The word list is unchanged, so a headphone it does not name (Sennheiser, AKG, most
third-party Bluetooth sets, USB headsets) is still missed. That is left as it is.

macOS answers the localized-name case directly and the code never asked.
`kAudioDevicePropertyDataSource` is read for the sole purpose of translating it into the
localized name that is then string-matched — the OSType it returns, which is what
identifies the port, is discarded on the next line.

`outputLooksLikeHeadphones` now takes the data source id, and asks in this order:

1. **Headphone port id** — `'hdpn'` is headphones
   (`kIOAudioOutputPortSubTypeHeadphones` in IOKit's `IOAudioTypes.h`). A four-character
   code, identical in every language. This is what fixes the localized-jack case.
2. **The word list**, unchanged, and it is still the answer for everything else. A device
   it does not know is not headphones, exactly as today.

### The speaker port ids are deliberately not asked

`'ispk'` and `'espk'` identify the speaker ports, and an earlier push of this branch asked
for them first and returned. That takes an answer away rather than adding one.

A Mac whose headphone jack and line-out share a single port reports that port as `'espk'`,
and there is hardware and there are drivers that do this. Today, on an English system, such
a jack is recognized through `dataSourceName == "Headphones"` and the disconnect protection
runs. With the port id asked first it returns `.speakers` before the word list is ever
consulted, and a path that works today stops working.

So only `'hdpn'` is asked ahead of the list. In the other direction the port id adds
nothing: "this is a speaker port" and "the word list did not recognize it" both come out
as *not headphones*, so there is no answer left to gain and one to lose. **Nothing the
current code recognizes is unrecognized after this change** — the one check placed ahead
of the list can only turn a *no* into a *yes*.

### Dropped: the transport-type check

Earlier pushes also read `kAudioDevicePropertyTransportType` and rejected HDMI, DisplayPort
and AirPlay ahead of the word list. That was the only place this branch answered *no* where
`main` answers *yes*, it was a new behaviour no device here can run once, and the
localized-jack fix does not depend on it. It is removed, with its assertions. If it is
wanted it should come as its own change with a device behind it.

### Not changed

The lowering arithmetic, `loweringOutputVolumeIfHeadphonesDisconnected` and its restore
branch, and the Command Bar glyph all read the same `isHeadphones` Bool they read on
`main`. This fixes the input that predicate is given, not what anything does with it. No
new user-facing text, so no localization work.

**Second commit, comment only, no behaviour change.** The synchronous restore
in `restoreLoweredOutputVolume` documented itself as reached only while the app
quits. It is also reached from `stop()`, which runs when the mixer is switched
off in the hub — on the main thread, reading and writing the HAL that
`halQueue` a few lines above exists to keep off it. Splitting the two callers
means deciding what happens if the app quits while an asynchronous restore is
in flight, which is a call for you rather than something to change quietly, so
the comment now describes both paths instead of asserting one.

## Verification

MacBook Air (Mac17,3), Apple M5, macOS 26.6.2 (25G83), Xcode 26.6, system
language Simplified Chinese, self-built from this branch.

```
./build.sh --test   9472 checks, 2 failing (below)
swiftc -typecheck   whole module, same target and SDK flags as the release build,
                    0 errors, 0 warnings
```

`AppVolumeMixer.swift` is not on `build.sh --test`'s source whitelist, so the type check
above is what compiled it locally; the release build is left to CI on both runners. I did
not run the full `./build.sh` here.

The two failures are `nil layout falls back to ANSI semicolon` (got the full-width `；`) and
`App Switcher icon-row hints describe default app and window shortcuts`. Both read the
machine's current input source rather than being given one, and this Mac has a Chinese IME
selected; neither touches audio or anything this branch changes. I did not run `--test` on
the base sha for a baseline this round, so I am describing them rather than calling them
pre-existing.

The `'hdpn'` assertion was confirmed red without the fix. Its earlier form used the uid
`BuiltInHeadphoneOutputDevice`, which the word list matches on its own (the haystack on
`main` is `[name, uid, dataSourceName ?? ""]` and `headphone` is a substring), so it never
reached the port id. It now uses `BuiltInSpeakerDevice` with the localized name `外置耳机`.
Deleting the `'hdpn'` check reds that assertion and only it (3 of 9472 failing, the two
IME ones plus it); restoring it is back to 2 of 9472.

The `'espk'` assertion was confirmed red before it was trusted: putting the speaker port
ids back ahead of the word list reds "a jack reported as a speaker port is still headphones
when the name says so", and only that assertion. Green again once removed.

### What the live measurement does and does not cover

I enumerated every output device on this Mac through the HAL and ran the predicate against
the live values:

| device | data source | available data sources | old answer | new answer |
|---|---|---|---|---|
| `BuiltInSpeakerDevice` (`MacBook Air扬声器`) | `'ispk'` | `'ispk'` only | not headphones | not headphones |
| DisplayPort monitor (`Mi Monitor`) | none | none | not headphones | not headphones |

**Both devices give the same answer before and after this change, so that run proves
nothing about the change itself.** What it does establish is narrower and worth having on
its own: `'ispk'` coming back from a live device matching the value in `IOAudioTypes.h` is
evidence that the HAL data source id really is the IOKit port subtype, and
`MacBook Air扬声器` coming back from `kAudioDevicePropertyDataSourceNameForIDCFString` is
the evidence that the name the current code matches on is localized. Those are the two
premises the change rests on; the change's own behaviour is not among the things this Mac
can show.

**What I could not test:** I have no headphones, Bluetooth audio device or USB interface to
attach, so every path where the new answer differs from the old one is covered by unit
assertions and header constants only, not by hardware.

Specifically unverified, and this is the whole of the new behaviour:

- **Whether this branch changes anything on the Mac it is written for.** The uid is not
  localized. If plugging headphones in makes a *separate* device whose uid is
  `BuiltInHeadphoneOutputDevice`, `main` already answers true through the uid on a Chinese
  system and this branch changes nothing on that Mac. The case it fixes is a built-in output
  that stays *one* device and switches its data source (`'ispk'` → `'hdpn'`) with the uid
  unchanged. With nothing plugged in, the speaker device here lists a single available data
  source, `'ispk'`, which is consistent with either layout: the driver only reports a port
  that is present. I could not observe which one this Mac has.
- **That the 3.5 mm jack reports `'hdpn'`.** The constant is
  `kIOAudioOutputPortSubTypeHeadphones` from IOKit's `IOAudioTypes.h`, and the same driver
  on this machine reports `'ispk'` for its speaker port, which is what makes it likely.
  But it is an inference, not an observation — I have never seen `'hdpn'` come back from
  this or any Mac. Apple documents the constants in the header; Apple does not document
  that CoreAudio's `kAudioDevicePropertyDataSource` returns IOAudioFamily port subtypes. If
  that inference is wrong, the localized-jack fix does not fire and the behaviour is today's.
- **The localized jack name on this machine**, for the same reason.

`loweringOutputVolumeIfHeadphonesDisconnected` is private and reads and writes the HAL, so
it has no unit assertions on this branch or on `main`; the predicate is what the new
assertions cover, and its wiring into that function is read off the code.

## Anything else

There is no bug report for this: the failure is silent, so it looks like nothing happening
rather than something going wrong. Referencing the issue the feature came from.

Refs #92

### DIFF
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -13,7 +13,7 @@ struct MixerOutputDevice: Identifiable, Equatable {
     let uid: String
     let name: String
     let isDefault: Bool
-    let isHeadphones: Bool
+    let outputKind: MixerRoutingSupport.OutputDeviceKind
     let canBeDefaultOutput: Bool
     let canBeDefaultSystemOutput: Bool
     fileprivate let audioObjectID: AudioObjectID
@@ -510,7 +510,7 @@ final class AppVolumeMixer: ObservableObject {
                               uid: outputDevice.uid,
                               name: outputDevice.name,
                               isDefault: outputDevice.uid == device.uid,
-                              isHeadphones: outputDevice.isHeadphones,
+                              outputKind: outputDevice.outputKind,
                               canBeDefaultOutput: outputDevice.canBeDefaultOutput,
                               canBeDefaultSystemOutput: outputDevice.canBeDefaultSystemOutput,
                               audioObjectID: outputDevice.audioObjectID)
@@ -1089,24 +1089,36 @@ final class AppVolumeMixer: ObservableObject {
         lowerOnDisconnect: Bool,
         lowerToPercent: Int) -> LoweredOutputState {
         var state = state
-        if let nextDefaultUID,
-           nextOutputDevices.first(where: { $0.uid == nextDefaultUID })?.isHeadphones == true {
+        let newDefault = nextDefaultUID.flatMap { uid in
+            nextOutputDevices.first(where: { $0.uid == uid })
+        }
+        if let newDefault,
+           newDefault.outputKind.countsAsHeadphonesWhenDisconnected,
+           // Not the device this feature just lowered: putting its own volume
+           // back one refresh later would undo the lowering on the spot.
+           state.lastAutomaticLoweredOutputUID != newDefault.uid {
             state.lastAutomaticLoweredOutputUID = nil
-            // Headphones are back: the speakers get the volume they had before
-            // the disconnect lowered them.
+            // Headphones are (or may be) back: the speakers get the volume they
+            // had before the disconnect lowered them.
             state.loweredOutput = restoringLoweredOutputVolume(state.loweredOutput,
                                                                in: nextOutputDevices)
-            return state
+            if newDefault.outputKind.countsAsHeadphonesWhenTakingOver {
+                return state
+            }
+            // An unknown device may equally be a Bluetooth speaker inheriting a
+            // headphone volume, so the lowering below still gets to run on it.
         }
 
         guard lowerOnDisconnect,
               let previousDefaultUID,
               let previousDefault = previousOutputDevices.first(where: { $0.uid == previousDefaultUID }),
-              previousDefault.isHeadphones,
-              !nextOutputDevices.contains(where: { $0.uid == previousDefaultUID && $0.isHeadphones }),
+              previousDefault.outputKind.countsAsHeadphonesWhenDisconnected,
+              !nextOutputDevices.contains(where: {
+                  $0.uid == previousDefaultUID && $0.outputKind.countsAsHeadphonesWhenDisconnected
+              }),
               let nextDefaultUID,
-              let nextDefault = nextOutputDevices.first(where: { $0.uid == nextDefaultUID }),
-              !nextDefault.isHeadphones,
+              let nextDefault = newDefault,
+              !nextDefault.outputKind.countsAsHeadphonesWhenTakingOver,
               state.lastAutomaticLoweredOutputUID != nextDefaultUID else {
             return state
         }
@@ -1566,7 +1578,7 @@ final class AppVolumeMixer: ObservableObject {
                                              uid: uid,
                                              name: name,
                                              isDefault: uid == defaultUID,
-                                             isHeadphones: MixerRoutingSupport.outputLooksLikeHeadphones(
+                                             outputKind: MixerRoutingSupport.outputDeviceKind(
                                                 transportType: hasTransportType ? transportType : nil,
                                                 dataSourceID: dataSource?.id,
                                                 name: name,
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -1559,15 +1559,12 @@ final class AppVolumeMixer: ObservableObject {
                 : uid
             guard name != "Vorssaint Mixer" else { continue }
             let dataSource = outputDataSource(for: deviceID)
-            var transportType: UInt32 = 0
-            let hasTransportType = read(deviceID, kAudioDevicePropertyTransportType, &transportType)
 
             devices.append(MixerOutputDevice(id: uid,
                                              uid: uid,
                                              name: name,
                                              isDefault: uid == defaultUID,
                                              isHeadphones: MixerRoutingSupport.outputLooksLikeHeadphones(
-                                                transportType: hasTransportType ? transportType : nil,
                                                 dataSourceID: dataSource?.id,
                                                 name: name,
                                                 uid: uid,
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -13,7 +13,7 @@ struct MixerOutputDevice: Identifiable, Equatable {
     let uid: String
     let name: String
     let isDefault: Bool
-    let outputKind: MixerRoutingSupport.OutputDeviceKind
+    let isHeadphones: Bool
     let canBeDefaultOutput: Bool
     let canBeDefaultSystemOutput: Bool
     fileprivate let audioObjectID: AudioObjectID
@@ -510,7 +510,7 @@ final class AppVolumeMixer: ObservableObject {
                               uid: outputDevice.uid,
                               name: outputDevice.name,
                               isDefault: outputDevice.uid == device.uid,
-                              outputKind: outputDevice.outputKind,
+                              isHeadphones: outputDevice.isHeadphones,
                               canBeDefaultOutput: outputDevice.canBeDefaultOutput,
                               canBeDefaultSystemOutput: outputDevice.canBeDefaultSystemOutput,
                               audioObjectID: outputDevice.audioObjectID)
@@ -1089,36 +1089,24 @@ final class AppVolumeMixer: ObservableObject {
         lowerOnDisconnect: Bool,
         lowerToPercent: Int) -> LoweredOutputState {
         var state = state
-        let newDefault = nextDefaultUID.flatMap { uid in
-            nextOutputDevices.first(where: { $0.uid == uid })
-        }
-        if let newDefault,
-           newDefault.outputKind.countsAsHeadphonesWhenDisconnected,
-           // Not the device this feature just lowered: putting its own volume
-           // back one refresh later would undo the lowering on the spot.
-           state.lastAutomaticLoweredOutputUID != newDefault.uid {
+        if let nextDefaultUID,
+           nextOutputDevices.first(where: { $0.uid == nextDefaultUID })?.isHeadphones == true {
             state.lastAutomaticLoweredOutputUID = nil
-            // Headphones are (or may be) back: the speakers get the volume they
-            // had before the disconnect lowered them.
+            // Headphones are back: the speakers get the volume they had before
+            // the disconnect lowered them.
             state.loweredOutput = restoringLoweredOutputVolume(state.loweredOutput,
                                                                in: nextOutputDevices)
-            if newDefault.outputKind.countsAsHeadphonesWhenTakingOver {
-                return state
-            }
-            // An unknown device may equally be a Bluetooth speaker inheriting a
-            // headphone volume, so the lowering below still gets to run on it.
+            return state
         }
 
         guard lowerOnDisconnect,
               let previousDefaultUID,
               let previousDefault = previousOutputDevices.first(where: { $0.uid == previousDefaultUID }),
-              previousDefault.outputKind.countsAsHeadphonesWhenDisconnected,
-              !nextOutputDevices.contains(where: {
-                  $0.uid == previousDefaultUID && $0.outputKind.countsAsHeadphonesWhenDisconnected
-              }),
+              previousDefault.isHeadphones,
+              !nextOutputDevices.contains(where: { $0.uid == previousDefaultUID && $0.isHeadphones }),
               let nextDefaultUID,
-              let nextDefault = newDefault,
-              !nextDefault.outputKind.countsAsHeadphonesWhenTakingOver,
+              let nextDefault = nextOutputDevices.first(where: { $0.uid == nextDefaultUID }),
+              !nextDefault.isHeadphones,
               state.lastAutomaticLoweredOutputUID != nextDefaultUID else {
             return state
         }
@@ -1578,7 +1566,7 @@ final class AppVolumeMixer: ObservableObject {
                                              uid: uid,
                                              name: name,
                                              isDefault: uid == defaultUID,
-                                             outputKind: MixerRoutingSupport.outputDeviceKind(
+                                             isHeadphones: MixerRoutingSupport.outputLooksLikeHeadphones(
                                                 transportType: hasTransportType ? transportType : nil,
                                                 dataSourceID: dataSource?.id,
                                                 name: name,
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -1141,9 +1141,17 @@ final class AppVolumeMixer: ObservableObject {
         return nil
     }
 
-    /// Teardown path only, and deliberately synchronous: `stopAll()` runs while
-    /// the app is quitting, so anything the mixer still owes the system has to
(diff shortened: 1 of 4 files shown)


PathGao:
Port id over localized name is the right key, and the live `'ispk'` / `MacBook Air扬声器` pair is what settles it. Checked against `main`: one call site for `outputLooksLikeHeadphones`, and `kAudioDevicePropertyTransportType` is read nowhere else — this is the whole surface.

One gap, and it's your own argument. The trade-offs say the asymmetry inverts for the device that *takes over*, which is why the built-in speaker ports are matched positively. Bluetooth isn't. Lowering needs `!nextDefault.isHeadphones`, so: headphones come off, a paired Bluetooth speaker becomes default, it now reads as headphones, and the lowering that runs today is skipped — the speaker plays at headphone volume. That is the accident the feature exists to prevent, on a path that works before this change.

One Bool can't serve both consumers: the disconnect side wants unknown-Bluetooth = headphones, the takeover side wants the opposite. Three states (headphones / speakers / unknown), each consumer reading `unknown` its own way, is the smallest thing that fits.

I have no Bluetooth device either, so this is read off the code, not hardware.

**Impact:** with the option on, a headphones → Bluetooth-speaker handoff stops lowering the new default; it lowers today.
**Merge:** needs a call on the Bluetooth takeover case first.

PathGao:
@PathGao You are right, and the three states are what I pushed.

The gap: `outputLooksLikeHeadphones` returning `true` for Bluetooth answered the disconnect side and inverted the takeover side. A Bluetooth speaker becoming the default read as headphones, `!nextDefault.isHeadphones` failed, and the lowering that runs today was skipped — the speaker played the room at headphone volume.

`OutputDeviceKind` is now `headphones` / `speakers` / `unknown`. The word list moved ahead of the Bluetooth check, so a Bluetooth output it names is still `.headphones` and only what is left is `.unknown`; every non-Bluetooth transport keeps today's answer of "not headphones". Two properties on the enum, one per consumer: `countsAsHeadphonesWhenDisconnected` (unknown counts — a missed pair of headphones is issue #92) and `countsAsHeadphonesWhenTakingOver` (unknown does not — it may be a speaker in the room, so it gets lowered). Lowering a Bluetooth headset that we could not name is the cost, and it is the direction that only ever goes down.

I checked the other readers of the answer rather than only the one you named. There are three, not two. The third is the early return in `loweringOutputVolumeIfHeadphonesDisconnected` that restores the speakers and re-arms the feature when headphones come back. Leaving it strict would have cost an unnamed Bluetooth headset both: the protection would fire once and the speakers would sit low afterwards. It now runs for `unknown` as well, and returns early only for a certain answer, so an unknown device falls through to the lowering. It also skips the device this feature just lowered — otherwise a Bluetooth device lowered as a suspected speaker would have its own volume restored on the next refresh and undo the lowering immediately. `CommandBarCatalog` is the fourth reader and renders exactly what it did before: not certainly speakers, headphones glyph.

Verified: `./build.sh --test` — 9482 checks, the 2 IME-dependent ones failing (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints ...`); this Mac has a Chinese IME selected and neither assertion touches audio. `swiftc -typecheck` over the whole module with the release flags, 0 errors, 0 warnings, which is what compiled `AppVolumeMixer.swift` and `CommandBarCatalog.swift` — neither is on the `--test` whitelist, and I did not run the full `./build.sh`. The seven new assertions were confirmed red first, in a standalone build of `MixerRoutingSupport.swift`: putting Bluetooth back to `.headphones` reds the `.unknown` assertion, making `countsAsHeadphonesWhenTakingOver` permissive reds the takeover one, both green on restore.

Not verified: I have no Bluetooth audio device either, so the handoff itself is read off the code, not run. And `loweringOutputVolumeIfHeadphonesDisconnected` is private and touches the HAL, so it has no unit assertions here or on `main` — the new ones cover the three states and the two readings, not the wiring into that function.

PathGao:
Three states land it, and the restore early-return is a reader I missed. Grepped `main` for it: those four sites in `AppVolumeMixer` plus `CommandBarCatalog` are all of them, so the coverage claim holds.

The body gets the glyph wrong. It says an unknown Bluetooth output "looks the same as before" — on `main` `isHeadphones` is `false` there and it renders `hifispeaker`; on the branch `.unknown != .speakers` renders `headphones`. The word list has no jbl / sonos / sennheiser, so most Bluetooth speakers now show a headphones icon, and unlike the lowering this needs no setting turned on.

```swift
icon: .symbol(device.outputKind == .headphones ? "headphones" : "hifispeaker"),
```

Keeps `main`'s answer for unknown and still fixes the localized jack.

Second, also undeclared: `unknown` counting on the *previous* default means a device this feature lowered can now trigger the next lowering. Bluetooth speaker lowered, switched off, built-in speakers lowered in turn — and `state.loweredOutput` overwrites the Bluetooth entry, so it reconnects at the lowered volume with nothing left to restore it. Unreachable on `main`, where the lowered device was always `!isHeadphones`. Down-only, so possibly a cost you take; it just isn't written down.

No Bluetooth hardware here either, both read off the code.

**Impact:** most Bluetooth speakers get a headphones icon in the command bar, no setting required.
**Merge:** needs the `CommandBarCatalog` line first.

PathGao:
@PathGao Both of those are fixed in `48461345`, by removing the half they came from rather than patching it.

**Removed: the Bluetooth third state.** `OutputDeviceKind`, `countsAsHeadphonesWhenDisconnected`, `countsAsHeadphonesWhenTakingOver`, the rewritten restore branch and the `state.lastAutomaticLoweredOutputUID != newDefault.uid` guard are gone; the predicate is a `Bool` again and `loweringOutputVolumeIfHeadphonesDisconnected` is now identical to `main` line for line. That takes your glyph regression with it — `CommandBarCatalog` is back to `main`'s line and is out of the diff — and the cascading lowering you found second, since an `unknown` previous default is what made it reachable. The reason for removing rather than patching: of the four readers, three ended up doing what `main` does even with the third state in place, the one new behaviour was on the disconnect side of a Bluetooth handoff, and there is no Bluetooth audio device here to run it once. Both defects you found were in the part that had never executed. It should come back as its own change with hardware behind it.

**Removed: the speaker port ids.** `'ispk'`/`'espk'` were asked first and returned, which took an answer away instead of adding one. A Mac whose jack and line-out share one port reports `'espk'`, and today that jack is recognized through `dataSourceName == "Headphones"` — asking the port first made it speakers before the word list ever ran. In the other direction the port id adds nothing, since "speaker port" and "the list did not know it" both come out as not-headphones, so the constants went too.

**Left:** `'hdpn'` ahead of the word list, the transport-type check, and the list unchanged behind them. Diff is +122/−24 over 3 files, from +228/−39 over 4.

**The part I'd most like you to check** is a difference I found in the half that stayed, now written up under "What this stops recognizing" in the body. The transport check overrides the name, so an AirPlay speaker named `Living Room Beats` is headphones on `main` and speakers here. I traced both readers and read it as pointing the safe way at both — as the device taking over, `main` hits the restore early-return and never lowers it, which is #92; as the device going away, `main` counts it as a headphone disconnect and lowers something for no reason. One missing lowering, one spurious one, both corrected. The cost is that someone deliberately moving to that speaker after unplugging gets it at `lowerToPercent`, and the narrower real loss is an AirPlay or HDMI endpoint that actually feeds headphones, which is now called speakers on both sides.

That reading is off the code only. The DisplayPort monitor here is on one of those transports but matches no word in the list, so it cannot exercise the difference, and I have no device that can. If you read the direction the other way, the transport check is the part to drop — the localized-jack fix is `'hdpn'` alone and does not depend on it.

`./build.sh --test`: 9476 checks, 2 failing, both the input-source ones — `nil layout falls back to ANSI semicolon` (got the full-width `；`) and `App Switcher icon-row hints describe default app and window shortcuts`. This Mac has a Chinese IME selected and neither touches audio. Whole-module `swiftc -typecheck` on the release target and SDK flags: 0 errors, 0 warnings, which is what compiled `AppVolumeMixer.swift` since it is not on the `--test` whitelist. The `'espk'` assertion was confirmed red first: putting the speaker port ids back ahead of the list reds "a jack reported as a speaker port is still headphones when the name says so", and only that one.

PathGao:
Removing the half rather than patching it is the right call, and the `'espk'` reasoning holds.

**The `'hdpn'` assertion cannot fail.** Its uid is `BuiltInHeadphoneOutputDevice`, and the haystack includes the uid, so the word list matches `headphone` as a substring. Delete both lines of the new check and that assertion stays green — it never reaches `'hdpn'`. It covers the whole point of this branch, and it is the one you did not confirm red first.

That takes the premise with it. The uid is not localized. Your live table shows the speaker side is `BuiltInSpeakerDevice`, so if the jack device really is `BuiltInHeadphoneOutputDevice`, `main` already answers true on a Chinese system through the uid and nothing here changes. A Mac that reports one device with a switching data source still fails. Which one you have decides whether this branch does anything.

Transport check: drop it. Your own rule from `48461345` — a new behaviour with no device to run it once lands as its own change. It is the only yes→no direction, and `'hdpn'` does not need it.

**Impact:** unproven; the assertion that would prove it passes on `main` too.
**Merge:** needs the `'hdpn'` assertion red without the fix first.

PathGao:
@PathGao All three taken, in `49dc9a96`.

`'hdpn'` assertion. You are right that it could not fail: `main`'s haystack is `[name, uid, dataSourceName ?? ""]` joined, folded and lowercased, and `"headphone"` is a substring of `BuiltInHeadphoneOutputDevice`. It now uses uid `BuiltInSpeakerDevice`, data source `'hdpn'`, name `外置耳机`, nothing the list can match. Confirmed red: deleting the `if dataSourceID == headphonesDataSourceID { return true }` line and running `./build.sh --test` gives 3 of 9472 failing, that assertion plus the two IME ones; restored, 2 of 9472.

Premise. I measured instead of arguing. A standalone CoreAudio probe on this Mac, headphones not plugged in: `BuiltInSpeakerDevice` / `MacBook Air扬声器` / transport `'bltn'` / data source `'ispk'` / available data sources `'ispk'` only, and the DisplayPort monitor with no data source. The live table in the body holds. It does not settle the question. With nothing plugged in the driver reports only the port that is present, so a single `'ispk'` entry is consistent with both layouts. If the jack appears as a separate `BuiltInHeadphoneOutputDevice`, `main` already answers true through the uid on a Chinese system and this branch changes nothing on that Mac. If the built-in output stays one device and switches its data source to `'hdpn'`, this is the fix. I could not observe which one this Mac is, and the body now says so under "Specifically unverified", first bullet.

Transport check dropped: the `kAudioDevicePropertyTransportType` read in `AppVolumeMixer`, the HDMI/DisplayPort/AirPlay switch, the `transportType:` parameter, its four assertions (AirPlay Beats, DisplayPort monitor, USB Jabra, Bluetooth Sennheiser) and the "What this stops recognizing" section. What is left is `'hdpn'` ahead of the unchanged word list. Diff is +92/−24 over the same 3 files. Title changed to match, since transport is no longer part of it.

Verification: `./build.sh --test` 9472 checks, 2 failing, `nil layout falls back to ANSI semicolon` and `App Switcher icon-row hints describe default app and window shortcuts`, both input-source dependent, neither touches audio. `swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk -I Sources/VMStatisticsCompat -I Sources/HIDEventSystem` over every file under `Sources/Vorssaint`: 0 errors, 0 warnings, which is what compiled `AppVolumeMixer.swift` since it is not on the `--test` whitelist. `MixerRoutingSupport.swift` is on it. Full `./build.sh` not run.

Not verified: `'hdpn'` coming back from a real jack, and which device layout this Mac has when headphones are plugged in. Both need headphones I do not have.

PathGao:
@PathGao Measured with headphones plugged in, on the same MacBook Air (M5, macOS 26.6.2, Chinese UI), CoreAudio device list read from a standalone probe:

```
* uid=BuiltInHeadphoneOutputDevice name=外置耳机 transport='bltn' dataSource='hdpn' 外置耳机
      offers 'hdpn' 外置耳机
  uid=BuiltInSpeakerDevice name=MacBook Air扬声器 transport='bltn' dataSource='ispk' MacBook Air扬声器
      offers 'ispk' MacBook Air扬声器
```

The jack is a separate device, not a data-source switch on the speaker device, and its uid is `BuiltInHeadphoneOutputDevice`. `main`'s `outputLooksLikeHeadphones` folds `[name, uid, dataSourceName]` into one haystack and already answers true on that uid, which is what the old `'hdpn'` assertion staying green without the fix had shown. So the localized-name failure this branch was opened for does not happen on this hardware: the disconnect branch is entered on a Chinese system today.

That leaves the branch fixing a Mac that reports one device whose data source flips to `'hdpn'` with a uid that says nothing about headphones. I have no evidence such a Mac exists, and the body's premise was mine and was wrong. Recommending this be closed rather than merged; the `'hdpn'` line is cheap but a change without a reproduced defect behind it is the kind this repo turns down.

Not verified: any Mac other than this one, and Bluetooth or USB headsets, which were never in scope here.

PathGao:
Closing: measured above, the defect this was opened for does not occur on the hardware it was written on, and no report exists for it.

Issue #961, opened 2026-08-25
App Switcher does not switch


Berat-Dzhevdetov:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

The App Switcher with preview doesn't want to switch over to Terminal. I would have to manually open the terminal by clicking on it on the dock

### Steps to Reproduce

1. Have Vorssaint turned on and App Switcher enabled
2. Open browser (or any program)
3. Open Terminal
4. Try to go from the "other program" to the Terminal using short cut cmd+Tab

The Terminal never gets on top of the program

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

English

### Mac and displays

Mac M3 Air 13" with second monitor LG ultra 4k

### Installation Method

DMG

tbergeron:
I have 3.3.5 and the same still happens. There are certain apps that the app switcher just won't switch to (which seems random because /sometimes/ it works) ie. ChatGPT, Fastmail, Obsidian, Zed amongst others.

xcorp:
I'm also on 3.3.5 (have also tested 3.4.0-beta.2.1), for me the problem is global and not related to any specific apps, the issues started after upgrading to macOS27
Completely uninstalling Vorssaint and clearing all permissions work for a while and then cmd+tab stops working.
If I disable Vorssaint app switcher (or just toggle off "Replace macOS cmd+Tab") the standard macOS switcher takes over as expected.
Restarting Vorssaint or rebooting the computer doesn't help, nor that just toggle the permissions in "System Center"

The only thing I've found that helps is to click "Clear all permissions" under Advanced in Vorssaint settings, running `brew uninstall vorssaint", rebooting and then re-installing with brew.


Update:
The other features work, like dock preview, so it's not an issue with grabbing screenshots from apps, just a "capture keypresses"

Also just tried "Clear all permissions" and "Uninstall Vorssaint" completely" from the advanced settings, then followed by
```
rm -f ~/Library/Preferences/com.vorssaint.utils.plist                                                                                                                                                                                                                                                                                  
rm -rf ~/Library/Application\ Support/com.vorssaint.utils
rm -rf ~/Library/Caches/com.vorssaint.utils

brew uninstall vorssaint
```

[...]

PR #1222, opened 2026-09-01
perf(switcher): stop re-reading Accessibility on every minimize restore pulse

PathGao:
## Summary

One Accessibility cost the App Switcher pays on the main thread at the worst
possible moment, removed without changing any decision the code makes.

A minimize taken over from the switcher schedules a dense burst of restore
pulses — the completion burst alone is 151 timers across 450ms. Every pulse
called `restoreSourceAfterTargetMinimizeIntent`, which resolved the target
window (`kAXWindows` copy plus one `_AXUIElementGetWindow` per window), read its
focused window (two more AX reads) and looked up an `NSRunningApplication`
*before* handing all of it to `shouldRestoreSourceAfterTargetMinimizeIntent` —
whose first real check is `frontmostPID == sourcePID`. Because the pulses keep
firing after the restore succeeds, that check rejects most of them and every
read was wasted. With eight windows on the target app that is roughly 1,660
synchronous AX round trips crammed into the 450ms the app spends animating its
own minimize.

The fix: the three Accessibility-backed inputs of
`shouldRestoreSourceAfterTargetMinimizeIntent` are now `@autoclosure`. The guard
order inside the function is unchanged, so the same inputs produce the same
answer; what changes is that they are only evaluated once the cheap comparisons
have failed to settle it.

Trade-offs:

- Reordering the reads at the call site would have been a smaller diff than the
  autoclosures, but it leaves the next caller free to make the same mistake —
  the function's signature would still say "hand me these values".
  `@autoclosure` makes paying early impossible to express, and it makes the
  property testable: the tests below count evaluations.
  `shouldContinueFocusRetry` looks like the same anti-pattern and was checked,
  but is not one: its caller needs the minimized state unconditionally to feed
  `SwitcherWindowFocusRetryState.observe`, so there is nothing to defer.
- Pulse density is untouched: it was tuned deliberately in "Fix App Switcher
  latency and click cancellation" and this change is not an argument about it.

Dropped from the original branch: threading a once-resolved `AXUIElement` from
`activate()` down through the minimized check, the prepare pass, the focus pass,
the staging pass and the minimize-restore observer. Short-circuiting the
re-resolve at every step is what removed `main`'s healing of a window element
the app rebuilt under the pass, and the `liveWindow` fallback that restored the
healing put a `kAXRole` read in front of every step that took a handed-down
element — turning "five window enumerations to zero" into "five enumerations to
five `kAXRole` reads", about 40 round trips per ⌘Tab. That does not pay for
seven new parameters, a new helper, a check that can only be written as a
regular expression over the source, and an element-lifetime rule with no
mechanical guard. The predicate was also the wrong one: a handle answering a
`kAXRole` read does not establish that a later write lands on the same window,
and Apple documents no such thing, while `axElement` in the same file compares
`AXWindowResolver.windowID(for:)` against the window ID for the same one round
trip and answers the identity question directly.

## Verification

MacBook, macOS 26 (Tahoe), single display, arm64, in a worktree rebased onto
`main` @ `f25d1aa8`.

- `./build.sh --test`: `9473 checks`, zero warnings, four of them new here. Two
  checks fail on this Mac on clean `main` as well — `nil layout falls back to
  ANSI semicolon` and `App Switcher icon-row hints describe default app and
  window shortcuts` — because they read the machine's live input source, a
  Chinese IME. Nothing else changed state.
- `WindowActivator.swift` is not in the `--test` file list, so `--test` alone
  never compiles the call site. Type-checked separately over the whole app
  source set with the same target and SDK `build.sh` uses
  (`swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk`):
  clean, no diagnostics.
- The four new assertions in `Tests/MetricsTests.swift` count how many times the
  expensive inputs are evaluated: zero when the frontmost check alone settles
  the pulse, and the focused-window read skipped once the target reports
  minimized. Confirmed they actually fail: forcing both autoclosures at the top
  of the function turned exactly those two count assertions red and nothing
  else, and restoring the order turned them green again.

What I could not test:

- **No profile.** The round-trip figure is counted off the code path and the
  window count, not measured with Instruments. The direction is certain
  (strictly fewer AX calls, none added); the magnitude is arithmetic, not a
  measurement.
- **No live run of the AX paths at all.** The switcher and the minimize takeover
  need a signed build and a real desktop; this was verified by compilation and
  by the pure-logic tests only. Reviewer verification worth doing: use the
  switcher to raise a window and click its minimize button, and confirm the
  previous app comes back exactly as before.
- **No multi-window stress case.** The 8-window figure that motivates this was
  not reproduced; behaviour with an app holding many windows is unverified on
  hardware.

## Anything else

Refs #961


PathGao:
Equivalence rewrite holds — truth table matches, each autoclosure forced once at most. The evaluation-count tests are the right way to pin it.

One gap. The boundary is at `asyncAfter`, but the rebuild hazard you name is likelier *inside* the synchronous pass: `prepareWindowForActivation` deminimizes and raises, `activateApp` runs, and only then does `focusWindow` use the element resolved before all of it. `main` re-resolved there, so a rebuilt element healed itself. Now it short-circuits on the stale handle, every `AXUIElementSetAttributeValue` returns `.invalidUIElement` into a discarded result, and it still returns `true` — silently no focus. I didn't reproduce a rebuilding app either; the claim is just that `main` was safe by construction here and this isn't. Fix that keeps the win: `focusWindow` already reads `kAXMinimized` first — on `.invalidUIElement`, fall through to `axElement(windowID:in:)`.

Also: the element handed to `SwitcherWindowMinimizeRestore` lives on as `observedWindow` until `invalidate()`, so "reused only inside the synchronous pass" isn't literally true. Same element `main` resolved a moment later, no behavior change — but that sentence is what the next reader will trust.

**Impact:** target window only, on apps that rebuild elements on deminiaturize — focus silently fails to land.
**Merge:** needs the invalid-element fallback in `focusWindow` first.

PathGao:
Release build since verified on this branch: `./build.sh` completes with zero warnings and produces the bundle. `WindowActivator.swift` is not in the `--test` file list, so `--test` alone never compiled the change — this closes that gap.

Signing needed a workaround unrelated to this branch: an invalid identity in the keychain makes `build.sh` exit 1 before the bundle stage. That is fixed separately in the build-signing PR; with it, this branch builds clean end to end.

PathGao:
@PathGao Right on both counts, and the first one is worse than a `focusWindow` fix
alone would cover. `prepareWindowForActivation` is what provokes the rebuild, and
four steps run after it on the handed-down element, not one: the `windowMinimizedState`
that feeds `SwitcherWindowFocusRetryState.observe` on both the fullscreen and the
ordinary path, `focusWindow`, and `stageSourceBehindTargetIfNeeded`. A stale handle
in the first turns a real minimized state into "could not be queried" and changes what
the retries decide; in the last it raises and re-focuses nothing. Patching only
`focusWindow` would have left three.

So the fallback sits in the resolve step every one of them already shares. `liveWindow`
returns the handed-down element when it still answers a `kAXRole` read and falls back to
`axElement(windowID:in:)` when it does not; the five `axWindow ?? axElement(...)` sites
now go through it, including `SwitcherWindowMinimizeRestore` (`axElementForMinimizeRestore`
was only a shim for that one call and is gone). Cost is one attribute read per handed-down
element instead of a `kAXWindows` copy and a walk, and `axWindow == nil` — every caller
outside `activate()` — is byte-for-byte the old path.

On `observedWindow`: you are right that the sentence was wrong, and it is corrected in the
body rather than in code. That element is registered with the `AXObserver` at the same point
in the pass where the observer used to resolve it for itself, so its lifetime did not move;
the body now says so instead of claiming the reuse ends with the synchronous pass.

Verification: `./build.sh --test` gives `9474 checks` with the two input-source checks that
fail on clean `main` on this Mac (a Chinese IME) and nothing else. `./build.sh` completes
with zero warnings, which is the run that actually compiles `WindowActivator.swift`. The new
check is a source-shape one — no `?? axElement(` survives anywhere in the file — because the
behaviour needs an app that rebuilds its windows and I still cannot produce one. Confirmed
red: putting `axWindow ?? axElement(...)` back at one site turned exactly that check red and
nothing else, and `liveWindow` turned it green again.

Not verified: still no live AX run, no profile, and no app that actually rebuilds its
elements on deminiaturize — the fallback is reasoned and compiled, not exercised. Worth a
reviewer doing: ⌘Tab to another app's window and confirm focus lands on that exact window,
then raise a window through the switcher and click its minimize button.

`#1217` and `#1214` touch neighbouring switcher and AX code; neither hands an element across
a step, so nothing here reaches into them.

PathGao:
Both gaps are closed, and wider than I asked: all seven synchronous resolves go through `liveWindow`, not just `focusWindow`. Deleting `axElementForMinimizeRestore` is what makes the shape check real — `axElement` is `private`, so with the shim gone `SwitcherWindowMinimizeRestore` has no route to it.

One swap. `liveWindow` checks `kAXRole` answers, which proves the app is alive, not that the handle still points at this window. `AXWindowResolver.windowID(for: axWindow) == windowID` is the same single round trip and is the exact predicate `axElement` uses, so reuse and fallback agree on which window is the right one.

Follow-up, not a blocker: `focusAfterSpaceHop` and `activateSource` each call `prepareWindowForActivation` then `focusWindow` back to back — two `kAXWindows` copies of the same window in one synchronous pass. `activateSource` is on the restore path this PR is named after. Grepped `main`; those are the only two left.

Still no live AX run from me — the fallback is compiled and shape-checked, not exercised.

**Impact:** none. Same decisions, fewer AX round trips.
**Merge:** ship it.

ref #961

PathGao:
@PathGao Dropped the second half of this branch in `dc81382e`. Not a force-push — the removal is its own commit, so the review above still reads against what it reviewed.

**What went:** the target element resolved once in `activate()` and threaded through seven parameters, `liveWindow(_:windowID:in:)`, and the source-shape assertion that pinned it. What stays is the `@autoclosure` change in `shouldRestoreSourceAfterTargetMinimizeIntent` alone.

**Why:** your review of that half is what removed the case for it. Threading the element short-circuited the re-resolve that used to heal a rebuilt handle, and `liveWindow` — the fix for that — puts a `kAXRole` read in front of every step that receives one. The half's account goes from "five window enumerations to zero" to "five enumerations to five `kAXRole` reads", about 40 round trips per ⌘Tab. That does not pay for seven parameters, a helper, a regex-shaped assertion, and an element-lifetime rule with no mechanical guard — `SwitcherWindowMinimizeRestore` holds `observedWindow` until `invalidate()`, and the `asyncAfter` boundary is stated in a comment and nowhere else.

The probe was also the wrong one. "This handle still answers a `kAXRole` read" is not documented to mean a later write lands on the same window, and I could not produce an app that rebuilds its window elements on deminiaturize, so that fallback was compiled and shape-checked but never executed. `axElement` already compares `AXWindowResolver.windowID(for:)` against the target id at the same one round trip and answers identity rather than liveness. If this comes back, that is the predicate it should use, with an app that actually reproduces the rebuild.

**Title changed** — `and activation step` described only the half that is gone.

**Checks:** `TESTS FAILED (2 of 9473)`, both the input-source-dependent pair (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints…`); 9474 → 9473 is the deleted shape assertion. `WindowActivator.swift` is outside the `--test` whitelist, so a whole-module `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` at the build's target and SDK: exit 0, no diagnostics. The tree after the revert is byte-identical to the `@autoclosure` commit.

**Worth re-checking:** that nothing in `WindowActivator` still reads as if an element were being handed down — the revert restored `axElementForMinimizeRestore`, and I want a second pair of eyes on whether the restored shim and the `@autoclosure` call sites agree.

PathGao:
Right call — `liveWindow`'s `kAXRole` probe answered liveness, and every step it guarded was asking identity.

Checked the residue you asked about. Nothing is handed down: all seven `axElement(windowID:in:)` call sites — `windowMinimizedState`, `setWindowOrigin`, `setWindowMinimized`, `closeWindow`, `prepareWindowForActivation`, `stageSourceBehindTargetIfNeeded`, `focusWindow` — build their own `axApp` first, and no parameter in the file carries an `AXUIElement`. `axElementForMinimizeRestore` has exactly one caller, `SwitcherWindowMinimizeRestore.init`, resolving from the `axApp` it just created. The two autoclosures pass `(windowID, pid)`, not elements. Shim and call sites agree.

The body's `shouldContinueFocusRetry` note holds: `state.observe(targetMinimizedState:)` runs unconditionally, so there is nothing to defer there.

Not a blocker: `restoreSourceAfterTargetMinimize(_:)`, the non-Intent twin, is `fileprivate` and has no callers in the tree — and it eagerly builds `matchesTargetBundle`, the shape you just made lazy. Unless it is kept on purpose, it will attract this same fix again.

Still no live AX run from me.

**Impact:** none — the tree is byte-identical to the commit I already read.
**Merge:** ship it.

ref #961

PR #1223, opened 2026-09-01
chore(clipboard): drop the unused single-entry delete alias

PathGao:
## Summary

Removes `ClipboardHistoryService.removeSelectedQuickEntry()`, a three-line wrapper that
nothing calls and whose name states the opposite of what it does.

`bb612b17` added multi-selection deletion and moved every call site to
`removeSelectedQuickEntries()`. The singular wrapper was left behind, forwarding to the
plural one. So the name promises "delete the selected entry" while the body deletes the
whole batch — `removeSelectedQuickEntries()` resolves its targets through
`quickEntriesForPrimaryAction()`, which returns the full batch selection when one exists.

Two ways to resolve that in this one symbol: give the singular name a real singular body,
or delete it. Deleting is the right one — a genuinely single-entry delete has no caller
asking for it, and the batch path is already the behaviour the UI wants. Adding one would
be speculative API, and keeping the current wrapper keeps a name that lies.

Scope: this removes one unreachable symbol. It does not fix the singular-name-plural-body
shape in general — see **Anything else** for the two live methods that still have it.

No behaviour change: nothing referenced the symbol, so nothing was rerouted.

**Not changed:** the `⌥⌫` / `⌘⌫` bindings still delete the whole selection. That is
deliberate — `bb612b17`'s changelog entry describes exactly that ("supports
Command-Delete and Option-Delete to remove selected items at once"). This PR only removes
unreachable code.

**Not changed:** `CHANGELOG.md`.

## Verification

MacBook (Mac17,3, Apple M5), macOS 26.6.2 (25G83), Swift 6.3.3, self-built from this
branch — not a release build.

Zero callers, checked before deleting rather than inferred from reading:

- `grep -rn "removeSelectedQuickEntry\b" Sources Tests` — one hit, the definition line
  itself. The only other occurrences of the stem in the repo are `removeSelectedQuickEntries`
  (`ClipboardQuickPanelView.swift:232`, `ClipboardHistoryService.swift:443` definition,
  `:1212` key handler).
- Repo-wide search for the name as a string literal: no hits, so it is not reached through
  `#selector`, `NSSelectorFromString`, or KVC. `ClipboardHistoryService` is a
  `final class ... ObservableObject` with no protocol requiring the symbol.

Commands:

- `./build.sh` — full `-O` build, 0 warnings, 0 errors; 43,537,688-byte `build/Vorssaint` produced
- `./build.sh --test` — `TESTS OK (9399 checks)` against this branch's base `b0dde656`; this PR adds no assertions

**What I could not test:** nothing behavioural — the removed code had no reachable path to
exercise. No new test was added; deleting unreachable code leaves no proposition to assert.
Note that `ClipboardHistoryService.swift` is not in `build.sh --test`'s source whitelist, so
the full `./build.sh` above is the only compile gate this change actually passes through.

## Anything else

The same shape — a singular name whose body acts on the whole batch — is still live in two
methods on `ClipboardHistoryService`, and both have callers, so neither is touched here:

- `copySelectedQuickEntry()` (`:417`) — called from `ClipboardQuickPanelView.swift:225`
  (footer paste button), `:510` (click on an already-selected row), and the key handler at
  `ClipboardHistoryService.swift:1183` (plain `⏎`).
- `copySelectedQuickEntryOnly()` (`:427`) — called from `ClipboardQuickPanelView.swift:229`
  (footer copy button) and the key handler at `ClipboardHistoryService.swift:1179` (`⇧⏎`)
  and `:1195` (`⌘C`).

Both resolve targets through the same `quickEntriesForPrimaryAction()` (`:1299`) and fan
out to `copyQuickEntries` / `copyOnlyQuickEntries` when the batch holds more than one
entry. Their behaviour is what the UI wants; only the names are singular. Renaming live
methods with six call sites is a separate change and does not belong in a dead-code
removal, so it is deliberately left out of this PR rather than overlooked.

### DIFF
--- a/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
+++ b/Sources/Vorssaint/Services/Clipboard/ClipboardHistoryService.swift
@@ -452,10 +452,6 @@ final class ClipboardHistoryService: ObservableObject {
         save()
     }
 
-    func removeSelectedQuickEntry() {
-        removeSelectedQuickEntries()
-    }
-
     /// Where the pointer sat when the keyboard last moved the selection. Rows
     /// scrolling under a still pointer report hover, and hover would otherwise
     /// take the preview back from the row the arrow keys chose.


PathGao:
Zero-caller claim holds. On `main` the only occurrences of the stem are `removeSelectedQuickEntries` at `ClipboardHistoryService.swift:443`, `:1216` and `ClipboardQuickPanelView.swift:232` — no string-literal or selector reference anywhere either, so the deletion is safe. The `⌥⌫` / `⌘⌫` path goes through the plural and is untouched.

One thing this leaves behind. The same name-lies-about-body shape is still live twice: `copySelectedQuickEntry()` and `copySelectedQuickEntryOnly()` both resolve targets through `quickEntriesForPrimaryAction()`, so both act on the whole batch under a singular name — and unlike the wrapper you deleted, those have callers (`ClipboardQuickPanelView.swift:225` / `:229` / `:510`, plus the key handler at `:1183` / `:1187`). So the argument in the body doesn't actually stop at this symbol: the PR removes the one instance of the trap that was unreachable and keeps the two that aren't. Not a reason to hold this one — renaming live methods is a separate change — but worth naming so it doesn't read as closed.

**Impact:** none. Nothing referenced the symbol.
**Merge:** ship it.

PathGao:
@PathGao You are right that the body argued wider than the diff. Fixed in the body, no code change.

Checked both on this head (`cafa816b`). `copySelectedQuickEntry()` (`ClipboardHistoryService.swift:417`) and `copySelectedQuickEntryOnly()` (`:427`) both open with `let selectedEntries = quickEntriesForPrimaryAction()` (`:1299`) and fan out to `copyQuickEntries` / `copyOnlyQuickEntries` when the count is above one — same shape as the wrapper this PR removes, minus the forwarding. Both have callers. One correction to your line numbers: they were read off `main`, and this branch is four lines shorter, so the key-handler sites sit at `:1183` (plain `⏎` → `copySelectedQuickEntry`) and `:1179` (`⇧⏎` → `…Only`), not `:1187` / `:1183`. There is also a third key-handler site you did not name, `:1195` (`⌘C` → `…Only`). With `ClipboardQuickPanelView.swift:225`, `:229` and `:510` that is six live call sites across the two methods.

Taking your word that renaming live methods is a separate change, so I left them alone and instead added an `## Anything else` section naming both, their call sites, and why they stay. The Summary now says explicitly that this PR removes one unreachable symbol and does not close the shape in general. Also corrected a stale line number in Verification — the `removeSelectedQuickEntries` key-handler site is `:1212` on this branch, not `:1216`.

No commit and no push: this was body-only, so the tree is unchanged and the build results already in Verification still describe this head. Nothing new was compiled or run for this update. I did not search for a follow-up issue to file for the rename — there is no open issue covering the batch-versus-single naming, so nothing to reference.

PathGao:
Corrections check out. Six live call sites is right, and I did miss the ⌘C one.

On today's `main` the three key handlers read `:1185` (⇧⏎ → `…Only`), `:1189` (⏎ → `copySelectedQuickEntry`), `:1201` (⌘C → `…Only`) — two above the numbers I gave you, so `main` grew two lines since I read it. Your branch numbers are self-consistent against your base; nobody should try to reconcile the three sets. Zero callers still holds on current `main`: one hit, the definition line.

One for the follow-up, not this PR. `togglePinSelectedQuickEntry()` sits in the same key-handler block (⌥P) but resolves through `selectedQuickEntry`, not `quickEntriesForPrimaryAction()`. So with a batch selected, ⌥⌫ deletes all of it and ⌘C copies all of it while ⌥P pins only the cursor row. Honest name, but it is the odd one out on the batch axis — unless single-row pinning under a batch is deliberate, which I can't tell from source. Worth folding into the rename issue. And you're right that nothing open covers it: I read the full open list, no issue mentions batch-versus-single on the clipboard panel.

**Impact:** none. Body-only update, tree unchanged.
**Merge:** ship it.

PR #1224, opened 2026-09-01
fix(quit-protection): size the confirmation HUD to its longest translation

PathGao:
## Summary

The quit/close confirmation HUD had a hardcoded 300pt panel with 12pt insets on both
sides, leaving 276pt for a single-line `NSTextField`. Its text is localized and
`String(format:)`-composed with the shortcut symbol, and no translation had ever been
measured against that box.

Measured every shipped combination — 13 languages x `holdHUDFormat` / `doubleHUDFormat` /
`extraHUDFormat` (with the widest arguments: `⌘Q`, `⌘W`, `⇧⌘Q`…`⌃⌘W`), plus `cancelHint`
and `releaseHint` — at the HUD's own fonts (`.systemFont(ofSize: 13, weight: .semibold)`
for the title, `.systemFont(ofSize: 10.5)` for the detail). Eight strings overflow 276pt,
all of them `doubleHUDFormat`:

- fr, `Appuyez à nouveau sur ⌘W pour quitter/fermer` — 299.5pt (widest overall)
- ptBR, `Pressione ⌘W novamente para encerrar/fechar` — 296.5pt
- ru, `Нажмите ⌘W ещё раз для выхода/закрытия` — 291.8pt
- de, `⌘W erneut drücken zum Beenden/Schließen` — 280.1pt

(plus the `⌘Q` variant of each, 277–297pt). A French or Brazilian-Portuguese user in
double-press mode gets the end of the instruction clipped by up to 24pt.

The fix sizes the panel from the labels that draw the text. `show(...)` fills them first,
then `fittingSize(_:)` takes `max(title.fittingSize.width, detail.fittingSize.width) + 24`,
floored at 300, and `setContentSize` applies it before positioning. English and the other
nine languages keep the current 300pt exactly; the widest case becomes a 324pt panel.

Asking the labels rather than measuring the strings matters because the labels are
`NSTextField`s and their cells lay text out themselves. An `NSString.size(withAttributes:)`
of the same string at the same font does not include any horizontal inset the cell adds,
and the widest translation leaves only 0.5pt of slack — a 2pt-per-side cell inset would
put fr and ptBR straight back to clipped. Going through the view removes the question
rather than answering it, and it is how every other panel here sizes itself: `QuickToolHUD`,
`BrightnessOSD`, `CommandBarService`, `ShelfService` and the rest all read a view's
`fittingSize`.

Every other `size(withAttributes:)` call in the tree was checked and none has this problem:
`MenuBarRenderer`, `ScreenshotRenderer`, `ScreenshotSelectionController`, `RecorderZoomLane`
and `SwitcherSupport.titleWidth` each measure text they then draw themselves with the same
attributes, so measuring and rendering are the same code path there. This HUD was the only
place where a measured string was handed to a label.

Considered pinning the constant to the measured maximum (bump 300 to 324) — smaller diff,
but it widens the HUD for every language to accommodate four, and it goes stale silently
the next time a translation is edited. Also considered wrapping to two lines or letting
the labels truncate: both change what the HUD is. It stays a one-line pill; corner radius,
fill, stroke, fonts, insets, and the 48pt height are untouched.

**Not changed:** any `*Strings.swift`. Shortening translations to fit a fixed box would be
fixing the wrong side.

**Not changed:** `CHANGELOG.md`.

**Separate observation, not touched here:** `title.textColor = .white` and the
`NSColor(calibratedWhite: 0.09, alpha: 0.95)` fill are hardcoded, so the HUD stays dark in
Light Mode. That may well be intended for a transient overlay — it is a design call, so I
left it alone rather than guess.

## Verification

MacBook (Mac17,3, Apple M5), macOS 26.6.2 (25G83), Swift 6.3.3.

`./build.sh --test` — `TESTS FAILED (2 of 9403)`. The two are
`nil layout falls back to ANSI semicolon` (`got "⌃⌥⌘ ；", expected "⌃⌥⌘ ;"`) and
`App Switcher icon-row hints describe default app and window shortcuts`; both read the
machine's live keyboard input source, which here is a Chinese IME — hence the full-width
`；`. Neither touches this file. The other 9401 checks pass, 9399 of which were the count
before this branch: the four added are the new HUD assertions.

`Sources/Vorssaint/UI/QuitProtection/QuitProtectionHUD.swift` is outside `build.sh --test`'s
source whitelist, so it is not compiled here; CI covers it on both runners. `swiftc -parse`
on the file is clean.

New assertions in `Tests/MetricsTests.swift`, read as source shape since the file cannot be
compiled into the test target:

- the HUD source is readable and non-trivial (the byte count is in the failure message, so
  the block cannot pass vacuously on a missing file)
- the width comes from `title.fittingSize.width` and `detail.fittingSize.width`
- no `size(withAttributes:` remains in the file
- inside `show(...)`, the label fill precedes the measurement, which precedes the resize —
  that order is what makes the measurement see the real text

Red-first, against the two spellings that actually existed rather than an invented one:
on `main` (`b0dde656`, fixed 300pt panel) two of the four go red; on this branch's first
commit (`5824c559`, the string-measurement version) three of the four go red. Both green
on the current tip.

Measurement probe (throwaway, compiled against the repo's own `QuitProtectionStrings.swift`
so the strings are the shipped ones): 260 language x format x detail combinations, 16
clipped by the old 276pt text area, widest text 299.5pt. Re-run against the sizing formula:
0 clipped, widest panel 324pt, none below the 300pt floor.

**What I could not test:** the on-screen result. The HUD was not exercised live in
French/Portuguese/Russian/German, so the claim rests on the width measurement plus the
source-shape assertions, not on a screenshot. The full `./build.sh` was not run — with the
signing keychain locked it blocks on a system unlock dialog. And the specific thing that
prompted this rework — whether an `NSTextFieldCell` adds horizontal inset — is still not
measured; the fix routes around the question instead of answering it.

## Anything else

The two font constants added in the first commit (`ContentView.titleFont` /
`detailFont`) existed only so the standalone measurement could use the same fonts as the
labels. Nothing measures separately any more, so they are gone and the labels set their
fonts inline again, as on `main`.

#892 and its duplicate #1158 asked for the ⌘Q/⌘W protection itself. The HUD sized
here is the only thing that tells those reporters the protection engaged, so they are
linked the way #1218 links them — related surface, not the fix track, which is #1200's.

Refs #892
Refs #1158


PathGao:
Measuring all 13 languages × 3 formats before touching the constant is the right order, and `needsDisplay = true` in `update` is the bit most people miss when a panel changes width.

One thing: `fittingSize` measures with `NSString.size(withAttributes:)`, but the text is drawn by `NSTextFieldCell`. The body says the shared fonts mean measuring and rendering "cannot drift apart" — the fonts can't, the sizing path still can. The widest string leaves 0.5pt of slack (300pt of text area against 299.5pt of text), so a 2pt-per-side cell inset would put fr/ptBR back to clipped. I could not verify whether a label cell adds one — no AppKit to run here.

Measure the labels instead: update first, then size from them.

```swift
content.update(title: title, detail: detail)
size = Self.fittingSize(content)   // max(title.fittingSize.width, detail.fittingSize.width) + 24, floored at 300
panel?.setContentSize(size)
```

The object that renders does the measuring, and it matches how every other panel here sizes itself — `QuickToolHUD`, `BrightnessOSD`, `CommandBarService` all go through `fittingSize`.

**Impact:** none outside this HUD; worst case is the four overflowing languages stay clipped by a few pt, which is what the PR set out to fix.
**Merge:** needs the width to come from the labels (or a measurement showing the cell adds no horizontal inset) first.

PathGao:
@PathGao You were right that the sizing path could drift even though the fonts could not.
`NSString.size(withAttributes:)` returns the glyph run's width; the label draws through
`NSTextFieldCell`, which lays the text out itself and is free to inset it. At 0.5pt of
slack on the widest string, 2pt per side would have put fr and ptBR back where they
started — the fix would have missed exactly the cases it exists for.

Rather than measure whether the cell insets, the width now comes out of the labels, so it
cannot matter. `show(...)` fills them first, then `fittingSize(_:)` reads
`max(title.fittingSize.width, detail.fittingSize.width) + 24`, still floored at 300, and
`setContentSize` applies it before positioning. Same shape as `QuickToolHUD`,
`BrightnessOSD` and `CommandBarService`. The two font constants the old measurement needed
are gone with it; the labels set their fonts inline again, as on `main`.

I checked the other five `size(withAttributes:)` sites for the same shape —
`MenuBarRenderer`, `ScreenshotRenderer`, `ScreenshotSelectionController`,
`RecorderZoomLane`, `SwitcherSupport.titleWidth`. Each measures text it then draws itself
with the same attributes, so there is no second renderer to disagree with. This HUD was
the only place a measured string was handed to a label.

The file is outside `build.sh --test`'s whitelist, so the check is a source-shape one in
`MetricsTests.swift`: the width comes from `title.fittingSize` and `detail.fittingSize`,
no `size(withAttributes:` is left in the file, and inside `show(...)` the fill precedes
the measure which precedes the resize — that order is what makes the measurement see the
real text. It also asserts the source read is non-empty with the byte count in the message,
so it cannot pass on a file it failed to open. Verified red against both spellings that
really existed: two of the four go red on `main` (`b0dde656`), three of four on this
branch's first commit (`5824c559`), all green on the tip.

`./build.sh --test` gives `TESTS FAILED (2 of 9403)` here, both in assertions that read
the machine's live keyboard input source — this machine's is a Chinese IME, so the
semicolon comes back as `；`. Neither is near this file. The full `./build.sh` was not run:
the signing keychain is locked here and codesign blocks on a system unlock dialog.

Still not verified: the on-screen result. No live run in French/Portuguese/Russian/German,
so this rests on the measurement and the assertions, not a screenshot. And the original
question — whether an `NSTextFieldCell` adds horizontal inset — is still unanswered; the
change routes around it instead.

PathGao:
Width comes out of the labels now, and `positionPanel` reads `size` after the resize so the centring follows it. The 24pt is one constant shared with `layout()` instead of a second copy.

One thing the body has wrong: "This HUD was the only place where a measured string was handed to a label." `ScrollingTitle` does the same — `SwitcherSupport.titleWidth` measures with `NSString.size(withAttributes:)`, and SwiftUI `Text` draws it.

```swift
// ScrollingTitle.scrollingLine
let run = measured + Self.gap   // the two Text copies size themselves
```

If SwiftUI's width differs from the measurement, `run` doesn't match where the second copy actually starts, so the marquee seams or jumps at the wrap. The same number also decides `overflows`, i.e. whether it scrolls at all.

I have not measured whether SwiftUI `Text` and `NSString.size` agree at `.system(size:weight:)` — no AppKit here, so that is a shape match, not a reproduced bug. Out of scope for this diff; it wants its own issue rather than a wider PR.

**Impact:** none from this diff.
**Merge:** ship it.

Issue #1153, opened 2026-08-30
Every filter tap stalls the front session after fast user switching; #1075 fixed two of ~22


iltonandrew:
Splitting this out of #1075 so its PR can stay the size of its evidence.

### The mechanism is not scroll-specific

#1075's diagnosis: an instance left running in a switched-away session still owns an active filter tap, so the window server routes the front session's events through a process that cannot answer and waits out the ~4s tap timeout on each one. PR #1145 fixes that for the two scroll services — the ones with a reporter, a dump, and a symptom.

But the process owns roughly twenty-two `.defaultTap` event taps, and the mechanism applies to every one of them. Two others carry `scrollWheel` in their mask (`MouseButtonShortcutService`, `WindowLayoutService`, plus `CleaningModeManager` while active); most of the rest carry `keyDown` or mouse masks (`SuperKeyService`, `KeyboardDebounceService`, `TextSnippetService`, `AppSwitcher`, `FinderCutPaste`, `MiddleClickService`, …), which would stall the front session's typing and clicking the same way scrolling stalls today. Nothing outside PR #1145 observes fast user switching.

### What #1145 leaves ready

`SessionActivity` (the watcher) and `SessionActivitySupport.tapShouldRun` (the pure gate) are general; the two scroll services show the wiring — subscribe in init, route to `syncWithPreferences()`, guard the timeout re-arm, and make `stop()` invalidate the port (ten tap owners already invalidate; the scroll services joined them in #1145). Gating each remaining service is the same small change, but twenty services in one PR is not reviewable, and several have their own lifecycle wrinkles (threaded taps in `TextSnippetService`, the `SuperKey` mapping guard).

### Not verified

Same caveat as #1075: nobody involved so far has two logged-in sessions, so the per-service impact is inferred from the shared mechanism, not measured. A fast-user-switching user with Super Key or snippets enabled could confirm whether typing in the other account stutters the way scrolling dies.

Refs #1075.

iltonandrew:
Sharper census from PathGao on #1075 after #1145 merged: 21 other `.defaultTap` creation sites across 17 services still build an active filter tap and never hand it back on resign; the three `.listenOnly` taps (`DockPreview`, `AutoQuit`, `MusicLaunchBlocker`) are unaffected and out of scope. Structural, not observed — the reporter's dump held only the four scroll taps.

PathGao:
[...]

PR #1225, opened 2026-09-01
fix(taps): invalidate event tap ports on teardown

PathGao:
## Summary

Ten event tap owners switched their tap off on teardown but never handed the port back. `tapEnable(false)`, `CFRunLoopRemoveSource`, `tap = nil` — and the tap stays registered.

I assumed ARC covered this: `CGEvent.tapCreate` returns a `CFMachPort`, so dropping the last Swift reference should release the port and take the tap with it. It does not. Measured on `origin/main` with a probe that replicates the exact teardown these services use:

- The mach port name the `CFMachPort` wrapped still carries receive and send rights after the properties are nil'd; with `CFMachPortInvalidate` added, the same name reports `KERN_INVALID_NAME` immediately.
- `CGGetEventTapList` still reports the tap against the process, `enabled=false`, for the rest of the process's life. 20 start/stop cycles leave 20 registered taps and 20 leaked port names. With `CFMachPortInvalidate`, 20 cycles leave zero.
- Same numbers for `.defaultTap` and `.listenOnly`, and the same with no run loop source at all — so it is not the source holding the port. `CFGetRetainCount` on the `CFMachPort` reads 5 while the service holds one reference: the last Swift reference is nowhere near the last reference.

That means a tap owner accumulates one registration per enable/disable cycle, and every one of those cycles is user-reachable (each `syncWithPreferences()` on a settings change runs `stop()` then `start()`).

The fix is one line per teardown, placed where the thirteen services that already do this put it: after `tapEnable(false)` and `CFRunLoopRemoveSource`, before the property is cleared, and on the tap thread for the four threaded owners.

`Tests/MetricsTests.swift` gets the rule rather than a list: scan every file that calls `CGEvent.tapCreate` and require at least as many `CFMachPortInvalidate` calls as taps created, counted on comment-stripped source. The existing per-service assertion only covers the owners that also do session gating, so a new tap owner could be added tomorrow without tripping anything.

Counting per tap rather than per file matters for the three owners that create more than one: `BrightnessService` and `SuperKeyService` own two taps each and `WindowLayoutService` three, so a file-level check would let one invalidate answer for all of them. Stripping comments first matches the per-service check twenty-five lines above, which strips them so that prose naming the API cannot satisfy the rule. All 24 tap owners are green under the counted rule, and each of the three multi-tap files invalidates every tap it creates.

The sweep also has to say how much it read. A file with no literal `CGEvent.tapCreate` is skipped, so if the call ever moves behind a helper — the consolidation shape #937 and #938 describe — every owner drops out, the offender list stays empty and the check passes having read no tap owner at all. That is the same vacuous-scan failure being fixed on #1229, and it takes the same shape of fix: count the owners the sweep actually reached, assert that count above zero beside the empty offender list, and carry the number into the failure message. It reads 24 today.

The margin on `MouseClickDebounceService` — one tap, two invalidations — is not headroom. Both of its invalidations are on that one tap, in `stop()` and in `runEventTap()`'s exit path, so a second tap added to that file would pass for free. It is the only file on `main` where the counts diverge, and the comment now says so rather than leaving the number to be read as slack.

### What this does not change

The registered leftover is `enabled=false`. Whether the window server also *stalls* the front session on a disabled tap after fast user switching — the symptom in #1075 / #1153 — is not something I could measure: that needs two logged-in sessions. What is measured here is the registration and port leak, which holds regardless of fast user switching.

`MusicLaunchBlocker`, `AutoQuitService` and `DockPreviewService` are `.listenOnly` and were called out as out of scope for the stall mechanism. They are included anyway because the leak is identical and the guardrail is a rule about all tap owners, not about filter taps.

`QuitProtectionService` is not here because `f25d1aa8` already added its `CFMachPortInvalidate`. It was going to be this PR's one guardrail exception; the rebase made the exception unnecessary, so the guardrail ships with no allowances at all.

No session gating is added here. That is the per-service work #1153 describes and it belongs in its own PRs.

### Why this is one PR and not ten

+11 source lines, +35 test lines, no behavior branch — well under the +110 median, and every hunk is the same line for the same measured reason. The argument for splitting is reviewability, and there is nothing to review per file that is not already true of the first file.

Splitting would also make the guardrail worse. It can only land once, and it can only be green once every tap owner invalidates, so a split would have to ship it carrying a shrinking exception list — allowances that mean nothing except "not done yet", which is the exact shape the guardrail exists to prevent. Landing it together lets it ship with no exceptions.

## Verification

Mac17,3 (Apple M5), macOS 26.6.2 (25G83), self-built from source, release configuration.

```
./build.sh                     compiled clean, 0 warnings, 41.5 MB build/Vorssaint (run before the rebase)
./build/Vorssaint --selftest   SELFTEST OK
./build.sh --test              9470 checks, 2 failures
```

The two failures are `nil layout falls back to ANSI semicolon` and `App Switcher icon-row hints`. Both are outside this change: they read the machine's live input source, which is a Chinese IME here. Baseline for that claim is this branch's merge base `f25d1aa8`, run in a throwaway worktree on this machine: 9469 checks, the same 2 failures and nothing else. The scanned-owner assertion replaces the previous one rather than adding to the count, so the branch stays at 9470.

The guardrail was verified to bite rather than assumed to, once for each of the two shapes it now covers. Dropping one of `BrightnessService`'s two invalidations — the multi-tap shape a per-file check cannot see — produced:

```
every event tap owner invalidates its port on teardown: ["Services/Display/BrightnessService.swift (2 taps, 1 invalidated)"]
```

Commenting out the `WindowMaximizer` line, so the string is still in the raw file but not in the code, produced:

```
every event tap owner invalidates its port on teardown: ["Services/WindowMaximizer.swift (1 taps, 0 invalidated)"]
```

Both go green again when the line is restored, and the restored tree is back to the same 2 failures.

The scanned-owner assertion was bitten the same way, and the old one was bitten alongside it to show the difference. Changing the searched literal to `CGEvent.tapCreateViaHelper` — every owner disappearing behind a helper, in one edit — gives, on the assertion as it now stands:

```
every event tap owner invalidates its port on teardown, across 0 scanned owners: []
```

The same edit against the previous assertion, which guarded `!tapOwnerSources.isEmpty`, ran green: 9470 checks, 2 failures, the vacuous sweep passing unremarked. Restoring the literal returns the branch to green.

What I could not test:

- The fast-user-switching symptom itself. That needs two logged-in sessions; this machine has one.
- The full release compile of the ten service files **on the rebased tree**. `./build.sh --test` builds a hand-listed subset with `-Onone`, and most of these service files are not on that list, so locally they were only compiled by the full `./build.sh` run — which was made before the rebase onto `f25d1aa8`. The change is one added line per file with no new symbols, but the authoritative full compile of the final tree is CI's two jobs, not mine.
- Runtime behavior of the four threaded owners with a `stop()` racing the tap thread. The placement there is identical to what `TextSnippetService` and `SuperKeyService` already do, but I exercised it only through the probe.

## Anything else

Refs #1153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 - Uninstallation now verifies that sleep was restored and asks for admin authorization if the passwordless restore fails, preventing closed-lid sleep prevention from remaining permanently disabled on the Mac. Thanks to @mugurc.
 - Screenshot quick preview no longer takes keyboard focus when presented, keeping keystrokes in the active app until the preview is clicked. Thanks to @iltonandrew.
 - Smooth Scroll and Scroll Inverter now step aside during fast user switching and resume on return, preventing background scroll stalls. Thanks to @iltonandrew.
+- Turning features on and off repeatedly no longer leaves unused keyboard and mouse listeners registered with the system for the rest of the session. Thanks to @PathGao.
 - Mouse navigation, button shortcuts and middle click now step aside outside the active login session or after Accessibility access is removed, preventing input stalls.
 - Focus follows mouse now leaves chosen apps alone and tracks the final pointer position during drags before changing focus.
 - Cleaning Mode now ends when its login session leaves the screen, preventing its input lock from affecting another active user.
--- a/Sources/Vorssaint/Services/Audio/MusicLaunchBlocker.swift
+++ b/Sources/Vorssaint/Services/Audio/MusicLaunchBlocker.swift
@@ -120,6 +120,7 @@ final class MusicLaunchBlocker: ObservableObject {
         if let mediaKeyTapSource {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), mediaKeyTapSource, .commonModes)
         }
+        CFMachPortInvalidate(tap)
         mediaKeyTapSource = nil
         mediaKeyTap = nil
     }
--- a/Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift
+++ b/Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift
@@ -37,6 +37,7 @@ final class PreciseVolumeRollerService: ObservableObject {
         if let source {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), source, .commonModes)
         }
+        if let tap { CFMachPortInvalidate(tap) }
         tap = nil
         source = nil
         gate.reset()
--- a/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
+++ b/Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
@@ -653,6 +653,7 @@ final class AutoQuitService: ObservableObject {
         if let closeRequestRunLoopSource {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), closeRequestRunLoopSource, .commonModes)
         }
+        if let closeRequestTap { CFMachPortInvalidate(closeRequestTap) }
         closeRequestTap = nil
         closeRequestRunLoopSource = nil
     }
--- a/Sources/Vorssaint/Services/Display/BrightnessService.swift
+++ b/Sources/Vorssaint/Services/Display/BrightnessService.swift
@@ -694,6 +694,7 @@ final class BrightnessService: ObservableObject {
         if let keyTapSource {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), keyTapSource, .commonModes)
         }
+        CFMachPortInvalidate(tap)
         keyTapSource = nil
         keyTap = nil
     }
@@ -777,6 +778,7 @@ final class BrightnessService: ObservableObject {
             }
             CGEvent.tapEnable(tap: tap, enable: false)
             CFRunLoopRemoveSource(runLoop, source, .commonModes)
+            CFMachPortInvalidate(tap)
             if clearFunctionKeyThread() { installFunctionKeyTap() }
         }
     }
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -111,6 +111,7 @@ final class DockClickService {
         if let runLoopSource {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), runLoopSource, .commonModes)
         }
+        if let tap { CFMachPortInvalidate(tap) }
         tap = nil
         runLoopSource = nil
         for (_, sweep) in pendingSweeps { sweep.cancel() }
--- a/Sources/Vorssaint/Services/DockPreview/DockPreviewService.swift
+++ b/Sources/Vorssaint/Services/DockPreview/DockPreviewService.swift
@@ -345,6 +345,7 @@ final class DockPreviewService: ObservableObject {
         if let runLoopSource {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), runLoopSource, .commonModes)
         }
+        if let tap { CFMachPortInvalidate(tap) }
         tap = nil
         runLoopSource = nil
         cancelPendingHover()
--- a/Sources/Vorssaint/Services/Finder/FinderRenameService.swift
+++ b/Sources/Vorssaint/Services/Finder/FinderRenameService.swift
@@ -131,6 +131,7 @@ final class FinderRenameService {
 
             CGEvent.tapEnable(tap: tap, enable: false)
             CFRunLoopRemoveSource(runLoop, source, .commonModes)
+            CFMachPortInvalidate(tap)
             if clearEventTapThread() { installTap() }
         }
     }
--- a/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
+++ b/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
@@ -186,6 +186,7 @@ final class KeyboardDebounceService: ObservableObject {
 
             CGEvent.tapEnable(tap: tap, enable: false)
             CFRunLoopRemoveSource(runLoop, source, .commonModes)
+            CFMachPortInvalidate(tap)
             eventLock.withLock {
                 state.reset()
             }
--- a/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
+++ b/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
@@ -380,6 +380,7 @@ final class AppSwitcher: ObservableObject {
 
             CGEvent.tapEnable(tap: tap, enable: false)
             CFRunLoopRemoveSource(runLoop, source, .commonModes)
+            CFMachPortInvalidate(tap)
             if clearEventTapThread() { installTap() }
         }
     }
--- a/Sources/Vorssaint/Services/WindowMaximizer.swift
+++ b/Sources/Vorssaint/Services/WindowMaximizer.swift
@@ -42,6 +42,7 @@ final class WindowMaximizer: ObservableObject {
         if let runLoopSource {
             CFRunLoopRemoveSource(CFRunLoopGetMain(), runLoopSource, .commonModes)
         }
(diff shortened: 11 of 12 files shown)


PathGao:
Coverage checks out — I grepped `CGEvent.tapCreate` across `Sources/Vorssaint`: 28 sites in 24 files, and these 10 are exactly the ones with no `CFMachPortInvalidate` on main.

The guardrail is weaker than the body claims, twice over:

**Per file, not per tap.** `BrightnessService` owns 2 taps, `SuperKeyService` 2, `WindowLayoutService` 3 — one `CFMachPortInvalidate` anywhere in the file passes, so dropping one of Brightness's two stays green. The bite test used `WindowMaximizer`, a single-tap file, so it doesn't cover that shape.

**Raw source, not stripped code.** The per-service check 25 lines above strips `//` lines on purpose, "so prose naming the API cannot answer for it". Here a comment mentioning the function satisfies the rule.

Counting on the stripped text fixes both:

```swift
let code = source.components(separatedBy: "\n")
    .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
    .joined(separator: "\n")
guard code.contains("CGEvent.tapCreate") else { continue }
if code.components(separatedBy: "CFMachPortInvalidate").count
    < code.components(separatedBy: "CGEvent.tapCreate").count {
    tapOwnersWithoutInvalidate.append(file)
}
```

Green on all 24 tap owners as they stand today.

Nothing runtime-verified on my side, and I didn't work through the threaded owners' teardown races.

**Impact:** no behavior change; the guardrail silently exempts the three multi-tap files.
**Merge:** needs the guardrail counting per tap on stripped source first.

ref #1153

PathGao:
@PathGao Both holes are closed in `7b21a941`. The guardrail now strips `//` lines first, counts `CGEvent.tapCreate` and `CFMachPortInvalidate` on the stripped text, and flags a file when the invalidate count is lower than the tap count. Your snippet, with the counts carried into the failure message so a partially covered file is legible rather than just named.

On the three files it was exempting: they were already correct, so this is a check gained, not a bug found. `BrightnessService` invalidates both — `removeKeyTap()` for the session tap, the tap thread's teardown for the function-key tap. `SuperKeyService` invalidates both in `runEventTap()`'s exit path, key tap then mouse tap. `WindowLayoutService` invalidates all three, one per `stopDirectionalTap()` / `stopEdgeSnapTap()` / `stopGestureTap()`. All 24 tap owners pass under the counted rule; the widest margin is `MouseClickDebounceService` at 1 tap and 2 invalidations.

Verified by breaking it twice rather than once, since there are now two shapes to bite. Dropping one of `BrightnessService`'s two invalidations gives `["Services/Display/BrightnessService.swift (2 taps, 1 invalidated)"]` — the case that stayed green before. Commenting out the `WindowMaximizer` line, leaving the string in the raw file but not in the code, gives `["Services/WindowMaximizer.swift (1 taps, 0 invalidated)"]`. Both restore to green, and the restored tree is back to the same 2 pre-existing failures on this machine.

`./build.sh --test`: 9470 checks, 2 failures, both the input-source ones that reproduce on clean `main` here. No source file changed in this commit, so no full `./build.sh` rerun — the caveats about the rebased tree's full compile and the threaded owners' teardown races stand exactly as they did, and I still have not run anything against two logged-in sessions.

PathGao:
The counted rule reads right and the two bite tests cover both shapes.

It can still go green covering nothing: `!tapOwnerSources.isEmpty` checks the *file* list, not the owner count. Files with no literal `CGEvent.tapCreate` are `continue`d, so if the call moves behind a helper — the consolidation shape #937 / #938 describe — all 24 owners drop out and the assertion passes having scanned zero. That's the failure the body rules out by design: "the absence of a tap owner that never invalidates, not a list". One line:

```swift
guard taps > 0 else { continue }
tapOwners += 1
```

then assert `tapOwners > 0` next to the empty list.

`MouseClickDebounceService` at 1/2 isn't headroom, it's the same hole one level down: both invalidations are on the one tap, `stop()` and `runEventTap()`'s exit path. A second tap added to that file passes for free. It's the only file on main where the counts diverge — not worth guarding, worth a clause in the comment so nobody reads the margin as slack.

I grepped `main` for both counts. I did not run the test, and the runtime caveats stand as they were.

**Impact:** no behavior change; the guardrail goes vacuous under a helper extraction.
**Merge:** needs the scanned-owner count asserted first.

ref #1153

PathGao:
@PathGao Closed in `81cd9481`, your snippet as written: `tapOwners += 1` after the `guard taps > 0`, and the assertion now reads `tapOwners > 0 && tapOwnersWithoutInvalidate.isEmpty`, with the count in the message so a shrunken sweep says how far it shrank. It reads 24 today.

You are right that this is the file list guarding the owner list — a sweep that skips every file without the literal cannot be defended by the fact that files existed. I took the shape from #1229's scans rather than inventing one, since they report the count for the same reason.

Bitten both ways. Changing the searched literal to `CGEvent.tapCreateViaHelper` — the helper extraction in one edit — gives `every event tap owner invalidates its port on teardown, across 0 scanned owners: []`. The same edit against the previous assertion ran green, 9470 checks and 2 failures, the vacuous sweep passing unremarked. Restoring the literal returns to green.

On `MouseClickDebounceService`: agreed, not headroom, and I did not guard it. Both invalidations are on the one tap, `stop()` and `runEventTap()`'s exit path, so a second tap there passes for free. That is now a clause in the comment naming the file, so the divergence is documented rather than read as slack.

I checked the rest of the file for the same shape. This is the only whole-tree sweep in `Tests/MetricsTests.swift` that filters files by content before collecting: the `waitUntilAllOperationsAreFinished` and application-role scans read every file and already guard `!appSources.isEmpty`, which is the right guard when nothing is skipped, and the availability-writer scan asserts an exact set, which an empty scan fails. The `uiFiles` borderless-menu scan is the one that needed the same treatment, and #1229 is already carrying it — I left it alone rather than collide.

`./build.sh --test`: 9470 checks, 2 failures, the two input-source ones. I dropped "pre-existing on main" from the body and gave a baseline instead — merge base `f25d1aa8` in a throwaway worktree on this machine, 9469 checks, the same 2 and nothing else. No source file changed here, so no full `./build.sh`; the caveats about the rebased tree's full compile, the threaded owners' teardown races, and the two-session symptom stand unchanged.

PathGao:
Both holes are closed. I re-checked the coverage claims against `main` rather than taking them: 28 taps across 24 files, 14 already invalidating (18 calls), and the 10 without are exactly the files in this diff. After the patch `BrightnessService` is 2/2, `SuperKeyService` 2/2, `WindowLayoutService` 3/3, `MouseClickDebounceService` 1/2 with both calls on the one tap. The two unguarded `CFMachPortInvalidate(tap)` sites sit under a `guard let tap`, so nothing optional slips through.

One leftover, not a blocker. The per-service loop 25 lines above still asserts `code.contains("CFMachPortInvalidate")` for its six owners. All six are single-tap, so `invalidations >= taps` is the same predicate — the new sweep strictly subsumes it, and that line is now the second copy of one rule. Unless it is deliberate belt-and-braces against the sweep being weakened, it is six checks saying nothing new.

I did not verify #1229's `uiFiles` guard, and the runtime caveats stand as they were.

**Impact:** none.
**Merge:** ship it.

ref #1153

PathGao:
Both closed, and the two bite tests are the right ones — the old assertion running green on the vacuous sweep is the part that proves it.

I reproduced the rule against `main` independently: 28 `CGEvent.tapCreate` across 24 files, 18 `CFMachPortInvalidate` across 14. The 10 files at zero are exactly this PR's 10, `BrightnessService` is the only one of them at 2 taps needing 2 lines, and `MouseClickDebounceService` at 1/2 is the only divergence among the 13 that already invalidated. Every count in the body checks out.

The full-compile caveat is discharged rather than still standing: Build & selftest and Swift 6.0.3 are both green on the rebased tree, which is the authoritative compile you deferred to. What stays unverified is the two-session symptom and the threaded owners' teardown races, and neither gates this — the registration and port leak is measured independently of both.

I did not run the test. My counts are a static reproduction of the rule over `main`, not a run of it.

**Impact:** no behavior change beyond the ports being handed back; the guardrail now fails on a partially covered file and on a sweep that reads nothing.
**Merge:** ship it.

ref #1153

Issue #1045, opened 2026-08-26
vorssaint appears to block certain mouse clicks


tomas65107:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When playing and using Minecraft in full-screen, it randomly stops receiving mouse event clicks when vorssaint is active (see video).
When i turn off the vorssaint, it works normally.

I have the following things under Mouse and keyboard enabled:
Invert mouse scrolling and Three finger click acts as middle click
THO i have tested this with them disabled and it still does this.

Also, as shown in the video, it starts receiving mouse clicks after a few seconds, but the bug returns randomly or when switching focused windows (sometimes).

I launched mc with Modrinth but tried it with official launcher and it did not solve the problem.
And also this bug is not limited to the lower part of the screen, it bugs out anywhere on screen

Thanks for any tips in advance!

### Steps to Reproduce

1. Have vorssaint open
2. Launch minecraft
3. Fullscreen it
4. Right after fullscreening try to click anything

### Feature Area

Mouse and keyboard

### Screenshots / Recordings

https://github.com/user-attachments/assets/591bd14c-bbe5-4353-83c8-08fa28b9492c

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

English

### Installation Method

Homebrew

PathGao:
#960 is an open change for exactly this: clicks getting swallowed in borderless fullscreen games, which is how Minecraft runs. Nothing has changed on your Mac yet, and I can't promise it covers your case.

One thing that would help pin it down: leave Vorssaint on, but turn off the Dock features (the window previews you get from hovering or clicking a Dock icon), then play for a while. If the clicks stop vanishing, that's the culprit. If they still vanish, say so — that rules it out and points somewhere else.

tomas65107:
I tried that and for a quick test it appears that the issue was resolved. Im sorry, i was not aware of the second issue. I will post upd ates here if something changes.

PathGao:
That's the answer I needed. The Dock previews are what's eating the clicks, and #960 is the open change aimed at exactly that path, so it's the most likely fix. One quick test isn't proof, but it's a strong enough signal to go on.

[...]

PR #1228, opened 2026-09-01
perf(radial-menu): keep the side-button tap cheap

PathGao:
## Summary

Every side-button event ran a full profile decode.

`RadialMenuSupport.claimsMouseButton` and the wheel's own tap callback both answered
"which button opens the wheel?" by calling `decodeProfiles`, which reads the stored blob,
JSON-decodes every profile with all of its items, and then runs `NSImage(data:)` over
every custom icon it finds — up to `maxStoredIconBytes` (64 KB) of PNG per item, up to 12
items per wheel, per profile. Three tap callbacks ask that question:
`MouseNavigationService`'s HID tap (whose mask includes `.otherMouseDragged`, so it asks
on every event of a held-button drag), `MouseButtonShortcutService`'s tap on every
`.otherMouseDown`, and the radial menu's own tap on every press and release. Holding a
side button and moving the mouse therefore ran the whole decode at the mouse's event
rate, synchronously in the callbacks, on the main run loop — where overrunning the tap
timeout costs dropped events.

The buttons are now decoded on their own: `RadialMenuSupport.claimedMouseButtons` reads
the same blob through a `Decodable` that has one field, `mouseButton`, so the items and
their icons are never touched. It keeps the same legacy single-button fallback
`decodeProfiles` migrates from, so the two answers cannot drift for a user who has not
saved a profile yet. `claimsMouseButton` routes through it, which fixes all three callers
at once; the wheel's tap decodes the profiles only after a press has matched a summoner
and only to open the wheel; and `syncMouseTap` uses it when its caller has no profiles in
hand. `decodeProfiles` keeps its name and its callers, and gains a note on the cost and
on what to call from a callback instead; `claimsMouseButton`'s doc comment claimed "pure
defaults reads", which had not been true since profiles landed, and now says what it
does.

Splitting the question in two let the two answers disagree inside one click, and the
first revision of this branch let that disagreement reach the app under the pointer. A
profile that keeps a valid `mouseButton` but fails the full decode — a bad `color`, a
non-array `items` — is dropped by `sanitizedProfiles` and kept by the one-field read. The
press took the early `return Unmanaged.passUnretained(event)` that guarded `beginSession`,
while its release fell past that branch to `return nil`, so the app got half a click,
which the comment right above promises never happens. Only the cheap read decides that
now: the profile match no longer returns, it only decides whether a wheel opens, so a
press whose wheel cannot be decoded opens nothing and is still swallowed with its
release. The `guard` became an `if let` for that reason.

Trade-offs:

- A cache of the claimed buttons with invalidation on defaults change would also have
  removed the cost, but it adds state to keep correct across profile edits, migration and
  the capture row. Decoding one small field is already cheap enough that there is nothing
  left to cache.
- Renaming `decodeProfiles` to something that says what it costs would make the wrong
  call harder to write, at five call sites plus tests and a conflict with every other
  branch that touches the file. The doc comment says it instead.

### Split out: the session gate

This branch also handed the tap back across a fast user switch. That half is now #1270.
It references a different issue (#1153, not #1045), shares no premise with the decode,
and carries a **Not tested** note this half does not — there is no second logged-in
account here. It went out whole: the `SessionActivity.shared.onChange` registration with
its `endSession()`, the `SessionActivitySupport.tapShouldRun` wrapper around the sync
gate, the feature check that wrapper needed, the re-arm's session question, and the
assertion for the resign path. `AXIsProcessTrusted()` is back where `main` has it, in the
tap-creation guard.

## Verification

```
./build.sh --test      TESTS OK (9537 checks)
```

`RadialMenuService.swift` is outside the `--test` compile whitelist, so the whole of
`Sources/Vorssaint/**` was type-checked separately with the build's own target and SDK
(`arm64-apple-macosx14.0`, `MacOSX26.sdk`, plus the two `-I` compat paths). The full
`./build.sh` was not run: with the signing keychain locked, `codesign` blocks on an
unlock dialog.

Three assertions cover this half. Two are behavioural and run against the real decoders:
the cheap read agrees with the full decode on a profile set that carries a custom icon,
and it falls back to the legacy single-button key exactly where the full decode does,
both for a missing blob and for a corrupt one. The third is a source-shape check on
`handleMouseTap`, because that path is private, takes a live `CGEvent` and cannot be
called from a test: once the cheap read has claimed the button, nothing past that point
hands an event back, which is what keeps the down and the up together.

**Not tested.** The event-rate cost is read off the code path — no instrumented profile
with custom icons was timed, so the improvement is "the decode no longer happens", not a
measured millisecond figure. The half-click path itself was not reproduced with a live
corrupt blob and a real side button.

## Anything else

`MouseNavigationService.handle` asks `RadialMenuSupport.claimsMouseButton`,
`MouseButtonShortcutSupport.claimsButton` and `MouseButtonShortcutService.isCaptureActive`
again on every event of a gesture, including the release, so a claim that changes while a
side button is held can split a click there too. That is on `main` already and
independent of this branch, so it is left alone here. `MouseButtonShortcutService`,
`MiddleClickService` and `DockClickService` all record what the press decided and let the
release follow it, so none of them has the shape.

#1045 reports mouse clicks dropped at random in a full-screen game, coming back after a
few seconds, with "three finger click acts as middle click" switched on. That is the
shape of a filter tap overrunning the tap timeout and being disabled by the window
server, and `MouseNavigationService`'s HID tap — whose mask includes
`.otherMouseDragged` — is one of the three callbacks that ran the full profile decode on
every event of a held-button drag. Linked as a candidate cause, not a confirmed one:
that report has not been retested, and #836 asks the reporter for a Dock-clicks-only
retest first.

Refs #1045

### DIFF
--- a/Sources/Vorssaint/Services/MouseNavigation/MouseNavigationService.swift
+++ b/Sources/Vorssaint/Services/MouseNavigation/MouseNavigationService.swift
@@ -211,7 +211,8 @@ final class MouseNavigationService: ObservableObject {
         // other side button keeps navigating. While the shortcut capture row
         // is listening, every press belongs to it, including a button with no
         // mapping yet, or the capture could never see the button it is asked
-        // to watch for.
+        // to watch for. Both claims are asked again on every Drag of a held
+        // side button, so both have to stay cheap.
         if RadialMenuSupport.claimsMouseButton(buttonNumber)
             || MouseButtonShortcutSupport.claimsButton(buttonNumber)
             || MouseButtonShortcutService.isCaptureActive {
--- a/Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
+++ b/Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
@@ -72,7 +72,14 @@ final class RadialMenuService: ObservableObject {
     private var activationObserver: NSObjectProtocol?
     private var promptedForAccessibility = false
 
-    private init() {}
+    private init() {
+        // A filter tap owned by a switched-away login session stalls input in
+        // the session on screen. Hand the mouse tap back on resign and build
+        // it again from preferences when this session comes back.
+        SessionActivity.shared.onChange { [weak self] _ in
+            self?.syncMouseTap()
+        }
+    }
 
     var sessionActive: Bool { !stack.isEmpty }
 
@@ -128,16 +135,28 @@ final class RadialMenuService: ObservableObject {
     // app under the pointer; alive only while a button is configured, torn
     // down with the feature)
 
-    private func syncMouseTap(profiles: [RadialMenuProfile]? = nil) {
+    /// Whether the button tap has a job right now. The capture row needs it
+    /// even with the feature switched off, so it can tell a button this app
+    /// cannot see from one that is simply set to something else.
+    private func mouseTapWanted(profiles: [RadialMenuProfile]?) -> Bool {
         let defaults = UserDefaults.standard
-        let currentProfiles = profiles ?? RadialMenuSupport.decodeProfiles(
+        let hasAnyButton = profiles.map { list in
+            list.contains { RadialMenuMouseTrigger.sanitized($0.mouseButton).buttonNumber != nil }
+        } ?? !RadialMenuSupport.claimedMouseButtons(
             defaults.data(forKey: DefaultsKey.radialMenuProfiles),
             defaults: defaults
-        )
-        let hasAnyButton = currentProfiles.contains {
-            RadialMenuMouseTrigger.sanitized($0.mouseButton).buttonNumber != nil
-        }
-        guard hasAnyButton || isReportingMouseButtons else {
+        ).isEmpty
+        let enabled = AppFeature.radialMenu.isAvailable
+            && defaults.bool(forKey: DefaultsKey.radialMenuEnabled)
+        return (hasAnyButton && enabled) || isReportingMouseButtons
+    }
+
+    private func syncMouseTap(profiles: [RadialMenuProfile]? = nil) {
+        guard SessionActivitySupport.tapShouldRun(
+            featureWanted: mouseTapWanted(profiles: profiles),
+            accessibilityGranted: AXIsProcessTrusted(),
+            sessionIsActive: SessionActivity.shared.isActive
+        ) else {
             tearDownMouseTap()
             return
         }
@@ -146,7 +165,7 @@ final class RadialMenuService: ObservableObject {
         if let mouseTap, !CGEvent.tapIsEnabled(tap: mouseTap) {
             tearDownMouseTap()
         }
-        guard mouseTap == nil, AXIsProcessTrusted() else { return }
+        guard mouseTap == nil else { return }
         let mask = (CGEventMask(1) << CGEventType.otherMouseDown.rawValue)
             | (CGEventMask(1) << CGEventType.otherMouseUp.rawValue)
         guard let tap = CGEvent.tapCreate(
@@ -184,7 +203,18 @@ final class RadialMenuService: ObservableObject {
 
     private func handleMouseTap(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            if let mouseTap { CGEvent.tapEnable(tap: mouseTap, enable: true) }
+            let shouldRearm = SessionActivitySupport.tapShouldRun(
+                featureWanted: mouseTapWanted(profiles: nil),
+                accessibilityGranted: AXIsProcessTrusted(),
+                sessionIsActive: SessionActivity.shared.isActive
+            )
+            if shouldRearm, let mouseTap {
+                CGEvent.tapEnable(tap: mouseTap, enable: true)
+            } else {
+                // Invalidating the port from its own callback stack is unsafe;
+                // finish this callback fail-open, then release the tap.
+                DispatchQueue.main.async { [weak self] in self?.syncMouseTap() }
+            }
             return Unmanaged.passUnretained(event)
         }
         let pressed = Int(event.getIntegerValueField(.mouseEventButtonNumber))
@@ -202,17 +232,17 @@ final class RadialMenuService: ObservableObject {
             lastMouseButtonSeen = pressed
         }
         let defaults = UserDefaults.standard
-        let profiles = RadialMenuSupport.decodeProfiles(
+        let button = Int64(pressed)
+        // Every press and release of every extra button lands here, so the
+        // cheap question comes first: the wheels themselves are decoded only
+        // once a press turns out to be a summoner, and only to open one.
+        guard RadialMenuSupport.claimedMouseButtons(
             defaults.data(forKey: DefaultsKey.radialMenuProfiles),
             defaults: defaults
-        )
-        guard let matchingProfile = profiles.first(where: {
-            RadialMenuMouseTrigger.sanitized($0.mouseButton).buttonNumber == Int64(pressed)
-        }) else {
+        ).contains(button) else {
             return Unmanaged.passUnretained(event)
         }
 
-        let button = Int64(pressed)
         // The source lives on the main run loop, so this already runs on
         // main; acting synchronously keeps a quick click ordered (the down
         // opens the wheel before its own up arrives). The claimed button is
@@ -221,6 +251,15 @@ final class RadialMenuService: ObservableObject {
         // caption promises exactly that).
         if type == .otherMouseDown {
             if !sessionActive {
+                let profiles = RadialMenuSupport.decodeProfiles(
+                    defaults.data(forKey: DefaultsKey.radialMenuProfiles),
+                    defaults: defaults
+                )
+                guard let matchingProfile = profiles.first(where: {
+                    RadialMenuMouseTrigger.sanitized($0.mouseButton).buttonNumber == button
+                }) else {
+                    return Unmanaged.passUnretained(event)
+                }
                 beginSession(for: matchingProfile, hold: false, heldButton: button)
             } else if !holdPhase {
                 endSession()
--- a/Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
+++ b/Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
@@ -72,18 +72,7 @@ final class RadialMenuService: ObservableObject {
     private var activationObserver: NSObjectProtocol?
     private var promptedForAccessibility = false
 
-    private init() {
-        // A filter tap owned by a switched-away login session stalls input in
-        // the session on screen. Hand the mouse tap back on resign and build
-        // it again from preferences when this session comes back.
-        // The session goes with it: the tap is what ends a held wheel, so a
-        // wheel still open when the tap is handed back would never see its
-        // release and would come back stuck in hold phase.
-        SessionActivity.shared.onChange { [weak self] active in
-            if !active { self?.endSession() }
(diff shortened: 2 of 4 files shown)


PathGao:
The split is right, and routing `claimsMouseButton` through it fixes all three tap callers in one place. Grepped `decodeProfiles` on main: what is left is settings, the menu panel and `presentPreview`, nothing in a callback.

One thing. `handleMouseTap` now asks "is this mine?" with `claimedMouseButtons` and "which wheel?" with `decodeProfiles`, and those two can disagree. A profile carrying a valid `mouseButton` that fails the full decode (bad `color`, non-array `items`) is dropped by `sanitizedProfiles` and kept by the cheap read. The down then hits the new `return Unmanaged.passUnretained(event)` while the up falls past that branch to `return nil`, so the app under the pointer gets half a click, which the comment right below promises never happens. On main both halves passed through. Corrupt blob only, so reach is small, but so is the fix: return `nil` from that guard, the button is claimed either way.

Not verified: no second account here either, so the session half is read, not run. CI has not reported yet.

**Impact:** a profile that fails the full decode while keeping its mouseButton splits a click for the app under the pointer.
**Merge:** needs the down/up mismatch closed first.

ref #1153

PathGao:
@PathGao The split was mine: `handleMouseTap` ended up with two answers to one question, and
both were allowed to decide whether the click was passed on. A
profile that keeps a valid `mouseButton` but fails the full decode is dropped by
`sanitizedProfiles` and kept by the one-field read, so the press hit the new
`return Unmanaged.passUnretained(event)` and the release fell past that branch to `return nil`.

Closed the way you described. The profile match no longer returns anything; it is an `if let`
that decides whether a wheel opens, and the claimed-button guard above it is now the only
thing in that function that decides pass or swallow. A press whose wheel cannot be decoded
opens nothing and is swallowed with its release, which is what the comment above it already
promised.

On the other taps. `MouseNavigationService.handle` does have the shape: it re-asks
`claimsMouseButton`, `claimsButton` and `isCaptureActive` on every event of a gesture, the
release included, so a claim that changes while a side button is held splits a click there
too. It is on `main` and nothing on this branch touches that function, so I left it and noted
it in the body rather than widening this PR — say the word if you want it in here.
`MouseButtonShortcutService` is the counter-example: `consumedButtons` makes Up and Drag
follow their Down, and its comment says exactly why. `MiddleClickService`
(`middleButtonHeld` / `suppressedButtonSequence`) and `DockClickService` (its pending-click
state) do the same. I considered copying `consumedButtons` here and did not: one read already
answers for both halves, and a held-button set would be new state that a release lost to a
tap timeout could strand.

Verified with `./build.sh --test`: `TESTS FAILED (2 of 9473)`, the same two input-source
checks that fail on clean `main` here, no third. Ran the full `./build.sh` this time too —
`Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift` is not in the `--test`
whitelist, so the branch's earlier "full build not needed" line was wrong and the body now
says so. The release compile is clean, zero warnings; the run then stops in the signing step
with `errSecInternalComponent` because this checkout has no code-signing identity, which is
the machine and not the diff. The new check reads `handleMouseTap` and asserts nothing after the claimed-button
guard hands an event back; against the previous commit's source it is red, against this one
green. `handleMouseTap` is private and needs a live `CGEvent`, so that is a source-shape
check, not a synthesised click — the closest runnable thing I could leave behind.

Still not run: the fast-user-switching half, no second account here. Searched the tracker for
the split before writing this. Nothing reports the wheel's side button giving an app half a
click — the three closed side-button reports are all other mechanisms (a button never
detected, a thumb wheel read as one button, back/forward needing repeated presses) — so
`Refs #1153` stays as the only ref.

PathGao:
Both closed. The claimed-button guard is now the only pass-through left after the capture blocks, and that capture block is symmetric on its own (`consumedButtons` in the shortcut tap swallows the Up with its Down), so nothing in that function splits a click any more. The "cannot drift" claim in the body holds for a nameable reason: `sanitizedProfiles` never returns empty, so `decodeProfiles` reaches its legacy branch on exactly the two conditions `claimedMouseButtons` does — nil data, or a decode that throws.

Leave `MouseNavigationService.handle` out. It is on `main`, independent, and dragging it in re-opens a diff that is now correct.

One thing the new teardown opens. `tearDownMouseTap()` on resign leaves `stack`, `holdPhase` and `holdButton` set, so a wheel held open by its side button when the session switches away never gets its release. Coming back, the summoner press is swallowed (`holdPhase` is true) and the next release runs whatever slice is highlighted. `suspend()` in this same file already pairs the teardown with `endSession()`, and `MouseNavigationService.stop()` clears `passThroughButtons` for the same reason. The closure has the flag:

```swift
SessionActivity.shared.onChange { [weak self] active in
    if !active { self?.endSession() }
    self?.syncMouseTap()
}
```

Not verified: no second account here either, so the resign path is read, not run.

**Impact:** a wheel held open by its side button across a fast user switch comes back stuck in hold phase, and the next release of that button runs a slice the user did not ask for.
**Merge:** needs the session ended when the tap is handed back.

ref #1153

PathGao:
Both halves return `nil` on every path out of the down branch now — that closes it. Gating the tap through `tapShouldRun` also matches the five services already doing it.

Agreed on not widening. `MouseNavigationService.handle` needs the `consumedButtons` shape, which is new state and a separate risk; file it on its own so it doesn't ride along here.

One exception to the new comment's "never one half of a click": `MouseButtonShortcutService.isCaptureActive` returns `passUnretained` three lines above the claim guard, and it flips at runtime. Down swallowed by the wheel → capture row opens → the Up is passed through there, and the shortcut tap passes it on too, because its `consumedButtons` never saw the Down. Same on main, so not a merge gate — but the new check only reads the region after `if type == .otherMouseDown {`, so it cannot see it. Gating that early return on `type == .otherMouseDown` would close it; the capture row only reads Downs.

Not verified: whether a capture row can appear while a side button is held.

**Impact:** none new — the corrupt-blob split is closed.
**Merge:** ship it.

ref #1153

PathGao:
@PathGao Confirmed and fixed in `921ed7ea`. This is the resign gate from your 10:11 review, not the down/up split — that one is closed and the later "ship it" was answering the round before this comment.

Read the three fields before touching anything. `tearDownMouseTap()` only disables the port, invalidates it and clears `mouseTap` / `mouseTapSource` / `isWatchingMouseButton`; `stack`, `holdPhase` and `holdButton` are untouched, and `endSession()` is the only entry that clears all three together with the monitors and the panel. Nothing else on the resign path reaches them: `activationObserver` only fires on `didActivateApplicationNotification` for another app in this session, which a fast user switch does not produce, and the `SuperKeyService.onHoldEnded` hook is installed only for a shortcut-summoned session with modifiers. So a wheel open when the session resigns keeps `holdPhase == true` with the tap that would have ended it gone. Coming back, `handleMouseTap` takes the `else if !holdPhase` branch on the summoner press and swallows it, and the next release runs `endHoldPhase()` on whatever slice the pointer is over.

The closure now has the flag and the shape you gave: `if !active { endSession() }` before `syncMouseTap()`, which is the same pair `suspend()` already makes in this file.

Left behind: one source-shape check that reads the `SessionActivity.shared.onChange` registration and asserts it ends the session. Verified both ways — green on this commit, red when the closure is reverted to `{ [weak self] _ in self?.syncMouseTap() }`. `./build.sh --test`: `TESTS FAILED (2 of 9474)`, the same two input-source checks that fail on clean `main` on this machine (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints`), no third. `RadialMenuService.swift` is outside the `--test` whitelist, so I also ran `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` with the build's own `arm64-apple-macosx14.0` / `MacOSX26.sdk` flags: clean, no warnings.

Not verified: still no second account here, so the resign path is read, not run.

Worth re-checking after this: whether ending the session on resign is right for a *sticky* wheel — one opened by a press and left up with no button held. It now closes across a switch instead of being there on return. I think that is correct (the panel belongs to a session that is no longer on screen, and `suspend()` treats it the same way), but it is a behavior change for the non-held case and it is the part of this commit that is judgement rather than mechanism.

Refs #1153

PathGao:
@PathGao Rebased onto `main` at `a952b829` and split, head is now `366ef7be`.

**The session gate half is out, in #1270** (`366ef7be`). It references #1153
where this half references #1045, the two share no premise, and only one of
them carries a *Not tested* note — there is no second logged-in account here,
and that note should be read against one change rather than two. Your 10:11
review of the resign gate and the `endSession()` follow-up in `921ed7ea` moved
across whole, so nothing from that thread is lost; #1270 opens with it already
applied.

What went with it: the `SessionActivity.shared.onChange` registration and its
`endSession()`, the `SessionActivitySupport.tapShouldRun` wrapper around
`syncMouseTap`, the re-arm's session question, and the resign assertion.
`AXIsProcessTrusted()` is back where `main` has it, in the tap-creation guard.

Two things I want to flag about the seam, because neither is obvious from the
diff:

- **`mouseTapWanted` went with the gate, not with this half.** With the gate
  gone it had one caller, so it was the gate's shape rather than the decode's.
  The cheap read is folded back into `syncMouseTap` inline. What stays here is
  exactly the `decodeProfiles` → `claimedMouseButtons` substitution.
- **The feature check went with the gate too.** `syncWithPreferences` already
  guards on `AppFeature.radialMenu.isAvailable && radialMenuEnabled` and calls
  `suspend()`, so on `main` the check is redundant. It stops being redundant
  the moment the resign handler calls `syncMouseTap()` directly, which is
  exactly what #1270 adds — without it the tap comes back on a session switch
  with the feature switched off. It belongs to the branch that creates the
  caller that needs it.

**#1270 is chained on this branch**, because the session gate wraps the same
`syncMouseTap` guard the cheap read rewrites. `Depends on #1228` is in its
body. If this one merges first, that branch needs a rebase before it will
apply — I would rather that than have the two conflict in one function where
the resolution could silently keep the expensive decode.

**Tests.** `./build.sh --test` → `TESTS OK (9537 checks)`. The two
machine's input source is ANSI now. `swiftc -typecheck` over all of
`Sources/Vorssaint/**` is clean; `RadialMenuService.swift` is not in the
`--test` whitelist. Full `./build.sh` not run — signing keychain locked.

**Where to look first.** The three assertions left here. Two run against the
real decoders and are the ones I trust — the cheap read agrees with the full
decode over a profile set carrying a custom icon, and falls back to the legacy
button key for both a missing blob and a corrupt one. The third is the
source-shape check on `handleMouseTap` holding both halves of a claimed click,
which is the defect you caught at 10:13; it is a text check because that path
is private and needs a live `CGEvent`. If the half-click is worth a real test,
the way to get one is to lift the down/up decision out of `handleMouseTap` into
`RadialMenuSupport`, which is a change of its own.

PathGao:
The split is the right shape, and the `if let` keeps both halves of a claimed click — nothing left in that function splits one.

One thing the trade-off section rests on. `JSONDecoder` scans the whole document to find each element's `mouseButton`, and `customIconData` is base64 *inside that same document* — roughly 87 KB of text per stored 64 KB icon. What went away is the base64 decode, `NSImage(data:)` and the item graph, which is the expensive part. What is left is O(blob bytes) per event, not O(profiles), so "the items and their icons are never touched" and "there is nothing left to cache" are both stronger than what the code does.

It also runs twice per event on the navigation path: `MouseNavigationService.handle` asks `claimsMouseButton`, then `MouseButtonShortcutSupport.claimsButton` asks it again in its first guard, for any button the wheel does not own — that is every `.otherMouseDragged` of a held side button.

Not verified: I did not time it. Worth timing against an icon-heavy blob before crediting #1045.

**Impact:** none new — the residual is follow-up work, not a regression.
**Merge:** merged; nothing to gate.

ref #1045

Issue #162, opened 2026-07-04
[Bug]: Existing Shelf data is erased upon every update to the app


Sujan-003:
### Description

All the data that is stored in the shelf like links and text is completely erased when I update Vorssaint.

### Steps to Reproduce

1. Update Vorssaint 
2. Open shelf

### Vorssaint Version

3.1.7

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

PR #1229, opened 2026-09-01
fix(shelf): keep the saved shelf when its store fails to decode

PathGao:
### A store that failed to decode emptied the shelf and deleted its files

`restoreItems` read the saved blob with `try? JSONDecoder().decode([ShelfPersistedItem].self, from: data)`, which is all-or-nothing: `id`, `kind` and `title` are non-optional and `Kind` is a strict raw-value enum, so one entry with a missing field or a kind written by a newer build threw and left the restored list empty. Two irreversible things then ran on that empty list — `needsPersistAfterRestore(restoredIsEmpty: true, …)` saved it back over the real store, and `sweepOwnedFiles(keeping: [])` deleted every file in the shelf's own payload directory older than the launch cutoff. One entry this build could not read cost the whole shelf, including the pasted images and text it had written to disk itself.

Fixed at the shape, not at the one call site:

- `ShelfPersistedItem` decodes through a `FailableShelfPersistedItem` wrapper, the construct `RadialMenuSupport` already uses for menu items and profiles. A bad entry drops itself and the rest of the list survives, at any nesting depth.
- `ShelfPersistenceSupport.load(_:)` is the only way into the store and returns `ShelfStoreLoad`, not an array. `.unreadable` is a case the caller has to answer for, so "the blob did not decode" can no longer pass for "the shelf is empty". A blob that decoded to a list where nothing survived counts as unreadable too: that is a downgrade past a format change, not a shelf the user emptied.
- A blob that decoded with entries missing is `.partial`, its own case. An entry that dropped itself still owns its payload file in the shelf's directory, and the blob that survived still points at that file, so a `.partial` store is treated like `.unreadable`: no save-back, no sweep. The drop is counted at every depth, not off the top-level `decoded.count`, because a child of a batch owns a payload file the same way a top-level entry does. This is unlike a `sanitized` drop, whose file is already gone by definition and whose sweep is the point.
- Restore therefore reaches the save-back and the sweep only for a store it read whole.

Four states are now distinct where they used to be one: no blob at all (first launch), a blob that decoded whole (possibly to an empty list — a shelf the user emptied, and safe to save over), a blob that decoded with entries dropped, and a blob that does not decode at all. Only the first two may be written over and swept behind.

`ShelfPersistedItem` writes its eight `CodingKeys` by hand next to eight stored properties, and a ninth added without a case is invisible in both directions at once: the synthesized encoder has no key to write it under, and the hand-written `init(from:)` never reads it back. The round trip already in the suite could not see that — every item it encodes leaves `bookmark` and `children` at their defaults, so a field that never survives compares equal anyway. A second round trip encodes one item with all eight fields set to non-default values and compares the decoded item against it.

### Tradeoffs

- Considered guarding only the `restoreItems` call site with `data != nil && decoded == nil`. Rejected: it leaves the all-or-nothing decode expressible, so the next reader of the store writes the same bug. An enum makes the failure a case the compiler forces a caller to handle.
- Considered making the bare decode impossible outright. It cannot be — `ShelfPersistedItem` has to stay `Decodable` for the failable wrapper to work — so the nearest available guard is a source-shape check in the tests that no file under `Sources/` decodes the array directly. It keys off the persisted type's name, which is a rename away from going quiet; that is the weaker half of this change. It now also reports how many files it read and fails on zero, so a run from outside the repo root cannot pass it by scanning nothing.
- Considered saving a repaired list when only some entries were unreadable. Not done, and now the payload files behind those entries survive too: rewriting the store would make a downgrade lossy, and sweeping behind it would make it permanent.
- `restoreCompleted` still flips to true on an unreadable or partial store, so a drop or a delete the user makes afterwards does save over it. Blocking saves outright would leave the shelf silently read-only. What that later save cannot do any more is take the dropped entries' files with it in the same launch.
- The full-field round trip is only as wide as the instance it encodes. It catches a ninth field when that instance sets it to something other than its default, which is what the comment beside it asks the next reader to do; a ninth field left at its default in the test still passes. Making it independent of that would need reflection over the stored properties, a different check that is not in here.
- Not touched here: the `maxLeaves` ceiling drops restored items past 200 leaves and then sweeps their payload files, which is the same trade on a store too large rather than one too new. It predates this change and is left as it is.
- Not touched here either: restore still builds each item's icon with `NSWorkspace.icon(forFile:)` and its bookmark with `url.bookmarkData()` on the main thread, once per restored item, up to 200 — and `sanitized` deliberately keeps items on unmounted volumes, which are the paths most likely to make LaunchServices wait. Deferring those needs a placeholder icon and a second async patch channel, and it changes what launch looks like; that belongs in its own change.

## Verification

macOS 26.x on Apple silicon.

- `./build.sh --test`: `TESTS FAILED (2 of 9480)`, and both are the pre-existing pair (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints…`) that read the machine's current input source and flip with the active input method. They fail the same way on `origin/main` at `f25d1aa8` on this machine.
- `swiftc -typecheck` over the whole `Sources/Vorssaint` set, with the target, SDK and compat flags `build.sh` compiles with: clean, no diagnostics. `ShelfService.swift` is not in the `--test` file list, so this is the only thing here that reads it. The full `./build.sh` release compile finished with no warnings on the previous head of this branch; what changed in that file since is a removal, restoring the line already on `main`.
- Every check here was confirmed red before being trusted, in one run with the fixes broken on purpose. Making `load` return `.items` instead of `.partial` reddened both partial checks (`found []`). Replacing the restore guard with one that lets everything but `.unreadable` through reddened the source-shape pin on `restoreItems`. Pointing the two directory scans at a path that does not exist reddened both on their file counts (`across 0 scanned files`) where before the fix they passed on an empty scan. All four were restored and are green. The new round trip was seen red the same way: adding a ninth stored property to `ShelfPersistedItem` without a `CodingKeys` case, and setting it on the encoded item, failed that check and nothing else.
- The unreadable, partial and whole-store paths were driven through `ShelfPersistenceSupport.load` in the unit tests, including a batch whose child has a kind this build does not know.

What was not tested: the sweep itself never ran in a test — `ShelfService` is not in the `--test` set, so the rule that restore reaches `sweepOwnedFiles` only past the whole-store guard is pinned by source shape, not by executing it. No real `shelfItems` blob was corrupted and no app was relaunched against it, and no payload file was watched surviving a downgrade on disk. No second login session, no external or network volume mounted to exercise the unmounted-volume branch, and no multi-display or multi-Space layout.

## Anything else

The drag half that was on this branch has been taken back off it. `dockOriginProbeLimit = 5` narrowed "ask on every dragged event" to "ask at most five times past the start" on the strength of one line of reasoning — that five events is 40–80 ms — with nothing measuring when a Dock stack actually surfaces its window number, and no list of the gestures the old code recognised that the new one does not. `dragBeganInDock` feeds `ShelfInteractionSupport.isContentDrag(beganInDock:)`, so a number guessed too low is a user-visible regression on Dock drags, and no check here covers it. The pasteboard-answer memo beside it is clean on its own — `NSPasteboard.changeCount` is documented, and it takes one full pasteboard enumeration per caller per event down to one per gesture — but it is the same half of the branch, and it is worth a real before-and-after count of its own rather than a ride in behind the restore fix. Both are now #1265, where the guessed five has a section of its own.

One unrelated test carried the same vacuous-scan bug as the new source-shape guard and is fixed in the same commit: the borderless-menu rule enumerated `Sources/Vorssaint/UI` and passed on an empty list. It now requires the scan to have read at least one file, the way the `waitUntilAllOperationsAreFinished` and application-role scans beside it already did.

Refs #162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project are documented here. The format follows
 ## [Unreleased]
 
 ### Summary
-Vorssaint adds quit and close protections, expands screen text recognition, capture magnifier, clipboard, Super key, Dock Preview, App Switcher, Window Layout, mouse controls, snippet controls and the disk image installer, broadens app update and safe cleanup discovery, and hardens shortcut capture and key caps, permission guide recovery, input session switching, Accessibility timeouts, Volume Mixer audio rendering, helper process attribution and teardown, Super key shutdown, process termination, update and uninstallation teardown, window handling and shared on-screen parsing, pasteboard restoration, sensor selection, Cleaning Mode unlock, favicon downloads, the What's New showcase video download and stopping a recording. It also improves mouse scrolling feel, disk space and system monitor metrics, Command Bar responsiveness, Settings, Scratchpad, Screenshot Editor, floating panels and several menu bar behaviors.
+Vorssaint adds quit and close protections, expands screen text recognition, capture magnifier, clipboard, Super key, Dock Preview, App Switcher, Window Layout, mouse controls, snippet controls and the disk image installer, broadens app update and safe cleanup discovery, and hardens shortcut capture and key caps, permission guide recovery, input session switching, Accessibility timeouts, Volume Mixer audio rendering, helper process attribution and teardown, Super key shutdown, process termination, update and uninstallation teardown, window handling and shared on-screen parsing, pasteboard restoration, the docked shelf's saved items, sensor selection, Cleaning Mode unlock, favicon downloads, the What's New showcase video download and stopping a recording. It also improves mouse scrolling feel, disk space and system monitor metrics, Command Bar responsiveness, Settings, Scratchpad, Screenshot Editor, floating panels and several menu bar behaviors.
 
 ### Added
 - Settings now includes optional protections for Command-Q and Command-W with customizable hold duration, double press, extra modifier requirements and per-app scopes. Thanks to @RuanMD.
@@ -86,6 +86,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 - Text snippets now keep their trigger buffer when typed on the macOS Accessibility Keyboard. Thanks to @fermincasagrande.
 - Dock click actions and previews now stay behind fullscreen content and sample high-rate pointer movement without overloading the hover tap. Thanks to @iltonandrew.
 - Clipboard history now trims against the encoded file size before saving, preventing its store from becoming unreadable.
+- The docked shelf now keeps its saved items and the text and images it stored for them when its saved list cannot be read completely, instead of clearing the shelf and deleting those files. Thanks to @PathGao.
 - Kill Process now sorts consistently and revalidates the exact process before terminating it, preventing a recycled process identifier from targeting a different process.
 - Music launch blocking now fails open when its media-key listener is unavailable.
 - Radial Menu favicon downloads now stay on the link origin and reject cross-origin redirects, oversized payloads and unsafe image dimensions.
--- a/Sources/Vorssaint/Services/Shelf/ShelfService.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfService.swift
@@ -201,15 +201,6 @@ final class ShelfService: ObservableObject {
     /// as a content drag (issue #240).
     private var sawGestureStart = false
     private var dragBeganInDock = false
-    /// Dragged events spent re-asking whether this gesture started in the
-    /// Dock, capped so a long drag does not pay for the answer at event rate.
-    private var dockOriginProbes = 0
-    private static let dockOriginProbeLimit = 5
-    /// One gesture's answer to "does the drag pasteboard hold something the
-    /// Shelf can keep", remembered against the change count that produced it.
-    /// The contents cannot change while the count stands still, so enumerating
-    /// them once per gesture is enough.
-    private var droppableContentAnswer: (changeCount: Int, hasContent: Bool)?
     private var dragSourceBundleIdentifier: String?
     private var activeInternalDragIDs: [UUID] = []
     private var internalDragWasMerged = false
@@ -432,15 +423,7 @@ final class ShelfService: ObservableObject {
                 if !self.sawGestureStart {
                     self.beginDragGesture(with: event)
                 }
-                // Where the gesture started is settled at `beginDragGesture`;
-                // re-asking costs a window-list copy plus a running-app lookup
-                // on every dragged event. A Dock stack whose mouse-down never
-                // reached the monitor can still surface its window number a
-                // few events late, so keep asking for a short while.
-                if !self.dragBeganInDock, self.dockOriginProbes < Self.dockOriginProbeLimit {
-                    self.dockOriginProbes += 1
-                    self.dragBeganInDock = self.eventBelongsToDock(event)
-                }
+                self.dragBeganInDock = self.dragBeganInDock || self.eventBelongsToDock(event)
                 let defaults = UserDefaults.standard
                 if defaults.bool(forKey: DefaultsKey.shelfShakeToOpen) {
                     self.handleDrag(event)
@@ -511,19 +494,11 @@ final class ShelfService: ObservableObject {
 
     private func isContentDragActive() -> Bool {
         let pasteboard = NSPasteboard(name: .drag)
-        let changeCount = pasteboard.changeCount
         return ShelfInteractionSupport.isContentDrag(
             baselineChangeCount: dragBaselineChangeCount,
-            changeCount: changeCount,
+            changeCount: pasteboard.changeCount,
             beganInDock: dragBeganInDock,
-            hasDroppableContent: {
-                if let answer = droppableContentAnswer, answer.changeCount == changeCount {
-                    return answer.hasContent
-                }
-                let hasContent = pasteboardHasDroppableContent(pasteboard)
-                droppableContentAnswer = (changeCount, hasContent)
-                return hasContent
-            })
+            hasDroppableContent: { pasteboardHasDroppableContent(pasteboard) })
     }
 
     /// Opens a gesture at its first observed event: the mouse-down when it
@@ -532,7 +507,6 @@ final class ShelfService: ObservableObject {
     private func beginDragGesture(with event: NSEvent) {
         sawGestureStart = true
         dragBaselineChangeCount = NSPasteboard(name: .drag).changeCount
-        dockOriginProbes = 0
         dragBeganInDock = eventBelongsToDock(event)
         dragSourceBundleIdentifier = sourceBundleIdentifier(for: event)
     }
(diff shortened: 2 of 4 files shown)


PathGao:
`load` as the only door in, with `.unreadable` as a case the caller has to answer, kills the class and not the instance. The recursion through `children` checks out.

The partial path still loses files, and it's the case the body claims to cover. "Rewriting the store would make a downgrade lossy" — the sweep is what makes it lossy, not the rewrite. An entry the failable decode drops (unknown kind written by a newer build) still has its payload file sitting in `ShelfFiles`. It isn't in `keptPaths` and it predates `sweepCutoff`, so `sweepOwnedFiles` deletes it. The blob survives the downgrade pointing at a file that is gone. That is unlike a `sanitized` drop, where the file is already missing by definition. `load` knows it happened — `decoded.count != items.count`. Carry that out and skip the sweep the way `.unreadable` does.

Smaller: the source-shape guard goes green having checked nothing. `enumerator(atPath: "Sources")` returns nil off the repo root, `rawShelfStoreDecoders` stays empty, the expect passes. Count the files it visited and require more than zero.

Not verified here: no build run, and I read the drag changes only.

**Impact:** a downgrade past a format change permanently deletes the dropped entries' pasted images and text, while the blob it kept still points at them.
**Merge:** needs the sweep skipped when the decode dropped entries first.

PathGao:
@PathGao Right on both.

`ShelfStoreLoad` has a third case now. A blob that decoded with entries missing comes back as `.partial(items)`, and `restoreItems` guards on `case .items`, so a partial store is left exactly where `.unreadable` leaves it: blob untouched, payload directory untouched. I skipped the save-back with it and not only the sweep. With only the sweep skipped, a launch where `liveItemCount > 0` still writes the truncated list back, and the next launch decodes that whole, finds the file outside `keptPaths` and deletes it — the same loss, one launch later.

The drop is counted at every depth rather than off `decoded.count != items.count`. A child of a batch owns a payload file the same way a top-level entry does, and the top-level count cannot see a child that the `children` decode dropped, so `load` compares the entries the blob describes (walked over the raw JSON, following `children`) with the entries it read back. The batch blob in the tests is the case that separates the two.

`ShelfPersistenceSupport.load` has one caller and `sweepOwnedFiles` has one, both inside `restoreItems`, so there is no second site of this shape. The nearest relative I left alone is the `maxLeaves` ceiling: restore drops restored items past 200 leaves and then sweeps their payload files, which is the same trade on a store too large rather than one too new. It predates this branch, so it is not in here; happy to take it if you want it.

The source-shape guard now counts the files it read and fails on zero. The borderless-menu scan had the same bug — `enumerator(atPath: "Sources/Vorssaint/UI")` with a bare `unpinnedBorderlessMenus.isEmpty` — and now requires a non-empty scan, the way the `waitUntilAllOperationsAreFinished` and application-role scans beside it already did. Those two were the only vacuous ones in the suite: the other enumerator is that guarded pair, and the `availabilityKey` scan asserts an exact non-empty set, so an empty scan already fails it.

Verified: `./build.sh --test` reports 9479 checks with only the two input-source failures that also fail on `main` here, and the full `./build.sh` compiles the release set with no warnings, which is the only thing that builds `ShelfService.swift`. Every check was seen red first, in one run with the four fixes broken on purpose, then restored.

Not verified: nothing here ran the sweep. `ShelfService` is outside the `--test` set, so "restore reaches `sweepOwnedFiles` only past the whole-store guard" is pinned by source shape, not by executing it. I did not corrupt a real `shelfItems` blob, relaunch and watch a payload file survive on disk, and the drag changes are still unexercised by hand.

PathGao:
Checked the guard's placement rather than taking it: `restoreCompleted = true` sits above it, so a later drop still saves and the shelf isn't silently read-only. Counting at depth is right — the top-level count can't see a child the `children` decode dropped.

`maxLeaves`: not here. It looks like the same trade, but the fix isn't the same shape — an over-200 store is readable, so the answer can't be "skip the save-back", it has to be "stop dropping items out of the blob you then rewrite". Own issue.

One thing this adds: `CodingKeys` now lists the eight fields by hand, next to eight stored properties. Add a ninth and it silently stops being written *and* read, on the type this PR exists to protect. One round trip kills the class:

```swift
let full = ShelfPersistedItem(id: UUID(), kind: .file, title: "t", text: "x", url: "u",
                              path: "p", bookmark: Data([1]), children: [])
let round = (try? JSONEncoder().encode(full))
    .flatMap { try? JSONDecoder().decode(ShelfPersistedItem.self, from: $0) }
expect(round == full, "every persisted field survives a round trip")
```

Not verified: nothing built here, and that snippet is unrun.

**Impact:** nothing here breaks. The residual path you already declared — one drop plus one relaunch and the dropped entries' files go anyway — needs the raw entry kept and written back, which is its own change.
**Merge:** ship it.

PathGao:
@PathGao Split the drag half out in `9b9e9aef`, and added the round-trip you asked for in `532706b7`. Not a force-push — both are their own commits on top of what you reviewed.

**What went:** `dockOriginProbes` / `dockOriginProbeLimit = 5` and `droppableContentAnswer`. `ShelfService.swift` is back to `dragBeganInDock = dragBeganInDock || eventBelongsToDock(event)`. Nothing pinned either of them, so no assertion had to move. The half is saved as a `format-patch` and will come back as its own pull request.

**Why:** the probe budget narrows "ask for the whole gesture" to "ask five times", and the only thing standing behind five is one sentence of arithmetic — 40–80 ms — with no measurement and no list of what a late Dock stack would now be missed on. `dragBeganInDock` feeds `ShelfInteractionSupport.isContentDrag(beganInDock:)`, so guessing that number wrong is a visible drag regression, not a slow path. The pasteboard memo beside it is clean on its own — `NSPasteboard.changeCount` is documented, and it turns one full enumeration per event per caller into one per gesture — but it belongs to the same half and should arrive with a measurement rather than ride along with a fix to the store.

**On the `CodingKeys` gap you named:** `ShelfPersistedItem` hand-writes all eight, so a ninth field would silently stop being read *and* written, on the type this pull request exists to protect. `532706b7` encodes an instance with all eight set to non-default values and compares the decode field by field. Confirmed red by adding a ninth stored property outside `CodingKeys` and setting it in the test instance — one assertion by name, `3 of 9480`.

One limit, checked rather than assumed: adding the field and *not* setting it in the test instance leaves the assertion green. A round-trip only covers what it encodes. That is in Tradeoffs now; a reflective guard would be a different change.

**Checks:** `TESTS FAILED (2 of 9480)`, both the input-source-dependent pair. `ShelfService.swift` is outside the `--test` whitelist, so a whole-module `swiftc -typecheck` over `Sources/Vorssaint` at the build's target and SDK: exit 0, no diagnostics.

**One thing to know:** this branch now reports conflicting, and it did before I touched it — `git merge-tree origin/main d192a8b4` conflicts the same way. The conflict is `Tests/MetricsTests.swift`, from #1246 and #1244 landing. I have not rebased, since #1230 is the change that removes the roster those conflicts keep landing on.

PR #1230, opened 2026-09-01
test: enumerate event-tap owners instead of listing six by hand

PathGao:
# Enumerate event-tap owners instead of listing six by hand

## Summary

The session-switch guard in `Tests/MetricsTests.swift` walked a hardcoded list
of six files. The app has twenty-four files that call `CGEvent.tapCreate`. So
the guard covered a quarter of the taps it was written for, and a service
adding a tap tomorrow inherited nothing — the list was itself a membership
roster nobody had a reason to update, which is the same failure the guard
exists to catch, one level up. The taps that never joined the session gate are
exactly the ones that were missing from the list.

The loop now reads its subjects out of `Sources/`: every `.swift` file
containing `CGEvent.tapCreate` is checked for the four things the guard already
asserted — `SessionActivity.shared.onChange`, `SessionActivity.shared.isActive`
between `tapDisabledByTimeout` and `return`, an Accessibility check, and
`CFMachPortInvalidate` on teardown. Registering is no longer optional because
there is nothing to register.

Two consequences of enumerating:

- Listen-only taps are excused from the session checks, and the excuse is
  re-derived from the source on every run rather than written down: a file whose
  tap is `.listenOnly` and never `.defaultTap` hands every event straight back,
  so an off-screen session cannot swallow the keystrokes of the account on
  screen. Change one to `.defaultTap` and it joins the gated set the same
  second.
- The permission half of the gate is now matched through either public spelling
  — `SessionActivitySupport.tapShouldRun(`, which takes it as an argument, or
  `AXIsProcessTrusted()` for the services that predate the helper. The old rule
  named four specific files because `ScrollInverter` asks through
  `Permissions.shared.accessibility`; that carve-out was a second hand-kept
  roster inside the first.

Thirteen services are outside the gate today. They are listed in
`knownUngatedTaps` with the harm each one does while its line stands, not with
a reason it is allowed — none of them is exempt. The list is how they get
reported rather than forgiven; rewriting fourteen services in the same change
that lands the enumeration is the kind of pull request that never gets read.
Nothing may be added to it, and each fix deletes its line. A separate check
fails if an entry stops naming a file that creates a tap, so a line cannot
outlive its subject and start excusing whatever moves in under that path.

Trade-offs:

- **Text-shape test rather than a unit test.** The tap owners need the event
  chain to run, so the proposition here is an absent one — "no file creates a
  tap outside the gate" — which is not reachable from a callable seam. That is
  what the existing comment on this block already says; enumerating strengthens
  a guard that had already picked the right tool.
- **The known-ungated list is a floor, not a ratchet.** A file on it that starts
  following the gate is not flagged for removal. A strict set-equality check
  would keep the list from rotting, but it would also break every in-flight
  branch that fixes one of these thirteen the moment either side merges. The
  cheap staleness check above covers the case that actually corrupts the guard.
- **The staleness check reads the path, not the prose.** It catches an entry
  whose file stops creating a tap. It cannot catch an entry whose *description*
  goes stale, which is what happened to three of them below. Linting the prose
  would be a second guard over a thirteen-line list; re-reading the list when a
  tap fix merges is cheaper.
- **Public symbols only.** Every needle is an exported symbol
  (`CGEvent.tapCreate`, `SessionActivity.shared.onChange`, `CFMachPortInvalidate`,
  `AXIsProcessTrusted`, `SessionActivitySupport.tapShouldRun`) so a rename inside
  a service cannot turn the guard red without changing behaviour. Comments are
  stripped before matching, so prose naming an API cannot answer for it.
- **No source file is touched.** The thirteen fixes are separate changes.

### Rebased onto #1225

This branch previously carried a `CFMachPortInvalidate` in `MusicLaunchBlocker`,
`AutoQuitService` and `DockPreviewService` — the three listen-only owners the
port check newly reached. #1225 (`f7d70935`) landed that call in all three, so
after the rebase the branch contributes none of them and the diff is one test
file.

Two things changed with it:

- Three `knownUngatedTaps` entries named a missing `CFMachPortInvalidate`
  alongside the session gap — `FinderRenameService`, `AppSwitcher`,
  `WindowMaximizer`. #1225 fixed the port in all three. Each entry is supposed to
  name the harm that stands while the line does, so the port half came off and
  the session gap stayed. The `WindowLayoutService` entry said "two" taps where
  the file has three, which you flagged earlier — corrected in the same commit.
  The remaining nine entries read the same as before: all thirteen still lack
  `SessionActivity.shared.onChange` on today's `main`.
- The per-owner port check is out of this loop (`61f04555`). #1225's own sweep in
  the same test asks every owner for the port, counted per tap rather than per
  file, with no exemption for the reported gaps — a strictly stronger question,
  and green. One caveat stays: this loop walks `Sources`, the port sweep walks
  `Sources/Vorssaint`, and `Package.swift`'s two other targets (`HIDEventSystem`,
  `VMStatisticsCompat`) create no tap today. A tap landing there would get the
  session questions here and never the port question. Same fact, two roots.

### The listen-only excuse, narrowed to the checks it can justify

Handing every event back is a reason to skip the *session* checks: an off-screen
session cannot swallow the keystrokes of the account on it. It is not a reason to
skip the port. Switching a tap off leaves the process owning it, which is what the
window server waits on, and that holds whether the tap modified anything or not.

The port is the sweep's question now, asked of every owner. This loop asks only
the session checks, and the listen-only excuse covers exactly those.

| owner | session checks asked here | port asked by #1225's sweep |
|---|---|---|
| in `knownUngatedTaps` | no | yes |
| listen-only, not on the list | no | yes |
| everything else | yes | yes |

## Verification

macOS 26, Apple silicon, own build from this branch, rebased onto `ea374e1e`.

```
./build.sh --test    TESTS OK (9562 checks) on this branch
                     TESTS OK (9528 checks) on a clean tree at origin/main
                     +45 checks, zero failures on either side
```

Both runs were green. The two input-source-dependent failures this repo sees on
some machines (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row
hints…`) did not fire in either run on this machine.

Falsified the guard rather than only running it: added a throwaway
`Sources/Vorssaint/Services/ZZProbeTapService.swift` containing nothing but a
bare `.defaultTap` `CGEvent.tapCreate` and no teardown. Four assertions failed
against it without anyone registering it anywhere — the three per-owner checks by
that file's name, plus #1225's sweep reporting `1 taps, 0 invalidated` across 25
scanned owners. Deleted, back to `TESTS OK (9562 checks)`.

Re-derived the roster against today's `main` rather than trusting the version
written before #1225, #882 and #1219 merged: 24 tap owners, all under
`Sources/Vorssaint`, so the `>= 20` floor holds with four to spare; all 13
`knownUngatedTaps` paths still create a tap, so the staleness check passes; the
three listen-only exemptions are still exactly `MusicLaunchBlocker`,
`AutoQuitService` and `DockPreviewService`; and the eight owners that are neither
listed nor listen-only all pass every check, including `CleaningModeManager` and
`MouseClickDebounceService`, which the old six-name list never reached.

Still open from earlier review, both carried and neither touched here: the
re-arm slice reads only the first `tapDisabledByTimeout` site (`.dropFirst().first`),
and assertion 2 still greps for the literal `SessionActivity.shared.isActive`
rather than also accepting a lock-sampled `SessionActivitySupport.tapShouldRun`.
Both bite only once a `knownUngatedTaps` line is deleted, so they belong to the
branch that deletes it.

Not tested: nothing here runs a tap. The behaviour these assertions stand in
for — a tap handed back across a fast user switch — needs a second logged-in
account and was not exercised. The full `./build.sh` was not run. No source file
is changed, so there is nothing outside the `--test` compile whitelist to worry
about this time.

## Anything else

`RadialMenuService.swift` and `BrightnessService.swift` each have a branch in
flight adding the gate. Both are listed in `knownUngatedTaps` here; whichever
side merges second should delete the corresponding line. No ordering
dependency — the list is tolerant of a member that already complies.

#1233, #1234, #1235 and #1236 all currently report `mergeable=false` against
`main`, and the conflict is the same six-name list this change deletes. All
eleven files they touch are already named in `knownUngatedTaps`. Either ordering
resolves it — merging this last means taking this branch's side, as laid out
above; merging it first turns each of those four rebases into a deletion of the
matching `knownUngatedTaps` lines. Flagging only because the four are blocked
today and this is what unblocks them.

Refs #1153

### DIFF
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21758,12 +21758,18 @@ struct MetricsTests {
             let code = source.components(separatedBy: "\n")
                 .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
                 .joined(separator: "\n")
-            // A listen-only tap hands every event straight back, so a session
-            // that is off screen cannot swallow the keystrokes of the account
-            // on it, and the window server does not wait on the port. That is
-            // the whole reason these are excused, and it is re-derived from the
-            // source on every run: flip one to .defaultTap and it joins the
-            // gated set the same second, with nothing to remember.
+            // Releasing the port is asked of every owner, listen-only ones
+            // included: switching a tap off leaves the process holding it, and
+            // that is what the window server waits on. Nothing about handing
+            // events back changes who owns the port.
+            expect(code.contains("CFMachPortInvalidate"),
+                   "\(tapOwner) hands its tap back rather than only disabling it")
+            // The session gate is a different question, and a listen-only tap
+            // is genuinely excused from it: it returns every event unchanged,
+            // so a session that is off screen cannot swallow the keystrokes of
+            // the account on it. Re-derived from the source on every run --
+            // flip one to .defaultTap and it joins the gated set the same
+            // second, with nothing to remember.
             if code.contains(".listenOnly") && !code.contains(".defaultTap") { continue }
             if knownUngatedTaps[tapOwner] != nil { continue }
             expect(code.contains("SessionActivity.shared.onChange"),
@@ -21778,10 +21784,6 @@ struct MetricsTests {
             expect(code.contains("SessionActivitySupport.tapShouldRun(")
                     || code.contains("AXIsProcessTrusted()"),
                    "\(tapOwner) does not keep a modifying tap alive after Accessibility is lost")
-            // Switching a tap off leaves the process owning it, which is what
-            // the window server waits on; teardown must invalidate the port.
-            expect(code.contains("CFMachPortInvalidate"),
-                   "\(tapOwner) hands its tap back rather than only disabling it")
         }
         // A line that outlives its file stops naming a real gap and starts
         // excusing whatever moves in under that path.
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21758,10 +21758,15 @@ struct MetricsTests {
             let code = source.components(separatedBy: "\n")
                 .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
                 .joined(separator: "\n")
-            // Releasing the port is asked of every owner, listen-only ones
-            // included: switching a tap off leaves the process holding it, and
-            // that is what the window server waits on. Nothing about handing
-            // events back changes who owns the port.
+            // The reported gaps below are exempt from every check here,
+            // the port included: each is a named line waiting on its own fix,
+            // and asking a question the entry already answers turns the list
+            // into a wall of red that says nothing new.
+            if knownUngatedTaps[tapOwner] != nil { continue }
+            // Releasing the port is asked of every owner off that list,
+            // listen-only ones included: switching a tap off leaves the
+            // process holding it, and that is what the window server waits
+            // on. Nothing about handing events back changes who owns the port.
             expect(code.contains("CFMachPortInvalidate"),
                    "\(tapOwner) hands its tap back rather than only disabling it")
             // The session gate is a different question, and a listen-only tap
@@ -21771,7 +21776,6 @@ struct MetricsTests {
             // flip one to .defaultTap and it joins the gated set the same
             // second, with nothing to remember.
             if code.contains(".listenOnly") && !code.contains(".defaultTap") { continue }
-            if knownUngatedTaps[tapOwner] != nil { continue }
             expect(code.contains("SessionActivity.shared.onChange"),
                    "\(tapOwner) rebuilds its tap when the session comes back")
             let rearm = code.components(separatedBy: "tapDisabledByTimeout")
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21698,34 +21698,96 @@ struct MetricsTests {
             encoding: .utf8)) ?? ""
         expect(mouseAccelerationSource.contains("SessionActivitySupport.isOnConsole("),
                "mouse acceleration shares the safe initial session-state fallback")
-        for tapOwner in ["Sources/Vorssaint/Services/ScrollInverter.swift",
-                         "Sources/Vorssaint/Services/SmoothScrollService.swift",
-                         "Sources/Vorssaint/Services/MouseNavigation/MouseNavigationService.swift",
-                         "Sources/Vorssaint/Services/MouseButtons/MouseButtonShortcutService.swift",
-                         "Sources/Vorssaint/Services/MiddleClick/MiddleClickService.swift",
-                         "Sources/Vorssaint/Services/QuitProtection/QuitProtectionService.swift"] {
+        // The owners are read out of the tree, not listed here. A roster kept
+        // by hand is itself a membership list nobody maintains: the six names
+        // this loop used to carry were a sixth of the taps in the app, and the
+        // ones missing from it were exactly the ones that never joined the
+        // gate. Enumerating means the next person to call CGEvent.tapCreate
+        // cannot forget to register — there is nothing to register.
+        let tapOwnerFiles: [String] = {
+            guard let walker = FileManager.default.enumerator(atPath: "Sources") else { return [] }
+            var owners: [String] = []
+            for case let entry as String in walker where entry.hasSuffix(".swift") {
+                let path = "Sources/" + entry
+                guard let source = try? String(contentsOfFile: path, encoding: .utf8),
+                      source.contains("CGEvent.tapCreate") else { continue }
+                owners.append(path)
+            }
+            return owners.sorted()
+        }()
+        // An empty sweep would pass every check below in silence.
+        expect(tapOwnerFiles.count >= 20,
+               "the CGEvent.tapCreate sweep reaches the sources from the test's working directory")
+        // These taps were already outside the gate when the sweep replaced the
+        // roster, and none of them is exempt: the list is how they get reported
+        // rather than forgiven, so the enumeration can land without a rewrite
+        // of fourteen services in one change. Each value names the harm that
+        // stands while the line does. Nothing may be added — a tap written from
+        // now on follows the gate — and every fix deletes its line.
+        let knownUngatedTaps: [String: String] = [
+            "Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift":
+                "volume keys stay swallowed for the account switched in",
+            "Sources/Vorssaint/Services/Display/BrightnessService.swift":
+                "brightness keys stay swallowed for the account switched in",
+            "Sources/Vorssaint/Services/DockClick/DockClickService.swift":
+                "Dock clicks stall on the timeout of a switched-away session",
+            "Sources/Vorssaint/Services/Finder/FinderCutPaste.swift":
+                "Command-X in the other account's Finder reaches a dead tap first",
+            "Sources/Vorssaint/Services/Finder/FinderRenameService.swift":
+                "as FinderCutPaste, and its teardown also skips CFMachPortInvalidate",
+            "Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift":
+                "every keystroke of the account switched in pays the tap timeout",
+            "Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift":
+                "the radial trigger holds the mouse chain of a session it cannot serve",
+            "Sources/Vorssaint/Services/ShortcutRecordingTap.swift":
+                "a recorder left open across a switch re-arms into the other account",
+            "Sources/Vorssaint/Services/Snippets/TextSnippetService.swift":
+                "snippet expansion watches the other account's typing",
+            "Sources/Vorssaint/Services/SuperKey/SuperKeyService.swift":
+                "the Super Key holds both a key and a mouse tap across the switch",
+            "Sources/Vorssaint/Services/Switcher/AppSwitcher.swift":
+                "the switcher keeps its Tab tap and never invalidates the port",
+            "Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift":
+                "two window-layout taps stay in the other account's key chain",
+            "Sources/Vorssaint/Services/WindowMaximizer.swift":
+                "the maximizer keeps its tap and never invalidates the port",
+        ]
+        for tapOwner in tapOwnerFiles {
             let source = (try? String(contentsOfFile: tapOwner, encoding: .utf8)) ?? ""
             expect(!source.isEmpty, "\(tapOwner) reads back for its session-switch check")
             let code = source.components(separatedBy: "\n")
                 .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
                 .joined(separator: "\n")
+            // A listen-only tap hands every event straight back, so a session
+            // that is off screen cannot swallow the keystrokes of the account
+            // on it, and the window server does not wait on the port. That is
+            // the whole reason these are excused, and it is re-derived from the
+            // source on every run: flip one to .defaultTap and it joins the
+            // gated set the same second, with nothing to remember.
+            if code.contains(".listenOnly") && !code.contains(".defaultTap") { continue }
+            if knownUngatedTaps[tapOwner] != nil { continue }
             expect(code.contains("SessionActivity.shared.onChange"),
                    "\(tapOwner) rebuilds its tap when the session comes back")
             let rearm = code.components(separatedBy: "tapDisabledByTimeout")
                 .dropFirst().first?.components(separatedBy: "return").first ?? ""
             expect(rearm.contains("SessionActivity.shared.isActive"),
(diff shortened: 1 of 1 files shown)


PathGao:
Enumerating is right, and the falsification pass is what makes it trustworthy. Sweep confirmed against main: 24 tap files, 13 listed, 3 listen-only, 8 fully checked — the old six all still covered.

**The listen-only `continue` also skips `CFMachPortInvalidate`.** Its stated reason covers event swallowing and the window-server wait, not teardown. All three excused files — `DockPreviewService`, `AutoQuitService`, `MusicLaunchBlocker` — stop with `tapEnable(false)` + `CFRunLoopRemoveSource` and never invalidate, so each stop leaks the port. Hoist that `expect` above both `continue`s.

**Three listed files own several taps each** (`BrightnessService` 2, `SuperKeyService` 2, `WindowLayoutService` 3) and every check is file-scoped. Brightness's fix is in flight — the day that line goes, the file greens with one of two taps gated. Same shape in the rearm slice: `.dropFirst().first` reads only the first `tapDisabledByTimeout` site; loop the whole `dropFirst()`.

**"Nothing may be added" is prose.** `expect(knownUngatedTaps.count <= 13)` makes it mechanical; deletions still pass, so no in-flight branch breaks.

Didn't run the suite — counts are grep over main.

**Impact:** none on shipped behaviour; the guard proves less than the body claims.
**Merge:** needs the listen-only excuse narrowed to the session checks first.

ref #1153

PathGao:
Six branches now fix taps this sweep exempts, so here is the ordering and what changes here when they land.

**Merge this one last.** Each of the six adds its files to the hand-written roster this PR deletes wholesale, so every one of them conflicts with this branch as add-vs-delete-block. Taking this branch's side resolves it, and the matching `knownUngatedTaps` entries come out at the same time:

| lands | entries to drop here |
|---|---|
| radial-menu tap PR | RadialMenuService |
| brightness PR | BrightnessService — see the carve-out below |
| window-layout PR | WindowLayoutService, WindowMaximizer |
| finder PR | FinderCutPaste, FinderRenameService |
| keyboard PR | KeyboardDebounceService, SuperKeyService, ShortcutRecordingTap |
| switcher/snippet/volume/dock PR | AppSwitcher, TextSnippetService, PreciseVolumeRollerService, DockClickService |

That is all thirteen entries. If every one merges, `knownUngatedTaps` is empty and the sweep stands on its own.

**One assertion needs widening.** Assertion 2 greps the timeout slice for the literal `SessionActivity.shared.isActive`. Three services run their tap on a dedicated thread, where reading that Bool is a cross-thread read of state the main queue writes — `BrightnessService`, `FinderCutPaste` and `FinderRenameService` therefore sample `SessionActivitySupport.tapShouldRun` under the lock that holds the tap and re-arm off the sample. Correct code that this assertion calls a violation.

Two of the other threaded services took a third route instead: hand the event back, hop to the main queue, and ask there — honest and matching the literal. That works, but it is not the only correct answer, so the assertion should accept a re-arm that reads a lock-sampled `tapShouldRun` as well. Otherwise the guard pushes every threaded tap toward one shape for the guard's convenience.

**One reason string is already stale.** The WindowLayoutService entry says "two window-layout taps"; there are three — the directional gesture tap is the one it misses, and the maintainer's own census on #1153 already says three. The sweep counts them correctly; only the prose is wrong.

**A question this sweep answered.** #1153 asks whether ShortcutRecordingTap stays armed after its sheet closes. It does: `ShortcutRecorderButton` closes on `didResignActive`, but `CommandBarService.beginCapturingShortcut()` is a second caller and nothing hides that panel on resign, so the tap survives a switch returning nil for every key.

PathGao:
Pushed a correction, and it gives this PR a dependency.

**The listen-only excuse was too broad.** It waived all four checks. Handing every event back unchanged is a real reason to skip the session gate — an off-screen session cannot swallow the keystrokes of the account on screen. It is not a reason to skip releasing the port: switching a tap off leaves the process owning it, and that is what the window server waits on, whether or not the tap modified anything. The invalidate check now runs ahead of both exemptions; the gate checks still sit behind them.

**#1225 is exactly the missing half.** With the check widened, ten owners fail it on current main:

```
MusicLaunchBlocker  PreciseVolumeRoller  AutoQuit      Brightness   DockClick
DockPreview         FinderRename         KeyboardDebounce  AppSwitcher  WindowMaximizer
```

Those are the same ten files #1225 adds `CFMachPortInvalidate` to, with nothing left over on either side. So: **Depends on #1225.** With it merged this check is green; without it, ten reds that are all genuine.

That match is also the argument for widening it. Three of those ten are the listen-only taps this sweep was about to excuse — and someone had already concluded independently that they need the port released.

Rebased onto `ed882cd3`. Baseline is clean now that the Command-table keycap fix landed: the two input-source failures are gone, and `./build.sh --test` reports 9545 checks with only the ten above outstanding.

## Anything else

Depends on #1225

PathGao:
@PathGao Narrowed the listen-only excuse to the session checks, and the head that was here is red — worth saying first because it is measured, not read.

`bf2bcef2` moved the port check ahead of *both* exemptions. `./build.sh --test` on it gives `TESTS FAILED (10 of 9545)`: MusicLaunchBlocker, PreciseVolumeRoller, AutoQuit, Brightness, DockClick, DockPreview, FinderRename, KeyboardDebounce, AppSwitcher, WindowMaximizer, all on `hands its tap back rather than only disabling it`. That commit touches only the test file, so every listed owner that leaks its port fails at once.

Seven of those ten are `knownUngatedTaps` entries. Each is a named line already waiting on its own fix, so asking it a question its entry answers adds red that says nothing new. The port check now sits between the two exemptions instead: the reported list skips everything, and every other owner is asked for the port — listen-only included, which is the excuse your line said to narrow. Session checks still stop at the listen-only `continue`, since handing every event back genuinely does answer that one.

The three listen-only owners the check newly reaches now invalidate in their single teardown path (each nils the tap right after, so no dead port survives to be re-enabled). `./build.sh --test` is `TESTS OK (9532 checks)`. Verified red first: deleting the new invalidate from `DockPreviewService` gives exactly one failure naming that file.

Pushed as a commit on top of `bf2bcef2` rather than a rewrite, so nothing that was here is lost — if you want the wide reading instead, it is ten one-line service fixes, seven of them in files #1233, #1234, #1235, #1236 and the brightness branch currently own, so it wants to land after those rather than across them.

Not run: the full `./build.sh`. The three changed services are outside the `--test` whitelist, so a green `--test` is not evidence they type-check; CI compiles them on both runners. Nothing here runs a tap, and the fast-user-switch behaviour still needs a second account.

PathGao:
Between the two exemptions is the better placement, and the measurement carries it: seven of those ten reds are lines the list already answers.

Confirmed against main — 28 taps in 24 files, exactly three `.listenOnly` and never `.defaultTap`: `MusicLaunchBlocker`, `AutoQuitService`, `DockPreviewService`. The narrowed check reaches exactly the three you fixed, no fourth owner slips in. The bare `CFMachPortInvalidate(tap)` is fine — `removeMediaKeyTap` opens with `guard let tap = mediaKeyTap`.

**The body contradicts its own diff.** Trade-offs still ends with "**No source file is touched.** The thirteen fixes are separate changes." Three are touched. Drop the bullet; the new listen-only section already says what changed.

**#1225 stopped being a dependency and became a conflict.** It adds `CFMachPortInvalidate` to ten files, three of them these. Add-vs-add, one hunk each — whoever lands second drops three redundant hunks. Worth a line in the merge, not a hold.

Carried over, both only bite once a `knownUngatedTaps` line is deleted: the WindowLayout entry says "two" taps where main has three, and the rearm slice still reads only the first `tapDisabledByTimeout` site.

Counts are grep over main; didn't run the suite.

**Impact:** three services stop leaking their port on teardown; guard goes from 6 files to 8 fully checked plus 3 port-only.
**Merge:** ship it.

ref #1153

PathGao:
@PathGao Rebased onto `ea374e1e`, and the three source hunks are gone — the branch is one test file now.

**The `**No source file is touched.**` bullet finally matches its diff.** It did not by deletion: #1225 (`f7d70935`) landed `CFMachPortInvalidate` in all three of `MusicLaunchBlocker`, `AutoQuitService` and `DockPreviewService`, so the rebase left this branch with nothing to contribute there. Checked each one on `main` rather than trusting the stat line — `MusicLaunchBlocker.swift:123`, `AutoQuitService.swift:679`, `DockPreviewService.swift:348`. All that would have remained was the comment and a reformat, so the conflicts resolved to `main`'s side. Head is `0086e473`.

**Roster re-derived against today's `main`, not the version written before #1225, #882 and #1219.** 24 tap owners, all under `Sources/Vorssaint`, so the `>= 20` floor holds with four to spare. All 13 `knownUngatedTaps` paths still create a tap, so the staleness check passes. The three listen-only exemptions are still exactly the same three files. The eight owners that are neither listed nor listen-only all pass every check, `CleaningModeManager` and `MouseClickDebounceService` included.

Four entries were wrong, all prose, all in `0086e473`:

| entry | said | now |
|---|---|---|
| `FinderRenameService` | "its teardown also skips `CFMachPortInvalidate`" | port half dropped, session gap kept |
| `AppSwitcher` | "never invalidates the port" | same |
| `WindowMaximizer` | "never invalidates the port" | same |
| `WindowLayoutService` | "two window-layout taps" | three — the one you flagged |

**The staleness check could not have caught any of them.** It reads the path, not the description: an entry whose file stops creating a tap fails, an entry whose *harm* gets fixed underneath it does not. That is the gap #1225 walked through. I left it as a documented trade-off rather than adding a second guard to lint thirteen lines of prose.

**One thing I did not act on, because it is your call.** The per-owner `CFMachPortInvalidate` check in this loop is now redundant. #1225's own sweep in the same test asks all 24 owners for the port, counted per tap rather than per file, with no exemption for the reported gaps — strictly stronger than what this loop asks, and green. The reported-gap `continue` above it exists only to keep that weaker copy quiet. Both come out in one delete if you want it; I did not do it unasked, since it undoes the placement you confirmed at 08:38.

Carried over untouched, both bite only once a `knownUngatedTaps` line is deleted: the re-arm slice still reads only the first `tapDisabledByTimeout` site, and assertion 2 still greps the literal `SessionActivity.shared.isActive` instead of also accepting a lock-sampled `SessionActivitySupport.tapShouldRun`.

**Tests.** `./build.sh --test` → `TESTS OK (9573 checks)`. Clean tree at `origin/main` → `TESTS OK (9528 checks)`. +45 checks, zero failures on either side; the two input-source-dependent failures did not fire on this machine in either run. Falsified rather than only run: dropped a throwaway `ZZProbeTapService.swift` holding nothing but a bare `.defaultTap` `CGEvent.tapCreate` into `Sources/` — five reds, the four per-owner checks by name plus your sweep reporting `1 taps, 0 invalidated` across 25 scanned owners. Deleted, back to green.

**Ordering.** You said merge this last, and taking this branch's side resolves the add-vs-delete either way. Worth one line: #1233, #1234, #1235 and #1236 all report `mergeable=false` today against the same six-name list, and every one of the eleven files they touch is already in `knownUngatedTaps`. Landing this first turns those four rebases into line deletions instead of roster merges. Your call, not a hold.

**Re-check here first:** the four corrected entries. They are the only thing in this branch that a merge somewhere else can silently falsify, and nothing in the suite will tell you when it happens.

PathGao:
@PathGao Dropped the per-owner CFMachPortInvalidate check from the enumeration loop (61f04555); the reported-gap exemption goes back to covering only the session-gate checks. What it asked is strictly covered by main's tapOwnersWithoutInvalidate sweep (MetricsTests 21918-21942): that one counts invalidations per tap rather than per file and has no exemption list, so a gap this check could have found already goes red there. TESTS OK (9562 checks).

PathGao:
Verified the claim against main: the sweep at `MetricsTests` 21920–21942 counts per tap, strips comments, and carries no exemption list, so it does subsume what the deleted check asked. The count agrees — 9573 → 9562 is 11, exactly the owners that check still reached (24 minus the 13 listed).

**The two sweeps enumerate different roots.** This loop walks `Sources`; the port sweep walks `Sources/Vorssaint`. `Package.swift` declares three targets there — `HIDEventSystem` and `VMStatisticsCompat` are outside the port sweep entirely. Neither creates a tap today, so nothing is red, but "strictly covered" holds per question, not per file set: a tap landing in `Sources/HIDEventSystem` gets the gate questions from this loop and is never asked for the port. Same shape as the roster this PR deletes — one fact, two copies, drifting apart quietly.

**The body still describes the check that is gone.** The listen-only table says the port is asked of listen-only and of everything else; the loop asks nobody. The bullet still offers "it can come out … say so and it is a one-line delete". And the falsification reads "Five assertions failed … the four per-owner checks" — now four and three, against 9562, not 9573.

Counts are grep over main; didn't run the suite.

**Impact:** none on shipped behaviour.
**Merge:** ship it — the body edit can ride the merge.

ref #1153

PathGao:

PathGao:
Reopened as #1297.

Issue #301, opened 2026-07-19
External Monitor Blacking Out


danilo-alm:
### Description

Since the introduction of this display brightness adjustment, my external display has been blacking out, sometimes randomly, sometimes when I'm adjusting the brightness in the slider (just happened before I opened this issue). After it blacking out, waiting doesn't do anything, opening the lid of my MacBook also doesn't. To restore it, I unplug my usb-c hub and plug it in again.
I see this might be a quite difficult problem to solve, as there's little indication of how to reproduce or what might be causing it, not even a way of directly accusing Vorssaint of being the culprit, but I wanted to put this here just to let you know, in case you might know the cause now or sometime in the future. For now, I'll just uninstall this module if it happens too often.

### Steps to Reproduce

There's no clear path to reproduce it.

### Vorssaint Version

3.1.14

### macOS Version

macOS 26 (Tahoe)

### Installation Method

dmg

vorssaint:
@danilo-alm 3.1.15 stops the app from re-reading a dimmed display as its baseline, which is what kept your monitor dark. if it ever blacks out again after updating please ping me here

danilo-alm:
I got home, used the computer plugged in for ~20mins and seconds after I used the brightness keys to check if it was working properly (that other issue), it blacked out. If that seems odd and difficult to diagnose, may I suggest you include, in a future release, an option to manually send logs / diagnostics to the developer. You could attribute a specific ID to them, and we would disclose this ID in the issues to you. If setting up a server for that is too many sweats, the option could just easily pack the last x hours logs into a zip. 
I suggest that because other app I've used called symphonium (android) did it this way.

vorssaint:
3.2.0 carries the brightness work from this thread. Blackouts come down to the specific monitor and cable, so I cannot call it done from here. Update and see how yours behaves, I am keeping this open.

danilo-alm:
Update: I've had moments where I woke my laptop from suspension (closed, with external display), and the external display was not working. Removing the cable and plugging it again solved. I did not have the monitor blacking out mid-use or when I was trying to change the brightness.

danilo-alm:
Happened again today as I was in a meeting. I had adjusted the brightness, but it did not happened directly upon adjusting.

danilo-alm:
Happened again when I was not in a meeting and had just resumed from suspend, so there's no particular situation I think

PathGao:
[...]

PR #1231, opened 2026-09-01
fix(brightness): check the display fingerprint on every gamma write and hand the key taps back on a user switch

PathGao:
## Summary

Two fixes in `BrightnessService`, both cases of a rule the file already had and
one place that did not follow it.

**A gamma curve could be written to the wrong monitor.** `GammaTable` carries a
`fingerprint` for one reason, written at its declaration: display numbers are
handed out again after a reconnection, so the fingerprint is what stops one
monitor's curve from being written to another. The remembered curves are
deliberately kept across a disconnection — a hub renegotiating or a cable
settling must not force a fresh reading from a screen that is still dimmed — so
"this number now belongs to a different monitor" is a state the code expects to
be in. Four places wrote a curve; three checked the fingerprint first and the
fourth, the restore that runs just before a display is switched off, took the
table by display number and wrote it.

Rather than add the missing check, the write itself now owns it. There is one
private `writeGammaBaseline(_:dimmedBy:)`, it looks the baseline up and compares
the fingerprint before it writes, and it is the only place in the file that
calls `CGSetDisplayTransferByTable`. The three callers no longer hold a raw
table to write, so the check is not something a call site can forget: restoring
and dimming are the same write with and without a scale factor. A test pins the
call count at one, so a future write that goes around the helper fails rather
than silently reintroducing the case.

Considered and not taken: adding the fingerprint comparison at the fourth site.
It fixes the reported instance and leaves the next writer free to repeat it —
which is what had already happened three writes in. Also considered: dropping
the curve for a display number whose monitor changed. That is the behaviour the
"kept across the gap" comment exists to prevent.

`dimmedDisplays` is deliberately left holding the display through that restore.
The rebuild reads it as `appliedByApp` to decide whether the saved dim comes
back when the display returns, which is what the surrounding comment asks for.

**Both brightness key taps stayed armed through a fast user switch.** `#1153`
names this service twice. The session-defined tap and the keystroke tap were
built from `running && preferences && Accessibility`, with no session term, and
`BrightnessService` never subscribed to `SessionActivity`, so a switch left both
taps in the event chain. The window server then routes the front account's
events through a process that cannot answer and waits out the tap timeout on
each one. The keystroke tap is the worse of the two: its mask is `keyDown |
keyUp`, so every keystroke in the session goes through it.

The wiring now follows `MouseNavigationService`: subscribe in `init`, route the
install decision through `SessionActivitySupport.tapShouldRun`, ask again before
a timeout re-arm, and invalidate the mach port on teardown rather than only
disabling it — a disabled tap the process still owns is still a tap the window
server owns a reference to.

The keystroke tap runs its callback on a thread of its own rather than on the
main run loop — that is why it exists separately, since waiting on the main run
loop is what made typing stutter once already. An earlier revision read
`SessionActivity.shared.isActive` on that thread, the way
`MouseClickDebounceService` did at the time. The session-tap branches that
landed since (#1234, #1235, #1236) settled three rules for a re-arm, and this
branch now follows them:

- The re-arm decision is made on the thread that writes the answer. The
  session flag is written on the main thread, so `routeFunctionKey` hops there
  with `DispatchQueue.main.async` and asks `SessionActivity.shared.isActive`
  and `AXIsProcessTrusted()` from that block; `handleKeyEvent` already runs on
  the main run loop and asks in place. Both are modifying taps, so both ask
  Accessibility, not only the session.
- A refused re-arm goes to `syncKeyTap` instead of leaving the tap disabled
  with its thread still up. The sync then stops it for good, or rebuilds it,
  from the same gate the preference sync uses. `handleKeyEvent` defers that
  hop, since invalidating a port from its own callback stack is unsafe.
- A restart pending at the tap thread's tail goes through that gate too.
  The two `if clearFunctionKeyThread() { installFunctionKeyTap() }` tails
  became `restartFunctionKeyTapOnMain()`, the shape of
  `KeyboardDebounceService.restartTapOnMain`: hop to the main thread, call
  `syncKeyTap`.

`Permissions.shared.accessibility` (a cached published value) became
`AXIsProcessTrusted()` in that decision, matching the four other services whose
taps modify events.

### Dropped from this branch: a sampled `functionKeyTapShouldRun`

An earlier revision kept the gate's answer in a `functionKeyTapShouldRun` field
and re-armed on `functionKeyTapShouldRun && !shouldStopFunctionKeyThread`.
`syncKeyTap` set it to `wanted && wantsKeyRouting`, which is the same expression,
three lines down, that chooses between `installFunctionKeyTap()` — which sets
`shouldStopFunctionKeyThread = false` — and `removeFunctionKeyTap()`. On every
path `syncKeyTap` drives, the conjunction said exactly what
`!shouldStopFunctionKeyThread` already said, and it added a way to go stale: the
two re-installs inside `runFunctionKeyTap`'s
`if clearFunctionKeyThread() { installFunctionKeyTap() }` never set it.

## Verification

MacBook, macOS 26, `MacOSX26.sdk`.

Rebased on `main`. Full `./build.sh`: exit 0, no warnings.
`./build.sh --test` → `TESTS OK (9602 checks)`, no failures.

The gamma guardrail was checked by breaking what it guards: reintroducing the
raw unchecked `CGSetDisplayTransferByTable` at the toggle site turns it red,
and it goes green again on restore. The three re-arm assertions (both taps ask
session and Accessibility on the main thread and hand a refusal to the sync;
the keystroke tap's hop precedes its read; no thread tail goes around the
gate) were written before the code and ran red first
(`TESTS FAILED (3 of 9602)`).

**Not tested.** There is one login session on this machine, so the fast user
switching behaviour is reasoned from the shared mechanism `#1153` describes, not
measured — same caveat the issue carries. No second monitor here either, so the
gamma path was not exercised against real hardware: a software-dimmed display,
switched off and back on with a different monitor taking its display number, is
the case the fix is about and it has not been reproduced.

## Anything else

Refs #1153

#301 reports an external display blacking out, sometimes while the brightness slider is
moving, recoverable only by unplugging the hub. The gamma half of this branch is one way
that can happen: a display number handed out again after a reconnection let the
pre-switch-off restore write a dimmed baseline to whichever monitor now holds the number.
#836 attributes #301 to DDC renegotiation, which this branch does not touch, so this is
an additional candidate path and not the fix that issue is waiting for.

Refs #301

### DIFF
--- a/Sources/Vorssaint/Services/Display/BrightnessService.swift
+++ b/Sources/Vorssaint/Services/Display/BrightnessService.swift
@@ -2,6 +2,7 @@
 // Copyright (C) 2026 Vorssaint
 
 import AppKit
+import ApplicationServices
 import IOKit.graphics
 import ObjectiveC.runtime
 import os
@@ -41,12 +42,13 @@ struct BrightnessDisplay: Identifiable, Equatable {
 /// buttons use, addressed per display through its I2C service.
 ///
 /// While display control is off there are no display observers, services or
-/// I2C traffic. Keyboard light state is read only when Quick toggles opens.
-/// While display control is on, the standing resources are one screen change
-/// observer and a pair of wake observers, and no timers; everything else
-/// happens when a slider moves, a panel opens or the Mac wakes. All I2C work
-/// runs on a serial queue with the pacing displays need, and slider drags
-/// coalesce to the newest value per display.
+/// I2C traffic; the one standing subscription is the session-activity handler
+/// that hands the key taps back on a user switch. Keyboard light state is read
+/// only when Quick toggles opens. While display control is on, the standing
+/// resources add one screen change observer and a pair of wake observers, and
+/// no timers; everything else happens when a slider moves, a panel opens or
+/// the Mac wakes. All I2C work runs on a serial queue with the pacing displays
+/// need, and slider drags coalesce to the newest value per display.
 final class BrightnessService: ObservableObject {
     static let shared = BrightnessService()
 
@@ -95,8 +97,9 @@ final class BrightnessService: ObservableObject {
     /// protocol its own buttons use.
     private static let wakeSettleDelay: TimeInterval = 3
     /// Media-key tap, alive while pointer routing or the optional overlay is
-    /// on and Accessibility is granted. Its mask covers system-defined events
-    /// only, so ordinary typing never touches it.
+    /// on, Accessibility is granted and this login session is the one on
+    /// screen. Its mask covers system-defined events only, so ordinary typing
+    /// never touches it.
     private var keyTap: CFMachPort?
     private var keyTapSource: CFRunLoopSource?
     /// Second tap for keyboards that send brightness as an ordinary key
@@ -116,6 +119,11 @@ final class BrightnessService: ObservableObject {
     /// Whether the app's own overlay stands in for the system's, sampled with
     /// the tap so the tap thread never reads published state.
     private var overlayReplacesNativeOSD = false
+    /// `SessionActivitySupport.tapShouldRun`'s answer for the keystroke tap,
+    /// sampled for the same reason: the timeout re-arm runs on the tap thread,
+    /// and a tap put back into a session that is no longer on screen stalls
+    /// the account that is (issue #1075).
+    private var functionKeyTapShouldRun = false
     /// Codes whose press this app consumed, so the matching release is
     /// consumed as well and the system never sees half a key.
     private var swallowedKeyCodes = Set<Int>()
@@ -217,7 +225,16 @@ final class BrightnessService: ObservableObject {
     /// panel must not put the same slow monitor probe behind itself again.
     private var rebuildingTopology: BrightnessSupport.DisplayTopology?
 
-    private init() {}
+    private init() {
+        // A filter tap owned by a switched-away login session keeps its place
+        // in the chain, so the window server waits out the tap timeout on
+        // every event the account on screen sends (issue #1075). Hand both key
+        // taps back on resign and build them from preferences again on the
+        // way in.
+        SessionActivity.shared.onChange { [weak self] _ in
+            self?.syncKeyTap()
+        }
+    }
 
     func setKeyboardLightEnabled(_ enabled: Bool) {
         guard keyboardLightEnabled != nil, let keyboardLightBridge else { return }
@@ -434,10 +451,7 @@ final class BrightnessService: ObservableObject {
                 }
                 // A software-dimmed screen should return with its clean gamma
                 // before the saved dim level is reapplied by the rebuild.
-                if let baseline = self.gammaBaselines[display.id] {
-                    CGSetDisplayTransferByTable(display.id, baseline.count, baseline.red,
-                                                baseline.green, baseline.blue)
-                }
+                self.writeGammaBaseline(display.id, dimmedBy: nil)
             }
 
             // The gamma curve above is work-queue state; the reconfiguration
@@ -582,13 +596,11 @@ final class BrightnessService: ObservableObject {
         }
     }
 
-    /// Writes every remembered curve back, skipping any display number that
-    /// now belongs to a different monitor. Runs on the work queue.
+    /// Writes every remembered curve back, each one only to the monitor it was
+    /// read from (see `writeGammaBaseline`). Runs on the work queue.
     private func restoreAllGamma() {
-        for (id, baseline) in gammaBaselines
-        where Self.displayFingerprint(id) == baseline.fingerprint {
-            CGSetDisplayTransferByTable(id, baseline.count, baseline.red,
-                                        baseline.green, baseline.blue)
+        for id in gammaBaselines.keys {
+            guard writeGammaBaseline(id, dimmedBy: nil) else { continue }
             Self.log.log("restored gamma baseline for display \(id)")
         }
         gammaBaselines = [:]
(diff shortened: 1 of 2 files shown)


PathGao:
Funnelling four writers into one `writeGammaBaseline` and pinning the count in a test is the right shape — it kills the class, not the instance. Checked against `main`: 4 write sites, 3 had the check, none outside this file.

`functionKeyTapShouldRun` restates what `shouldStopFunctionKeyThread` already says. `syncKeyTap` clears it in exactly the branch that calls `removeFunctionKeyTap()`, which sets the stop flag — so `functionKeyTapShouldRun && !shouldStopFunctionKeyThread` is the guard that was already there. It is stale-able in a way one flag was not: an `installFunctionKeyTap()` caller added outside `syncKeyTap` leaves it false and the tap silently stops re-arming.

The "one deliberate difference" paragraph does not match `main`. Only `ScrollInverter` re-arms on a bare `isActive` — `MouseNavigationService`, `MiddleClickService`, `QuitProtectionService` and `MouseButtonShortcutService` read the full `tapShouldRun` and stop the tap on the else instead of leaving it disabled in the chain, which is what your new `removeKeyTap` comment argues for. And `MouseClickDebounceService`, the other off-main tap thread, reads `isActive` and `AXIsProcessTrusted()` live from that thread.

Not verified: I read `main`, not the merge result, and cannot build or run two sessions. You note `--test` skips this file; whether `Swift 6.0.3 compatibility` is a real compile I cannot see.

**Impact:** none today; the second flag is a future stale-false hazard on the keystroke tap.
**Merge:** hold for Build & selftest to report.

ref #1153

PathGao:
Closing the build gap this PR's description flags: the release build was run on the final tree, not the earlier one. `./build.sh` completes with zero warnings and produces the bundle.

`BrightnessService.swift` is not in the `--test` file list, so that run is the only thing compiling this change — the same gap that caught the `CGGammaValue` type error noted above.

PathGao:
@PathGao Dropped `functionKeyTapShouldRun` in `e22c2260`. Not a force-push. The gamma fingerprint half is untouched — this is 13 lines out, 7 in, one file.

**Why:** the flag was set to `wanted && wantsKeyRouting` in `syncKeyTap`, and that is the same condition that decides between `installFunctionKeyTap()` (which sets `shouldStopFunctionKeyThread = false`) and `removeFunctionKeyTap()`. So on every path `syncKeyTap` drives, `functionKeyTapShouldRun && !shouldStopFunctionKeyThread` was already `!shouldStopFunctionKeyThread`. It carried no new information, and it carried a way to go stale: the two re-installs inside `runFunctionKeyTap` — `if clearFunctionKeyThread() { installFunctionKeyTap() }` — do not set it.

The argument I gave for it does not hold either. I wrote that the keystroke tap runs on its own thread and so cannot read `SessionActivity` directly. `MouseClickDebounceService` is the same shape — its own `Thread`, its own run loop, its own `shouldStopTapThread` — and its `tapDisabledByTimeout` branch reads `SessionActivity.shared.isActive && AXIsProcessTrusted()` straight off the tap thread. The re-arm here now does the same, so both taps in this file read the gate the same way and there is one less copy of the state.

**Checks:** `TESTS FAILED (2 of 9470)`, both the input-source-dependent pair; the count is unchanged because no assertion pinned the flag. `BrightnessService.swift` is outside the `--test` whitelist, so a whole-module `swiftc -typecheck` at the build's target and SDK: exit 0.

**Worth re-checking, and a question:** the tap-owner loop in `MetricsTests` pins `rearm.contains("SessionActivity.shared.isActive")`, and `BrightnessService.swift` now satisfies that. Adding it to the loop would be an addition rather than a removal so I left it out of this commit — say if you want it in, here or in whichever branch lands the roster change.

PathGao:
@PathGao Rebased on main (both BrightnessService conflicts take main's CFMachPortInvalidate) and brought the two key-tap re-arms in line with #1234-#1236: routeFunctionKey hops to the main thread before asking session and Accessibility, a refused re-arm on either tap goes to syncKeyTap instead of staying disabled, and the two thread tails restart through restartFunctionKeyTapOnMain rather than around the gate. Three assertions ran red first (TESTS FAILED (3 of 9602)); full ./build.sh exit 0, TESTS OK (9602 checks).

PathGao:
Rebase reads clean, and both re-arms now match #1234-#1236: hop, ask, hand a refusal to `syncKeyTap`. Routing the two thread tails through the gate is what `functionKeyTapShouldRun` was standing in for, so dropping it was right.

**`KeyboardDebounceService.restartTapOnMain` is not on `main`.** The doc comment and the PR body both cite it. That service restarts by calling `start()` straight from the tap thread (both `shouldRestart` branches in `runEventTap`), and its own re-arm asks nothing. The shape you copied is `SuperKeyService.startOnMain` / `TextSnippetService.startOnMain` — point at one of those.

**Both re-arms hand-roll the gate.** `syncKeyTap` asks `SessionActivitySupport.tapShouldRun`; the re-arms inline `SessionActivity.shared.isActive, AXIsProcessTrusted()` and drop `featureWanted`. MouseNavigation, MiddleClick, MouseButtonShortcut, QuitProtection and SmoothScroll use it in both places. No behaviour gap — `keyTap == nil` and `shouldStopFunctionKeyThread` carry the preference term — but the new assertion pins this second copy, and `tapShouldRun(...)` still contains the literal it greps for.

Your question: yes, add the file to the `tapOwner` loop, it passes as written. That loop only reads the first `tapDisabledByTimeout` block, so keep the two-tap assertion.

Not verified: I read `main`, not the merge result.

**Impact:** none.
**Merge:** ship it.

ref #1153
ref #301

PathGao:

PathGao:
Reopened as #1299.

PathGao:
Reopened as #1304.

PR #1232, opened 2026-09-01
fix(build): only count signing identities that can actually sign

PathGao:
## Summary

`build.sh` and `Tools/setup-signing.sh` both decided whether the stable
self-signed identity `Vorssaint Utils Signing` was usable by reading a
`security find-identity` listing. A listing is not that answer, in either
spelling:

- **Without `-v`** it includes certificates that cannot sign. A stale one is
  handed to `codesign`, which answers `errSecInternalComponent`, and the
  script exits 1 at "Assembling and signing bundle" after a clean release
  compile. The ad-hoc fallback that exists for this case sits in the `else`
  arm and is never reached. On `--dev` the same misreading makes the guard
  skip `Tools/setup-signing.sh`, and the script's own front door then prints
  `✓ Signing identity already installed.` over the dead certificate.
- **With `-v`** it answers "is this certificate trusted", which a self-signed
  one never is — nothing signs it but itself. On a machine where `codesign`
  signs with the identity happily, `find-identity -v -p codesigning` reports
  zero valid identities. Deciding by that flag would send every local build to
  the ad-hoc fallback the identity exists to prevent, and have
  `setup-signing.sh` delete and reissue a perfectly good certificate on every
  run, changing the designated requirement each time.

So neither flag is the fix. Both scripts now ask `codesign` itself: sign a
throwaway copy of `/bin/echo` with the identity and believe the result. The
probe passes `--strip-disallowed-xattrs` like every other signing call in
`build.sh`: a copy into `$TMPDIR` carries `com.apple.provenance`, the xattr
the header names as invalidating codesign, and a probe that failed on it
would read as "no usable identity" and sign the build ad-hoc.
`build.sh` asks once through `legacy_identity_installed` and reuses the
answer; `setup-signing.sh` asks through `identity_can_sign` at its front door
and again as the post-import read-back, so an import `codesign` cannot see
fails there instead of surfacing as a silent ad-hoc build later.

The probe unlocks the dedicated keychain first. That keychain is not the login
keychain, so it is locked again after every login and nothing in the repo ever
reopens it — `setup-signing.sh` unlocks once at creation and exits early
forever after. The first signing build after a reboot therefore stopped on a
GUI prompt for a keychain password that is a literal string inside
`setup-signing.sh`, which no contributor can be expected to know. A keychain
that will not open now counts as no identity, rather than letting the probe
raise that dialog; `--dev` then rebuilds it through `setup-signing.sh`.

Two defects in that arrangement were fixed in a later round:

- **The cached verdict outlived the repair.** `--dev` asks the guard first,
  records "no", and only then runs `setup-signing.sh`. Nothing cleared
  `LEGACY_IDENTITY_USABLE`, so every later call site read the stale "no": the
  build that had just created a working identity signed itself ad-hoc and
  printed `run Tools/setup-signing.sh`. That is #1123's failure on the one
  path this branch adds the repair for. The cache is cleared where the repair
  ends, before anything reads it again.
- **`setup-signing.sh` could not have repaired anything either.** Its probe
  declared `local probe status=1`. zsh keeps `status` as a second name for
  `$?` and holds it read-only, so the declaration aborts the script at the
  first call to `identity_can_sign`, before it decides anything: the script
  exits 1 with `read-only variable: status` and `build.sh` prints its ad-hoc
  warning. The counter is named `signed`.

`git grep find-identity` reaches `build.sh` (2 — the Developer ID lookup,
which correctly keeps `-v`, and the helper), `Tools/setup-signing.sh` (2, both
through `identity_can_sign`) and the test. `.github/` has none;
`Tools/ci-setup-signing.sh` imports a certificate and never enumerates. No
other `local status` exists in any script in the tree, and
`LEGACY_IDENTITY_USABLE` is the only cached signing verdict.

Releases are not affected by any of this: the release job signs with the
Developer ID from the repository secret and `Verify release signature` pins
`certificate leaf[subject.OU] = "3D485NHW29"`, so a release cannot silently
ship ad-hoc. The damage was to everyone building locally.

### The shape check, and what was removed from it

Four rounds after the behaviour fix went in, the script shape check kept
growing. Each of those rounds reported `Impact: none today`: they changed the
test and two comments, and no behaviour changed. Those additions are removed
in `9755ed9f`, because what they bought was drift that might happen later and
what they cost was a mechanism that has to be maintained and can fail in
silence — two closures parsing zsh by string matching (`probeBody`, slicing a
function between `name() {` and a `}` at column 0; `shellLiterals`, reading
`NAME="` prefixes), and a set of index-order assertions built on top of them.
A rename, a continuation line or an indented closing brace makes either
closure return nothing, and then every assertion above it passes vacuously,
which is the failure mode those rounds kept finding in the assertions they
replaced.

Removed: the probe extracted by function name and required to be defined
exactly once; `find-identity` counted file-wide and required to equal the
count inside the probe; one `codesign` call and one verdict line with an
index ordering between them; `LEGACY_KEYCHAIN`/`KC` and
`LEGACY_KEYCHAIN_PASSWORD`/`KCPASS` pinned across the two files and each
required to be assigned exactly once; and the cache-reset check widened from
the repair invocation to every invocation.

Two more went with them because they could not be made to fail. The unlock
read over `Tools/setup-signing.sh` is answered by the creation-time unlock
next to `security create-keychain`, so deleting the probe's own unlock left it
green — it is dropped rather than left asserting nothing, with a comment
saying why. The identity lookup is kept, but as what it is: the input set the
`-v` ban reads, so that ban cannot pass vacuously, rather than a claim that
each script still asks.

What remains is the part with a behaviour fix behind it, each one re-checked
by reverting its own shell fix and watching it go red:

| Assertion | Drift that turns it red |
| --- | --- |
| no `-v` on this identity | restoring `-v` in `legacy_identity_installed` |
| `codesign --force --sign` in both scripts | stubbing the `setup-signing.sh` probe out; in `build.sh` the bundle and helper signing calls answer the same read, so the probe stub stays green there |
| `--strip-disallowed-xattrs` on every such call | dropping it from either probe |
| `build.sh` unlocks the keychain itself | deleting the unlock from the probe |
| no zsh `local status` | restoring `local probe status=1` |
| the cache is cleared after the repair | deleting `LEGACY_IDENTITY_USABLE=""` |
| the repair is invoked at all | removing the `./Tools/setup-signing.sh` call |

### Trade-offs

- The probe costs one `codesign` call per build, not per call site, because
  the answer is cached. That cache is the reason the repair has to invalidate
  it, and the shape check pins that it does.
- The guard matches on `find-identity`, `unlock-keychain`,
  `codesign --force --sign` and `--strip-disallowed-xattrs`, which are the
  `security` and `codesign` CLIs' own contract, not on helper names. The
  `codesign` read splits each line into tokens, so flag order does not matter
  to it. It reads each script as a whole rather than
  the probe body: a rename or a reindent then cannot make it read nothing, at
  the price of a second call site being able to answer for the probe. That is
  the trade the removal above takes on purpose.
- Nothing changes for a machine with a valid Developer ID: that lookup is a
  different question and still uses `-v`. What changes is that a machine with
  only an unusable self-signed certificate signs ad-hoc and builds instead of
  failing, and on `--dev` gets a working identity created for it first.
- `setup-signing.sh` is idempotent against a *usable* identity rather than any
  certificate of that name. Re-running it where the certificate has gone stale
  deletes and recreates the `vorssaint-signing` keychain, which is what the
  repair is for; where one works it is still a no-op.

## Verification

macOS 26.x, Apple silicon.

- `./build.sh --test`: 9479 of 9481 pass. The two failures are
  `nil layout falls back to ANSI semicolon` and
  `App Switcher icon-row hints describe default app and window shortcuts`;
  both read this machine's live input source, which is a Chinese IME. They
  fail identically on this branch's base, `16216d56`, run in a separate
  worktree in an earlier round: 9467 of 9469 there, same two names.
- Red-first on every assertion that is left, one drift at a time, each
  reverted back to green afterwards:
  - restoring `-v` in `legacy_identity_installed` fails "no lookup of the
    self-signed identity asks find-identity for valid identities only, found
    1 that do";
  - `local probe status=1` back in `identity_can_sign` fails
    "Tools/setup-signing.sh declares zsh's read-only `status` as a local";
  - deleting `LEGACY_IDENTITY_USABLE=""` fails "build.sh clears the cached
    identity verdict after running the repair";
  - deleting the unlock out of `legacy_identity_installed` fails "build.sh
    unlocks the dedicated signing keychain itself";
  - replacing the `codesign --force --sign` call in `identity_can_sign` with
    a stub fails "Tools/setup-signing.sh asks codesign whether the identity
    can sign before relying on it";
  - dropping `--strip-disallowed-xattrs` from either probe fails that
    script's "passes --strip-disallowed-xattrs to every codesign --force
    --sign call, found 1 without it", one script at a time;
  - removing the `./Tools/setup-signing.sh` invocation fails "an
    identity-less Developer build invokes Tools/setup-signing.sh itself".
- Two drifts stay green, measured rather than assumed. Deleting the unlock
  out of `identity_can_sign` in `Tools/setup-signing.sh` leaves the suite at
  2 of 9481, the two input-source failures and nothing else, because the
  creation-time unlock answers a file-wide read. That is why the assertion
  for that file is gone rather than kept. Stubbing the probe's `codesign` out
  of `legacy_identity_installed` in `build.sh` also stays at 2 of 9481: once
  the probe passes the same flags as the bundle and helper signing calls,
  those calls answer the file-wide read for it. Before the flag was added
  the probe was the only `codesign --force --sign` literal in `build.sh`, so
  that assertion was probe-specific by accident of flag order, not by design.
- `cp /bin/echo` into `$TMPDIR`, and into a `mktemp -d` directory under it,
  was measured with `xattr -l`: `/bin/echo` itself has no xattrs, the copy
  has `com.apple.provenance`. Whether that xattr alone makes `codesign` fail
  was not measured, for the reason below.
- `zsh -n build.sh` and `zsh -n Tools/setup-signing.sh`: both parse.
- The release claim in the comments was checked against the two files it
  names, not from memory: `Tools/ci-setup-signing.sh` lines 12-31 import
  `SIGNING_CERT_P12` and never create a certificate, and
  `.github/workflows/release.yml` runs it with `REQUIRE_SIGNING: "1"` and then
  verifies against `anchor apple generic and certificate
  leaf[subject.OU] = "3D485NHW29"` at four call sites.
- Evidence from earlier rounds still stands: the `status` defect was
  reproduced by extracting the real `identity_can_sign` body and running it
  under `zsh` with `security` and `codesign` stubbed — `local probe status=1`
  aborts at that line with `read-only variable: status`, exit 1, before the
  probe runs; with `signed` the same body returns usable for a stub that signs
  and unusable for one that fails.

**What this branch changes has never been run against real `security` or
`codesign`.** The new criterion is "`codesign` can sign with the identity
**and** the dedicated keychain can be unlocked". Neither half has been
exercised on any machine:

- No `security` command, no `Tools/setup-signing.sh`, no `--dev`, and no full
  `./build.sh` has been run here since the criterion changed. The signing
  keychain on this machine is locked, and each of those reaches `codesign` or
  `security` against it and blocks on a SecurityAgent password dialog that
  never returns. All evidence for the new criterion is stubs plus `zsh -n`.
  The `-v` measurements that motivated dropping that flag are in the PR
  discussion, from an earlier round on an unlocked keychain — they are
  evidence against `-v`, not evidence for the unlock-and-sign probe. The
  `--strip-disallowed-xattrs` on the probe has likewise never been run: the
  xattr on the copy is measured, the probe signing with and without the flag
  is not.
- No machine with a valid Developer ID identity. That branch was not taken.

The consequence is worth naming, because it is #1123's failure mode. In
`build.sh`:

```sh
if security find-identity -p codesigning 2>/dev/null | grep -q "$LEGACY_IDENTITY" \
    && { [[ ! -f "$LEGACY_KEYCHAIN" ]] \
        || security unlock-keychain -p "$LEGACY_KEYCHAIN_PASSWORD" \
            "$LEGACY_KEYCHAIN" 2>/dev/null; }; then
```

A failed unlock is read as "no usable identity". On a non-`--dev` build there
is no repair path, so the build signs ad-hoc and the developer's
Accessibility and Screen Recording grants stop surviving rebuilds — the exact
#1123 symptom, now reachable through a keychain whose password drifted, whose
path moved, or which the user relocated. Whether that is a real risk or a
theoretical one is exactly what has not been measured. The check that would
settle it is one run of `./build.sh --dev` on a Mac whose signing keychain is
locked, followed by `codesign -dv --verbose=4 build/Vorssaint.app` to see
which identity it actually used.

## Anything else

Refs #1123

### DIFF
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21280,10 +21280,12 @@ struct MetricsTests {
         // before signing. The needle is the invocation at the start of a
         // command line: the ad-hoc fallback's advice string also names the
         // script, and must not satisfy this check.
-        let runsSigningSetup = buildScript.components(separatedBy: "\n").contains {
+        let invokesSigningSetup: (String) -> Bool = {
             $0.range(of: #"^\s*(if\s+!?\s*)?\./Tools/setup-signing\.sh"#,
                      options: .regularExpression) != nil
         }
+        let runsSigningSetup = buildScript.components(separatedBy: "\n")
+            .contains(where: invokesSigningSetup)
         expect(runsSigningSetup,
                "an identity-less Developer build invokes Tools/setup-signing.sh itself")
 
@@ -21329,32 +21331,59 @@ struct MetricsTests {
         // time where it creates it, so a file-wide search answers yes with the
         // probe's own unlock deleted — and then the first signing build after
         // a reboot is back on the SecurityAgent dialog this branch removes.
-        for (script, lines, probeName) in
-            [("build.sh", buildScriptLines, "legacy_identity_installed"),
-             ("Tools/setup-signing.sh", signingSetupLines, "identity_can_sign")] {
+        for (script, lines, probeName, verdict) in
+            [("build.sh", buildScriptLines, "legacy_identity_installed",
+              "LEGACY_IDENTITY_USABLE=yes"),
+             ("Tools/setup-signing.sh", signingSetupLines, "identity_can_sign", "signed=0")] {
+            // Once, not at least once: a second definition later in the file is
+            // the one the shell runs, and everything below here reads the first.
+            let definitions = lines.filter {
+                $0.trimmingCharacters(in: .whitespaces).hasPrefix("\(probeName)() {")
+            }
+            expect(definitions.count == 1,
+                   "\(script) defines \(probeName)() exactly once, found \(definitions.count)")
             let probe = probeBody(lines, probeName)
             expect(!probe.isEmpty, "\(script) still defines \(probeName)()")
             let probeCode = probe.joined(separator: "\n")
+            let namedQueries = lines
+                .filter { $0.contains("find-identity") }
+                .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
             let identityQueries = probe
                 .filter { $0.contains("find-identity") }
                 .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
             expect(!identityQueries.isEmpty,
                    "\(script)'s \(probeName) still looks the stable identity up by name")
+            // Presence is not the whole guarantee. The defect this branch fixes
+            // was call sites deciding off the raw listing, copy-pasted five
+            // times; a sixth added beside the probe leaves the probe intact and
+            // brings the bug straight back. The listing may be asked for this
+            // identity only where the answer is then handed to codesign.
+            expect(namedQueries.count == identityQueries.count,
+                   "\(script) asks find-identity for the stable identity only inside "
+                    + "\(probeName), found \(namedQueries.count - identityQueries.count) "
+                    + "such lookup(s) outside it")
             // This one stays file-wide on purpose: it forbids a spelling
             // rather than requiring a line, and -v against this identity is
             // wrong wherever it is written.
-            let trustQueries = lines
-                .filter { $0.contains("find-identity") }
-                .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
-                .filter { $0.contains("-v") }
+            let trustQueries = namedQueries.filter { $0.contains("-v") }
             expect(trustQueries.isEmpty,
                    "\(script) looks the self-signed identity up without asking find-identity "
                     + "for valid identities only, found \(trustQueries.count) that do")
             expect(probeCode.contains("unlock-keychain -p"),
                    "\(script)'s \(probeName) unlocks the dedicated signing keychain itself")
-            expect(probeCode.contains("codesign --force --sign"),
-                   "\(script)'s \(probeName) asks codesign whether the identity can sign "
-                    + "before relying on it")
+            // Requiring the codesign call is not enough by itself: a line
+            // setting the verdict unconditionally leaves the call in place and
+            // answers yes without it. Exactly one call, exactly one verdict,
+            // and the verdict no earlier than the call it is meant to read.
+            let signingProbes = probe.indices.filter {
+                probe[$0].contains("codesign --force --sign")
+            }
+            let verdicts = probe.indices.filter { probe[$0].contains(verdict) }
+            expect(signingProbes.count == 1 && verdicts.count == 1
+                    && verdicts[0] >= signingProbes[0],
+                   "\(script)'s \(probeName) reaches `\(verdict)` only through asking codesign "
+                    + "whether the identity can sign, found \(signingProbes.count) call(s) and "
+                    + "\(verdicts.count) verdict line(s)")
             // zsh keeps `status` as a second name for $?, and it is read-only:
             // `local status=1` aborts the script on the first call to the
             // function that declares it, so the probe never runs at all.
@@ -21375,37 +21404,47 @@ struct MetricsTests {
         // then goes true and the unlock is skipped without a word, leaving
         // codesign to meet a locked keychain and raise the SecurityAgent
         // dialog. Pin both values against each other, not just the call.
-        let shellLiteral: (String, [String]) -> String? = { name, lines in
+        // Every assignment, not the first: reading one and comparing it would
+        // pass a second assignment added below it, and the shell obeys the last
+        // one to run. So the value has to be written exactly once per script.
+        let shellLiterals: (String, [String]) -> [String] = { name, lines in
             lines.compactMap { line -> String? in
                 let trimmed = line.trimmingCharacters(in: .whitespaces)
                 guard trimmed.hasPrefix("\(name)=\"") else { return nil }
                 return String(trimmed.dropFirst(name.count + 2).prefix { $0 != "\"" })
-            }.first
+            }
         }
         for (what, inBuildScript, inSetupScript) in
             [("password", "LEGACY_KEYCHAIN_PASSWORD", "KCPASS"),
              ("path", "LEGACY_KEYCHAIN", "KC")] {
-            let fromBuild = shellLiteral(inBuildScript, buildScriptLines)
-            let fromSetup = shellLiteral(inSetupScript, signingSetupLines)
-            expect(fromBuild?.isEmpty == false && fromBuild == fromSetup,
+            let fromBuild = shellLiterals(inBuildScript, buildScriptLines)
+            let fromSetup = shellLiterals(inSetupScript, signingSetupLines)
+            expect(fromBuild.count == 1 && fromBuild.first?.isEmpty == false
+                    && fromBuild == fromSetup,
                    "build.sh's \(inBuildScript) and Tools/setup-signing.sh's \(inSetupScript) "
-                    + "are the same signing keychain \(what), found \(fromBuild ?? "none") "
-                    + "and \(fromSetup ?? "none")")
+                    + "are each assigned once and name the same signing keychain \(what), "
+                    + "found \(fromBuild) and \(fromSetup)")
         }
 
         // The signing probe is asked once and cached. Running the repair is the
         // only thing that changes its answer, so the cache has to be dropped
         // there: otherwise --dev creates a working identity and then signs the
         // build ad-hoc off the "no" recorded a moment earlier.
-        if let repair = buildScriptLines.firstIndex(where: { $0.contains("Tools/setup-signing.sh") }) {
+        // Every invocation, located the same way runsSigningSetup locates one:
+        // a bare substring also matches the ad-hoc fallback's advice string,
+        // which names the script and repairs nothing. And a second repair site
+        // added later is the same bug again, so each one carries its own reset.
+        let repairs = buildScriptLines.indices.filter { invokesSigningSetup(buildScriptLines[$0]) }
+        expect(!repairs.isEmpty,
+               "build.sh still runs Tools/setup-signing.sh when no identity is usable")
+        for repair in repairs {
             let afterRepair = Array(buildScriptLines[(repair + 1)...])
             let nextRead = afterRepair.firstIndex { $0.contains("legacy_identity_installed") }
                 ?? afterRepair.count
             expect(afterRepair[..<nextRead].contains { $0.contains("LEGACY_IDENTITY_USABLE=\"\"") },
-                   "build.sh clears the cached identity verdict after running the repair, "
+                   "build.sh clears the cached identity verdict after running the repair at "
+                    + "`\(buildScriptLines[repair].trimmingCharacters(in: .whitespaces))`, "
                     + "before anything reads it again")
-        } else {
-            expect(false, "build.sh still runs Tools/setup-signing.sh when no identity is usable")
         }
 
         // The setup script must run against the stock /usr/bin/openssl, which
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21280,12 +21280,10 @@ struct MetricsTests {
         // before signing. The needle is the invocation at the start of a
         // command line: the ad-hoc fallback's advice string also names the
         // script, and must not satisfy this check.
-        let invokesSigningSetup: (String) -> Bool = {
+        let runsSigningSetup = buildScript.components(separatedBy: "\n").contains {
             $0.range(of: #"^\s*(if\s+!?\s*)?\./Tools/setup-signing\.sh"#,
                      options: .regularExpression) != nil
         }
-        let runsSigningSetup = buildScript.components(separatedBy: "\n")
-            .contains(where: invokesSigningSetup)
         expect(runsSigningSetup,
                "an identity-less Developer build invokes Tools/setup-signing.sh itself")
 
@@ -21315,75 +21313,29 @@ struct MetricsTests {
         // setup-signing.sh knows.
         let buildScriptLines = buildScript.components(separatedBy: "\n")
             .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("#") }
-        // The lines of a top-level shell function, closing brace at column 0.
-        let probeBody: ([String], String) -> [String] = { lines, name in
-            guard let start = lines.firstIndex(where: {
-                $0.trimmingCharacters(in: .whitespaces).hasPrefix("\(name)() {")
-            }) else { return [] }
-            let rest = lines[(start + 1)...]
-            guard let end = rest.firstIndex(of: "}") else { return [] }
-            return Array(rest[..<end])
-        }
-        // Asserted per script, not over the pair: a union is satisfied by one
-        // surviving line in either file, so deleting the lookup out of
-        // setup-signing.sh would pass on build.sh's. And per probe, not over
-        // the whole script: setup-signing.sh unlocks the keychain a second
-        // time where it creates it, so a file-wide search answers yes with the
-        // probe's own unlock deleted — and then the first signing build after
-        // a reboot is back on the SecurityAgent dialog this branch removes.
-        for (script, lines, probeName, verdict) in
-            [("build.sh", buildScriptLines, "legacy_identity_installed",
-              "LEGACY_IDENTITY_USABLE=yes"),
-             ("Tools/setup-signing.sh", signingSetupLines, "identity_can_sign", "signed=0")] {
-            // Once, not at least once: a second definition later in the file is
-            // the one the shell runs, and everything below here reads the first.
-            let definitions = lines.filter {
-                $0.trimmingCharacters(in: .whitespaces).hasPrefix("\(probeName)() {")
-            }
-            expect(definitions.count == 1,
-                   "\(script) defines \(probeName)() exactly once, found \(definitions.count)")
-            let probe = probeBody(lines, probeName)
-            expect(!probe.isEmpty, "\(script) still defines \(probeName)()")
-            let probeCode = probe.joined(separator: "\n")
-            let namedQueries = lines
-                .filter { $0.contains("find-identity") }
-                .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
-            let identityQueries = probe
-                .filter { $0.contains("find-identity") }
-                .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
-            expect(!identityQueries.isEmpty,
-                   "\(script)'s \(probeName) still looks the stable identity up by name")
-            // Presence is not the whole guarantee. The defect this branch fixes
-            // was call sites deciding off the raw listing, copy-pasted five
-            // times; a sixth added beside the probe leaves the probe intact and
-            // brings the bug straight back. The listing may be asked for this
-            // identity only where the answer is then handed to codesign.
-            expect(namedQueries.count == identityQueries.count,
-                   "\(script) asks find-identity for the stable identity only inside "
-                    + "\(probeName), found \(namedQueries.count - identityQueries.count) "
-                    + "such lookup(s) outside it")
-            // This one stays file-wide on purpose: it forbids a spelling
-            // rather than requiring a line, and -v against this identity is
-            // wrong wherever it is written.
-            let trustQueries = namedQueries.filter { $0.contains("-v") }
-            expect(trustQueries.isEmpty,
-                   "\(script) looks the self-signed identity up without asking find-identity "
-                    + "for valid identities only, found \(trustQueries.count) that do")
-            expect(probeCode.contains("unlock-keychain -p"),
-                   "\(script)'s \(probeName) unlocks the dedicated signing keychain itself")
-            // Requiring the codesign call is not enough by itself: a line
-            // setting the verdict unconditionally leaves the call in place and
-            // answers yes without it. Exactly one call, exactly one verdict,
-            // and the verdict no earlier than the call it is meant to read.
-            let signingProbes = probe.indices.filter {
-                probe[$0].contains("codesign --force --sign")
-            }
-            let verdicts = probe.indices.filter { probe[$0].contains(verdict) }
-            expect(signingProbes.count == 1 && verdicts.count == 1
-                    && verdicts[0] >= signingProbes[0],
-                   "\(script)'s \(probeName) reaches `\(verdict)` only through asking codesign "
-                    + "whether the identity can sign, found \(signingProbes.count) call(s) and "
-                    + "\(verdicts.count) verdict line(s)")
+        // The set the -v ban reads, and the reason it cannot pass vacuously.
+        // It is a union across the two scripts on purpose: this is the input
+        // for a negative, not a claim that each script still asks.
+        let identityQueries = (buildScriptLines + signingSetupLines)
+            .filter { $0.contains("find-identity") }
+            .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
+        expect(!identityQueries.isEmpty,
+               "the stable identity is still looked up by name in these two scripts")
+        let trustQueries = identityQueries.filter { $0.contains("-v") }
+        expect(trustQueries.isEmpty,
+               "no lookup of the self-signed identity asks find-identity for valid "
+                + "identities only, found \(trustQueries.count) that do")
+        // build.sh's only unlock is the probe's own, so deleting it fails here.
+        // Tools/setup-signing.sh unlocks a second time where it creates the
+        // keychain, which answers a file-wide read on its own — so the probe's
+        // unlock is left unasserted there rather than asserted vacuously.
+        expect(buildScriptLines.joined(separator: "\n").contains("unlock-keychain -p"),
+               "build.sh unlocks the dedicated signing keychain itself")
+        for (script, lines) in [("build.sh", buildScriptLines),
+                                ("Tools/setup-signing.sh", signingSetupLines)] {
+            let code = lines.joined(separator: "\n")
+            expect(code.contains("codesign --force --sign"),
+                   "\(script) asks codesign whether the identity can sign before relying on it")
             // zsh keeps `status` as a second name for $?, and it is read-only:
             // `local status=1` aborts the script on the first call to the
             // function that declares it, so the probe never runs at all.
@@ -21398,53 +21350,19 @@ struct MetricsTests {
                     + "the script, found \(readOnlyLocals.count)")
         }
 
-        // Each unlock names the keychain and its password with a literal of
-        // its own. Password drift fails the unlock and the build signs ad-hoc
-        // in silence; path drift is worse, because `[[ ! -f "$LEGACY_KEYCHAIN" ]]`
-        // then goes true and the unlock is skipped without a word, leaving
-        // codesign to meet a locked keychain and raise the SecurityAgent
-        // dialog. Pin both values against each other, not just the call.
-        // Every assignment, not the first: reading one and comparing it would
-        // pass a second assignment added below it, and the shell obeys the last
-        // one to run. So the value has to be written exactly once per script.
-        let shellLiterals: (String, [String]) -> [String] = { name, lines in
-            lines.compactMap { line -> String? in
-                let trimmed = line.trimmingCharacters(in: .whitespaces)
-                guard trimmed.hasPrefix("\(name)=\"") else { return nil }
-                return String(trimmed.dropFirst(name.count + 2).prefix { $0 != "\"" })
-            }
-        }
-        for (what, inBuildScript, inSetupScript) in
-            [("password", "LEGACY_KEYCHAIN_PASSWORD", "KCPASS"),
-             ("path", "LEGACY_KEYCHAIN", "KC")] {
-            let fromBuild = shellLiterals(inBuildScript, buildScriptLines)
-            let fromSetup = shellLiterals(inSetupScript, signingSetupLines)
-            expect(fromBuild.count == 1 && fromBuild.first?.isEmpty == false
-                    && fromBuild == fromSetup,
-                   "build.sh's \(inBuildScript) and Tools/setup-signing.sh's \(inSetupScript) "
-                    + "are each assigned once and name the same signing keychain \(what), "
-                    + "found \(fromBuild) and \(fromSetup)")
-        }
-
         // The signing probe is asked once and cached. Running the repair is the
         // only thing that changes its answer, so the cache has to be dropped
         // there: otherwise --dev creates a working identity and then signs the
         // build ad-hoc off the "no" recorded a moment earlier.
-        // Every invocation, located the same way runsSigningSetup locates one:
-        // a bare substring also matches the ad-hoc fallback's advice string,
-        // which names the script and repairs nothing. And a second repair site
-        // added later is the same bug again, so each one carries its own reset.
-        let repairs = buildScriptLines.indices.filter { invokesSigningSetup(buildScriptLines[$0]) }
-        expect(!repairs.isEmpty,
-               "build.sh still runs Tools/setup-signing.sh when no identity is usable")
-        for repair in repairs {
+        if let repair = buildScriptLines.firstIndex(where: { $0.contains("Tools/setup-signing.sh") }) {
             let afterRepair = Array(buildScriptLines[(repair + 1)...])
             let nextRead = afterRepair.firstIndex { $0.contains("legacy_identity_installed") }
                 ?? afterRepair.count
             expect(afterRepair[..<nextRead].contains { $0.contains("LEGACY_IDENTITY_USABLE=\"\"") },
-                   "build.sh clears the cached identity verdict after running the repair at "
-                    + "`\(buildScriptLines[repair].trimmingCharacters(in: .whitespaces))`, "
+                   "build.sh clears the cached identity verdict after running the repair, "
                     + "before anything reads it again")
+        } else {
+            expect(false, "build.sh still runs Tools/setup-signing.sh when no identity is usable")
         }
 
         // The setup script must run against the stock /usr/bin/openssl, which
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21294,30 +21294,39 @@ struct MetricsTests {
             .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("#") }
         let signingSetupCode = signingSetupLines.joined(separator: "\n")
 
-        // MARK: Signing identity checks only ever count identities that can sign
-        // `security find-identity` without -v lists certificates that cannot
-        // sign, an expired self-signed one included. Matching one of those made
-        // build.sh prefer an unusable identity over the ad-hoc fallback, and
-        // codesign then failed the whole build with errSecInternalComponent --
-        // on a machine whose only fault was a stale certificate left in the
-        // keychain. setup-signing.sh asks the same question twice and both
-        // answers matter: the front-door check decides whether the repair runs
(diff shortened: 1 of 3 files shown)


PathGao:
Four of the five call sites are right, and pinning the flag rather than the helper name is the correct guard.

The `--dev` half doesn't land. `Tools/setup-signing.sh` asks the same unvalidated question at its own front door:

```sh
# setup-signing.sh:28
if security find-identity -p codesigning 2>/dev/null | grep -q "$IDENTITY"; then
    echo "✓ Signing identity already installed."; exit 0
```

build.sh now correctly decides no usable identity exists and calls the repair — which sees the dead certificate, prints ✓, exits 0. build.sh then signs ad-hoc. That is #1123's failure reached by a new route, with a success message on top. Line 64, the post-import read-back, has the same spelling and can green-light an import codesign cannot see.

The test only reads `build.sh`, so it catches neither. `signingSetupCode` is already loaded three lines below for the `-legacy` check — same filter, same expect.

I didn't run `--dev` either; this is read from source, not observed.

**Impact:** `--dev` against a keychain holding a stale legacy cert still builds ad-hoc, so Accessibility grants still die on every rebuild.
**Merge:** needs `-v` in `Tools/setup-signing.sh` first.

ref #1123

PathGao:
@PathGao You're right, and the second line matters as much as the first.

`Tools/setup-signing.sh` asked the unvalidated question twice. Line 28 is the front door: with `build.sh` fixed it correctly calls the repair, the repair sees the stale certificate, prints `✓ Signing identity already installed.`, exits 0, and `build.sh` signs ad-hoc — #1123's failure with a success message on top, exactly as you read it. Line 64 is the post-import read-back, whose entire job is to prove `codesign` can see what was just imported; without `-v` it green-lights an import `codesign` cannot use. Both now ask with `-v`.

I edited those two in place rather than adding a helper. Two calls in a 68-line script don't need one, and the shape check covers them either way.

Those two are the only other call sites. `git grep find-identity` over the whole tree reaches `build.sh` (2, both behind the helper), `Tools/setup-signing.sh` (2), and the test. `.github/` has none, and nothing else enumerates identities — no `find-certificate`, no other `-p codesigning` reader.

The check now reads both files, as you suggested — `signingSetupCode` was already loaded below for the `-legacy` check, so it moved above the identity block and the line array feeds both. It also asserts each script still asks the question at all, so deleting a call site can't pass for fixing it.

Verified: `./build.sh --test` is green apart from the two pre-existing failures on this machine that read the live input source. Red-first on the new half specifically — I restored line 28 to the unvalidated spelling, got `found 1 without -v`, reverted, green. That confirms the check genuinely reads the second file rather than passing on `build.sh` alone. `zsh -n` parses both scripts. Full `./build.sh` also re-run: `EXIT=0`, zero warnings, and it took the ad-hoc branch on this machine as expected.

Not verified: I still have not run `--dev`, and I did not run `Tools/setup-signing.sh` itself. Reaching either means letting it delete and rewrite a keychain on this machine. Both changes are the same substitution as the other four, but they are read from source, not observed — same position you took.

PathGao:
@PathGao The `-v` half of this branch is wrong, and I have replaced it. Both scripts now ask codesign whether the identity can sign, rather than asking `find-identity` a question it does not answer.

`find-identity -v` means "the certificate passes trust evaluation". A self-signed certificate never does — nothing signs it but itself — so the validated search excludes the very identity `Tools/setup-signing.sh` creates. On my machine, with that identity installed and working:

```
$ security find-identity -v -p codesigning
     0 valid identities found

$ security find-identity -p codesigning
  1) DF40E24B... "Vorssaint Utils Signing" (CSSMERR_TP_NOT_TRUSTED)

$ security find-certificate -c "Vorssaint Utils Signing" -p | openssl x509 -checkend 0
Certificate will not expire            # notAfter=Aug 29 2036

$ codesign --force --sign "Vorssaint Utils Signing" ./probe
./probe: replacing existing signature  # signs fine
```

So on this branch as it stood, `legacy_identity_installed` was false on every machine that has the identity, and:

- every local build signed ad-hoc — the exact loss of Accessibility and Screen Recording grants the identity exists to prevent;
- `setup-signing.sh` never took its front door, so every `--dev` build deleted the keychain and issued a *new* certificate, changing the designated requirement each time;
- its post-import read-back could never pass, so the script would exit 1 right after importing a good identity.

Releases are not affected: the release job signs with the Developer ID from the repository secret and `Verify release signature` pins `certificate leaf[subject.OU] = "3D485NHW29"`, so a release cannot silently ship ad-hoc. The damage was to everyone building locally.

The original complaint on this branch stands — an unvalidated listing includes certificates that cannot sign, which is how a stale one failed a build with `errSecInternalComponent`. The fix is to stop reading the listing as a verdict: `legacy_identity_installed` now unlocks the keychain, signs a throwaway copy of `/bin/echo`, and caches the answer for the rest of the build. `identity_can_sign` in `setup-signing.sh` does the same at both call sites.

The unlock is a second defect, and the one that sent me here. The signing keychain is not the login keychain, so it is locked again after every login, and nothing in the repo ever unlocks it — `setup-signing.sh` unlocks once at creation and exits early forever after. The first signing build after a reboot therefore stops on a GUI dialog asking for the keychain password, which is a literal string inside `setup-signing.sh` and not something a contributor can be expected to know. Every developer building this repo hits it. A keychain that will not open is now treated as no identity at all, rather than letting codesign raise that dialog; `--dev` then rebuilds it through `setup-signing.sh`.

Verified against a throwaway certificate in a scratch keychain, so the real one was never touched:

| case | result |
|---|---|
| keychain unlocked | usable |
| keychain locked, as after every login | unlocked and signed, no prompt |
| keychain locked, password does not match | unusable, no prompt |
| identity absent | unusable |

`./build.sh` then picked the identity for real: `signing with legacy self-signed identity: Vorssaint Utils Signing`, exit 0.

`./build.sh --test`: 9475 of 9477 pass. The two failures are this machine's input source, not this branch — `nil layout falls back to ANSI semicolon` gets `⌃⌥⌘ ；` and the App Switcher hint expects `⌘ \``. Both read the current input method where they should read the layout, which is what #1220 is about. They fail on the parent commit for the same reason and will pass on a US-layout runner.

PathGao:
The `-v` diagnosis is right, and asking codesign instead of reading a listing is the correct question. Two problems in the new half.

**The cache outlives the repair.** `--dev` calls `legacy_identity_installed` in the guard, caches `no`, then runs `setup-signing.sh` — and nothing clears `LEGACY_IDENTITY_USABLE`. Every later call site reads the cached `no`, so the build that just created a working identity signs ad-hoc and prints `run Tools/setup-signing.sh`. That is #1123 again, on the exact path this PR adds the repair for. One line after that `fi`:

```sh
LEGACY_IDENTITY_USABLE=""
```

**`local probe status=1` in `identity_can_sign`.** zsh documents `status` as another name for `?`, so `return $status` reads the exit code of the preceding `rm -rf` — always 0 — and the front door then always reports usable. I could not run zsh here, so this is read from the manual, not observed: `zsh -c 'f(){ local status=1; true; return $status }; f; echo $?'` settles it.

The body still describes the `-v` fix throughout ("Both now ask with `-v`"), which is now the opposite of the diff.

**Impact:** `--dev` on a machine with no usable identity still builds ad-hoc, so Accessibility grants still die on every rebuild.
**Merge:** needs the cache reset first.

ref #1123

PathGao:
@PathGao Both are fixed, and the second one was worse than it read from the manual.

The cache: `LEGACY_IDENTITY_USABLE=""` now sits one line after that `fi`, inside the `--dev` block, so the repair invalidates the "no" the guard recorded a moment earlier. Your reading was exact — every later call site was reading the stale verdict and signing ad-hoc right after creating a working identity.

`local probe status=1` does not merely return the wrong value. zsh holds `status` read-only, so the declaration aborts the script at that line: `identity_can_sign` never decides anything, `setup-signing.sh` exits 1 with `read-only variable: status` before its front door, and `build.sh` prints the ad-hoc warning. Same destination as your prediction, one step earlier. I could not source the real script here, so I extracted the `identity_can_sign` body out of the file and ran it under `zsh` with `security` and `codesign` stubbed: the old spelling aborts at line 9 with exit 1; renamed to `signed`, the same body returns usable for a stub that signs and unusable for one that fails. Nothing touched a keychain.

That was the only `local status` in the tree — `git grep -E "local[[:space:]].*status"` over every `.sh` reaches only that one line, and `build.sh`'s two counters are `child_status` and `test_status`. `LEGACY_IDENTITY_USABLE` is likewise the only cached signing verdict; `ADAPTIVE_SKIP` is the other empty-string global and is about actool. `Tools/ci-setup-signing.sh` imports a certificate and never enumerates identities, so it has neither shape.

The shape check grew two assertions: no `local` named `status` in either script, and the cached verdict must be cleared between the repair line and the next read of `legacy_identity_installed`. Red-first on both — removing the reset and restoring `status` turned exactly those two red, restoring turned them green. `./build.sh --test` is 9478 of 9480; the two failures read this machine's Chinese input source and are identical on the branch's base `16216d56` (9467 of 9469, same two names), which I ran in a separate worktree rather than calling them pre-existing from memory.

The body is rewritten — it described the `-v` fix throughout, which the diff has not done since the previous push.

Still not run here: any `security` command, `Tools/setup-signing.sh`, `--dev`, or the full `./build.sh`. The signing keychain on this machine is locked, and each of those reaches `codesign` or `security` against it and blocks on a SecurityAgent dialog that never returns. Same position as last round, now including the full build.

PathGao:
The `status` finding is better than I had it: read-only aborts the script, not just returns the wrong value. Cache reset and rename both land where you say.

**`identityQueries` is the union, not per-script.** `!identityQueries.isEmpty` runs over `buildScriptLines + signingSetupLines`, so one surviving line in `build.sh` satisfies it — deleting the `find-identity` guard out of `identity_can_sign` stays green. Its own message says "both scripts"; the body says "requires each script to still ask the question at all". The per-script loop is three lines below, move the filter into it.

**The keychain password is a second copy now.** `LEGACY_KEYCHAIN_PASSWORD` in `build.sh`, `KCPASS` in `Tools/setup-signing.sh`, same literal, nothing pinning them together. Change one and the probe either fails the unlock (silent ad-hoc) or, on a stale path, skips it and raises the SecurityAgent dialog this change exists to avoid. The check pins that both files call `unlock-keychain -p`, not that they pass the same thing.

Also: the `-v`-breaks-releases line in `legacy_identity_installed` and its copy in the test say CI imports the same self-signed certificate. `ci-setup-signing.sh` imports the Developer ID secret and `release.yml` pins `subject.OU = "3D485NHW29"`, as your body says. Both comments claim the opposite.

Read from source, I ran nothing.

**Impact:** none today. Both gaps are drift the shape check advertises it catches and doesn't.
**Merge:** needs the per-script identity assertion first.

ref #1123

PathGao:
@PathGao All three land, and the union one was exactly as bad as you read it.

`!identityQueries.isEmpty` over `buildScriptLines + signingSetupLines` was satisfied by `build.sh` alone. I confirmed that rather than reasoning about it: with the old assertion in place I deleted the `find-identity` guard line out of `identity_can_sign` and ran `--test` — 2 of 9480 failed, the two input-source ones, nothing else. The filter now lives inside the per-script loop, and the same break fails `Tools/setup-signing.sh still looks the stable identity up by name`. The `-v` filter, `unlock-keychain -p`, `codesign --force --sign` and the `local status` check all moved under the same loop variable, so `readOnlyLocals` no longer re-selects the line array by comparing the script name to a string.

The password pinning reads both literals out of the two files and asserts they match. Changing `KCPASS` to `vorssaint-signing-2` fails with `found vorssaint-signing and vorssaint-signing-2`; restoring it goes green. I did not refactor the two scripts onto one variable — `setup-signing.sh` has to write the password literally anyway, since it is what creates the keychain, so sharing it would mean `build.sh` sourcing the setup script or a third file for one string. The check is the smaller fix, as you said.

On the release claim: you are right and both comments were wrong. `Tools/ci-setup-signing.sh` imports `SIGNING_CERT_P12` from a repository secret and never creates a certificate, and `release.yml` runs it with `REQUIRE_SIGNING: "1"` then verifies against `anchor apple generic and certificate leaf[subject.OU] = "3D485NHW29"` at four call sites — a self-signed certificate cannot satisfy `anchor apple generic`. So `-v` would never have reached a release. Both comments now say local builds only, which is what the body already said.

One place I checked and left alone: `CHANGELOG.md` line 2351, the 2.0.2 entry, says builds are signed with a self-signed identity, "shared certificate in CI". That is a shipped historical entry rather than a claim the code acts on, so rewriting it looked like widening the PR. Say the word if you want it corrected.

Verified: `./build.sh --test` is 9481 of 9483, the two failures being this machine's Chinese input source, identical on the base `16216d56` from the previous round's separate worktree. `zsh -n` parses both scripts. Still not run here: any `security` command, `Tools/setup-signing.sh`, `--dev`, or the full `./build.sh` — the locked signing keychain turns each into a SecurityAgent dialog that never returns. This round's diff is the test and two comments, so nothing in it changes behaviour.

PathGao:
The union fix lands, and the per-script loop is the right shape — `readOnlyLocals` no longer selecting its line array by comparing the script name to a string is better than what I asked for.

This round added two copies of one fact and pinned one of them. `LEGACY_KEYCHAIN` / `KC` is the other, and drift there fails worse than the password does:

```sh
&& { [[ ! -f "$LEGACY_KEYCHAIN" ]] \
    || security unlock-keychain -p "$LEGACY_KEYCHAIN_PASSWORD" ...
```

A path that no longer matches *skips* the unlock rather than failing it, so codesign meets a locked keychain and raises the SecurityAgent dialog this branch exists to remove. A wrong password costs a silent ad-hoc build; a wrong path costs the hang. `keychainPassword` is already the generic reader — one more pair, `("LEGACY_KEYCHAIN", "KC")`.

`LEGACY_IDENTITY` / `IDENTITY` is the third copy of that shape and predates the branch, so not this PR's to fix. `git grep` over the tree says those two lines are the only occurrences, and `Tools/ci-setup-signing.sh` builds its own path under `$TMP`, so nothing else drifts with them.

CHANGELOG 2351: leave it. A shipped entry no code reads.

Read from source, I ran nothing.

**Impact:** none today. The gap is exactly the drift the new check was added to catch.
**Merge:** needs `LEGACY_KEYCHAIN` pinned against `KC` first.

ref #1123

PathGao:
Both fixes land, and the per-script loop is the right shape.

**The unlock assertion is a union inside one file.** `code.contains("unlock-keychain -p")` over `Tools/setup-signing.sh` is satisfied by the creation-time unlock next to `create-keychain`, which is not the one that matters. Delete the unlock out of `identity_can_sign` and the suite stays green while the first signing build after a reboot goes back to the SecurityAgent dialog. Same shape as `identityQueries` last round, one level down — scope it to the probe body, not the file.

**`LEGACY_KEYCHAIN` is the copy you didn't pin.** This PR creates it — `main`'s build.sh has no keychain path — so `KC` now has a twin one line above the password you did pin. Drift it and `[[ ! -f "$LEGACY_KEYCHAIN" ]]` goes true, the unlock is skipped in silence, same dialog. Worse than password drift, which at least fails the unlock. Your closure already reads it: `keychainPassword("LEGACY_KEYCHAIN", buildScriptLines)` against `keychainPassword("KC", signingSetupLines)`.

CHANGELOG 2353: leave it. It records what 2.0.2 shipped, which was true then.

Read from source, I ran nothing.

**Impact:** none today. Both are drift the shape check advertises catching and doesn't.
**Merge:** needs the unlock assertion scoped to the probe first.

ref #1123

PathGao:
@PathGao Both land, and the unlock one was green under the exact break you described.

The unlock assertion is scoped to the probe now, and so are the other two presence checks, because they had the same shape one file away. `code.contains("unlock-keychain -p")` over `Tools/setup-signing.sh` was answered by the creation-time unlock next to `security create-keychain`. I deleted the unlock out of `identity_can_sign` and ran `--test` with the old assertion in place: 2 of 9483, the two input-source ones and nothing else. With the new one the same break fails `Tools/setup-signing.sh's identity_can_sign unlocks the dedicated signing keychain itself`. `probeBody` pulls the lines between `<name>() {` and the closing brace at column 0, and `find-identity`, `unlock-keychain -p` and `codesign --force --sign` all read that instead of the file. Stubbing out the probe's `codesign` line fails its own message, so narrowing did not blunt it. The two negatives stay file-wide: `-v` on this identity and `local status` are wrong wherever they are written, so width is the strength there, not the gap.

`LEGACY_KEYCHAIN` / `KC` goes through the reader you pointed at, as a second pair beside the password. Changing `KC` to `vorssaint-signing-2.keychain-db` fails `build.sh's LEGACY_KEYCHAIN and Tools/setup-signing.sh's KC are the same signing keychain path`; the old suite was green on that drift, since there was no path assertion at all.

Then I re-read every assertion this PR adds against the one question — what does it read, and could the thing it protects be deleted while it stays green — since this is the third round of the same defect. The three presence checks are the ones that failed it. Cleared, with why: the `-v` and `local status` filters forbid a spelling across the file; the two literal pins take the first `NAME="` assignment per script, and `LEGACY_KEYCHAIN_PASSWORD` cannot answer for `LEGACY_KEYCHAIN` nor `KCPASS` for `KC`, since both need the `="` immediately after the name; the cache-reset check slices `build.sh` between the `Tools/setup-signing.sh` invocation and the next read of `legacy_identity_installed`, so the `LEGACY_IDENTITY_USABLE=""` declaration at the top of the file sits outside the window and cannot satisfy it, and deleting the invocation trips the `else` arm rather than passing. Locating the probe by function name is the one helper name the check reads, and a rename fails `still defines identity_can_sign()` instead of passing on an empty body.

`LEGACY_IDENTITY` / `IDENTITY` I left alone, as you said — it predates the branch. CHANGELOG 2353 left as is.

Verified: `./build.sh --test` is 9484 of 9486, the two failures being this machine's Chinese input source, identical on the base `16216d56` from the earlier round's separate worktree. `zsh -n` parses both scripts before and after each drift, and each drift was reverted back to green. Still not run here: any `security` command, `Tools/setup-signing.sh`, `--dev`, or the full `./build.sh` — the locked signing keychain turns each into a SecurityAgent dialog on the machine's screen that never returns. This round's diff is the test only.

PathGao:
Probe-scoping is right, and keeping the two negatives file-wide — forbidding a spelling rather than requiring a line — is the correct split.

**The check pins the probe, not exclusivity.** Nothing bounds how many times the raw listing gets asked. Add a sixth signing path spelled the old way —

```sh
elif security find-identity -p codesigning 2>/dev/null | grep -q "$LEGACY_IDENTITY"; then
```

— and everything stays green: `identityQueries` is still non-empty inside the probe, and `trustQueries` only forbids `-v`. That is #1123 back, by the same copy-paste that put five of these in `build.sh` to begin with. Your own predicate closes it:

```swift
expect(lines.filter { $0.contains("find-identity") }
            .filter { $0.contains("$LEGACY_IDENTITY") || $0.contains("$IDENTITY") }
            .count == identityQueries.count,
       "\(script) asks find-identity for the stable identity only inside \(probeName)")
```

1 == 1 in both files today: `developer_id_identity`'s `-v` line carries no `$LEGACY_IDENTITY`.

Smaller: the cache-reset needle is a bare `contains("Tools/setup-signing.sh")`, while the comment above `runsSigningSetup` says the ad-hoc advice string names the script too. Reuse that regex instead of relying on the invocation coming first.

Read from source, I ran nothing.

**Impact:** none today. The gap is the regression the check exists to catch.
**Merge:** needs the exclusivity assertion first.

ref #1123

PathGao:
@PathGao The exclusivity gap is real and your line reproduces it exactly. I added the sixth call site verbatim — `elif security find-identity -p codesigning 2>/dev/null | grep -q "$LEGACY_IDENTITY"; then` in `sign_bundle` — and ran `--test`: 2 of 9489, the two input-source ones and nothing else. Every assertion in the set stayed green under it. Your predicate is in, counting `find-identity` lines naming the identity file-wide against the count inside the probe, and it fails with "found 1 such lookup(s) outside it". 1 == 1 in both scripts today.

The cache-reset needle now goes through the same regex `runsSigningSetup` uses, extracted into one closure. Worth confirming that it mattered: a bare `"Tools/setup-signing.sh"` matches three lines in `build.sh` — the invocation on 122, the failure warning on 123, and the ad-hoc advice on 601 — while the regex matches only 122.

Then the audit, both questions this time. Three more assertions failed the second one, so they are fixed here rather than next round:

- **The two literal pins read the first `NAME="` assignment.** A second assignment below it is what the shell obeys, and the check could not see it. Both values must now be written exactly once per script. Adding `LEGACY_KEYCHAIN="…vorssaint-signing-2.keychain-db"` under the first fails with both values listed; the previous version was green on it.
- **The cache-reset check used `firstIndex`.** A second repair site added later carried no reset requirement — the same "first match answers for all" shape as the union. It runs over every invocation now. A second `./Tools/setup-signing.sh` with no reset after it fails.
- **Requiring `codesign --force --sign` allowed an unconditional verdict beside it.** The call stays, the answer stops depending on it. Exactly one call and one verdict now, verdict no earlier than the call. `LEGACY_IDENTITY_USABLE=yes` added after `rm -rf "$PROBE_TMP"` fails with "found 1 call(s) and 2 verdict line(s)".

And the probe function must be defined once: a second `legacy_identity_installed() {` later in the file is what the shell runs while `probeBody` reads the first, so all five body-scoped checks would have been reading a function nothing calls. It fails with "found 2".

Every assertion, and which question each answers, is in a table in the body's Verification section. Two of them answer only the addition question — the `-v` and `local status` negatives — because a negative has nothing to delete, which is the same reason they stay file-wide. The probe's unlock is the one positive with no addition case: a second unlock inside the probe is idempotent, so absence is its only failure mode. I would rather name that than invent an assertion for it.

Verified: `./build.sh --test` is 9489 of 9491, the two failures this machine's Chinese input source, identical on the base `16216d56` from an earlier round's separate worktree. `zsh -n` parses both scripts before and after each of the five drifts, and each drift was reverted back to green. Still not run here: any `security` command, `Tools/setup-signing.sh`, `--dev`, or the full `./build.sh` — the locked signing keychain turns each into a SecurityAgent dialog on the machine's screen that never returns. This round's diff is the test only.

PathGao:
@PathGao Removed the last four rounds of shape checks in `9755ed9f`. The
behaviour fix and the assertions that guard it stay.

**What went.** `probeBody` and `shellLiterals` — the two closures that parse
zsh by string matching — and everything built on them: the probe extracted by
function name and required to be defined exactly once, `find-identity` counted
file-wide against the count inside the probe, one `codesign` call and one
verdict line with an index ordering between them, `LEGACY_KEYCHAIN`/`KC` and
`LEGACY_KEYCHAIN_PASSWORD`/`KCPASS` pinned across the two files and each
required to be assigned exactly once, and the cache-reset check widened from
the repair invocation to every invocation. 110 lines of test, no lines of
script.

**Why.** Every one of those rounds reported `Impact: none today`, and the body
said the same thing: they changed the test and two comments, no behaviour
changed. What they bought is drift that might happen later. What they cost is
a parser for a language the test does not run — `probeBody` slices between
`name() {` and a `}` at column 0, `shellLiterals` reads a `NAME="` prefix — and
a rename, a continuation line or an indented brace makes either return
nothing, at which point every assertion above it passes vacuously. That is the
failure mode those four rounds kept finding in the assertions they replaced,
one level up each time.

**What stays**, each re-checked by reverting its own shell fix and watching it
go red: no `-v` on this identity; `codesign --force --sign` asked in both
probes; `build.sh` unlocking the keychain itself; no zsh `local status`; the
cache cleared after the repair; the repair invoked at all. Six drifts, six
reds, each reverted back to green. `./build.sh --test` is 9477 of 9479 — the
two failures are this machine's Chinese input source, `nil layout falls back
to ANSI semicolon` and the App Switcher icon-row hint, identical on the base
`16216d56`.

Two assertions I dropped rather than kept, because I could not make them fail.
Deleting the unlock out of `identity_can_sign` leaves a file-wide read of
`Tools/setup-signing.sh` green — the creation-time unlock next to
`security create-keychain` answers it — so that one is gone with a comment
saying why, instead of asserting nothing. The identity lookup is kept but
described as what it is: the input set the `-v` ban reads, not a claim that
each script still asks.

**Where to look next, and it is not the test.** The new criterion —
`codesign` can sign **and** the dedicated keychain unlocks — has never run
against a real `security` or `codesign` on any machine. Everything behind it
is stubs and `zsh -n`, which the body has said since the round it was
introduced. And in `build.sh` a failed unlock is read as "no usable identity":

```sh
&& { [[ ! -f "$LEGACY_KEYCHAIN" ]] \
    || security unlock-keychain -p "$LEGACY_KEYCHAIN_PASSWORD" \
        "$LEGACY_KEYCHAIN" 2>/dev/null; }; then
```

On a non-`--dev` build there is no repair arm, so that silently signs ad-hoc,
which is #1123's failure reached through a keychain whose password drifted or
whose path moved. One run of `./build.sh --dev` on a Mac with a locked signing
keychain, then `codesign -dv --verbose=4 build/Vorssaint.app` to see which
identity it used, settles more than the four rounds did.

The body is rewritten: the removed assertions are out of it, the Verification
section lists the six drifts and the one that stays green, and the untested
half of the criterion is spelled out rather than left as a line at the bottom.

ref #1123

PathGao:
Removing them is right, and the ask was mine. An assertion built on a parser that can silently return nothing is worse than no assertion — that is the same defect those rounds kept finding, so four rounds of it was the wrong direction. The six that stay all have a shell fix behind them.

One gap sits in the half you named. All twelve other signing calls in `build.sh` pass `--strip-disallowed-xattrs`. The two probes are the only ones that don't:

```sh
cp /bin/echo "$PROBE_TMP/probe"
/usr/bin/codesign --force --sign "$LEGACY_IDENTITY" "$PROBE_TMP/probe"
```

build.sh's own header says why that flag exists: copies pick up xattrs (`com.apple.provenance` and friends) that invalidate codesign. If the probe copy carries one, the probe fails, and a failing probe reads as "no usable identity" — ad-hoc build on a machine whose identity signs fine. #1123 reached through the new criterion, on the path nobody has run.

Read from source. I could not run `codesign` or `xattr` here, so whether `cp /bin/echo` into `$TMPDIR` actually carries a disallowed xattr is unmeasured. The flag costs nothing either way.

**Impact:** if the probe copy carries a disallowed xattr, every local build signs ad-hoc.
**Merge:** needs `--strip-disallowed-xattrs` on both probes first.

ref #1123

PathGao:
@PathGao Both probes pass `--strip-disallowed-xattrs` in `c529c5b1`, in the same position as the other twelve calls. No comment on the probe lines: none of the other signing calls carry one, the header is the only place the flag is explained.

Measured rather than read from source: `cp /bin/echo "$TMPDIR/probe-xattr-test" && xattr -l "$TMPDIR/probe-xattr-test"` prints `com.apple.provenance` on the copy, and so does a copy into a `mktemp -d` directory under `$TMPDIR`. `/bin/echo` itself has none. So the probe copy does carry the xattr the header names. Whether `codesign` refuses it without the flag I could not run, same locked-keychain reason as before.

The existing assertion was `code.contains("codesign --force --sign")`, a literal, and the flag goes between `--force` and `--sign`, so adding it turned that needle red on its own. I extended that check instead of adding a new one. It now splits each line into tokens, takes every line with `codesign`, `--force` and `--sign` on it, requires at least one, and requires `--strip-disallowed-xattrs` among the tokens of each. Flag order is irrelevant to it. It is a file-wide read like the five that stayed, no function slicing.

Red-first, each with `./build/metrics-tests` after the `./build.sh --test` build, each restored to green afterwards: the flag dropped from `legacy_identity_installed` fails `build.sh passes --strip-disallowed-xattrs to every codesign --force --sign call, found 1 without it`; dropped from `identity_can_sign` fails the same message for `Tools/setup-signing.sh`; the `codesign` in `identity_can_sign` stubbed out fails `Tools/setup-signing.sh asks codesign whether the identity can sign before relying on it`.

One drift that used to be red is now green, and the body says so instead of hiding it. Stubbing the probe's `codesign` out of `legacy_identity_installed` leaves the suite at 2 of 9481: the bundle and helper calls at lines 189, 572 and 574 now read the same as the probe, so they answer the file-wide read for it. Before this round the probe was the only `codesign --force --sign` literal in `build.sh`, so that assertion was probe-specific by accident of flag order. Making it probe-specific again would mean matching `$PROBE_TMP` or the function name, which is the parser the last round removed, so I did not.

`./build.sh --test`: 9479 of 9481, the two failures this machine's Chinese input source as before. `zsh -n build.sh` and `zsh -n Tools/setup-signing.sh` parse. Body updated: the probe paragraph, the drift table, the trade-off line about what the guard matches, the counts, and the not-run list now name the flag.

Still not run here: `codesign`, any `security` command, `Tools/setup-signing.sh`, `--dev`, the full `./build.sh`. The probe signing with and without the flag is unmeasured.

ref #1123

PathGao:
The xattr measurement is the right kind of evidence, and naming the drift that went green instead of burying it is the right call.

One gap in the new check, not a blocker. It needs `codesign`, `--force` and `--sign` on one physical line. Seven of build.sh's thirteen signing calls wrap `--sign` onto the continuation — every one carrying `--identifier` or `--options runtime`. The probe's line already wraps; add `--identifier` to it and it drops out of `signingCalls` entirely, while five other lines keep both assertions green. The body says a rename or reindent "cannot make it read nothing" — a wrap can. Joining trailing `\` before the filter is five lines and can only under-join, so it isn't the parser you removed.

The thing worth doing is still the one run. `security unlock-keychain -p vorssaint-signing ~/Library/Keychains/vorssaint-signing.keychain-db` should not raise a dialog: `-p` is the non-interactive form, and listing identities never needs the private key. If it returns, the probe won't prompt either and `--dev` is runnable. If it blocks anyway, say so and I'll drop this to ship it.

**Impact:** the unlock condition in `legacy_identity_installed` has never executed; on a non-`--dev` build a wrong read there is a silent ad-hoc build.
**Merge:** needs one `./build.sh --dev` run first.

ref #1123

PathGao:
@PathGao Ran the new criterion on a real machine, this head (c529c5b1). On this Mac `security find-identity -v -p codesigning` reports 0 valid identities, which is the case #1123 describes, and `./build.sh --dev` still went through the probe and signed with the legacy identity: build log says `signing with legacy self-signed identity: Vorssaint Utils Signing`, and `codesign -dv` on `build/stage/Vorssaint (Developer).app` shows `Authority=Vorssaint Utils Signing`; `codesign --verify --deep --strict` passes. Also measured your xattr point: a `cp /bin/echo` into $TMPDIR here carries `com.apple.provenance`, so the `--strip-disallowed-xattrs` on both probes matters on this machine, not only for consistency.

PathGao:
That closes the half nobody had run. `--dev` on a machine reporting 0 valid identities is exactly #1123's shape, and it signed with the identity, so the probe and the unlock both executed against real `security` and `codesign` rather than stubs. The xattr measurement settles the flag too.

One thing the run doesn't distinguish, and it doesn't need to hold the merge: whether the keychain was locked going in. If it was already open, what's proven is that `unlock-keychain -p` returns non-interactively, not the post-reboot case the branch exists for. That belongs in the body's not-run list rather than in another round.

The line-wrap gap in `signingCalls` stays as described. Non-blocking, and I said last round I'd drop it to ship.

**Impact:** none open. The one unmeasured branch is a failed unlock on a non-`--dev` build, which the body already names.
**Merge:** ship it.

ref #1123

PathGao:

PathGao:
Reopened as #1300.

PR #1233, opened 2026-09-01
fix(window-layout): hand the window taps back across a fast user switch

PathGao:
## Summary

Window Layout owns three `.defaultTap` event taps and the green-button
maximizer owns a fourth. None of them followed the login session:

| tap | mask | created |
| --- | --- | --- |
| move/resize gesture | left+right down/dragged/up | while the feature is on |
| edge snap | left down/dragged/up | while the feature is on |
| directional gesture | scroll, left/right down, keyDown | only while the hotkey is held |
| maximizer | left down/up | while the feature is on |

All four run on the main run loop, so the re-arm reads
`SessionActivity.shared.isActive` in place, the way
`MouseNavigation/MouseNavigationService.swift` does. The taps that run a
thread of their own — Finder, keyboard, Super Key, snippets, switcher — hop to
the main queue for the same answer, because that is where the flag is written.

the screen, and a filter tap created there keeps its place in the chain. The
window server still routes every matching event through a process that cannot
answer and waits out the tap timeout on each one, so the account actually in
use pays it (#1153). Each of the four now asks
`SessionActivitySupport.tapShouldRun` in both places that can put a tap into
the chain — the preference sync and the timeout re-arm — and subscribes to
`SessionActivity.shared.onChange` so the taps come back with the session.

The maximizer's two permission sources disagreed inside one path. `handle`
refuses to re-arm on `AXIsProcessTrusted() == false` and then, from the same
async block, calls `syncWithPreferences()` — which asked the timer-polled
`Permissions.shared.accessibility` and could `start()` the tap straight back
on a value that had not caught up with the revoke. `syncWithPreferences()` now
asks `AXIsProcessTrusted()` directly, the way
`WindowLayoutService.syncWithPreferences` already does. Nothing else changes:
the only other caller path is the `Permissions.shared.$accessibility` sink in
`AppDelegate`, which fires after the grant, so the direct read is the fresh
one there too.

Two smaller fixes in the same code:

- The directional gesture cancelled itself on Escape from inside its own tap
  callback, removing the run loop source and invalidating the port on that
  stack. The repository states three times that this is unsafe
  (`MouseButtonShortcutService.swift`, `CleaningModeManager.swift`,
  `MiddleClickService.swift`), and the edge-snap tap in this same file already
  defers to the main queue. Escape now does the same, as do the three new
  teardown paths, so this change does not reintroduce the shape it removes.
- The directional tap had no `tapDisabledByTimeout` handling at all. A tap the
  window server disabled stayed disabled and the gesture went quiet until the
  hotkey was released.

Trade-offs:

- The directional *hotkey* is gated too, not just its tap. Gating only the tap
  would leave `beginDirectionalGesture()` free to create one while the session
  is off screen; taking the hotkey away removes the only path to it, in one
  line, at the same place the other three gates live.
- The re-arm branches call `syncWithPreferences()` after teardown, following
  `MouseNavigationService`, so a tap released by a timeout that raced the
  resign notification is rebuilt if the session was in fact still on screen.
- No new guard file was written. `MetricsTests.swift` already carries the
  session-switch guard as a loop over tap owners; both files were added to it
  and two of its assertions were changed, because as written they did not
  reach the code this pull request adds:
  - the re-arm slice split on the *first* `tapDisabledByTimeout` in the file,
    which in `WindowLayoutService.swift` is `observeDirectionalEvent`, so the
    edge-snap and gesture re-arms went unchecked. It now checks every slice.
  - the `AXIsProcessTrusted()` assertion sat behind an `if tapOwner.contains(…)`
    naming four other services, so it skipped both new files. The gate now
    admits them, and a second check rejects the polled mirror at the gate.

    That second check reads the gate's own argument — everything between
    `SessionActivitySupport.tapShouldRun(` and its `sessionIsActive:` label,
    for every call in the file — rather than banning
    `Permissions.shared.accessibility` across the source. A whole-file ban
    holds for the eight files behind this gate today and breaks on the taps
    that come next: `FinderCutPaste` and `AppSwitcher` both cache that mirror
    on purpose on their per-keystroke path, each with a comment saying a live
    TCC round-trip per key is what the cache exists to avoid. The gate's
    argument is where a stale answer actually does harm, and it is the only
    place the check now looks.

    It is asked of the files whose re-arm refuses on `AXIsProcessTrusted()`,
    which is where the two answers can disagree, rather than off a second
    hand-kept name list. `ScrollInverter.swift` re-arms on the session alone
    and never consults Accessibility, so there is no pair of sources to
    disagree there; it passes `accessibilityGranted:
    Permissions.shared.accessibility` legitimately and stays exempt.

  PR #1230 replaces the owner list with a sweep of every file containing
  `CGEvent.tapCreate`; this change is what lets both files drop out of its
  `knownUngatedTaps` table.

Two riders that were in this branch are now on their own:

- the directional gesture's 60 Hz feedback timer moving to common run loop
  modes — an independent UI defect, visible without any second account, that
  none of the assertions here reach — is #1267.
- `focusedTarget(for:)` looking its app up by pid instead of re-bridging
  `NSWorkspace.runningApplications` — a performance change with no measurement
  behind it and one behaviour difference on a terminated pid — is #1268.

A third finding from the same audit was left out entirely: caching
`WindowEdgeSnapSupport.isSystemTilingEnabled`, which rebuilds a `UserDefaults`
suite and reads three keys on every left mouse down. PR #1208 is rewriting that
exact computed property (and makes it read a second suite), so the cache should
be rebased onto it rather than conflict with it.

## Verification

MacBook Pro, macOS 26. Rebased onto `main` at `a952b829`.

```
./build.sh --test      TESTS OK (9550 checks)
```

`--test` compiles a whitelist that does not include either changed source
file — that list is almost entirely `*Support.swift` and `*Strings.swift` — so
the whole of `Sources/Vorssaint/**` was type-checked separately with the same
target and SDK `build.sh` uses. The full `./build.sh` was not run: with the
signing keychain locked, `codesign` blocks on an unlock dialog that never
returns.

The two input-source failures reported on the earlier revision
(`nil layout falls back to ANSI semicolon` and `App Switcher icon-row hints`)
did not appear this round; the machine's input source is ANSI now rather than
a Chinese IME. They read the current input source and are unrelated to this
diff either way.

The changed assertions were checked by breaking the code they are meant to
hold, not by argument. Each break was applied alone and reverted:

- `sessionIsActive: SessionActivity.shared.isActive` replaced with `true` in
  the **edge-snap** re-arm — the second of the three, the one the old
  first-slice split could not see. Fails `WindowLayoutService.swift does not
  re-arm any disabled tap into a switched-away session`.
- the maximizer's `syncWithPreferences()` put back on the polled mirror. Fails
  `WindowMaximizer.swift asks Accessibility directly wherever it decides to run
  a tap`. Restored: green.
- the gate-argument form was also checked against the case that made the
  whole-file ban wrong: `ScrollInverter.swift` reads the mirror in its gate
  today and is caught by this check when it is not exempt, and the
  per-keystroke caches in `FinderCutPaste` and `AppSwitcher` are not caught by
  it at all.

**What was not tested.**

- **The behaviour itself.** There is no second login session on this machine,
  so the fast user switch this change is about was never performed. What is
  verified is that the wiring matches the services that already carry it.
- The revoke race is not reproduced either. Reaching it needs the window server
  to disable the maximizer's tap inside the polling interval that follows a
  revoked Accessibility grant; the fix is the read the sibling service already
  makes at the same point.
- The bundle assembly and signing step of `./build.sh`, for the reason above.
- The Escape teardown is a divergence from a rule this repository states three
  times, not an observed crash. Nothing was seen to fail before the change.
- The directional tap's new timeout handling was not exercised: reaching it
  needs the window server to disable a tap that only exists while a hotkey is
  physically held.

## Anything else

Refs #1153

### DIFF
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
@@ -310,11 +310,7 @@ final class WindowLayoutService: ObservableObject {
         guard let onScreenWindowIDs = onScreenWindowIDs() else { return nil }
         for pid in pids {
             let isFocusedOwnApp = pid == ownPID && hasFocusedResizableOwnWindow
-            // One lookup by pid, not a fresh bridge of every running app on
-            // each turn of a list that can hold dozens of them. The edge-snap
-            // drag already resolves its app this way.
-            guard let app = NSRunningApplication(processIdentifier: pid),
-                  !app.isTerminated,
+            guard let app = NSWorkspace.shared.runningApplications.first(where: { $0.processIdentifier == pid }),
                   isFocusedOwnApp
                     || (app.activationPolicy == .regular && !app.isHidden
                         && app.bundleIdentifier != ownBundleID)
@@ -759,16 +755,9 @@ final class WindowLayoutService: ObservableObject {
                                                       action: nil,
                                                       manualOverride: nil)
         showDirectionalIndicator(at: NSEvent.mouseLocation, action: nil)
-        // The gesture's own tap takes leftMouseDown and rightMouseDown, so a
-        // held mouse button puts the main run loop in event tracking and a
-        // default-mode timer stops firing: the indicator and the preview would
-        // freeze on the frame before the press. Common modes keep the 60 Hz
-        // feedback running for as long as the hotkey is held.
-        let timer = Timer(timeInterval: 1.0 / 60.0, repeats: true) {
+        directionalTimer = Timer.scheduledTimer(withTimeInterval: 1.0 / 60.0, repeats: true) {
             [weak self] _ in self?.updateDirectionalGesture()
         }
-        directionalTimer = timer
-        RunLoop.main.add(timer, forMode: .common)
         startDirectionalTap()
     }
 
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
@@ -107,7 +107,7 @@ final class WindowLayoutService: ObservableObject {
         let wantsDirectional = SessionActivitySupport.tapShouldRun(
             featureWanted: available
                 && UserDefaults.standard.bool(forKey: DefaultsKey.windowDirectionalEnabled),
-            accessibilityGranted: trusted,
+            accessibilityGranted: AXIsProcessTrusted(),
             sessionIsActive: sessionIsActive
         )
         wantsDirectional ? registerDirectionalHotkey() : unregisterDirectionalHotkey()
@@ -115,7 +115,7 @@ final class WindowLayoutService: ObservableObject {
         let wantsGesture = SessionActivitySupport.tapShouldRun(
             featureWanted: available
                 && UserDefaults.standard.bool(forKey: DefaultsKey.windowGestureEnabled),
-            accessibilityGranted: trusted,
+            accessibilityGranted: AXIsProcessTrusted(),
             sessionIsActive: sessionIsActive
         )
         wantsGesture ? startGestureTap() : stopGestureTap()
@@ -124,7 +124,7 @@ final class WindowLayoutService: ObservableObject {
             featureWanted: available
                 && UserDefaults.standard.bool(forKey: DefaultsKey.windowEdgeSnapEnabled)
                 && !WindowEdgeSnapSupport.isSystemTilingEnabled,
-            accessibilityGranted: trusted,
+            accessibilityGranted: AXIsProcessTrusted(),
             sessionIsActive: sessionIsActive
         )
         wantsEdgeSnap ? startEdgeSnapTap() : stopEdgeSnapTap()
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
@@ -310,7 +310,11 @@ final class WindowLayoutService: ObservableObject {
         guard let onScreenWindowIDs = onScreenWindowIDs() else { return nil }
         for pid in pids {
             let isFocusedOwnApp = pid == ownPID && hasFocusedResizableOwnWindow
-            guard let app = NSWorkspace.shared.runningApplications.first(where: { $0.processIdentifier == pid }),
+            // One lookup by pid, not a fresh bridge of every running app on
+            // each turn of a list that can hold dozens of them. The edge-snap
+            // drag already resolves its app this way.
+            guard let app = NSRunningApplication(processIdentifier: pid),
+                  !app.isTerminated,
                   isFocusedOwnApp
                     || (app.activationPolicy == .regular && !app.isHidden
                         && app.bundleIdentifier != ownBundleID)
@@ -755,9 +759,16 @@ final class WindowLayoutService: ObservableObject {
                                                       action: nil,
                                                       manualOverride: nil)
         showDirectionalIndicator(at: NSEvent.mouseLocation, action: nil)
-        directionalTimer = Timer.scheduledTimer(withTimeInterval: 1.0 / 60.0, repeats: true) {
+        // The gesture's own tap takes leftMouseDown and rightMouseDown, so a
+        // held mouse button puts the main run loop in event tracking and a
+        // default-mode timer stops firing: the indicator and the preview would
+        // freeze on the frame before the press. Common modes keep the 60 Hz
+        // feedback running for as long as the hotkey is held.
+        let timer = Timer(timeInterval: 1.0 / 60.0, repeats: true) {
             [weak self] _ in self?.updateDirectionalGesture()
         }
+        directionalTimer = timer
+        RunLoop.main.add(timer, forMode: .common)
         startDirectionalTap()
     }
 
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
@@ -80,30 +80,53 @@ final class WindowLayoutService: ObservableObject {
     private let resizeGestureUpdateInterval: TimeInterval = 1.0 / 60.0
     private let edgeSnapSampleInterval: TimeInterval = 1.0 / 30.0
 
-    private init() {}
+    private init() {
+        // A filter tap owned by a switched-away login session keeps its place
+        // in the chain and stalls input for the account on screen. Hand every
+        // tap back on resign and build them again from the preferences on the
+        // way back in.
+        SessionActivity.shared.onChange { [weak self] _ in
+            self?.syncWithPreferences()
+        }
+    }
 
     func syncWithPreferences() {
         let available = AppFeature.windowLayout.isAvailable
         let trusted = AXIsProcessTrusted()
+        let sessionIsActive = SessionActivity.shared.isActive
+        // Carbon hotkeys are registered per session, so the plain layout
+        // shortcuts need no session gate: they own no tap.
         let wantsShortcuts = available
             && UserDefaults.standard.bool(forKey: DefaultsKey.windowLayoutShortcutsEnabled)
             && trusted
         wantsShortcuts ? registerHotkeys() : unregisterHotkeys()
 
-        let wantsDirectional = available
-            && UserDefaults.standard.bool(forKey: DefaultsKey.windowDirectionalEnabled)
-            && trusted
+        // The directional hotkey is gated all the same, because holding it is
+        // what creates the third tap. Dropping the hotkey is how that tap stays
+        // out of a session it cannot serve.
+        let wantsDirectional = SessionActivitySupport.tapShouldRun(
+            featureWanted: available
+                && UserDefaults.standard.bool(forKey: DefaultsKey.windowDirectionalEnabled),
+            accessibilityGranted: trusted,
+            sessionIsActive: sessionIsActive
+        )
         wantsDirectional ? registerDirectionalHotkey() : unregisterDirectionalHotkey()
 
-        let wantsGesture = available
-            && UserDefaults.standard.bool(forKey: DefaultsKey.windowGestureEnabled)
-            && trusted
+        let wantsGesture = SessionActivitySupport.tapShouldRun(
+            featureWanted: available
+                && UserDefaults.standard.bool(forKey: DefaultsKey.windowGestureEnabled),
+            accessibilityGranted: trusted,
+            sessionIsActive: sessionIsActive
+        )
         wantsGesture ? startGestureTap() : stopGestureTap()
 
-        let wantsEdgeSnap = available
-            && UserDefaults.standard.bool(forKey: DefaultsKey.windowEdgeSnapEnabled)
-            && !WindowEdgeSnapSupport.isSystemTilingEnabled
-            && trusted
+        let wantsEdgeSnap = SessionActivitySupport.tapShouldRun(
+            featureWanted: available
+                && UserDefaults.standard.bool(forKey: DefaultsKey.windowEdgeSnapEnabled)
+                && !WindowEdgeSnapSupport.isSystemTilingEnabled,
+            accessibilityGranted: trusted,
+            sessionIsActive: sessionIsActive
+        )
         wantsEdgeSnap ? startEdgeSnapTap() : stopEdgeSnapTap()
     }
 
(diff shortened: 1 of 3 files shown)


PathGao:
Gating the directional *hotkey* instead of its tap is the right call — `beginDirectionalGesture()` is only reachable from the Carbon handler, so one line really does close the path. And all three `WindowLayoutService` teardowns already invalidated; only the maximizer didn't. Both checked against `main`.

**The guard covers less than the body claims.** The re-arm slice in `MetricsTests.swift` splits on the *first* `tapDisabledByTimeout`, which in this file is `observeDirectionalEvent` — the edge-snap and gesture re-arms go unchecked. And the `AXIsProcessTrusted()` assertion sits behind an `if tapOwner.contains(...)` naming four other services, so it skips both new files; your own break-test count (2+3=5) confirms it never fired. #1230 is about to drop both out of `knownUngatedTaps` on the strength of this.

**The maximizer's two permission sources now disagree inside one path.** `handle` refuses to re-arm on `AXIsProcessTrusted() == false`, then the same async block calls `syncWithPreferences()`, which asks the timer-polled `Permissions.shared.accessibility` and can `start()` the tap straight back. `WindowLayoutService.syncWithPreferences` reads `AXIsProcessTrusted()` directly and has no such gap.

No second login session here either — this is read against `main` plus the diff, nothing run.

**Impact:** the maximizer tap can rebuild itself right after an AX-revoked timeout; the session-switch fix itself holds.
**Merge:** needs the re-arm assertion to reach all three `WindowLayoutService` sites (or the coverage claim in the body corrected) first.

ref #1153

PathGao:

PathGao:
@PathGao Both halves of that are right, and the second one is a real defect rather than a wording problem, so I closed it in the code instead of softening the body.

`WindowMaximizer.syncWithPreferences()` now reads `AXIsProcessTrusted()` instead of `Permissions.shared.accessibility`, which is what `WindowLayoutService.syncWithPreferences` already does. That removes the disagreement inside the timeout path: `handle` refuses to re-arm on the live answer, and the `syncWithPreferences()` that follows it in the same async block no longer starts the tap back on the polled one. The only other caller is the `Permissions.shared.$accessibility` sink in `AppDelegate`, which fires after the grant, so the direct read is the fresh value there too — no relaunch regression.

The guard was widened in the two places you named, both inside the existing loop in `MetricsTests.swift`:

- the re-arm slice took `components(separatedBy: "tapDisabledByTimeout").dropFirst().first`, so only the first tap in a file was checked. It now maps every slice and uses `allSatisfy`, which reaches all three `WindowLayoutService` sites.
- the `AXIsProcessTrusted()` gate now admits both new files, and a second assertion in it rejects `accessibilityGranted: Permissions.shared.accessibility`. The file-level `contains("AXIsProcessTrusted()")` on its own would have passed on the maximizer even with the defect in place, since `handle` already calls it.

Where else the shape appears: `WindowLayoutService.swift` is the only tap owner in `Sources/` with more than one `tapDisabledByTimeout` besides `BrightnessService.swift` (2, not in the list yet), so the widened split is a no-op for the six services already there and for the single-tap files the sibling session-switch PRs are adding — it just holds for them from now on. `ScrollInverter.syncWithPreferences` does read the polled `Permissions.shared.accessibility`, but its timeout branch re-arms on the session alone and never asks about Accessibility, so it has no second source to disagree with; it stays out of the AX gate and out of this diff. I left it alone rather than widen this pull request.

Verified by breaking each one and watching it go red. Replacing `sessionIsActive: SessionActivity.shared.isActive` with `true` in the **edge-snap** re-arm — the second of the three, the one the old first-slice split could not see — fails `WindowLayoutService.swift does not re-arm any disabled tap into a switched-away session`. Putting `syncWithPreferences()` back on `Permissions.shared.accessibility` fails `WindowMaximizer.swift asks Accessibility directly everywhere it decides to run a tap`. Both breaks together: 4 failures (the 2 new ones plus the 2 pre-existing input-source ones). Restored: 2. `./build.sh --test` is 9483 of 9485, and I ran the full `./build.sh` as well since neither source file is in the compile whitelist. The release compile of the whole target finished with zero warnings and zero errors, so `WindowMaximizer.swift` did compile; the run then failed in the bundle step signing the fan-control helper with `errSecInternalComponent`, a keychain failure on this machine rather than anything in the diff. I did not re-run it, so no bundle was produced here.

Not verified: the behaviour. Still no second login session here, and the revoke race needs the window server to disable the maximizer's tap inside the polling interval that follows a revoked grant. What is checked is that both permission reads in that path now come from the same place, and that the guard fails when either is put back.

PathGao:
The three re-arm slices really are all covered now — I walked the split against `main` and each of the three `WindowLayoutService` sites lands before its first `return`. Read, not run.

**The second assertion cannot fail on the defect it was written for.** It rejects `accessibilityGranted: Permissions.shared.accessibility`, but the line on `main` is

```swift
if wanted, Permissions.shared.accessibility {
```

no `accessibilityGranted:` prefix, so that string was never in the file. The break-test reintroduced the *guarded* spelling, not the original one, which is why it went red. Put the real line back and the suite stays green. Across all of `Sources/`, that exact string appears only in `ScrollInverter.swift` and `WindowEnumerator.swift`, neither of which is behind the gate — so as written the assertion is true for every file it runs on.

Drop the prefix:

```swift
expect(!code.contains("Permissions.shared.accessibility"),
```

Free, because of the six files behind that `if`, the bare symbol appears in exactly one on `main` — the maximizer line this PR deletes. Comment lines are already stripped by the `hasPrefix("//")` filter, so the new `//` reference in `WindowMaximizer.swift` doesn't trip it.

**Impact:** none at runtime; the guard is a no-op against the regression it names.
**Merge:** needs the assertion to drop the `accessibilityGranted: ` prefix first.

ref #1153

PathGao:
@PathGao You're right and the reading is exact: the assertion matched a string that was never in the file. `main` has `if wanted, Permissions.shared.accessibility {`, no label, so the break-test only went red because it reintroduced the *guarded* spelling rather than the original one. As written the check was true everywhere it ran.

Fixed, but by the symbol rather than by dropping one prefix off one spelling: `expect(!code.contains("Permissions.shared.accessibility"), …)`. What is wrong on this path is that the polled mirror is reached from it at all — a different argument label or a local holding the read is the same defect, and the prefix-free match is no longer tied to how the call site happens to be written.

Confirmed it does not fire on legitimate reads. Of the six files behind that gate — `MouseNavigationService`, `MouseButtonShortcutService`, `MiddleClickService`, `QuitProtectionService`, `WindowLayoutService`, `WindowMaximizer` — none contains `Permissions.shared.accessibility` on `main` except the maximizer line this branch deletes, and none contains it now. The only occurrence left in any of them is the comment in `WindowMaximizer.swift` explaining the choice, and the `hasPrefix("//")` filter strips it. `ScrollInverter.swift` does pass `accessibilityGranted: Permissions.shared.accessibility` legitimately, for the reason we settled earlier — it is in the loop but outside this gate, so it is untouched. No extra scoping was needed.

Red-first in each spelling I now claim to catch, one at a time, each reverted: the labelled argument; `main`'s own bare `if wanted, Permissions.shared.accessibility {`; and a local, `let granted = Permissions.shared.accessibility` passed as `accessibilityGranted: granted`. All three fail `WindowMaximizer.swift asks Accessibility directly everywhere it decides to run a tap` — 3 of 9485 each, restored 2 of 9485.

`./build.sh --test` is 9483 of 9485. The two are `nil layout falls back to ANSI semicolon` and `App Switcher icon-row hints`; both read the machine's live input source, a Chinese IME here. I ran `--test` on this branch's merge base with `main` (`16216d56`) in a throwaway worktree and it fails the same two and nothing else. I did not run the full `./build.sh` this round — the signing keychain is locked here and `codesign` blocks on an unlock dialog — so neither changed source file was compiled locally; CI covers both. Behaviour is still unverified for the same two reasons as before: no second login session, and the revoke race needs the window server to disable the tap inside the polling interval after a revoked grant.

PathGao:
The symbol-level match is the right fix, and it holds against `main`: of the six files behind that gate only `WindowMaximizer.swift:33` reads the mirror, and this branch deletes that line — `WindowLayoutService.swift` has zero. The new `//` comment is stripped by the `hasPrefix("//")` filter, and `Permissions` declares exactly one `accessibility` member, so the prefix-free match has nothing to false-positive on. Walked `src/`, not run.

One correction that doesn't move your conclusion: `WindowMaximizer.shared.syncWithPreferences()` has more callers than the AppDelegate sink — `FeatureRuntime.swift:162` and `MenuPanelView.swift:1509`. All of them are safe for the same reason, since `AXIsProcessTrusted()` is never staler than the mirror.

Not a blocker, and outside this diff: `ShelfService.fadeOutAndHide` is the only other `1.0 / 60.0` `Timer.scheduledTimer` in `Sources/` — the same default-mode shape you just fixed here. Whether its fade can overlap event tracking I did not check.

**Impact:** none outstanding; the guard now fails on the defect it names.
**Merge:** ship it.

ref #1153

PathGao:
@PathGao Rebased onto `main` at `a952b829` and split, head is now `db4506f2`.

**The rebase.** `#1225` landed `CFMachPortInvalidate` in both files, so the
duplicate hunk this branch carried is gone. The rest of the diff applied
unchanged: `WindowLayoutService` still has 3 invalidations for 3 taps and
`WindowMaximizer` 1 for 1, so the repo-wide sweep in `MetricsTests` counts
clean without either of this branch's lines.

**Two riders removed** (`503b254a`). The `### Also in this file` section is
gone from the body along with them:

- the 60 Hz feedback timer moving to common run loop modes is #1267. It is an
  independent UI defect — the indicator freezes while a mouse button is held,
  visible on one account with no switching involved — and none of this
  branch's assertions reached it. It went out with an assertion of its own,
  sliced to the gesture's start so the settle timer that has been in common
  modes all along cannot answer for it.
- `focusedTarget(for:)` looking its app up by pid is #1268. It is a
  performance change with no measurement, and it has a behaviour difference
  the old form did not: `runningApplications` never lists a terminated
  process, `NSRunningApplication(processIdentifier:)` does, which is why it
  needed `!app.isTerminated`. That is worth weighing on its own rather than on
  the strength of a session-gate review.

**The assertion changed shape** (`73c5699c`). The symbol-level match you
signed off at 08:30 holds for the eight files behind that gate, and it breaks
on the taps arriving behind this one. `FinderCutPaste.handle` and
`AppSwitcher.handle` both read `Permissions.shared.accessibility` on purpose
on their per-keystroke path, each with a comment saying a live TCC round-trip
per key is what the cache exists to avoid. A whole-file ban reads those as the
defect.

It now reads the gate's own argument — everything between
`SessionActivitySupport.tapShouldRun(` and its `sessionIsActive:` label, for
every call in the file. That is the only place the stale answer does harm, and
it catches a different argument label or a local reading the mirror the same
way the symbol match did. It is asked of the files whose re-arm refuses on
`AXIsProcessTrusted()`, which is where the two sources can disagree, rather
than off a second hand-kept name list; `ScrollInverter` re-arms on the session
alone and stays exempt, which is the same conclusion you reached at 08:30 by
naming it.

Verified both ways: putting `Permissions.shared.accessibility` back into
`WindowMaximizer`'s gate turns it red on its own; restored, green. Also
checked that it does *not* fire on the two per-keystroke caches, which is the
case the previous form got wrong.

**Tests.** `./build.sh --test` → `TESTS OK (9550 checks)`. The two
input-source failures I reported earlier (`nil layout falls back to ANSI
semicolon`, `App Switcher icon-row hints`) did not appear — this machine's
input source is ANSI now rather than the Chinese IME. `swiftc -typecheck` over
all of `Sources/Vorssaint/**` is clean; neither changed file is in the
`--test` whitelist. The full `./build.sh` was not run: the signing keychain is
locked and `codesign` blocks on the unlock dialog.

**Where to look first.** The gate-argument slice. It assumes every
`tapShouldRun(` call writes its three labels in order, which Swift enforces,
but it is a text slice and it is the only thing now standing between a gate
and the polled mirror across eight files. If you would rather have the
name-list form back for the six files that can take it, say so and I will
split it that way.

PathGao:
The rebase and the split are right, and the argument slice beats the name list — don't go back to it.

It can't see `WindowLayoutService`'s gates, though. All three in `syncWithPreferences` pass `accessibilityGranted: trusted`, the local hoisted at the top of that function; only the re-arms call `AXIsProcessTrusted()` inline. The slice reads that argument text, finds no mirror, and passes whatever `trusted` came from. Put `let trusted = Permissions.shared.accessibility` on that line and three of the four taps start on the polled value, green. The symbol form you replaced did catch that spelling — you broke it that way yourself.

Assert the live call instead of banning the mirror:

```swift
expect(gateArguments.allSatisfy { $0.contains("AXIsProcessTrusted()") },
```

Every gate in the eight files already reads that way except those three, so the cost is inlining the call there in place of `trusted` (`wantsShortcuts` still uses it). `ScrollInverter` stays exempt on the same file-level condition and the per-keystroke caches are still outside the slice — both cases that made you change shape survive. Read against `main`, not run.

**Impact:** none at runtime; the guard is blind to the mirror in the file holding three of the four taps.
**Merge:** needs the gate check to assert `AXIsProcessTrusted()` first.

ref #1153

PathGao:
@PathGao The three syncWithPreferences gates now call AXIsProcessTrusted() inline (638a73be), and the slice asserts the live call instead of banning the mirror. Red first: with the assertion flipped and the source untouched, TESTS FAILED (1 of 9550) naming WindowLayoutService alone; after the source change, full ./build.sh exit 0, TESTS OK (9550 checks). The local trusted stays for wantsShortcuts, which owns no tap.

PathGao:
Done, and it holds. I walked every `tapShouldRun(` call site on `main`: the six files already in the loop all pass `accessibilityGranted: AXIsProcessTrusted()` inline, so the new assertion is a no-op for them and a live check for your two. `ScrollInverter` is the only one with no `AXIsProcessTrusted()` anywhere in the file, so the file-level `if` is what exempts it — that carve-out is doing the work, not decoration. `trusted` is left with one reader, `wantsShortcuts`, which registers Carbon hotkeys and owns no tap. Read against `main`, not run.

One non-blocker, one word: `gateArguments.allSatisfy` is true on an empty array. The re-arm check three lines up guards with `!rearms.isEmpty` for that reason. All eight files have at least one `tapShouldRun(` today so it can't fire now, but a file whose gate gets inlined later would go quiet instead of red.

```swift
expect(!gateArguments.isEmpty && gateArguments.allSatisfy { $0.contains("AXIsProcessTrusted()") },
```

**Impact:** none.
**Merge:** ship it.

ref #1153

PathGao:

PathGao:
Reopened as #1304.

PR #1234, opened 2026-09-01
fix(finder): hand the Finder taps back across a user switch

PathGao:
# Hand both Finder taps back on a user switch

## Summary

`FinderCutPaste` and `FinderRenameService` are near-identical twins: same threaded
`runEventTap()`, same `CFRunLoopRun()` shape, both long-lived `.defaultTap` keyboard
taps. Two things were wrong across the pair.

**Neither observed fast user switching.** Both gated only on the feature switch and
Accessibility, so an instance left running in a switched-away session kept its place in
the event chain: the window server still routes the front account's keystrokes through a
process that cannot answer and waits out the tap timeout on each one (#1153). Both now
subscribe to `SessionActivity` in `init`, ask `SessionActivitySupport.tapShouldRun` in
the preference sync, and re-arm after a timeout only through that answer.

Both callbacks run on a thread of their own, and `SessionActivity.shared.isActive` is
written on the main thread, so the re-arm cannot read it where it runs. The callback
finishes fail-open and hops to the main queue, and the main thread either puts the tap
back or hands it to the sync to be stopped. That is what the keyboard, Super Key,
snippet and switcher taps do in the same position; the taps that live on the main run
loop read the flag in place because their callback is already there. This file already
hands its candidate keys to the main thread with `DispatchQueue.main.sync`, so the hop
is not a new relationship between the two threads.

Accessibility is asked of the system in that gate, not of `Permissions.shared`, whose
answer is a poll up to `PermissionPollingSupport.interval` old: the re-arm hands a
revoked tap to the sync to be stopped, and a stale grant there would install it straight
back. The mirror stays on `FinderCutPaste`'s per-keystroke path in `handle`, where the
comment explains it is cached to keep a live TCC round-trip off every key.

**The tap thread's own restart went around that gate.** Both files build the tap in
two places. When a stop and a start cross, `installTap()` records the pending start and
the tap thread's tail rebuilds on the spot — from the tap thread, on a decision taken
before the session handed over. `removeTap()` clears that flag under the lock, so the
window is the gap between `clearEventTapThread()` reading `true` and `installTap()`
running: a resign landing there finds `tapThread` already nil, takes the `!threadExists`
branch and resets `shouldStopTapThread`, and the tail then builds a fresh filter tap into
the session that has just left the screen. That is the stall this pull request exists to
end, reached through the one door it had left open, in the same file.

Both tails now hop to the main thread and call `syncWithPreferences()` — the same rule
the re-arm follows, and the shape `MouseClickDebounceService.finishEventTapThread` uses.
The third `clearEventTapThread()` call in each file, on the `tapCreate` failure path,
deliberately drops the pending restart and is unchanged: neither service retries a failed
creation. `AppSwitcher` and `TextSnippetService` have the same window and the same fix in
#1236, and the restart check in `MetricsTests` now asks it of every owner in the roster
that runs a tap thread.

**`FinderRenameService` disabled its port on teardown without invalidating it.** That
line is now on `main` through #1225, so what is left of it here is the removal of its
write-only `runLoopSource` field, which its twin never had.

### Trade-offs

- **Asking on the main thread vs. sampling the gate.** An earlier revision sampled
  `tapShouldRun`'s answer into a field under the tap lock, following
  `BrightnessService`. That precedent no longer exists — the sampled flag was withdrawn
  from #1231 for carrying no information its sync did not already have — and the sample
  is a second copy of the session state in the process. The cost of asking instead is
  one main-queue hop before the tap comes back, in a branch that only runs after the tap
  has already been out of the chain for its full timeout. Over that hop the keys reach
  Finder untapped, which is how the feature is off.
- **Reading `SessionActivity.shared.isActive` straight from the tap thread**, which is
  what `MouseClickDebounceService` does today. Not followed: it satisfies the wording of
  the guard without being correct on the thread it runs on.
- **Tying the observer's lifetime to the tap's.** Rejected: on a session switch the tap
  goes down and the staged cut is deliberately kept, and dropping the observer with the
  tap would leave the HUD without its background auto-hide.
- **Collapsing the two files onto a shared tap helper.** Not done. The remaining
  differences are deliberate (`.headInsertEventTap` vs `.tailAppendEventTap`, one filters
  its own synthetic pastes, one rewrites the event), so a shared helper would arrive
  carrying flags for each of them.

Both files join the shared tap-owner sweep in `MetricsTests.swift` rather than carrying
their own checks, so the re-arm rule is stated once for every owner in it. The earlier
revision's own block said the opposite of what the sweep says — it required the re-arm
*not* to name `SessionActivity.shared.` — and is gone with the sample it was written for.
The gate check added alongside reads the argument between
`SessionActivitySupport.tapShouldRun(` and its `sessionIsActive:` label rather than
banning `Permissions.shared.accessibility` across the file, because `FinderCutPaste`'s
per-keystroke cache is the same symbol used on purpose.

### Dropped from this branch: the cached front app

An earlier revision kept the frontmost bundle id on the tap thread so ⌘X, ⌘C and ⌘V
outside Finder could be turned down without waiting on `DispatchQueue.main.sync`. The
cache is refreshed by the activation observer **on the main queue** — the very thread the
reject exists to skip — so through a stall it names the app the user just left.

Stale is free for ⌘C and only ⌘C. It is not free for the other two: ⌘X swallows the key
and cuts, ⌘V moves the marked files or writes a pasteboard image out as a file, and
Finder has no native answer for either, so a stale reject makes them silently do nothing.
Narrowing the cache to ⌘C alone was the fix, and it took the win with it — the case this
started from named ⌘C and ⌘V as the two most-pressed combinations on a Mac.

What was left is smaller than it reads. `handle()` guards on
`NSWorkspace.shared.frontmostApplication?.bundleIdentifier == finderBundleID` before it
switches on the key code, so under a stale reject the main thread was letting ⌘C through
anyway. That remainder has never been measured, and it was bought with an `NSLock`, a
cache field refreshed on the thread it exists because it stalls, and an observer
condition widened to `cutPasteEnabled || pasteImageAsFileEnabled` — which left a staged
cut live behind a switched-off Cut & Paste and needed a follow-up of its own. Both the
widened condition and the `clearMarks()` split it forced are back to what `main` has.

If it is worth having, it is worth its own pull request: a measurement of the `main.sync`
on the ⌘C path, and an account of why refreshing the cache on the main thread is
acceptable when a stalled main thread is the premise.

## Verification

MacBook Pro, macOS 26. Rebased onto `main` at `a952b829`.

```
./build.sh --test      TESTS OK (9553 checks)
```

`FinderCutPaste.swift` and `FinderRenameService.swift` are both outside the `--test`
compile whitelist, so `--test` is no evidence that they compile. The whole of
`Sources/Vorssaint/**` was type-checked separately against the same target and SDK the
app build uses:

```sh
swiftc -typecheck -target arm64-apple-macosx14.0 \
    -sdk /Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk \
    -Xfrontend -interface-compiler-version -Xfrontend 6.3.2 \
    -I Sources/VMStatisticsCompat -I Sources/HIDEventSystem \
    Sources/Vorssaint/**/*.swift
```

Clean. The full `./build.sh` was **not** run: with the signing keychain locked,
`codesign` blocks on an unlock dialog and never exits.

The two input-source failures reported on the earlier revision (`nil layout falls back to
ANSI semicolon`, `App Switcher icon-row hints`) did not appear this round; the machine's
input source is ANSI now rather than a Chinese IME. They read the current input source
and are unrelated to this diff either way.

Verified both ways, each break applied alone and reverted: replacing the main-thread hop
in `FinderCutPaste` with a direct re-arm turns `FinderCutPaste.swift does not re-arm a
disabled tap into a switched-away session` red on its own; putting `installTap()` back on
`FinderRenameService`'s tail turns `FinderRenameService.swift has no tap-thread restart
that goes around the session gate` red on its own.

**Not tested.** I have no second logged-in account, so the session-switch behaviour is
inferred from the mechanism in #1153, not measured — the same caveat that issue carries.
The revoke race is not reproduced either: reaching it needs the window server to disable
one of these taps inside the polling interval that follows a revoked grant. The
restart-tail window is a read of the lifecycle code, not a reproduction — it needs a stop
and a start to cross at the moment of a handover, and the resign to land inside the gap
between two lock acquisitions. macOS 26,
Apple silicon, `--test` and `-typecheck` only.

## Anything else

Refs #1153

### DIFF
--- a/Sources/Vorssaint/Services/Finder/FinderCutPaste.swift
+++ b/Sources/Vorssaint/Services/Finder/FinderCutPaste.swift
@@ -12,12 +12,9 @@ import SwiftUI
 /// shortcuts only while Finder is frontmost and no text field is being edited,
 /// so renaming and text editing keep working untouched.
 ///
-/// A keystroke is answered on the tap thread where a wrong answer would cost
-/// nothing: which app is in front is kept there, and ⌘C outside Finder — which
-/// this service never swallows anyway — is handed straight back. ⌘X and ⌘V
-/// cross to the main thread, where the marks, the pasteboard change count and a
-/// fast Accessibility role check decide; the slow parts (reading the Finder
-/// selection, moving files) run off both threads.
+/// The decision to swallow a keystroke is made synchronously (in-memory marks +
+/// the pasteboard change count + a fast Accessibility role check); the slow
+/// parts (reading the Finder selection, moving files) run off the tap thread.
 /// Requires Accessibility, and Automation consent for Finder on first use.
 final class FinderCutPaste: ObservableObject {
     static let shared = FinderCutPaste()
@@ -86,13 +83,6 @@ final class FinderCutPaste: ObservableObject {
     private var imagePasteInProgress = false
     private var appObserver: NSObjectProtocol?
 
-    /// Which app is in front, written by the activation observer on the main
-    /// thread and read by the tap callback on its own. Nil whenever the
-    /// observer is not installed, so a cached answer can never outlive the
-    /// thing that maintains it.
-    private let routeLock = NSLock()
-    private var frontmostBundleID: String?
-
     private static let finderBundleID = "com.apple.finder"
     private static let syntheticPasteMarker: Int64 = 0x564F5249
     private static let maxRawImageBytes = 64 * 1024 * 1024
@@ -128,21 +118,6 @@ final class FinderCutPaste: ObservableObject {
         showHUD = UserDefaults.standard.object(forKey: DefaultsKey.finderCutPasteShowHUD) as? Bool ?? true
         pasteImageAsFileEnabled = available
             && UserDefaults.standard.bool(forKey: DefaultsKey.finderPasteImageAsFile)
-        // The observer is what keeps the front app the tap thread reads
-        // current, so it follows either shortcut path rather than cut and paste
-        // alone, and it goes up before the tap: a live tap with nothing
-        // maintaining that answer behind it hands ⌘C back everywhere.
-        if cutPasteEnabled || pasteImageAsFileEnabled {
-            installAppObserver()
-        } else {
-            removeAppObserver()
-        }
-        // The marks belong to cut and paste, so they go on that switch alone.
-        // Sharing the observer's condition would leave a staged cut live when
-        // cut and paste is turned off while image paste keeps the observer up:
-        // the HUD stays on screen, the change count still matches, and the next
-        // ⌘V moves the files the user thought they had un-staged.
-        if !cutPasteEnabled { clearMarks() }
         let wanted = SessionActivitySupport.tapShouldRun(
             featureWanted: cutPasteEnabled || pasteImageAsFileEnabled,
             accessibilityGranted: Permissions.shared.accessibility,
@@ -152,6 +127,12 @@ final class FinderCutPaste: ObservableObject {
             installTap()
         } else {
             removeTap()
+        }
+        if cutPasteEnabled {
+            installAppObserver()
+        } else {
+            removeAppObserver()
+            clearMarks()
         }
         refreshPanel()
     }
@@ -167,8 +148,6 @@ final class FinderCutPaste: ObservableObject {
 
     private func installAppObserver() {
         guard appObserver == nil else { return }
-        // The app already in front sends no activation of its own.
-        cacheFrontmostApplication()
         appObserver = NSWorkspace.shared.notificationCenter.addObserver(
             forName: NSWorkspace.didActivateApplicationNotification,
             object: nil,
@@ -183,18 +162,11 @@ final class FinderCutPaste: ObservableObject {
             NSWorkspace.shared.notificationCenter.removeObserver(observer)
             appObserver = nil
         }
-        routeLock.withLock { frontmostBundleID = nil }
     }
 
     private func handleApplicationActivation() {
-        cacheFrontmostApplication()
         guard cutPasteEnabled else { return }
         refreshPanel()
-    }
-
-    private func cacheFrontmostApplication() {
-        let bundleID = NSWorkspace.shared.frontmostApplication?.bundleIdentifier
-        routeLock.withLock { frontmostBundleID = bundleID }
     }
 
     // MARK: - Event tap
@@ -293,10 +265,9 @@ final class FinderCutPaste: ObservableObject {
         }
     }
 
-    /// Runs on the tap thread. Every key except a plain Command-X/C/V returns
-    /// after reading only the event, and Command-C outside Finder after one
-    /// cached string as well; the rare candidate is handed to the main thread
-    /// where the service's UI and pasteboard state live.
+    /// Runs on the tap thread. Every key except plain Command-X/C/V returns
+    /// after reading only the event itself; the rare candidate is handed to
+    /// the main thread where the service's UI and pasteboard state live.
     private func route(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
             // The gate the preference sync applied, asked again: a tap re-armed
@@ -320,25 +291,6 @@ final class FinderCutPaste: ObservableObject {
               !flags.contains(.maskControl), !flags.contains(.maskAlternate),
               keyCode == Key.x || keyCode == Key.c || keyCode == Key.v
         else { return Unmanaged.passUnretained(event) }
-
-        // ⌘C is one of the two most-pressed combinations on the machine, and
-        // outside Finder the answer is always no. Learning that on the main
-        // thread would make every copy anywhere wait for whatever this app is
-        // drawing, so the activation observer keeps the answer over here.
-        //
-        // Only ⌘C. The observer refreshes this on the main queue, which is the
-        // thread this reject exists because it is busy, so through that stall
-        // the cache still names the app the user just left. A stale reject is
-        // free here — the main thread passes ⌘C through in every branch, and
-        // the pending-cut marks it would have dropped are dropped again by the
-        // change-count check on the next ⌘V. It is not free for the other two:
-        // ⌘X swallows the key and cuts, ⌘V moves the marked files or writes the
-        // pasteboard image out as a file, and Finder has no native answer for
-        // either, so a stale reject makes them silently do nothing.
-        if keyCode == Key.c,
-           routeLock.withLock({ frontmostBundleID }) != Self.finderBundleID {
-            return Unmanaged.passUnretained(event)
-        }
 
         var verdict: Unmanaged<CGEvent>?
         DispatchQueue.main.sync {
--- a/Sources/Vorssaint/Services/Finder/FinderCutPaste.swift
+++ b/Sources/Vorssaint/Services/Finder/FinderCutPaste.swift
@@ -12,12 +12,12 @@ import SwiftUI
 /// shortcuts only while Finder is frontmost and no text field is being edited,
 /// so renaming and text editing keep working untouched.
 ///
-/// A keystroke is answered on the tap thread wherever it can be: which app is
-/// in front is kept there, so ⌘C and ⌘V outside Finder are handed straight
-/// back. Only those combinations inside Finder cross to the main thread, where
-/// the marks, the pasteboard change count and a fast Accessibility role check
-/// decide; the slow parts (reading the Finder selection, moving files) run off
-/// both threads.
+/// A keystroke is answered on the tap thread where a wrong answer would cost
+/// nothing: which app is in front is kept there, and ⌘C outside Finder — which
+/// this service never swallows anyway — is handed straight back. ⌘X and ⌘V
+/// cross to the main thread, where the marks, the pasteboard change count and a
+/// fast Accessibility role check decide; the slow parts (reading the Finder
+/// selection, moving files) run off both threads.
 /// Requires Accessibility, and Automation consent for Finder on first use.
 final class FinderCutPaste: ObservableObject {
     static let shared = FinderCutPaste()
@@ -131,7 +131,7 @@ final class FinderCutPaste: ObservableObject {
         // The observer is what keeps the front app the tap thread reads
         // current, so it follows either shortcut path rather than cut and paste
         // alone, and it goes up before the tap: a live tap with nothing
-        // maintaining that answer behind it hands every shortcut back.
+        // maintaining that answer behind it hands ⌘C back everywhere.
         if cutPasteEnabled || pasteImageAsFileEnabled {
             installAppObserver()
         } else {
@@ -288,10 +288,10 @@ final class FinderCutPaste: ObservableObject {
         }
     }
 
-    /// Runs on the tap thread. Every key except a plain Command-X/C/V pressed
-    /// in Finder returns after reading only the event and one cached string;
-    /// the rare candidate is handed to the main thread where the service's UI
-    /// and pasteboard state live.
+    /// Runs on the tap thread. Every key except a plain Command-X/C/V returns
+    /// after reading only the event, and Command-C outside Finder after one
+    /// cached string as well; the rare candidate is handed to the main thread
+    /// where the service's UI and pasteboard state live.
     private func route(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
             // The gate the preference sync applied, asked again: a tap re-armed
@@ -316,13 +316,24 @@ final class FinderCutPaste: ObservableObject {
               keyCode == Key.x || keyCode == Key.c || keyCode == Key.v
         else { return Unmanaged.passUnretained(event) }
 
-        // ⌘C and ⌘V are the two most-pressed combinations on the machine, and
+        // ⌘C is one of the two most-pressed combinations on the machine, and
         // outside Finder the answer is always no. Learning that on the main
         // thread would make every copy anywhere wait for whatever this app is
-        // drawing, so the activation observer keeps the answer over here. The
-        // main thread reads the live value again before acting on it.
-        guard routeLock.withLock({ frontmostBundleID }) == Self.finderBundleID
-        else { return Unmanaged.passUnretained(event) }
+        // drawing, so the activation observer keeps the answer over here.
+        //
+        // Only ⌘C. The observer refreshes this on the main queue, which is the
+        // thread this reject exists because it is busy, so through that stall
+        // the cache still names the app the user just left. A stale reject is
+        // free here — the main thread passes ⌘C through in every branch, and
+        // the pending-cut marks it would have dropped are dropped again by the
+        // change-count check on the next ⌘V. It is not free for the other two:
+        // ⌘X swallows the key and cuts, ⌘V moves the marked files or writes the
+        // pasteboard image out as a file, and Finder has no native answer for
+        // either, so a stale reject makes them silently do nothing.
+        if keyCode == Key.c,
+           routeLock.withLock({ frontmostBundleID }) != Self.finderBundleID {
+            return Unmanaged.passUnretained(event)
+        }
 
         var verdict: Unmanaged<CGEvent>?
         DispatchQueue.main.sync {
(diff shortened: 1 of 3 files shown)


PathGao:
Sampling the gate instead of reading `isActive` from the tap thread is right, and the `CFMachPortInvalidate` divergence between the twins is a real catch.

**The cache is authoritative for rejection.** The activation notification that refreshes `frontmostBundleID` is delivered on the main queue — the thread this PR says is busy enough to matter. Click into Finder mid-stall, press ⌘X, and the cut silently doesn't happen; before this it was slow and correct. The body's "never the authority" only covers false positives. Laziest fix that keeps the win: fast-reject `Key.c` and `Key.v` only. ⌘X isn't hot, and its miss is the invisible one.

**`BrightnessService` has no `SessionActivity` reference** — it isn't the precedent. The service that actually has your shape is `MouseClickDebounceService.handle`, which still reads `SessionActivity.shared.isActive` from its own tap thread. Not this PR's job; worth knowing it's three, not two.

Neither file is in `--test`'s compile set and I can't run anything, so nothing here is verified beyond reading `main`.

**Impact:** ⌘X in Finder can silently no-op while the main thread stalls.
**Merge:** needs ⌘X off the cached reject first.

ref #1153

PathGao:

PathGao:
@PathGao Closed the ⌘X gap, and it turned out to cover ⌘V too.

The cached front app is refreshed by the activation observer on the main queue — the same thread the reject exists to skip — so through a stall it names the app the user just left. Stale is free for ⌘C and only ⌘C: the main thread passes it through in every branch, and the pending-cut marks it would have dropped are dropped again by the change-count check on the next ⌘V. ⌘X swallows the key and cuts, ⌘V moves the marked files or writes a pasteboard image out as a file, and Finder answers for neither, so a stale reject makes both silently do nothing. Your line named ⌘X; ⌘V sits on the same reject with the same consequence, so both now ask the main thread and the cheap path keeps the combination it was actually worth.

The guard is the only thing in `route` that turns a candidate down before the main thread, so there is no second site of this in the file. `MetricsTests` now pins the rule by source shape: the slice ahead of `DispatchQueue.main.sync` must reject on `Key.c` and must not mention `Key.x` or `Key.v`. Verified red first — widening the reject back over `Key.x` produced that assertion and nothing else, restoring returned it to green.

`./build.sh --test` is fully green here, 9476 checks. Worth noting for the numbers in this thread: the two assertions reported red on earlier runs of this branch pass now — they read the machine's live input source, so they are not a baseline either way.

What I did not do: the full `./build.sh` was not run for this commit. `FinderCutPaste.swift` is outside the `--test` whitelist, so a green `--test` is not evidence it type-checks; CI compiles it on both runners and that is the check to read here, not this machine. The stale-reject path was not reproduced live either — reaching it needs the main thread genuinely stalled at the moment of a ⌘X — so it is pinned by source shape, not by a timed run. The session-switch half is still unverified for want of a second account.

PathGao:
⌘V was the right catch — same reject, same silent no-op — and pinning it by source shape is what makes it stick.

**`clearMarks()` rode along with the observer.** The body frames the widening as an observer change ("with image paste alone the tap existed and the observer did not"), but `clearMarks()` shared that `else` branch, so its condition widened too. Turn Finder Cut & Paste off with image paste still on and a cut staged: on `main` the marks were dropped, now they survive. `refreshPanel` keeps the HUD up for a feature that is off, and `markedChangeCount` is untouched — so after re-enabling, the next ⌘V still matches the change count and moves files staged before the user switched the feature off. Split the branch:

```swift
if cutPasteEnabled || pasteImageAsFileEnabled { installAppObserver() } else { removeAppObserver() }
if !cutPasteEnabled { clearMarks() }
```

Read off `main` plus the diff; I ran nothing.

**Impact:** with image paste on, disabling cut & paste leaves a staged cut live — HUD stays up, and a later ⌘V moves those files.
**Merge:** needs `clearMarks()` back on `cutPasteEnabled` alone first.

PathGao:
@PathGao Right — `clearMarks()` rode the observer's condition, and that is a data bug, not a lifecycle nit. Split exactly as you wrote it: `installAppObserver()` / `removeAppObserver()` stay on `cutPasteEnabled || pasteImageAsFileEnabled`, and `if !cutPasteEnabled { clearMarks() }` is its own statement. With image paste on, turning cut & paste off now drops the marks again, `resetCutState` zeroes `markedChangeCount`, and the `refreshPanel` it ends with takes the HUD down with them.

Traced both settings' teardown for the same conflation. `FinderRenameService` owns no marks. `suspend()` drops observer, tap and marks together on purpose — it runs before the app resets its own permissions, so everything goes. Inside `handle`, `Key.x` guards on `cutPasteEnabled`, the `Key.c` mark-drop guards on `cutPasteEnabled`, and the image path guards on `pasteImageAsFileEnabled`; those were already separate key by key. The observer's install condition is the only combined one left and stays combined, which was the point of widening it — the tap thread needs the front app cached under either feature.

`MetricsTests` pins the split by source shape: the last `if` ahead of `clearMarks()` in `syncWithPreferences` must name `cutPasteEnabled` and must not name `pasteImageAsFileEnabled`. Verified red first — restoring `clearMarks()` to the observer's `else` produced `turning Finder cut & paste off clears the staged cut on its own switch alone` and no other new failure; splitting it again returned to green.

`./build.sh --test`: 9477 checks, 2 failing — `nil layout falls back to ANSI semicolon` and `App Switcher icon-row hints describe default app and window shortcuts`. Both are red on this branch's base, `16216d56`, in a throwaway worktree on this machine right now, so they are the machine's live input source (a Chinese IME) and not this branch. Not verified: `FinderCutPaste.swift` is outside the `--test` whitelist, so it is not compiled here — CI covers it on both runners. The full `./build.sh` was not run; with the signing keychain locked codesign blocks on an unlock dialog. The surviving-cut path was not reproduced live either, and the session-switch half is still unverified for want of a second account.

PathGao:
Both fixes are in, and the assertions parse against the new shape — the `clearGuard` slice lands on `!cutPasteEnabled {` and `cachedReject` on `keyCode == Key.c,`.

One thing worth recording: the stale ⌘C reject is safer than the body argues. `handle` guards on `NSWorkspace.shared.frontmostApplication?.bundleIdentifier == finderBundleID` *before* the `Key.c` mark-drop, so on a stale reject the main thread would have returned passthrough without touching the marks anyway. The change-count fallback is a second net, not the first one.

**`CFMachPortInvalidate` is one divergence in this pair, not in the repo.** Eight more files remove a CGEvent tap's run-loop source and never invalidate the port: `BrightnessService` (`removeKeyTap` and `runFunctionKeyTap`, two sites), `AppSwitcher.runEventTap`, `KeyboardDebounceService`, `DockClickService`, `DockPreviewService`, `MusicLaunchBlocker`, `PreciseVolumeRollerService`, `WindowMaximizer`. I read the first four; the last four are grep only — `CFRunLoopRemoveSource` present, `CFMachPortInvalidate` absent from the whole file. Same mechanism #1153 names, so it belongs there, not here.

Read off `main` plus the diff; I ran nothing.

**Impact:** none.
**Merge:** ship it.

ref #1153

PathGao:
@PathGao Dropped the cached front app in `113bc8f8`. Not a force-push — the removal is its own commit on top of what you reviewed. What is left is the session gate alone, and the title says only that now.

**What went:** `routeLock`, `frontmostBundleID`, `cacheFrontmostApplication()` and its three call sites, the `Key.c` early return in `route()`, and the observer condition widened to `cutPasteEnabled || pasteImageAsFileEnabled`. `clearMarks()` goes back into its `else` — checked rather than assumed: on `main` it already sits in the `else` of `if cutPasteEnabled`, and it only needed splitting out because the widened condition dragged it along. With the condition back, the split has nothing left to do.

**Why:** your review is what took the case for it apart. The cache was authoritative for rejection, so ⌘X had to go back to asking the main thread, and then ⌘V with it — and ⌘C and ⌘V were the two combinations the summary named. Then the guard on `NSWorkspace.shared.frontmostApplication?.bundleIdentifier == finderBundleID` above the `Key.c` mark-drop turned out to mean the main thread returns passthrough on a stale reject anyway. What was left was one shortcut, never profiled, paid for with a lock, a cache field refreshed *on the main thread* — the thread whose stalling is the reason it exists — and a widened observer condition that produced a second data defect on the way through. To come back it needs its own pull request, a measured `main.sync` on the ⌘C path, and an answer for refreshing the cache off the main thread.

**One thing worth your eye.** The branch had four source-shape assertions around this. I put all four back against the reverted source: only `only ⌘C is turned down from the cached front app` goes red. The other three — the `didActivateApplicationNotification` one, the `beforeMainThread` one, and the `clearGuard` one — are green on `main` too. They never pinned anything this branch added. I deleted all four anyway, since the comments on them describe code that is gone, but three of them were decoration from the day they were written.

**Checks:** `TESTS FAILED (2 of 9473)`, both the input-source-dependent pair. Neither Finder file is in the `--test` whitelist, so a whole-module `swiftc -typecheck` over all 424 sources at the build's target and SDK: exit 0, no warnings.

**Not done:** no rebase. `FinderRenameService.swift`'s `CFMachPortInvalidate` is already on `main` via #1225 and will drop out then, and the roster this branch touches in `MetricsTests` is what #1230 removes.

PathGao:
@PathGao Rebased onto `main` at `a952b829`, head is now `ac909643`. The
sampled gate is gone (`c608abe4`).

**Why it went.** The sample was written against `BrightnessService`, and that
precedent no longer exists — the sampled flag was withdrawn from #1231 for
carrying no information its own sync did not already have. What was left here
was a second copy of the session state kept under the tap lock, for a branch
that only runs after the tap has already been out of the chain for its full
timeout.

Both taps now do what the keyboard, Super Key, snippet and switcher taps do:
the callback finishes fail-open, hops to the main queue — where the flag is
written — and the main thread either re-arms or hands the tap to the sync to
be stopped. `FinderCutPaste` already hands its candidate keys to main with
`DispatchQueue.main.sync`, so this is not a new relationship between the two
threads, and over the hop the keys reach Finder untapped, which is how the
feature is off.

I did check the third option before choosing: `MouseClickDebounceService` on
`main` reads `SessionActivity.shared.isActive` straight from its own tap
thread, same threaded shape, same `shouldStopTapThread`. It is the shortest
diff and it is a cross-thread read of a plain `Bool` written on main. Not
followed, and not touched either — it is not in this diff.

**Accessibility.** The gate asks `AXIsProcessTrusted()` in both files now,
matching #1233, #1235 and #1236. The mirror stays where `handle` caches it on
the per-keystroke path, with its comment.

**The assertion.** This branch's own block said the opposite of what the
shared sweep says — it required the re-arm *not* to name
`SessionActivity.shared.` — so it went out with the sample. Both files join
the shared tap-owner loop instead, and the gate check there reads the
argument between `tapShouldRun(` and `sessionIsActive:` rather than banning
the symbol across the file, precisely so `handle`'s deliberate cache is not
read as the defect.

The `CFMachPortInvalidate` hunk is gone too — #1225 landed it. What is left of
that half is the removal of `FinderRenameService`'s write-only
`runLoopSource` field.

**Tests.** `./build.sh --test` → `TESTS OK (9549 checks)`. The two
machine is on ANSI now. `swiftc -typecheck` over all of `Sources/Vorssaint/**`
is clean; neither file is in the `--test` whitelist. Full `./build.sh` not
run — signing keychain locked.

Verified both ways: replacing the hop with a direct re-arm turns
`FinderCutPaste.swift does not re-arm a disabled tap into a switched-away
session` red on its own.

**Where to look first.** The hop's cost, on this file specifically. A tap
timeout here means the main thread was busy — that is what
`DispatchQueue.main.sync` in `route` runs into — and the re-arm now queues
behind that same busy main thread. The feature stays off for one more
main-queue turn than the sample would have taken, exactly when it is most
likely to have timed out. I think that is the right trade for not keeping a
second copy of the session state, but it is the part I would push on.

**Still open in this file, not fixed here.** The tap-thread tail still
restarts with `if clearEventTapThread() { installTap() }`, which rebuilds the
tap without asking the gate if a stop and a start crossed at handover time.
#1236 fixes that shape in `AppSwitcher` and `TextSnippetService`; the same
line is here, in `FinderRenameService`, in `SuperKeyService` and in
`KeyboardDebounceService`. I left it alone rather than widen this branch —
say the word and it is four one-line changes across the three PRs.

PathGao:
@PathGao Answering your call on the restart path — both tails are gated now, in
`e437653b`. Not a force-push; it is one commit on top of what you last saw.

**Confirmed reachable before touching it**, and it is the same window you already
accepted the fix for in `AppSwitcher`, not a weaker one. `removeTap()` clears
`pendingTapRestart` under the lock, so this is not a straight miss — the gap is between
`clearEventTapThread()` reading `true` and releasing the lock and `installTap()` running.
A resign landing there finds `tapThread` already nil, takes the `!threadExists` branch and
resets `shouldStopTapThread`, and the tail then builds a fresh filter tap into the session
that has just left the screen. `AppSwitcher.removeTap` and `FinderCutPaste.removeTap` are
the same shape line for line, so the reachability argument is identical across all six
files.

Both tails hop to main and call `syncWithPreferences()` now. The `tapCreate` failure path
(`_ = clearEventTapThread()`) drops the pending restart deliberately and is untouched —
neither service retries a failed creation.

`MetricsTests` asks it of every roster owner that runs a tap thread again, not only the
files that adopted the helper. Verified both ways: putting `installTap()` back on
`FinderRenameService`'s tail turns it red on its own.

**The cost I raised last round still stands, and this adds to it.** A tap timeout in this
file means the main thread was busy — that is what `route`'s `DispatchQueue.main.sync`
runs into — and the re-arm now queues behind that same busy main thread, so the feature
stays off for one more main-queue turn than the sample would have taken, exactly when it
is most likely to have timed out. The restart tail now queues there too. I think both are
the right trade for not keeping a second copy of the session state and not rebuilding a
tap on a stale decision, but it is the same cost paid twice and it is the part I would
push on.

**`restartTapOnMain` is six copies now and I left it that way.** The shared content is one
stdlib call, `DispatchQueue.main.async { [weak self] in self?.syncWithPreferences() }`. A
free function taking a closure is that call with a wrapper and the caller still writes
`[weak self]`; the only form that would actually shorten anything is a protocol with a
`syncWithPreferences()` requirement and a default extension, which is the gated-tap-owner
layer over six classes that share nothing else. What each copy carries that is worth
having is its doc comment, and those differ per file — `SuperKeyService`'s says the hop
was already there and only the question was missing. Moving the method would move the
comments away from the tails they explain. Say so if you would rather have it in
`SessionActivitySupport` anyway.

`./build.sh --test` → `TESTS OK (9553 checks)`, no failures; the two input-source ones are
still absent, this machine is on ANSI. `swiftc -typecheck` over all of
`Sources/Vorssaint/**` clean. Full `./build.sh` not run — signing keychain locked.

**Where to look first.** Whether the restart tail should publish anything about the tap
being down over the hop. These two have no `isRunning` to publish, so nothing was needed;
`KeyboardDebounceService` in #1235 did and I added it there. If you disagree with that
asymmetry it is the seam between the two branches.

PathGao:
Both tails through `syncWithPreferences`, and the walk through `removeTap`'s `!threadExists` branch is what makes it a fix rather than a nit.

**`FinderCutPaste` does have an `isRunning`.** `CutPasteSettings` draws its "active now" row off it. It is computed off the lock, not `@Published`, so it never publishes on its own — over the hop the tap is down and the row still says active, and on the way back `syncWithPreferences` publishes before the tap thread sets `tap`, so the row goes and stays gone. `TextSnippetService.isRunning` is the same shape. The asymmetry with #1235 isn't "nothing to publish", it's "computed, so stale unnoticed". Cosmetic, and not new here.

**No other roster member runs a tap thread.** `if code.contains("clearEventTapThread")` can only fire for the two files this PR adds; `AppSwitcher`, `TextSnippetService`, `SuperKeyService` and `KeyboardDebounceService` all carry the line and none are in that list.

**The second `expect` hard-codes a name.** The ban half accepts `startOnMain()`, then the split on `private func restartTapOnMain` yields `""` for anything else and fails it. Both `SuperKeyService` and `TextSnippetService` have `startOnMain` today, so #1236 renames or goes red.

Read off `main` plus the diff; I ran nothing.

**Impact:** none on the tap lifecycle; Settings can show "active now" for a tap that is down.
**Merge:** ship it.

ref #1153

PathGao:

PathGao:
Reopened as #1304.

PR #1235, opened 2026-09-01
fix(keyboard): hand the keyboard, Super Key and recording taps back across a user switch

PathGao:
## Summary

Three keyboard event taps stayed in the chain across a fast user switch. They
now follow the session the way the mouse taps already do: handed back on
resign, rebuilt from the preferences on the way in, and the timeout re-arm asks
whether this session is still the one on screen before putting a tap back.

**`KeyboardDebounceService`** is the one that mattered most. It creates its tap
at the HID stage with `.defaultTap`, which is ahead of the split into login
sessions, and `KeyboardDebounceState` keys its bounce windows on the key code
alone — nothing in it knows which account is typing. A session left running
behind a switch therefore does not merely stall the account on screen; it
answers that account's keystrokes out of timing state built from someone else's
typing, and its answer can be `nil`, which is a keystroke that never arrives.
Its teardown also only disabled the tap and never called `CFMachPortInvalidate`.
#1225 has since landed the call after the run loop returns; what is left here
is the second one, in `stop()`, so a teardown that reaches the port before the
thread does hands it back too. That is the pair its twin
`MouseClickDebounceService` has. The re-arm asks that twin's full question:
`AXIsProcessTrusted()` as well as the session, because a modifying tap whose
permission went away while it was disabled has to end rather than go back in.

The branch that question falls through to has to be able to end it, and it
could not. `syncWithPreferences` read Accessibility off
`Permissions.shared.accessibility`, a value polled at
`PermissionPollingSupport.interval` (60s); the twin passes `AXIsProcessTrusted()`.
Inside that window the gate still computed true, so the sync called `start()`,
which saw the tap thread still up, published running and returned. The tap was
neither back in the chain nor stopped, and `isRunning` went on saying it was
live. The gate now asks the system directly, so the sync reaches `stop()` and
publishes not-running.

**`SuperKeyService`** owns two taps (session-level for keys, HID for mouse
buttons) plus a `hidutil` keyboard mapping. The mapping is the part that is
easy to miss: `hidutil` writes it for the whole Mac, so a session that keeps it
across a switch hands the account switched in a source key that does nothing.
`stop()` already clears the mapping, so gating the sync covers both halves at
once. Both taps are re-armed through one guarded path.

That path asked the session only. `start()` carries the Accessibility guard, but
a re-arm never reaches `start()`, so a grant revoked mid-session left both taps
disabled — same modifying `.defaultTap`, same failure as the debounce one — with
`isRunning` still true behind them. It now asks `AXIsProcessTrusted()` too. The
sync it falls through to sees the disabled tap as dead through its existing
`CGEvent.tapIsEnabled` check, tears down and rebuilds, and `start()`'s guard
clears the leftover mapping and publishes not-running.

**`ShortcutRecordingTap`** swallows every key while a field is listening, so a
recorder still open across a switch is the worst-behaved of the three for as
long as it lasts. #1153 left this one open with "it only exists while someone
is recording a shortcut in Settings … unless it stays armed after the sheet
closes, which I didn't check". Checked: the Settings fields do stop themselves
(`ShortcutRecorderButton.observeExits` subscribes to `didResignActive`), but the
command bar's shortcut-capture card is a second caller, and nothing hides that
panel on resign or deactivation. The tap is shared, so the guard is in the tap.

Dropping the tap is not the whole exit, though, and the first version of this
branch did only that. Everything that calls `ShortcutRecordingTap.begin` also
called `ShortcutCapture.begin()`, which switches off `HotkeyManager`, Shelf,
clipboard, sound output and the window-layout shortcuts; only
`ShortcutCapture.end()` gives them back. Switching away with the command bar's
capture card open therefore came back with the card still up, the tap gone and
every global shortcut in the app dead. Both session-driven teardowns — the
`onChange` handler and the timeout branch that declines to re-arm — now go
through one exit that ends the capture too. The five ordinary exits (the command
bar's `endCapturingShortcut`, `ShortcutRecorderButton.stopRecording` and its
`deinit`) already paired the two and are untouched; `ShortcutCapture.end()` is
idempotent, so whichever of them runs after a session switch is a no-op.
Nothing here hides or re-opens the capture card: giving the keys back early is
the direction `ShortcutCapture` documents as the safe one, and the card's own
local-monitor fallback still records.

That exit also has to be installed before the tap is asked for, and it was not.
The `onChange` registration sat inside `if tap == nil`, below
`guard AXIsProcessTrusted()`, so on a Mac without the grant `begin` returned at
the guard and nothing was ever listening for the switch — while
`ShortcutCapture.begin()` had already switched the app's shortcuts off. The
capture card still records through its caller's own key monitor, the documented
fallback, so nothing looks wrong until the session resigns and every global
shortcut stays dead with no exit left to give them back: round one's symptom
reached through the other door. The registration is now above the guard. It is
`observingSession`-guarded and kept for the process, so on a Mac that does have
the grant the first tap-creating recording installed it anyway; the gap was the
stretch before that, not a permanent state. Both halves of
`endForSessionSwitch()` are safe with no tap — `tearDown()` touches nil state
and `ShortcutCapture.end()` guards on `isCapturing`.

### The tap thread's own restart

Both threaded services build the tap in two places, and neither second place asked
the gate. When a stop and a start cross, the pending start is recorded and the tap
thread's tail rebuilds on a decision taken before the session handed over. `stop()`
clears that flag under the lock, so the window is the gap between
`clearEventTapThread()` reading `true` and the restart running: a resign landing there
finds the thread already gone, resets `shouldStopTapThread`, and the tail builds a
fresh tap into the session that has just left the screen.

The two were missing different amounts of it.

`KeyboardDebounceService` called `start()` straight from the tap thread, so both halves
were wrong — the thread and the question. It now hops and syncs. `isRunning` is published
false before the hop rather than only on the no-restart branch, so the gap between the two
is not a green "Active now" tick over a tap that is down; the sync publishes it true again
if it starts one. `publishRunning` already drops a stale generation and the two main-queue
blocks run in order, so nothing else is needed to keep that honest.

`SuperKeyService` already hopped: `startOnMain()` was a `DispatchQueue.main.async` onto
`start()`. What it did not do is ask. `start()` guards on `AXIsProcessTrusted()` alone and
the session conjunct lives in `syncWithPreferences`, which the restart skipped — so a
restart that crossed a handover rebuilt both taps in a switched-away session with the
grant still in place, which is the common case. It calls the sync now, and is renamed
`restartTapOnMain` to say so: "on main" was never the part that was missing.

The third `clearEventTapThread()` call in each file, on the `tapCreate` failure path,
deliberately drops the pending restart and is unchanged: neither service retries a failed
creation. `AppSwitcher` and `TextSnippetService` have the same window and the same fix in
#1236, and `FinderCutPaste` and `FinderRenameService` in #1234.

**One of the two is not pinned.** The restart check in `MetricsTests` asks this of every
owner in the roster that runs a tap thread, and it catches `SuperKeyService` — pointing
its helper back at `start()` turns it red. It does not reach
`KeyboardDebounceService`'s tail, which branches on a local
(`let shouldRestart = clearEventTapThread()` … `if shouldRestart { … }`) rather than
spelling the call inline the way the other five do. Reverting both of its tails to
`start()` leaves the suite green, which I checked rather than assumed. Pinning it would
mean matching a variable name and its indentation, or reshaping the tail to suit the
check; neither is worth it, so it is named here instead.

### Trade-offs

The re-arm in the first two runs on the tap's own thread, not the main run
loop. `SessionActivity.isActive` is written on the main thread, so reading it
from there would be a cross-thread read of main-thread state — the reason
`BrightnessService` samples the answer under `keyThreadLock` instead. Two ways
out were available and a third was rejected:

- **Sample the gate under the existing lock** (the `BrightnessService` shape).
  Decides instantly, but the tap thread then acts on an answer from the last
  preference sync.
- **Ask on the main thread** (chosen). The callback hands the event back
  fail-open, then hops to main, where the flag is written, and either re-arms or
  re-syncs. The answer cannot be stale relative to the `onChange` handler,
  because both run on the same thread in order. The cost is one main-queue hop
  before the tap comes back; the tap had already been out of the chain for its
  full timeout by then, and `KeyboardDebounceService` now resets its debounce
  state over that gap so the untapped stretch is not read as silence and the
  next press as a bounce.
- **Read `SessionActivity.shared.isActive` straight from the tap thread**, which
  is what `MouseClickDebounceService` does today. Rejected: it satisfies the
  wording of the guard without being correct on the thread it runs on.

The chosen option is now the answer across the four branches that gate a tap on
the session. #1234 sampled the gate instead and has moved to this shape;
`FinderCutPaste`, `FinderRenameService`, `AppSwitcher` and `TextSnippetService`
all hop to the main thread in the same position now, and the taps that live on
the main run loop — `WindowLayoutService`, `WindowMaximizer`,
`ShortcutRecordingTap`, `PreciseVolumeRollerService`, `DockClickService` — read
the flag in place because their callback is already there. One rule: the re-arm
decision is made on the thread that writes the answer.

Both Accessibility gates route through the existing stop paths rather than just
declining to re-arm, because `isRunning` has live readers: `@ObservedObject`
bindings in `KeyboardDebounceSettings`, `SuperKeySettings` and
`ShortcutsSettings`. A stale `true` is a green "Active now" tick over a dead tap
and a Super Key alternative offered in the shortcut rows for a tap that is not
running. Closing a gate without reporting the state is what the previous
revision did. No new write of `isRunning` was added — the existing `stop()` and
`start()` paths publish it at the right moment once the gates are honest.

`ShortcutRecordingTap` runs on the main run loop, so it reads the flag directly,
and defers only the teardown — invalidating a port from inside its own callback
stack is unsafe. Deferring it also puts `ShortcutCapture.end()` on the main
thread, which is where it belongs.

Nothing here adds a retry on `tapCreate` failure. None of the three retried
before: the debounce and Super Key services publish "not running" and stop,
and the recording tap returns `false` so the field falls back to plain view
events. The bounded retry `f25d1aa8` gave the scroll taps exists because those
retried without a bound; adding one where none existed would be a separate
change.

The three files are added to the existing session-switch guard in
`Tests/MetricsTests.swift`, which pins all four halves of the wiring as text.
Four more assertions pin the capture pairing: that one exit exists, that it
calls both halves, and that each of the two session paths reaches it. The
debounce re-arm's `AXIsProcessTrusted()` is asserted on the re-arm slice
itself rather than on the file, so a mention elsewhere cannot answer for it.
That re-arm-slice assertion now covers `SuperKeyService` alongside it, and one
more pins the debounce gate to `AXIsProcessTrusted()` on the
`syncWithPreferences` slice, so the polled cache cannot come back. Two further
assertions pin the session half of the same two gates, on those same slices:
`sessionIsActive: SessionActivity.shared.isActive` in the debounce gate, and the
`&& SessionActivity.shared.isActive` conjunct in Super Key's `enabled`. The
re-arm declining is only half of each fix — the sync answering `onChange` is
what actually reaches `stop()` on a resign, and for Super Key that conjunct is
the whole of it. `ShortcutRecordingTap` needs no gate assertion of that kind: it has no
sync, and both of its session paths are already pinned. It gains one of its own
instead, on the slice of `begin` above the Accessibility guard, so the
tap-creation path cannot answer for where the session handler is registered.

## Verification

```
./build.sh --test      TESTS OK (9559 checks)
```

None of `KeyboardDebounceService.swift`, `SuperKeyService.swift` or
`ShortcutRecordingTap.swift` is in the `--test` compile list, so a green
`--test` did not compile them. The whole of `Sources/Vorssaint/**` was
type-checked separately with the build's own target and SDK
(`arm64-apple-macosx14.0`, `MacOSX26.sdk`, plus the two `-I` compat paths) and
is clean. The full `./build.sh` was not run for this revision: with the signing
keychain locked, `codesign` blocks on an unlock dialog. (It was run and
reported on the earlier head `0f5e2de8` in the thread.)

The two input-source failures reported on earlier revisions (`nil layout falls
back to ANSI semicolon`, `App Switcher icon-row hints`) did not appear this
round; the machine's input source is ANSI now rather than a Chinese IME. They
read the current input source and are unrelated to this diff either way.

Every assertion added on this branch was checked red first, and each one on its
own, by removing the real conjunct from the source rather than by misspelling
it: deleting `&& SessionActivity.shared.isActive` from
`SuperKeyService.syncWithPreferences` turns only "the Super Key sync gate stops
both taps when the session leaves" red — every earlier assertion, including the
re-arm-slice ones, stays green under that break, which is the hole it was added
for; passing `sessionIsActive: true` in the debounce gate turns only "the
keyboard debounce gate stops the tap when the session leaves" red; putting
`accessibilityGranted: Permissions.shared.accessibility` back turns "the
keyboard debounce gate asks the system for Accessibility, not the poll cache"
red on its own; dropping `AXIsProcessTrusted()` from the Super Key re-arm turns
"SuperKeyService.swift does not put a modifying tap back after Accessibility is
lost" red on its own; restoring the previous `ShortcutRecordingTap.swift` turns
the four capture assertions red together; and "the recording tap watches the
session even when Accessibility denies it a tap" is red with the registration
back below the guard and green with it above. For the restart tails: pointing
`SuperKeyService`'s helper back at `start()` turns "SuperKeyService.swift restarts
its tap thread through the gate on the main thread" red on its own; reverting
`KeyboardDebounceService`'s two tails leaves the suite green, which is the gap
described above.

**Not tested:** there is no second login session on this machine, and
Accessibility was not revoked by hand — the app was never launched. Every claim
about what the account switched in experiences, and about what happens after a
revoked grant, is read off `KeyboardDebounceState`, the `hidutil` call,
`ShortcutCapture`'s callers, the two services' `start()`/`stop()` paths and
HID-versus-session tap placement. None of it was reproduced. Whether the window
server lets a tap re-enable at all once Accessibility is gone is not something I
checked either; the change makes the app stop asking, which is correct in both
cases. The same caveat stands on #1075 and #1153 themselves.

## Anything else

The capture card itself is left alone. Nothing hides the command bar panel on
session resign, so the card is still on screen for the account that comes back —
it just no longer holds the keyboard or the app's shortcuts. Hiding the panel
belongs to whoever owns the command bar's session behaviour, not to a tap.

`ScrollInverter` is the one remaining `SessionActivitySupport.tapShouldRun`
caller still passing `Permissions.shared.accessibility`, and its re-arm gates on
the session without asking Accessibility. Every other caller
(`SmoothScrollService`, `MouseNavigationService`, `MouseButtonShortcutService`,
`MiddleClickService`, `QuitProtectionService`, `MouseClickDebounceService`) asks
the system. That shape landed with its own review in #1145 and is not in this
branch's diff, so it is left alone here rather than widening this PR into a
third feature.

Refs #1075
Refs #1153

### DIFF
--- a/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
+++ b/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
@@ -27,12 +27,27 @@ final class KeyboardDebounceService: ObservableObject {
                                                 globalWindowMs: Defaults.defaultKeyboardDebounceWindowMs,
                                                 keyWindows: [:])
 
-    private init() {}
+    private init() {
+        // This tap sits at the HID stage, ahead of the split into login
+        // sessions, so a session that is switched away still sees the keys of
+        // the account on screen — and this one answers them from state that was
+        // built out of another account's typing, swallowing what it reads as a
+        // bounce. Hand the tap back on resign and build it again from the
+        // preferences on the way in (issue #1075).
+        SessionActivity.shared.onChange { [weak self] _ in
+            self?.syncWithPreferences()
+        }
+    }
 
     func syncWithPreferences() {
-        let nextConfig = KeyboardDebounceConfig(
-            enabled: AppFeature.keyboardDebounce.isAvailable
+        let shouldRun = SessionActivitySupport.tapShouldRun(
+            featureWanted: AppFeature.keyboardDebounce.isAvailable
                 && UserDefaults.standard.bool(forKey: DefaultsKey.keyboardDebounceEnabled),
+            accessibilityGranted: Permissions.shared.accessibility,
+            sessionIsActive: SessionActivity.shared.isActive
+        )
+        let nextConfig = KeyboardDebounceConfig(
+            enabled: shouldRun,
             globalWindowMs: Defaults.sanitizedKeyboardDebounceWindow(
                 UserDefaults.standard.integer(forKey: DefaultsKey.keyboardDebounceWindowMs)
             ),
@@ -44,7 +59,7 @@ final class KeyboardDebounceService: ObservableObject {
             config = nextConfig
         }
 
-        if nextConfig.enabled, Permissions.shared.accessibility {
+        if shouldRun {
             start()
         } else {
             stop()
@@ -109,6 +124,10 @@ final class KeyboardDebounceService: ObservableObject {
 
         if let tap = snapshot.tap {
             CGEvent.tapEnable(tap: tap, enable: false)
+            // Switching a tap off leaves this process owning the port, and the
+            // port is what the window server waits on. Handing the tap back
+            // means leaving the chain, not going quiet in it.
+            CFMachPortInvalidate(tap)
         }
         if let runLoop = snapshot.runLoop {
             CFRunLoopPerformBlock(runLoop, CFRunLoopMode.commonModes.rawValue) {
@@ -212,6 +231,15 @@ final class KeyboardDebounceService: ObservableObject {
         }
     }
 
+    /// Puts the tap back after the window server disabled it, unless a stop is
+    /// already on its way and the port is about to go. Main thread.
+    private func rearmDisabledTap() {
+        let currentTap = lifecycleLock.withLock {
+            shouldStopTapThread ? nil : tap
+        }
+        if let currentTap { CGEvent.tapEnable(tap: currentTap, enable: true) }
+    }
+
     private func publishRunning(_ running: Bool, generation: UInt) {
         let update = { [weak self] in
             guard let self else { return }
@@ -231,10 +259,24 @@ final class KeyboardDebounceService: ObservableObject {
 
     private func handle(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            let currentTap = lifecycleLock.withLock {
-                tap
+            // Keys flowed untapped while the tap was out of the chain, so the
+            // gap must not be read as a quiet stretch and the next press as a
+            // bounce.
+            eventLock.withLock {
+                state.reset()
             }
-            if let currentTap { CGEvent.tapEnable(tap: currentTap, enable: true) }
+            // A tap put straight back is a tap put back into whatever session
+            // is on screen now (issue #1075). The flag that answers that is
+            // written on the main thread, and this callback runs on the tap's
+            // own thread, so the question is asked where the answer lives
+            // rather than read across the two.
+            DispatchQueue.main.async { [weak self] in
+                if SessionActivity.shared.isActive {
+                    self?.rearmDisabledTap()
+                } else {
+                    self?.syncWithPreferences()
+                }
+            }
             return Unmanaged.passUnretained(event)
         }
 
--- a/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
+++ b/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // Copyright (C) 2026 Vorssaint
 
+import ApplicationServices
 import Combine
 import CoreGraphics
 import Foundation
@@ -270,8 +271,11 @@ final class KeyboardDebounceService: ObservableObject {
             // written on the main thread, and this callback runs on the tap's
             // own thread, so the question is asked where the answer lives
             // rather than read across the two.
+            // Accessibility is asked here for the same reason the mouse twin
+            // asks it: this tap modifies events, so a revoked permission has to
+            // end it rather than put it back. The sync then stops it for good.
             DispatchQueue.main.async { [weak self] in
-                if SessionActivity.shared.isActive {
+                if SessionActivity.shared.isActive, AXIsProcessTrusted() {
                     self?.rearmDisabledTap()
                 } else {
                     self?.syncWithPreferences()
--- a/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
+++ b/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
@@ -163,11 +163,8 @@ final class KeyboardDebounceService: ObservableObject {
             }
             guard !shouldStopBeforeCreatingTap else {
                 let shouldRestart = clearEventTapThread()
-                if shouldRestart {
-                    start()
-                } else {
-                    publishRunning(false, generation: generation)
-                }
+                publishRunning(false, generation: generation)
+                if shouldRestart { restartTapOnMain() }
                 return
             }
 
@@ -218,11 +215,8 @@ final class KeyboardDebounceService: ObservableObject {
                 state.reset()
             }
             let shouldRestart = clearEventTapThread()
-            if shouldRestart {
-                start()
-            } else {
-                publishRunning(false, generation: generation)
-            }
+            publishRunning(false, generation: generation)
+            if shouldRestart { restartTapOnMain() }
         }
     }
 
@@ -237,6 +231,18 @@ final class KeyboardDebounceService: ObservableObject {
             pendingStartAfterStop = false
             return shouldRestart
         }
+    }
+
+    /// Runs on the tap thread once its run loop has stopped. A restart that
+    /// `start` left pending goes back through `syncWithPreferences` instead of
+    /// straight to `start`: the session flag is written on the main thread, and
+    /// a tap rebuilt into a session that is no longer on screen stalls the
+    /// account that is (issue #1075). `isRunning` is published false first, so
+    /// the gap between the two is not a green tick over a tap that is down;
+    /// the sync publishes it true again if it starts one. Its twin
+    /// `MouseClickDebounceService.finishEventTapThread` hops the same way.
+    private func restartTapOnMain() {
+        DispatchQueue.main.async { [weak self] in self?.syncWithPreferences() }
     }
 
     /// Puts the tap back after the window server disabled it, unless a stop is
--- a/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
+++ b/Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift
@@ -44,7 +44,14 @@ final class KeyboardDebounceService: ObservableObject {
         let shouldRun = SessionActivitySupport.tapShouldRun(
             featureWanted: AppFeature.keyboardDebounce.isAvailable
                 && UserDefaults.standard.bool(forKey: DefaultsKey.keyboardDebounceEnabled),
-            accessibilityGranted: Permissions.shared.accessibility,
+            // Asked of the system and not of `Permissions.shared`, whose
+            // answer is a poll up to `PermissionPollingSupport.interval` old.
+            // The re-arm hands a revoked tap to this sync to be stopped, and a
+            // stale grant here would leave `shouldRun` true, so `start()` would
+            // see the thread still up, publish running and return — a tap that
+            // is neither back in the chain nor stopped. The mouse twin asks the
+            // same way.
+            accessibilityGranted: AXIsProcessTrusted(),
             sessionIsActive: SessionActivity.shared.isActive
         )
         let nextConfig = KeyboardDebounceConfig(
--- a/Sources/Vorssaint/Services/ShortcutRecordingTap.swift
+++ b/Sources/Vorssaint/Services/ShortcutRecordingTap.swift
@@ -14,7 +14,9 @@ import Foundation
 ///
 /// The tap lives only while a field records, plus the tail of a key still
 /// held when recording ends, so its release and autorepeats cannot reach the
-/// app as a fresh press of the recorded combination. Only one field ever
+/// app as a fresh press of the recorded combination. A fast user switch ends it
+/// too: a tap left behind by a session that is no longer on screen swallows the
+/// keys of the account that is. Only one field ever
 /// records at a time (the ShortcutCapture invariant), so one static tap is
 /// enough. Main thread only. Without Accessibility, begin fails and the
 /// field falls back to plain view events, which is how it always worked.
@@ -33,6 +35,14 @@ enum ShortcutRecordingTap {
     /// super key does lets a field record the combination the way it will be
     /// pressed later, instead of asking for the chosen modifiers by hand.
     private static var superState = SuperKeySupport.State()
+    /// Registered on the first recording and kept for the process. A recorder
+    /// that is still open when this login session stops being the one on screen
+    /// would otherwise keep a tap that swallows every key in the chain the
+    /// account switched in is typing into, and the timeout re-arm below would
+    /// put it back there (issue #1075). Only the command bar's capture card
+    /// reaches that state — the shortcut fields in Settings already stop on
+    /// `didResignActive` — but the tap is shared, so the guard is here.
+    private static var observingSession = false
 
     /// Starts swallowing key events and delivering each fresh press to the
     /// handler. Returns false when the tap cannot exist (no Accessibility),
@@ -50,6 +60,13 @@ enum ShortcutRecordingTap {
         }
         if tap == nil {
             guard AXIsProcessTrusted() else { return false }
+            if !observingSession {
+                observingSession = true
+                SessionActivity.shared.onChange { isActive in
+                    guard !isActive else { return }
+                    tearDown()
+                }
+            }
             let mask = (CGEventMask(1) << CGEventType.keyDown.rawValue)
                 | (CGEventMask(1) << CGEventType.keyUp.rawValue)
             guard let created = CGEvent.tapCreate(
@@ -116,7 +133,13 @@ enum ShortcutRecordingTap {
 
     private static func handle(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            if let tap { CGEvent.tapEnable(tap: tap, enable: true) }
+            if SessionActivity.shared.isActive, let tap {
+                CGEvent.tapEnable(tap: tap, enable: true)
+            } else {
+                // Invalidating the port from its own callback stack is unsafe;
+                // finish this callback fail-open, then release the tap.
+                DispatchQueue.main.async { tearDown() }
+            }
             return Unmanaged.passUnretained(event)
         }
         let keyCode = event.getIntegerValueField(.keyboardEventKeycode)
--- a/Sources/Vorssaint/Services/ShortcutRecordingTap.swift
+++ b/Sources/Vorssaint/Services/ShortcutRecordingTap.swift
@@ -58,15 +58,21 @@ enum ShortcutRecordingTap {
         if let tap, !CGEvent.tapIsEnabled(tap: tap) {
             tearDown()
         }
+        // Registered on the way in, before the tap is asked for: the exit this
+        // installs is the recording's, not the tap's. Without Accessibility
+        // the lines below give up and the caller records through its own
+        // monitor, but `ShortcutCapture.begin()` has already switched the app's
+        // global shortcuts off, so a resign with no handler registered leaves
+        // them off with nothing left to turn them back on.
+        if !observingSession {
+            observingSession = true
+            SessionActivity.shared.onChange { isActive in
+                guard !isActive else { return }
+                endForSessionSwitch()
+            }
+        }
         if tap == nil {
             guard AXIsProcessTrusted() else { return false }
-            if !observingSession {
-                observingSession = true
-                SessionActivity.shared.onChange { isActive in
-                    guard !isActive else { return }
-                    endForSessionSwitch()
-                }
-            }
             let mask = (CGEventMask(1) << CGEventType.keyDown.rawValue)
                 | (CGEventMask(1) << CGEventType.keyUp.rawValue)
             guard let created = CGEvent.tapCreate(
--- a/Sources/Vorssaint/Services/ShortcutRecordingTap.swift
+++ b/Sources/Vorssaint/Services/ShortcutRecordingTap.swift
@@ -64,7 +64,7 @@ enum ShortcutRecordingTap {
                 observingSession = true
                 SessionActivity.shared.onChange { isActive in
                     guard !isActive else { return }
-                    tearDown()
+                    endForSessionSwitch()
                 }
             }
             let mask = (CGEventMask(1) << CGEventType.keyDown.rawValue)
@@ -113,6 +113,20 @@ enum ShortcutRecordingTap {
         DispatchQueue.main.asyncAfter(deadline: .now() + 1.0, execute: watchdog)
     }
 
+    /// The session leaving ends the recording, not only the tap, and the two
+    /// are not the same teardown: whoever called `begin` also called
+    /// `ShortcutCapture.begin()`, which left `HotkeyManager`, Shelf, clipboard,
+    /// sound-output and window-layout shortcuts switched off. Dropping the tap
+    /// alone comes back with the capture card still up and every global
+    /// shortcut in the app dead. Nothing here reopens the card; giving the keys
+    /// back early is the safe direction, and `ShortcutCapture.end()` is
+    /// idempotent, so the caller's own exit still runs unchanged.
+    /// Main thread, like `ShortcutCapture`.
+    private static func endForSessionSwitch() {
+        tearDown()
+        ShortcutCapture.end()
+    }
+
     private static func tearDown() {
         drainWatchdog?.cancel()
         drainWatchdog = nil
@@ -138,7 +152,7 @@ enum ShortcutRecordingTap {
             } else {
                 // Invalidating the port from its own callback stack is unsafe;
                 // finish this callback fail-open, then release the tap.
-                DispatchQueue.main.async { tearDown() }
+                DispatchQueue.main.async { endForSessionSwitch() }
             }
             return Unmanaged.passUnretained(event)
         }
--- a/Sources/Vorssaint/Services/SuperKey/SuperKeyService.swift
+++ b/Sources/Vorssaint/Services/SuperKey/SuperKeyService.swift
@@ -235,7 +235,7 @@ final class SuperKeyService: ObservableObject {
             let runLoop = CFRunLoopGetCurrent()
             lifecycleLock.withLock { tapRunLoop = runLoop }
             guard !lifecycleLock.withLock({ shouldStopTapThread }) else {
-                if clearEventTapThread() { startOnMain() }
+                if clearEventTapThread() { restartTapOnMain() }
                 return
             }
 
@@ -329,7 +329,7 @@ final class SuperKeyService: ObservableObject {
                 if let mouseSource { CFRunLoopRemoveSource(runLoop, mouseSource, .commonModes) }
                 CFMachPortInvalidate(mouseTap)
             }
-            if clearEventTapThread() { startOnMain() }
+            if clearEventTapThread() { restartTapOnMain() }
         }
     }
 
@@ -349,8 +349,14 @@ final class SuperKeyService: ObservableObject {
         }
     }
 
-    private func startOnMain() {
-        DispatchQueue.main.async { [weak self] in self?.start() }
+    /// Runs on the tap thread once its run loop has stopped. A restart that
+    /// `start` left pending goes back through `syncWithPreferences` instead of
+    /// straight to `start`: the hop to the main thread was already here, but
+    /// `start` asks only `AXIsProcessTrusted()`, so a restart that crossed a
+    /// session handover rebuilt both taps into a session that is no longer on
+    /// screen (issue #1075). The sync is the only place the session gate lives.
+    private func restartTapOnMain() {
+        DispatchQueue.main.async { [weak self] in self?.syncWithPreferences() }
     }
 
     private func tapDidStart(_ startedTap: CFMachPort) {
--- a/Sources/Vorssaint/Services/SuperKey/SuperKeyService.swift
+++ b/Sources/Vorssaint/Services/SuperKey/SuperKeyService.swift
@@ -15,7 +15,8 @@ import IOKit.hidsystem
 /// F18; the event tap then keeps that key to itself and adds the user's
 /// chosen modifiers to whatever is pressed while it is down. Nothing is
 /// installed while the feature is off, and the mapping is always taken back
-/// out when the feature goes off or the app quits. Requires Accessibility:
+/// out when the feature goes off, when this login session stops being the one
+/// on screen, or when the app quits. Requires Accessibility:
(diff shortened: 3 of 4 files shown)


PathGao:
The threading argument is the right one: asking on main instead of sampling under the lock is what keeps the answer non-stale against `onChange`.

**`ShortcutRecordingTap` tears down half the recording state.** Every other exit pairs `end()` with `ShortcutCapture.end()`; the new session handler calls the private `tearDown()` alone. `ShortcutCapture.begin()` disabled `HotkeyManager`, Shelf, clipboard, sound-output and window-layout shortcuts, and nothing turns them back on. Switch away with the command bar's capture card open, switch back: card still up, tap gone, every global shortcut in the app dead. Its own doc says shortcuts returning early is the safe direction.

```swift
guard !isActive else { return }
tearDown()
ShortcutCapture.end()
```

**The debounce re-arm didn't follow its twin.** `MouseClickDebounceService.handle` gates on `config.enabled && isActive && AXIsProcessTrusted() && !stopping`; `rearmDisabledTap` asks `isActive` and `!shouldStopTapThread`. This is a modifying tap too, and the dropped `AXIsProcessTrusted()` is what the test asserts for the other four.

Not verified: `hidutil` clear timing on the async `stop()`, and no second login session here either.

**Impact:** app-wide global shortcuts stay dead after a switch away with the command bar capture card open.
**Merge:** needs the `ShortcutCapture.end()` pairing first.

ref #1075
ref #1153

PathGao:

PathGao:
@PathGao Right on both counts, and the first one was a real hole: the session handler dropped the tap and left `ShortcutCapture` mid-capture, so the app came back with `HotkeyManager`, Shelf, clipboard, sound output and the window-layout shortcuts all still switched off and nothing left to switch them on.

I went looking for every begin and every end rather than patching the one line. There are two begins — `CommandBarService.beginCapturingShortcut` and `ShortcutRecorderButton.beginRecording` — and five ends: `endCapturingShortcut`, `stopRecording`, the button's `deinit` (both branches), and the two teardowns inside the tap itself. The five external ones already pair `ShortcutRecordingTap.end()` with `ShortcutCapture.end()`; the two internal ones did not. Both are session-driven — the `onChange` handler you quoted, and the `tapDisabledByTimeout` branch that declines to re-arm into a switched-away session and instead tears the tap down on the main queue. That second one has the same consequence and the same fix, so both now go through one `endForSessionSwitch()` that calls `tearDown()` and then `ShortcutCapture.end()`. Nothing else changed: `begin`'s rebuild of a dead tap and the drain watchdog still call `tearDown()` alone, because neither of those is the recording ending.

The card is deliberately left on screen. `ShortcutCapture.end()` is idempotent, so whichever ordinary exit the user reaches afterwards is a no-op, and the card's local-monitor fallback still records — the same fallback that covers a machine without Accessibility. Hiding the command bar panel on resign is a command-bar behaviour and I did not want it inside a tap; it is called out in the body.

Second point taken as written: the re-arm now asks `SessionActivity.shared.isActive, AXIsProcessTrusted()`, and the else branch already re-syncs, which stops the tap. That needed `import ApplicationServices`, which the file did not have. The test asserts it on the re-arm slice rather than on the whole file, so a mention somewhere else cannot answer for it.

Four assertions cover the pairing (one exit exists, it calls both halves, and each session path reaches it) plus one for the re-arm, all in the existing session-switch guard in `Tests/MetricsTests.swift`, with lines trimmed so indentation is not pinned. Confirmed red before green: restoring the previous `ShortcutRecordingTap.swift` turns all four capture assertions red at once, and removing `AXIsProcessTrusted()` turns the fifth red on its own; both go green again on restore. `./build.sh --test` is 9484/9486, the two failures being the pre-existing input-source ones on this machine. The full `./build.sh` also ran on this branch and exited 0 with the bundle produced, which is what compiles these services at all; I only kept the tail of that log, so I am not repeating the earlier zero-warnings claim on my own run, and its fan-helper codesign step hit `errSecInternalComponent` against this machine's self-signed identity.

Still not verified, same as before: no second login session here, so none of this was reproduced. I read the behaviour off `ShortcutCapture`'s callers, not off a switch. Your `hidutil` clear-timing note on the async `stop()` is also still open — I did not touch that path.

PathGao:
Both fixes land: one `endForSessionSwitch()` for both session paths is the right shape, and the assertions were checked red-first.

**The re-arm's `AXIsProcessTrusted()` has no follow-through.** Body says the else branch "re-syncs, which stops it" — it doesn't. `syncWithPreferences` passes `accessibilityGranted: Permissions.shared.accessibility`, a 60s-polled cache (`PermissionPollingSupport.interval`); the twin passes `AXIsProcessTrusted()`. Inside that window `shouldRun` is still true, so sync calls `start()`, which sees `tapThread != nil` and returns after publishing running. The disabled tap is neither re-armed nor stopped.

```swift
accessibilityGranted: AXIsProcessTrusted(),
```

**`SuperKeyService.rearmDisabledTaps` didn't follow.** It asks the session only. `start()` has the AX guard, but the re-arm never reaches `start()` — same modifying `.defaultTap`, same PR. Not in the re-arm-slice assertion either, so nothing catches it.

Not verified: no second session here, and whether the window server lets a tap re-enable after AX is revoked is your argument, not something I checked.

**Impact:** after Accessibility is revoked mid-session, both taps stay disabled while `isRunning` publishes true.
**Merge:** needs both AX gaps closed first.

ref #1075
ref #1153

PathGao:
@PathGao Both gaps were real and they were the same gap twice: the round-one change taught two re-arms to decline, and neither decline had anything downstream that could actually end the tap.

`KeyboardDebounceService` now passes `AXIsProcessTrusted()` to `tapShouldRun` instead of `Permissions.shared.accessibility`. Your read of the failure is exactly what the code does: inside the 60s poll window `shouldRun` stays true, `start()` finds `tapThread != nil`, publishes running and returns, so the disabled tap is neither re-armed nor stopped. With the system asked directly the sync reaches `stop()`, which publishes not-running.

`SuperKeyService.rearmDisabledTaps` now asks `AXIsProcessTrusted()` alongside the session. The else branch there works out on its own: `syncWithPreferences` already treats a tap failing `CGEvent.tapIsEnabled` as dead, so it stops and calls `start()`, whose Accessibility guard clears the leftover `hidutil` mapping and sets `isRunning = false`. No new stop path was needed.

I checked `isRunning`'s readers before routing the fixes through the stop paths rather than just widening the guards. It has three live `@ObservedObject` bindings — `KeyboardDebounceSettings`, `SuperKeySettings`, and `ShortcutsSettings`, where `superKey.isRunning` decides whether the shortcut rows offer the Super Key alternative. A stale `true` is a green "Active now" over a dead tap and an alternative offered for a tap that is not running, which is the visible half of your impact line. Nothing new writes `isRunning`; the existing `stop()`/`start()` publishes are correct once the gates are.

Same shape elsewhere: I went through every `SessionActivitySupport.tapShouldRun` caller. `SmoothScrollService`, `MouseNavigationService`, `MouseButtonShortcutService`, `MiddleClickService`, `QuitProtectionService` and `MouseClickDebounceService` all ask the system. `ScrollInverter` is the one left on the polled cache, and its re-arm gates on the session without asking Accessibility at all. That landed in #1145 with its own review and is not in this branch's diff, so I left it and called it out in **Anything else** rather than pulling a third feature into this PR — say the word if you want it here instead.

Two assertions, both checked red first and on their own. Putting `accessibilityGranted: Permissions.shared.accessibility` back turns only "the keyboard debounce gate asks the system for Accessibility, not the poll cache" red; dropping `AXIsProcessTrusted()` from the Super Key re-arm turns only "SuperKeyService.swift does not put a modifying tap back after Accessibility is lost" red. The debounce one is pinned on the `syncWithPreferences` slice and the Super Key one on the re-arm slice, so a mention elsewhere in either file cannot answer for them. `./build.sh --test` is 9486/9488 on this head; the same two assertions fail on the branch's base `16216d56` in a throwaway worktree (9467/9469) because they read this machine's live input source under a Chinese layout, so that is the machine and not the tree.

Not verified, and I have corrected the body to say so: I did not run the full `./build.sh` for this revision, so neither service file was compiled locally — CI covers both runners. Still no second login session here, and I did not revoke Accessibility by hand; the app was never launched. Your point about whether the window server lets a tap re-enable at all after a revoke is still untested — the change makes the app stop asking, which is the right answer either way. Your `hidutil` clear-timing note on the async `stop()` is also still open; I did not touch that path.

PathGao:
Both gates land, and routing them through `stop()` instead of just widening the guard is right — `isRunning`'s readers are why.

**The session half of the sync gates isn't pinned, and for Super Key that's the whole first-order fix.** The guard asserts the re-arm slice in both files and never looks at the sync. Drop `&& SessionActivity.shared.isActive` from `SuperKeyService.syncWithPreferences` and everything stays green while resign → sync → `stop()` is gone: `enabled` is true, no tap reads dead, `start()` returns on `tapExists`, both taps keep their place. The debounce gate slice has the same hole one step down — it pins `accessibilityGranted:` and not `sessionIsActive:`. One line each, beside the assertions you added:

```swift
expect(keyboardDebounceGate.contains("sessionIsActive: SessionActivity.shared.isActive"), …)
```

Rest checks out: the five ordinary capture exits do pair both halves, the twin invalidates in the same two places, and `ShortcutCapture.end()` guards on `isCapturing`, so the permanent resign handler is a no-op outside a recording.

Not verified: no second session, nothing run — read off `main` plus the diff.

**Impact:** none today; both gates are correct as written.
**Merge:** needs the Super Key session conjunct pinned first.

ref #1075
ref #1153

PathGao:
@PathGao Right, and the Super Key one was the worse half: the round-two guard
asserted the re-arm slice in both files, and the re-arm is not what ends a tap
on a resign. The sync is. For Super Key the session conjunct in `enabled` is the
entire resign path — without it nothing reads dead, `start()` returns on
`tapExists`, both taps keep their place and the `hidutil` mapping stays written
for the whole Mac.

Rather than paste the two lines I went through every gate this branch touches or
claims to cover. The debounce gate depends on both halves, so it now pins
`accessibilityGranted: AXIsProcessTrusted()` and
`sessionIsActive: SessionActivity.shared.isActive` on the `syncWithPreferences`
slice. Super Key's gate depends on the session only — Accessibility lives in
`start()`, which this branch does not touch — so the `&& SessionActivity.shared.isActive`
conjunct is pinned on the `let enabled` slice, up to `guard enabled else`, so the
re-arm's own call to the same property elsewhere in the file cannot answer for
it. `ShortcutRecordingTap` is the third file and has no sync at all: its resign
path is the `onChange` handler and its timeout branch, and both were already
pinned last round, so there is no third hole to close. The six older tap owners
in the same loop reach `tapShouldRun` and are covered by the earlier PRs' own
tests; `ScrollInverter` is still the odd one out and is still called out in
**Anything else** rather than pulled in here.

Red-first on both, each on its own, and by removing the real conjunct rather
than misspelling it. Deleting `&& SessionActivity.shared.isActive` from
`SuperKeyService.syncWithPreferences` — your break, verbatim — turns only "the
Super Key sync gate stops both taps when the session leaves" red, and every
round-one and round-two assertion stays green under it, which is the proof the
addition was needed. Passing `sessionIsActive: true` in the debounce gate turns
only "the keyboard debounce gate stops the tap when the session leaves" red.
Both go green on restore. `./build.sh --test` is 9488/9490 on this head, the two
failures being the input-source assertions that also fail on the base
`16216d56` (9467/9469) under this machine's Chinese layout.

No source changed this round, only `Tests/MetricsTests.swift`, so nothing new
needs compiling; neither service file is in the `--test` list and I did not run
the full `./build.sh`, so CI is what compiles them. Still no second login
session here and the app was never launched, so none of this was reproduced, and
your `hidutil` clear-timing note on the async `stop()` is still open.

PathGao:
Both slices are cut right — `let enabled` to `guard enabled else` cannot be answered by the re-arm's own call.

**`SuperKeyService.stop()` disables both taps without invalidating them.** `KeyboardDebounceService` got both halves this round to match the mouse twin; Super Key has only the one at `runEventTap`'s tail. This PR is what first routes a resign into that `stop()`, so the branch's own argument — off is not out of the chain — now applies to a path it does not cover. The guard's file-level `CFMachPortInvalidate` check is answered by that tail.

**Without Accessibility the capture card has no exit at all.** `beginCapturingShortcut` calls `ShortcutCapture.begin()`, then `ShortcutRecordingTap.begin`, which returns on `guard AXIsProcessTrusted()` before it reaches `observingSession`. No tap, no observer, and Settings' `didResignActive` does not cover the command bar — switch away mid-capture without the grant and every global shortcut stays off. Round one's symptom through the other door. Registering above `if tap == nil` closes it; `tearDown()` and `ShortcutCapture.end()` are both safe with no tap.

Not verified: no second session, nothing run.

**Impact:** on a Mac without Accessibility, switching away with the capture card open leaves every global shortcut dead.
**Merge:** needs the no-Accessibility capture exit first.

ref #1075
ref #1153

PathGao:
Both gates are pinned on the right slice, and the slicing fails closed — I traced `let enabled` → `guard enabled else` and `syncWithPreferences()` → `let nextConfig` against the trimmed source; a reorder empties the slice rather than passing it.

Checked the two claims I'd left unverified:

- The capture card's fallback is real. `CommandBarService`'s `keyMonitor` matches `.capturingShortcut` and calls the same `handleCaptureKey`, so a card left up is the no-Accessibility state, not a dead card.
- Giving the keys back on resign puts no tap back. `ShortcutCapture.end()` → `FeatureRuntime.sync(featuresToSilenceWhileRecording)` reaches `WindowLayoutService.suspendShortcuts()`, which is `unregisterHotkeys()` — Carbon, not a tap — and every other binding in that list is a session-gated sync that stops.

One correction: `FocusFollowsMouseService` is an eighth `tapShouldRun` caller your list of seven missed. It already asks the system and owns no tap, so nothing to do; the count was just short.

Nothing left worth a round five.

Not verified: no second session, nothing run. Both build checks were still pending when I looked, and the `hidutil` clear on `stop()` is still async.

**Impact:** none.
**Merge:** ship it.

ref #1075
ref #1153

PathGao:
@PathGao Confirmed and fixed in `b7f9dc6e`. This is the no-Accessibility capture exit from your 10:04 review — the second of the two you raised there. The later "ship it" was answering the round before it.

Read the path before moving the line. `begin` resets the drain state, then enters `if tap == nil`, where `guard AXIsProcessTrusted() else { return false }` came first and the `observingSession` registration second. Without the grant the guard returns and the registration never runs, but both callers — `CommandBarService.beginCapturingShortcut` and `ShortcutRecorderButton.beginRecording` — have already called `ShortcutCapture.begin()`, which sets `isCapturing` and switches off `HotkeyManager`, Shelf, clipboard, sound output and the window-layout shortcuts. The card still records: `CommandBarService`'s `keyMonitor` matches `.capturingShortcut`, which is the fallback that file's comment promises. So nothing is visibly wrong until the session resigns — at which point no handler exists, `ShortcutCapture` stays mid-capture, and every global shortcut in the app stays off. Settings' fields survive this on `didResignActive`; the command bar's card has no equivalent.

On your other question, what happens after the grant arrives: the registration is `observingSession`-guarded and kept for the process, so the first recording that actually creates a tap installs it and every switch after that is covered. The hole is the stretch before that first successful `begin`, not a permanent state — which is why it reads as an ordering bug rather than a missing feature. I put that in the commit message rather than only "as reviewed".

Fix is the placement you suggested: the registration moved above `if tap == nil`, so it runs on every `begin` regardless of the grant. Both halves of `endForSessionSwitch()` are safe with no tap — `tearDown()` touches nil state, and `ShortcutCapture.end()` guards on `isCapturing`, so outside a recording the handler is a no-op, which is the same reason the permanent handler was already safe.

Left behind: one assertion sliced to the part of `begin` above `guard AXIsProcessTrusted()`, asserting `SessionActivity.shared.onChange` appears there, so the tap-creation path cannot answer for it. Verified both ways — red with the registration back below the guard, green with it above, and it is the only assertion that moves. `./build.sh --test`: `TESTS FAILED (2 of 9491)`, the same two input-source checks that fail on the base under this machine's Chinese layout (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints`), no third. `ShortcutRecordingTap.swift` is outside the `--test` whitelist, so I also ran `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` with the build's own `arm64-apple-macosx14.0` / `MacOSX26.sdk` flags and the two `-I` compat paths: clean.

Not verified: no second login session here, the app was never launched, and I did not revoke Accessibility by hand. The path is read, not run.

Left open on purpose, since you raised it in the same comment: `SuperKeyService.stop()` disabling both taps without invalidating them is not in this commit. It is the other half of that review and a different file; say the word if you want it here rather than on its own.

Worth re-checking after this: the registration now fires on `ShortcutRecorderButton.beginRecording` too, on a Mac with no grant, where Settings' `didResignActive` already ends the recording. Both exits are idempotent so the pair is a no-op, but it is the one new interaction this ordering creates.

Refs #1075
Refs #1153

PathGao:
@PathGao Rebased onto `main` at `a952b829`, head is now `6d6325c2`. No source
change of my own — the rebase is the whole of it.

**What the rebase took.** #1225 landed the `CFMachPortInvalidate` after the run
loop returns in `KeyboardDebounceService`, so that hunk resolved away. The one
in `stop()` survives and is not a duplicate: it is the pair
`MouseClickDebounceService` has, so a teardown that reaches the port before the
thread does hands it back too. `KeyboardDebounceService` is now 2 invalidations
on 1 tap and `SuperKeyService` 2 on 2, which is what the repo-wide sweep
expects. The body was corrected — it claimed both lines, and only one is this
branch's now.

**The rejected option in the trade-offs is now the answer across the four
branches.** #1234 sampled the gate under a lock; that has moved to the
main-thread hop this branch chose, because the precedent it cited
(`BrightnessService`) was withdrawn from #1231. #1236 re-armed provisionally
on the tap thread and then corrected on main; that has moved too, because
putting the tap back first is putting it back into the switched-away session
for as long as the correction takes.

So the six own-thread taps — `FinderCutPaste`, `FinderRenameService`,
`KeyboardDebounceService`, `SuperKeyService`, `TextSnippetService`,
`AppSwitcher` — all leave the tap out of the chain and ask on main, and the
six main-run-loop taps read the flag in place because their callback is
already on the thread that writes it. One rule: the decision is made where the
answer lives. `MouseClickDebounceService` is the one file still reading it
across threads; it is on `main`, not in any of these diffs, and I have not
touched it.

Same for the permission source: `AXIsProcessTrusted()` at every gate across
the four branches now. This one already did.

**Tests.** `./build.sh --test` → `TESTS OK (9555 checks)`. The two
input-source failures I reported on earlier revisions did not appear this
round; this machine's input source is ANSI now rather than the Chinese IME.
`swiftc -typecheck` over all of `Sources/Vorssaint/**` is clean; none of the
three files is in the `--test` whitelist. Full `./build.sh` not run — signing
keychain locked, `codesign` blocks on the unlock dialog.

**Where to look first.** `MetricsTests` now carries two overlapping conditions
for the same rule: the older `if tapOwner.contains("MouseNavigation") || …`
requiring `AXIsProcessTrusted()` somewhere in the file, and this branch's
`if tapOwner.contains("KeyboardDebounce") || tapOwner.contains("SuperKey")`
requiring it in the re-arm slice. The second is strictly stronger. I left both
rather than restructure a block three other open branches are also editing,
but #1230 replaces that roster wholesale, so it may be worth collapsing them
there instead of here.

**Still open in these files, not fixed here.** `SuperKeyService`'s tap-thread
tail restarts with `if clearEventTapThread() { startOnMain() }` and
`KeyboardDebounceService`'s with a bare `start()`, both of which rebuild the
tap without asking the gate when a stop and a start cross at handover time.
#1236 fixes that shape in its own two files and names these as not its to
fix. Two one-line changes if you want them here.

PathGao:
@PathGao Answering your call on the restart path — both tails are gated now, in
`10fafdce`. Not a force-push; one commit on top of what you last saw.

**I checked each one was reachable before touching it, and they were not equally
broken.**

`KeyboardDebounceService` called `start()` straight from the tap thread. Both halves
wrong: the thread and the question. It hops and syncs now.

`SuperKeyService` was the one worth checking rather than assuming. `startOnMain()` was
already `DispatchQueue.main.async { start() }` — the hop was there. What was missing is
the question: `start()` guards on `AXIsProcessTrusted()` alone, and the
`&& SessionActivity.shared.isActive` conjunct you pinned lives in `syncWithPreferences`,
which the restart skipped. So a restart that crossed a handover rebuilt **both** taps in a
switched-away session with the grant still in place — the common case, not the revoked
one. Renamed `restartTapOnMain`, because "on main" was never the part that was missing.

The window is the same one in all six files, and the same one you already accepted the fix
for in `AppSwitcher`: `stop()` clears the pending flag under the lock, so the gap is
between `clearEventTapThread()` reading `true` and the restart running. A resign landing
there finds the thread already gone, resets `shouldStopTapThread`, and the tail builds a
fresh tap into the session that has just left. The `tapCreate` failure path
(`_ = clearEventTapThread()`) drops the pending restart deliberately and is untouched.

**One behaviour change beyond the hop, in the debounce service.** `isRunning` was
published false only on the no-restart branch. With the restart deferred there is now a
gap where the tap is down, so it is published false before the hop and the sync publishes
it true again if it starts one. That is your own argument from the earlier round — a stale
`true` is a green "Active now" tick over a dead tap — applied to the gap this change
creates. `publishRunning` already drops a stale generation and the two main-queue blocks
run in order, so nothing else was needed.

**One of the two is not pinned, and I did not reshape the code to pin it.**
`MetricsTests` asks the restart question of every roster owner that runs a tap thread
again. It catches `SuperKeyService` — pointing its helper back at `start()` turns it red on
its own. It does **not** reach `KeyboardDebounceService`'s tail, which branches on a local
(`let shouldRestart = clearEventTapThread()` … `if shouldRestart { … }`) instead of
spelling the call inline the way the other five do. I reverted both of its tails to
`start()` with the helper left in place and the suite stayed green, so this is measured,
not assumed. Pinning it would mean matching a variable name and its indentation, or
reshaping the tail to suit the check. Named in the body instead.

`./build.sh --test` → `TESTS OK (9559 checks)`, no failures; the two input-source ones are
still absent, this machine is on ANSI. `swiftc -typecheck` over all of
`Sources/Vorssaint/**` clean. Full `./build.sh` not run — signing keychain locked.

**Where to look first.** `SuperKeyService`'s restart now goes through the full sync rather
than `start()`, which means it re-reads the source key, the modifiers and the solo action,
and can take the `sourceChanged` branch into `stop()` before starting. That is strictly
more correct after a handover, but it is more work than the old restart did and it is the
one place where routing through the sync is not just a gate check.

PathGao:
Renaming it `restartTapOnMain` was right — the hop was never the missing half.

**The sync's `deadTap` branch fires again on the restart.** `sourceChanged`, the one you named, can't fire: `self.source` was already stored by the sync that set `pendingTapRestart`, so the restart sync reads it equal. `mouseTapRefusals` can — it deliberately outlives `clearEventTapThread`, so with `mouseTap == nil` the restart sync computes `mouseDead = (mouseTapRefusals == 1)` and takes `stop()` before `start()`: hidutil mapping pulled and re-applied, wake observer dropped, `isRunning` blipped. That is the condition that caused the restart in the first place.

**The debounce's new `publishRunning(false)` cannot land.** A pending restart exists only because `stop()` ran, and `stop()` bumps `lifecycleGeneration` and publishes false itself, inline on main. The tail's generation is stale by construction, so the publish is dropped — the gap it describes was already closed. The twin spends that generation on the other side: `finishEventTapThread` checks it before syncing.

Not verified: no second session, nothing run.

**Impact:** after a single mouse-tap refusal, the Super Key mapping is pulled and re-applied once more on the rebuild.
**Merge:** ship it.

ref #1075
ref #1153

PathGao:

PathGao:
Reopened as #1304.

PR #1236, opened 2026-09-01
fix: hand the switcher, snippet, volume and Dock taps back across a user switch

PathGao:
## Summary

Four `.defaultTap` event taps stayed in the event chain after a fast user
switch. A filter tap owned by a switched-away login session keeps its place in
the chain, so the window server still routes the front account's events through
a process that cannot answer and waits out the ~4s tap timeout on each one
(#1075). `AppSwitcher`, `TextSnippetService`, `PreciseVolumeRollerService` and
`DockClickService` never subscribed to `SessionActivity`.

Each service now does what `MouseNavigationService` does: subscribe to
`SessionActivity.shared.onChange` in `init`, route its preference sync through
`SessionActivitySupport.tapShouldRun`, and ask the same question before a tap
the window server disabled goes back into the chain. The
`CFMachPortInvalidate` calls this branch carried are on `main` through #1225
and are gone from the diff.

**Where the session is asked.** Two of the four run their tap on their own
thread (`AppSwitcher`, `TextSnippetService`) because a `keyDown` tap on the main
run loop turns any main-thread stall into system-wide keyboard lag (#275).
`SessionActivity.isActive` is written on the main thread, so those two must not
read it from the tap callback. An earlier revision re-armed provisionally under
the lifecycle lock and then corrected on the main thread; that is the one shape
that cannot hold, because putting the tap back first is putting it back into
whatever session is on screen now — the stall this branch exists to end — for as
long as the main queue takes to correct it. Both now leave the tap out of the
chain, hop to the main queue, and either re-arm there or hand the tap to the
sync to be stopped. That is what the keyboard, Super Key and Finder taps do in
the same position. `PreciseVolumeRollerService` and `DockClickService` run on
the main run loop, so their callback is already on the thread that writes the
flag and they read it in place.

Sampling the answer into a lock-protected field, the way #1234 did, was the
other option and is not taken anywhere now: that precedent was
`BrightnessService`'s flag, which has since been withdrawn from #1231, and the
sample is a second copy of state the process already has.

**Which Accessibility.** All four gates ask `AXIsProcessTrusted()`. It was mixed
before — `PreciseVolumeRollerService` asked the system while `AppSwitcher`,
`DockClickService` and `TextSnippetService` read `Permissions.shared`, whose
answer is a poll up to `PermissionPollingSupport.interval` old. The re-arm hands
a tap it declines to put back to the sync to be stopped, so a stale grant there
installs it straight back. The cached mirror stays on the per-key paths in
`AppSwitcher.handle` and `beginPendingSession`, where the comments explain that
a live TCC round-trip per keystroke is what they avoid.
`PreciseVolumeRollerService.start()` loses its own `AXIsProcessTrusted()` guard:
the gate asks that now, and the not-granted path lands on the same `stop()` the
guard was calling by hand.

**The tap-thread restart.** Those same two services restart their tap thread
from the thread's own tail. When a stop and a start cross,
`clearEventTapThread()` reports the pending start and the tail used to rebuild
the tap on the spot — `installTap()` in the switcher, `start()` in the snippet
service — without going through the gate. If the session hands over while that
restart is in flight, the resign handler's teardown finds the thread already
gone and no-ops, and the tail then installs a fresh filter tap into a session
that is no longer on screen. Both tails now hop to the main thread and call
`syncWithPreferences()`, which is the shape
`MouseClickDebounceService.finishEventTapThread` uses. The third
`clearEventTapThread()` call in each file, on the `tapCreate` failure path,
deliberately drops the pending restart and is unchanged — none of these services
retries a failed creation.

**`AppSwitcher` after #882.** That branch replaced `recoverTapIfNeeded` with
`reconcileTakeover`, and the post-wake recheck is still the second place that
can install the tap, so it asks the same gate. Both readers now go through one
private property rather than each spelling the condition out. It is named
`tapIsWanted`, not `tapShouldRun`: it *calls*
`SessionActivitySupport.tapShouldRun`, and #1234 used that same name for a
stored sample, so one name meant three things across three open branches.

`Tests/MetricsTests.swift` gains the four paths in the existing session-switch
roster, which pins the four-part wiring as source shape, plus a restart-tail
check. That check asks every owner in the roster that runs a tap thread, so a
restart added to another gated owner has to answer the same question. The four
other own-thread taps that restarted the old way have since been fixed on the
branches that own them — `FinderCutPaste` and `FinderRenameService` in #1234,
`SuperKeyService` and `KeyboardDebounceService` in #1235 — so whichever of the
three lands first, the roster only ever holds files that already answer it.
`MouseClickDebounceService` is already gated and its tail already hops.

PR #1230 replaces that hand-written roster with a sweep of every file containing
`CGEvent.tapCreate`; these four files are in its `knownUngatedTaps` exemption
table and those entries should be dropped when the two land together.

## Verification

```
./build.sh --test      TESTS OK (9553 checks)
```

None of the four service files is in the `--test` compile whitelist, so a green
`--test` never compiled them. The whole of `Sources/Vorssaint/**` was
type-checked separately with the build's own target and SDK
(`arm64-apple-macosx14.0`, `MacOSX26.sdk`, plus the two `-I` compat paths) and
is clean. The full `./build.sh` was not run this round: with the signing
keychain locked, `codesign` blocks on an unlock dialog. On an earlier revision
it compiled the app and the fan helper with zero warnings and then failed at the
signing step with `errSecInternalComponent`, so the compile is verified and the
bundle step is not.

The two input-source failures reported on earlier revisions (`nil layout falls
back to ANSI semicolon`, `App Switcher icon-row hints`) did not appear this
round; the machine's input source is ANSI now rather than a Chinese IME.

Both guardrails were checked by breaking them. For the wiring: removing the
session read from `AppSwitcher`'s timeout branch and the `onChange`
subscription from `DockClickService` turned the run red with those named
failures, and green again on restore. For the restart tails: putting
`installTap()` back in `AppSwitcher`'s two tails turned the run red with
`AppSwitcher.swift has no tap-thread restart that goes around the session gate`,
and green again on restore.

**Not tested:** the behaviour itself. There is no second login session on this
machine, so nothing here was observed under a real fast user switch — the
mechanism is inherited from #1075 and #1145, not measured. In particular the
restart-tail race is a read of the lifecycle code, not a reproduction: it needs a
stop and a start to cross at the moment of a session handover. Snippet expansion,
Dock click actions, the volume roller and the switcher were not exercised
interactively either; the tap lifecycle is what changed, so the ordinary paths
should be unaffected, but that is an argument, not a run. The revoke race is
not reproduced either: reaching it needs the window server to disable one of
these taps inside the polling interval that follows a revoked grant.

## Anything else

Refs #1153
Refs #1075

`Sources/Vorssaint/Services/DockClick/DockClickService.swift` was also touched by
#1219 and `Sources/Vorssaint/Services/Switcher/AppSwitcher.swift` by #882; both
have landed and this branch is rebased on them.

### DIFF
--- a/Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift
+++ b/Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift
@@ -2,10 +2,12 @@
 // Copyright (C) 2026 Vorssaint
 
 import AppKit
+import ApplicationServices
 import Combine
 
 /// Turns coarse hardware volume wheel bursts into macOS' fine volume step.
-/// Active only while enabled and Accessibility is granted.
+/// Active only while enabled, Accessibility is granted and this login session
+/// is the one on screen.
 final class PreciseVolumeRollerService: ObservableObject {
     static let shared = PreciseVolumeRollerService()
 
@@ -15,12 +17,28 @@ final class PreciseVolumeRollerService: ObservableObject {
     private var source: CFRunLoopSource?
     private var gate = PreciseVolumeRollerGate()
 
-    private init() {}
+    private init() {
+        // A filter tap owned by a switched-away login session keeps its place
+        // in the chain, so the account on screen waits out the tap timeout on
+        // every volume key it presses (issue #1075). Hand the tap back on
+        // resign and build it again from preferences on the way in.
+        SessionActivity.shared.onChange { [weak self] _ in
+            self?.syncWithPreferences()
+        }
+    }
 
     func syncWithPreferences() {
         let wanted = AppFeature.mixer.isAvailable
             && UserDefaults.standard.bool(forKey: DefaultsKey.preciseVolumeRollerEnabled)
-        wanted ? start() : stop()
+        if SessionActivitySupport.tapShouldRun(
+            featureWanted: wanted,
+            accessibilityGranted: AXIsProcessTrusted(),
+            sessionIsActive: SessionActivity.shared.isActive
+        ) {
+            start()
+        } else {
+            stop()
+        }
     }
 
     func suspend() {
@@ -44,11 +62,6 @@ final class PreciseVolumeRollerService: ObservableObject {
     }
 
     private func start() {
-        guard AXIsProcessTrusted() else {
-            removeTap()
-            tapFailed = false
-            return
-        }
         if let tap, !CGEvent.tapIsEnabled(tap: tap) {
             CGEvent.tapEnable(tap: tap, enable: true)
         }
@@ -85,8 +98,12 @@ final class PreciseVolumeRollerService: ObservableObject {
 
     private func handle(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            if AXIsProcessTrusted(), let tap {
+            if AXIsProcessTrusted(), SessionActivity.shared.isActive, let tap {
                 CGEvent.tapEnable(tap: tap, enable: true)
+            } else {
+                // Invalidating the port from its own callback stack is unsafe;
+                // finish this callback fail-open, then release the tap.
+                DispatchQueue.main.async { [weak self] in self?.syncWithPreferences() }
             }
             return Unmanaged.passUnretained(event)
         }
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -139,7 +139,7 @@ final class DockClickService {
 
     private func handle(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            if SessionActivity.shared.isActive, let tap {
+            if AXIsProcessTrusted(), SessionActivity.shared.isActive, let tap {
                 CGEvent.tapEnable(tap: tap, enable: true)
             } else {
                 // Invalidating the port from its own callback stack is unsafe;
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -8,7 +8,8 @@ import CoreGraphics
 
 /// Adds optional actions when the active app's Dock icon is clicked: minimize
 /// its windows, hide the app, or cycle its windows. The Dock's native behavior
-/// remains untouched for every other click. Requires Accessibility.
+/// remains untouched for every other click. Requires Accessibility, and the
+/// tap only exists while this login session is the one on screen.
 final class DockClickService {
     static let shared = DockClickService()
 
@@ -60,15 +61,26 @@ final class DockClickService {
     private static let syntheticEventMarker: Int64 = 0x564F5253
     private var pendingSweeps: [pid_t: DispatchWorkItem] = [:]
 
-    private init() {}
+    private init() {
+        // A filter tap owned by a switched-away login session keeps its place
+        // in the chain, so every click the account on screen makes waits out
+        // this tap's timeout (issue #1075). Hand the tap back on resign and
+        // build it again from preferences on the way in.
+        SessionActivity.shared.onChange { [weak self] _ in
+            self?.syncWithPreferences()
+        }
+    }
 
     func syncWithPreferences() {
         let minimizeEnabled = UserDefaults.standard.bool(forKey: DefaultsKey.dockClickMinimize)
         let hideEnabled = UserDefaults.standard.bool(forKey: DefaultsKey.dockClickHide)
         let cycleEnabled = UserDefaults.standard.bool(forKey: DefaultsKey.dockClickCycleWindows)
-        if AppFeature.dockClick.isAvailable,
-           (minimizeEnabled || hideEnabled || cycleEnabled),
-           Permissions.shared.accessibility {
+        if SessionActivitySupport.tapShouldRun(
+            featureWanted: AppFeature.dockClick.isAvailable
+                && (minimizeEnabled || hideEnabled || cycleEnabled),
+            accessibilityGranted: Permissions.shared.accessibility,
+            sessionIsActive: SessionActivity.shared.isActive
+        ) {
             start()
         } else {
             stop()
@@ -123,7 +135,13 @@ final class DockClickService {
 
     private func handle(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            if let tap { CGEvent.tapEnable(tap: tap, enable: true) }
+            if SessionActivity.shared.isActive, let tap {
+                CGEvent.tapEnable(tap: tap, enable: true)
+            } else {
+                // Invalidating the port from its own callback stack is unsafe;
+                // finish this callback fail-open, then release the tap.
+                DispatchQueue.main.async { [weak self] in self?.syncWithPreferences() }
+            }
             return Unmanaged.passUnretained(event)
         }
         // The replayed down of a press that became a drag: the Dock must
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -75,10 +75,14 @@ final class DockClickService {
         let minimizeEnabled = UserDefaults.standard.bool(forKey: DefaultsKey.dockClickMinimize)
         let hideEnabled = UserDefaults.standard.bool(forKey: DefaultsKey.dockClickHide)
         let cycleEnabled = UserDefaults.standard.bool(forKey: DefaultsKey.dockClickCycleWindows)
+        // Accessibility is asked of the system and not of `Permissions.shared`,
+        // whose answer is a poll up to `PermissionPollingSupport.interval` old.
+        // The re-arm hands a tap it declines to put back to this sync to be
+        // stopped, and a stale grant here would install it straight back.
         if SessionActivitySupport.tapShouldRun(
             featureWanted: AppFeature.dockClick.isAvailable
                 && (minimizeEnabled || hideEnabled || cycleEnabled),
-            accessibilityGranted: Permissions.shared.accessibility,
+            accessibilityGranted: AXIsProcessTrusted(),
             sessionIsActive: SessionActivity.shared.isActive
         ) {
             start()
--- a/Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
+++ b/Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
@@ -8,7 +8,8 @@ import CoreGraphics
 /// Text snippets: typing a trigger replaces it with its expansion, with
 /// {{date}}, {{time}}, {{datetime}} and {{clipboard}} filled in. The key tap,
 /// the observers and the snippet cache only exist while the feature is on;
-/// off means nothing lives. Requires Accessibility (the tap).
+/// off means nothing lives. Requires Accessibility (the tap), and the tap is
+/// handed back while this login session is not the one on screen.
 final class TextSnippetService {
     static let shared = TextSnippetService()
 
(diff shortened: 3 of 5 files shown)


PathGao:
Pulling `tapShouldRun` into one property and having `recoverTapIfNeeded` read it is the right call — the wake path reinstalling into a switched-away session was a real second decider.

A third installer still doesn't ask. `AppSwitcher.runEventTap` ends `if clearEventTapThread() { installTap() }`, and `TextSnippetService.runEventTap` ends `if clearEventTapThread() { startOnMain() }` → `start()`. Neither goes through the gate.

Reachable: the tap thread runs `clearEventTapThread()` (reads `pendingStartAfterStop == true`, clears `tapThread`), the resign handler's `removeTap()` then sees `threadExists == false` and no-ops, and the thread goes on to `installTap()` and builds a fresh tap in the switched-away session. `routeCanStartSession` is false so the panel stays shut, but the tap is back in the chain filtering `keyDown` — #1075 again, and nothing corrects it until the next switch.

Laziest fix is to route both tails through the gate instead of around it: `startOnMain()` calling `syncWithPreferences()`, and AppSwitcher's tail hopping to main before `installTap()`.

Not verified: #1230's `knownUngatedTaps` table isn't in this diff.

**Impact:** switcher and snippet taps can reinstall into a switched-away session when a thread restart is in flight at the moment of the switch.
**Merge:** needs the two `clearEventTapThread()` tails gated first.

ref #1153
ref #1075

PathGao:

PathGao:
@PathGao Both tails are gated now, in `54545da`.

The gap was the third installer you named: `clearEventTapThread()` reports a
start that crossed a stop, and the tail rebuilt the tap right there —
`installTap()` in `AppSwitcher.runEventTap`, `startOnMain()` → `start()` in
`TextSnippetService.runEventTap`. Neither asked the gate, so a session handover
landing between the `clearEventTapThread()` and the rebuild left the resign
handler's teardown with no thread to stop and the tail free to put a fresh
filter tap into the switched-away session.

Both tails now go `DispatchQueue.main.async { syncWithPreferences() }` — the
main thread is where `SessionActivity.isActive` is written and where the gate
already lives, so the restart re-asks the same question the preference sync
asks. I went to `syncWithPreferences` rather than a main-thread hop in front of
`installTap` because `MouseClickDebounceService.finishEventTapThread` already
does exactly that, and it keeps `tapShouldRun` the switcher's single decider
instead of adding a fourth reader. Each file had two such tails, not one — the
early return when `shouldStopTapThread` was already set before `tapCreate` has
the same window — and both are routed. The third `clearEventTapThread()` call in
each file is the `tapCreate` failure path, which drops the pending restart on
purpose; none of these services retries a failed creation, so I left it.

Grepping `clearEventTapThread()` across the repo turns up seven files, so I went
through the other five rather than assume:

- `MouseClickDebounceService` — gated, and its tail already hops to main and
  calls `syncWithPreferences` (with a generation check on top). Nothing open.
- `SuperKeyService`, `KeyboardDebounceService` — same tail shape and the same
  window, but neither is session-gated at all yet; #1235 covers both, so I did
  not touch them.
- `FinderCutPaste`, `FinderRenameService` — same, and #1234's.

So the shape has five sites that need the gate; two are in this PR, three are in
those two branches, and whichever lands last should get the same tail treatment.
Worth checking there, since the gate has to exist before the tail can ask it.

Verified: `./build.sh --test` is 9487/9489, the two known input-source failures
that are red on `origin/main` too. `./build.sh` was run as well, since none of
these service files are in the `--test` compile list — the release compile of the
app, the fan helper, the icon and the asset catalog all came out with zero
warnings, but the run then died at signing with `errSecInternalComponent` on the
fan-control helper (`security find-identity -v` finds no valid identity in this
shell), so the bundle was never assembled. Compile verified clean, bundle step
not — I did not re-run it. The new assertions were confirmed red before I trusted
them: putting `installTap()` back in the switcher's tails and `start()` back in
the snippet hop turned the run red with `AppSwitcher.swift has no tap-thread
restart that goes around the session gate` and `TextSnippetService.swift
restarts its tap thread through the gate on the main thread`, green again on
restore. The check is scoped by `code.contains("clearEventTapThread")` inside
the existing roster loop, so it applies to any gated owner that grows a thread
restart, not just these two.

Not verified: the race itself. There is still no second login session here, and
this one needs a stop and a start to cross at the instant of a handover, so it is
a read of the lifecycle code rather than a reproduction. #1230's
`knownUngatedTaps` table is still not in this diff.

PathGao:
Both tails are gated now, and routing them through `syncWithPreferences` instead of a hop in front of `installTap` was the right pick — it keeps one decider rather than adding a fourth reader.

Main moved under the branch, and that's the conflict. `recoverTapIfNeeded` is gone; #882's `reconcileTakeover` replaced it, and it spells the predicate by hand again with no session term:

```swift
guard AppFeature.switcher.isAvailable,
      UserDefaults.standard.bool(forKey: DefaultsKey.switcherEnabled),
      Permissions.shared.accessibility else { restoreNativeHotkeys(); return }
if !isTapLive { restoreNativeHotkeys(); removeTap(); installTap() }
```

`recoverTapAfterWake` calls it twice, immediately and again 3s later. So the rebase drops your `guard tapShouldRun` onto a function that no longer exists, and the wake path is ungated again. `guard tapShouldRun else { restoreNativeHotkeys(); return }` is the same one-liner.

The rest of the conflict is already yours: `CFMachPortInvalidate` is on main in `AppSwitcher` (before `clearEventTapThread`), `DockClickService.stop` and `PreciseVolumeRollerService.removeTap`. Drop those three hunks, keep the gating. Read off main's source, not a rebase run.

Smaller: `MouseClickDebounceService` isn't in the roster, and the tail check wants a func literally named `restartTapOnMain`, so the one gated owner that already does this can't join without a rename.

**Impact:** after the rebase, waking while switched away reinstalls the switcher tap — #1075 on the wake path, the case this PR set out to close.
**Merge:** needs the rebase with `reconcileTakeover` gated.

ref #1153
ref #1075

PathGao:
@PathGao Rebased onto `main` at `a952b829`, head is now `18acd8ef`. Three
things changed on top of the rebase (`ac9e1c2e`).

**The rebase.** #1225 landed `CFMachPortInvalidate` in all four files, so those
hunks are gone from the diff. #1219 landed under `DockClickService` and #882
under `AppSwitcher`; the note in the body about whichever lands second needing
a rebase is settled — this is that rebase. `recoverTapIfNeeded` no longer
exists, so the second installer this branch gates is now #882's
`reconcileTakeover`, which is still the post-wake recheck and still the second
place that can put the tap in.

**The provisional re-arm is gone.** `AppSwitcher` and `TextSnippetService`
re-armed under the lifecycle lock and then corrected on main. That is the one
of the three shapes across these branches that cannot hold: putting the tap
back first is putting it back into whatever session is on screen now, which is
the stall this branch exists to end, for as long as the main queue takes to
correct it. Both now leave the tap out of the chain, hop, and either re-arm
there or hand the tap to the sync — the shape #1235 already had and #1234 has
moved to. The main-run-loop pair (`PreciseVolumeRollerService`,
`DockClickService`) reads the flag in place because the callback is already on
the thread that writes it. `AXIsProcessTrusted()` is asked in the same place
for the same reason, which the re-arms here were not doing.

**The permission source.** This branch was internally inconsistent —
`PreciseVolumeRollerService` asked the system, `AppSwitcher`, `DockClickService`
and `TextSnippetService` read the 60-second poll. All four gates ask
`AXIsProcessTrusted()` now, matching #1233, #1234 and #1235. The mirror stays
on `AppSwitcher.handle` and `beginPendingSession`, where the comments say a
live TCC round-trip per key is what they avoid.

**`tapShouldRun` renamed to `tapIsWanted`.** The property called
`SessionActivitySupport.tapShouldRun` while sharing its name, and #1234 used
the same name for a stored sample, so one name meant three things across three
open branches. The sample is gone from #1234 and the static helper keeps the
name; the local one is `tapIsWanted`.

**Tests.** `./build.sh --test` → `TESTS OK (9553 checks)`. The two
machine's input source is ANSI now. `swiftc -typecheck` over all of
`Sources/Vorssaint/**` is clean; none of the four files is in the `--test`
whitelist. Full `./build.sh` not run this round — signing keychain locked.
Verified both ways: putting `installTap()` back on `AppSwitcher`'s tail turns
`AppSwitcher.swift has no tap-thread restart that goes around the session
gate` red on its own.

**Where to look first — I narrowed one of your assertions and want that
checked.** The restart-tail check fired for any owner in the roster with a
`clearEventTapThread`. That is the right rule, and as written it turns red on
#1234 and #1235 the moment either lands: `FinderCutPaste`,
`FinderRenameService` and `SuperKeyService` still restart with a bare
`installTap()` / `startOnMain()`, and `KeyboardDebounceService` reaches
`start()` through a local, which the string form would not even catch.

Rather than fix four files across three branches inside this one, I scoped the
check to the files that adopted `restartTapOnMain` — so the half-fix is what
it catches, one tail careful and another not — and named the four in the
comment. That leaves the defect standing in those four rather than pinned. If
you would rather have it pinned repo-wide and the four fixed, that is four
one-line changes and I will put them on the branches that own the files.

PathGao:
@PathGao The restart check is back to your original scope, in `4673089b`. Not a
force-push; one commit on top of what you last saw.

I narrowed it last round to `code.contains("restartTapOnMain")` because four other
own-thread taps still restarted around the gate and widening it would have turned this
branch red the moment #1234 or #1235 landed. Your call was to fix the four rather than
narrow the check, so they are fixed on the branches that own them —
`FinderCutPaste` and `FinderRenameService` in #1234 (`e437653b`), `SuperKeyService` and
`KeyboardDebounceService` in #1235 (`10fafdce`) — and the check asks every roster owner
that runs a tap thread again.

Landing order does not matter: each of the three branches carries the check for its own
files, and the roster only ever holds files that already answer it.

One thing that came out of doing it, worth having here since this is where the check
lives. **`SuperKeyService` was not the same defect as the other three.** Its
`startOnMain()` already hopped to the main thread; what it skipped was the question —
`start()` guards on `AXIsProcessTrusted()` alone and the session conjunct lives in
`syncWithPreferences`. So the check as written, which looks for `startOnMain()` on the
tail, would have passed a file whose hop was right and whose gate was absent. The second
half of the check — that `restartTapOnMain`'s body reaches `syncWithPreferences` — is what
catches it, and it is the half that earned its keep.

**And one file the check does not reach.** `KeyboardDebounceService`'s tail branches on a
local rather than spelling the call inline, so reverting both of its tails to `start()`
leaves the suite green — I checked rather than assumed. Pinning it would mean matching a
variable name and its indentation, so it is named in #1235's body instead of being chased
into the source.

`./build.sh --test` → `TESTS OK (9553 checks)`, no failures. `swiftc -typecheck` over all
of `Sources/Vorssaint/**` clean. Full `./build.sh` not run — signing keychain locked.
Verified both ways again after widening: putting `installTap()` back on `AppSwitcher`'s
tail turns it red on its own.

**Where to look first.** The `MetricsTests` loop body is now edited by four open branches
at once — the roster array, the modifying-tap condition, the gate-argument check and this
restart check. Whichever lands first, the rest need a rebase there, and the resolutions
are all "keep both". #1230 replaces the roster wholesale, so if you would rather take that
first the four rebases get simpler rather than harder.

PathGao:
@PathGao One line on top (`ec2e1736`): `DockClickService`'s in-callback re-arm now asks `AXIsProcessTrusted()` ahead of the session flag, the same shape as `PreciseVolumeRollerService`. The 18:57 note that all four owners ask the grant in the same place was true of three; this makes it four. `./build.sh` exit 0, `./build.sh --test` TESTS OK (9553 checks), unchanged count since no assertion pins that conjunct.

PathGao:
Confirmed against `main`: all four re-arms ask `AXIsProcessTrusted()` now, and every `installTap()` / `start()` call site in `AppSwitcher` and `TextSnippetService` reaches the gate — the sync, `reconcileTakeover`, and both thread tails each. No fifth installer. That's grep, not a build.

Two one-liners on the conjunct you just added, neither blocking.

**Nothing pins it for these four.** The roster's `if tapOwner.contains("MouseNavigation") || ...` still names only the original four, so the fact your last two commits establish can regress silently — that is why the count didn't move. Adding the four new paths to that disjunction is the whole fix.

**The grant goes first here, last everywhere else.** `MouseClickDebounceService`, `AppSwitcher` and `TextSnippetService` all read the session flag first and let it short-circuit. As written, a tap disable in a switched-away session pays a live TCC round-trip on the main run loop inside a callback the window server is blocking on — the stall this branch exists to end. Swap the conjuncts in `DockClickService` and `PreciseVolumeRollerService`.

**Impact:** none from the new line itself; the ordering costs one TCC call per tap disable while switched away.
**Merge:** ship it.

ref #1153
ref #1075

PathGao:

PathGao:
Reopened as #1304.

Issue #566, opened 2026-08-09
Retain terminal output/log of last action in homebrew


AB-boi:
### Problem or Motivation

The app is an indispensable part of my day to day workflows! Thanks for it!

It would be nice to have the option to see the terminal output or log of the last action done in homebrew so that we can catch any errors if necessary. Currently, the log disappears after a short while.

Idk if there is already a setting that persists the log when something throws an error, but I think the last log should persist for more certainty.

### Proposed Solution

A button that shows logs/last n logs/last 1 log

Logs remain open if there was an error

### Feature Area

Other

PathGao:
Triage note, part of a batch pass over the open issues. The labels here were applied by a contributor, not a maintainer ruling.

The log UI you are asking for is not missing — it is being erased on a timer.

`HomebrewOperationStatusView` (`Sources/Vorssaint/UI/HomebrewOperationStatusView.swift`) already exists and is already wired into Settings at `HomebrewSettings.swift:552`, as well as the menu bar panel and the uninstaller. What removes it is `HomebrewManager.scheduleCompletedOperationCleanup` (`Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift:588`): it schedules a work item that sets `self.log = ""` (`:615`) after 8 seconds on success, 6 on cancel and 20 on failure or needs-terminal (`:597-605`).

So the cheap version is two edits: stop clearing `log` in that work item while still clearing `operationStatus`, and add a "view last log" entry next to the status view at `HomebrewSettings.swift:552`. Note the failure branch already sets `clearsWholeStatus = false`, so the retain-on-error behaviour you asked for is one branch away from existing.

Worth deciding while implementing: keep the last log only, or a short ring of the last few — the storage is a single `String` today, so "last one" is the smaller change.

Borisserz:
Proposed in https://github.com/vorssaint/vorssaint-utils/pull/$(gh pr list --repo vorssaint/vorssaint-utils --head Borisserz:feat/homebrew-keep-last-operation-log --json number -q '.[0].number')

Keeps the last Homebrew operation log after the completion timer clears status chrome, with a View last log control in Settings.

PR #1237, opened 2026-09-01
fix(homebrew): bound the streaming command's wait, output and log

PathGao:
## Summary

A Homebrew operation that never finishes used to take the whole Homebrew pane
with it, and a long one grew without bound in three places.

**The process was waited on with `waitUntilExit()`, which takes no timeout.**
That wait ran on `HomebrewManager`'s *serial* `workQueue`, so one stuck `brew
install` parked every later `refreshInstalled`, `search`, `refreshOutdated`,
`select` and `packageManagingApplication` behind it, permanently.
`runStreaming` now waits on a semaphore its `terminationHandler` signals — the
same shape as `run` a few lines above — bounded at 30 minutes, and escalates
through the existing `stop(_:finished:)` when the bound expires. The bound is
the whole fix: that semaphore blocks the same thread `waitUntilExit()` blocked,
it is just finite.

**The retained output was unbounded.** `runStreaming` accumulated every byte
for the completion handler, which reads the last few lines of it. It now keeps
a 4 MB tail. The tail and not the head, because the error this output is read
for is the last thing brew prints. Decoding moved to `String(decoding:as:)` so
a dropped head cutting a multi-byte character in half cannot take the whole
output — and the error message with it — down with it.

**The published log was unbounded and republished whole.** `appendLog` did
`log.append(text)` on an `@Published` string for every stdout chunk. brew's
download phase is a curl progress bar refreshing on `\r` many times a second,
and `brew upgrade --cask --greedy` across several packages runs to megabytes;
every append pushed the entire string to `HomebrewOperationStatusView`, which
laid all of it out again. The log is now cut back to its last 64 KB.

**Three regular expressions were compiled per output chunk on the main
thread.** `updateOperationStatus` calls `phase` + `progressFraction` +
`activity` for every chunk, and both `progressFraction` and `stripANSI` built
a fresh `NSRegularExpression` on each call. They are `static let` now.

### What each bound does when it trips

The three bounds do not behave alike, and only one of them ends the operation.

- **The 30-minute wait aborts the operation.** `stop(_:finished:)` sends
  SIGTERM, waits 2 s, then SIGKILL, then waits 2 s more. The completion
  handler is reached with a non-zero status and `cancelRequested` false, so
  `perform` takes its last branch: the operation card goes to **Failed** and
  `errorMessage` is whatever `visibleError` finds in the retained tail. It is
  not reported as cancelled, because the person did not cancel it. brew is
  killed mid-command, so the package can be left half-installed exactly as it
  would be after a crash or a cancel — rerunning the action is the fix, and
  that is unchanged from the existing cancel path.
- **The 4 MB output cap truncates silently and the operation keeps running.**
  The head is dropped, no marker is inserted, and the read handler keeps
  draining so the child can never block on a full pipe. Nothing downstream
  loses a signal to it: `visibleError`, `needsTerminalFallback` and
  `untrustedTapName` all look for things brew prints as it stops, which are in
  the tail by construction.
- **The 64 KB log cap truncates visibly and the operation keeps running.** The
  head is dropped back to a whole-line boundary and a leading `…` is left in
  its place, so the pane and "copy log" both show that something was cut.

**No child is left orphaned by a tripped bound.** SIGKILL cannot be ignored,
`activeProcess` is cleared in the completion handler, and the drain that
follows is itself capped at one second before the read handle is closed — so
even a child wedged in an uninterruptible wait releases the serial queue
instead of holding it. What survives a kill is what already survived a cancel:
brew's own grandchildren (`curl`, `git`) are not in a process group this
signals, and that shape is untouched here.

### Scope against #971

The previous round of this section was wrong, in the direction of claiming
more than the diff does. Corrected:

**#971 names two sites, and this PR touches neither.**

- `BoundedProcessRunner.run` — a `waitUntilExit()` parked on
  `DispatchQueue.global` and then *abandoned*, because `run` returned on
  timeout without ever collecting it. That is the one that accumulates: seven
  call sites reach it on this head, several of them periodic samplers, so
  stranded waiters piled up on their own until the 64-thread pool had nothing
  left.
- `WindowEnumerator.listWindows` — `waitUntilAllOperationsAreFinished()` with
  no timeout, on the main thread, which is what turns a full pool into a dead
  app rather than a slow one.

Both are already bounded on current `main`, which is why the issue carries
`fixed (pending release)`.

**What this PR fixes is a third occurrence of the shape, at a site #971 never
mentions, and one that could not have produced the reported freeze.** The
reporter's own notes say Homebrew is uninstalled on his machine, so this call
site was dead for him. More to the point, this wait runs on Homebrew's private
*serial* `workQueue` and is never abandoned — it holds one thread at a time and
cannot count towards the 64. Unbounded it starved that one queue, which is the
defect in the Summary above, not the pool exhaustion #971 reports.

**The same is true of the two `waitUntilExit()` calls still on `main`**, so the
previous round's "the pool can still fill from those two" was wrong on both:

- `DiskImageInstallerService.run` (`DiskImageInstallerService.swift:303`) —
  `readDataToEndOfFile()` then `waitUntilExit()`, no timeout, on that service's
  own serial `workQueue`. A stalled `spctl -a` parks that queue for good, one
  thread, no accumulation.
- `JunkCleaner.bootoutUserAgent` (`JunkCleaner.swift:284`) — called from a
  sequential `for` loop inside a single `DispatchQueue.global` block, once per
  login-item plist. A hung `launchctl bootout` parks that one clean run.

So the precise claim: **this PR fixes neither defect #971 reports, and #971
should not take `half` on account of it.** The `Refs #971` line has therefore
been removed: on merge that line is what puts a report on the fix track and
tells its reporter which release carries the fix, and neither is true here.
The shape and the diagnosis still come from #971, which is why it is named
throughout this section — as a reference, not as a claim to have fixed it.

Trade-offs:

- **Timeout value: 30 minutes, deliberately loose.** A large cask on a slow
  line legitimately runs for many minutes, and cutting off a real install is
  worse than the stall this exists to end. The value only has to be finite,
  because what it protects is the serial queue behind it. A tight bound
  modelled on `brewReadTimeout` (30s) was rejected for that reason.
- **The log keeps its tail, so a very long install loses its beginning**, and
  "copy log" then hands over the last 64 KB with a leading `…`. The
  alternative — a ring of chunks, or spilling to a file — is a data structure
  and a lifetime for a string that is already erased 8 seconds after the
  operation ends.
- The output cap and the log cap stayed separate limits because they answer
  to different readers: 4 MB feeds the error parser, 64 KB feeds a `Text`
  view.
- `appendingToLog` decides its cut in `Characters`, the unit `suffix(limit)`
  trims in. An earlier round decided it on `utf8.count` alone, which mixed
  units: on multi-byte text the guard fired while `suffix(limit)` was already
  the whole string, so each call dropped one leading line and prepended `…`
  instead of capping anything. brew's output is ASCII so it never bit that
  path, but a localised tool's error line would have reached it.
- **The Character walk is now screened, so this change does not undo its own
  regex hoist.** `log.count` is O(n) per chunk on the main thread, in a
  change that hoisted three regexes off that same path for exactly that
  reason. The guard is now `guard log.utf8.count > limit * 2, log.count >
  limit * 2`: `utf8.count` is O(1) on a native string and a string never has
  more Characters than UTF-8 bytes, so the cheap test can only turn away a
  chunk no trim was due for, and `log.count` runs only for the few that get
  past it. The decision stays in Characters. Nothing observable separates
  the two orders, so nothing pins the screen: the comment above the guard
  is what records it. A source-shape assertion that pinned the guard's
  spelling was added and then dropped, because it went red on an equivalent
  spelling and stayed green on the same O(n) walk anywhere else.
- **The per-chunk `String(data:encoding:.utf8)` in the read handler is
  untouched, and it is a real defect.** It returns nil when a read boundary
  splits a multi-byte character, and the entire chunk is then dropped — from
  the log and from the progress parser both. That line predates this PR, and
  the fix is not the lenient decode used for the final output: decoding
  leniently per chunk trades a lost line for a `U+FFFD` in the middle of one.
  Doing it correctly means carrying the trailing partial bytes over to the
  next chunk, which is state this change has no other reason to introduce.
  Flagged rather than fixed; either version can be had here or in a follow-up.

## Verification

```sh
./build.sh --test     # 9490 checks, 0 failures
```

Run on this head. `./build.sh --test` compiles a whitelist that contains
`HomebrewSupport.swift` but **not** `HomebrewManager.swift`, so that file is
not compiled here at all and CI covers it on both runners.

The only change to `HomebrewManager.swift` across the review rounds was the block comment above the
`terminationHandler` (later rounds touched `HomebrewSupport.swift` and the tests only) — it stated the defect wrongly
(pool exhaustion rather than starvation of this one serial queue, and the
mechanism rather than the bound as the fix), and it is what the two remaining
sites would be copied from.

Every new assertion was verified red-first when it was added — broken, seen to
fail, restored, seen green:

- `Homebrew operation log measures its limit in the unit it trims in` —
  dropping `log.count` so `utf8.count` becomes the decision rather than the
  screen fails it.
- `Homebrew operation log keeps its tail instead of growing without bound` —
  neutering the length check in `appendingToLog` fails it.
- `Homebrew progress parser cleans activity lines` (existing) — breaking the
  hoisted ANSI pattern fails it.

**Not tested:** none of it against a real slow or hung `brew`. The timeout
path, the SIGTERM/SIGKILL escalation and the 4 MB output cap were reasoned
from the code and from the matching paths in `run` and
`BoundedProcessRunner`, not exercised — including the Failed-not-Cancelled
routing described above, which was traced through `perform` rather than
observed. No multi-package `brew upgrade --cask --greedy` was run to watch the
log actually reach the 64 KB cap in the UI. No GUI testing was done at all.
The claims about `DiskImageInstallerService` and `JunkCleaner` in *Scope
against #971* are read from the code on `main`, not run.

## Anything else

The two `waitUntilExit()` calls outside this PR's files are left untouched on
purpose. Each starves its own queue and neither can fill the pool, as set out
in *Scope against #971* above, so neither is a #971 site and neither is urgent
on that account — they are still unbounded waits and still worth bounding.
Reaching them means `DiskImageInstallerService` and `JunkCleaner`, neither of
which this change otherwise goes near, and a separate change is what the
review called that.

No `Refs`. #566 asks for the Homebrew log of the last action to be kept instead of wiped,
and this does not do that: `scheduleCompletedOperationCleanup` still sets `log = ""` after
a success, a cancel and a failure, and nothing here goes near it. The 64 KB cap is a
constraint on #566 rather than a piece of it — if the log is ever kept, the cap is what
decides what "kept" means — so it belongs in a comment there, where `Refs` semantics cannot
put the report on the fix track for a build in which the log still vanishes.

### DIFF
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
@@ -778,11 +778,13 @@ final class HomebrewManager: ObservableObject {
                     DispatchQueue.main.async { onOutput(chunk) }
                 }
             }
-            // Watched through the termination handler, never `waitUntilExit()`:
-            // that parks a worker of the shared 64-thread pool for the whole
-            // command, and one hung `brew install` used to hold this serial
-            // queue for good, with every later refresh, search and ownership
-            // lookup stuck behind it (see BoundedProcessRunner, issue #971).
+            // The bound is the fix. This wait blocks the serial queue's thread
+            // exactly as `waitUntilExit()` did — one at a time, never
+            // abandoned, so it cannot fill the shared pool the way #971's
+            // waits do. Unbounded it starved this one queue instead: a `brew
+            // install` that never returned left every later refresh, search
+            // and ownership lookup behind it for good. The semaphore is only
+            // how a timeout is put on a wait `waitUntilExit()` will not take.
             let finished = DispatchSemaphore(value: 0)
             process.terminationHandler = { _ in finished.signal() }
             do {
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
@@ -668,6 +668,19 @@ final class HomebrewManager: ObservableObject {
     private static let brewReadTimeout: TimeInterval = 30
     private static let processTerminationGrace: TimeInterval = 2
 
+    /// Timeout for an operation the person started (install, upgrade,
+    /// uninstall). Deliberately far beyond any real command: a large cask on
+    /// a slow line legitimately runs for many minutes, and cutting one of
+    /// those off is worse than the stall this exists to end. It only has to
+    /// be finite, because the wait holds the serial queue and everything
+    /// behind it.
+    private static let brewOperationTimeout: TimeInterval = 30 * 60
+
+    /// How much of an operation's output is kept for the completion handler,
+    /// which reads the last few lines of it. The download phase is a curl
+    /// progress bar that can run to megabytes.
+    private static let brewOperationOutputLimit = 4 * 1024 * 1024
+
     /// SIGTERM is cooperative. Escalate only when a command ignores it so a
     /// cancelled or timed-out operation cannot keep the serial queue forever.
     private static func stop(_ process: Process, finished: DispatchSemaphore? = nil) {
@@ -752,12 +765,26 @@ final class HomebrewManager: ObservableObject {
                     return
                 }
                 lock.lock()
+                // Keep draining once the retained tail is full, so the child
+                // can never block on a full pipe or turn its output into
+                // memory use. The tail is kept and not the head because the
+                // error this is read for is the last thing brew prints.
                 output.append(data)
+                if output.count > Self.brewOperationOutputLimit * 2 {
+                    output.removeFirst(output.count - Self.brewOperationOutputLimit)
+                }
                 lock.unlock()
                 if let chunk = String(data: data, encoding: .utf8) {
                     DispatchQueue.main.async { onOutput(chunk) }
                 }
             }
+            // Watched through the termination handler, never `waitUntilExit()`:
+            // that parks a worker of the shared 64-thread pool for the whole
+            // command, and one hung `brew install` used to hold this serial
+            // queue for good, with every later refresh, search and ownership
+            // lookup stuck behind it (see BoundedProcessRunner, issue #971).
+            let finished = DispatchSemaphore(value: 0)
+            process.terminationHandler = { _ in finished.signal() }
             do {
                 try process.run()
             } catch {
@@ -770,19 +797,24 @@ final class HomebrewManager: ObservableObject {
                 self.activeProcess = process
                 if self.cancelRequested { Self.stop(process) }
             }
-            process.waitUntilExit()
+            if finished.wait(timeout: .now() + Self.brewOperationTimeout) == .timedOut {
+                Self.stop(process, finished: finished)
+            }
             _ = drained.wait(timeout: .now() + 1)
             pipe.fileHandleForReading.readabilityHandler = nil
             try? pipe.fileHandleForReading.close()
             lock.lock()
-            let finalOutput = String(data: output, encoding: .utf8) ?? ""
+            // Decoded leniently: a dropped head can cut a multi-byte
+            // character in half, and losing the whole output over that would
+            // take the error message with it.
+            let finalOutput = String(decoding: output, as: UTF8.self)
             lock.unlock()
-            completion(process.terminationStatus, finalOutput)
+            completion(process.isRunning ? -1 : process.terminationStatus, finalOutput)
         }
     }
 
     private func appendLog(_ text: String) {
-        log.append(text)
+        log = HomebrewProgressParser.appendingToLog(log, text)
     }
 
     private func appleScriptString(_ value: String) -> String {
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
@@ -517,7 +517,11 @@ enum HomebrewProgressParser {
                                limit: Int = logLimit) -> String {
         var log = existing
         log.append(text)
-        guard log.utf8.count > limit * 2 else { return log }
+        // Counted in Characters, the same unit `suffix(limit)` trims in. A
+        // byte count here would fire on multi-byte text while `suffix` was
+        // already the whole string, dropping a leading line per call instead
+        // of capping anything.
+        guard log.count > limit * 2 else { return log }
         let tail = log.suffix(limit)
         let wholeLines = tail.drop { !$0.isNewline }.dropFirst()
         return "…\n" + (wholeLines.isEmpty ? tail : wholeLines)
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
@@ -517,11 +517,14 @@ enum HomebrewProgressParser {
                                limit: Int = logLimit) -> String {
         var log = existing
         log.append(text)
-        // Counted in Characters, the same unit `suffix(limit)` trims in. A
-        // byte count here would fire on multi-byte text while `suffix` was
-        // already the whole string, dropping a leading line per call instead
-        // of capping anything.
-        guard log.count > limit * 2 else { return log }
+        // The cut is decided in Characters, the same unit `suffix(limit)`
+        // trims in: a byte count deciding it would fire on multi-byte text
+        // while `suffix` was already the whole string, dropping a leading
+        // line per call instead of capping anything. The byte count leads
+        // because it is O(1) and never below the Character count, so it
+        // turns away the chunks that would otherwise pay for an O(n) walk
+        // on the main thread. Order matters; keep the cheap test first.
+        guard log.utf8.count > limit * 2, log.count > limit * 2 else { return log }
         let tail = log.suffix(limit)
         let wholeLines = tail.drop { !$0.isNewline }.dropFirst()
         return "…\n" + (wholeLines.isEmpty ? tail : wholeLines)
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift
@@ -496,10 +496,36 @@ enum HomebrewAnalytics {
 }
 
 enum HomebrewProgressParser {
+    /// Compiled once. Every one of these runs on the main thread for every
+    /// chunk a running operation emits, and brew's download phase is a curl
+    /// progress bar that refreshes many times a second.
+    private static let percentRegex = try? NSRegularExpression(
+        pattern: #"([0-9]{1,3}(?:\.[0-9]+)?)%"#)
+    private static let ansiRegex = try? NSRegularExpression(
+        pattern: #"\u001B\[[0-9;?]*[ -/]*[@-~]"#)
+
+    /// How much of a running operation's log is kept. `brew upgrade --cask
+    /// --greedy` across several packages emits megabytes of progress bars,
+    /// and every append republishes the whole string for the view to lay out
+    /// again.
+    static let logLimit = 64 * 1024
+
+    /// The log with `text` appended, cut back to its last `limit` characters
+    /// once it grows past twice that. The tail is what says where the
+    /// operation is, so the head is what goes.
+    static func appendingToLog(_ existing: String, _ text: String,
+                               limit: Int = logLimit) -> String {
+        var log = existing
+        log.append(text)
+        guard log.utf8.count > limit * 2 else { return log }
+        let tail = log.suffix(limit)
+        let wholeLines = tail.drop { !$0.isNewline }.dropFirst()
+        return "…\n" + (wholeLines.isEmpty ? tail : wholeLines)
+    }
+
     static func progressFraction(in output: String) -> Double? {
         var latest: Double?
-        let pattern = #"([0-9]{1,3}(?:\.[0-9]+)?)%"#
-        guard let regex = try? NSRegularExpression(pattern: pattern) else { return nil }
+        guard let regex = percentRegex else { return nil }
         let range = NSRange(output.startIndex..<output.endIndex, in: output)
         regex.enumerateMatches(in: output, range: range) { match, _, _ in
             guard let match,
@@ -601,9 +627,7 @@ enum HomebrewProgressParser {
     }
 
     private static func stripANSI(_ value: String) -> String {
-        guard let regex = try? NSRegularExpression(pattern: #"\u001B\[[0-9;?]*[ -/]*[@-~]"#) else {
-            return value
-        }
+        guard let regex = ansiRegex else { return value }
         let range = NSRange(value.startIndex..<value.endIndex, in: value)
         return regex.stringByReplacingMatches(in: value, range: range, withTemplate: "")
     }
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -11514,18 +11514,6 @@ struct MetricsTests {
         expect(HomebrewProgressParser.appendingToLog(String(repeating: "日本語テキスト\n", count: 2),
                                                      "日本語テキスト\n", limit: 32) == multibyteLog,
                "Homebrew operation log measures its limit in the unit it trims in")
-        // Both counts answer the same question, but `log.count` walks
-        // graphemes on the main thread for every chunk a running operation
-        // emits. Nothing observable separates the two orders, so the order
-        // is pinned here instead.
-        let homebrewSupportSource = (try? String(
-            contentsOfFile: "Sources/Vorssaint/Services/Homebrew/HomebrewSupport.swift",
-            encoding: .utf8)) ?? ""
-        let logCapGuard = homebrewSupportSource.components(separatedBy: "\n").first {
-            $0.trimmingCharacters(in: .whitespaces).hasPrefix("guard log.")
-        } ?? ""
-        expect(logCapGuard.contains("log.utf8.count > limit * 2, log.count > limit * 2"),
-               "Homebrew operation log screens on the O(1) byte count before walking Characters")
 
         let homebrewJSON = """
         {
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -11501,6 +11501,14 @@ struct MetricsTests {
                == "Error: Cask failed",
                "Homebrew progress parser hides command lines from visible errors")
 
+        expect(HomebrewProgressParser.appendingToLog("head\n", "tail\n", limit: 64) == "head\ntail\n",
+               "Homebrew operation log is left alone while it fits")
+        let cappedLog = HomebrewProgressParser.appendingToLog(
+            String(repeating: "old\n", count: 200), "last line\n", limit: 32)
+        expect(cappedLog.hasSuffix("last line\n") && cappedLog.utf8.count <= 96
+                && !cappedLog.hasPrefix("old"),
+               "Homebrew operation log keeps its tail instead of growing without bound")
+
         let homebrewJSON = """
         {
           "formulae": [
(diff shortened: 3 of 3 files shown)


PathGao:
`runStreaming` now matches `run` line for line — same `terminationHandler` + semaphore, same `stop(_:finished:)`, same `isRunning ? -1 :` guard before `terminationStatus`.

**The shape survives in two other places.** On `main` these are the only remaining `waitUntilExit`, both unbounded:

- `DiskImageInstallerService.run` — `readDataToEndOfFile()` then `waitUntilExit()`, no timeout, on that service's serial `workQueue`. A stalled `spctl -a` parks it for good.
- `JunkCleaner.bootoutUserAgent` — `waitUntilExit()` on `DispatchQueue.global`, once per plist.

So referencing #971 marks it fixed on merge while the pool can still fill from those two. Either scope the claim in the body, or let #971 take `half`.

**`appendingToLog` mixes units** — the guard counts `utf8.count`, `suffix(limit)` counts Characters. On multi-byte-heavy text the guard fires while `suffix(limit)` is already the whole string, so it drops one leading line per call instead of capping. `log.count > limit * 2` puts both sides on the same unit. Won't bite on brew's ASCII output, and the new test is ASCII-only so it wouldn't catch it either.

**Impact:** none new — every change here is a bound that was missing before.
**Merge:** ship it.

ref #971

PathGao:
@PathGao Both notes addressed, one in code and one in the body.

**The unit mix is fixed.** `appendingToLog` now guards on `log.count`, the same
unit `suffix(limit)` trims in, as you suggested. New assertion `Homebrew
operation log measures its limit in the unit it trims in` covers it with
multi-byte text: 24 Characters, 66 UTF-8 bytes, `limit: 32` — under a
Character limit, over a byte one. Red-first confirmed: putting `utf8.count`
back fails that one check and only that one (1 of 9490), restoring it goes
green at 9490/0. The existing ASCII assertion is unchanged and still passes,
so it did not need replacing, only a sibling.

**The other two `waitUntilExit()` sites are left alone**, per the addendum rule
against widening a `ready to merge` diff — you called them a separate change
yourself. I verified your claim rather than taking it: `git grep waitUntilExit`
on `origin/main` returns exactly `DiskImageInstallerService.swift:303`,
`JunkCleaner.swift:284`, the site this PR replaces, and the comment in
`BoundedProcessRunner.swift:63`. So the body now scopes the claim instead —
there is a *Scope against #971* section naming both remaining sites and saying
plainly that the `Refs #971` line is partial. Whether that means `half` on the
issue or a follow-up per site is your call, not this PR's.

**Separately, I checked what each of the three bounds does when it trips**,
since the body only said what they cap and not what the person sees. They do
not behave alike and the body now says so: the 30-minute wait *aborts* (SIGTERM,
2 s, SIGKILL, 2 s, then the `else` branch in `perform` marks it **Failed**, not
Cancelled, since `cancelRequested` is false — brew can be left half-installed
exactly as after a cancel); the 4 MB output cap truncates *silently* and lets
the command run on; the 64 KB log cap truncates *visibly*, with a leading `…`
at a line boundary. Worth having in the body because "silently truncates a
`brew upgrade`" and "kills a `brew upgrade`" are very different things to read
about after the fact.

No bound can orphan a child. SIGKILL cannot be ignored, `activeProcess` is
cleared in the completion handler, and the `drained` wait that follows is
capped at one second before the read handle is closed — so even a child wedged
in an uninterruptible wait releases the serial queue rather than holding it.
brew's own grandchildren (`curl`, `git`) survive a kill because nothing here
signals the process group, but that is unchanged from the existing cancel path
and I did not touch it.

Verified with `./build.sh --test` only — 9490 checks, 0 failures, including
the two assertions that read the live input source. No full `./build.sh` per
the repo's local-dev rule, so `HomebrewManager.swift` is not compiled here and
CI covers it. Not verified: anything against a real slow or hung `brew`. The
timeout path, the SIGTERM/SIGKILL escalation, the 4 MB cap and the
Failed-not-Cancelled routing were all traced through the code, not exercised,
and no GUI testing was done.

PathGao:
The multibyte assertion pins the unit rather than a value. Right shape.

**The pool framing is wrong at all three sites.** On `main` both `workQueue`s (Homebrew, DiskImageInstaller) are serial, and `bootoutUserAgent` runs in a sequential `for` loop inside one `DispatchQueue.global` block — each holds one thread at a time and cannot accumulate. What fills the pool in #971 is `BoundedProcessRunner`'s *abandoned* waits, already fixed. So the new comment's "parks a worker of the shared 64-thread pool" and *Scope against #971*'s "the pool can still fill from those two" are both wrong: each site starves its own queue, which is what your Summary says.

Same root: `finished.wait(timeout:)` blocks the thread `waitUntilExit()` blocked. The timeout is the entire fix, as your Trade-offs section says — the comment claims otherwise, and it is the comment the other two sites will copy.

`String(data:encoding:.utf8)` three lines above the new lenient decode still nils when a read boundary splits a character, losing the whole chunk from log and parser.

`log.count` is an O(n) grapheme walk per chunk on the main thread, in a PR that hoisted regexes for that reason. `log.utf8.count > limit * 2, log.count > limit * 2` gates it cheaply — bytes ≥ characters, so it never skips a trim.

**Impact:** none on behaviour.
**Merge:** needs the #971 scope claim and the inline comment corrected first — they decide whether #971 takes `half` and what the other two sites copy.

ref #971
ref #566

PathGao:
@PathGao Both corrections made. You were right on both counts, and the #971
claim was wrong by more than the framing.

**The scope claim.** I read #971 again instead of reasoning from the shape.
It names two sites: `BoundedProcessRunner.run`, where the wait is *abandoned*
by a caller that returned on timeout and therefore accumulates, and
`WindowEnumerator.listWindows`, whose untimed
`waitUntilAllOperationsAreFinished()` on the main thread is what turns a full
pool into a dead app. Both are already bounded on `main` — hence
`fixed (pending release)`. This PR touches neither. It bounds a third
occurrence of the shape at a site the issue never mentions; the reporter even
says Homebrew is uninstalled on his machine, so this call site was dead for
him. And it could not have produced his symptom: the wait is on Homebrew's
private serial `workQueue` and is never abandoned, so it holds one thread at a
time and cannot count towards the 64. So the honest answer to the `half`
question is **neither** — this PR fixes no part of #971, and the section now
says that. I kept the `Refs #971` line because removing a `Refs` line is your
call, not mine, and flagged in the body that it is the thing to drop if the
automation reads `Refs` as "fixes".

Same correction for the other two sites: `DiskImageInstallerService.run` is on
that service's own serial `workQueue`, and `bootoutUserAgent` is called from a
sequential `for` inside one `DispatchQueue.global` block. One thread each, no
accumulation. "The pool can still fill from those two" was wrong.

**The comment.** Rewritten to say the bound is the fix and that the semaphore
is only how a timeout is put on a wait `waitUntilExit()` will not take — it
blocks the same thread. It now names #971's pool exhaustion as the thing this
site is *unlike*, so a copy of it into the other two sites carries the right
diagnosis rather than the wrong one.

**The two notes I did not act on**, both answered in *Trade-offs*:

- `log.count`'s grapheme walk — your gate is right and I checked the direction:
  a string never has more Characters than UTF-8 bytes, so `log.utf8.count >
  limit * 2` can only skip a trim that was not due. It is one line and I will
  push it on a word. It is not in this push because the verdict asked for two
  prose corrections and I did not want to re-open accepted code alongside them.
- the per-chunk `String(data:encoding:.utf8)` — a real defect and it predates
  this PR. Worth saying that the lenient decode is not the fix for it: applied
  per chunk it trades a lost line for a `U+FFFD` in the middle of one. Correct
  is carrying the trailing partial bytes into the next chunk, which is state
  this change has no other reason to add. Happy to do either, here or separately.

**Verified:** `./build.sh --test` on this head — 9490 checks, 0 failures. No
assertion could have moved: the only source change is a block comment in
`HomebrewManager.swift`, which `--test`'s whitelist does not compile at all, so
CI covers that file on both runners. **Not verified:** nothing new was
exercised. The `DiskImageInstallerService` and `JunkCleaner` claims above are
read from the code on `main`, not run, and nothing here was tried against a
real slow or hung `brew`. No GUI testing.

PathGao:
@PathGao Dropped `Refs #971`.

Your scope question had an answer neither of us had written down: this PR fixes **neither** of the two sites #971 names. `BoundedProcessRunner.run` and `WindowEnumerator.listWindows` are both already bounded on `main` and the issue carries `fixed (pending release)`. This wait is on Homebrew's own serial `workQueue`, one thread at a time, never abandoned — it starves that one queue rather than filling the shared pool. So #971 takes neither `half` nor anything else on account of this.

Since `Refs` is what puts a report on the fix track and tells its reporter which release carries the fix, leaving the line in would have labelled #971 and notified its reporter for a fix that is not in here. It is gone. #971 is still named throughout *Scope against #971* — as where the shape and the diagnosis came from, which is what it actually is.

`Refs #566` stays, and it is the weaker of the two: the body says plainly that this does not do what #566 asks (the wipe timer is untouched) and only changes what would be kept. If you read `Refs` strictly as "fixes", that line goes the same way — say so and I will drop it.

PathGao:
Right call on #971, and *Scope against #971* now says the useful thing: this bounds a third instance of the shape at a site the issue never names.

**`Refs #566` goes too, same reason.** #566 asks for the log to survive the action. It doesn't: `scheduleCompletedOperationCleanup` still sets `log = ""` 8 s after a success, 6 s after a cancel, 20 s after a failure, and this diff doesn't go near it. So on merge that line labels #566 `fixed (pending release)`, tells its reporter which release carries the fix, and closes his issue two weeks later — for a build where the log still vanishes while he is reading it. Worse than the #971 case, because he will actually go and check.

The 64 KB cap is a constraint on #566, not a piece of it: if the log is ever kept, the cap is what decides what "kept" means. That belongs in a comment on #566, where `Refs` semantics can't reach it.

The `log.utf8.count` gate still isn't in, and it doesn't block — a cost note, not a defect.

**Impact:** none on behaviour.
**Merge:** needs `Refs #566` dropped first — one line, and it is the same automation that made `Refs #971` wrong.

PathGao:
@PathGao The guard is in, in `db37d61c`. It was one line and you were right that
it goes in on a word.

**What changed.** `HomebrewProgressParser.appendingToLog` now reads

```swift
guard log.utf8.count > limit * 2, log.count > limit * 2 else { return log }
```

`utf8.count` is O(1) on a native string — and `log` is native, it is built by
`var log = existing; log.append(text)` — and a string never has more Characters
than UTF-8 bytes, so the byte count can only turn away chunks no trim was due
for. `log.count` runs for the few that get past it. The decision is still in
Characters, which is the unit `suffix(limit)` trims in, so the multibyte case
you asked for is unmoved. This was the half of the change that was cancelling
the other half: hoisting three regexes off the per-chunk main-thread path and
then putting an O(n) grapheme walk back on it.

**What is pinned.** Nothing observable separates the two orders — the guard is
pure screening — so a behavioural assertion cannot see it. The order is pinned
by reading the guard line out of `HomebrewSupport.swift` instead:
`Homebrew operation log screens on the O(1) byte count before walking
Characters`. Verified both ends: restoring `guard log.count > limit * 2` fails
that one alone (1 of 9491); dropping `log.count` so the byte count becomes the
decision fails it together with `Homebrew operation log measures its limit in
the unit it trims in` (2 of 9491).

**What the body says now.** The Trade-offs bullet no longer claims the cost is
unpaid, because it is paid. The earlier bullet about the unit mix now says an
earlier round decided the cut on `utf8.count` *alone*, so it does not read as
contradicting the new guard. Verification is 9491 checks and lists the new
assertion.

```sh
./build.sh --test     # TESTS OK (9491 checks)
```

Neither of the two input-method-dependent assertions came up red on this run.

**Where to look next:** whether a source-shape assertion is the right pin here
at all. It is the one thing in this PR no input can falsify, and by your own
standard in #1214 an assertion that can never be red for a behavioural reason
is close to decoration. If you would rather have nothing there than a text
match on a guard line, say so and it comes out.

PathGao:
@PathGao `Refs #566` is out — body only, no code change.

You are right that it is the same automation that made `Refs #971` wrong, and the consequence is worse here for the reason you gave: on merge that line labels #566 `fixed (pending release)` and tells its reporter which release carries the fix, for a build where `scheduleCompletedOperationCleanup` still sets `log = ""` eight seconds after a success. He would go and check.

The paragraph that carried it now says plainly that this is a constraint on #566 and not a piece of it, and that the cap belongs in a comment on the issue rather than in a `Refs` line. I have not posted that comment on #566 — say if you want it there, and what it should claim.

Nothing else on the branch moved: head is still `db37d61c`, `TESTS OK (9491 checks)`.

Your other question from that round still stands open and I have not answered it: whether a source-shape assertion is the right pin for the byte-count guard at all, given it is the one thing here no input can falsify. By the standard you set in #1214 it is close to decoration, and I would rather you decide that than have me argue for the check I just added.

PathGao:
Dropping `Refs #566` was right, and the paragraph replacing it says the useful thing.

**Don't post the #566 comment.** The cap is already documented where whoever implements #566 will meet it — `logLimit` and the comment on `appendingToLog`. On the issue it would be a second copy of that, on a thread whose reader is waiting for his log to stop vanishing. He'd read it as progress.

**Drop the source-shape assertion.** It pins one line of text, not the property. Rewriting the guard as an equivalent `if … { return log }`, renaming `limit`, or hoisting the screen into a helper all turn it red with no regression; introducing the same O(n) main-thread walk elsewhere in the parser leaves it green. Wrong in both directions.

It also makes the guard a three-copy fact — the code, the comment above it that already says "Order matters; keep the cheap test first", and now a string in a test. The comment is the better half: it says why, and it cannot go stale-red.

That leaves the screen unpinned, which is the honest state. It is an optimisation with no observable, and a comment is what records those.

**Impact:** none on behaviour.
**Merge:** needs the source-shape assertion dropped first — the check is harmless in itself, but it is the template the next unobservable optimisation gets pinned with.

PathGao:
@PathGao Dropped the byte-count shape assertion (12 lines in MetricsTests, ff9dfeaf). It pinned one line of source text: an equivalent spelling of the guard goes red, an O(n) count anywhere else stays green. Code and the HomebrewSupport.swift comment stay. TESTS OK (9490 checks).

PathGao:
Right call, and for the reason that matters: it pinned a spelling, not a property.

**The body still describes it as present**, in three places:

- *Trade-offs*: "Nothing observable separates the two orders, so a source-shape assertion pins the order instead of a behavioural one." Nothing pins it now — that was the point of dropping it.
- *Verification*: the first red-first bullet is the deleted assertion, described as failing 1 of 9491.
- The recorded run is `9491 checks`; this head is 9490.

Merged as it stands, the Verification section vouches for a check that isn't in the branch, and the next person to read it goes looking for it. The accurate line is the one you already argued: the screen is unpinned, and the comment above the guard is what records it.

Nothing else moved. The multibyte assertion still fixes the decision in Characters, and the byte screen can only turn away chunks no trim was due for.

**Impact:** none on behaviour.
**Merge:** needs the three stale sentences in the body corrected first — one number, and two claims about a check that no longer exists.

PathGao:
@PathGao Body corrected, no code change; head is still `ff9dfeaf`.

The Trade-offs bullet now says the screen is unpinned and that the comment above the guard is what records it, with one sentence on why the shape assertion was dropped. The Verification section no longer lists that assertion, and the multibyte bullet no longer claims a second failure from it. The recorded run is `9490 checks`, which is what this head produces.

Nothing was re-run: the branch did not move, and the three edits are prose about a check that is not in it.

PathGao:
All three are corrected, and the Trade-offs bullet lands on the right claim: the screen is unpinned and the comment above the guard is what records it.

**One stale sentence left, non-blocking.** Verification still says "The last round's only source change was the block comment above the `terminationHandler` in `HomebrewManager.swift`". That was true at the comment-rewrite round, but `db37d61c` changed `HomebrewSupport.swift` and `ff9dfeaf` changed `MetricsTests.swift` since. What it supports still holds — `--test` does not compile `HomebrewManager.swift` and CI covers that file — so only the round attribution is wrong, and nobody acts on it after merge.

Nothing else moved. The multibyte assertion still fixes the decision in Characters, and the byte screen can only turn away chunks no trim was due for.

**Impact:** none on behaviour.
**Merge:** ship it.

PathGao:

PathGao:
Reopened as #1295.

Issue #893, opened 2026-08-24
Paste as plain text by default


chriscarpenter12:
### Problem or Motivation

I could be misunderstanding how this feature works, but to actually paste as plain text you need to use the defined shortcut key instead of just command+V

### Proposed Solution

If enabled behave like [Pure Paste](https://sindresorhus.com/pure-paste) and have an option keyboard shortcut to past as original?

### Feature Area

Clipboard and files

AB-boi:
It works as intended but what you are asking for is essentially a feature enhancement for people that never want to retain the formatting and always paste as plain text.

chriscarpenter12:
Which is why I said I think I misunderstood the current behavior. Initially when I saw it say "Paste as plain text" I thought it'd be the default behavior. Maybe what I'm looking for is an option to behave as the default with a modifier key to paste as original. Essentially acting as Pure Paste does.

chriscarpenter12:
Was a fix for this actually released? I don't see anything in the changelog indicating a behavior difference

PR #1238, opened 2026-09-01
fix(paste-plain): read the clipboard on the shared lane and say when the paste fails

PathGao:
## Summary

Three problems in `PastePlainService.performPastePlain()`, all on the path a
press of ⌥⇧⌘V takes. The third was opened by the fix for the first.

**The clipboard was read on the main thread.** `plainText(from:)` asks
`NSPasteboard.general` for `.string`, then `.rtf`, then `.html`, and a
pasteboard read has no time limit: content can be promised and rendered only
when somebody asks for it, so a source app that is busy holds the reader for
as long as it likes. That is not a corner case here — this feature exists
precisely because the clipboard holds rich text some other app put there, and
Word, Excel and Illustrator are all promising providers. One press against a
busy source froze the app until it answered. `GeneralPasteboardAccess` exists
for this, and says so in its own header: a caller waiting on the main thread
is a frozen app. Nine other places already go through the lane, including
`ScreenshotService.openClipboardImage()` doing the same thing one directory
over. This one now uses the lane's two-argument overload and the rest of the
function runs in the completion, which is already back on the main queue.

**This is not the last main-thread read of the general pasteboard**, and an
earlier version of this body said it was. That claim is withdrawn. Two reads
remain after this branch, both in snippet expansion, both the same expression
`TextSnippetSupport.needsClipboard(...) ? NSPasteboard.general.string(forType: .string) : nil`:

- `TextSnippetService.expand` — deliberate, and documented in place: the
  expansion is decided while the triggering key event still belongs to the
  tap callback, so it cannot wait for a lane and hand the event back in time.
- `SnippetLibraryService.insert` — not documented as deliberate, and the
  typing it feeds is already deferred 0.15 s by
  `DispatchQueue.main.asyncAfter`, so nothing there needs the answer
  synchronously. It is a separate change and this branch does not touch it.

Nothing else in `Sources/` reads the general pasteboard's *contents* on the
main thread. Every other main-thread `NSPasteboard.general` is either a write
(`clearContents` / `setString` / `writeObjects`, which cannot stall on
promised content) or a `changeCount` read — `ScreenshotService.autoCopy`, twice
— which does not render a promise. `ShelfService` does read
`readObjects` / `propertyList` / `types` / `NSImage(pasteboard:)` on the main
thread, but never on `NSPasteboard.general`: its sources are the drag
pasteboard and a private one `CommandBarCatalog.keepOnShelf` creates.

The frontmost app is resolved after the read rather than before it. Both paste
routes land in whichever app is in front when the paste happens, so the target
has to be read at that moment; capturing it before the read would have pressed
a menu item in an app the user had left. What stops that from following the
user into an app they switched to is the deadline below, which does not let the
paste run at all once the frontmost app has changed.

**The completion had no deadline.** Not waiting on the lane was only half of
moving the read off the main thread. A pasteboard read has no time limit, and
neither does `GeneralPasteboardAccess`: it hands the answer back whenever the
answer arrives. So a provider that stalls no longer freezes the app, it just
pastes late — and since the target app is resolved inside the completion, a
read that comes back ten seconds later puts stale clipboard text into whatever
app the user has moved to by then. That trades a freeze for a paste nobody
asked for, in a window nobody was looking at.

The lane has no timeout facility for callers to share; the two other callers
that need one bound their own completion. `ClipboardHistoryService` does it
with a generation counter plus a 5 s `scheduleCaptureTimeout`, which frees its
in-flight flag and lets the stale completion identify itself as abandoned. This
call site needs no counter — there is nothing to reset and no second request to
distinguish — so it carries the same idea in the two values a stale completion
can be recognised by, both captured on the press:

- `ProcessInfo.processInfo.systemUptime`, monotonic so a clock change cannot
  move the deadline, bounded at 2 s. Short because this sits on a key press:
  past two seconds the user has already read the shortcut as having failed, and
  a paste arriving after that is a surprise whatever app it lands in.
- `NSWorkspace.shared.frontmostApplication?.processIdentifier`, the app the
  shortcut was pressed in. `performPastePlain` runs on the main thread —
  `QuickToolHotkey`'s Carbon handler dispatches every press to
  `DispatchQueue.main` — so the frontmost app is readable there, on the press,
  not only in the completion.

Both have to still hold or the completion returns without pasting. A deadline
alone would only stop the waiting; it would not stop the late completion from
pasting when it finally ran. Refusing on either is what makes a late arrival a
no-op instead of a misdirected paste, and it is the app check that closes the
wrong-window case, since a stalled provider that answers at 1.9 s is inside the
deadline and can still land in an app the user has left.

A refused completion beeps. It is the only failure on this path the user waited
seconds for, so of all of them it is the one that should not be silent.

The predicate is `QuickToolsSupport.pastePlainReadIsStillCurrent(elapsed:requestedApp:frontmostApp:)`,
next to `isMatchStyleEquivalent` in the same MARK section. It is there rather
than inline because `QuickToolsSupport.swift` is in the list `./build.sh --test`
compiles and `PastePlainService.swift` is not, so the rule can be asserted by
running it instead of by grepping the source.

**The other `TransientPaste` caller does not have this exposure.**
`TextSnippetService` is the only other one, and its clipboard read is
synchronous on the main thread (deliberate and documented: the expansion is
decided while the triggering key event still belongs to the tap callback). With
no async hop between the trigger and the paste there is no completion that can
arrive late, so there is nothing to bound, and it does not route through
`performPastePlain` — no shared function to fix once. What both callers do share
is the lane read inside `TransientPaste.paste` itself, which snapshots every
type of every pasteboard item and can stall on the same promised content. The
guard in `performPastePlain` does not cover that read: it runs before the
paste is queued, so a snapshot that stalls after it has passed still ends in a
⌘V posted seconds later, into whatever app is in front by then. That late
paste is a failure this branch introduced by making the first read
asynchronous, so the same test is asked once more where it matters.
`TransientPaste.paste` takes an optional `stillWanted: (() -> Bool)?` and asks
it in `postPasteShortcut`, right before the ⌘V key-down is posted — after
`setString`, so a refusal goes through the existing `completion(false)` path:
`didFail` beeps, and `scheduleRestore` puts the original clipboard back the
same way it does after a paste. `PastePlainService` passes the same pid
predicate, measured from the same press, but against its own limit,
`QuickToolsSupport.pastePlainPostDeadline`, not the read's. Raised in review:
between the two checks sit the paste's own waits, and the read's 2 s cannot
absorb them. The read may have used its whole deadline; the first press in
an app walks its menu bar once in `pressNativeMatchStyleItem`; and
`postPasteWhenModifiersReleased` waits up to 100 × 0.015 s + 0.06 s for the
shortcut's own modifiers to come up. None of that is a stall, so a first
⌥⇧⌘V held a beat would have beeped at the post instead of pasting. The
paste-leg limit is the read deadline, plus that modifier wait
(`pastePlainModifierReleaseBound`, 1.56 s, repeated in `QuickToolsSupport`
because `TransientPaste.swift` is outside the `--test` list, and pinned to
the now-named poll constants in `TransientPaste` by an assertion that reads
them back), plus one second for the menu walk in a responsive app: 4.56 s.
The walk's own worst case, 600 elements × 0.35 s of AX timeout, is not
budgeted here and is bounded where it lives, by its element cap and
messaging timeout; a walk that eats the whole second is a target whose AX is
stalling, and a beep after four and a half seconds is the right report for
that press. The point of the post-point check is the pid comparison plus a
bound that a snapshot read stalled for seconds still trips, not a tight one.
`TextSnippetService` passes nothing, and a nil `stillWanted` never refuses,
so its behavior is unchanged; its snapshot read stays unbounded, and bounding
it would change what happens when a snippet fails to expand. That is a
separate change and this branch does not take it.

**Every way the paste could fail was silent.** `TransientPaste.paste` gives up
six ways — another paste already running, an incomplete pasteboard snapshot,
the change count moving while the snapshot is read, `setString` failing, the
CGEvent not being constructible, and being called off the main thread — and
this call site discarded the return value and passed no `didFail`. All six
looked the same from outside: press the shortcut, nothing happens. Twelve
lines further up, the Accessibility guard in this same function already beeps
rather than swallow the shortcut for that exact reason, and
`TextSnippetService`, the only other caller of `TransientPaste`, passes
`didFail`.

Now `didFail` reports the four the paste discovers on its lane, with a beep
for each. The synchronous `false` deliberately does not beep. It carries two
rejections — called off the main thread, and a paste already in flight — and
this call site can only ever see the second one: `pastePlain(_:)` runs inside
the lane completion, which `GeneralPasteboardAccess` dispatches to the main
queue, so the `Thread.isMainThread` guard cannot trip here. A paste already in
flight still ends in a paste of this same clipboard — `TransientPaste` holds
`isPerforming` true through `postPasteWhenModifiersReleased`, which polls up
to ~1.5 s while the shortcut's own modifiers are still down, and a second
press inside that window is coalesced into the first. Beeping there put an
error sound on a press that worked. The first version of this branch did beep
on it; that beep is gone.

`TextSnippetService` keeps reading the same `false` as failure and is
unchanged. For it a busy lane means the snippet genuinely did not expand and
the trigger key has to fall through to normal typing, which is what its
`? nil : Unmanaged.passUnretained(event)` already does; it does not beep on
either rejection, so there is no second site with this defect.

Considered and not taken: reading the pasteboard on the main thread but with a
timeout — `NSPasteboard` has no such API, which is why the serial lane exists
at all. Also considered showing a `QuickToolHUD` instead of beeping on
failure; the beep matches what the permission guard four lines up already
does, and a HUD for a paste that did not happen is more interruption than the
event is worth.

Raised in review and settled there: saying something when the clipboard holds
nothing this feature can flatten. `guard let plain, !plain.isEmpty else { return }`
still returns silently, so ⌥⇧⌘V with an image on the clipboard does nothing and
says nothing. Reviewed and closed with plain ⌘V as the precedent — it is silent
against a clipboard with no text too — so no follow-up pull request is opened
for it and the guard stays as it is.

## Verification

- `./build.sh --test` on this branch, merged with `main`: `TESTS OK (9612 checks)`, no failures.
  Not compared against a base count: the two assertions that read the machine's
  live input source move with the active input method, so a count is only
  meaningful measured on the same machine at the same moment. The failure list
  is empty, which is the claim that matters.
- `QuickToolsSupport.swift`, which holds the new predicate, is in the list
  `./build.sh --test` compiles, so the six assertions below that call
  `pastePlainReadIsStillCurrent` run the real function rather than reading
  its source.
- `PastePlainService.swift` and `TransientPaste.swift` are outside that list
  (it holds no `*Service.swift`, and not `TransientPaste.swift` either), so
  they are typechecked separately: `swiftc -typecheck` over all 427 files of
  `Sources/Vorssaint` with the target, SDK and include flags `build.sh` uses.
  Exit 0, no errors and no warnings. Their compiled build is left to CI on
  both runners.
- The full `./build.sh` ran after the merge with `main`, on the head before
  the paste-leg limit was split (c5a4ece): exit 0, no warnings. It has not
  been re-run since; the split touched no file the typecheck above does not
  cover.
- Every assertion on this branch was confirmed red before being kept, one
  broken edit at a time, each reverted to green afterwards:
  - putting the read back on `NSPasteboard.general` reddens the lane assertion
    and leaves the others green;
  - restoring `_ = TransientPaste.shared.paste(...)` without `didFail` reddens
    the reporting assertion;
  - restoring `if !started { NSSound.beep() }` reddens only the coalesced-paste
    assertion;
  - dropping `elapsed < pastePlainReadDeadline` from the predicate reddens only
    *a read that answers long after the press is abandoned in that app too*;
  - dropping `requestedApp == frontmostApp` reddens only *a read landing after
    an app switch does not paste into the new app*;
  - deleting the guard from `performPastePlain` while leaving the predicate in
    place reddens both the wiring assertion and the beep count, which is the
    pair that catches a predicate that exists but is never called;
  - the two `stillWanted` assertions were written before the code and ran
    red first (`TESTS FAILED (2 of 9608)`): one pins the `stillWanted` check
    inside `postPasteShortcut` ahead of `keyDown.post`, the other pins
    `PastePlainService` passing it in;
  - the four paste-leg assertions went red one break at a time: setting
    `pastePlainPostDeadline` equal to the read deadline reddens only *a paste
    whose modifiers were held past the read deadline still posts*; making
    the call site pass the read deadline to the paste leg reddens only *the
    read leg and the paste leg are checked against their own limits*;
    changing `TransientPaste`'s `postDelay` from 0.06 to 0.05 reddens only
    the assertion that reads the poll constants back; dropping
    `elapsed < limit` from the predicate reddens the two paste-leg predicate
    assertions together with the read's *long after the press* one, which is
    the three that share the bound.

**Not tested.** I did not run the app. The freeze this fixes was reasoned from
the code and from the rule `GeneralPasteboardAccess` already states, not
reproduced against a stalled provider. The late paste the deadline prevents was
likewise not staged: it needs a promising app that stops answering for seconds
on demand, which I cannot arrange, so the two seconds, the 4.56 s and the
pid comparison are argued from the code path and asserted as a pure
function, not measured against a real stall. The one second allowed for the
menu walk is a judgement, not a measurement: I have not timed the walk in
any app. I have not heard any of the beeps.

`pressNativeMatchStyleItem` still walks the front app's menu bar on the main
thread, bounded at 600 elements × 0.35 s of AX timeout and cached per pid. That
walk is pre-existing and this branch does not shorten it. What changed (33debfac)
is what happens after it: the function now takes `stillWanted` and asks it
right before `kAXPressAction`, on `pastePlainPostDeadline`, so a walk that ran
long or ended after an app switch beeps and returns true instead of pressing an
item on the pid captured before the walk. Returning true rather than false is
deliberate: the synthesized fallback would only post the same late clipboard
into the same wrong app. That is the fourth beep in the file, and the fourth
paste-leg assertion pins the check ahead of the press.

The limit that leg accepts is looser than its own path: `pastePlainPostDeadline`
includes `pastePlainModifierReleaseBound`, and a menu press never waits on
modifiers, so the native leg can accept up to ~4.5 s where it can only spend
about 3. Loose rather than wrong; the pid half of the check carries it, and one
limit for both legs beats a third constant.

## Anything else

Refs #887

#893 reports not being able to tell how paste-as-plain-text behaves and asks for it to be
the default. The default is not changed here. What this adds is the missing feedback — a
silent no-op now says it failed — which is the half of that report about not knowing
whether the feature ran.

Refs #893


PathGao:
Resolving the frontmost app after the read rather than before is the right call, and the easiest thing to get wrong when a sync read goes async.

**"This was the last main-thread read" doesn't hold.** `SnippetLibraryService.insert` still does

```swift
? NSPasteboard.general.string(forType: .string)
```

on the main thread, same promised-content stall. `TextSnippetService.expand` has the same read but documents it as a deliberate trade; the snippet library one doesn't. A separate PR for it is fine — the claim in the body isn't.

**An empty or non-text clipboard stays silent.** `guard let plain, !plain.isEmpty else { return }`. By your own argument for the beep, pressing ⌥⇧⌘V with an image on the clipboard is a seventh way for the feature to do nothing without saying so.

**Not verified:** `pressNativeMatchStyleItem` still walks the front app's menu bar on the main thread, bounded at 600 elements × 0.35 s of AX timeout and cached per pid. Pre-existing and untouched here, but it sits on the same key press as the freeze you're fixing. I read `main` for the other call sites; I ran nothing.

**Impact:** none. The diff is sound.
**Merge:** ship it.

ref #887

PathGao:
Resolving the frontmost app inside the completion rather than before the read is the right call — the target has to follow an app switch that happens mid-read.

**"This was the last main-thread read" is wrong.** `SnippetLibraryService.insert` still does `needsClipboard(...) ? NSPasteboard.general.string(forType: .string) : nil` on main, and the typing it feeds is already deferred 0.15 s, so nothing there needs the answer synchronously. `TextSnippetService.expand` has the same expression, but its comment argues the tap callback needs it now. Narrow the claim, or take the snippet library too.

**`isPerforming` is the one of the six that shouldn't beep.** `TransientPaste` holds it true through `postPasteWhenModifiersReleased`, which polls up to ~1.5 s while the shortcut's own modifiers are still down. A second press in that window beeps, then the first paste lands anyway. The other five end with nothing pasted; this one doesn't.

Not verified: I didn't run it either, and `pressNativeMatchStyleItem` still walks the AX menu bar on main (600 elements × 0.35 s cap) right after the completion — bounded, unlike the read, but the path isn't off main end to end.

**Impact:** none on the fix. The beep fires on a paste that still succeeds.
**Merge:** needs the "last main-thread read" claim corrected first.

ref #887

PathGao:
@PathGao Both are addressed — one in the body, one in the code.

**The claim is withdrawn and replaced with what I actually found.** I grepped
every `NSPasteboard.general` in `Sources/` and split them by what they do. Two
main-thread reads of the general pasteboard's contents remain, exactly the two
you named: `TextSnippetService.expand`, deliberate and documented in place, and
`SnippetLibraryService.insert`, which is not documented and whose typing is
already deferred 0.15 s. The body now names both and says the snippet library
one is a separate change this branch does not take. Everything else that
touches `NSPasteboard.general` on main is a write or a `changeCount` read
(`ScreenshotService.autoCopy`, twice); neither renders a promise.
`ShelfService` does read `readObjects` / `propertyList` / `types` on main, but
on the drag pasteboard and on the private one `CommandBarCatalog.keepOnShelf`
makes, never on `.general`.

**The beep on `isPerforming` is gone.** You are right that it is the one of the
six that shouldn't fire, and it turned out to be the only one the synchronous
return could reach here: `pastePlain(_:)` runs inside the lane completion,
which `GeneralPasteboardAccess` dispatches to the main queue, so the
`Thread.isMainThread` guard can never trip at this call site. `if !started
{ NSSound.beep() }` was therefore beeping on nothing but the coalesced press.
Removing it leaves all four genuine failures reported through `didFail` and
nothing reporting a paste that succeeded.

I checked the other caller before touching anything shared, and deliberately
left `TransientPaste` alone. `TextSnippetService` reads the same `false` and
must keep reading it as failure — a busy lane there means the snippet did not
expand and the trigger key has to fall through to normal typing — and it does
not beep on either rejection, so there is no sibling site with this defect and
no reason to widen `paste`'s return type.

Verified: `./build.sh --test` gives `TESTS OK (9490 checks)`, no failures.
`PastePlainService.swift` is outside the `--test` whitelist, so I typechecked
it with `swiftc -typecheck` over all 425 files of `Sources/Vorssaint` using
`build.sh`'s target, SDK and include flags — exit 0, no errors, no warnings.
The new assertion was confirmed red first: restoring `if !started
{ NSSound.beep() }` fails it and leaves the other two green. I did not run the
full `./build.sh` — codesign blocks on a locked keychain here — and I did not
run the app, so the removed beep was traced through `isPerforming` and
`postPasteWhenModifiersReleased` rather than provoked by double-pressing the
shortcut. `pressNativeMatchStyleItem` is still on main and still untouched;
the body says so.

On the empty or non-text clipboard: I left `guard let plain, !plain.isEmpty
else { return }` silent and said why in the body rather than adding a seventh
beep. ⌘V with no text is silent too, and this is a behaviour change to a path
the branch otherwise leaves alone, so it reads better as its own pull request.
Say the word and I will open one.

One thing for you to settle: this PR currently carries both `incomplete` and
`ready to merge`.

PathGao:
Both fixed, and the body now says what's actually in `Sources/` instead of what we assumed. 

One thing the async hop opens: **the completion has no deadline.** `ClipboardHistoryService` guards its lane reads with a generation token and `scheduleCaptureTimeout` at 5 s, exactly so a wedged read's completion is abandoned instead of acting late. `performPastePlain` has none — and since the frontmost app is resolved *inside* the completion (right for a 50 ms read), a read that answers 10 s later pastes stale clipboard text into whatever app the user has moved to by then. That trades a freeze for an unrequested paste, and it's the one failure left that says nothing, which is the half of this PR I'd least want silent.

```swift
let deadline = ProcessInfo.processInfo.systemUptime + 2
GeneralPasteboardAccess.shared.async({ Self.plainText(from: .general) }) { [weak self] plain in
    guard ProcessInfo.processInfo.systemUptime < deadline else { NSSound.beep(); return }
```

Not verified: I read `main` for the sibling guard and for `TransientPaste`'s four `didFail` paths; I ran nothing.

Labels: drop `ready to merge`, keep `incomplete`. Empty clipboard: don't open that PR — plain ⌘V is silent there too, that's precedent enough.

**Impact:** a stalled provider can land a paste in an app the user has since switched to.
**Merge:** needs a deadline on the lane completion first.

ref #887
ref #893

PathGao:
@PathGao Deadline added, and it does more than stop waiting.

**The gap.** `performPastePlain` had no bound on the lane completion, so the
async hop that fixed the freeze moved the cost rather than removing it: a
stalled provider answers whenever it likes, and because the target app is
resolved inside the completion, a read that returns ten seconds later pastes
into whatever app the user moved to. Your `ClipboardHistoryService` comparison
is the right one — that is exactly what its generation counter plus
`scheduleCaptureTimeout` prevents.

**What closed it.** The lane has no shared timeout facility; both callers that
need one bound their own completion, so this one does too. It needs no
generation counter — nothing to reset, no second request to tell apart — so it
captures the two things a stale completion is recognised by, both on the press:
`ProcessInfo.processInfo.systemUptime`, bounded at your 2 s, and
`NSWorkspace.shared.frontmostApplication?.processIdentifier`. `performPastePlain`
runs on main (`QuickToolHotkey`'s Carbon handler dispatches every press to
`DispatchQueue.main`), so the frontmost app is readable there and not only in
the completion. Both must still hold or the completion returns without pasting,
and beeps.

The second half is the part I'd argue for beyond your snippet. A deadline on
its own stops the waiting but not the acting: a read that answers at 1.9 s is
inside the deadline and can still land in an app the user has left. The pid
comparison is what turns a late arrival into a no-op rather than a misdirected
paste. Resolving the target inside the completion stays correct — it is now
only reachable while the target is the app the press happened in — and the doc
comment there says so instead of the old claim that an app switch moves the
target with it.

**Where else the shape appears.** `TextSnippetService` is the only other
`TransientPaste` caller and does not have this exposure: its clipboard read is
synchronous on main, so there is no completion that can arrive late, and it
does not route through `performPastePlain`. What both callers do share is the
lane read inside `TransientPaste.paste` itself — the snapshot walks every type
of every item and can stall on the same promised content. On this path it is
now unreachable while stalled, since the guard runs first; for snippet
expansion it is unbounded. Bounding it changes what a failed expansion does, so
I left it and named it in the body rather than widen this.

**Verification.** `./build.sh --test` gives `TESTS OK (9495 checks)`, no
failures. I did not compare against a base count — the two assertions that read
the machine's live input source move with the active input method, so the empty
failure list is the claim, not the number. The predicate lives in
`QuickToolsSupport`, which `--test` compiles, so the four new assertions run it
rather than grep it; `PastePlainService.swift` is outside that list and is
typechecked with `swiftc -typecheck` over all 425 files of `Sources/Vorssaint`
using `build.sh`'s target, SDK and include flags — exit 0, no errors, no
warnings, compiled build left to CI on both runners. No full `./build.sh`:
codesign blocks on a locked keychain here.

Each new assertion was confirmed red first, one broken edit at a time. Dropping
`elapsed < pastePlainReadDeadline` reddens only the "answers long after the
press" one; dropping `requestedApp == frontmostApp` reddens only the "landing
after an app switch" one; deleting the guard from `performPastePlain` while
leaving the predicate in place reddens both the wiring assertion and the beep
count, which is the pair that catches a predicate that exists but is never
called. All green again after reverting.

**Not verified.** I did not run the app and did not stage the stall — it needs
a promising app that stops answering for seconds on demand. The 2 s and the pid
comparison are argued from the code path and asserted as a pure function, not
measured against a real provider. `pressNativeMatchStyleItem` is still on main
and still untouched.

The empty-clipboard question is closed per your ruling: no follow-up PR, plain
⌘V's silence is the precedent, and the body now records that rather than
offering to open one. I have not touched the labels.

PathGao:
The pid comparison is the right addition — a deadline alone would have let a 1.9 s read paste into an app the user had left.

**The guard is on the wrong read.** It bounds `plainText`, which asks for three types. The read that stalls is `TransientPaste.snapshot` — `item.data(forType:)` for every type of every item, which is exactly where promised content renders. It goes on the lane *after* your guard passes, with no deadline and no pid check. The body says it is "unreachable while stalled, since the guard runs first"; the guard runs before it is enqueued, not before it stalls.

**And the window is wider than the snapshot.** `postPasteWhenModifiersReleased` polls up to 100 × 0.015 s before posting, so even with an instant snapshot ~1.6 s separates the pid check from `keyDown.post`.

The check has to sit at the last point before the post:

```swift
func paste(_ text: String, stillWanted: (() -> Bool)? = nil, …)
// postPasteShortcut, before willPost?():
guard stillWanted?() ?? true else { completion(false); return }
```

nil for `TextSnippetService`. It vetoes after `setString`, so that path still has to restore.

Not verified: I read `main` for `TransientPaste` and the command bar call site; I ran nothing.

**Impact:** a stalled snapshot or held modifiers can still land the paste in an app the user switched to.
**Merge:** needs the check moved to just before the ⌘V post first.

ref #887
ref #893

PathGao:
@PathGao Merged main (Tests/MetricsTests.swift add/add, both sides kept) and moved the late-paste check to where it matters: TransientPaste.paste takes an optional stillWanted closure, asked in postPasteShortcut right before the ⌘V key-down; a refusal takes the existing completion(false) path (beep, clipboard restored). PastePlainService passes its 2 s + pid predicate, TextSnippetService passes nothing and is unchanged. Two new assertions ran red first (TESTS FAILED (2 of 9608)); full ./build.sh exit 0, TESTS OK (9608 checks).

PathGao:
Moving the check into `postPasteShortcut` is the right spot, and routing the refusal through `completion(false)` means `didFail` and `scheduleRestore` already do the right thing — no new failure path added.

**The 2 s now spans the code's own waits.** Between the two checks sit `pressNativeMatchStyleItem`'s AX menu walk (up to 600 elements × 0.35 s messaging timeout, cached only after the first miss per pid), the snapshot lane read, and `postPasteWhenModifiersReleased` polling 100 × 0.015 + 0.06 s while the shortcut's own modifiers are still down. None of that is the stall the deadline was argued against — "past two seconds the user has already read the shortcut as having failed" — but all of it counts against it. First ⌥⇧⌘V in an app with a deep menu bar, held a beat, and elapsed is past 2 at the post point: beep, nothing pasted. My ask did this — the clock came along with the check.

Give the paste leg its own limit instead of reusing the read's:

```swift
static func pastePlainReadIsStillCurrent(
    elapsed: TimeInterval,
    limit: TimeInterval = QuickToolsSupport.pastePlainReadDeadline, …)
// paste leg passes limit: pastePlainReadDeadline + 1.56 — the poll's own bound
```

Not verified: I read `main` for the walk and the poll; I ran nothing.

**Impact:** a slow first press in an app beeps instead of pasting.
**Merge:** needs the post-point deadline split from the read's first.

ref #887
ref #893

PathGao:
@PathGao The post-point check has its own limit now. `pastePlainReadIsStillCurrent` takes `limit:` defaulting to `pastePlainReadDeadline`; the read leg keeps that, the paste leg passes `QuickToolsSupport.pastePlainPostDeadline`. Both measure from the press and keep the pid comparison.

The paste-leg limit is derived, not a literal: read deadline (2 s) + `pastePlainModifierReleaseBound` (100 × 0.015 s + 0.06 s = 1.56 s, the wait `postPasteWhenModifiersReleased` makes) + 1 s for the first-press menu walk in a responsive app, 4.56 s. The modifier bound is repeated in `QuickToolsSupport` because `TransientPaste.swift` is outside the `--test` list; the poll literals in `TransientPaste` are now named constants and an assertion reads them back from the file and compares the sum, so the two cannot drift. The walk's own worst case (600 × 0.35 s) is not budgeted: it is bounded by its element cap and messaging timeout where it lives, and a walk that eats the whole second is a target whose AX is stalling, which a beep after 4.5 s reports correctly. The body's trade-offs paragraph says the same.

Verification: `./build.sh --test` on 3986f0de, `TESTS OK (9612 checks)`, four new. Red first, one break at a time, restored to green after each: `pastePlainPostDeadline = pastePlainReadDeadline` reddens only *a paste whose modifiers were held past the read deadline still posts*; the call site passing the read deadline to the paste leg reddens only *the read leg and the paste leg are checked against their own limits*; `postDelay` 0.06 → 0.05 in `TransientPaste` reddens only the read-back assertion; dropping `elapsed < limit` from the predicate reddens the two paste-leg predicate assertions plus the read's *long after the press* one (3 of 9612). `swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk -I Sources/VMStatisticsCompat -I Sources/HIDEventSystem` over all 427 files of `Sources/Vorssaint`: exit 0, no output. `PastePlainService.swift` and `TransientPaste.swift` are typechecked only, not compiled; the full `./build.sh` was not re-run on this head.

Not verified: the app was not run. The 4.56 s and the one-second walk allowance are argued from the code, not timed in any app; I have not measured a menu walk or heard a beep.

ref #887
ref #893

PathGao:
Splitting the limit is right, and deriving it from `pastePlainModifierReleaseBound` with a read-back assertion beats the literal I suggested.

**The native menu route has no late check.** `pressNativeMatchStyleItem()` runs first in `pastePlain` and returns before `TransientPaste` is reached, so `stillWanted` never applies to it. It resolves the front app itself, walks the menu bar (600 elements × 0.35 s, uncached on the first press per pid — and uncached on every press when `kAXMenuBarAttribute` itself times out, since that path returns without recording a miss), then presses an item on the pid it captured before the walk. A press on a backgrounded app still lands there. The body says the deadline "does not let the paste run at all once the frontmost app has changed" — that holds for the ⌘V leg only.

The closure is in scope one line up:

```swift
guard stillWanted() else { NSSound.beep(); return true }
return AXUIElementPerformAction(item, kAXPressAction as CFString) == .success
```

Not verified: I read `main` for the walk and its cache; I ran nothing.

**Impact:** a slow menu walk can press Paste and Match Style in the app the user just left.
**Merge:** needs the same check on the native menu route first.

ref #887
ref #893

PathGao:
@PathGao The native route now takes stillWanted and asks it right before kAXPressAction (33debfac), on the post-point limit. A refusal beeps and returns true rather than falling through to the synthesized paste, which would only land the same late clipboard in the same wrong app. Red first: the new assertion and the beep count (3 to 4) both failed against the old source, TESTS FAILED (2 of 9613); after the change, full ./build.sh exit 0, TESTS OK (9613 checks).

PathGao:
Returning true on refusal rather than falling through is the right call — the fallback would only have posted the same late clipboard into the same app.

**The body still says this function is untouched.** "Also untouched and unverified … `pressNativeMatchStyleItem` still walks the front app's menu bar … and it is pre-existing. It runs before the paste-leg check" — it now takes `stillWanted`, beeps, and returns true without pressing. Someone reading that paragraph won't go looking for a fourth beep inside it. One paragraph to fix.

**The native leg is budgeted for a wait it never makes.** `pastePlainPostDeadline` is read + `pastePlainModifierReleaseBound` + walk, but a menu press doesn't wait on modifiers — the comment right above says held keys don't matter to it. So this leg accepts up to 4.56 s when its own path can only spend about 3. Loose rather than wrong; the pid half still carries it.

Not verified: I read the diff and `main` for the cache path; I ran nothing.

**Impact:** none on behaviour. The body under-describes what the branch now changes.
**Merge:** needs the "untouched" paragraph corrected first.

ref #887
ref #893

PathGao:
@PathGao Body corrected: the paragraph that called pressNativeMatchStyleItem untouched now describes the stillWanted check ahead of the press, the fourth beep, and why a refusal returns true. The budget point is written in as loose rather than wrong: the native leg never waits on modifiers, so it accepts about 4.5 s where it can spend about 3; the pid half carries it and I kept one limit for both legs rather than add a third constant. No code change on this push.

PathGao:
Body reads right now — the paragraph names the `stillWanted` check ahead of the press, the fourth beep, and why a refusal returns true, and the budget point is written in as loose rather than wrong.

Walked the refusal path once more on this head. The guard in `postPasteShortcut` sits ahead of `willPost?()`, so a refused paste never unregisters the hotkey and there is nothing left half-done; `completion()` still runs either way, so `scheduleRestore` puts the original clipboard back the way a failed post does. Four `NSSound.beep()` sites in `PastePlainService`, matching what the assertion pins. Nothing left to flag.

Not verified: I read the diff and `main`; I ran nothing. CI is green on both runners.

Labels updated to match: `incomplete` off, `ready to merge` on.

**Impact:** none.
**Merge:** ship it.

ref #887
ref #893

PathGao:

PathGao:
Reopened as #1301.

Issue #396, opened 2026-07-30
[Bug]: Eject all disks doesn't seem to work


kimonostereo:
### Description

I've tried multiple times to use the Eject all disks feature and it doesn't seem to work. It always says `No external disk ready to eject` even though there are multiple disks attached to my MacBook Pro.

### Steps to Reproduce

1. Attache external storage
2. Open Vorssaint
3. Click on quick toggles icon
4. Choose Eject all disks

Disks are not ejected.

### Vorssaint Version

3.1.15

### macOS Version

macOS 26 (Tahoe)

### Installation Method

DMG

vorssaint:
3.2.0 goes by the bus the drive hangs off instead of asking for removable media, so a fixed external should show up now. Only a physical drive reproduces this and I do not have one here, so let me know what you get. Not closing yet.

kimonostereo:
I updated and tried it with both SD card and external hard drives and SSD on my MacBook Pro and it worked to eject all of them. Thanks for the update!

PR #1239, opened 2026-09-01
fix(quick-tools): honour excluded volume UUIDs everywhere

PathGao:
## Summary

**A volume UUID in the eject exclusion list was honoured in some places and
ignored in others.** `Defaults` documents that key as "volume names/UUIDs
excluded from Eject all disks" and `QuickTogglesSupport.isExcluded` matches
four forms: name, UUID, full mount path and mount directory name. Three
places ask the question, and two of them asked it weakly.

`volumeUUID` and `mountPath` were declared with a default of `nil`, so a
caller could omit them and get a quietly weaker match with nothing to flag it.
`QuickTogglesService` passed all four. `DiskProtectionService` — the panel's
disk section, which is where the eject buttons live — passed only name and
mount path. Someone who typed a volume UUID into the exclusion list got it
respected by "Eject all disks" and ignored by the panel. Rather than fix that
one call site, the two defaults are gone, so the omission no longer compiles.
That is what pointed at `DiskProtectionService`, and carrying the UUID over to
it meant `DiskDeviceReading` had to have one: `DiskSampler` already builds
every reading from `FileManager.mountedVolumeURLs`, so it is one more resource
key (`.volumeUUIDStringKey`) and one more field passed through. `MountedVolume`
and `DiskDeviceReading` are unchanged for every other caller — the optional
field keeps them omittable in the memberwise initializer, so the synthetic
preview disk in `MenuBarRenderer` needed no edit.

The third place the compiler could not reach.
`DiskExclusionsList.mountedCandidateDrives`, the picker in Settings that offers
connected drives to add, did not call `isExcluded` at all: it compared the
volume's name against the list and nothing else. A drive excluded by UUID —
or by mount path — kept being offered as a candidate, as if it were not
excluded. It now reads `.volumeUUIDStringKey` like the other two enumerations
and asks `isExcluded` with name, UUID and mount path, so all three sites give
the same answer for the same drive. Those are every site: a grep for the
exclusion set finds `QuickTogglesService`, `DiskProtectionService` and this
picker; `addExcludedVolume` and `removeExcludedVolume` also compare strings,
but they deduplicate the user's own list entries rather than test a drive.

Considered and not taken: giving `DiskProtectionService` an explicit
`volumeUUID: nil` with a comment saying why it cannot pass one. That is a
smaller diff, but it leaves the user-visible inconsistency in place and turns
a bug into a documented bug. Also not taken: removing the defaults from
`shouldOfferEject` as well. Its four exclusion arguments are one coherent
all-or-nothing group, and `DiskExclusionsList` deliberately calls it with none
of them because it only wants the "is this drive ejectable at all" half — the
exclusion half is now the separate `isExcluded` call right below it.

## Verification

- `./build.sh --test` on this branch: `TESTS OK (9489 checks)`, no failures.
  That count is one lower than the previous push of this branch, which is the
  wake assertion that left with its fix. The branch's base is `ed882cd3`; I did
  not re-run `--test` there in this round, so I am not quoting a baseline count.
- Of the files this branch changes, only `QuickTogglesSupport.swift`,
  `DiskSupport.swift` and the tests are in the whitelist `--test` compiles.
  `DiskSampler.swift`, `DiskProtectionService.swift` and
  `DiskExclusionsList.swift` are not compiled here; CI covers them on both
  runners.
- The removed defaults were confirmed to be the guard they are meant to be: a
  reduced two-function case compiles silently when `volumeUUID` has a default
  and errors with `missing argument for parameter 'volumeUUID' in call` when
  it does not.
- The picker assertion is a source-shape check, because the file is outside
  the `--test` whitelist. Rather than spend a second build on it, its predicate
  was evaluated directly against the pre-fix content of the file
  (`git show HEAD~1:…/DiskExclusionsList.swift`, comment lines stripped the
  same way the assertion strips them): zero matches before the change, both
  matches after. The assertion is red on the old file and green on the new one.

**Not tested.** I did not run the app. I have no second volume with a UUID
mounted here, so neither the panel nor the picker honouring a UUID exclusion
was confirmed on hardware — only that the value now reaches `isExcluded` in
all three places. `./build.sh` was not run here: with the signing keychain
locked, codesign blocks on an unlock dialog and never exits, so the release
build and the zero-warning check are left to CI.

## Anything else

This branch opened with two unrelated small fixes on it. The share-link expiry
one — nine lines observing `NSWorkspace.didWakeNotification` in
`ScreenshotShareService`, plus its assertion — has been taken off and sent as
#1266, so a question about the shared signature here does not hold it behind.
Nothing that stays touches that file, and the two share no code.

No issue in the tracker describes this defect. The closest, #396
("Eject all disks doesn't seem to work"), is closed and was a different
cause — the removable/ejectable flags hiding every fixed external drive —
so I have not referenced it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 - Cleaning Mode now offers an option in Settings to keep the screen visible with a discreet corner indicator instead of blacking out the screen.
 - Mouse button shortcuts can now switch Spaces or open Mission Control and App Exposé by holding an extra button and dragging. Thanks to @iltonandrew.
 - Window Layout now offers configurable window and screen gaps, so snapped windows can keep a preset distance from each other and from the screen edge. Thanks to @marcelharinck.
-- Eject all disks can now exclude specific drives in Settings, keeping backup drives and permanent storage mounted when ejecting other external disks.
+- Eject all disks can now exclude specific drives in Settings, keeping backup drives and permanent storage mounted when ejecting other external disks. Thanks to @PathGao.
 - Dock Preview now includes an option in Settings to quit an app from a thumbnail's × button instead of closing only that window, and closes the preview as soon as the quit request is accepted. Thanks to @arefshal.
 - Copy Text from Screen now includes an option in Settings to remove line breaks and join recognized lines as a single paragraph with script-aware spacing. Thanks to @ywu73.
 - Clipboard history now offers retention limit options for 10,000 items and unlimited storage.
--- a/Sources/Vorssaint/Services/Metrics/DiskProtectionService.swift
+++ b/Sources/Vorssaint/Services/Metrics/DiskProtectionService.swift
@@ -51,6 +51,7 @@ final class DiskProtectionService: ObservableObject {
         guard !excludedVolumes.isEmpty else { return false }
         let set = Set(excludedVolumes.map { $0.lowercased() })
         return QuickTogglesSupport.isExcluded(volumeName: disk.name,
+                                              volumeUUID: disk.volumeUUID,
                                               mountPath: disk.mountPath,
                                               excludedVolumes: set)
     }
--- a/Sources/Vorssaint/Services/Metrics/DiskSampler.swift
+++ b/Sources/Vorssaint/Services/Metrics/DiskSampler.swift
@@ -53,6 +53,7 @@ final class DiskSampler {
             return DiskDeviceReading(id: volume.mountPath,
                                      name: volume.name,
                                      mountPath: volume.mountPath,
+                                     volumeUUID: volume.volumeUUID,
                                      bsdName: metadata.bsdName,
                                      wholeDisk: wholeDisk,
                                      ioCounterID: counter?.id,
@@ -164,6 +165,7 @@ final class DiskSampler {
     private struct MountedVolume {
         var name: String
         var mountPath: String
+        var volumeUUID: String?
         var totalBytes: UInt64
         var freeBytes: UInt64
         var rawFreeBytes: UInt64?
@@ -187,6 +189,7 @@ final class DiskSampler {
             .volumeIsRemovableKey,
             .volumeIsEjectableKey,
             .volumeIsLocalKey,
+            .volumeUUIDStringKey,
         ]
         guard let urls = FileManager.default.mountedVolumeURLs(includingResourceValuesForKeys: Array(keys),
                                                                options: [.skipHiddenVolumes]) else {
@@ -214,6 +217,7 @@ final class DiskSampler {
             let identity = statfsIdentity(for: url.path)
             return MountedVolume(name: name.isEmpty ? url.path : name,
                                  mountPath: url.path,
+                                 volumeUUID: values.volumeUUIDString,
                                  totalBytes: total,
                                  freeBytes: min(free, total),
                                  rawFreeBytes: rawFree,
--- a/Sources/Vorssaint/Services/Metrics/DiskSupport.swift
+++ b/Sources/Vorssaint/Services/Metrics/DiskSupport.swift
@@ -26,6 +26,9 @@ struct DiskDeviceReading: Identifiable, Equatable {
     var id: String
     var name: String
     var mountPath: String
+    /// Stable across renames and remounts, so the eject exclusion list matches
+    /// on it as well as on the name and the mount path.
+    var volumeUUID: String?
     var bsdName: String?
     var wholeDisk: String?
     var ioCounterID: String?
--- a/Sources/Vorssaint/Services/QuickTools/QuickTogglesSupport.swift
+++ b/Sources/Vorssaint/Services/QuickTools/QuickTogglesSupport.swift
@@ -71,9 +71,12 @@ enum QuickTogglesSupport {
 
     /// Whether a volume matches any entry in the user's exclusion list by
     /// name (case-insensitive), volume UUID, full mount path or mount directory name.
+    /// None of the four forms is optional to pass: an entry the user typed is
+    /// honoured or ignored depending on which identifiers the caller happened
+    /// to hand over, so a caller that has none says so with an explicit nil.
     static func isExcluded(volumeName: String?,
-                           volumeUUID: String? = nil,
-                           mountPath: String? = nil,
+                           volumeUUID: String?,
+                           mountPath: String?,
                            excludedVolumes: Set<String>) -> Bool {
         guard !excludedVolumes.isEmpty else { return false }
         if let name = volumeName?.trimmingCharacters(in: .whitespacesAndNewlines), !name.isEmpty {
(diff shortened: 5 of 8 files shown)


PathGao:
Removing the two defaults is the right shape: the omission now fails to compile instead of waiting for a reviewer, and that is what surfaced `DiskProtectionService`. The wake path matches the file's own `Task { @MainActor }` idiom, and `refresh()` really does recompute off `Date()`, so the re-arm holds.

One gap on "everywhere". `DiskExclusionsList.mountedCandidateDrives` runs its own exclusion test rather than calling `isExcluded`:

```swift
if !trimmed.isEmpty, !currentExcluded.contains(trimmed.lowercased()), …
```

Name only, and its `keys` set has no `.volumeUUIDStringKey`. So a volume excluded by UUID keeps showing up in the Settings exclusions picker as if it were not excluded — the same identifier-dependent answer you removed from the eject path. The body dismisses this file, but that argument covers its `shouldOfferEject` call; this line is a second, separate exclusion test, and the compiler cannot reach it. Same fix: add the key, call `isExcluded(volumeName:volumeUUID:mountPath:excludedVolumes:)`.

Unless the picker is meant to dedupe by name only — then the title should not say everywhere.

**Impact:** cosmetic. A UUID-excluded drive still appears in the exclusions picker; the eject paths are correct after this PR.
**Merge:** needs `mountedCandidateDrives` matching on the UUID too, or the title narrowed.

PathGao:
@PathGao You are right, and the argument in the body only covered the
`shouldOfferEject` call. `mountedCandidateDrives` ran a second, separate
exclusion test one line below it — `!currentExcluded.contains(trimmed.lowercased())`
— which is name-only, so a drive excluded by UUID or by mount path kept being
offered in the picker. The picker is not meant to dedupe by name; it is meant
to hide drives that are already excluded, and it was simply asking a weaker
question than the two eject paths. So I made it match rather than narrow the
title.

`.volumeUUIDStringKey` is now in the picker's key set, matching
`QuickTogglesService.ejectableVolumeURLs`, and the ad-hoc test is replaced by
`QuickTogglesSupport.isExcluded(volumeName:volumeUUID:mountPath:excludedVolumes:)`.
That picks up the mount path and mount directory name forms too, which the
name-only test also missed. `shouldOfferEject` is still called with none of the
exclusion arguments, deliberately, since that call only wants the "is this
drive ejectable at all" half.

I grepped for every place that decides whether a drive is excluded before
choosing. Three test a drive: `QuickTogglesService` (all four forms, already
correct), `DiskProtectionService.isExcluded(disk:)` (correct after the previous
commit) and this picker. All three now route through the same function.
`DiskProtectionService.addExcludedVolume` and `removeExcludedVolume` also
compare strings case-insensitively, but they deduplicate entries in the user's
own list rather than test a drive, so I left them alone.

`./build.sh --test` gives `TESTS OK (9490 checks)` with no failures, one more
check than the previous push. `DiskExclusionsList.swift` is outside the
`--test` whitelist, so the new assertion is a source-shape check like the wake
one. I did not spend a second build to see it red; instead I evaluated its
predicate directly against the pre-fix content of the file, stripping comment
lines the same way the assertion does — zero matches before, both matches
after, so it is red on the old file and green on the new one.

Not verified: I still have no second volume with a UUID mounted here, so the
picker hiding a UUID-excluded drive is confirmed only as far as the UUID now
reaching `isExcluded`. I did not run the app, and `./build.sh` was not run —
the locked signing keychain makes codesign block on an unlock dialog, so the
release build and the zero-warning check are CI's.

PathGao:
Fix is right, and the stronger of the two options — the picker was asking a weaker question than the eject paths, so making it match is what the title's "everywhere" actually claims.

I re-grepped `main` rather than take the coverage claim: `isExcluded` has three callers (`QuickTogglesService` via `shouldOfferEject`, `DiskProtectionService.isExcluded(disk:)`, this picker), `shouldOfferEject` has exactly two call sites, `DiskDeviceReading(` has two construction sites. Holds.

Two non-blocking notes.

`shouldOfferEject` keeps all four defaults, so the hazard you removed from `isExcluded` still lives one layer up: a caller can pass `volumeName:` and `excludedVolumes:` and silently skip the UUID. Nothing is wrong today — `QuickTogglesService` is its only caller that passes exclusions, and it passes all four.

The new wake observer discards its token; the other wake observers store theirs (`wakeObserver`, `observers`) and remove them. `ScreenshotShareService` is a `static let shared` so it never deinits — no leak, just not the same shape.

**Impact:** none beyond the fix. Picker, panel and Eject all disks now give the same answer for the same drive.
**Merge:** ship it.

PathGao:
@PathGao Split in `2926727a`. The wake half is off this branch and is now #1266.

**Removed here:** the `NSWorkspace.didWakeNotification` observer in
`ScreenshotShareService.init()` (9 lines) and the source-shape assertion that
pinned it in `MetricsTests`. The shared `// MARK:` line above the remaining
assertions was narrowed to match.

**Why:** the summary on this branch opened by calling these two unrelated
fixes, and they share no code. What stays alters a signature across five files
— `isExcluded` losing its defaults, a UUID carried through `DiskDeviceReading`
and `DiskSampler` into `DiskProtectionService`, and the picker rewritten to ask
the shared test. The wake half is nine lines in a file none of that touches,
and by its own body its visible effect is small: `ScreenshotSettings` already
refreshes in `onAppear` at both places it lists links, so a stale list is only
reachable when the panel was left open across a sleep. Bundled, that small fix
waits on review of the signature change; apart, neither blocks the other.

**What stays:** `QuickTogglesSupport.isExcluded` without its `volumeUUID` and
`mountPath` defaults, `DiskSupport`/`DiskSampler` carrying
`.volumeUUIDStringKey` through to `DiskProtectionService`, and
`DiskExclusionsList.mountedCandidateDrives` calling `isExcluded` instead of
comparing names. `shouldOfferEject` keeps its four defaults for the reason
already in the body. Title and summary updated; the wake paragraphs and the
wake verification bullet are gone rather than left describing code that is not
here.

`./build.sh --test` on `2926727a`: `TESTS OK (9489 checks)`, no failures —
one check fewer than the previous push, which is the wake assertion that left
with its fix. Neither of the two input-source flakes appeared in this run.
`./build.sh` was not run: the signing keychain is locked here and codesign
blocks on its unlock dialog.

Worth rechecking after this: that the assertion count on this branch is short
by exactly the one check, and that #1266 carries the observer verbatim — the
observer and its assertion moved unchanged, so nothing should differ between
what left here and what arrived there.

PathGao:
Split is right — nothing that stays touches `ScreenshotShareService`, so the two never needed each other.

The count holds. This diff adds exactly two `expect` calls to `MetricsTests`; the previous push had three. 9490 → 9489 is the wake assertion and nothing else.

#1266 I can't see, so "verbatim" stays unverified. The near half I can confirm: `ScreenshotShareService.swift` has zero hunks here, and `didWakeNotification` does not appear in that file on `main` — this branch has no orphaned half waiting on the observer.

Two things I checked this round that I hadn't before. The picker passes `currentExcluded`, already `Set(protection.excludedVolumes.map { $0.lowercased() })` — same shape `DiskProtectionService` builds, so `isExcluded`'s lowercased comparisons line up. And `DiskDeviceReading(` has three construction sites, not the two I said last time: `MetricsTests` builds one too, and it compiles unedited for the same memberwise-default reason as `MenuBarRenderer`.

**Impact:** none. Same fix as the last push, minus nine unrelated lines.
**Merge:** ship it.

PR #1241, opened 2026-09-01
fix(update): bound the What's New showcase video download

PathGao:
## Summary

The What's New showcase video was downloaded with a bare
`URLSession.shared.downloadTask`, with nothing bounding how much it could
write.

`timeoutInterval: 12` on the request is the idle timeout — the gap allowed
between two arrivals of data — so a response that trickles a byte at a time
never trips it. `downloadTask` writes the body straight to disk, whatever its
size. The SHA-256 check happens *after* the download completes, so it decides
whether the file is used, not how much of it lands.

The fix is the answer this directory already has.
`UpdateService` solved the identical problem with
`BoundedUpdateDownloadDelegate`, which writes to a scratch file and abandons
the task once it passes a limit, and with `downloadCeilingBytes`, whose
comment says it exists "only to stop a response that never ends". Rather than
write a second delegate for the same job, that one is now shared (`private`
dropped, doc comment added) and the loader runs its download through it on a
session that sets `timeoutIntervalForResource`, which does bound the whole
transfer.

Teardown then has to cover both ways the loader can go away. `cancel()`
invalidates the session rather than just cancelling the task, but its only
caller is `.onDisappear` in `UpdateShowcaseIntroView`, and the loader is a
`@StateObject` that SwiftUI can drop without `.onDisappear` ever running. On
that path the delegate completion returns at `guard let self else { return }`
before `finishTasksAndInvalidate()`, so the session is never invalidated — and
a live session holds its delegate for the life of the process, while it is the
delegate's own `deinit` that closes and deletes the scratch file. A `deinit` on
the loader invalidates the session on the release path too. It cancels rather
than finishes, because `finishTasksAndInvalidate()` would let an abandoned
download run to completion first; the completion closure holds `self` weakly
and that reference is already nil by then, so the teardown cannot resurrect the
loader.

That `deinit` is also the last exit path that could strand a scratch file.
Every other path either moves the file to the cache or ends in the delegate's
`complete()`, which deletes it — including the untrusted-checksum and
failed-move paths, where the file survives `complete()` and is deleted by the
delegate's `deinit` once the session releases it.

Severity is low: the URL is a hardcoded HTTPS
GitHub release address, so reaching this needs the TLS channel or the CDN
behind it to be under someone else's control. This is a missing bound on an
untrusted-length response, not a vulnerability. The leak it now closes is not
user-visible either — one session, one delegate and one temporary file per
abandoned download, until the app quits.

Trade-offs:

- **The 200 MB ceiling is reused rather than given a tighter, video-specific
  value.** A demo mp4 is far smaller and a dedicated constant could say so,
  but that would be a second answer to the question `downloadCeilingBytes`
  already answers, and the next reader would have to work out why the two
  differ. The bound exists to end an endless response, and any finite number
  does that.
- **`BoundedUpdateDownloadDelegate` was shared in place, not moved to its own
  file.** Moving it would be a larger diff in `UpdateService.swift` for no
  behavioural difference; it stays beside its first caller in the same
  directory.
- **No Content-Length pre-check was added.** The delegate stores
  `expectedContentLength` but does not refuse on it, and adding that would
  change `UpdateService`'s behaviour too. The byte counter already stops the
  write, which is the bound that matters.
- 120 s resource timeout: generous for a file of this size, chosen above the
  neighbouring services' 20 s because those fetch JSON and this fetches video.

## Verification

```sh
./build.sh --test     # 9489 checks, 0 failures
```

Two assertions were added beside the existing update-download checks. They
read `UpdateShowcaseMedia.swift` and assert that the loader's `deinit` calls
`session?.invalidateAndCancel()` and does not call
`finishTasksAndInvalidate()`. That is a source-shape check, the idiom this
suite already uses for code it cannot link; a `deinit` on a `@StateObject` is
not reachable from a standalone test binary. Both were confirmed red first:
against the previous revision of the file the first fails, and against a
variant whose `deinit` finishes instead of cancelling, both fail.

`UpdateShowcaseMedia.swift` is not in the `--test` whitelist, so it is not
compiled here; CI covers it on both runners. The full `./build.sh` was not run
— with the signing keychain locked it blocks on an unlock dialog.

**Not tested:** no download was performed. The showcase asset is fetched only
when the running version equals `UpdateShowcaseInfo.releaseVersion` (3.1.4)
and What's New opens, which this build does not satisfy. The bounded path, the
cancel path, the new release path and the resource timeout were read, not
observed; the delegate they share is exercised in production by
`UpdateService`'s own download.

## Anything else

The same shape was checked everywhere else it could occur. `UpdateService` and
`SpeedTest` own delegate-backed sessions but are process-lifetime singletons,
so neither can be released with a session running. The favicon downloader in
`RadialMenuSupport` is its own session's delegate, so the session keeps it
alive to the end of the transfer, and it holds no scratch file. Every other
`URLSession` in the app is delegate-less or `URLSession.shared`. This loader
was the only one that could be dropped mid-transfer.

`UpdateService.check()` pulling the releases JSON through `URLSession.shared`
with no resource timeout and no ceiling is a real observation, but it is a
different file and a different failure, and adding it here would widen a PR
that is already about the showcase download. It is left for a separate change.

### DIFF
--- a/Sources/Vorssaint/Services/Update/UpdateService.swift
+++ b/Sources/Vorssaint/Services/Update/UpdateService.swift
@@ -474,7 +474,10 @@ final class UpdateService: ObservableObject {
     }
 }
 
-private final class BoundedUpdateDownloadDelegate: NSObject, URLSessionDataDelegate {
+/// Writes a response to a scratch file and abandons it once it passes
+/// `byteLimit`, so a body that never ends cannot fill the disk. Shared by the
+/// app update download and the What's New showcase video.
+final class BoundedUpdateDownloadDelegate: NSObject, URLSessionDataDelegate {
     private let byteLimit: Int64
     private let progress: (Int64, Int64?) -> Void
     private let completion: (URL?, URLResponse?, Error?) -> Void
--- a/Sources/Vorssaint/Services/Update/UpdateShowcaseMedia.swift
+++ b/Sources/Vorssaint/Services/Update/UpdateShowcaseMedia.swift
@@ -69,7 +69,7 @@ final class UpdateShowcaseMediaLoader: ObservableObject {
     }
 
     @Published private(set) var state: State = .idle
-    private var task: URLSessionDownloadTask?
+    private var session: URLSession?
 
     func load() {
         if case .ready = state { return }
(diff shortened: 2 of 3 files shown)


PathGao:
Reusing `BoundedUpdateDownloadDelegate` rather than writing a second one is the right call, and the idle-vs-resource timeout diagnosis is correct.

One gap. The body says `cancel()` "invalidates the session ... so the session cannot outlive the loader holding its delegate." That holds only when `cancel()` runs, and its one caller is `.onDisappear` in `UpdateShowcaseIntroView`. If the loader is released without it, the completion hits `guard let self else { return }` and returns before `finishTasksAndInvalidate()`: the session is never invalidated, keeps its delegate for the life of the process, and the delegate's `deinit` — the thing that deletes the scratch file — never runs. `URLSession.shared` couldn't leak that way, so it's new.

```swift
deinit { session?.invalidateAndCancel() }
```

Separately, `UpdateService.check()` pulls the releases JSON through `URLSession.shared` with no resource timeout and no ceiling — same shape, smaller stakes.

I ran no download either; the bounded, cancel and invalidate paths are read, not observed.

**Impact:** a loader deallocated mid-download without `.onDisappear` leaks a session, its delegate and one scratch file until the app quits. Nothing user-visible.
**Merge:** needs the `deinit` first.

PathGao:
@PathGao Right on both counts — thanks. `cancel()` was the only teardown and
`.onDisappear` was its only caller, and the loader is a `@StateObject`, which
SwiftUI can drop without that ever running. On that path the completion returns
at `guard let self else { return }` before `finishTasksAndInvalidate()`, the
session stays live holding the delegate, and the delegate's `deinit` — the
thing that deletes the scratch file — never runs.

Added the `deinit` you wrote. Two details I made sure of. It calls
`invalidateAndCancel()`, not `finishTasksAndInvalidate()`, so an abandoned
download stops instead of running to completion first; and it cannot resurrect
`self`, because the completion holds `self` weakly and that reference is
already nil by the time `deinit` runs — the cancellation delivers
`didCompleteWithError`, `complete(location: nil, …)` deletes the file, and the
completion closure then returns at its own `guard`. Once the session is
invalidated it releases the delegate, so the delegate's `deinit` runs on the
paths where the file is still on disk: the untrusted-checksum guard and the
failed `moveItem`, both of which leave a non-nil `tempURL` that `complete()`
does not delete. That makes the loader's `deinit` the last exit path that could
strand a scratch file.

Checked the same shape elsewhere. `UpdateService` and `SpeedTest` own
delegate-backed sessions, but both are `static let shared` singletons, so their
`deinit` is unreachable and the session cannot outlive them. `FaviconDownload`
in `RadialMenuSupport` passes `delegate: self`, so the session keeps it alive
to the end of the transfer, and it buffers in memory with no scratch file.
Every other `URLSession` in `Sources/` is delegate-less or `URLSession.shared`.
The showcase loader was the only one that could be released mid-transfer.

Verification: `./build.sh --test`, 9489 checks, 0 failures. Two assertions
added next to the existing update-download checks, reading
`UpdateShowcaseMedia.swift` and asserting the `deinit` calls
`session?.invalidateAndCancel()` and not `finishTasksAndInvalidate()` — the
source-shape idiom the suite already uses, since a `deinit` on a `@StateObject`
is not reachable from the standalone test binary. Confirmed red first: against
the previous revision of the file the first assertion fails, and against a
variant whose `deinit` finishes rather than cancels, both fail.

Not verified: I ran no download either, and `UpdateShowcaseMedia.swift` is
outside the `--test` whitelist so it is not compiled locally; CI covers it. The
release path itself is read, not observed.

Your note on `UpdateService.check()` fetching the releases JSON through
`URLSession.shared` with no resource timeout and no ceiling is fair, and I left
it alone — different file, different failure, and folding it in would widen a
PR that is about the showcase download. Say the word if you would rather it
came along here than as its own change.

PathGao:
Keep `UpdateService.check()` out of here. Different failure — a JSON body buffered in memory, not an unbounded write to disk — and `check()` runs on every launch, so reworking its session touches a path this PR otherwise doesn't go near. Own change, `priority: low`.

Verified the two numbers the body asserts. `ReleaseAssets/vorssaint-3.1.4-showcase-1.mp4` is 2,656,929 bytes, so 120 s is a floor of ~22 KB/s — "generous" is right. And I redid the `URLSession` sweep on `main`: 14 sites, the delegate-backed ones are `UpdateService`, `SpeedTest` (`private override init`, so no second instance can exist) and `FaviconDownload`. Your conclusion holds, and the `deinit` reads correctly — session → delegate → weak self, no cycle, so it is reachable.

One correction to the reasoning, not the code. `FaviconDownload` isn't safe *because* the session retains it through the transfer — that alone is the leak you just fixed. It's safe because `finish()` calls `finishTasksAndInvalidate()` on every exit path, including the `completionHandler(.cancel)` branch in `didReceive response:`. Say it that way, or someone later reads "the session keeps it alive" as the safety argument and drops the invalidation.

**Impact:** none beyond the diff.
**Merge:** ship it.

PR #1242, opened 2026-09-01
fix(recorder): lock the typing samples and run the sampler monitors on the main thread

PathGao:
## Summary

`RecorderTypingSampler` appends to a plain `[Double]` from its key monitor while `stop()` reads
that same array to build the track, with nothing between them. The stop order is
`engine.stop()` → `pointer.stop()` → `typing.stop()`, and the monitor is only removed inside
that last call, so every keystroke in that window is a concurrent append into a copy-on-write
buffer being copied — a crash or a corrupt track, not a stale value. Its twin
`RecorderPointerSampler` already guards the same shape with an `NSLock`; this gives the typing
sampler the same lock.

The lock covers `startedAt` as well as the buffer. `start()` was writing it and the monitor
callback was reading it with nothing between them, so only the buffer append had actually moved
inside the lock — half of what the pointer sampler does, which writes `startedAt` in `start()`
and reads it in `record()` both under its lock. `startedAt` is the origin every recorded time is
measured from, so a torn or stale read of it shifts the whole typing track, not one sample.
`record()` now takes the lock once and does the `pauseClock.eventTime` conversion and the append
inside it, which is the pointer sampler's shape line for line.

`RecorderSession.start`/`stop` are `nonisolated async`, so awaiting them from
`Task { @MainActor }` runs their bodies on the cooperative pool (SE-0338), which put
`NSEvent.addGlobalMonitorForEvents` and `removeMonitor` on a background thread. Verified with a
standalone reproduction on this toolchain: the caller is on the main thread, the awaited body
is not. The two sampler calls now hop back with `MainActor.run`.

Trade-offs. The lock is `NSLock` rather than a serial queue or an actor because that is what
the pointer sampler next to it already uses for the identical shape, and the append is a few
nanoseconds. Calling `pauseClock.eventTime` while the sampler's lock is held nests two locks,
which the pointer sampler already does; `RecorderPauseClock` never calls back into a sampler, so
there is no cycle. For the threading, marking `RecorderSession.start`/`stop` `@MainActor` was the
smaller edit but would have moved every `await` in them — and the `writerQueue.sync {}` in
`stop()` — onto the main thread, so the hop is scoped to the two calls that need AppKit.
Clearing the typing buffer in `stop()` mirrors what the pointer sampler does and stops a second
`start()` from inheriting the first recording's keystrokes.

The only other `startedAt` on this shape is `ScreenRecorderService`'s, and it is not the same
one: it is written in `recordingDidStart()` and read from `tickElapsed()` and `togglePause()`,
all on the main thread, with no monitor callback touching it. It is left alone.

Dropped from the branch: stamping an event by `NSEvent.timestamp` instead of
`CACurrentMediaTime()`. Earlier commits here changed only the event side, while `startedAt` and
the `now` that `togglePause()` hands to `RecorderPauseClock.pause(at:)`/`resume(at:)` both stay
on `CACurrentMediaTime()`, which leaves three subtractions across two clocks. That
`NSEvent.timestamp` shares a time base with `CACurrentMediaTime()` has no Apple documentation
behind it and was never checked against a real recording, and `RecorderSupport.elapsed(since:at:)`
opens with `guard time > origin else { return 0 }`, so a mismatched base is clamped to zero
silently. Separately, the synthetic `.leftMouseDown` that `DockClickService.handleDragged` posts
carries no timestamp of its own, and what a zero-timestamp event should do here was never
settled. Earning it back means moving `startedAt` and the pause clock to the same source,
deciding the zero-timestamp case, and measuring the delivery delay on one real recording; with
none of the three done the whole half comes out. `RecorderPointerSampler.swift` is unchanged by
this branch as a result, and the source-shape check that walked both samplers now walks only the
typing track: it was added here for a change that no longer exists on that file. The pointer
sampler's own lock discipline is therefore pinned by nothing, which is left to whoever changes
that file next.

That narrowed check is a brace counter, and it counts every `{` in the file, a comment's or a
string literal's included. One unbalanced push shifts the stack, so a `withLock` entry outlives
its own block and the next append is read as a locked one: the check goes on passing while the
lock it watches is gone. It desyncs green, which is the one direction a source-shape check must
not fail in, so the walk now also asserts the file parses to balanced braces.

## Verification

`./build.sh --test` on this branch: `TESTS OK (9491 checks)`, no failures. Both files this
change touches — `RecorderTypingTrack.swift` and `Tests/MetricsTests.swift` — are in the
`--test` whitelist, so both were compiled here.

`ScreenRecorderService.swift` is outside that whitelist and was not compiled here; CI covers it
on both runners. The full `./build.sh` was not run: with the signing keychain locked, `codesign`
blocks on an unlock dialog that never returns.

The assertions were confirmed red before they were trusted. They are source-shape checks over
`RecorderTypingTrack.swift` alone: reverting the typing sampler to an unlocked append turns the
buffer check red, and reverting the `startedAt` write out of the lock turns the start-time check
red, both naming the file. Re-adding an assertion on `event.timestamp` after the drop above turns
red as well, which is why it is gone rather than kept.

The brace-balance assertion was confirmed the same way, and it was written because the failure it
covers was reproduced rather than assumed. A `{` inside a comment in the locked block, with the
only `times.append(` moved outside the lock, left both lock assertions green - a real unguarded
append, reported as locked. With the assertion in place that same file is
`TESTS FAILED (1 of 9491)` naming the parse, and the untouched file is `TESTS OK (9491 checks)`.

Not tested: no actual recording was made. The data race needs typing during the stop window to
reproduce, which is timing-dependent by nature. Single display, no second login session, no
external mouse.

## Anything else

No open or closed issue reports this. The recorder issues on the tracker are about window
capture, editor handles, shortcuts and webcam overlay, and none of them describes a corrupt or
shifted typing track, and the three long-lived index issues do not carry it either. Nothing is
referenced rather than referencing something adjacent, since the fix-status bot would then track
the wrong report.

One review note is outside this change and stays open on purpose. `RecorderTypingSampler`
`deinit` calling `stop()` runs `NSEvent.removeMonitor` on whatever thread releases the sampler,
which is the same off-main AppKit call the `MainActor.run` hop fixed on the normal path; the
pointer sampler has no `deinit` to compare against, so making the two agree is its own change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project are documented here. The format follows
 ## [Unreleased]
 
 ### Summary
-Vorssaint adds quit and close protections, expands screen text recognition, capture magnifier, clipboard, Super key, Dock Preview, App Switcher, Window Layout, mouse controls and snippet controls, broadens app update and safe cleanup discovery, and hardens shortcut capture and key caps, permission guide recovery, input session switching, Accessibility timeouts, Volume Mixer audio rendering, helper process attribution and teardown, Super key shutdown, process termination, update and uninstallation teardown, window handling and shared on-screen parsing, pasteboard restoration, sensor selection, Cleaning Mode unlock, favicon downloads and the What's New showcase video download. It also improves mouse scrolling feel, disk space and system monitor metrics, Command Bar responsiveness, Settings, Scratchpad, Screenshot Editor, floating panels and several menu bar behaviors.
+Vorssaint adds quit and close protections, expands screen text recognition, capture magnifier, clipboard, Super key, Dock Preview, App Switcher, Window Layout, mouse controls and snippet controls, broadens app update and safe cleanup discovery, and hardens shortcut capture and key caps, permission guide recovery, input session switching, Accessibility timeouts, Volume Mixer audio rendering, helper process attribution and teardown, Super key shutdown, process termination, update and uninstallation teardown, window handling and shared on-screen parsing, pasteboard restoration, sensor selection, Cleaning Mode unlock, favicon downloads, the What's New showcase video download and stopping a recording. It also improves mouse scrolling feel, disk space and system monitor metrics, Command Bar responsiveness, Settings, Scratchpad, Screenshot Editor, floating panels and several menu bar behaviors.
 
 ### Added
 - Settings now includes optional protections for Command-Q and Command-W with customizable hold duration, double press, extra modifier requirements and per-app scopes. Thanks to @RuanMD.
@@ -105,6 +105,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 - The battery icon in the menu bar now preserves its rectangular aspect ratio when split into its own item instead of rendering as a square. Thanks to @Yahddyyp.
 - Cleaning Mode now uses a forgiving 6-second press window for the Escape unlock gesture and resets the count when modifier keys are pressed while wiping. Thanks to @iltonandrew.
 - Dock previews now move to the vacated screen edge when an auto-hiding Dock slides away instead of floating detached. Thanks to @iltonandrew.
+- Stopping a recording while you are still typing no longer risks a crash or a recording whose typing moments land in the wrong place. Thanks to @PathGao.
 
 ## [3.3.3-beta.3] - 2026-08-26
 
--- a/Sources/Vorssaint/Services/Recorder/RecorderPointerSampler.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderPointerSampler.swift
@@ -10,8 +10,7 @@ import QuartzCore
 /// three reasons: it needs no permission at all, it costs about a hundred
 /// nanoseconds per sample, and a filter that runs offline wants its input on
 /// an even grid anyway. The presses come from a mouse-only global monitor,
-/// which is ungated unlike key events, and which stamps every press with the
-/// moment it happened rather than the moment it was handed over.
+/// which is exact and, unlike key events, ungated.
 ///
 /// Nothing here exists between recordings: the thread, the monitor and the
 /// buffer are all created in `start()` and gone after `stop()`.
@@ -68,10 +67,8 @@ final class RecorderPointerSampler {
             return self.generation
         }
 
-        // Mouse-only monitors need no Accessibility grant. The press carries
-        // its own moment, which is what gets recorded: the callback can be
-        // handed over late behind the main thread's own work, and a click
-        // ring that lands after the click is visible.
+        // Mouse-only monitors need no Accessibility grant, and a press is
+        // never coalesced the way a move is, so these timestamps are exact.
         clickMonitor = NSEvent.addGlobalMonitorForEvents(
             matching: [.leftMouseDown, .leftMouseUp, .rightMouseDown, .rightMouseUp]
         ) { [weak self] event in
@@ -165,7 +162,7 @@ final class RecorderPointerSampler {
     private func record(_ event: NSEvent, generation: Int) {
         lock.withLock {
             guard running, self.generation == generation else { return }
-            guard let time = pauseClock.eventTime(event.timestamp, since: startedAt)
+            guard let time = pauseClock.eventTime(CACurrentMediaTime(), since: startedAt)
             else { return }
             let isDown = event.type == .leftMouseDown || event.type == .rightMouseDown
             let click = RecorderMotion.Click(time: time, isDown: isDown)
--- a/Sources/Vorssaint/Services/Recorder/RecorderPointerSampler.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderPointerSampler.swift
@@ -10,7 +10,8 @@ import QuartzCore
 /// three reasons: it needs no permission at all, it costs about a hundred
 /// nanoseconds per sample, and a filter that runs offline wants its input on
 /// an even grid anyway. The presses come from a mouse-only global monitor,
-/// which is exact and, unlike key events, ungated.
+/// which is ungated unlike key events, and which stamps every press with the
+/// moment it happened rather than the moment it was handed over.
 ///
 /// Nothing here exists between recordings: the thread, the monitor and the
 /// buffer are all created in `start()` and gone after `stop()`.
@@ -67,8 +68,10 @@ final class RecorderPointerSampler {
             return self.generation
         }
 
-        // Mouse-only monitors need no Accessibility grant, and a press is
-        // never coalesced the way a move is, so these timestamps are exact.
+        // Mouse-only monitors need no Accessibility grant. The press carries
+        // its own moment, which is what gets recorded: the callback can be
+        // handed over late behind the main thread's own work, and a click
+        // ring that lands after the click is visible.
         clickMonitor = NSEvent.addGlobalMonitorForEvents(
             matching: [.leftMouseDown, .leftMouseUp, .rightMouseDown, .rightMouseUp]
         ) { [weak self] event in
@@ -162,7 +165,7 @@ final class RecorderPointerSampler {
     private func record(_ event: NSEvent, generation: Int) {
         lock.withLock {
             guard running, self.generation == generation else { return }
-            guard let time = pauseClock.eventTime(CACurrentMediaTime(), since: startedAt)
+            guard let time = pauseClock.eventTime(event.timestamp, since: startedAt)
             else { return }
             let isDown = event.type == .leftMouseDown || event.type == .rightMouseDown
             let click = RecorderMotion.Click(time: time, isDown: isDown)
--- a/Sources/Vorssaint/Services/Recorder/RecorderTypingTrack.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderTypingTrack.swift
@@ -22,6 +22,8 @@ struct RecorderTypingTrack: Codable, Equatable {
 /// The monitor exists only while recording and remembers timing, never keys.
 final class RecorderTypingSampler {
     private let pauseClock: RecorderPauseClock
+    /// Written by `start()` and read by the monitor callback, so it lives
+    /// under the same lock as the buffer it times events against.
     private var startedAt: CFTimeInterval = 0
     private var globalMonitor: Any?
     private var localMonitor: Any?
@@ -34,7 +36,7 @@ final class RecorderTypingSampler {
 
     func start() {
         guard globalMonitor == nil, localMonitor == nil else { return }
-        startedAt = CACurrentMediaTime()
+        lock.withLock { startedAt = CACurrentMediaTime() }
         globalMonitor = NSEvent.addGlobalMonitorForEvents(matching: .keyDown) { [weak self] event in
             self?.record(event)
         }
@@ -61,8 +63,10 @@ final class RecorderTypingSampler {
         // The key's own moment, not the moment the callback was scheduled:
         // the main thread is busy redrawing the indicator while recording,
         // and this timestamp starts an auto zoom in the finished video.
-        guard let time = pauseClock.eventTime(event.timestamp, since: startedAt) else { return }
-        lock.withLock { times.append(time) }
+        lock.withLock {
+            guard let time = pauseClock.eventTime(event.timestamp, since: startedAt) else { return }
+            times.append(time)
+        }
     }
 
     deinit {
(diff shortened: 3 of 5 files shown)


PathGao:
The SE-0338 diagnosis is right and rarely caught, and scoping the hop to the two calls rather than marking the method `@MainActor` is the right trade.

1. The lock is half the sibling's shape. `RecorderPointerSampler` writes `startedAt` under the lock in `start()` and reads it under the lock in `record()`. The typing sampler still does both unlocked — only the append moved. Benign today only because the hop puts writer and reader on the main thread, which means the hop is what makes it safe, and the test asserts the lock, not the hop.

2. `event.timestamp` is "when it happened" for hardware events. The app posts its own `.leftMouseDown`: `DockClickService.handleDragged` creates one without setting `timestamp`, and `replaySpacesPress` in `MouseButtonShortcutService` had to restamp a replayed press explicitly. Those now decide where the click ring lands. I have not verified what `CGEventPost` does with a zero timestamp — question, not a claim.

3. `RecorderTypingSampler.deinit` calls `stop()` → `NSEvent.removeMonitor` on whatever thread releases it. Same off-main AppKit call the hop just fixed; the pointer sampler has no deinit.

**Impact:** none from the diff as written; (2) is the one that could change recorded output.
**Merge:** needs the `startedAt` guard in the typing sampler first.

PathGao:
@PathGao You are right that the lock was half the sibling's shape. Only the append had moved; `start()` still wrote `startedAt` unlocked and the key monitor still read it unlocked, so the field the whole track is measured from was the one part left racing. `ca81382` writes it inside the lock in `start()` and folds the `pauseClock.eventTime` conversion and the append into one `lock.withLock` in `record()`, which is `RecorderPointerSampler.record` line for line. The hop is no longer what makes it safe.

Deliberately a lock and not a filter. A `guard time >= startedAt` would have closed the same sentence and been wrong: `NSEvent.timestamp` is the event's own moment, so it can fall before the origin without being stale — a synthesized press carrying no timestamp of its own (your point 2), or a local-monitor event already sitting in the queue when `start()` ran. Both are legitimate samples and `elapsed` already clamps them to 0. Staleness is decided by how long the monitor lives, not by comparing two clocks, so the guard I added protects the read of `startedAt` and drops nothing.

I checked the rest of the shape. `startedAt` on this pattern exists in three places: the pointer sampler, which was already correct and is untouched; the typing sampler, fixed here; and `ScreenRecorderService`, which is written in `recordingDidStart()` and read in `tickElapsed()` and `togglePause()`, all main thread, no monitor callback — not the same shape, left alone. The nesting of `pauseClock`'s lock inside the sampler's is new for the typing sampler but not new to the file pair: the pointer sampler already does it, and `RecorderPauseClock` never calls back into a sampler.

Verified with `./build.sh --test`: `TESTS OK (9495 checks)`, no failures. The new assertion extends the sampler shape check already in `MetricsTests` and counts, per file, `startedAt` writes and `event.timestamp` reads that sit outside a `withLock` block. Because it reads source text I ran it against `7926545f` as well as against the fix: on `7926545f` the typing sampler scores two unlocked sites and it fails while the pointer sampler passes, and with the fix both pass. Red before green, and red only where the defect was.

Not verified: no recording was made, so the visible half of this — where the click ring, the press punch and a typing-triggered zoom land — is still untested by me. The race itself needs typing inside the stop window and is timing-dependent. `./build.sh` in full was not run; the two files this round edits are both in the `--test` whitelist and were compiled, `RecorderPointerSampler.swift` and `ScreenRecorderService.swift` are not in it and are also not edited in this round.

Your points 2 and 3 are untouched and I did not fold them in. What `CGEventPost` writes into a zero `timestamp` decides where the ring lands for the app's own presses and I would rather not guess at it. And `deinit` calling `stop()` puts `removeMonitor` on whatever thread releases the sampler, which is the same off-main AppKit call the hop fixed on the normal path — the pointer sampler has no `deinit`, so making the two agree is a change of its own rather than a line in this one.

PathGao:
`ca81382` closes point 1 — `startedAt` under the lock on both sides, and red on `7926545f` makes the assertion worth something. The lock-not-filter argument holds: `elapsed` clamps `time <= origin` to 0, so nothing is dropped.

You don't need to know what `CGEventPost` does to close point 2. `DockClickService.handleDragged` builds its `.leftMouseDown` with `CGEvent(mouseEventSource:...)` and never sets `timestamp`. If that arrives as 0, `eventTime(0, since: startedAt)` fails the `time > origin` guard and returns 0 — the ring lands at the start of the recording, which is worse than the delay this PR fixes, by the PR's own argument. `replaySpacesPress` restamping by hand is evidence that posting does not restamp.

One line, correct either way, in both `record`s:

```swift
let stamp = event.timestamp > 0 ? event.timestamp : CACurrentMediaTime()
```

Not verified: that a self-posted press actually reaches the global monitor during a recording. I read the post path, not a recording.

**Impact:** a dock-icon drag while recording puts the click ring at t=0.
**Merge:** needs the zero-timestamp fallback in both samplers first.

PathGao:
@PathGao Dropped the event-timestamp half in `9a8deadb`. Not a force-push — the removal is its own commit.

**What went:** both samplers go back to `CACurrentMediaTime()`, and the two assertions that pinned `event.timestamp` go with them. `RecorderPointerSampler.swift` is now byte-identical to the merge base and has left the file list entirely. The lock half stays — `NSLock` over `startedAt` and the buffer, the guarded write and the guarded read-and-clear in `stop()` — as does `MainActor.run` in `ScreenRecorderService`.

**Why:** stamping only the event side left three cross-clock subtractions. `startedAt` is still `CACurrentMediaTime()` in both samplers, and `togglePause()` feeds `pauseClock.pause(at:)`/`resume(at:)` from `CACurrentMediaTime()` too, so `eventTime` compares one clock against the other. That `NSEvent.timestamp` and `CACurrentMediaTime()` share a base is not something Apple documents, and no real recording was ever made on this branch. The failure is silent either way: `RecorderSupport.elapsed(since:at:)` clamps with `guard origin.isFinite, time.isFinite, time > origin else { return 0 }`, so a mismatched time base reads as zero rather than as an error. And `DockClickService.handleDragged` synthesises its `.leftMouseDown` with only `.eventSourceUserData` set — what a zero timestamp does through `CGEventPost` was still an open question on this branch when I wrote it.

Three things would have to be true to bring it back: `startedAt` and the pause interval moved to the same clock, a defined answer for zero-timestamp events, and one real recording that measures how late the main thread actually stamps.

**Checks:** `TESTS OK (9493 checks)`; the input-source-dependent pair did not fire this run. Both edited files are in the `--test` whitelist. Falsified rather than assumed: putting the `event.timestamp` assertions back against the reverted source reddens both by name, `TESTS FAILED (2 of 9495)`.

**Worth re-checking:** the `MetricsTests` loop still walks `RecorderPointerSampler.swift` asserting its lock discipline, and that file is no longer in this branch's diff — a guard on code this PR does not touch. I am inclined to narrow it to the typing sampler; say if you would rather it stay wide.

PathGao:
@PathGao Narrowed the sampler source-shape check in `7de47698`. Again a plain commit, no force-push.

**What changed:** the loop in `MetricsTests.swift` walked both `RecorderTypingTrack.swift` and `RecorderPointerSampler.swift`. It now reads the typing track only, as straight-line code rather than a one-element loop.

**Why:** that loop is this branch's own — it arrived with the lock, and it covered two files because the branch changed two files. After `9a8deadb` the pointer sampler is byte-identical to the merge base and out of the file list, so the loop was asserting the lock discipline of code this branch does not touch. A guard belongs to the change that earns it.

**What stays:** both checks, on the typing track, where the `NSLock`, the locked append and the locked `startedAt` write all come from this branch. Verified both ways: unlock the append and `RecorderTypingTrack.swift appends to its sampler buffer only under the lock` fails; take `startedAt` back out of the lock and `RecorderTypingTrack.swift writes the recording's start time under the lock` fails; restore both and it is green.

**Known choice:** `RecorderPointerSampler.swift` now has no source-shape guard on its lock. It had none before this branch either. If that discipline is worth pinning it is a separate change with its own reason, not something this one should carry in passing.

`./build.sh --test`: `TESTS OK (9490 checks)` — down from 9493, which is the three expectations the dropped iteration was running. The two input-method-dependent checks (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints…`) were green on this machine's current layout.

Body updated: the check count, and the Verification paragraph that described the assertions as covering both sampler files.

Best place to re-check: whether dropping the pointer sampler's lock guard is acceptable to you, or whether you would rather it kept a guard of its own on a separate branch.

PathGao:
Narrowing was right, and no, don't open a branch for it.

I read what the loop was actually asserting on `RecorderPointerSampler.swift`: `let startedAt = CACurrentMediaTime()` is excluded by name, `self.startedAt = startedAt` sits inside `lock.withLock { () -> Int in`, and both `.append(` sites are under the lock. It was green before this branch and would stay green — dropping it costs no coverage, only upkeep on a file you don't touch.

Worth knowing about the check you kept: it desyncs green, not red. `blockIsLocked` pushes on every `{`, including one inside a comment or a string literal, and a single unbalanced push leaves the `withLock` entry on the stack for the rest of the file, so `contains(true)` then reports every later append as locked. One line of insurance after the loop:

```swift
expect(blockIsLocked.isEmpty, "\(samplerPath) parses to balanced braces")
```

Not verified: I read the pointer sampler on `main` by hand, I did not run the old two-file loop against it.

**Impact:** none.
**Merge:** ship it, assuming `Build & selftest` comes back green — it had not reported when I read this.

PathGao:
@PathGao Added the brace-balance guard in `aa11a78a`, and corrected the title. Plain commit, no force-push.

I reproduced the desync before trusting it, because "it can drift" and "it hides a real violation" are not the same claim. Inside the locked block in `record()` I put a comment containing a single `{`, and moved the only `times.append(` outside the lock. That is one genuinely unguarded append and nothing else. `./build.sh --test` came back `TESTS OK (9490 checks)` — both lock assertions green, because the spurious push made the withLock entry outlive its own block and the loose append was counted as a locked one. So it does not just drift, it passes over the exact failure it exists to catch.

With `expect(blockIsLocked.isEmpty, "\(samplerPath) parses to balanced braces")` after the loop, the same injected file reports `TESTS FAILED (1 of 9491)` naming the parse, and with the injection reverted the run is `TESTS OK (9491 checks)`. Nothing outside `Tests/MetricsTests.swift` changed in this commit.

The title was describing code that left the branch. `and stamp input with the event's own time` was the half `9a8deadb` removed; what is here now is the lock over the typing buffer and `startedAt`, plus the `MainActor.run` hop for the two sampler start/stop calls. Title is now `fix(recorder): lock the typing samples and run the sampler monitors on the main thread`. I reread the Summary for the same residue and found none — the timestamp half only appears in the paragraph that explains why it was dropped, which is still accurate. The body picks up the new assertion, its two-ended check and the corrected count.

Worth rechecking after this: the balance assertion is a guard on the walk, not on the sampler. It says the file parses; it still says nothing about `RecorderPointerSampler.swift`, whose lock discipline this branch stopped pinning when the check narrowed.

`./build.sh --test` on the tip: `TESTS OK (9491 checks)`, no failures — neither known input-method false red (`nil layout falls back to ANSI semicolon`, `App Switcher icon-row hints describe default app and window shortcuts`) appeared on this machine today. `ScreenRecorderService.swift` is still outside the `--test` whitelist and unchanged by this commit.

PathGao:
The two-ended reproduction is what makes the balance guard worth having. "it can drift" and "it passes over the exact failure it exists to catch" are different claims, and you separated them before trusting either.

I checked whether the guard covers the whole hole, and it does. A `}` pop can only remove a frame, never add a `true`, so a brace *deficit* can only turn a locked append into a loose one — false red, which announces itself. A surplus `{` is the only route to false green, and `blockIsLocked.isEmpty` is exactly that.

One piece of residue the narrowing left, in the code rather than the body: `!samplerLine.contains("let startedAt")` excludes a local that exists only in `RecorderPointerSampler.swift`. There is none in `RecorderTypingTrack.swift`, and the walk no longer reads the pointer sampler, so the clause guards a case the check can't see and its comment explains a file that left the branch. Fine to keep as insurance; the comment is the part that's now wrong.

Not verified: I read `main`'s copy of both samplers and grepped `Sources/` for `let startedAt`. I did not run the walk.

**Impact:** none.
**Merge:** ship it once `Build & selftest` and `Swift 6.0.3 compatibility` report — both were still blank when I read this.

PR #1243, opened 2026-09-01
perf(recorder): walk the click track once per plan instead of once per frame

PathGao:
## Summary

`RecorderComposer.makePlan` solves the whole pointer and zoom timeline before the first frame is
drawn, and four of its per-frame lookups walked their entire list every time:
`RecorderMotion.pressScale`, `ringProgress`, `focus` and `anchorWeight`. Ten minutes at 60 fps is
thirty six thousand frames, and a recording of real use carries a four figure click list, so the
export spent tens of millions of iterations — most of them on clicks that had faded out a minute
earlier or had not happened yet.

The frame loop asks in time order and both lists are in time order, so these are now cursored
walks: one pass across the whole loop instead of one pass per frame. The same file already
solved this exact problem this way in `resampled`, `resampledShapes` and `resampledVisibility`;
this brings the four stragglers over.

Measured on the extracted functions with `-O`, ten minutes of recording, 2914 clicks, 36000
frames: the press punch and the click ring together drop from 0.307 s to 0.001 s, with an
identical checksum over every frame.

This has to be a pure algorithmic swap — any difference in output is a bug, not a trade-off — so
each query keeps exactly one implementation. The old signatures still exist and now call the
cursored one with a fresh cursor. Four things needed care to stay bit-identical:

- Neither the press punch nor the click anchor may stop early while a button is held, because a
  press from long before this moment still weighs on it until its release arrives. They stop
  early only with nothing held, and their cursors step forward only past a release.
- Their trailing "still held at the end" check used the scan's own state, which the cursor no
  longer reproduces from an arbitrary start. It is now read off `clicks.last` directly, which is
  the same answer: every release clears that state, so only a final press can leave it set.
- `anchorWeight`'s cursor is carried by `anchored` over its own frame index rather than over the
  composer's `sourceTimes`, so its ascending order is true by construction. It is the only caller
  that carries one; every other caller, including every test, goes through the fresh-cursor
  signature.
- That fresh-cursor signature is equal to the scan it replaced **for a list in ascending time
  order, and only for one**. Every one of the four stops at the first entry too late to matter,
  so on a list that is not ascending they answer from the part before that break — and a fresh
  cursor is the case where the break lands earliest, not a protection against it. With clicks
  `[(10, down), (10.05, up), (0.5, down), (0.6, up)]` at `time = 0.55`, the old scan gives an
  anchor weight of 1 and a fresh cursor gives 0. The doc comments said the equivalence held
  unconditionally; they now carry the qualifier, and four assertions pin it.

  The real guarantee is that no producer emits a list out of time order: the sampler appends in
  event order, `RecorderPointerTrack.decode` round-trips the encoder's order, `focusClusters`
  emits ascending. That is why there is still no fallback branch, and it is a weaker statement
  than the one the comments used to make.

  Detecting a non-ascending *query* and resetting the cursor would not have made the old claim
  true: resetting gives the fresh path, which is exactly the path the counterexample breaks.
  Detecting a non-ascending *list* takes the whole-list scan the break exists to avoid, once per
  query. Neither buys anything a caller can reach today, so neither is here.

Trade-offs. A binary search per frame would have needed no ordering assumption about the frames
themselves, but it cannot start mid-list for the press punch or the anchor, whose answers depend
on the press/release pairing in the whole prefix. Sorting or defensively rescanning on unordered
input was rejected as well: unordered input would change the output. The ordering itself comes
from the encoder, not from there being a single producer — `RecorderPointerTrack.decode` builds
`clicks` too, and checks only `isFinite`, but it round-trips what the sampler appended in event
order.

## Verification

`./build.sh --test` on this branch: `TESTS OK (9498 checks)`, no failures.

The equivalence is asserted, not argued: the test keeps a copy of the scans that were replaced
and compares them frame by frame with `!=` on the raw `Double` — no tolerance — over 5972 frames
against 558 clicks built to include bursts, held buttons, presses that are never released and one
at the very end. Both the carried cursor and a fresh one are compared, for the press punch, the
click ring, the focus lookup and the click anchor. `anchored` itself is compared too: the blended
path it returns is checked point for point against the same blend driven by the full scan, at the
same `Double` step it uses internally.

That generated click list previously ended on a release, which left the trailing "ended mid-drag"
branch of all four queries unasserted. It now ends on a press that is never released.

Four more assertions pin the qualifier rather than the unqualified claim: each of the four
queries, on the counterexample list above, answers from before its break — anchor 0, punch scale
1, ring `nil`, focus the first cluster. The equivalence assertion above is worded "on a click
list in time order" so the test file no longer states the unqualified form either.

The new assertions were confirmed to fail on purpose, the same way the earlier ones were, on
extracted copies of the shipped functions driven by the same generated click list the test
builds. Advancing the anchor cursor past every click rather than only past faded-out releases
turns it red on 5264 of 5972 frames; stopping the scan early while a button is held, on 771;
dropping the `clicks.last` check, on 59. All three go green again when reverted. That last one is
red only because the list now ends on a held press — on the old list it was 0, which is what the
missing coverage looked like.

The four new qualifier assertions were confirmed red the same way: on extracted copies with the
early break removed — which is what widening the claim back to the unqualified form would take —
all four fail (anchor 1, punch 0.8, ring 0.125, focus the second cluster), and all four pass with
the break in place. The ascending equivalence was re-run exhaustively over every up/down pattern
up to six clicks, carried and fresh against the full scan: zero mismatches.

`RecorderMotion.swift` is in the `--test` whitelist and was compiled and run here.
`RecorderComposerPlan.swift` is not, and is untouched by this round; CI covers it on both runners.

Not tested: no video was exported. The claim being made is that the output is unchanged, so there
is nothing new to look at in a finished recording, and the timing figure above comes from the
extracted functions rather than from a real export. No profile of the anchor pass on a real
export was taken either — the argument for it is the same one the other three rest on, that the
work per frame is now bounded by the window rather than by the list.

## Anything else

No open issue reports a slow export, so there is nothing to reference. Searched the bug list and
the three summary issues for the symptom and for this PR number; the recorder issues that exist
(1147, 843) are about what is captured, not what the export costs.

### DIFF
--- a/Sources/Vorssaint/Services/Recorder/RecorderComposerPlan.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderComposerPlan.swift
@@ -122,6 +122,9 @@ extension RecorderComposer {
             var amountTargets = [Double](repeating: 1, count: frames)
             var travelX = [Double](repeating: 0.5, count: frames)
             var travelY = [Double](repeating: 0.5, count: frames)
+            // `sourceTimes` never goes backwards, so the lookups below walk
+            // their lists once across the whole loop instead of per frame.
+            var focusCursor = 0
             for index in 0..<frames {
                 let time = sourceTimes[index]
                 let state = RecorderTimeline.zoomState(at: time, segments: segments)
@@ -135,7 +138,9 @@ extension RecorderComposer {
                     travelY[index] = RecorderMotion.travelParameter(exactFocus: Double(aimed.y),
                                                                     zoom: state.amount)
                 } else {
-                    let focus = RecorderMotion.focus(at: time, clusters: clusters)
+                    let focus = RecorderMotion.focus(at: time,
+                                                     clusters: clusters,
+                                                     cursor: &focusCursor)
                     travelX[index] = RecorderMotion.travelParameter(focus: Double(focus.x))
                     travelY[index] = RecorderMotion.travelParameter(focus: Double(focus.y))
                 }
@@ -165,11 +170,16 @@ extension RecorderComposer {
         var pressScale = [Double](repeating: 1, count: frames)
         var ring = [Double](repeating: -1, count: frames)
         if showsPointer {
+            var clickCursor = RecorderMotion.ClickCursor()
             for index in 0..<frames {
                 let time = sourceTimes[index]
-                pressScale[index] = RecorderMotion.pressScale(at: time, clicks: track.clicks)
+                pressScale[index] = RecorderMotion.pressScale(at: time,
+                                                              clicks: track.clicks,
+                                                              cursor: &clickCursor)
                 if document.showsClickRing,
-                   let progress = RecorderMotion.ringProgress(at: time, clicks: track.clicks) {
+                   let progress = RecorderMotion.ringProgress(at: time,
+                                                              clicks: track.clicks,
+                                                              cursor: &clickCursor) {
                     ring[index] = progress
                 }
             }
--- a/Sources/Vorssaint/Services/Recorder/RecorderMotion.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderMotion.swift
@@ -207,12 +207,17 @@ enum RecorderMotion {
     }
 
     /// Where a caller asking in time order got to in the click list, on the
-    /// same terms as the press punch below: a fresh cursor gives exactly what
-    /// a full scan gives, and carrying one is only valid when the times asked
-    /// for never go backwards. It may only step past a RELEASE, because a
-    /// press still weighs on every moment until its own release arrives.
-    /// `anchored` is the only caller that carries one, over its own frame
-    /// index, so the ordering is true by construction rather than by promise.
+    /// same terms as the press punch below: for a click list in ascending time
+    /// order a fresh cursor gives exactly what a full scan gives, and carrying
+    /// one is only valid when the times asked for never go backwards. The
+    /// qualifier is the early break — out of order the scan stops short of
+    /// clicks that would have counted, and a fresh cursor is no protection
+    /// against that. The sampler appends in order and `decode` round-trips the
+    /// encoder's order, so nothing builds such a list today.
+    /// It may only step past a RELEASE, because a press still weighs on every
+    /// moment until its own release arrives. `anchored` is the only caller
+    /// that carries one, over its own frame index, so the order the moments
+    /// are asked in is true by construction rather than by promise.
     static func anchorWeight(at time: Double, clicks: [Click], cursor: inout Int) -> Double {
         var weight: Double = 0
         var pressedAt: Double?
@@ -467,8 +472,11 @@ enum RecorderMotion {
     }
 
     /// The same answer, for a caller that asks in time order. `cursor` starts
-    /// at zero and is carried across the calls; a fresh one gives exactly what
-    /// a full scan gives.
+    /// at zero and is carried across the calls; for clusters in ascending time
+    /// order a fresh one gives exactly what a full scan gives. Out of order it
+    /// stops at the first cluster past the moment, where the scan went on to
+    /// take the last one; `focusClusters` emits ascending, so nothing here
+    /// builds that list.
     static func focus(at time: Double, clusters: [FocusCluster], cursor: inout Int) -> CGPoint {
         guard let first = clusters.first else { return CGPoint(x: 0.5, y: 0.5) }
         var current = cursor == 0 ? first.center : clusters[cursor - 1].center
@@ -577,9 +585,14 @@ enum RecorderMotion {
     /// Where a caller asking in time order got to in the click list, so an
     /// export does not rescan every click for every frame: an hour of busy
     /// recording is thirty six thousand frames against a four figure click
-    /// list. A fresh cursor gives exactly what a full scan gives; carrying one
-    /// is only valid when the times asked for never go backwards, which the
-    /// composer's frame loop guarantees.
+    /// list. For a click list in ascending time order a fresh cursor gives
+    /// exactly what a full scan gives; carrying one is only valid when the
+    /// times asked for never go backwards, which the composer's frame loop
+    /// guarantees. Both halves want `clicks` ascending: these scans stop at
+    /// the first click too late to matter, so out of order they answer from
+    /// the part before the break, fresh cursor or not. The sampler appends in
+    /// order and `decode` round-trips the encoder's order, so nothing builds
+    /// such a list today.
     struct ClickCursor {
         fileprivate var punch = 0
         fileprivate var ring = 0
--- a/Sources/Vorssaint/Services/Recorder/RecorderMotion.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderMotion.swift
@@ -202,21 +202,49 @@ enum RecorderMotion {
     static let clickAnchorWindow: Double = 0.12
 
     static func anchorWeight(at time: Double, clicks: [Click]) -> Double {
+        var cursor = 0
+        return anchorWeight(at: time, clicks: clicks, cursor: &cursor)
+    }
+
+    /// Where a caller asking in time order got to in the click list, on the
+    /// same terms as the press punch below: a fresh cursor gives exactly what
+    /// a full scan gives, and carrying one is only valid when the times asked
+    /// for never go backwards. It may only step past a RELEASE, because a
+    /// press still weighs on every moment until its own release arrives.
+    /// `anchored` is the only caller that carries one, over its own frame
+    /// index, so the ordering is true by construction rather than by promise.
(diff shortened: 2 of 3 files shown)


PathGao:
Right shape for the equivalence argument: one implementation per query, old signature calling it with a fresh cursor, so "fresh cursor == full scan" is structural rather than hoped for. The punch's two hazards — can't stop early while a button is held, trailing check read off `clicks.last` — are both real and both handled. I checked the monotonicity premise as well: `keptRanges` builds its ranges ascending so `sourceTimes` never goes back, and `focusClusters` emits in order.

Four stragglers, not three. `RecorderMotion.anchorWeight` is the same full scan of `track.clicks`, called from `anchored` over `smoothed.indices` (strictly increasing times), and `makePlan` runs it at the top of the same `showsPointer` block. Same list, same samples × clicks magnitude, same cursor shape, untouched. The title says once per plan; that path is still once per sample.

Minor: "the only producer is the sampler" — `RecorderPointerTrack.decode` also builds `clicks`, checking only `isFinite`. It round-trips the sampler's order so nothing breaks today, but the ordering guarantee comes from the encoder, not from there being a single producer.

**Impact:** none on output. The export still pays the per-sample scan in `anchored`.
**Merge:** needs `anchorWeight` cursored the same way first.

PathGao:
@PathGao The fourth straggler is done. `RecorderMotion.anchorWeight` now takes the same cursor as
the punch, and `anchored` carries one across its own loop, so the export no longer walks the click
list once per output frame anywhere.

The rule had to be the punch's rule, because the shape is the punch's shape: the cursor steps
forward only past a release, since a press still weighs on every moment until its own release
arrives, and the scan stops early only with nothing held. The trailing "ended mid-drag" answer is
read off `clicks.last` for the same reason it is in the punch — the cursor cannot reproduce the
scan's own `pressedAt` from an arbitrary start, and only a final press can leave that state set.

On ordering: `anchored` is the only caller that carries a cursor, and it asks over `Double(index)
* step` on its own frame index, so the ascending order is a property of the loop rather than a
promise about the caller. That is a stronger footing than the punch and the ring have, which
depend on `sourceTimes` being ascending. Everything else that calls `anchorWeight` — the whole
test file — goes through the fresh-cursor signature, which is one implementation away from a full
scan, so an out-of-order query is answered correctly by construction rather than by a fallback.
That is why there is no fallback branch: there is no caller that can reach the carried form out
of order.

I checked the other two things the same shape touches. The minor point about `decode` is now in
the body, worded as the ordering coming from the encoder rather than from a single producer.
And the generated click list the cursor tests run against ended on a *release*, which meant the
trailing-press branch of all four queries was never asserted — the body claimed a press "at the
very end" and there wasn't one. It now ends on a press that is never released, so that branch is
covered for the punch and the anchor both.

Verification: `./build.sh --test` → `TESTS OK (9494 checks)`. The new assertions were confirmed
red on purpose the same way the earlier ones were, on extracted copies of the shipped functions
driven by the same generated list the test builds — advancing the anchor cursor past every click
rather than only past faded releases: red on 5264 of 5972 frames; stopping early while held: 771;
dropping the `clicks.last` check: 59. All green again on revert. That last figure is 0 on the old
list, which is exactly the coverage gap above. I also ran the equivalence exhaustively over every
up/down pattern up to six clicks: zero mismatches, carried and fresh.

Not verified: no export was run, so this rests on the frame-by-frame equivalence rather than on
looking at a finished video, and I have no profile of the anchor pass on a real export — the
argument for it is the same one the other three rest on. `RecorderComposerPlan.swift` is untouched
this round and is outside the `--test` whitelist; CI covers it on both runners.

PathGao:
Fourth straggler is closed the right way — `anchored` carrying its own cursor over `Double(index) * step` is a stronger ordering footing than the punch's. I grepped `main` for the rest: `anchorWeight`, `focus`, `pressScale` and `ringProgress` are reached only from `RecorderComposerPlan`, `anchored` and the tests, so that is all four, no sibling call site left behind.

One claim doesn't hold. "A fresh cursor gives exactly what a full scan gives" is in both doc comments and in the body, and it isn't true: the fresh path still early-breaks, and still reads the trailing press off `clicks.last` instead of the scan's own `pressedAt`. `[(10, down), (10.05, up), (0.5, down), (0.6, up)]` at `time = 0.55` — the old scan returns 1, a fresh cursor breaks on the first click and returns 0. `decode` checks only `isFinite`, so that list is representable.

It matters because that claim is the stated reason there is no fallback: "every other caller goes through the fresh-cursor signature, which is a full scan." It isn't one. The real guarantee is "no caller today asks out of order" — weaker, and nothing enforces it. One clause in the comment fixes it.

**Impact:** none today. A later out-of-order caller gets a wrong answer while the comment promises otherwise.
**Merge:** needs the fresh-cursor equivalence qualified to ascending input first.

PathGao:
@PathGao The counterexample is right and the claim is gone. `[(10, down), (10.05, up),
(0.5, down), (0.6, up)]` at `0.55` gives 1 from the old scan and 0 from a fresh cursor, and the
same list breaks the punch (0.8 vs 1) and the ring (0.125 vs nil); a non-ascending cluster list
breaks `focus` the same way. All three doc comments now say the equivalence holds for a list in
ascending time order, and name what the real guarantee is: no producer emits one out of order —
the sampler appends in event order, `decode` round-trips the encoder's, `focusClusters` emits
ascending. Weaker than what was written, which is the point.

I did not take the detect-and-reset route, because it does not make the claim true. Resetting the
cursor produces the fresh path, and the fresh path is exactly what the counterexample breaks — a
reset is a fallback to the thing being fallen back from. Detecting a non-ascending *list* is the
only thing that would catch it, and that needs a whole-list scan per query, which is the work the
break exists to avoid. So: qualifier in the comments, and assertions that pin the qualifier.

Four of them, one per query, on your list: anchor 0, punch scale 1, ring `nil`, focus the first
cluster. Red-first confirmed on extracted copies by removing the early break — which is what
widening the claim back to the unqualified form would take — all four fail (1, 0.8, 0.125, second
cluster) and all four pass with the break in place. The existing equivalence assertion was
reworded to "on a click list in time order" so nothing in the test file states the unqualified
form either. I also re-ran the ascending equivalence exhaustively over every up/down pattern up
to six clicks, carried and fresh: zero mismatches.

No behaviour changed this round — comments and assertions only. `./build.sh --test` →
`TESTS OK (9498 checks)`, up four from 9494. `RecorderComposerPlan.swift` is still untouched and
outside the `--test` whitelist; CI covers it on both runners. Still no export was run, so the
unchanged-output claim continues to rest on the frame-by-frame equivalence rather than on a
finished video.

PathGao:
The qualifier landed where it needed to and not further — three doc comments name ascending order as the precondition, and I re-derived all four counterexample assertions on `[(10,d),(10.05,u),(0.5,d),(0.6,u)]`: anchor 0, punch 1, ring nil, focus the first cluster. Declining detect-and-reset is right; a reset *is* the fresh path.

Three of the four got pinned, not four. `freshMatches` compares the fresh-cursor signature against the scan for the punch, the ring and the anchor. `focus` is only compared carried, so nothing in the test file asserts what its doc comment now promises — that a fresh cursor equals the scan on an ascending cluster list. One more clause in that loop's `if`.

Same loop: it runs against `cluster`, two entries at t=0 and t≈0.033, so 597 of its 600 frames sit past the last one. The other three get 558 clicks over 5972 frames.

Minor: the body's "every test goes through the fresh-cursor signature", and the `anchorWeight` doc comment's "`anchored` is the only caller that carries one" — the anchor equivalence test carries `anchorCursor` too. Ascending, so nothing breaks.

**Impact:** none. The cursored `focus` is correct; the claim about it is what isn't pinned.
**Merge:** needs the fresh-cursor comparison extended to `focus` first.

PathGao:
@PathGao Added the fresh-cursor comparison for focus in the same loop as its carried one (0abb963c), so all four cursor queries are now checked both ways. Body: the closing issue mentions lost their # (1147, 843) so they read as references, not fixes. TESTS OK (9498 checks).

PathGao:
All four queries are checked both ways now, and the focus loop is the right place for it: the carried path hits the `cursor - 1` initialiser at frame 1 and the fresh path is compared on every frame, so both states that can diverge are covered. `cluster` is two entries (t=0 and t=1/30), so 597 of the 600 frames sit past the last one — thin, but frames 0–2 are the ones that matter and they are in there.

Body reads as mentions rather than fixes, and the reference check is green.

Two sentences from last round survive and I am not holding on them: the body's "every other caller, including every test, goes through the fresh-cursor signature", and `anchorWeight`'s "`anchored` is the only caller that carries one". The anchor equivalence test carries `anchorCursor` too — ascending, so only the sentences are wrong.

Still no export run, so the unchanged-output claim rests on the frame-by-frame equivalence, same as every round.

**Impact:** none.
**Merge:** ship it.

PathGao:

PathGao:
Reopened as #1305.

PR #1244, opened 2026-09-01
perf(command-bar): stop refolding the query and rereading preferences on every keystroke

PathGao:
## Summary

Three things the command bar did on the main thread that it did not have to do.

**Every keystroke folded the query about a thousand times.** `searchRows`
builds one candidate per pooled row (900-1200 of them), and two of the things
it asks for each row started by folding what was typed:
`CommandBarQueryMemory.boost(query:)` and `CommandBarPreferences.aliasHit(_:query:)`.
Folding strips invisibles, folds diacritics, lowercases, splits and joins - four
heap allocations - so one keystroke paid for about a thousand foldings of the
same few letters, and the session memory returned zero for all of them until the
person had chosen something. Both now take an already-folded query, `searchRows`
folds once for the whole pass, and an empty memory answers without folding at
all. The comment three hundred lines above already said folding a thousand
titles per keystroke is the one thing that could make typing feel heavy; this is
the same cost, arriving through the query instead of the titles.

**Every keystroke re-read six preferences from disk.** The pins, the aliases,
the hidden keys, the usage counts and the switched-off sources are all computed
properties that hit `UserDefaults` and decode a string or a JSON blob on each
access, and `searchRows` runs off the field's `didSet`. The class already keeps
`pinCache` and `shortcutCache` for exactly this reason ("parsing the same JSON
ninety times for one frame") - `isPinned` used the cache while `searchRows` used
the raw property. The remaining four join the same cache, filled by
`reloadPreferenceCaches()` on every opening. The raw properties are renamed
`storedPins` / `storedAliases` / `storedHiddenKeys` and are now only read by the
four helpers that change one of those lists, where a read-modify-write has to
start from disk; `disabledSourcesRaw` is gone, since nothing outside the cache
wanted it.

**The battery and the memory pressure were read on the main thread.** Both sit
in `systemAnswerEntries`, which `rebuildCatalog()` calls synchronously one to
four times per opening, and both cross into the kernel (IOKit power sources, the
mach VM statistics). The three neighbouring facts in the same list - storage,
Wi-Fi, the Finder automation check - were already read in the background and
picked up on the next pass. These two now follow `refreshWiFiState`, sharing one
background hop.

**A background pass now stores its whole sample before it decides whether to
redraw.** `refreshSystemAnswers` compares the three memory fields the row shows
- `used`, `appUsed`, `total` - and returned before the assignment when they
matched, so `compressed`, `cached` and `swapUsed` would have kept whatever
reading they had when those three last moved. Nothing reads those three today,
which makes it a trap rather than a bug, and the fix is to assign first and
guard after. The same shape was already next door on `main`:
`refreshStorageAnswer` guards on `free` alone and stores a tuple that also
carries `total`, which the storage row does show, in its caption and in the
percentage. Both are reordered, since fixing only the one this branch
introduced would leave the shape standing one function above it.
`refreshWiFiState` and `refreshAutomationStatus` compare the whole value they
store, so they are already correct and are untouched.

Trade-offs:

- The four new caches take the lifetime the pins already had: read when the bar
  opens, trusted until it opens again. Nothing in Settings announces a write to
  them, and nothing needs to. `show()` calls `reloadPreferenceCaches()`, so
  every edit is picked up the next time the bar opens, and the bar is not on
  screen while those edits are made - a mouse-down anywhere outside the panel
  hides it, including one in another window of this app, so the click that
  flips a switch or presses a remove button has already hidden the bar. This is
  the lifetime the compact-mode toggle in the same view already relies on, and
  its comment says exactly that. No Settings-side callback is added here.
- The links blob is still decoded per keystroke, and so is the per-item fold on
  top of it: `scriptArgument` folds the query once per saved script link, and
  `matchingScriptLinks` walks the saved links twice per keystroke - once inside
  `matchingScriptLink`, once for the set of rows the answer replaces. Neither is
  fixed here. The cost is bounded by how many scripts the person saved, not by
  the pool, and the filter tests `kind == .script` before it folds anything, so
  a person with no scripts pays nothing and a person with five pays ten foldings
  against the roughly two thousand this removes. `CommandBarCatalog.build()`
  also still decodes the same blob from disk on every rebuild, so the fix worth
  making is one that removes that duplicate decode rather than adding a seventh
  cache beside it - a follow-up, not half of one here.
- The battery and memory rows now show the values from the previous opening for
  the moment before the background pass lands, which is what the storage row has
  always done. The alternative - keeping them synchronous and accepting the
  kernel calls - is the thing being fixed.

## Verification

`./build.sh --test` → `TESTS OK (9494 checks)`. The branch's own base,
`ed882cd3`, was run in a throwaway worktree and reports `TESTS OK (9487
checks)`; the seven added are the ones below, and there were no failures on
either side.

The two folded-query overloads are pinned by assertions that the folded and the
unfolded reading agree on an accented query, and that an empty memory is worth
nothing. The assign-before-guard order is pinned by a source-shape check over
both passes, in the style the file already uses for the shortcut-capture pair:
it slices `refreshStorageAnswer` and `refreshSystemAnswers` out of
`CommandBarService.swift` and asserts the store appears before
`guard changed else { return }`. All of these were confirmed to fail first -
making each overload look up the wrong key turns those red, and putting the
guard back in front of either assignment turns the ordering checks red with one
named failure per pass. Restoring both turns everything green.

`--test` compiles a whitelist. Of the files changed here it contains
`CommandBarPreferences.swift` and `CommandBarQueryMemory.swift` only;
`CommandBarService.swift`, `CommandBarCatalog.swift` and `SystemInfo.swift` are
not compiled here and CI covers them on both runners. The source-shape assertion
reads `CommandBarService.swift` as text, so it runs against the real file either
way.

Not tested: no run on real hardware. The main-thread cost this removes was read
off the call graph, not off a profile, so the claim is that the work is gone,
not a measured millisecond count. The background hop itself - the timing of the
battery and memory rows filling in - was not exercised by hand, and no GUI
testing was done.

## Anything else

#1245 depends on this branch and has to land after it.

Refs #887

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project are documented here. The format follows
 ## [Unreleased]
 
 ### Summary
-Vorssaint adds quit and close protections, expands screen text recognition, capture magnifier, clipboard, Super key, Dock Preview, App Switcher, Window Layout, mouse controls and snippet controls, broadens app update and safe cleanup discovery, and hardens shortcut capture and key caps, permission guide recovery, input session switching, Accessibility timeouts, Volume Mixer audio rendering, helper process attribution and teardown, Super key shutdown, process termination, update and uninstallation teardown, window handling and shared on-screen parsing, pasteboard restoration, sensor selection, Cleaning Mode unlock and favicon downloads. It also improves mouse scrolling feel, disk space and system monitor metrics, Settings, Scratchpad, Screenshot Editor, floating panels and several menu bar behaviors.
+Vorssaint adds quit and close protections, expands screen text recognition, capture magnifier, clipboard, Super key, Dock Preview, App Switcher, Window Layout, mouse controls and snippet controls, broadens app update and safe cleanup discovery, and hardens shortcut capture and key caps, permission guide recovery, input session switching, Accessibility timeouts, Volume Mixer audio rendering, helper process attribution and teardown, Super key shutdown, process termination, update and uninstallation teardown, window handling and shared on-screen parsing, pasteboard restoration, sensor selection, Cleaning Mode unlock and favicon downloads. It also improves mouse scrolling feel, disk space and system monitor metrics, Command Bar responsiveness, Settings, Scratchpad, Screenshot Editor, floating panels and several menu bar behaviors.
 
 ### Added
 - Settings now includes optional protections for Command-Q and Command-W with customizable hold duration, double press, extra modifier requirements and per-app scopes. Thanks to @RuanMD.
@@ -29,6 +29,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 
 ### Changed
 - Window controls, Dock clicks and mouse exceptions now share consistent on-screen window parsing while preserving their existing behavior. Thanks to @PathGao.
+- The Command Bar now stays responsive while typing, doing less work on each keystroke and reading the battery and memory levels in the background. Thanks to @PathGao.
 - System monitor process breakdowns now normalize per-app CPU usage against total active processor cores and cap grouped GPU and energy percentages at 100%. Thanks to @pergioa.
 - The docked shelf now uses a calibrated trigger area and requires a brief hover over the collapsed pill before expanding, preventing fast drags across the menu bar from opening the full card.
 - Smooth scrolling now feels consistent on standard and high-refresh displays, with adjustable speed and response and no lost wheel distance.
--- a/Sources/Vorssaint/Services/CommandBar/CommandBarCatalog.swift
+++ b/Sources/Vorssaint/Services/CommandBar/CommandBarCatalog.swift
@@ -1059,7 +1059,7 @@ enum CommandBarCatalog {
                                     bar: CommandBarFeatureStrings) -> [CommandBarEntry] {
         var entries: [CommandBarEntry] = []
 
-        if let battery = SystemInfo.batterySnapshot() {
+        if let battery = cachedBattery {
             let value = "\(battery.percent)%"
             let detail = battery.isCharging
                 ? bar.answerBatteryCharging
@@ -1075,7 +1075,7 @@ enum CommandBarCatalog {
         }
 
         if AppFeature.monitorMemory.isAvailable,
-           let memory = SystemInfo.memoryUsage(), memory.total > 0 {
+           let memory = cachedMemory, memory.total > 0 {
             let metric = Defaults.sanitizedMonitorMemoryMetric(
                 UserDefaults.standard.string(forKey: DefaultsKey.monitorMemoryMetric) ?? "")
             let selected = MetricFormat.selectedMemory(used: memory.used,
@@ -1150,6 +1150,13 @@ enum CommandBarCatalog {
     /// Filled in by the service from a background pass; nil until the first
     /// one lands, which simply means the row is not offered yet.
     static var cachedBootVolumeSpace: (free: UInt64, total: UInt64)?
+
+    /// The battery and the memory pressure are read the same way and for the
+    /// same reason: both cross into the kernel (IOKit power sources, the mach
+    /// VM statistics), and the bar opens on a keystroke.
+    static var cachedBattery: BatteryInfo?
+    static var cachedMemory: (used: UInt64, appUsed: UInt64, total: UInt64,
+                              compressed: UInt64, cached: UInt64, swapUsed: UInt64?)?
 
     static func readBootVolumeSpace() -> (free: UInt64, total: UInt64)? {
         let url = URL(fileURLWithPath: "/")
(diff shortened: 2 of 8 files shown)


PathGao:
The invalidation half is the part that could have gone wrong, and it holds. I grepped `main`: those four `@AppStorage` setters are the only writers of pins/aliases/hidden/disabled outside the service, `recordUse` / `resetRanking` / `forgetLearnedRanking` all reach a cache update, and backup restore relaunches. Ordering `disabledCache` ahead of `reloadFileSearchCaches()` is required — that call asks `isEnabled(.files)`.

Two things:

`refreshSystemAnswers` guards the assignment, so when used/appUsed/total match, `cachedMemory` keeps `compressed` / `cached` / `swapUsed` from an older sample. Nothing reads those today, so it is a trap rather than a bug — assign, then guard on the three fields the row actually shows.

The same per-item fold survives next door: `scriptArgument` folds the query once per saved script link, and `matchingScriptLinks` walks `savedLinks` twice per keystroke (inside `matchingScriptLink`, then again for the drop list). The body scopes out the links *decode*, not this.

Didn't verify the background hop on hardware either.

**Impact:** none beyond the four Settings writers now landing immediately.
**Merge:** ship it.

ref #887

PathGao:
@PathGao Both notes answered, one in code.

**The memory guard.** Fixed, and at the shape rather than the site. `refreshSystemAnswers` now stores the sample and guards afterwards, so `compressed` / `cached` / `swapUsed` cannot sit at an older reading behind the three fields the row shows. `refreshStorageAnswer` one function above had the same shape already on `main` — it guards on `free` alone and stores a tuple that also carries `total`, and unlike the memory case the storage row does read `total`, for the caption and the free-percentage — so it is reordered too. `refreshWiFiState` and `refreshAutomationStatus` compare the whole value they store and are untouched. `CommandBarExtras.cachedWiFiPower` and `finderAutomationDenied` are scalars, so there is nothing else of this shape in the file.

**The script-link fold.** Not fixed, deliberately, and now written into the body rather than left implied. It is bounded by how many scripts the person saved rather than by the pool, and `matchingScriptLinks` tests `kind == .script` before it folds anything — no scripts costs nothing, five scripts costs ten foldings against the roughly two thousand this removes. Fixing it properly wants the links blob cached first, and that needs the same Settings-side invalidation from an editor that also announces nothing, so it is one follow-up rather than half of one landed here.

**The body.** Corrected on the point your impact line names. The old text said a switched-off source would otherwise take effect only on the next opening, as though that were the hole being closed. `show()` calls `reloadPreferenceCaches()`, so all four edits land at the next opening with or without the call; and the bar is not on screen when they run, because a mouse-down outside the panel hides it and that includes one in another window of this app, which is the click that flipped the switch. So the four calls are insurance that keeps the writers honest about a cache they feed, not a fix for a visible stale value — the same reason the compact-mode toggle beside them is allowed to skip it, which its comment already says. That also settles whether landing immediately can reorder: there is no observer, and on the writing side `togglePin` / `setAlias` / `toggleHidden` all read `storedPins` / `storedAliases` / `storedHiddenKeys` off disk before writing, so neither side can write back a stale copy. The one real consequence is cost — `syncWithPreferences()` re-registers the global shortcut and every row hotkey, now four more times — and it is the call the file-scope writers beside them already make.

**Verified.** `./build.sh --test` → `TESTS OK (9494 checks)`; base `ed882cd3` in a throwaway worktree → `TESTS OK (9487 checks)`, no failures either side. The new order is held by a source-shape check over both passes, in the style the file already uses for the shortcut-capture pair. Red first: putting the guard back in front of either assignment gives one named failure per pass, restoring both goes green.

**Not verified.** No hardware run, so the background hop is still unmeasured — I did not check the battery and memory rows filling in on a real machine, and the four Settings-to-service notifications were not exercised by hand. `--test` compiles a whitelist that of the changed files holds only `CommandBarPreferences.swift` and `CommandBarQueryMemory.swift`; the service, the catalog, `SystemInfo.swift` and `CommandBarSettings.swift` are not compiled here and CI covers them. The source-shape assertion reads the service as text, so it runs against the real file regardless.

#1245 sits on this branch and has to land after it.

PathGao:
Reordering `refreshStorageAnswer` too was the right call — fixing only the one this branch introduced would have left the shape standing one function above it. I checked the claim it rests on: `MetricFormat.selectedMemory` picks between `used` and `app` and nothing else, so `compressed` / `cached` / `swapUsed` really are unread and this really was a trap rather than a bug. `refreshWiFiState` and `refreshAutomationStatus` do compare what they store, whole.

One note, not for this PR. The source-shape test exists because the background-hop-then-maybe-rebuild block is now four copies — those two plus the two untouched — and a text assertion keeps two of the four from drifting instead of removing the drift. One helper taking the read and the comparison would kill the class. Unless keeping them apart is deliberate, that belongs with the links-cache follow-up.

The marker pins `cachedMemory = memory` only; `cachedBattery = battery` is unpinned, harmless while the battery is compared whole. No hardware run on my side either.

**Impact:** none beyond the four Settings writers landing immediately.
**Merge:** ship it.

ref #887

PR #1245, opened 2026-09-01
perf(command-bar): fold each list of rows once per opening, not once per background pass

PathGao:
## Summary

One opening of the command bar rebuilt the whole search index seven to ten
times.

`index(_:)` folds every row's title and keywords, and the comment above it says
it happens there "once per rebuild" so that typing never pays for it. But it was
handed the flat concatenation of nine lists, and eight of those lists arrive at
different moments: the catalog and the running apps are built when the panel
appears, then the installed apps, the Mac's Settings panes, the open windows, the
front app's menus, the selection and the running processes each land from their
own background pass and each call `indexEntries()` on the way in. Every one of
those rebuilt all nine lists from nothing - 1000 to 1500 rows, two foldings each -
and all of it landed in the few hundred milliseconds while the panel was
appearing and the person started typing.

Each list now keeps its own folded copy and only the list that was assigned again
is folded. The invalidation is the assignment itself: each of the nine stored
properties carries a `didSet` that drops its folded copy, so no write site has a
separate "tell the index what changed" step to forget. `PoolSection` also becomes the single statement of
what the pool contains and in what order, which `indexableEntries` was repeating.

`rebuildCatalog()` stopped rebuilding the emoji rows every time. There are about
fifteen hundred of them - by far the largest thing being folded - and they read
exactly two things: the language, because their names are translated, and whether
Accessibility is granted, because typing at the caret is what the permission is
for and every emoji row carries the warning when it is missing. Both are
remembered and compared, so a grant mid-launch rebuilds them the same way a
change of language does. An earlier revision of this branch watched only the
language, which left those rows warning about a permission the person had just
given until the app was restarted; that is fixed here.

No other row list has that shape. Every other row carrying the permission warning
lives in `catalog`, `appEntries` or `selectionEntries`, all built again on every
opening. The one list besides the emoji rows that survives an opening is the
Mac's Settings panes, keyed by the language they were read in - those rows carry
no permission, no setting and no live state.

Trade-offs:

- The three lookup maps are still built from nothing on every call, deliberately.
  The dangerous version of this change is one that keeps `entriesByStableKey`
  across passes and lets a row that has left its list stay pressable - the menus
  of the app that was in front a moment ago are the case that matters, and the
  comment guarding it is still there. Merging cached folded lists into fresh maps
  keeps that property by construction: an id that is no longer in any list cannot
  reappear. What is saved is the folding, which is where the allocations are.
- A name the person gave is folded into the copies, so `aliasCache` clears all of
  them when it changes. That is one comparison per opening rather than a rule to
  remember at each write site.
- The `didSet` sits on each declaration rather than at the write sites, which
  covers every assignment to a list that has one - but nothing holds the pool at
  nine. `PoolSection` has nine cases today and nine properties carry a `didSet`;
  a tenth case whose array carries none would keep its first folded copy for the
  life of the service, because `index()` only folds a section whose entry is
  `nil`, and it would serve stale rows silently. That is a rule to remember when
  the pool grows. A general guard - deriving the invalidation from the section
  list instead of the declaration - is a different change and does not belong in
  a branch about how often the folding runs.
- Not done: caching the saved links. `searchRows(for:)` still decodes
  `commandBarLinks` per keystroke, and the decode worth removing is not that one
  on its own - `CommandBarCatalog.build()` decodes the same key off disk on every
  rebuild, one line after `reloadPreferenceCaches()` runs. Caching only the
  ranking side leaves the duplicate standing and adds a seventh cache beside it.
  The change worth making hands one decoded list to both, and needs no Settings
  callback to do it. It belongs in its own branch.
- Not done: keeping the copies across openings. `clearIndex()` on a new
  presentation leaves them, so the Settings panes - the one list that survives an
  opening - are folded once per launch, but everything else is folded once per
  opening. Going further would mean holding folded text for rows that are already
  gone, for a cost that is now paid once instead of ten times.

## Verification

Rebased onto `main` at `5a4afb15`, now that #1244 has landed. `./build.sh --test`
on this branch tip → `TESTS OK (9529 checks)`, no failures - which includes
#1244's own source-shape check that `refreshStorageAnswer` and
`refreshSystemAnswers` store their whole sample before
`guard changed else { return }`. That check is green, so the rebase kept `main`'s
reading of both passes rather than the earlier one this branch was cut from.

Two assertions were added, in `Tests/MetricsTests.swift`, next to the existing
Command Bar source-shape checks and in the same style. The first is that an emoji
row still reads `Permissions.shared.accessibility`; the second is that the guard
inside `rebuildCatalog()` which decides to keep those rows names both that grant
and the language, and stores the grant it compared against. Both were confirmed
red first: the same logic run against the commit before the emoji fix reports the
guard half false, and against this tip reports both true.

The store had to be named on its own. Deleting `emojiAccessibility = accessibility`
leaves every other asserted substring in place, so the check passed while the
field stayed `nil`, `emojiAccessibility != accessibility` stayed true and the
fifteen hundred emoji rows were folded again on every opening - the cost this
branch exists to remove, back with the suite green. That is now a fourth clause,
verified both ways on this branch: with the line deleted the run is
`TESTS FAILED (1 of 9529)` naming the emoji-rows check, and with it restored
`TESTS OK (9529 checks)`.

`CommandBarService.swift` is outside the `--test` whitelist, so the test binary
does not compile it. `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` with
the build's own target and SDK flags finishes clean, with no warnings; CI covers
the real build on both runners. The full `./build.sh` was not run: with the
signing keychain locked `codesign` blocks on an unlock dialog and never exits.

Not tested: no run on real hardware. The count of index passes per opening is
read off the call graph rather than measured, and the stale warning was read off
`@Published accessibility` in `Permissions` and the guard's inputs, not off an
emoji row changing after a real grant. The paths that would show it - opening the
bar with menus, windows and a selection all landing, and granting Accessibility
with the bar already open - were not exercised by hand.

## Anything else

No issue reports the cost this removes, and searching the tracker for the
symptom found none to reference, so there are no `Refs` lines. It came out of
reading the opening path, not out of a report.

#1244 has landed and this is rebased onto it, so the diff here is two commits
against `main` and touches `CommandBarService.swift` and `Tests/MetricsTests.swift`
only. The battery and memory work, the preference caches and the query folding
all belong to #1244 and no longer appear.

Dropped in the rebase: the `linksCache` commit and the two `CommandBarSettings`
callbacks that came with it - see the trade-off above for why.


PathGao:
The `didSet` invalidation is the right shape — a stored array fires it on in-place mutation too, so no write site can forget it, and `PoolSection` makes the pool one statement instead of two.

The emoji skip misses one input. `CommandBarCatalog.emojiEntries` sets `trouble: Permissions.shared.accessibility ? nil : .needsPermission`, so those rows depend on more than the language — the body and the new comment both say they don't. Grant Accessibility mid-launch and ~1500 emoji rows keep the orange "needs permission" subtitle until relaunch; every other row carrying that flag lives in `catalog`, `appEntries` or `selectionEntries`, all rebuilt per opening, so they clear. It also makes `freshFullEntry(forStableKey:)` hand back a cached emoji entry, which its own comment says it never does.

Remember the grant the way `builtLanguage` is remembered:

```swift
if emojiEntries.isEmpty || builtLanguage != L10n.shared.language
    || emojiAccessibility != Permissions.shared.accessibility {
```

Not verified: no hardware run — I read the flip off `@Published accessibility` in `Permissions`, not off a row changing.

**Impact:** after granting Accessibility, emoji rows show a stale permission warning for the rest of the launch.
**Merge:** needs the emoji guard to include the Accessibility flip; then behind #1244.

PathGao:
@PathGao Closed the named gap.

The gap: `rebuildCatalog()` kept the emoji rows on the language alone, and
`CommandBarCatalog.emojiEntries` sets `trouble: Permissions.shared.accessibility ? nil
: .needsPermission`. Grant Accessibility mid-launch and about fifteen hundred rows kept
the orange subtitle until relaunch, and `freshFullEntry(forStableKey:)` handed back a
cached emoji row against its own comment. The grant is now remembered in
`emojiAccessibility` next to `builtLanguage` and compared the same way, so either one
moving rebuilds the rows - your snippet, with the read hoisted into a local so the guard
and the store cannot drift apart.

Where else the shape appears: I walked every `trouble:` site in `CommandBarCatalog` and
every list in `PoolSection`. Your reading holds - the other permission-carrying rows are
in `catalog` (`toggleEntries`, `actionEntries`, `snippetEntries`), `appEntries` or
`selectionEntries`, all assigned on every opening. The only other list that survives an
opening is `macSettingsEntries`, keyed by `macSettingsLanguage`; `CommandBarCatalog.macSettingsEntries`
builds rows with no `trouble`, no `isActive` and no defaults read, so it has nothing that
can flip underneath it. The folded copies need no extra key: `foldedSections` is dropped
by the `didSet` on the assignment, and folding reads only the row's own title, keywords
and alias.

Verification: `./build.sh --test` on the tip → `TESTS OK (9493 checks)`. The new assertion
is in `Tests/MetricsTests.swift` beside the existing Command Bar source-shape checks -
one half pins that an emoji row still reads `Permissions.shared.accessibility`, the other
that the guard names both that grant and the language, so removing either input breaks the
test rather than silently un-keying the cache. Red first: running the same slicing and
substring logic against the previous commit of this branch reports the guard half false
and the fixed tree true, so the assertion is not vacuous. `CommandBarService.swift` is
outside the `--test` whitelist and was not compiled here; CI covers it on both runners,
and the full `./build.sh` was not run because codesign blocks on a locked keychain.

Not verified: no hardware run. I did not watch an emoji row lose its warning after a real
grant - the fix is read off the guard's inputs and off `@Published accessibility` in
`Permissions`, same limit as your own note.

The body now carries `**Depends on #1244.**` under `## Anything else`. I left the `blocked`
label alone.

PathGao:
Guard closed, and hoisting the read into a local is better than what I wrote. I also checked the other half of my own note: the emoji `run` closure captures only the character and re-checks `AXIsProcessTrusted()` itself, so a cached row out of `freshFullEntry` is harmless by construction, not just by the guard.

Three things, none of them emoji.

`searchRows(for:)` still decodes `commandBarLinks` off UserDefaults on every keystroke — the one defaults read in that function the four new caches skipped, and a JSON array decode, heavier than the string splits that did get cached.

The battery and memory rows moving to a background pass is not in the body. What the person sees: on the first opening after launch neither row exists until the pass lands. Same as storage, whose comment says so out loud; `cachedBattery`'s doesn't.

`CommandBarSettings` now carries two rules. The new comment says writing the key alone leaves a switch waiting for the next opening; the compact-mode toggle above says no callback is needed because the bar rereads on open and Settings has already hidden it. `save` and `removeLink` follow the second. One is wrong.

Not verified: outside the diff I read `main`'s `src/`, not the branch tip. No hardware run, no compile here.

**Impact:** none new — one decode left on the keystroke path, and a first-opening gap on two rows that is not written down.
**Merge:** needs the links decode off the keystroke path; then behind #1244.

PathGao:
@PathGao All three, and the guard note is right - the emoji `run` closure re-checking `AXIsProcessTrusted()` itself is why a cached row was only ever a cosmetic problem.

**The links decode.** `searchRows(for:)` now ranks against `linksCache`, decoded in `reloadPreferenceCaches()` beside the aliases, pins, hidden keys, sources and usage counts. That helper runs immediately before `rebuildCatalog()`, which builds the link rows off the same key, so the two readings of `commandBarLinks` are of the same stored value at the same moment - the same relationship they had when both read disk, kept rather than introduced. `DefaultsKey.commandBarLinks` now appears once in `CommandBarService.swift` and a test holds it to one.

Where else the shape appears: `searchRows(for:)` has no `UserDefaults.standard` read left at all, which was the last one. I walked the other `@AppStorage` keys on the Settings page for the write half. Seven controls hand the new value to the service; the compact-mode toggle deliberately does not, and `save`/`removeLink` were the two that wrote a cached key without saying so. Both now do.

**The two rules.** You are right that one is wrong, and it is mine. The bar hides on the first click outside its panel - `installMonitors` hides on a local click outside, a global click, and any other app activating - so it can never be on screen while a Settings control is being operated. The compact-mode comment states the true rule and it holds for every control on that page. My comment on the three list removals claimed writing the key alone would leave a switch waiting for the next opening; it would not. They call the service so a decoded copy does not outlive the write that replaced it, and the comment now says that instead. Nothing was removed - the calls are cheap and the invariant they keep is real, it just is not the one I wrote down.

**The first-opening gap.** Written into the body now, under its own sentence rather than buried in the trade-offs: on the first opening of a launch the battery row and the memory row are not in the list until the background pass lands, and the list refreshes when it does. Every opening after that has them. `cachedBattery` and `cachedMemory` carry the sentence `cachedBootVolumeSpace` already had.

**Verification.** `./build.sh --test` on the tip → `TESTS OK (9497 checks)`. Two assertions added beside the emoji one, same source-shape style: one that the ranking pass reads the cache and `reloadPreferenceCaches()` is what fills it, one that both link writers call `syncWithPreferences()`. Red first - running the same slicing and substring logic against the previous commit of this branch reports both halves false and this tip true, so neither is vacuous. The three files touched are outside the `--test` whitelist, so instead of the full `./build.sh` I ran `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` with the build's own target and SDK flags: clean, no warnings. The full build still is not run here - codesign blocks on the locked keychain.

Not verified: no hardware run. I did not watch the first opening of a launch to see the battery row arrive late, and I did not save a link with the bar somehow visible - the claim that it cannot be is read off the monitors in `installMonitors`, not off the screen.

PathGao:
@PathGao Rebased onto `main` at `5a4afb15`, now that #1244 has landed, and force-pushed. Two commits left against `main`, touching `CommandBarService.swift` and `Tests/MetricsTests.swift` only: the per-list folded index, and the emoji rebuild on an Accessibility grant.

Three things worth saying about what the rebase resolved.

**The battery and memory passes took `main`'s side.** This branch was cut from #1244's first commit, so it carried the reading where `guard changed else { return }` stood before `CommandBarCatalog.cachedBattery = battery`, without the source-shape check #1244 added when it reordered them. That is now a real conflict rather than a latent one, and it is resolved to `main`: the store comes first and #1244's check is green in the 9529 above. Nothing about those two passes appears in this diff any more.

**`CommandBarQueryMemory` took `main`'s side too.** The folded overloads are `main`'s, including dropping the empty-picks short circuit from the string overload. This branch does not touch that file.

**The links cache is gone, with its two `CommandBarSettings` callbacks.** It was added while this was under review, and it cached a duplicate rather than removing one: `CommandBarCatalog.build()` still decodes `commandBarLinks` off disk on every rebuild, one line after `reloadPreferenceCaches()` runs. The decode worth removing hands one list to both the rows and the ranking, and needs no Settings callback to do it. That belongs in its own branch. `searchRows(for:)` decodes per keystroke again, as on `main`.

`./build.sh --test` → `TESTS OK (9529 checks)`. `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` with the build's target and SDK flags is clean, since `CommandBarService.swift` is outside the `--test` whitelist. Still no run on real hardware.

PathGao:
Rebase reasoning checks out, including the part that reverses my own merge condition: `CommandBarCatalog.swift:180` and `CommandBarService.swift:1215` are the two `commandBarLinks` decodes, so caching the ranking side would have made three. Dropping it was right. `blocked` comes off — #1244 landed.

Two gaps, both in the guard rails rather than the behavior.

The body says removing either input breaks the test. Deleting `emojiAccessibility = accessibility` does not: all four substrings survive, the field stays `nil`, the guard is always true, and the fifteen hundred rows fold on every opening again — the cost this PR exists to remove, gone silently. One clause:

```swift
&& rebuildCatalogCode.contains("emojiAccessibility = accessibility")
```

The body also says a future list cannot be added to the pool and quietly serve stale rows. The `didSet` is the step to forget; it sits on the declaration instead of the write sites. A new `PoolSection` case whose array has no `didSet` keeps its folded copy forever. Nine today, nothing holds them to nine.

Not verified: I read `main`'s `src/`, not the branch tip. No compile, no hardware run.

**Impact:** none — both are about what a later change can break unnoticed.
**Merge:** needs the store pinned in the test; then ship it.

PathGao:
@PathGao Pinned the store in `62051826`, and rewrote the two body claims you named. Plain commit, no force-push.

I checked the failure before adding the clause. Deleting `emojiAccessibility = accessibility` from `rebuildCatalog()` and running `./build.sh --test` gave `TESTS OK (9529 checks)` — all four substrings survive the deletion, the field stays `nil`, `emojiAccessibility != accessibility` is true on every call, and the fifteen hundred rows fold again on every opening with the suite green. You had it exactly right. With `&& rebuildCatalogCode.contains("emojiAccessibility = accessibility")` added, the same deletion reports `TESTS FAILED (1 of 9529)` naming the emoji-rows check, and restoring the line is `TESTS OK (9529 checks)`. Count is unchanged because the clause joined the existing `expect` rather than adding one.

The body said removing either input would break the test. It would not, and that sentence is gone; the Verification section now says which four things the check names and what the fourth one was added for.

On the `didSet`: no code change, as you scoped it. The Summary claimed a future list could not be added and quietly serve stale rows, and that claim does not hold — the `didSet` is on the declaration, so it covers every write to a list that has one, but nothing holds the pool at nine. `PoolSection` has nine cases and nine properties carry a `didSet`; a tenth case whose array carries none keeps its first folded copy for the life of the service, because `index()` only folds a section whose entry is `nil`. That is now written into Tradeoffs as a rule to remember when the pool grows, together with why the general fix — deriving invalidation from the section list rather than the declaration — is a different change and not this one.

Worth rechecking after this: the check now pins the guard's three reads and its one write, all by source text. It still says nothing about `emojiEntries` being assigned only inside that branch — a write to it somewhere else would defeat the whole thing and the check would stay green.

`./build.sh --test` on the tip: `TESTS OK (9529 checks)`, no failures, and neither known input-method false red appeared on this machine today. `CommandBarService.swift` is unchanged by this commit and still outside the `--test` whitelist.

PathGao:
The store pin is right, and dropping the Summary claim instead of patching it was the better call — nothing holds the pool at nine, and a rule in Tradeoffs beats a sentence that isn't true.

Your recheck: the write outside that branch is already there. `syncWithPreferences()` sets `emojiEntries = []` on the uninstall path, and it is safe today only because of `emojiEntries.isEmpty ||` — the one clause of the guard your check does not name.

Delete that clause and the suite stays green. Uninstall the Command Bar from the hub and reinstall it in the same launch: `noteHubChange()` calls `rebuildCatalog()`, the language and the grant are both unchanged, so the rows are never rebuilt and stay empty until relaunch. First launch does not need the clause — `builtLanguage` is `nil` there — so that path is its only job.

Same `expect`, fifth clause:

```swift
&& rebuildCatalogCode.contains("emojiEntries.isEmpty")
```

Not verified: read off `main`'s `src/`, not the branch tip. No compile, no hardware run, and I did not exercise a hub uninstall and reinstall.

**Impact:** none today — the clause is in the guard. It is the input the test cannot see going missing.
**Merge:** needs the `isEmpty` clause in the same expect; then ship it.

PathGao:
@PathGao Pinned the emojiEntries.isEmpty arm in the rebuild-guard assertion (623681d7) and merged origin/main (e26d9f99). The conflict was add/add at the end of MetricsTests.swift: this branch's emoji-rebuild block against main's quit-HUD width checks (bfd534e7) and dropped-identifier checks (e1fc98e1); all three blocks are kept. TESTS OK (9601 checks); whole-module swiftc -typecheck of Sources/Vorssaint exit 0.

PathGao:
Both arms pinned, and merging rather than rebasing was the right call with the branch already reviewed — the three test blocks are independent additions landing at the same place, nothing to reconcile beyond position.

Read the merged diff, nothing left to flag. `PoolSection.allCases` still lists the nine in the old concatenation's order, so `freshFullEntry`'s `.last` picks the same row it always did, and the assertion now names all three guard inputs plus the read and the store.

One body sentence, no code. `PoolSection` is called the single statement of what the pool contains and in what order. It is that for the index. The ranking pool in `searchRows(for:)` states its own membership and order and is not derived from it — no `killProcessEntries`, `quitEntries` only on the quit verb, menus and emoji only from two letters, clipboard and file rows appended after. A tenth list has two places to be added, not one, which is worth saying right where the tradeoff note below it is about that exact failure.

Not verified: I read `main`'s `src/` for `searchRows(for:)` — it is not in this diff, so it is unchanged on the branch. No compile, no hardware run.

**Impact:** none.
**Merge:** ship it.

PathGao:

PathGao:
Reopened as #1296.

PR #1246, opened 2026-09-01
perf(uninstaller): look up embedded code where macOS puts it

PathGao:
# Stop repeating the installed-app walks, and keep the one that denies an ownership claim

## Summary

Two hot paths in the uninstaller and the cleaner did a full recursive walk of
app bundles to answer questions that need a handful of directory reads.

**Uninstaller.** `allBundleIDs` and `signingIdentity` each opened
`FileManager.enumerator(at:)` on an app bundle with no depth limit and no
pruning. The `inspected <= 256` cap looked like a bound but only counted
directories that had *already* matched `.appex` / `.xpc` / a login item, and an
ordinary app matches none — so the enumerator ran to the bottom of the bundle
every time, doing a `resourceValues` call per entry.

macOS only ever loads embedded executable code from three places, so when the
question is *what does the app being uninstalled own*, look there directly:
`Contents/PlugIns`, `Contents/XPCServices`, `Contents/Library/LoginItems`, plus
one nesting level for a login item helper's own extensions.
`UninstallerSupport.embeddedCodeURLs` does that, and the four lookups that read
the target itself use it.

**The other direction keeps the walk.** The same two functions are also run over
every *other* installed app, and that result is what withholds a claim:
`applicationBundleIdentifiers` feeds `sharedDataIsExclusive`, and the loop in
`exclusiveGroupIDs` collects the groups declared elsewhere. A miss there does
not shrink what gets claimed, it grows it — an owner the uninstaller cannot see
reads as "nobody else owns this". Nothing forces a product to nest code in the
reserved directories: Sparkle ships its updater XPC services under
`Contents/Frameworks/Sparkle.framework/Versions/*/XPCServices`, so an installed
app that owns `org.sparkle-project.*` only through the framework would have gone
unseen and its shared folder would have been trashed while uninstalling someone
else. Those two lookups run `UninstallerSupport.nestedCodeURLs`, which is the
old whole-bundle walk, kept for exactly this.

That is the cost of the fix, stated plainly: the sweep over every installed app
is as slow as it was before this PR. What is left of the uninstaller saving is
the target's own bundle — four walks of one app become directory reads — and
`removeSelected` no longer calling `knownApplicationURLs` unconditionally while
every consumer of the result sat behind `mayClaimSharedData`. It is now computed
inside that branch, matching what `select()` already did.

**Cleaner.** `mayRemove` called `installedBundleIDs()` — three levels of the
application folders, a `Bundle(url:)` per app found, plus a sweep of
`NSWorkspace.runningApplications` — from inside `cleanSelected`'s removal loop,
so it ran once per leftover and again per item in the Finder retry pass. The
scan already builds that set and passes it down; only the removal pass threw it
away.

Hoisting it to one build per pass fixed the recomputation but moved the walk
onto *every* clean, including the ones that never consult it. Leftover rows
start unchecked — `CleanerPolicy.precheckLeftovers` is `false` — so the default
selection holds no leftover rows, and the common clean would have gone from zero
walks to one in a change about removing walks. The oracle is now built only when
the selection contains a leftover row:

```swift
let installed = chosen.contains { $0.category == .leftovers }
    ? Self.installedBundleIDs() : []
```

`mayRemove` reads `installed` inside its `.leftovers` branch and nowhere else,
so the empty set changes nothing for the other categories, and the Finder retry
pass filters a subset of the same selection. The scan path keeps its
unconditional build: `scanLeftovers` and `scanOrphanedLaunchPlists` both consume
it and both categories run on every scan.

### What a wrongly claimed leftover costs

It goes to the Trash. Not unlinked — `trashItem` — so it is recoverable, but the
user was told the folder belonged to the app they just removed, so nobody has a
reason to look. The folders in question are `Application Support`, `Containers`
and preference files: another installed product silently loses its accounts,
licence, or documents, and finds out the next time it launches. That asymmetry
is the whole reason the denial side is not allowed to be an approximation. A
missed claim leaves a folder behind and the user can delete it later; a wrong
claim takes something the user never chose to remove.

### Trade-offs

The fixed-location lookup is deliberately narrower than the walk: an `.appex` or
`.xpc` parked somewhere non-standard is not found. On the target that only
shrinks what is claimed — a leftover goes unlisted, or an application group is
not treated as exclusively owned, so a file is *kept*. That is the direction the
uninstaller is allowed to be wrong in, and it is the only direction the narrow
lookup is used.

The alternative considered for the denial side was a bounded walk — a depth cap
plus `skipDescendants()` on `Contents/Resources`. It was rejected for the same
reason as the fixed-location lookup: any bound is a guess about where a third
party put its code, and being wrong about that deletes someone's data. The
`.prefix(256)` cap stays at both call sites, and `nestedCodeURLs` stops at 256
matches itself, so a pathological bundle still bounds the signature work.

The cleaner change is behaviour-preserving. Nothing can install an application
mid-clean whose survival this pass would have had to respect, which is the only
way a per-item recomputation could have differed from a per-pass one. The
leftover gate rests on `mayRemove` reading the oracle under `.leftovers` alone;
that premise is pinned by a check rather than left to the reader.

## Verification

`./build.sh --test` on macOS 26 (Darwin 25.6.0), Apple silicon: `TESTS OK (9496
checks)`.

Four checks cover the embedded-code lookups against one temporary bundle holding
a plug-in, an XPC service, a login item helper with its own nested extension, a
`.bundle`, a symlink, and two extensions parked outside the reserved directories
— one buried in `Resources`, one in the `Framework/XPCServices` shape Sparkle
uses. Two pin `embeddedCodeURLs` to the reserved locations and its nesting
depth; two pin `nestedCodeURLs` finding the two strays as well, and stopping at
its limit.

`AppUninstaller.swift` is not in the `--test` whitelist, so the two denial sites
are pinned by source-shape checks that read the file at run time, in the style
already used here for `JunkCleaner`: `applicationBundleIdentifiers` and the
`exclusiveGroupIDs` loop each have to pass `exhaustive: true`. Both were
confirmed red on their own break against the already-built test binary, no
rebuild involved — dropping the argument from either call reddened that check
alone, and both went green on revert.

The two `nestedCodeURLs` checks needed a recompile to break, so I confirmed them
outside `build.sh`: the real `UninstallerSupport.swift` plus a `main.swift`
carrying the same fixture and the same two assertions, compiled at the build's
target and SDK. Delegating `nestedCodeURLs` to `embeddedCodeURLs` reddened the
first; removing the limit break reddened the second; the unmodified file was
green. That is the assertion logic exercised against the shipped function, not a
run of `build.sh --test` in a red state.

The three cleaner source-shape checks from the previous round are unchanged and
still green.

**What I could not test.** `AppUninstaller.swift` and `JunkCleaner.swift` are
not compiled by the test binary, so a green `--test` run is not evidence that
they build; CI's Build & selftest covers that on both runners. Locally I ran a
whole-module `swiftc -typecheck` over `Sources/Vorssaint/**/*.swift` with the
same target and SDK the build uses, which completed with no errors and no
warnings — that covers type checking of both files, not the full build. The full
`./build.sh` was not run.

Neither uninstalling an app nor running a clean was exercised against real
files. No app was removed, so the claim that the denial now sees a
framework-nested owner is read off `nestedCodeURLs` and a fixture in that shape,
not off a real Sparkle-shipping app. Both lookups have still only been run
against the test fixture, never against a real installed bundle. The 288851
figure in the original report is a `find` count on this machine only, and I did
not profile before and after — the argument is the removed work, not a measured
wall-clock delta, and the denial walk that stays is the larger half of it.

## Anything else

No issue reports either symptom. I searched all 800 issues for the cleaner,
uninstaller, leftover and junk terms, for the shared-data and application-group
vocabulary, and for the performance words (`slow`, `lag`, `freeze`, `hang`,
`performance`), and checked the three long-lived index issues for this PR number
and for both symptoms. The cleaner and uninstaller issues that exist are feature
requests, not reports of this cost or of a wrongly removed file, so there is no
`Refs` line to add.


PathGao:
The `mayClaimSharedData` gate on `knownApplications` holds — all three consumers (`exclusiveOwnedBundleIDs`, `exclusiveBundleIDs`, `exclusiveGroupIDs`) sit behind that same branch, so the empty array can't leak into a claim. Pinning the extensions parked outside the reserved directories as *expected* misses is the right test to have.

One regression in the cleaner. `installedBundleIDs()` used to run only inside `mayRemove`'s `.leftovers` branch. It now runs for every clean, and `CleanerPolicy.precheckLeftovers = false` means the default selection has no leftover rows at all — so the common clean goes from zero walks to one, in a PR about removing walks. Guard it:

```swift
let installed = chosen.contains { $0.category == .leftovers }
    ? Self.installedBundleIDs() : []
```

`mayRemove` reads `installed` only under `.leftovers`, and `stubborn` is a subset of `chosen`, so `[]` is safe there.

The "these two files may not compile" caveat is answered by CI: Build & selftest is green. I only checked `embeddedCodeURLs` against the test fixture, not a real bundle.

**Impact:** any clean without leftover rows selected — which is the default — pays an application-folder walk it did not before.
**Merge:** needs the leftovers guard first.

PathGao:
@PathGao Guard added in `131b0f61`, in the form you wrote it.

```swift
let installed = chosen.contains { $0.category == .leftovers }
    ? Self.installedBundleIDs() : []
```

The gap was mine to begin with: hoisting the oracle out of the removal loop
fixed the per-item recomputation but paid for it on every clean, and with
`CleanerPolicy.precheckLeftovers = false` the default selection has no leftover
row to justify the walk.

I checked the other side of the walk before adding the guard. `installedBundleIDs()`
has two callers. The one in `scan()` is unconditional and has to stay that way:
`scanLeftovers(installed:)` and `scanOrphanedLaunchPlists(installed:)` both consume
it and both categories run on every scan, so there is no selection to gate on there.
That leaves `cleanSelected` as the single site, and the three `mayRemove` calls
inside it all read the one hoisted value, so the guard covers the removal loop, the
login-item retry and the Finder pass together. `mayRemove` reads `installed` only
inside its `if item.category == .leftovers` branch, which is what makes `[]` safe
for everything else.

`JunkCleaner.swift` is outside the `--test` whitelist, so the check is three
source-shape assertions in `MetricsTests`, in the style already used for
`AutoQuitService` and `DockClickService`. They pin the gate and the premise it
rests on: the clean builds the oracle only behind the leftover test, and
`mayRemove` reads it exactly once, inside the leftovers branch. That second half
matters more than the first — if a later change reads the oracle outside the
branch, the guard turns from a perf gate into a safety hole silently.

Each assertion was confirmed red on its own break. Because they read the file at
run time and `JunkCleaner.swift` is not linked into the test binary, I could flip
the source and re-run the already-built `build/metrics-tests` without a rebuild:
dropping the ternary back to an unconditional call reddened the first, moving the
single oracle read above the `.leftovers` branch reddened the third, and adding a
second read inside the branch reddened the second. All green again on revert.
`./build.sh --test` on the final tree: `TESTS OK (9492 checks)`.

Not verified: no clean was run against real files, so the claim that a leftover-free
selection still removes exactly what it did is read off `mayRemove`, not observed.
I did not run the full `./build.sh`; a whole-module `swiftc -typecheck` over
`Sources/Vorssaint/**/*.swift` at the build's target and SDK came back with no
errors and no warnings, which covers type checking of `JunkCleaner.swift` and
`AppUninstaller.swift` but is not the build. Your point about `embeddedCodeURLs`
still standing on the fixture rather than a real bundle is unchanged — I have not
closed that one, and it is now stated in the body.

PathGao:
Guard is in, and the `installedBundleIDs()` audit holds — I re-grepped `main`: two callers, and `scan()` really can't be gated. The source-shape checks fail closed if the file moves or the branch is refactored.

The body overclaims one thing. "It cannot make the uninstaller claim a file belonging to another product" — `applicationBundleIdentifiers` runs `allBundleIDs` over *every installed app*, and `sharedDataIsExclusive` uses that set to **deny** a claim. Narrowing the lookup narrows the denial side too, and two paths never intersect candidates with the target's own `allBundleIDs`, so they don't shrink in step: `exclusiveEvidenceBundleIDs` (evidence IDs come off the scan rows, not the walk) and `exclusiveOwnedBundleIDs(requiresCurrentOwnership: false)`.

Concrete: Sparkle ships its XPC services at `Frameworks/Sparkle.framework/Versions/*/XPCServices`. Installed app B owns `org.sparkle-project.*` only there; uninstalling A now sees no other owner and trashes the shared row. Your fixture pins that exact shape as an expected miss.

Not verified: I did not run it, and I don't know how often evidence identifiers land on framework-nested code.

**Impact:** a leftover named by evidence and owned by another installed app only through framework-nested code is now claimed instead of kept.
**Merge:** needs the body's safety claim narrowed, or the denial side kept broad.

PathGao:
@PathGao Took the second option: the denial side is broad again, in `c3b6e8e3`.
I did not narrow the prose, because the prose was describing the wrong half of
the change and the gap you found is a file getting trashed, not a sentence.

You named `applicationBundleIdentifiers`. The same shape is one function below
it: `exclusiveGroupIDs` loops over `knownApplications` running
`signingIdentity(in: url, requireValidSignature: false)`, collects
`claimedElsewhere`, and subtracts. Identical direction, identical consequence —
a group container kept by an owner the narrow lookup cannot see. Both are fixed;
the four lookups that read the target itself (`allBundleIDs` at `select`, the
one inside `exclusiveOwnedBundleIDs`, and `signingIdentity` on the target in
both paths) stay on the reserved-location path, where a miss can only leave a
leftover behind.

The old walk is restored as `UninstallerSupport.nestedCodeURLs`, and
`allBundleIDs` / `signingIdentity` take an `exhaustive` flag that picks between
the two. Your Sparkle example is now the fixture's
`Contents/Frameworks/Core.framework/XPCServices/Stray.xpc`: still an expected
miss for `embeddedCodeURLs`, and an expected hit for `nestedCodeURLs`.

Say the honest thing about the cost: this gives back the larger half of the
uninstaller's saving. The sweep over every installed app is as slow as it was
before the PR. What is left is the target's own bundle and the
`knownApplicationURLs` gate in `removeSelected`. The body says that in the
Summary now, along with what a wrong claim costs — the folder goes to the Trash
under the name of the app the user did remove, so the product that actually owns
it loses its data and nobody has a reason to go looking.

Checks: `TESTS OK (9496 checks)`. Two source-shape assertions pin
`exhaustive: true` at each denial site — `AppUninstaller.swift` is outside the
`--test` whitelist — and both were confirmed red on their own break against the
already-built test binary, no rebuild. The two new `nestedCodeURLs` assertions
needed a recompile to break, so I confirmed those outside `build.sh`: the real
`UninstallerSupport.swift` plus a `main.swift` with the same fixture and the
same two assertions, compiled at the build's target and SDK; delegating
`nestedCodeURLs` to `embeddedCodeURLs` reddened the first, removing the limit
break reddened the second, unmodified was green.

Not verified: no app was uninstalled, so the denial seeing a framework-nested
owner is read off the fixture and the code, not off a real Sparkle-shipping app.
I did not run the full `./build.sh`; a whole-module `swiftc -typecheck` over
`Sources/Vorssaint/**/*.swift` at the build's target and SDK came back with no
errors and no warnings, which covers type checking of both files but is not the
build. I also have no measurement of how often evidence identifiers land on
framework-nested code — that is the question your last comment left open, and
restoring the walk is what makes the answer not matter.

PathGao:
Both denial sites are covered. I grepped `main`: `allBundleIDs` and `signingIdentity` have six call sites between them, and `applicationBundleIdentifiers` plus the `exclusiveGroupIDs` loop are the only two that pass a URL that isn't the target. No third one hiding.

One residual, weaker than the last. `ownedBundleIDs` at `select` is now narrow, and it is also the key set for `knownApplicationURLs`: `urlsForApplications(withBundleIdentifier:)` runs once per key, and that is the only route into the denial roster for an app outside the three scanned roots that isn't running. Each candidate's own key is still queried, so the direct protection holds. What drops out is cross-ID — an app registered under an identifier the narrow lookup missed, which also owns a candidate that survived. Two coincidences, not one. If you want the symmetry it is one argument: exhaustive set as the roster key, candidates stay narrow, one extra walk of the target only.

Not verified: whether `urlsForApplications` returns the containing app for an embedded appex's identifier, which is what decides how much that lookup was carrying.

**Impact:** none on the single-condition path; the cross-ID case is third-order.
**Merge:** ship it.

PR #1247, opened 2026-09-01
perf: build the process tree from one ps and skip the same-volume rehash

PathGao:
# Two pieces of work neither feature needs

## Summary

Both changes remove work whose result is already known.

**Killing a process tree forked a helper per process.**
`KillProcessService.descendants(of:)` ran a BFS in which every pid in the
frontier got its own `Shell.run("/usr/bin/pgrep", ["-P", pid])`. Tearing down a
tree of a few hundred processes meant that many fork/exec round trips, run one
after another on a background queue, while the list had already optimistically
removed the rows — so the interface says the processes are gone and the app is
busy spawning `pgrep`.

The parent/child edges were already available: `snapshot()` runs
`ps -eo pid,ppid,pcpu,rss,comm`, `KillProcessEntry.ppid` is the parsed result,
and `groupedByApp` builds its tree from those edges in memory. `descendants` now
runs one `ps -eo pid,ppid` and hands the rows to a new pure
`KillProcessSupport.descendants(of:parents:)`, which does the BFS with no
subprocesses at all.

**Organizing a WhatsApp download hashed the same bytes twice.** The same-volume
branch of `moveVerified` called `moveItem` — a rename on one volume, pure
metadata, not a single byte read or written — and then ran SHA-256 over the
destination to check it against the digest. The digest was already taken from
the source at `WhatsAppDownloadOrganizer.swift:368`, and the destination path is
one nothing can be sitting at: `moveVerified`'s two callers pass either a path
from `uniqueDestination`, which only returns a name that does not exist, or, from
`replaceVerified`, the existing file's own path — which `replaceVerified` has
already emptied by moving that file to a staging name before it calls in.
`moveItem` throws rather than overwrite in either case. Organizing a 4 GB video
therefore read 8 GB to learn something a rename cannot change.

**The volume comparison is the whole guarantee, so it uses `isEqual`.** Skipping
the hash makes the same-volume decision a safety check rather than a hint: get it
wrong and `moveItem` falls back to a real cross-device copy that nothing then
verifies. That decision was comparing `String(describing:)` on two
`volumeIdentifier` objects. The type is opaque and defines only `isEqual`; its
`description` is an implementation detail, and two different volumes whose
identifiers happen to print alike would have taken the unverified path. It now
goes through a pure `WhatsAppDownloadSupport.isSameVolume(source:destination:)`
that calls `isEqual` and answers false when either identity is unknown — a wrong
answer there costs a verified copy, which is the direction worth being wrong in.

That is the app's second volume comparison. The first,
`CutPasteProgressSupport.isCrossVolume` behind `FinderCutPaste.volumeIdentity`,
already used `isEqual` and is untouched; its opposite fallback for an unknown
identity is right for what it decides, which is only whether a progress bar
appears. Those two are the only readers of `.volumeIdentifierKey` in the app.

### Trade-offs

**The `ps` snapshot races the same way `pgrep` did.** A grandchild forked
between the snapshot and the signal is missed. That is not new: the per-pid walk
had exactly the same window, only wider, because it took longer. Every target
still goes through `killBatch`, which re-checks `identityMatches` before it
signals anything, so a reused pid is not at risk either way. Reading
`self.entries` instead of shelling out at all was considered and rejected: with
grouping on, `groupedByApp` collapses rows, so the ppid edges there are
incomplete.

The in-memory BFS visits each pid once. That is slightly stricter than the
`pgrep` version, which relied only on the 4096 cap, and it means a circular or
self-parenting row in a malformed table terminates instead of burning the cap.

**`descendants` still reports an empty tree when `ps` fails.** A failed `ps`
gives `[]`, so `killTree` signals the parent alone and reports success, where
`snapshot()` in the same file returns `nil` for that case. Making it `[pid_t]?`
so `killTree` can bail is the better shape, and it is deliberately not in this
PR: it changes what killing a tree does when the system is in trouble, which is a
behaviour change with its own test surface, and the current text is not a
regression — the `pgrep` walk ignored exit status too.

**Only the same-volume verification is removed.** The cross-volume branch copies
real bytes and can genuinely corrupt them; its `sha256` check and its cleanup
are untouched. The two branches were answering the same question with the same
answer when only one of them had a reason to ask.

## Verification

`./build.sh --test` on macOS 26 (Darwin 25.6.0), Apple silicon: `TESTS OK (9493
checks)`, clean. Two checks drive `KillProcessSupport.descendants(of:parents:)`
from an explicit parent table — not from this machine's process list — covering
breadth-first order, a leaf, an unrelated branch, a self-parenting row and a
two-node cycle. Four more pin `WhatsAppDownloadSupport.isSameVolume`: equal
identifiers held by two different objects, two different identifiers, either or
both identities unknown, and two identities that print the same string while
being unequal.

Both sets were confirmed to fail before they were trusted. Replacing the
visited-set guard with a parent comparison turned the `descendants` checks red,
and they went green on revert; while doing that I found the separate self-parent
filter I had first written was dead — the visited set already covers it — so it
is gone rather than sitting there untested. For the volume checks I compiled the
real `WhatsAppDownloadSupport.swift` with `isEqual` swapped back to
`String(describing:) ==` and drove the same four assertions: the printed-alike
pair failed and the other three passed, and all four passed again with `isEqual`
restored. That last one is the check that pins the fix.

**What was not compiled here.** `KillProcessService.swift` and
`WhatsAppDownloadOrganizer.swift` are outside the `--test` whitelist, so the test
binary never builds them; CI covers them on both runners. The full `./build.sh`
was not run.

## Anything else

Nothing was exercised against a live system. I did not kill a real process tree,
so the claim that `ps -eo pid,ppid` yields the same descendants `pgrep -P` did is
read off the two implementations rather than compared on a running Mac, and I
have no before/after timing — the stalling figure is inferred from fork/exec
cost, not profiled. I did not move a real WhatsApp download, within a volume or
across two, so the same-volume claim still rests on rename semantics rather than
on an observed transfer.

What the removed hash actually protected against is worth stating, because it
sets what is now unverified. On a genuine same-volume move it could not fail: the
rename touches no bytes and the destination path is empty before it. Its only
real job was covering a wrong same-volume answer — a `moveItem` that quietly fell
back to copying across devices. That is exactly the case `isEqual` now decides,
so the remaining risk is not a bad comparison but a correct one on identifiers
the filesystem itself reports wrongly: two distinct volumes handing out one
identity, or a path whose volume changes between the two `resourceValues` reads.
Neither is checked, here or anywhere else in the app.

I did not verify what `String(describing:)` printed on volumes other than this
machine's. Locally the identifier arrives as `NSData` and its description is its
bytes, so the old comparison agreed with `isEqual` on APFS here — which is the
point: that agreement was a property of one class on one machine, not of the API.

### DIFF
--- a/Sources/Vorssaint/Services/KillProcess/KillProcessService.swift
+++ b/Sources/Vorssaint/Services/KillProcess/KillProcessService.swift
@@ -293,9 +293,9 @@ final class KillProcessService: ObservableObject {
     /// the per-pid `pgrep` walk raced the same way, and `killBatch` still
     /// re-checks every target's identity before it signals anything.
     private static func descendants(of pid: pid_t) -> [pid_t] {
-        let result = Shell.run("/bin/ps", ["-eo", "pid,ppid"])
-        guard result.status == 0 else { return [] }
-        let parents: [(pid: pid_t, ppid: pid_t)] = result.output
+        let listing = Shell.run("/bin/ps", ["-eo", "pid,ppid"])
+        guard listing.status == 0 else { return [] }
+        let parents: [(pid: pid_t, ppid: pid_t)] = listing.output
             .split(separator: "\n").dropFirst().compactMap { line in
                 let columns = line.split(separator: " ", omittingEmptySubsequences: true)
                 guard columns.count >= 2, let child = pid_t(columns[0]),
--- a/Sources/Vorssaint/Services/KillProcess/KillProcessService.swift
+++ b/Sources/Vorssaint/Services/KillProcess/KillProcessService.swift
@@ -283,24 +283,26 @@ final class KillProcessService: ObservableObject {
         }
     }
 
-    /// BFS over `pgrep -P` to collect every descendant of `pid`. Capped well
-    /// above any real process tree, so a pathological environment can't spin
-    /// this forever.
+    /// Every descendant of `pid`, walked in memory from a single `ps` parent
+    /// table. The BFS used to run `pgrep -P` once per pid it had found, so
+    /// tearing down a tree of a few hundred processes forked that many short
+    /// lived helpers one after another - seconds of stalling while the list
+    /// had already dropped the rows.
+    ///
+    /// A grandchild forked between this snapshot and the signal is missed;
+    /// the per-pid `pgrep` walk raced the same way, and `killBatch` still
+    /// re-checks every target's identity before it signals anything.
     private static func descendants(of pid: pid_t) -> [pid_t] {
-        var result: [pid_t] = []
-        var frontier = [pid]
-        while !frontier.isEmpty, result.count < 4096 {
-            var next: [pid_t] = []
-            for parent in frontier {
-                let output = Shell.run("/usr/bin/pgrep", ["-P", String(parent)]).output
-                next.append(contentsOf: output.split(separator: "\n").compactMap {
-                    pid_t($0.trimmingCharacters(in: .whitespaces))
-                })
+        let result = Shell.run("/bin/ps", ["-eo", "pid,ppid"])
+        guard result.status == 0 else { return [] }
+        let parents: [(pid: pid_t, ppid: pid_t)] = result.output
+            .split(separator: "\n").dropFirst().compactMap { line in
+                let columns = line.split(separator: " ", omittingEmptySubsequences: true)
+                guard columns.count >= 2, let child = pid_t(columns[0]),
+                      let parent = pid_t(columns[1]) else { return nil }
+                return (child, parent)
             }
-            result.append(contentsOf: next)
-            frontier = next
-        }
-        return result
+        return KillProcessSupport.descendants(of: pid, parents: parents)
     }
 
     // MARK: - Restart
--- a/Sources/Vorssaint/Services/KillProcess/KillProcessSupport.swift
+++ b/Sources/Vorssaint/Services/KillProcess/KillProcessSupport.swift
@@ -25,6 +25,31 @@ enum KillProcessSupport {
         return ascending ? comparison == .orderedAscending : comparison == .orderedDescending
     }
 
+    /// Every descendant of `root` from a flat parent table, breadth first so
+    /// the caller can reverse it and kill deepest first. A pid is visited
+    /// once, so a self-parenting or circular row cannot loop, and the total
+    /// is capped well above any real process tree.
+    static func descendants(of root: pid_t, parents: [(pid: pid_t, ppid: pid_t)]) -> [pid_t] {
+        var children: [pid_t: [pid_t]] = [:]
+        for row in parents {
+            children[row.ppid, default: []].append(row.pid)
+        }
+        var result: [pid_t] = []
+        var seen: Set<pid_t> = [root]
+        var frontier = [root]
+        while !frontier.isEmpty, result.count < 4096 {
+            var next: [pid_t] = []
+            for parent in frontier {
+                for child in children[parent] ?? [] where seen.insert(child).inserted {
+                    next.append(child)
+                }
+            }
+            result.append(contentsOf: next)
+            frontier = next
+        }
+        return result
+    }
+
     static func normalizedStartDescription(_ value: String) -> String? {
         let normalized = value.split(whereSeparator: \.isWhitespace).joined(separator: " ")
         guard !normalized.isEmpty,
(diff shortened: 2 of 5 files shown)


PathGao:
Good part is the test: an explicit parent table instead of this machine's process list, mutation-checked, and the dead self-parent filter deleted rather than left sitting untested.

**The volume check is now a safety guard, not a hint.** `moveVerified` compares `String(describing:)` on two opaque `volumeIdentifier` objects. Before this PR a false "same volume" cost nothing — `moveItem` falls back to copy+delete across devices and the sha256 caught it. Now that branch copies real bytes unverified. `volumeIdentifier` is meant to be compared with `isEqual`, and the repo already has that tested in `CutPasteProgressSupport.isCrossVolume` (used by `FinderCutPaste.volumeIdentity`). I haven't checked what `description` actually prints for that type, and that string is now the whole guarantee.

`descendants` returns `[]` when `ps` fails, so `killTree` kills the parent alone and reports success. `snapshot()` in the same file returns `nil` for exactly this case. Not a regression — the old walk ignored status too — but `[pid_t]?` would let `killTree` bail instead.

Body nit: "destination comes from `uniqueDestination`" covers one of two callers; `replaceVerified` passes `existing`. Still safe there, staging empties the path first.

**Impact:** same-volume organizer moves lose their only corruption check — harmless if the volume compare is exact, an unverified copy if it isn't.
**Merge:** needs the volume comparison switched to `isEqual` first.

PathGao:
@PathGao The volume compare is switched to `isEqual` in `e8a4b16`.

The gap was that skipping the hash promoted that comparison from a hint to the only thing standing between a real cross-device copy and an unverified move, while it was still comparing `String(describing:)` on two opaque `volumeIdentifier` objects. It now goes through a pure `WhatsAppDownloadSupport.isSameVolume(source:destination:)` — `isEqual`, and false when either identity is unknown, so an unknown identity buys a verified copy rather than an unchecked rename. That fallback is the opposite of `CutPasteProgressSupport.isCrossVolume`'s on purpose: there an unknown identity only costs a hidden progress bar.

You were right that it needed checking rather than assuming. `.volumeIdentifierKey` has exactly two readers in the app: this one and `FinderCutPaste.volumeIdentity`, which already fed `isCrossVolume` and is untouched. Nothing else in the branch compares volumes. On this machine the identifier arrives as `NSData`, whose description is its bytes, so the old compare happened to agree with `isEqual` on APFS here — which is the argument for the change, not against it: that was a property of one class on one machine.

Four checks pin the helper, and the red-first was done on the failure mode that matters. I compiled the real `WhatsAppDownloadSupport.swift` with `isEqual` swapped back to `String(describing:) ==` and drove the same four assertions: the pair that prints alike but is unequal went red, the other three stayed green, and all four were green again with `isEqual` restored. `./build.sh --test`: `TESTS OK (9493 checks)`, up from 9489.

The body nit is fixed — `replaceVerified` passing `existing` is now named, along with why it is safe. Your `ps`-failure note I answered in the body rather than acted on: `[pid_t]?` is the better shape, but it changes what killing a tree does when the system is already in trouble, and you called it not a regression, so it did not seem worth widening this PR — say the word and it is a small follow-up.

Not verified: `KillProcessService.swift` and `WhatsAppDownloadOrganizer.swift` are outside the `--test` whitelist, so neither was compiled here; the full `./build.sh` was not run. No real move was performed, on one volume or across two. What remains unverified after the fix is not the comparison but its inputs — two distinct volumes reporting one identity, or a path whose volume changes between the two `resourceValues` reads. Neither is checked here or anywhere else in the app.

PathGao:
`isSameVolume` is the right shape — pure, `isEqual`, `nil` falls to the verified copy. The compile-with-`String(describing:)`-restored run is what actually pins it. I re-grepped `.volumeIdentifierKey` on `main`: two readers, this one and `FinderCutPaste.volumeIdentity`. Your claim holds.

Your "not compiled here" caveat is already answered on this PR — `Build & selftest` is green, so both `KillProcessService.swift` and `WhatsAppDownloadOrganizer.swift` did build. Only the local `--test` run skipped them.

On `[pid_t]?`: don't. It changes what killing a tree does when the system is already in trouble, which needs its own tests, and this PR is about removing known work. Separate PR.

One place the perf claim now overreaches: `currentStartDescription` (`KillProcessService.swift:399`) still forks `ps -p <pid>` once per target, from `killBatch`. It only fires on EPERM, so a tree of your own processes never reaches it — but "forked a helper per process" stays true of the escalation path.

**Impact:** none. The same-volume decision is stricter than before; every other path is unchanged.
**Merge:** ship it.

Issue #1182, opened 2026-08-31
Cleaner exceptions


slalomsquid:
### Problem or Motivation

When I use the cleaner function, it deletes my browser's cache, it would be nice to be able to add the folder as an exception.

### Proposed Solution

It would be nice to have a list of exceptions that files and folders can be added to.

### Feature Area

Other

PathGao:
The cleaner does keep an internal list of things it won't touch, or won't tick by default, but you can't add to it yourself. So this is a real gap.

Two things that may help you today. Anything the cleaner removes goes to your Trash rather than being deleted outright, so the cache is recoverable if you haven't emptied it since. And in a manual run you can untick the browser's row before you press clean.

One thing I need from you: did this happen when you ran the cleaner yourself, or on the automatic schedule? The scheduled run cleans whatever a scan ticks by default and never shows you the list, so if that's what took your cache, unticking isn't available to you and the fix looks different.

slalomsquid:
Thanks for the help, the cleaner put it in the bin when doing the manual run, and I just recovered it from the bin when I realised. For that reason I have the automatic cleaner turned off just in case.

PR #1248, opened 2026-09-01
fix(cleaner): keep the scheduled clean from asking for an administrator password

PathGao:
# Keep the scheduled clean from asking for a password

## Summary

A scheduled clean can raise an administrator password prompt at the hour the
user picked precisely so they would not have to be there.

The path: `CleanerPolicy.precheckLoginItems` is `true`, so orphaned startup
items start checked. `scanOrphanedLaunchPlists` scans
`~/Library/LaunchAgents`, `/Library/LaunchAgents` and `/Library/LaunchDaemons`,
and the last two are root owned. `CleanerScheduler.runAutomaticCleanup` takes
the scan's pre-checked selection exactly as it stands and calls
`cleanSelected()`. `trashItem` cannot move a root-owned file, so those items
fail into `stubborn`, and `stubborn` goes to `trashViaFinder`, which is Finder's
delete — the standard administrator prompt. On a first run it is preceded by the
automation consent dialog as well.

So somebody who set "daily at 03:00" gets a password box on an empty desk, from
a scheduler whose own class comment promises only that it cleans the safe groups
into the Trash.

`cleanSelected` now takes `escalate`. `CleanerScheduler` passes
`CleanerPolicy.escalateOnAutomaticRun`, which is `false`; `CleanerView` passes
`true`. When it is false, whatever the Trash move refused counts as failed and
stays exactly where it is. The scheduler's class comment now states this, since
it was previously describing a pass that could not ask for anything.

`CleanerPolicy` is where the decision lives because that file already holds
"every judgment call of the cleaner in one place", and this is one — putting a
bare `false` at the call site would leave the reasoning nowhere.

**What a scheduled clean does not do, and now says.** It cannot remove
anything `trashItem` refuses — see *What the scheduled pass gives up* below for
the actual extent of that. Those items are not lost — they stay in the next
manual scan — but the first version of this branch let them fall out of sight
entirely: `JunkCleaner.Phase` carried `.done(freed, failed)`, `CleanerScheduler`
matched `case let .done(freed, _)`, and the outcome notification only ever
reported bytes. A pass that deferred every find read exactly like a pass that
found nothing.

`finishRun` and `notifyIfWanted` now carry the count, and the notification adds
one sentence naming it: "1 item stayed in place." / "%d items stayed in place."
It states the fact and stops there, because the count is not one thing — see
*What the scheduled pass gives up*. It is appended to the freed-bytes line when both apply, and "Nothing to
clean" is now reserved for a pass that moved nothing **and** left nothing,
rather than being printed over a run that left items behind. This reuses the
scheduler's existing notification rather than adding a channel or a setting, and
it matches `WhatsAppDownloadScheduler.finishRun`, which already threads its own
`failed` count into its notification body.

**The sentence is a singular/plural pair.** The count reaches 1 whenever a
single find is refused, which is the ordinary case, so a plural-only string
would read "1 items stayed in place" in the situation this was written for.
Every other counted string in `Localization.swift` ships a
pair — `cutMovedSingular`/`cutMovedPluralFormat`,
`urlCleanerRulesCountSingular`/`…PluralFormat`, the four
`shelfTooltip*Singular`/`*Plural` — and the one plural-only string documents in
a comment why it can never be one. This one now has its pair in all thirteen
languages. `CleanerSupport.leftInPlaceSentence(count:_:)` picks between them, so
the notification and the interface cannot end up wording the same fact
differently; `CleanerSupport` is also inside the `--test` whitelist, which is
what lets the tests pin the choice rather than only the format string.

Russian has three plural forms, so a two-string pair cannot cover it. Its plural
is reworded to the count-agnostic form the other Russian plurals in this file
already use (`cutMovedPluralFormat` is "Перемещено элементов: %d"), which reads
at every count above one. The languages with no number agreement — Turkish,
Japanese, Korean and the three Chinese variants — carry the same text in both
slots, exactly as their `shelfTooltip*Singular`/`*Plural` entries already do.

**The count is recorded, not only announced.** `Notifier.post` drops the note
quietly when macOS notifications are denied, and the notify toggle is off for
anyone who turned it off. `finishRun` was persisting `freed` but not `failed`,
and `CleanerView.lastRunLine` reads only `cleanerLastAutoFreed`, so with the
notification gone a pass that deferred every find still showed nothing but "Last
automatic cleanup <time>" — the exact ambiguity the notification change removed.
`finishRun` now writes the count under `cleanerLastAutoLeft` beside the bytes it
already wrote, and `lastRunLine` appends the same sentence from the same
function. This is the shape `whatsAppDownloadsLastCleanupFailed` already has:
persisted by the manager, read back with `@AppStorage` by two views. The key is
per-machine state like `cleanerLastAutoFreed`, so it joins
`SettingsBackupSupport.machineStateKeys` and never travels in a settings export.
No new setting and no new surface: the last-run line was already on screen
whenever a scheduled pass had run.

The same sentence covers the other work the automatic pass has always swallowed
silently, because it arrives in the same counter: a Trash that would not empty,
and an item the `mayRemove` guard rejected. With the count now on both the
notification and the last-run line, there is no second silent case left in the
scheduled path — the earlier claim to that effect held only while the
notification actually arrived. The one other early return — the ten-minute reschedule
when the user is mid-session — defers the whole run and retries it, so it
reports normally when it does fire.

`cleanSelected` also has no default for `escalate` any more. With
`escalate: Bool = true`, an automatic caller added later inherits the prompt by
forgetting an argument; without it, both call sites have to say which they are.
`WhatsAppDownloadManager.cleanSelected(automatic:)` is the same shape in this
repo and already has no default.

### What the scheduled pass gives up

The earlier rounds of this description named one category — the root-owned
orphaned launch daemon — and that understated the change. `stubborn` has exactly
one source:

```swift
do { try fm.trashItem(at: item.url, resultingItemURL: nil); freed += item.size }
catch { stubborn.append(item) }
```

That loop runs over **every non-Trash category in the selection**, so
`if !escalate { failed += stubborn.count }` defers whatever the Trash move
refused, in whichever category it sat, not root-owned launch plists alone.

What the only non-escalating caller can actually reach is narrower than that,
and worth stating exactly. `CleanerScheduler` is the sole `escalate: false`
call site, and it cleans the scan's pre-checked selection as it stands
(`Item.include = recommended`). So the categories an automatic pass can defer
are the pre-checked ones: **login items, logs, developer junk, and the caches
`CleanerPolicy.precheckCacheEntry` pre-checks.** Leftovers
(`precheckLeftovers == false`) and device backups (`precheckDeviceBackups ==
false`) never enter an automatic selection at all, and the Trash category never
enters `stubborn`. The two categories a manual clean escalates most visibly are
therefore *not* the ones this defers.

**Measured, on one Mac** — macOS 26 (Darwin 25.6.0), Apple silicon, one user
account, admin. A read-only probe compiled from this branch's own
`JunkCleaner.swift` (an extension in the same file, so the private scanners are
reachable; `trashViaFinder` and `emptyTrash` replaced by stubs that trap) ran
the real scanners and reported, per item, `FileManager.isDeletableFile`,
`access(parent, W_OK | X_OK)`, `st_flags` immutability, owner uid and volume:

| | items | that the Trash move would refuse |
|---|---|---|
| whole scan | 139 | 2 |
| what an automatic pass would select | 20 (14 caches, 5 logs, 1 developer) | **0** |

The two refusals are `/Library/Logs/MCXTools.log` and
`/Library/Logs/LKDC-setup.log` — uid 0, in a uid-0 directory, reached because
`Logs` is a leftover root under `/Library` as well as under `~/Library`. Both
are `leftovers`, so they are `recommended: false` and an automatic pass never
selects them; a manual clean still hands them to Finder exactly as before.

This machine has no orphaned launch plist at all, so the motivating case did
not appear in the scan. The mechanism behind it is confirmed at the roots:
`/Library/LaunchAgents` (2 plists) and `/Library/LaunchDaemons` (1) are uid 0
and not writable by the user, while `~/Library/LaunchAgents` (8 plists) is the
user's and writable.

**This is an estimate, not an observation.** Nothing was moved: `trashItem` was
never called, so no real error code was collected and the "Not tested" note
below still stands on that point. `isDeletableFile` plus the parent-directory
and immutable-flag checks model the POSIX reasons the move fails; they do not
model TCC, and the probe carried Terminal's privacy grants rather than the
app's, so a path the app has no Full Disk Access for would throw where the probe
says deletable. Treat the zero in the table as "no POSIX reason found on this
machine", not as proof the count is always zero.

**The branch is not narrowed to `loginItems`, deliberately.** Restricting
`if !escalate` to that one category would send a refused cache or log item to
`trashViaFinder` instead — which is the administrator prompt on an empty desk
that this pull request exists to prevent. The rule that holds is the general
one: an unattended pass does not escalate, whatever the category.

**The sentence had to change with it.** The count it reports is `failed`, and
`failed` also carries `mayRemove` refusals and a Trash that would not empty.
Neither of those yields to an administrator prompt, so "clean by hand to remove
it" pointed at a dead end for them. The sentence now states what happened and
stops, and an assertion holds the remedy clause out.

### Trade-offs

**This is a behaviour change, and it is the point of the pull request.** An
automatic clean can no longer remove anything the Trash move refuses, in any
pre-checked category — a root-owned orphaned launch daemon is the motivating
case, not the boundary. The
alternative — keeping the removal and suppressing the prompt — does not exist,
because the prompt *is* how macOS grants the right; there is no unattended way
to do it that would be appropriate here even if there were.

The counter-argument was that a scheduled clean then accomplishes less than the
same selection would by hand, and that is real. The position taken here is that
a cleaner promising unattended operation must not be the reason an
authentication dialog appears on a locked or unattended Mac, and that a deferred
find is cheaper than a surprise password box — provided the deferral is visible,
which is what the notification change above is for. Doing less silently is the
part that was not defensible.

The Trash category is unaffected — it is `recommended: false`, so an automatic
pass never selects it and `emptyTrash()` is never reached from the scheduler.

Both strings ship in all thirteen supported languages, since a missing one is a
failed build here.

## Verification

`./build.sh --test` on macOS 26 (Darwin 25.6.0), Apple silicon: `TESTS OK (9523
checks)`, run once on the final tree.

`JunkCleaner.swift`, `CleanerScheduler.swift` and `CleanerView.swift` are outside
the `--test` whitelist, so the test binary does not compile them. I type-checked
the whole app module against the same target and SDK the build uses
(`swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk`, plus the
`VMStatisticsCompat` and `HIDEventSystem` include paths): no errors, no warnings.
The full `./build.sh` was **not** run here; CI covers it on both runners.

Twelve checks, each confirmed red before it was trusted. The seven marked **new**
were re-confirmed red on this tree; the rest were confirmed on the previous push
and their code is unchanged. The red runs used the same compiler invocation
`--test` uses, extracted from `build.sh`, so `./build.sh` itself ran once.

- `CleanerPolicy.escalateOnAutomaticRun` is `false` — red when the constant is
  flipped to `true`.
- A source-shape check on `CleanerScheduler.swift`: exactly one `cleanSelected(`
  call, and it passes the policy rather than a literal — red when the argument is
  dropped.
- A source-shape check that `.done(freed, failed)` reaches
  `finishRun(freed:failed:)` — red when the match goes back to `.done(freed, _)`.
- **New.** A source-shape check that the failed count reaches
  `CleanerSupport.leftInPlaceSentence` — red when the scheduler formats the
  plural string directly again.
- **New.** `CleanerSupport.leftInPlaceSentence(count: 1, .enUS)` reads "1 item
  stayed in place." — red when the function returns the plural string at every
  count, which is what this branch shipped before.
- **New.** `CleanerSupport.leftInPlaceSentence(count: 2, .enUS)` keeps the plural — red
  when the function returns the singular at every count.
- **New.** Neither en-US sentence contains "by hand" — red when the remedy
  clause goes back in, which fails it together with the two exact-text checks
  above (3 of 9523).
- `expectFormat(strings.cleanerAutoLeftFormat, ["d"])` for every `AppLanguage` —
  red when en-US loses its `%d`.
- **New.** `expectFormat(strings.cleanerAutoLeftSingular, ["d"])` for every
  `AppLanguage` — red when the German singular is rewritten to spell the count
  out.
- **New.** A source-shape check that `CleanerScheduler.finishRun` writes
  `cleanerLastAutoLeft` — red when the `defaults.set` is removed.
- **New.** A source-shape check that `CleanerView` reads `cleanerLastAutoLeft` and
  appends `leftInPlaceSentence` — red when the block is removed from
  `lastRunLine`.
- **New.** `SettingsBackupSupport.machineStateKeys` contains `cleanerLastAutoLeft` — red
  when the key is left out and the count would travel to another Mac.

Plus the same source-shape check for the dropped default: `JunkCleaner` declares
`cleanSelected(escalate: Bool)` with no default and `CleanerView` passes
`escalate: true` — red both when the default is restored and when the manual call
site passes `false`.

**The manual path is unchanged, and I checked rather than assumed it.**
`CleanerView` was the only caller of `cleanSelected()` and now passes
`escalate: true`, which is exactly what the removed default resolved to; the
`escalate == true` branch in `cleanSelected` is the original `if !stubborn`
block unmodified. `finishRun` and `notifyIfWanted` are `private` to
`CleanerScheduler` and have no callers outside it, so nothing in the notification
change is reachable from a manual clean. `CleanerView`'s own `.done(freed,
failed)` handling and its `uninstallerSomeFailed` line are untouched.

**I never triggered the prompt.** I did not arm a schedule, wait for a fire time,
or watch a password dialog appear, and I did not see the new notification on
screen — no GUI testing was done. I have not seen the new last-run line either:
the singular and plural sentences and the persisted count are verified as values
and as source shape, not as pixels, and the thirteen translations are not
reviewed by a native speaker. The bug is derived by reading the chain
(`precheckLoginItems` → the `/Library` scan roots → `EPERM` → `stubborn` →
`trashViaFinder` → `AppleScriptRunner.consentToAutomate`), and I have not
moved a single real file with `trashItem`, so no error code from a refusal was
observed and `EPERM` rather than some other error is still an assumption. What
*was* measured is in *What the scheduled pass gives up*: a read-only estimate of
which scanned items the move would refuse, on one machine, which found none in
the selection an automatic pass would make and no orphaned launch plist on the
machine at all. The notification body
was not measured against Notification Centre's two-line truncation in the longer
languages.

## Anything else

No issue tracks this, so there is no `Refs` line. I searched the open and closed
issues for the scheduled clean, the administrator prompt, the automatic-cleanup
notification and the `bug` label, and read the three summary issues. The nearest
hits are the closed "Background File Cleaner" feature request the scheduled clean
came from, and the open "Cleaner exceptions" request; neither reports this
behaviour, and this fixes neither. Say the word if a reference should be added
anyway.

The `blocked` label is the maintainer's to clear: the question this was held on
has been answered, and the answer is implemented here.


PathGao:
Chain checks out against `main`: `scanOrphanedLaunchPlists` walks `/Library/LaunchAgents` and `/Library/LaunchDaemons` with `precheckLoginItems = true`, and `cleanSelected` has exactly two callers — `CleanerView` and the scheduler. No third automatic entry point.

**Deferred items are invisible.** `CleanerScheduler` matches `case let .done(freed, _)` and `finishRun` takes only `freed`, so the failed count never reaches `notifyIfWanted`. The body says they "stay checked in the next manual scan, which is the run where somebody is present" — but nothing tells anyone to open one. Threading `failed` through `finishRun` is the whole fix.

**`escalate: Bool = true` re-introduces this by omission.** `WhatsAppDownloadManager.cleanSelected(automatic:)` is the same shape here with no default, so both its call sites must decide. Drop the default, pass `escalate: true` at `CleanerView.swift`, and "added an automatic caller, forgot the argument" stops being possible. The test still earns the other half — that the scheduler passes the policy, not a literal.

**Impact:** manual path unchanged; a scheduled clean now leaves root-owned finds in place and says nothing.
**Merge:** hold for the maintainer's call on whether a scheduled clean may do less than the same selection by hand.

PathGao:
@PathGao The decision this was held on has been made by the maintainer: a scheduled clean **may** do less than the same selection run by hand, so the no-escalation behaviour stays as it is. What was not acceptable is that it did less without saying so. That is what `42e788a2` fixes, and it is the whole of the new commit.

The gap was exactly where you put it. `JunkCleaner.Phase` already carried `.done(freed, failed)`, but `CleanerScheduler` matched `case let .done(freed, _)` and `finishRun` took only `freed`, so the count died one line before the notification. A pass that deferred every find was indistinguishable from a pass that found nothing. `finishRun` and `notifyIfWanted` now carry `failed`, and the notification appends one sentence: "%d items stayed in place; clean by hand to remove them." I reused the scheduler's existing notification rather than adding a channel or a setting — `WhatsAppDownloadScheduler.finishRun` already threads its own `failed` into its body, so this is the shape the app has. One correctness detail fell out of it: `cleanerNothingFound` is now reserved for a pass that moved nothing **and** left nothing, instead of being printed over a run that had left items behind.

Same shape elsewhere, checked: `failed` in `cleanSelected` also collects a Trash that would not empty and items the `mayRemove` guard rejected. Those were equally invisible on the scheduled path before, and they route through the same sentence now, so there is no second silent case left. The only other early exit in `runAutomaticCleanup` is the ten-minute reschedule when the user is mid-session, which defers the whole run and reports normally when it fires. On your second point, `escalate` has lost its default: `JunkCleaner` declares `cleanSelected(escalate: Bool)`, `CleanerView` passes `true` (what the default resolved to), and forgetting the argument is now a compile error, matching `WhatsAppDownloadManager.cleanSelected(automatic:)`.

`./build.sh --test` once on the final tree: `TESTS OK (9504 checks)`. Five new checks, each seen red before I trusted it — the `.done(freed, failed)` → `finishRun(freed:failed:)` thread (red when the match drops `failed`), the count reaching `cleanerAutoLeftFormat` (red when the sentence loses its argument), the dropped default plus the manual call site (red both when the default returns and when `CleanerView` passes `false`), and `expectFormat(strings.cleanerAutoLeftFormat, ["d"])` across every `AppLanguage` (red when en-US loses its `%d`). The string ships in all thirteen languages. `CleanerScheduler.swift`, `JunkCleaner.swift` and `CleanerView.swift` are outside the `--test` whitelist, so I type-checked the whole module with the build's own target and SDK flags instead: clean, no warnings. I did not run the full `./build.sh` — CI covers it on both runners.

Manual path verified, not assumed: `CleanerView` was `cleanSelected()`'s only caller, the `escalate == true` branch is the original `if !stubborn` block unmodified, and `finishRun`/`notifyIfWanted` are private to the scheduler with no callers outside it. What I did not verify: anything on screen. No schedule was armed, no notification was seen, and I have not checked the new body against Notification Centre's two-line truncation in the longer languages.

The `blocked` label is yours to clear — I have not touched labels.

PathGao:
Decision noted. The thread lands where it needed to — `.done(freed, failed)` → `finishRun` → `notifyIfWanted`, no new channel, and `cleanerNothingFound` correctly narrowed. Two things.

**`cleanerAutoLeftFormat` is plural-only.** Every other count string in `Localization.swift` ships a pair: `cutMovedSingular`/`cutMovedPluralFormat`, `urlCleanerRulesCountSingular`/`…PluralFormat`, the four `shelfTooltip*Singular`/`*Plural`. The only plural-only one documents why in a comment (`shelfTooltipItemsFormat`, "a pile always has 2+"). One deferred launch daemon is the likely count here, so en-US reads "1 items stayed in place". Cheaper than a 13-string pair: reword so `%d` also reads at 1.

**"No second silent case left" holds only while the notify toggle is on.** `finishRun` persists `freed`, not `failed`, and `CleanerView.lastRunLine` reads only `cleanerLastAutoFreed` — so a pass that deferred everything still shows "Last automatic cleanup <time>", the exact ambiguity this removed from the notification. `Notifier.post` also drops quietly when notifications are denied, which that view already tracks. One more `defaults.set`, or narrow the claim.

**Impact:** wrong plural in all thirteen languages at the most likely count; with notifications off or denied, a deferred pass still looks like a clean one.
**Merge:** needs the singular case first.

PathGao:
@PathGao Both notes are closed in `ea75d754`.

**The singular.** `cleanerAutoLeftSingular` joins `cleanerAutoLeftFormat` in all thirteen languages, and `CleanerSupport.leftInPlaceSentence(count:_:)` picks between them. It lives in `CleanerSupport` rather than at the two call sites because `CleanerView` and `CleanerScheduler` both need the sentence and `CleanerSupport` is inside the `--test` whitelist, so the choice is pinned by a value assertion instead of only by source shape. Two details fell out: Russian has three plural forms, so its plural is reworded to the count-agnostic shape its own `cutMovedPluralFormat` ("Перемещено элементов: %d") already uses — a pair alone cannot cover it; and Turkish, Japanese, Korean and the three Chinese variants carry the same text in both slots, matching their existing `shelfTooltip*Singular`/`*Plural` entries.

**The silent path.** You were right that the claim held only while the notification arrived. `finishRun` now writes the count under `cleanerLastAutoLeft` next to the `freed` it already wrote, and `CleanerView.lastRunLine` appends the same sentence from the same function, so the deferred count survives both the notify toggle being off and `Notifier.post` dropping quietly when notifications are denied. That is one `defaults.set` and one `if` — no new setting, no new surface. The shape is `whatsAppDownloadsLastCleanupFailed`, which is persisted by the manager and read back with `@AppStorage` by two views. The key is per-machine like `cleanerLastAutoFreed`, so it also joins `SettingsBackupSupport.machineStateKeys` and never travels in a settings export.

Same shape elsewhere, checked: `cleanerAutoLeftFormat` had exactly two readers after this change and both now go through `leftInPlaceSentence` — the notification and the last-run line. The other counted strings in `Localization.swift` already ship pairs and are already selected at their call sites (`CutFeedbackView`, `URLCleanerSettings`, `ShelfSupport`); I did not touch them. The only remaining plural-only counted string is `shelfTooltipItemsFormat`, which documents in its comment why it can never be one.

`./build.sh --test` once on the final tree: `TESTS OK (9522 checks)`. Seven new or changed checks, each seen red before I trusted it — the singular at count 1 (red when the function returns the plural at every count, which is what this branch shipped), the plural at count 2 (red when it returns the singular at every count), `expectFormat(strings.cleanerAutoLeftSingular, ["d"])` across every `AppLanguage` (red when the German singular spells the count out), the scheduler reaching `leftInPlaceSentence` (red when it formats the plural string directly again), the `defaults.set` of `cleanerLastAutoLeft` (red when removed), `CleanerView` reading it and appending the sentence (red when the block goes), and the key being in `machineStateKeys` (red when it would travel to another Mac). The red runs used `--test`'s own compiler invocation extracted from `build.sh`, so `./build.sh` itself ran once. `CleanerScheduler.swift` and `CleanerView.swift` are outside the whitelist, so I type-checked the whole module with the build's target and SDK flags again: clean, no warnings.

What I did not verify: anything on screen. I have not seen the new last-run line or the notification, the thirteen translations are not reviewed by a native speaker, and I still have not checked the body against Notification Centre's two-line truncation in the longer languages.

PathGao:
The pair plus `leftInPlaceSentence` is the right consolidation — one sentence, two readers, and the count-agnostic Russian plural is the correct escape from a two-string pair.

One gap. `cleanerLastAutoLeft` is written only by `CleanerScheduler.finishRun`, so nothing clears it when the user does the clean it asks for. Pass defers 2 daemons at 03:00, user cleans by hand at 10:00 and authenticates — the last-run line still reads "2 items stayed in place; clean by hand to remove them" until the next scheduled pass, a week out on a weekly schedule. That is not the `whatsAppDownloadsLastCleanupFailed` shape you cited: `recordCleanup` runs on the manual and the automatic path, this one only on the automatic.

Clear it at the manual call site, not on phase — `CleanerView` observes `JunkCleaner.shared`, so an `onChange(of: cleaner.phase)` would also fire when a scheduled pass ends with the view open and clobber the count `finishRun` just wrote.

```swift
Button(...) { lastAutoLeft = 0; cleaner.cleanSelected(escalate: true) }
```

**Impact:** up to a week of the interface asking for a clean the user already did.
**Merge:** needs the manual call site to clear the count first.

PathGao:
@PathGao The difference set is bigger than this description said, and I measured
it instead of asserting it. Pushed as `de5e8031`.

**What `stubborn` actually is.** One source, `catch { stubborn.append(item) }`
on `fm.trashItem`, and that loop runs over **every non-Trash category in the
selection**. So `if !escalate { failed += stubborn.count }` defers whatever the
Trash move refused wherever it sat, not the root-owned launch plists this
description kept naming.

**What the one non-escalating caller can reach is narrower.**
`CleanerScheduler` is the only `escalate: false` call site and it cleans the
scan's pre-checked selection as it stands (`Item.include = recommended`). So the
reachable categories are the pre-checked ones — **login items, logs, developer
junk, and the caches `precheckCacheEntry` pre-checks**. Leftovers
(`precheckLeftovers == false`) and device backups never enter an automatic
selection, and Trash never enters `stubborn`.

**Measured, one Mac, nothing moved.** A read-only probe compiled from this
branch's own `JunkCleaner.swift` — an extension in the same file so the private
scanners are reachable, with `trashViaFinder` and `emptyTrash` replaced by stubs
that trap — ran the real scanners and reported `isDeletableFile`,
`access(parent, W_OK | X_OK)`, `st_flags`, owner uid and volume per item:

| | items | the Trash move would refuse |
|---|---|---|
| whole scan | 139 | 2 |
| what an automatic pass would select | 20 (14 caches, 5 logs, 1 developer) | **0** |

The two are `/Library/Logs/MCXTools.log` and `/Library/Logs/LKDC-setup.log`, uid
0 in a uid-0 directory, reached because `Logs` is a leftover root under
`/Library` as well as `~/Library`. Both are `leftovers`, so `recommended: false`
— an automatic pass never selects them and a manual clean still hands them to
Finder exactly as before. This machine has no orphaned launch plist at all, so
the motivating case did not appear; the mechanism is confirmed at the roots
instead — `/Library/LaunchAgents` (2 plists) and `/Library/LaunchDaemons` (1)
are uid 0 and not writable, `~/Library/LaunchAgents` (8) is the user's.

**This is an estimate.** `trashItem` was never called, so your `EPERM` note is
still open — no real error code was collected. `isDeletableFile` plus the
parent-directory and immutable-flag checks model the POSIX reasons only; they do
not model TCC, and the probe carried Terminal's grants rather than the app's, so
a path the app has no Full Disk Access for would throw where the probe says
deletable. The zero means "no POSIX reason found here", not "always zero".

**I did not narrow the branch to `loginItems`.** Doing that would send a refused
cache or log item to `trashViaFinder` instead, which is the administrator prompt
on an empty desk this PR exists to prevent. The general rule is the one that
holds. The description now carries the extent instead.

**The sentence changed with it.** The count it reports is `failed`, which also
carries `mayRemove` refusals and a Trash that would not empty. Neither yields to
an administrator prompt, so "clean by hand to remove it" pointed those at a dead
end. It is now "%d item stayed in place." in all thirteen languages — the clause
is gone, the count stays. New assertion `the deferred-count sentence names no
remedy a manual clean cannot deliver`; putting the clause back fails it with the
two exact-text checks (3 of 9523). The doc comments on
`escalateOnAutomaticRun`, `leftInPlaceSentence` and `notifyIfWanted` no longer
describe the deferral as root-owned launch plists.

```sh
./build.sh --test     # TESTS OK (9523 checks)
```

Neither of the two input-method-dependent assertions came up red on this run.

**Where to look next:** whether `failed` should be one counter at all. It is now
three different outcomes under one sentence — a Trash move the app may not make,
a `mayRemove` refusal, and an identity that changed between scan and clean — and
the third is the one I would expect most often on a real machine, since caches
and logs are rewritten constantly. Splitting it is a bigger change than this PR,
but the sentence is only honest because it stopped trying to say which.

PathGao:
Dropping the remedy clause is the right call: the count covers three outcomes and only one of them yields to a prompt, so a sentence that names no remedy is the only honest one. The measured table settles the extent, and labelling it an estimate rather than an observation is what makes it usable.

You are right that my clearing point dies with that clause. "Last automatic cleanup <time>. 2 items stayed in place." is a report about that pass and stays true after a manual clean. Withdrawn.

One new thing, and it is not in your diff — `main` moved under this branch. `cleanSelected` now builds an installed-apps oracle, and `MetricsTests` pins it by slicing the function out by name:

```swift
sourceBody(of: junkCleanerSource, from: "func cleanSelected()", to: "private static func mayRemove")
```

`sourceBody` returns `""` when the opening marker is missing, so `cleanSelected(escalate:)` empties that slice and the assertion under it goes red on merge. Not in your 9523 — your base predates it. The marker wants to be `"func cleanSelected("`. The same rebase decides whether your `!escalate` branch lands cleanly beside main's new `stubborn.filter { Self.mayRemove($0, installed: installed) }`.

**Impact:** `./build.sh --test` red on main after merge; no runtime behaviour.
**Merge:** needs the rebase and that marker first.

PathGao:
@PathGao Rebased onto main. The one conflict was JunkCleaner.swift where #1246 changed the stubborn filter to mayRemove($0, installed:); that form is kept inside the escalate branch. A separate commit changes the #1246 marker in MetricsTests from "func cleanSelected()" to "func cleanSelected(" so the body pin still matches the escalate: signature. Full ./build.sh exit 0, TESTS OK (9634 checks).

PathGao:
Rebase landed where it needed to. Verified against `main` rather than taken on trust:

- `sourceBody` takes the first match of the opening marker, and `func cleanSelected(` has no earlier occurrence in `JunkCleaner.swift` — the new doc comment does not contain it — so the #1246 pin slices the same region and its three `contains` still sit inside it.
- `JunkCleaner.cleanSelected` has exactly two callers on main, `CleanerView` and `CleanerScheduler`. No third, so the escalate decision is exhaustive.
- `cleanerLastAutoFreed` and `cleanerScheduleLastFormat` have one reader each, `CleanerView.lastRunLine`. The deferred sentence has no sibling surface it missed.

Keeping `mayRemove($0, installed:)` inside the escalate branch is the right resolution — that re-check only ever mattered ahead of the Finder hand-off.

Nothing left to flag. Side note: the bot's `Refs #1182` hint is wrong, that one asks for a user-editable exception list.

I did not build. `TESTS OK (9634)` and the type-check are your numbers, and neither of us has seen the sentence on screen.

**Impact:** none beyond the branch's own behaviour change, already described.
**Merge:** ship it.

PathGao:

PathGao:
Reopened as #1298.

PR #1250, opened 2026-09-01
feat(monitor): add SSD temperature and all-sensors view

maxiaoleking:
## What

Extend the System Monitor temperature section to discover SSD (Ts*/TN*) and all remaining SMC temperature sensors. Add an SSD temperature cell and a collapsible "All sensors" list to the System panel.

## Why

The current implementation only reads CPU (Tp/Te), GPU (Tg), and battery (TB) temperature keys. Many Macs expose additional sensors (SSD, platform controller, etc.) that are useful for thermal monitoring but invisible in the UI.

## Changes

- **SystemSnapshot**: add `ssdTemperature` and `allTemperatures` fields
- **SystemMonitorPanelNeeds**: add `ssdTemperature` / `allTemperatures` visibility flags
- **SamplingPlan**: add `needSSDTemperature` / `needAllTemperatures`
- **prepareSensorsIfNeeded()**: extend SMC key filter to include `Ts*` and `TN*` prefixes
- **refresh()**: read SSD temperature and collect all sensor readings when enabled
- **SystemSection**: add SSD temperature cell alongside CPU/GPU/Battery, plus a DisclosureGroup listing all discovered sensors sorted by temperature
- **Localization**: add `ssdLabel` and `allSensorsLabel` in all 13 languages

## Verification

- `./build.sh` compiles cleanly (release, no warnings)
- `./build/Vorssaint --selftest` → SELFTEST OK
- 14 files changed, 86 insertions, 2 deletions

## Notes

- SSD sensors show only when the hardware exposes them (nil = hidden, matching existing CPU/GPU/Battery behavior)
- The all-sensors view is gated behind a DisclosureGroup to avoid clutter
- No new dependencies; no changes to SMCClient, FanControl, or build system

### DIFF
--- a/Sources/Vorssaint/Core/Localization.swift
+++ b/Sources/Vorssaint/Core/Localization.swift
@@ -181,6 +181,8 @@ struct Strings {
     let cpuLabel: String
     let gpuLabel: String
     let batteryLabel: String
+    let ssdLabel: String
+    let allSensorsLabel: String
     let usageSection: String
     let memorySection: String
     let memoryPressure: String
@@ -1204,6 +1206,8 @@ extension Strings {
         cpuLabel: "CPU",
         gpuLabel: "GPU",
         batteryLabel: "Bateria",
+        ssdLabel: "SSD",
+        allSensorsLabel: "Todos os sensores",
         usageSection: "Uso de hardware",
         memorySection: "Memória",
         memoryPressure: "Pressão",
@@ -2190,6 +2194,8 @@ extension Strings {
         cpuLabel: "CPU",
         gpuLabel: "GPU",
         batteryLabel: "Battery",
+        ssdLabel: "SSD",
+        allSensorsLabel: "All sensors",
         usageSection: "Hardware usage",
         memorySection: "Memory",
         memoryPressure: "Pressure",
--- a/Sources/Vorssaint/Core/Localizations/Strings+ChineseSimplified.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+ChineseSimplified.swift
@@ -76,6 +76,8 @@ extension Strings {
         cpuLabel: "CPU",
         gpuLabel: "GPU",
         batteryLabel: "电池",
+        ssdLabel: "SSD",
+        allSensorsLabel: "所有传感器",
         usageSection: "硬件占用",
         memorySection: "内存",
         memoryPressure: "压力",
--- a/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalHK.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalHK.swift
@@ -76,6 +76,8 @@ extension Strings {
         cpuLabel: "CPU",
         gpuLabel: "GPU",
         batteryLabel: "電池",
+        ssdLabel: "SSD",
+        allSensorsLabel: "所有感應器",
         usageSection: "硬件使用量",
         memorySection: "記憶體",
         memoryPressure: "壓力",
--- a/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalTW.swift
+++ b/Sources/Vorssaint/Core/Localizations/Strings+ChineseTraditionalTW.swift
@@ -76,6 +76,8 @@ extension Strings {
         cpuLabel: "CPU",
         gpuLabel: "GPU",
         batteryLabel: "電池",
+        ssdLabel: "SSD",
+        allSensorsLabel: "所有感測器",
         usageSection: "硬體使用率",
         memorySection: "記憶體",
         memoryPressure: "壓力",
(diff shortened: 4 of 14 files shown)


PathGao:
The nil-hides-the-cell convention is followed and the `monitorUnavailable` guard was updated along with it. Three things:

**No carry-over.** The two new reads are the only temperature blocks without an `else`. CPU/GPU/battery each fall back to their cache when `take(.temperature)` strides past; these reset to nil / `[]`:

```swift
} else { next.ssdTemperature = self.ssdTemperatureCache?.value }
```

Any tick that skips temperature but still publishes (CPU usage samples every tick) blinks the SSD row and the whole All-sensors group out and back.

**`otherTempKeys` is empty by construction.** The `client.keys` predicate admits only CPU / `Tg` / `^TB[0-9]T$` / `Ts` / `TN`, and `knownNames` covers all five, so the difference is always empty. Body says "all remaining SMC temperature sensors (platform controller, etc.)"; what ships is CPU+GPU+battery+SSD. `name.hasPrefix("T")` in that predicate makes the claim true.

**Dead flags.** Nothing builds a `SystemMonitorPanelNeeds` with `ssdTemperature` / `allTemperatures` — MetricDetailView and MenuPanelView are the only producers — so `panelTemps` alone drives both plan fields. Drop the two fields.

No hardware here, so I can't confirm `Ts*`/`TN*` is what these Macs actually expose for SSD.

**Impact:** SSD row and sensor list flicker while the System panel is open. Nothing existing regresses.
**Merge:** needs the cache carry-over on both new reads first.

PathGao:
Updating the `monitorUnavailable` guard to include `ssdTemperature` is the sibling site most PRs like this miss — good.

**The all-sensors list can't contain anything new.** `prepareSensorsIfNeeded()` builds `all` from the same predicate you then partition it with, so `knownNames` (cpu ∪ `Tg` ∪ `TB` ∪ `Ts`/`TN`) covers every member of `all` by construction and `otherTempKeys` is always empty. The list ends up being exactly the four cells above it — the "platform controller, etc." keys the body mentions are never fetched. One line, without disturbing the grouping:

```swift
otherTempKeys = client.keys { $0.hasPrefix("T") && !knownNames.contains($0) }
```

Don't widen `all` itself — `batteryKeys` filters it with `hasPrefix("TB")` and would start swallowing keys that aren't `TB[0-9]T`.

**"Gated behind a DisclosureGroup" is UI-only.** `plan.needAllTemperatures = panelTemps` is true whenever the temperature section is visible, collapsed or not, so every user pays the sweep each temperature tick — and `allKeys` re-reads the cpu/gpu/battery/ssd keys that were just read for the cells. Gate the plan on expansion, or reuse the readings.

**Third copy of the plausibility window.** `> 1 && < 125` already lives in `maxTemperature(of:)`; the new `.filter` restates it. Put it in `temperatureReadings(of:)` once.

I checked the grouping against `main`, but I can't confirm `Ts*`/`TN*` are SSD sensors on real hardware.

**Impact:** the headline "all sensors" view shows only the four sensors already on screen; extra SMC reads per tick while the temperature section is open.
**Merge:** needs `otherTempKeys` to actually reach keys outside `all` first.

PathGao:
Sending this back: the headline feature does not exist as written.

`all` is built from `client.keys { isCPUTemperatureKey || hasPrefix("Tg") || ^TB[0-9]T$ || hasPrefix("Ts") || hasPrefix("TN") }`, then `knownNames` is the union of the cpu, gpu, battery and ssd partitions of that same set, and `otherTempKeys = all.filter { !knownNames.contains($0.name) }`. Every name `all` admits belongs to one of the four partitions, so `otherTempKeys` is empty by construction and "All sensors" can only ever show the four cells above it. Either widen `all` (any `T…` key, then classify) or drop the section.

Two smaller things: the two new reads lack the `else { cache }` branch the CPU/GPU/battery reads have, so the row blinks on ticks that do not sample temperature; and the description needs the Mac model and macOS build, plus how you established that `Ts*` / `TN*` are the SSD keys on that chip (`./build/Vorssaint --sensors` output would settle it).

vorssaint:
closing this one.
the all sensors list can only ever hold the cpu, gpu, battery and ssd sensors the cells above are already built from, because nothing else is ever picked up, so it never shows a new sensor.
a list of raw sensor codes is also not something the panel should show, since that kind of dump belongs to the sensors diagnostic on the command line.
the list re-reads every temperature sensor on each tick while the panel is open, collapsed or not, and the monitor is built to touch as few sensors as possible.
the ssd cell treats the ts and tn sensors as the drive, but where those prefixes are documented they sit on the palm rest and the logic board, so the cell can show another part's temperature.
the cell also ignores the disk family switch in the hub and has no hold over a bad read like the other cells, so it keeps sampling with disk monitoring off and can blink out on one failed read.
main has since moved the battery temperature into the power card, so a drive temperature belongs with the disk card rather than here.
a single ssd cell on the disk card, gated like the others and backed by a real reading of the keys on an actual mac, could still come back as a fresh pull request

Issue #1191, opened 2026-08-31
Add priority-based audio input and output selection


zamai:
## Problem

Vorssaint can select the current system output and remember one preferred microphone, but it cannot define an ordered list of preferred input and output devices. When devices connect or disconnect, there is no way to automatically select the highest-priority device that is currently available.

Related discussions cover parts of this problem:

- #93 added shortcut-based output cycling.
- #193 covers remembering a preferred microphone.
- #1003 discusses automatically selecting external-display audio when HDMI connects.

## Proposal

Add ordered priority lists for audio inputs and outputs, with automatic fallback to the next available device.

I currently use [AudioPriorityBar](https://github.com/tobi/AudioPriorityBar/) for this, and it works great. Since I also use Vorssaint—and it already has the Volume Mixer, Output switcher, and microphone selection—it feels like a natural home for this functionality.

I plan to contribute a PR for this soon. I am opening this issue first so the scope and behavior can be discussed before implementation.

PathGao:
Asking before writing the code is the right call here — the input and output halves are not symmetric.

Input already has the machinery. `MixerRoutingSupport.resolveInputDevice(preferredUID:availableUIDs:currentUID:)` is pure and takes one UID plus the set of available ones. Widen that to an ordered `[String]` and return the first available, and most of the input side is done — `AudioInputDeviceManager` re-runs it on every CoreAudio device notification, so connect/disconnect is already covered.

Output has nothing to widen. `AppVolumeMixer` tracks `currentOutputDeviceUID` and cycles via `nextSelectedOutputDeviceUID`; there is no persisted preference and no resolver. That half is new code, not a generalization.

Two things worth settling before you build, both for the maintainer to answer:

1. Storage. A new ordered-list key sitting next to `DefaultsKey.preferredInputDevice` puts the same fact in two places that can disagree. Migrating the existing key into a one-element list avoids that — acceptable, or should the single-mic setting stay as it is?
2. `preferredUnavailable` today means "your mic is gone". With a list it means "fell back to #3", which is not an error state. The mixer UI reads that flag.

Read from main only, nothing run.

PR #1253, opened 2026-09-01
feat(audio): add device priority lists

zamai:
## Summary

Adds the opt-in Audio device priority feature discussed in #1191. It uses ordered lists of discovered system outputs and microphones, retains disconnected devices by their last-known names, and leaves existing users unchanged until they install it. Both directions are enabled on first install; explicit off choices persist.

- A settled device connect or disconnect event applies the highest available device in each enabled list. Changing a list order or turning a priority switch back on also applies its highest available device after a short coalescing delay.
- A manual output or microphone choice remains active until the next inventory change or an explicit priority edit or enable action. Ordinary default-device changes and startup do not enforce the list or rewrite its order.
- Automatic CoreAudio writes run on the existing HAL queues. Refreshes publish the actual system defaults, and automatic output changes preserve explicit per-app routes.
- While microphone priority is enabled, the picker and list reflect the actual current input; the existing single preferred microphone remains dormant. A priority-selected microphone stays selected when Vorssaint quits and relaunches. When priority is disabled, the existing preferred-microphone behavior resumes.

The microphone list has its own preference key. The existing `preferredInputDevice` key remains unchanged, so installing or removing this feature does not migrate that setting. The feature adds no permissions, audio capture, network access, or runtime dependency. Settings backups and supported localizations include its controls.

## Verification

On an Apple M4 Pro Mac running macOS 26.6.2:

- `swift build` passed after merging current `main`.
- `./build.sh --test` passed all 60,005 checks, including the focused audio-priority suite.
- `./build.sh` and `./build.sh --dev` produced signed bundles from the merged branch; both executables passed `--selftest`.
- In a live developer-app check on this Mac before merging current `main`, reordering output and microphone lists and re-enabling each switch selected the top connected device. Manual selections stayed active without changing list order. A priority-selected microphone remained the macOS input through quit and relaunch. The original devices and developer settings were restored afterward. The merged branch was rebuilt and fully tested afterward.

Physical hot-plug combinations and Intel Macs have not been tested.

## Related issues

Refs #1191


PathGao:
The `processMonitoringEnabled` restart gate is right — priority-only genuinely avoids the process taps.

**1. The switcher shortcut rewrites the priority list.** `promoteOutputDevice` hangs off `setDefaultOutputDeviceUID(.manualUniversalSelection)`, but three call sites reach that node, not one: `MixerSection`, `switchToNextSoundOutput` (the output switcher shortcut) and a Command Bar row. Cycling outputs permutes the stored order into recency order and the dragged list is gone. Body says "manual device selection"; a cycle isn't one.

**2. A write that reports success but doesn't stick retries forever.** `setPriorityOutputDeviceUID` / `setCurrentInputDeviceUID` trust `noErr` and set the current UID optimistically. If the HAL accepts and doesn't apply, the listener publishes the old UID back and enforcement writes again — every 0.15s, unbounded. `isEnforcing` only guards synchronous reentry. Needs a read-back or a failed-target memory.

**3.** Priority input writes set `inputDeviceBeforeOverride`, so quit reverts the mic via `restoreOriginalInputDevice()`. Output has no matching revert; not in the body.

Diff truncated at 80 KB here — the list/drag view and tests are unreviewed.

**Impact:** priority users who also use the output switcher or Command Bar lose their configured order.
**Merge:** needs the promote hook narrowed to the picker call site first.

ref #1191

PathGao:
Sending this back; the 2026-09-01 12:16 ask was not addressed by the 12:51 push, which touched only `Defaults.swift` and the tests.

1. `AudioPriorityService.shared.promoteOutputDevice(device.uid)` still sits inside the shared `setDefaultOutputDeviceUID(source: .manualUniversalSelection)`. All three callers of `setUniversalOutputDeviceUID` on main reach it — the Settings picker, `switchToNextSoundOutput` (the output-switcher shortcut) and the Command Bar entry — so pressing the switch-output shortcut rewrites the order the user dragged into recency order. Promote only from the picker.
2. `enforce()` trusts `noErr`, sets `currentOutputDeviceUID` optimistically, and when the HAL accepts the write without applying it the sweep republishes the real default, which re-triggers `scheduleEnforcement()` every `eventEnforcementDelay` (0.15 s). Read the default back after the write, and remember a target that failed so it is not retried on every tick.

Then resolve the conflict. The storage question on #1191 (fold the preferred microphone into the list or keep two keys) is the maintainer's to answer; the branch picked one side unilaterally, and that is worth saying in the body.

zamai:
@PathGao Addressed the requested blockers in `c757984`:

1. Output priority promotion now lives only at the Settings/Menu panel Output picker call site. `setUniversalOutputDeviceUID` no longer promotes, so output-switcher cycling and Command Bar selection cannot rewrite the dragged list.
2. Output and input writes now read the CoreAudio default back before publishing success. Priority enforcement remembers the failed target together with the observed current default and available-device set, so an identical listener event cannot retry every 150 ms. A real target/default/availability/preference change permits a new attempt; the pure policy cases are covered in the test suite.
3. The PR body now states the separate microphone-list storage decision and the existing input-only last-writer restoration behavior explicitly.
4. Current `main` is merged and the three conflicts are resolved.

Verification after the merge: `swift build`, release `./build.sh`, `./build/Vorssaint --selftest`, and all 10,732 `./build.sh --test` checks pass. Ready for another review.

Sources/Vorssaint/Services/Audio/AudioPriorityService.swift
```diff
@@ -0,0 +1,357 @@
...
+        // services. These are already @Published on the main thread, so
+        // enforcement runs there too — no competing HAL listeners.
+        AppVolumeMixer.shared.$outputDevices
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in self?.scheduleEnforcement() }
+            .store(in: &cancellables)
+
+        AppVolumeMixer.shared.$currentOutputDeviceUID
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in self?.scheduleEnforcement() }
```
vorssaint: with output priority on, an output picked with the output switcher shortcut, in the command bar or in the sound menu flips and then flips back a moment later. the caption promises a switch only when a device connects or disconnects, so a choice made anywhere should stick until that happens, without reordering the list

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -103,12 +106,64 @@ final class AudioInputDeviceManager: ObservableObject {
...
+        // Record a setter that CoreAudio accepted even when the immediate
+        // read-back is stale. If the target appears asynchronously, the
+        // existing last-writer guard can still restore the original input on
+        // quit; when it never applies, that guard sees a different current UID
+        // and leaves the system alone.
+        if inputDeviceBeforeOverride == nil {
+            inputDeviceBeforeOverride = deviceBeforeOverride
+        }
+        appliedInputDeviceUID = device.uid
+        guard Self.defaultInputDeviceUID() == device.uid else {
```
vorssaint: this write and the read after it run on the main thread, right when a device is connecting or dropping, which is when they can hang and freeze the app. the preferred microphone already does its write off the main thread for that reason, can the automatic output and microphone switches go the same way

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -200,7 +255,7 @@ final class AudioInputDeviceManager: ObservableObject {
             savedUID: Defaults.sanitizedPreferredInputDeviceUID(
                 UserDefaults.standard.string(forKey: DefaultsKey.preferredInputDevice)),
             inputDeviceBeforeOverride: inputDeviceBeforeOverride,
-            mayApplyPreferred: !applyingPreferred)
+            mayApplyPreferred: !applyingPreferred && !inputPriorityIsActive)
```
vorssaint: with microphone priority on, the microphone picker keeps showing the old preferred choice as selected while the list marks a different microphone as current, and at launch the old choice is applied first and then switched away. while the list is on, the picker and the list need to agree on one microphone, and only that one should be applied

Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
```diff
@@ -493,16 +526,35 @@ final class AppVolumeMixer: ObservableObject {
             return false
         }
 
+        // A successful setter return does not guarantee that the HAL accepted
+        // the new default. Publish success only after reading the property
+        // back, otherwise priority enforcement would optimistically expose the
+        // target and retry it forever when the listener restores reality.
+        guard Self.defaultOutputDeviceUID() == device.uid else {
```
vorssaint: if the system applies the switch a moment after the write, this now shows the unavailable message for a pick that does end up selected. does that happen with any of your devices, and if it can, the picker should not report a failure for it

build.sh
```diff
@@ -40,8 +40,8 @@ for arg in "$@"; do
 done
 
 if (( DEV )); then
-    APP_NAME="Vorssaint (Developer)"
-    EXECUTABLE="VorssaintDeveloper"
+    APP_NAME="Vorssaint - Test"
+    EXECUTABLE="VorssaintTest"
```
vorssaint: this renames the local developer build and its executable, and the smoke script and the code comment follow it. together with the other edits to the smoke script this is unrelated to the priority lists, and the repo keeps the developer name, so can you drop these changes

Sources/Vorssaint/Core/Localization.swift
```diff
@@ -724,6 +724,22 @@ struct Strings {
...
+    let audioPrioritySection: String
+    let audioPriorityOutputEnable: String
+    let audioPriorityInputEnable: String
+    let audioPriorityOutputList: String
+    let audioPriorityInputList: String
+    let audioPriorityAddOutput: String
+    let audioPriorityAddInput: String
+    let audioPriorityMoveUp: String
+    let audioPriorityMoveDown: String
+    let audioPriorityRemove: String
```
vorssaint: add output, add microphone and remove are never shown anywhere, the list has no add or remove controls. can you drop these three strings from every language

Sources/Vorssaint/Core/SettingsBackupSupport.swift
```diff
@@ -40,6 +40,9 @@ enum SettingsBackupSupport {
         DefaultsKey.mixerHiddenApps,
         DefaultsKey.preferredInputDevice,
         DefaultsKey.soundOutputSwitcherDeviceUIDs,
+        DefaultsKey.audioPriorityOutputUIDs,
+        DefaultsKey.audioPriorityInputUIDs,
+        DefaultsKey.audioPriorityDeviceNames,
```
vorssaint: these three keys already have registered defaults, so they travel in the backup on their own. this list is only for keys without a registered default, can you take them out of it

build.sh
zamai: Addressed in ee817ae8. I reverted the unrelated app-name and UI-smoke changes.

Sources/Vorssaint/Core/SettingsBackupSupport.swift
zamai: Addressed in ee817ae8. Removed the three registered audio-priority keys

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
zamai: Agree, implemented: when input priority is active, the picker binds to the actual current input and concrete picker selections update that current device without overwriting the dormant preferred-microphone setting. Priority ownership is established before the first input refresh, so launch no longer applies the old preferred input and then switches again.

Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
zamai: Addressed in ee817ae8. The immediate output read-back was removed. A successful HAL write is followed by the normal asynchronous refresh, which publishes the actual default when CoreAudio reports it; delayed devices no longer produce a false unavailable/picker failure.

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
zamai: Addressed in ee817ae8. Automatic output and input writes now run on the existing dedicated HAL queues. The main thread only publishes the eventual refresh result or error, so device connect/disconnect work no longer performs blocking CoreAudio writes on main.

Sources/Vorssaint/Core/Localization.swift
zamai: Addressed in ee817ae8. Removed the unused fields and their values from every supported language.

Sources/Vorssaint/Services/Audio/AudioPriorityService.swift
zamai: Haven't thought about this initially, I guess I'll implement what's advertised on caption. 
Addressed in ee817ae8. Priority now observes settled eligible-device UID sets and enforces only when that inventory changes. Manual output or input changes, startup, toggles, and list reordering neither trigger enforcement nor mutate the order.

(other related feature might be "Keep Selected Input always a priority" or something like this but I guess this is out of scope)

zamai:
One remaining design fork from #1191 is how microphone preferences should be stored.

This implementation takes the following route:

- Keep the existing `preferredInputDevice` preference unchanged.
- While microphone priority is enabled, the ordered priority list owns selection and `preferredInputDevice` remains dormant.
- When priority is disabled or the feature is uninstalled, the existing preferred-microphone behavior resumes.

I chose this route because Audio device priority is opt-in: it avoids migrating an established preference, preserves behavior for existing users, and gives the two settings mutually exclusive ownership rather than letting them compete at runtime.

The other route I considered is migrating `preferredInputDevice` into the ordered list and making that list the single source of truth. That removes the second stored preference, but it also requires migration and a new definition for which microphone should be preferred when priority is disabled or uninstalled.

The PR currently assumes the first route. 
@vorssaint  If you prefer the single-source-of-truth route, please say so and I will rework the storage and transition behavior before merge.

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -119,6 +139,39 @@ final class AudioInputDeviceManager: ObservableObject {
...
+        refresh.discardInFlight()
+        halQueue.async { [weak self] in
+            let status = Self.setDefaultInputDevice(device.audioObjectID)
+            DispatchQueue.main.async {
+                guard let self, self.listenerInstalled else { return }
+                if status == noErr {
+                    if self.inputDeviceBeforeOverride == nil {
+                        self.inputDeviceBeforeOverride = deviceBeforeOverride
+                    }
+                    self.appliedInputDeviceUID = device.uid
```
vorssaint, line 161: with microphone priority on, quitting puts the microphone back to the one in use before the first automatic switch, and the next launch never applies the list again, so after every restart the mic stays on the wrong device until something is plugged in or unplugged. the microphone the list picked should still be the one in use after a restart

Sources/Vorssaint/Services/Audio/AudioPriorityService.swift
```diff
@@ -0,0 +1,332 @@
...
+        let sanitized = Defaults.sanitizedAudioPriorityUIDs(uids)
+        let defaults = UserDefaults.standard
+        if sanitized.isEmpty {
+            defaults.removeObject(forKey: DefaultsKey.audioPriorityOutputUIDs)
+        } else {
+            defaults.set(sanitized, forKey: DefaultsKey.audioPriorityOutputUIDs)
+        }
+        outputPriorityUIDs = sanitized
+        updateDeviceNames()
+    }
```
vorssaint, line 134: dragging a connected device to the top, or turning the switch back on, now changes nothing until something is plugged in or unplugged, so right after setup the list names one device while another stays in use. a change the user makes to the list itself should put its top connected device in use right away

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
zamai: Addressed in f0a0d712 (now on the PR branch). When microphone priority is active, quitting leaves the selected system input in place; when priority is off, the existing preferred-microphone restoration behavior still applies. I selected Insta360 Link 2 Pro from the priority list on a Mac, quit the developer app, confirmed macOS kept that input, then relaunched and confirmed the list still marked it current. The merged branch passes all 60,005 tests.

Sources/Vorssaint/Services/Audio/AudioPriorityService.swift
zamai: Addressed in f0a0d712 (now on the PR branch). Turning either priority switch on or changing its ordered list now schedules one coalesced selection of the highest available device; the 250 ms delay keeps a drag from issuing a switch at every hover position. Ordinary default-device changes still leave manual choices in place and do not reorder the list. I verified reorder and off/on behavior with MacBook Pro Speakers and BenQ RD280UG, then Wireless Mic Rx and Insta360 Link 2 Pro. The merged branch passes all 60,005 tests.

zamai:
@vorssaint Ready for another review. 

I addressed both remaining behavior comments in f0a0d712 
- editing either ordered list or re-enabling priority now selects its top available device, and a priority-selected microphone remains selected across quit and relaunch. 
- Manual device choices still remain in effect until the device inventory or priority preferences change.

Tests/FeatureCatalogTests.swift
```diff
@@ -380,7 +380,7 @@ enum FeatureCatalogTests {
             "mouseClickDebounce", "keyboardDebounce", "textSnippets", "superKey", "quitWindowProtection",
             "clipboardHistory", "pastePlain", "finderCutPaste", "finderRename", "shelf", "urlCleaner",
             "diskImageInstaller",
-            "mixer", "soundOutputSwitcher", "micMute", "musicBlock",
+            "mixer", "soundOutputSwitcher", "audioPriority", "micMute", "musicBlock",
```
vorssaint: with notch agents from main and audio priority from this branch the catalog now has 71 features, but the count check a few lines above still expects 70, so the features suite fails. can you bump it to 71

Sources/Vorssaint/UI/MenuPanel/MixerSection.swift
```diff
@@ -344,10 +346,22 @@ struct MixerSection: View {
 
     private var inputSelectionBinding: Binding<String> {
         Binding(
-            get: { inputManager.preferredInputDeviceUID ?? MixerRoutingSupport.systemDefaultSelectionID },
+            get: {
+                if audioPriority.inputPriorityEnabled {
```
vorssaint: the microphone menu in the dynamic island mixer still names and checks the saved preferred microphone, or the default entry when none is saved. with microphone priority on it can name a different microphone than this picker and the priority list, and picking one there switches the input while the check stays where it was. can that menu follow the same rule as this picker while microphone priority is on

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -252,7 +312,8 @@ final class AudioInputDeviceManager: ObservableObject {
             savedUID: Defaults.sanitizedPreferredInputDeviceUID(
                 UserDefaults.standard.string(forKey: DefaultsKey.preferredInputDevice)),
             inputDeviceBeforeOverride: inputDeviceBeforeOverride,
-            mayApplyPreferred: !applyingPreferred,
+            mayApplyPreferred: !applyingPreferred && !inputPriorityIsActive,
```
vorssaint: this was already in the earlier push and was missed then. with volume mixer uninstalled and audio priority installed this manager keeps running, so with the microphone switch off, including at launch, it applies the preferred microphone saved earlier in the mixer and puts it back every time the input is changed in system settings. nothing in the app shows that choice anymore or lets the user clear it. the saved preferred microphone should only apply while volume mixer is installed

Sources/Vorssaint/Services/Audio/AudioPriorityService.swift
```diff
@@ -0,0 +1,358 @@
...
+        cancellables.removeAll()
+        enforceDebounce?.cancel()
+        enforceDebounce = nil
+        settledOutputUIDs = nil
+        settledInputUIDs = nil
+        pendingOutputUIDs = nil
+        pendingInputUIDs = nil
+        pendingOutputPreferenceChange = false
+        pendingInputPreferenceChange = false
+        AudioInputDeviceManager.shared.setInputPriorityActive(false)
```
vorssaint: this was also there before and should have come with the earlier requests. uninstalling audio priority stops the service here but leaves its output and microphone switches on in memory until the next launch, and the mixer microphone picker still reads the microphone switch. for the rest of that session the picker shows the current input instead of the saved one, a pick is not saved, and when a preferred microphone is saved and connected each pick is taken back by it a moment later. can stop reset both switches so every menu that reads them follows the uninstall

Tests/FeatureCatalogTests.swift
zamai: Addressed in 4b0250fc after merging current main. The stable feature-catalog count now expects 71 entries, including Notch agents and Audio device priority. The feature suite passes all 934 checks.

Sources/Vorssaint/UI/MenuPanel/MixerSection.swift
zamai: Addressed in 4b0250fc. The Dynamic Island mixer now observes AudioPriorityService and uses the same selected-input helper as the panel picker. While microphone priority is on, its title and checkmark follow currentInputDeviceUID, and choosing a concrete microphone calls setCurrentInputDeviceUID without changing the dormant preferred microphone. The fallback row is also hidden while priority owns selection. The focused audio and mixer suites pass.

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
zamai: Addressed in 4b0250fc. Input resolution now receives Volume mixer availability. When Mixer is uninstalled, the saved preferred UID stays stored but is ignored: the manager follows the current system input, reports no preferred-device fallback, and never reapplies the hidden preference. Reinstalling Mixer makes the saved preference active again. A focused policy check covers the priority-only case.

Sources/Vorssaint/Services/Audio/AudioPriorityService.swift
zamai: Addressed in 4b0250fc. AudioPriorityService.stop() now clears both published enable flags in memory before releasing microphone ownership. The persisted switches and ordered lists remain intact for reinstall, but every live menu immediately returns to non-priority selection behavior after uninstall.

zamai:
@vorssaint Addressed all four comments from the latest review in `4b0250fc` and replied in each thread.

- Updated the feature catalog count to 71 after merging current `main`.
- Made the Dynamic Island microphone menu follow the actual current input while microphone priority is active.
- Prevented a saved Mixer microphone preference from applying while Volume mixer is uninstalled.
- Reset both live priority switches when Audio device priority is uninstalled, while retaining their stored values for reinstall.

The branch is current and mergeable. `swift build` passes; the focused Audio priority, Mixer, and feature suites pass all 1,239 checks; signed release and developer bundles build and both self-tests pass. The PR description records the full local test result and the current-main-only Notch/Switcher failures reproduced in a clean worktree.

The fork CI run is awaiting repository approval: https://github.com/vorssaint/vorssaint-utils/actions/runs/35878894464. The `incomplete` label also needs repository permission to clear.

Sources/Vorssaint/Services/Audio/AudioInputDeviceManager.swift
```diff
@@ -81,7 +85,14 @@ final class AudioInputDeviceManager: ObservableObject {
 
     func stop() {
         removeVolumeListeners()
-        restoreOriginalInputDevice()
+        // Priority selections are meant to survive a quit and the next
+        // launch. Only the singular preferred-microphone override is restored.
+        if !inputPriorityIsActive {
```
vorssaint: this was already in the earlier push and was missed then. a switch made by the microphone list, or a pick made in a mixer menu while the list is on, stays recorded as the app's own change after the microphone switch is turned off or audio priority is removed. when no saved preferred microphone takes over, as with default chosen in the mixer or without volume mixer, quitting while that microphone is still in use puts back the one from before the first switch, and removing audio priority without volume mixer does it right away. nothing picks the list's microphone again on the next launch. can those switches stay in use like a choice made in system settings, so only a change made by the saved preferred microphone is undone?

zamai: Addressed in 615d7ac8. setCurrentInputDeviceUID is used only by priority enforcement and Mixer selections while priority is active. After a successful HAL write it now clears, rather than creates, the preferred-microphone restoration record, so that microphone becomes the persistent system choice across disabling or removing priority and across quit. If the saved Mixer preference later takes over, applyPreferredInputDevice creates a fresh restoration record from that priority-selected microphone, so only the saved preference change is undone. The focused Mixer and Audio priority suites pass all 305 checks; the stamped signed developer bundle builds and passes self-test.

zamai:
yo-ho! so exited to get this merged into the app, this way I can consolidate my mac toolbox, droping https://github.com/tobi/AudioPriorityBar in faivour of swiss army knife 😄  

thanks! @vorssaint

Issue #291, opened 2026-07-19
Port management


zcyc:
### Problem or Motivation

manager local ports
manager port forward
manager ssh tunnel

### Proposed Solution

like [PortKiller](https://github.com/productdevbook/port-killer)

### Feature Area

Menu Bar

PathGao:
Triage note, part of a batch pass over the open issues. The labels are a contributor's read, not a maintainer ruling.

Three separate features are bundled here: listing local listening ports, port forwarding, and SSH tunnels. Which one you actually want changes the answer completely — a port list is a read-only enumeration, while forwarding and SSH tunnels need long-lived child processes and credential handling, which is a scope decision for a local utility rather than a straightforward feature.

Worth knowing: PR #753 (`feat(kill-process): show listening ports`) covered the listing half and was closed on 2026-08-24 without merging, so even the cheap part currently has no branch behind it.

Splitting this into one issue per item is what unblocks it. As a single issue it can be neither accepted nor declined, because the three parts have different answers.

PR #1260, opened 2026-09-01
feat(utilities): add port manager

MrDXTR:
## Summary

Adds a compact Port Manager utility for viewing TCP listening ports and the
processes using them.

Users can:

- View listening ports, protocols, addresses, processes, and PIDs
- Search by port, process, PID, or address
- Refresh the port list
- Gracefully terminate a process (requires the Kill Process feature)
- Force terminate a process after confirmation (requires the Kill Process feature)

> **Note:** Process termination actions (Kill / Force Kill) are gated behind the **Kill Process** feature (`AppFeature.killProcess`, an opt-in beta in the Features Hub). On a fresh install, Port Manager displays active listening ports in read-only mode until Kill Process is enabled.

The feature uses macOS's built-in `lsof` command and Darwin process signals,
with no runtime dependencies or external services. It is available both from
the Utilities panel and Settings.

## Verification

Tested on a MacBook Air with Apple M4 running macOS 26.6.1, build 25G76.

- Built the developer app successfully
- Ran the app and verified Port Manager behavior
- Verified port listing and filtering
- Verified kill buttons appear only when Kill Process is enabled
- Verified normal and force-kill confirmation flows
- Verified the compact scrollable UI
- `SELFTEST OK`
- `git diff --check` passes
- `./build.sh --test` passes


PathGao:
Clean split — 76-line service, no deps, and all 13 locales in from the start instead of English-only.

**1. The feature gate is special-cased.** `items(editing:)` carries `$0 == .portManager ||` because `.portManager` maps to `.killProcess`, which ships unavailable (beta, excluded in `availabilityDefaults`). So Port Manager becomes the only utility that can't be switched off in the hub, and `features(for: .portManager) -> []` keeps its Settings page permanently visible too. The body says "available both from the Utilities panel and Settings" but doesn't mention it's ungated. Add `case portManager` to `AppFeature` (group `.tools`) and both workarounds go away.

**2. Only one port per process survives.** `snapshot()` appends a row when it reaches the next `p` line, so every `n` overwrites the last one. A process on 3000 and 3001 shows one row, and the IPv4/IPv6 pair for the same port collapses. Append on `n`.

**3. Six locales swap PID and port.** `terminateMessageFormat` is called with `(port, pid)`, but tr/ja/ko/zhHans/zhTW/zhHK write PID first. Needs `%1$d`/`%2$d`.

I read the parser, didn't run it — #2 is unverified against real `lsof` output.

**Impact:** #2 and #3 hit every user. #1 makes the feature unswitchable.
**Merge:** needs the `AppFeature` case first.

ref #291

MrDXTR:
1. Added a dedicated `AppFeature.portManager` case in the `.tools` group, including its symbol, availability defaults, energy profile, Settings destination, and feature visibility. The special-casing workarounds are no longer needed.
2. Updated `snapshot()` to emit a row for each `n` record, so a process with multiple listening ports and separate IPv4/IPv6 endpoints is represented correctly. Also filtered exact duplicate records.
3. Corrected the positional format specifiers in Turkish, Japanese, Korean, Simplified Chinese, Traditional Chinese, and Hong Kong Chinese so the port is `%1$d` and PID is `%2$d`.

PathGao:
All three landed, and the real `AppFeature` case took both workarounds with it.

Two left, both "reuse what's already here":

**1. The kill path skips the guards its sibling has.** `KillProcessService` won't signal a pid unless `KillProcessSupport.isProtected(pid:name:)` clears it and `identityMatches` confirms the start time still matches the row — its own comment: "a PID alone is reusable and is not a safe identity for a destructive action confirmed from an old row." `PortManagerService.terminate` signals a pid from a snapshot taken before the alert, and discards the result, so an EPERM kill (that path escalates it through `AdminShell`) reads as a no-op.

**2. `snapshot()` hand-rolls Process+Pipe.** `Shell.run("/usr/sbin/lsof", args)` is one line and brings the timeout and bounded drain with it. Today `readDataToEndOfFile()` runs *after* `waitUntilExit()`, so a full pipe deadlocks and `isRefreshing` never clears; the blocking wait also parks a shared-pool thread, which is what `BoundedProcessRunner` exists to avoid (#971). It also frees the parser to be pure and testable like `NetworkProcessSupport.parseNettopCSV` — the `n` bug was one test away.

Minor, unverified: `lsof` truncates the command name at 9 chars unless you pass `+c0`.

**Impact:** #1 can signal a reused pid; #2 kills Refresh for the session when it hits.
**Merge:** needs #1 and #2 first.

ref #291

PathGao:
Sending this back for the two items from the 2026-09-01 review, both in `PortManagerService.swift`.

1. `process.waitUntilExit()` runs before `output.fileHandleForReading.readDataToEndOfFile()`. Once `lsof` writes more than the pipe buffer, it blocks on write, the app blocks on wait, and `isRefreshing` never clears. The repository has two runners for exactly this: `Shell.run` and `BoundedProcessRunner.run(_:_:timeout:)`, which drain while the process runs and kill on timeout. Use one of them.
2. `Darwin.kill(entry.pid, force ? SIGKILL : SIGTERM)` with only `pid > 1` and not-self as guards. `KillProcessSupport.isProtected(pid:name:path:)` exists to keep the window server, loginwindow and the like off the list, and `KillProcessService` already handles `EPERM` by escalating through `AdminShell`. Route the kill through that, or the port list becomes a second, unguarded kill tool.

The list itself is fine and the idea holds (read-only `lsof`, no dependency). Push those two and the review continues.

MrDXTR:
Thanks for the follow-up review. Both remaining items are addressed and pushed in `d39e958`.

1. Port Manager no longer signals the snapshot PID directly. Each row captures the process kernel start time, and termination now routes through `KillProcessService`. That reuses the existing protected-process checks, start-time identity validation, direct signal handling, and `EPERM` → `AdminShell` escalation for both normal and force termination. Rows without a valid identity, including protected processes, have their controls disabled.

2. Port discovery now uses `Shell.run("/usr/sbin/lsof", args)`, so it gets the repository’s bounded output drain and timeout behavior instead of waiting before reading the pipe. The parser was moved into the pure `PortManagerSupport.parseLsof` helper and is covered by regression tests for multiple ports, IPv4/IPv6 endpoints, and exact duplicate records.

Also included the minor `+c0` improvement so process names are not truncated by `lsof`, plus the Settings separator cleanup.

Verification completed:
- `./build.sh --test` passed: `TESTS OK (9600 checks)`
- `./build.sh --dev` completed successfully and produced the developer app
- `git diff --check` passes
- No runtime dependencies or external services were added

PathGao:
Routing the kill through `KillProcessService` was the right call — the identity check, the protected list and the `AdminShell` escalation all come along with it, and the parser is pure and tested now.

Three left, all "the sibling already handles this":

**1. A failed `lsof` wipes the list.** `snapshot()` returns `[]` on any non-zero status and `refresh` assigns it unconditionally. `Shell.run` returns `-1` on its 5s timeout, and `lsof` exits 1 whenever it prints a warning even with good rows above it (not verified here). `KillProcessService.refresh` refuses this exact thing — "must never wipe a list that was already showing good data". Return `[PortManagerEntry]?`, `nil` when `status < 0`, assign only when non-nil.

**2. The killed row stays until a manual refresh.** `finishKill` fires `completion` from the immediate `main.async`, so `refresh()` runs before anything is reaped — `terminate()` and `Darwin.kill` both return at once. The 0.5s reconcile under it exists for this. Move `completion?()` into the `asyncAfter` block.

**3. `PanelSection.utilities.featureGate` has no `.portManager`.** `isAvailable` is `featureGate.contains(where: \.isAvailable)`, so with every other utility off in the hub the section leaves the panel and takes the Port Manager button with it.

**Impact:** #1 blanks the list on a slow or noisy `lsof`; #2 makes every kill look like it did nothing.
**Merge:** needs all three first.

ref #291

MrDXTR:
All three issues have been resolved:

1. **Keep existing entries on failed/slow `lsof`**: `PortManagerService.snapshot()` now returns `[PortManagerEntry]?` (`nil` when `lsof` exits with non-zero or times out). `refresh()` only assigns `entries` when `result` is non-nil, preventing transient errors/timeouts from wiping existing rows.
2. **Reap killed processes before refreshing**: In `KillProcessService.finishKill`, `completion?()` was moved inside the 0.5s `asyncAfter` block so callers (like `PortManagerService`) only re-query `lsof` after Darwin has actually reaped the terminated process.
3. **Include Port Manager in Utilities feature gate**: Added `.portManager` to `PanelSectionID.utilities.featureGate` so the Utilities section and Port Manager button remain available in the menu panel even when all other utilities are turned off in the Hub.

PathGao:
All three landed — `nil` on failure, `completion` inside the reconcile, `.portManager` in the utilities gate.

**1. Port Manager ships on; Kill Process ships off.** `availabilityDefaults` excludes `.killProcess` and `isBeta` is true for it — termination is a deliberate opt-in beta. `.portManager` is in neither list, so it installs by default and its Kill / Force Kill buttons reach `KillProcessService.kill` with no availability check. Unless that's intended, add `.portManager` to the `availabilityDefaults` exclusion, or gate the buttons on `AppFeature.killProcess.isAvailable`.

**2. `status == 0` is stricter than asked.** `Shell.run` returns `-1` on timeout, but `lsof` also exits 1 when it merely prints a warning with good rows above it — then `snapshot()` returns `nil` every time and the view reads "No listening ports found. Try refreshing." Parse first; return `nil` only on `status < 0`, or on a non-zero exit that parsed nothing. `KillProcessService` keeps `hasLoadedOnce` for the same distinction.

The lsof-exits-1-on-warning behaviour is from the man page, not verified on a machine.

**Impact:** #1 hands every user an ungated force-kill; #2 can leave the list blank with no error.
**Merge:** needs #1 answered and #2 fixed.

ref #291

MrDXTR:
1. Kill buttons gated on `killProcess.isAvailable`** ([PortManagerView.swift](file:///Users/manav/Developer/vorssaint-utils/Sources/Vorssaint/UI/KillProcess/PortManagerView.swift)): went with gating the buttons rather than adding `.portManager` to `availabilityDefaults`. Port Manager ships on and is useful to everyone for inspecting ports; only the destructive Kill/Force Kill actions now require Kill Process to be enabled, consistent with it being the deliberate beta opt-in. No change to `availabilityDefaults` or `isBeta`.

2.  Parse-first, `nil` only on genuine failure** ([PortManagerService.swift](file:///Users/manav/Developer/vorssaint-utils/Sources/Vorssaint/Services/PortManager/PortManagerService.swift)): `snapshot()` now always parses first. `status < 0` (Shell.run timeout) → `nil` unconditionally. Non-zero with parsed rows (lsof exit-1-on-warning) → return those rows. Non-zero with an empty parse → `nil` so a genuine error doesn't replace good data with a blank list.

PathGao:
Everything from the last round landed: the buttons gate on `killProcess.isAvailable`, `snapshot()` parses before it judges the exit status, and `+c0` is in.

**1. The ship default is three grey buttons and no explanation.** `killProcess` is excluded from `availabilityDefaults`, so on a fresh install every row's Kill and Force Kill are disabled with nothing saying why and no path from here to the Hub toggle. The repo's answer elsewhere is to omit, not grey: `CommandBarCatalog` skips the kill row entirely under that same `isAvailable` check, and `features(for: .killProcess) == [.killProcess]` hides the whole Kill Process settings page. Wrap the button `HStack` in `if AppFeature.killProcess.isAvailable` — the two duplicated three-clause `.disabled` chains collapse with it.

**2. The list only covers the current user's processes.** Without root, `lsof -i` can't read other users' descriptors, so root-owned listeners never show and `%d open` undercounts. The empty state then advises "Try refreshing or changing your search," which isn't the reason. A word in `listeningCaption` covers it; I wouldn't add a mechanism. That's from the lsof man page, not verified on a machine.

**Impact:** #1 hits every fresh install; #2 makes the count and the empty state misleading.
**Merge:** needs #1 first.

ref #291

MrDXTR:
Both addressed:

**#1 — Buttons hidden instead of greyed** ([PortManagerView.swift](file:///Users/manav/Developer/vorssaint-utils/Sources/Vorssaint/UI/KillProcess/PortManagerView.swift)): wrapped the `HStack` in `if AppFeature.killProcess.isAvailable`. When Kill Process is off, the row shows only the port/process info with no buttons, nothing disabled, nothing to explain. The `!isAvailable` clause also falls out of both `.disabled` chains as a bonus, since it's now structural.

**#2 — Scope caveat in `listeningCaption`** ([PortManagerStrings.swift](file:///Users/manav/Developer/vorssaint-utils/Sources/Vorssaint/Core/PortManagerStrings.swift)): updated all 13 locales from e.g. `"Listening ports"` → `"Your listening ports"` (possessive phrasing, consistent across all locales). This appears as the panel tile subtitle and the Hub settings caption — right where a user would first see the count — and sets the expectation without a dedicated warning or new mechanism.

PathGao:
`+c0` turned out to matter more than a cosmetic fix — `KillProcessSupport.isProtected` matches on name, and `windowserver` and `loginwindow` are both past lsof's 9-char default truncation. Right call to take it.

One left.

**`panelUtilityPortManager` never reaches a settings backup.** The other panel utility toggles are all in `Defaults.registeredDefaults` (`panelUtilityScreenRecorder` and friends); this one is only a `DefaultsKey` constant. `SettingsBackupSupport.exportKeys()` is `registeredDefaults.keys` ∪ `availabilityDefaults.keys` ∪ `unregisteredPreferenceKeys`, so hiding Port Manager from the panel is dropped on export and comes back on the other Mac. The `@AppStorage(...) = true` default masks it inside the panel, and the backup assertions are per-key with no parity check, so the suite stays green. Add `DefaultsKey.panelUtilityPortManager: true,` beside the ScreenRecorder line.

Minor: the body still promises terminate and force-terminate. With the gate, a fresh install shows no kill buttons until Kill Process is on — worth a line so the release notes don't read as broken.

**Impact:** the panel visibility choice is lost across settings export/import; nothing else.
**Merge:** needs the `registeredDefaults` line first.

ref #291

MrDXTR:
Added `DefaultsKey.panelUtilityPortManager: true` to `Defaults.registeredDefaults` beside `panelUtilityScreenRecorder` so that the panel visibility setting is included in `SettingsBackupSupport.exportKeys()` and persists across backup export/import.

 Also updated the PR description / notes to clarify that process termination (Kill / Force Kill) from Port Manager requires the **Kill Process** feature to be enabled in the Features Hub (opt-in beta). On a fresh install, only the port/process listing is shown until Kill Process is turned on.

PathGao:
`panelUtilityPortManager` is in `registeredDefaults` now, so the panel visibility choice survives export/import. That was the last one.

I re-read the rest against `main`:

- `Defaults.sanitizedPanelItemOrder` appends cases missing from a saved order, so existing users get the tile without a migration.
- `KillProcessSupport.isProtected` guards by name; the path clause only backs three names it already covers. With `+c0` the lsof `c` field matches every entry, so calling it without a path loses nothing.
- `Shell.run` points `standardError` at the same pipe, so lsof's warnings reach `parseLsof`. They all start with `lsof:` or a space, and the parser drops any line whose first char isn't in `pcnPT` — no collision with the parse-first change.

Nothing left at the level of the earlier rounds. Not verified on hardware: the parser has the new fixture test, the rest of what I checked is read-only.

**Impact:** none.
**Merge:** ship it.

ref #291

PathGao:
dont worry to update branch。when it come to merge, we will update it。But may we will have some refactor recently so we hope you can come back later to resolve the conflict

Yash121l:
I built PortHole (https://github.com/Yash121l/PortHole) for this exact need before finding this PR, so I am keen to see it land. I will build this branch on my M3 (macOS 26.6.2) tonight and report back on the listing, the search and both kill flows.

Yash121l:
Tested `7eb1642` on an M3, macOS 26.6.2. `./build.sh --test` passes (38238 checks) and `--dev` builds. Against real `lsof -nP +c0 -iTCP -sTCP:LISTEN -F pcnPT` output the parser gives 19 rows matching the column-format cross-check, including IPv6 and `*` binds, a node cluster sharing a port, and a command with a space. Two listeners I started (`127.0.0.1:8765` and a `0.0.0.0` bind on 8766) appear with the right PID; note lsof reports the second as `*:8766`, so `0.0.0.0` never reaches the UI as a string. The start-time identity check refuses a kill once the PID is gone. I did not run a real kill since Kill Process ships unavailable and the fallback prompts for admin.

Four small things while reading it:

- `PortManagerSupport.swift:31-33`: when an `n` line has no trailing port, `port` keeps the previous line's value and a row is emitted pairing the new address with the old port. Latent under `-sTCP:LISTEN`, live the moment UDP or a wider `-i` is added. `guard let ... else { continue }` fixes it.
- `PortManagerService.swift:24`: trailing whitespace, so `git diff --check` does not pass as the description says.
- `FeatureCatalog.swift:342`: a stray `\` appended to an unrelated doc comment.
- `PortManagerService.swift:29`: `terminate` calls `KillProcessService.shared.kill` without checking `AppFeature.killProcess.isAvailable`; the gate lives only in the view.

Separately, and not for this PR: PortHole marks binds beyond localhost as exposed, infers labels from the real command line (a `node` on 5173 shows as Vite with its project folder), lists UDP bound sockets, has open-in-browser and copy actions, and can notify when a port appears exposed. @MrDXTR @PathGao would you take any of those as follow-ups once this lands? I will open them one issue each with the problem stated, and I can bring the PortHole implementations across.

Sources/Vorssaint/Core/FeatureCatalog.swift
```diff
@@ -338,7 +339,7 @@ extension AppFeature {
     }
 
     /// Registered defaults preserve existing features on update. New opt-in
-    /// features and explicit betas ship uninstalled.
+    /// features and explicit betas ship uninstalled.\
     static var availabilityDefaults: [String: Any] {
         Dictionary(uniqueKeysWithValues: allCases.map {
             ($0.availabilityKey,
```
vorssaint, line 365: new tools ship uninstalled here until someone turns them on in features, and kill process does the same. can you add this one to that list so it does not land in every panel and sidebar on the next update. the test that lists those features needs the same change

Sources/Vorssaint/Core/FeatureCatalog.swift
```diff
@@ -338,7 +339,7 @@ extension AppFeature {
     }
 
     /// Registered defaults preserve existing features on update. New opt-in
-    /// features and explicit betas ship uninstalled.
+    /// features and explicit betas ship uninstalled.\
```
vorssaint: a stray backslash got appended to this comment, can you drop it

Sources/Vorssaint/UI/PortManager/PortManagerView.swift
```diff
@@ -0,0 +1,132 @@
...
+                    .background(Color.primary.opacity(0.07), in: Capsule())
+                Button { service.refresh() } label: { Image(systemName: "arrow.clockwise") }
+                    .buttonStyle(.plain).help(strings.refresh)
+            }
+            .padding(.horizontal, 14).padding(.top, 10).padding(.bottom, 8)
+            HStack(spacing: 8) { Image(systemName: "magnifyingglass").foregroundStyle(.secondary); TextField(strings.filter, text: $service.query).textFieldStyle(.plain) }
+                .padding(9).background(Color.primary.opacity(0.06), in: RoundedRectangle(cornerRadius: 9))
+                .padding(.horizontal, 14).padding(.bottom, 8)
+            Divider()
+            if service.filteredEntries.isEmpty {
```
vorssaint, line 41: the list starts empty and the first lookup is still running when this shows, so opening the tool flashes no listening ports found before the rows arrive. the kill process page keeps a loaded once flag and shows a spinner while it refreshes. can you do the same here so the empty message only appears after a real answer

Sources/Vorssaint/UI/PortManager/PortManagerView.swift
```diff
@@ -0,0 +1,132 @@
...
+                Spacer()
+                Text(String(format: strings.openFormat, service.filteredEntries.count))
+                    .font(.system(size: 11, weight: .semibold, design: .rounded))
+                    .foregroundStyle(.secondary)
+                    .padding(.horizontal, 8).padding(.vertical, 5)
+                    .background(Color.primary.opacity(0.07), in: Capsule())
+                Button { service.refresh() } label: { Image(systemName: "arrow.clockwise") }
+                    .buttonStyle(.plain).help(strings.refresh)
+            }
+            .padding(.horizontal, 14).padding(.top, 10).padding(.bottom, 8)
```
vorssaint, line 30: the other tools hosted in this panel share one look, a small label header with the close button on the right, the controls in a card and the list capped at 260 points so it shrinks with few rows. this one brings a back arrow on the left, a larger header, its own search box and a fixed 245 point box. can you build the panel version on the same pieces the clipboard and app updates tools use

Sources/Vorssaint/UI/Settings/FeatureHubSettings.swift
```diff
@@ -841,6 +842,7 @@ extension AppFeature {
                 + FeatureStrings.whatsAppDownloads(L10n.shared.language).hubDescription
         case .uninstaller: return hub.descUninstaller
         case .killProcess: return FeatureStrings.killProcess(L10n.shared.language).hubDescription
+        case .portManager: return FeatureStrings.portManager(L10n.shared.language).listeningCaption
```
vorssaint: the hub describes each feature in a sentence, like the kill process row next to it. this reuses the panel caption, so the row only says whose ports they are. can you give it a proper description in every language

MrDXTR:
Thank you for the review and thorough feedback! All points have been addressed in the latest commit:

1. **Default Availability**:
   - Opted out `portManager` in `AppFeature.availabilityDefaults` alongside `killProcess`, `fanControl`, `focusFollowsMouse`, and `diskImageInstaller` so it ships uninstalled by default for new installs and existing update migrations.
   - Cleaned up the trailing backslash in the doc comment on `AppFeature`.

2. **Feature Hub Description**:
   - Added a dedicated `hubDescription` field to `PortManagerFeatureStrings` in place of reusing `listeningCaption`.
   - Provided natural, complete localized strings across all 13 supported languages (EN, PT-BR, TR, RU, ES, DE, FR, IT, JA, KO, ZH-Hans, ZH-TW, ZH-HK) describing the utility.
   - Updated `FeatureHubSettings.swift` to reference `hubDescription`.

3. **`lsof` Parser Robustness**:
   - In `PortManagerSupport.parseLsof`, guarded the address and port parsing on `n` records: if an address line lacks a valid port delimiter or number, `port` is reset and the line skipped to prevent pairing with a stale port from a preceding line.

4. **Process Termination Availability**:
   - Guarded `PortManagerService.terminate(_:force:)` with `AppFeature.killProcess.isAvailable` so processes cannot be terminated when the kill process feature is disabled.
   - Added `hasLoadedOnce` to `PortManagerService` to cleanly distinguish between the initial snapshot loading state and an empty port list.

5. **Menu Panel & Dynamic Island Integration**:
   - Created `PanelPortManagerView` adhering to the standard menu panel utility design pattern (`panelCard` styling, header with settings gear & close button, bounded scroll area).
   - Added self-contained scroll-to-top handling so opening the utility from the bottom of the Utilities list in constrained containers (like Dynamic Island / Notch) resets cleanly to the top header without clipping content.

6. **Test Coverage**:
   - Added automated tests in `MetricsTests.swift` verifying:
     - `portManager` is uninstalled by default in `availabilityDefaults`.
     - `parseLsof` correctly ignores endpoints without a port without inheriting prior state.
     - All 13 locales contain a non-empty `hubDescription`.

Sources/Vorssaint/UI/MenuPanel/PanelPortManagerView.swift
```diff
@@ -0,0 +1,252 @@
...
+                    .frame(width: 22, height: 22)
+            }
+            .buttonStyle(.plain)
+            .help(l10n.s.uninstallerCancel)
+        }
+    }
+
+    private var controls: some View {
+        Group {
+            if notchPresentation {
```
vorssaint: in the notch this switches to its own layout, with the search on one line, the count moved under the list and a shorter list, which the clipboard and app updates tools never do. can you drop the notch only branches so it reads like its neighbours there. if the 260 point list really does not fit the notch, that is a change for every hosted tool, not this one

Sources/Vorssaint/UI/MenuPanel/PanelPortManagerView.swift
```diff
@@ -0,0 +1,252 @@
...
+                }
+                .padding(.horizontal, 4)
+            }
+        }
+        .onAppear {
+            PanelInteractionState.shared.viewKeepsPopoverOpen = true
+            service.refresh()
+        }
+        .onDisappear { PanelInteractionState.shared.viewKeepsPopoverOpen = false }
+        .scrollEnclosingToTopOnAppear()
```
vorssaint: this reaches into the panel around it and scrolls the whole thing to the top whenever the tool opens, and none of the other hosted tools do that. if opening a tool from the bottom of the list needs the panel back at the top, that belongs to the panel for every tool, so please drop it here

Sources/Vorssaint/Core/PortManagerStrings.swift
```diff
@@ -0,0 +1,56 @@
...
+    let title: String
+    let filter: String
+    let openFormat: String
+    let empty: String
+    let emptyHint: String
+    let listeningCaption: String
+    let kill: String
+    let forceKill: String
+    let refresh: String
+    let back: String
```
vorssaint: the back button is gone now, so this string is never shown in any language. can you remove it

Sources/Vorssaint/Services/PortManager/PortManagerService.swift
```diff
@@ -0,0 +1,63 @@
...
+                             address: entry.address,
+                             pid: entry.pid,
+                             processName: entry.processName,
+                             startedAt: KillProcessService.startTime(for: entry.pid))
+        }
+        // lsof exits 1 when it prints a warning but still outputs good rows above
+        // it; treat that as a valid snapshot. Only return nil when non-zero AND
+        // the parse came up empty — that's a genuine failure we should not use to
+        // replace good data already on screen.
+        if result.status != 0 && parsed.isEmpty { return nil }
```
vorssaint: when nothing is listening the lookup exits with a failure code and no rows, so this returns nothing, the first load never counts as done and the spinner stays up forever instead of the empty message. can that case count as a real empty snapshot

Yash121l:
Pulled b5439fe and re-ran the suite on my M3: 54136 checks pass. The parser fix is right, an address line with no usable port now resets `port` instead of inheriting the previous one, and since the snapshot is `-sTCP:LISTEN` there are no `->` rows that could hand you a remote port instead. Default availability, the 13 `hubDescription` strings and the terminate guard all look good. I'll open the PortHole follow-ups as one issue each once this lands.

MrDXTR:
Thanks for the review! All four items from the review comments have been addressed in `92d8650`:

1. **Standardized Panel Layout & Removed Notch-Only Branching** (`PanelPortManagerView.swift`):
   - Dropped the notch-only branching, single-line controls switch, and custom height logic so the layout matches the shared pattern used across the other panel utilities (`PanelClipboardView`, `PanelAppUpdatesView`). The list height is standardized at `260pt`.

2. **Removed Scroll-to-Top Container Hack** (`PanelPortManagerView.swift`):
   - Removed the `scrollEnclosingToTopOnAppear` helper and its AppKit `NSViewRepresentable` that scrolled the parent container.

3. **Removed Unused `back` String** (`PortManagerStrings.swift`):
   - Deleted the unused `back: String` property and its translations across all 13 supported locales.

4. **Handled Empty Port Snapshots on Zero Listening Ports** (`PortManagerService.swift`):
   - Updated `snapshot()` so that when `lsof` exits with status 1 due to zero active listening sockets, it returns a valid empty array `[]` instead of `nil`. This lets `hasLoadedOnce` transition to `true` and display the empty-state message instead of spinning indefinitely.

Yash121l:
92d8650 is green here too, 54954 checks on an M3.

Sources/Vorssaint/Services/PortManager/PortManagerService.swift
```diff
@@ -0,0 +1,61 @@
...
+            PortManagerEntry(port: entry.port,
+                             protocolName: entry.protocolName,
+                             address: entry.address,
+                             pid: entry.pid,
+                             processName: entry.processName,
+                             startedAt: KillProcessService.startTime(for: entry.pid))
+        }
+        // lsof exits 1 when no listening sockets are found or when it prints a
+        // warning. Both cases yield a clean result: an empty list or the parsed rows.
+        // Only non-timeout negative codes count as an infrastructure failure.
```
vorssaint: the last sentence of this note says the opposite of the check a few lines up, where a negative code is the timeout and already gives up. can you reword it so the two agree

MrDXTR: Done in 61f37ef8 — reworded the comment so that it aligns with the negative status timeout check above.

Yash121l:
Filed the follow-ups I mentioned, one each so they can be taken or declined separately: #1784 UDP listeners, #1785 marking ports bound to all interfaces, #1786 identifying the process by command rather than executable name, #1787 row actions for copy and open, #1788 optional notifications. I checked the enhancement register first and none of them were already tracked. Happy to implement any of them; #1785 is the cheapest and #1788 the one most likely to be noisy, so I would not start there.

PR #1264, opened 2026-09-01
refactor(switcher): move the session state machine out of the event tap

PathGao:
## Summary

`AppSwitcher` owned the session state machine and the machinery around it in the same object: roughly 40 mutable fields (selection, generation counters, the Shift-back chord, pending closes, search query, scope) sitting next to a `CGEventTap` on its own thread, the `NSPanel` lifecycle, and the wake and keyboard-layout observers. Nothing could drive the state machine without standing up an event tap and a panel, so nothing tested it.

The decisions were already testable — `SwitcherSupport` is in the `--test` source list and holds 75 pure functions. The *ordering* of those decisions was not, and that is where the bugs are: of 53 `Services/Switcher` files touched by fix commits in the last 200, 29 landed in files the test binary does not compile.

This moves the session into `SwitcherSession.swift`, a value type with one entry point:

```swift
mutating func apply(_ event: Event, environment: SwitcherSessionEnvironment) -> SwitcherSessionOutcome
```

11 event cases in; an outcome carrying 16 effect cases plus `swallowsEvent` out. `AppSwitcher` keeps the tap, the panel, `NSScreen` placement, preview images, pointer hover and `decodeKey`, and becomes the thing that feeds events in and performs the effects that come back. `SwitcherSession` imports only `CoreGraphics` and `Foundation`.

Behaviour is unchanged. No timing, debouncing or ordering was altered.

**What did not move, and why.** The `@Published` mirrors `SwitcherView` binds to stay on `AppSwitcher` — they are now written only while performing a publish effect, and `SwitcherView` has a zero-line diff. `routeSessionActive` stays under `routeLock` because the tap thread routes every keystroke by it. The icon-row slide stays on the adapter because it reads a layout measured from `NSScreen`; its selection writes now route through the session so it cannot drift.

**On the size.** +1,558 / −689, which the template flags as the range where things stall. I could not find a smaller honest split: a partly-extracted state machine, with the fields divided between two owners, is worse than either end state. The 806 new lines are one file that did not exist; the deletions are the same logic leaving `AppSwitcher` (1,708 → 1,320).

**Trade-off considered and rejected.** Folding the `SwitcherSupport` statics into the session physically. Four of the 20 it uses (`appGroups`, `usesIconRowLayout`, `usesWindowRow`, `windowNavigationDelta`) are also called by `WindowEnumerator`, `WindowActivator` and the views, so moving them would have broken those callers for no gain. The other 16 now have no callers outside `SwitcherSession`.

## Verification

MacBook (Mac17,3, Apple M5), macOS 26.6.2 (25G83), single built-in display, one Space. Release build via `./build.sh`, not a signed distribution build.

```
./build.sh                      exit 0, no warnings
./build/Vorssaint --selftest    SELFTEST OK
./build.sh --test               TESTS OK (9621 checks)   — main is 9526, so +95
```

The +95 is 98 new checks minus 3 removed. The 3 removed were a source-shape assertion that read `AppSwitcher.swift` as text and compared string offsets to prove the scope was assigned before the session-start layout pass. That hazard is now designed out rather than guarded: `SwitcherSessionEnvironment` answers the scope-dependent layout questions from the scope the session holds, instead of taking a pre-answered value.

The new checks drive sequences, not individual functions: open → Tab ×3 → wrap → held-Tab → Shift-Tab ×4; the Shift-back chord at three different inter-key delays plus release and re-press; presses claimed before the window list exists, folded, bounded at 200, and discarded on release; close → close-again → commit → confirm, and commit landing on the item being closed; app termination walking the selection; letter commands versus an in-progress search; grid arrows with 3 columns; a window-scoped session receiving the Apps shortcut.

I checked the tests can fail rather than assuming it: six deliberate regressions introduced into `SwitcherSession`, five turned the suite red across nine labels. The sixth — teardown leaking `isSearchPinned` into the next session — was not caught, so the reset assertion was widened to cover all 12 fields, then the regression was reverted and the suite re-run green.

**What I could not test.** No interactive run. Everything above is the harness and a release build; I did not hold Shift and Tab through a real session, and I did not exercise it across multiple displays, multiple Spaces, a fast user switch, or wake from sleep — all paths this code touches. One built-in display and one Space is the whole of the hardware coverage here. That gap is the useful part of this section: the sequences are asserted, the real input path is not.

Rebased onto `5a4afb15`, which includes #882 (system Command-Tab takeover). That PR changed the same file. I checked each symbol it introduced (`applyNativeHotkeySuppression`, `applyNativeHotkeySuppressionIfTapLive`, `reconcileTakeover`, `restoreNativeHotkeys`, `nativeHotkeysToSuppress`, `SwitcherNativeHotkeys.apply`, `SwitcherNativeHotkeys.configuredShortcuts`) still appears the same number of times after the rebase as before it.

## Anything else

Three defects were found while extracting and deliberately left alone, since this change is meant to preserve behaviour:

- `SwitcherSupport.commitTargetID`'s closing-window branch is unreachable from the app. `commitSession()` guards `closingItemIDs.isEmpty` before calling it, so the argument is provably empty at the only call site and the function always returns `itemIDs[selectedIndex]`. What its comment describes is actually delivered by `commitPendingForClose` deferring the commit. `MetricsTests` exercises the branch directly, which is why it reads as live.
- `userNavigated` has 7 writes and no reads, before and after.
- `KeyCode.tab` is unused.

One defensive change that is not behaviour-preserving in the strictest sense: the old `quitSelectedApp` did `windows[..<selectedIndex]`, which traps if `selectedIndex > windows.count`. I could not find a reachable path to it and clamped it in `applyTerminatedApp`. No reachable behaviour changes.

No issue reference — I did not find an existing issue for this.

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
+++ b/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
@@ -8,28 +8,6 @@ import Combine
 import CoreGraphics
 import SwiftUI
 
-private struct SwitcherSourceContext {
-    let itemID: String?
-    let pid: pid_t
-    let windowID: CGWindowID?
-    let windowOwnerPID: pid_t?
-    let isFullscreen: Bool
-    /// Window server bounds of the source window; `.zero` for an app-only entry.
-    let frame: CGRect
-}
-
-/// A shortcut press already owned by the switcher while the window list is
-/// being assembled. The tap thread can cancel or add repeated navigation
-/// without ever waiting for the main thread or Accessibility.
-private struct SwitcherPendingSessionStart {
-    let generation: UInt64
-    let shortcut: GlobalShortcut
-    let scope: SwitcherSessionScope
-    let reversed: Bool
-    var additionalNavigation = 0
-    var commitWhenReady = false
-}
-
 /// The window switcher: a global event tap takes over the configured shortcut,
 /// and while its modifiers are held a non-activating panel cycles through real
 /// windows. Releasing commits, middle-clicking a card or pressing W closes the
@@ -38,6 +16,13 @@ private struct SwitcherPendingSessionStart {
 /// needs the modifier held). Esc and a click outside cancel. The panel joins
 /// every Space and fullscreen app, so the switcher is available wherever the
 /// user is.
+///
+/// What a session *is* lives in `SwitcherSession`; this class is the adapter
+/// around it. It owns the event tap, the panel and the observers, turns real
+/// input into session events, and performs the effects the session hands back.
+/// The `@Published` properties below are the session's state projected for
+/// SwiftUI: they are written only while performing an effect, never as a
+/// decision of their own.
 final class AppSwitcher: ObservableObject {
     static let shared = AppSwitcher()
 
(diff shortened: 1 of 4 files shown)


PathGao:
Deliberately breaking the state machine six ways to prove the suite can go red, then widening the reset assertion when one slipped through, is what makes the rest of this believable.

Both `Swift 6.0.3 compatibility` and `Build & selftest` are red on this head. The body reports `./build.sh` exit 0, no warnings — that was your toolchain; CI disagrees on both.

The diff I got is truncated at 80KB, so `SwitcherSession.handleKeyDown` and `requestClose` are not in it. Two things to confirm from those:

1. `handle(type:)` used to open with `guard sessionActive else { return nil }` — a key claimed on the tap thread stays swallowed even when teardown wins the race to main. That guard and its comment are gone; the return is now `outcome.swallowsEvent`. If `.keyDown` yields `.ignored` while inactive, that Tab reaches the front app.

2. `closeWindow` used to bail on `item.windowID == nil` before touching `closingItemIDs`. That guard now sits in `beginClosingWindow`, downstream of the session. If `requestClose` inserts the ID first, W on a Finder app entry (no window) leaves an ID in `closingItemIDs` that nothing removes — `commitSession` then latches `commitPendingForClose` and the release never commits.

**Impact:** unproven; both reads are of a truncated diff. CI red blocks it regardless.
**Merge:** needs both build checks green first.

PathGao:
Closing. This is a pure structural move with no user-visible change, and it sits on the same file as #1236, which fixes a real stall. Not worth holding a +1557 refactor open in front of that. The branch stays on the fork; if the split is wanted later it comes back smaller.

PR #1266, opened 2026-09-01
fix(quick-tools): recompute share link expiry on wake

PathGao:
## Summary

**Share links do not notice they expired during sleep.**
`ScreenshotShareService.scheduleExpiryRefresh` arms the expiry with
`DispatchQueue.main.asyncAfter(deadline: .now() + …)`, and `DispatchTime.now()`
is a monotonic clock that does not advance while the Mac is asleep. A deadline
up to 24 hours out, crossed by an eight-hour sleep, fires eight hours late, and
until it does `records` keeps listing links the server has already dropped as
live.

Thirteen other services in the app already observe
`NSWorkspace.didWakeNotification` for exactly this, in the same words — "a fire
date that passed during sleep never triggers its timer; recomputing catches the
miss". This one now does too, calling the same `refresh()` that already
recomputes against the wall clock and re-arms the deadline. The service is a
`@MainActor` singleton with a `private init()`, so the observer lives as long
as the process and needs no teardown, which is why the thirteen beside it do
not remove theirs either.

`RecordingShareService` carries the same bug by copy. Its `scheduleExpiryRefresh` is the same lines word for word, armed on the same monotonic clock, and `ScreenRecorderSettings` likewise calls `sharing.refresh()` in `onAppear` at both places it shows the list. The one difference is the horizon: recording links run out after six hours (`RecordingSharingSupport` accepts an expiry no further than 6h 5m out) rather than 24, so a lunch-break sleep is enough to cross the deadline. The same nine lines now sit in its `init()` before the first `refresh()`, and the shape check reads both files.

The visible effect is small, and I would rather say so than oversell it:
`ScreenshotSettings` calls `sharing.refresh()` in `onAppear` at both places it
shows the list, so a stale list is only reachable when the Settings panel was
left open across a sleep.

This was one of two unrelated fixes on #1239, whose own summary opened by
calling them that. The other one removes default arguments from a shared
exclusion test and carries a volume UUID through five files; this one is nine
lines, in two files that half does not touch. They are apart so that a question
about the shared signature does not hold this one behind it. #1239 keeps the
volume-UUID half and is unchanged otherwise.

## Verification

macOS 26.x on Apple silicon. Branched from `origin/main` at `ea374e1e`.

- `./build.sh --test` after merging `main`: `TESTS OK (9600 checks)`, no failures.
  The sleeping-clock check now runs once per share service, two checks in all.
- Both assertions were confirmed red before being kept: reverting either
  share service to the version on `main` and re-running the built test binary
  gives `TESTS FAILED (1 of 9600)` naming that file's check, and restoring the
  observer turns it green again.
- Neither share service is in the file list `--test` compiles, which is why
  the check is a source-shape one rather than a behavioural test.
  `swiftc -typecheck` was run over the whole `Sources/Vorssaint` set with the
  target, SDK and compat flags `build.sh` uses (`arm64-apple-macosx14.0`, the
  pinned `MacOSX26.sdk`, `-interface-compiler-version 6.3.2`, both compat
  include paths): clean, no diagnostics.
- `./build.sh` (full release build) exits 0 on the merged branch.

**Not tested.** I did not run the app and I did not sleep the Mac to watch a
link expire, so the wake path is verified as wiring and as source shape, not as
behaviour. The check asserts the observer is registered; it does not assert
that `refresh()` on wake produces a different list than the one that was
showing.

## Anything else

No issue in the tracker describes this. It is one half of #1239, which is also
where its review history is; that pull request now carries only the volume-UUID
fix.


PathGao:
The diagnosis is right, and confirming the check red before keeping it is the step most source-shape assertions skip.

`RecordingShareService` is the same file. Same `expiryRefresh: DispatchWorkItem?`, same `scheduleExpiryRefresh(now:)` arming `.now() + expiration.timeIntervalSince(now)`, same `@MainActor` singleton with `private init()`, and `ScreenRecorderSettings` shows its `records` behind `.onAppear { sharing.refresh() }` at the same two spots. No wake observer. Its links cap at 6h rather than 24h, so the window is smaller, not absent.

Same nine lines in its `init()`, and run the check over both paths instead of one — otherwise the next reader fixes one twin and leaves the other.

Smaller, your call: `CleanerScheduler` and `WhatsAppDownloadScheduler` pair the wake observer with `NSSystemClockDidChange`, `AppUpdatesService` does not, so 2 of 3 and not a house rule. A clock jump strands this deadline the same way sleep does.

**Impact:** none new. Recording share links keep the bug this removes from screenshot links.
**Merge:** needs the `RecordingShareService` half first.

PathGao:
@PathGao `RecordingShareService` had the same bug by copy: its `scheduleExpiryRefresh` is the same lines on the same monotonic clock, so `cca0d793` puts the identical nine-line wake observer in its `init()` and the shape check now reads both files. Then merged `main` (`78ec2b97`, add/add at the tail of `MetricsTests.swift`, all three blocks kept). `./build.sh` exit 0, `./build.sh --test` TESTS OK (9600 checks); reverting either share service to `main` and rerunning the built binary gives TESTS FAILED (1 of 9600) naming that file, restoring it goes green. Body updated.

PathGao:
Second half landed, and the shape check loops both paths, so it cannot go green with one twin fixed. That was the gap.

Coverage holds: on `main`, `expiryRefresh` exists in exactly these two files. There is no third share service to miss.

One sentence in the body is wrong. "the thirteen beside it do not remove theirs either" — most of them do. `SuperKeyService`, `AppSwitcher`, `AppUpdatesService`, `CleanerScheduler` and `AppVolumeMixer` all call `removeObserver` when their feature is switched off; they hold the token for uninstall, not for dealloc. The conclusion still stands — nothing switches a share service off, and `AssistiveKeyboard.bootstrap` is the existing precedent for discarding the token. Only the sentence needs fixing.

Not verified: I read source, not behaviour, and did not sleep a Mac either. `Swift 6.0.3 compatibility` and `Build & selftest` had not reported when I read this.

**Impact:** none new.
**Merge:** ship it.

PR #1267, opened 2026-09-01
fix(window-layout): keep the directional gesture's feedback running under a held button

PathGao:
## Summary

The directional gesture's indicator and snap preview freeze while a mouse
button is held.

`beginDirectionalGesture` drives both from a 60 Hz timer created with
`Timer.scheduledTimer`, which schedules only in the default run loop mode. The
gesture's own tap takes `leftMouseDown` and `rightMouseDown`, so pressing a
button during the gesture moves the main run loop into event tracking,
`updateDirectionalGesture()` stops firing, and the indicator and the preview
hold the frame from before the press until the button comes back up.

`RunLoop.main.add(timer, forMode: .common)` is what `scheduleSettleCheck` in
this same service already does. The feedback timer now does it too. No
tolerance was added — slack in a 60 Hz feedback timer is the thing being
fixed.

`WindowLayoutService.swift` is outside the `--test` whitelist, so a source
check pins the mode. It is sliced to the gesture's own start, between
`showDirectionalIndicator` and `startDirectionalTap`, so the settle timer that
has been in common modes all along cannot answer for it.

Split out of #1233, which had it as a rider under `### Also in this file`.
That change is about handing the window taps back across a fast user switch;
this one is an independent UI defect visible without any second account, and
none of that branch's assertions reached it.

## Verification

MacBook Pro, macOS 26.

```
./build.sh --test      TESTS OK (9534 checks)
```

`--test` compiles a whitelist that does not include `WindowLayoutService.swift`,

The new check was verified both ways. Reverting the timer to
`Timer.scheduledTimer` turns it red on its own; restored, the suite is green.

**What was not tested.** The freeze itself was not reproduced on screen, and
the fix was not watched working. What is verified is the mechanism — a
default-mode timer does not fire during event tracking, the tap that starts
that tracking is this gesture's own, and the sibling timer in the same file
already answers it this way.

## Anything else

Refs #1153

### DIFF
--- a/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
+++ b/Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
@@ -732,9 +732,16 @@ final class WindowLayoutService: ObservableObject {
                                                       action: nil,
                                                       manualOverride: nil)
         showDirectionalIndicator(at: NSEvent.mouseLocation, action: nil)
-        directionalTimer = Timer.scheduledTimer(withTimeInterval: 1.0 / 60.0, repeats: true) {
+        // The gesture's own tap takes leftMouseDown and rightMouseDown, so a
+        // held mouse button puts the main run loop into event tracking and a
+        // default-mode timer stops firing: the indicator and the snap preview
+        // hold the frame from before the press. The settle timer in this same
+        // service is already in common modes; the feedback timer now is too.
+        let timer = Timer(timeInterval: 1.0 / 60.0, repeats: true) {
             [weak self] _ in self?.updateDirectionalGesture()
         }
+        directionalTimer = timer
+        RunLoop.main.add(timer, forMode: .common)
         startDirectionalTap()
     }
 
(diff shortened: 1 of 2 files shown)


PathGao:
The slice is pinned correctly — between `showDirectionalIndicator` and `startDirectionalTap`, both separators unique in the file, so the settle timer at line 459 can't answer for it. Checked.

The mechanism doesn't hold. `observeDirectionalEvent` returns `nil` for `.leftMouseDown` and `.rightMouseDown`, on a `.cgSessionEventTap` / `.headInsertEventTap` / `.defaultTap` — the press is discarded before any app sees it, this one included. Nothing receives the mouse-down, so AppKit never enters event tracking and the cause the body names looks unreachable. Combined with the freeze never being reproduced, this may fix nothing, and the assertion message claims what the same function contradicts. Common modes is still the right mode for a 60 Hz UI timer.

Separately, not this PR: `ShelfService.startDockedWatchdog` polls `CGEventSource.buttonState` on a default-mode timer, and `ShelfTilesView` does call `beginDraggingSession` — a real drag source. I did not confirm the watchdog is armed during an internal drag.

**Impact:** none if merged; common modes is a superset of default.
**Merge:** needs the mechanism reconciled first — a repro, or the real cause in the body.

PathGao:
Closing. The mechanism in the body does not hold: the gesture tap returns nil for the button-down, so the event never reaches AppKit and the run loop never enters tracking mode while the timer is alive. The symptom was never reproduced either. The .common change is harmless but there is nothing for it to fix.

Issue #780, opened 2026-08-22
Vorssaint windows don't appear in screen captures


AB-boi:
### Description

In screenshots/recordings, vorrsaint windows don't appear despite disabling the hide vorssaint windows option.

### Steps to Reproduce

disable "hide vorssaint windows in screenshots" option
try screenshotting vorssaint's settings/screenshot editor windows → you can't

### Vorssaint Version

3.3.2

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

AB-boi:
### I would like to add a clarification:
Vorssaint's settings window is captureable but other windows like screenshot and screen-recording windows are not.

PR #1269, opened 2026-09-01
fix(screenshot): let the visibility preference reach the content windows

dlowzzxx:
Refs #780.

## Problem

`ScreenshotService.protectedWindowIDs` unioned every window the app puts on screen — selection overlays, the countdown and scrolling HUDs, editors, pins, the quick preview and the capture panels — before the "Hide Vorssaint windows" preference was ever consulted. The capture paths themselves honour the preference (`ScreenshotCapturePolicy.excludedWindowIDs` excludes everything when it is on, and otherwise only the protected list), so with the preference off the protected list re-excluded the editors, pins and panels anyway: the switch could never reach them. That matches the clarification on the issue — the Settings window, which was never in the protected list, was already capturable, while the screenshot and screen-recording windows were not.

## Change

The protected list now splits the app's own windows by role:

- **Workflow surfaces** — the screenshot selection overlay, the shared capture tool's selection overlay, the countdown HUD and the scrolling HUD — stay excluded whatever the preference says: they are the tool taking the capture, never its subject.
- **Content windows** — editors, pins, the quick preview — join the protected list only when "Hide Vorssaint windows" is on; with it off they stay capturable.

The decision lives in `ScreenshotCapturePolicy.protectedWindowIDs(hideVorssaintWindows:workflowWindowIDs:contentWindowIDs:)` beside the other capture-policy decisions, so `./build.sh --test` reaches it; `ScreenshotService.protectedWindowIDs` assembles the two sets and calls it.

The screen recorder builds its chrome from the same list (`ScreenRecorderService`), so its behaviour follows the preference too: preference on, the recording excludes the whole app as before; preference off, the capture panels stop being silently erased from recordings as well.

## Tests

Three behaviour tests against the new policy function: preference off keeps only the workflow excluded; preference on excludes the whole app; nothing on screen yields an empty list. Reverting the split to an unconditional union turns the first one red.

## What I could not test

This branch was written on a Windows machine, so it was not compiled or run here — not `./build.sh`, not `--selftest`, not `--test`. The whole path was read instead: both capture engines, the scrolling capture, the selection controller, the recorder's chrome assembly and the window picker. Two things most need a real Mac: that the migrated HUD and overlay window numbers resolve identically through `SCShareableContent` (the mechanics are unchanged, but I cannot run it), and a direction call on the recorder side — with the preference off, an open editor or a pinned capture will now appear in recordings, which is what the preference says but is a slightly wider effect than the issue's screenshots-only framing.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotCapturePolicy.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotCapturePolicy.swift
@@ -13,6 +13,19 @@ enum ScreenshotCapturePolicy {
         hideVorssaintWindows ? ownWindowIDs : ownWindowIDs.intersection(protectedWindowIDs)
     }
 
+    /// The own windows the capture paths keep out of the shot before the
+    /// visibility preference says anything (issue #780). The workflow
+    /// surfaces — the selection overlays and the countdown and scrolling
+    /// HUDs — are always excluded: they are the tool taking the capture,
+    /// never its subject. Content windows (editors, pins, the quick preview)
+    /// follow the preference, so with "Hide Vorssaint windows" off they stay
+    /// capturable instead of silently joining the protected list.
+    static func protectedWindowIDs(hideVorssaintWindows: Bool,
+                                   workflowWindowIDs: Set<CGWindowID>,
+                                   contentWindowIDs: Set<CGWindowID>) -> Set<CGWindowID> {
+        hideVorssaintWindows ? workflowWindowIDs.union(contentWindowIDs) : workflowWindowIDs
+    }
+
     static func canPickWindow(_ windowID: CGWindowID,
                               isOwnWindow: Bool,
                               hideVorssaintWindows: Bool,
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
@@ -44,20 +44,30 @@ final class ScreenshotService: ObservableObject {
     }
 
     private var protectedWindowIDs: Set<CGWindowID> {
-        var ids = session?.protectedWindowIDs ?? []
-        ids.formUnion(preview?.protectedWindowIDs ?? [])
-        for editor in editors {
-            ids.formUnion(editor.protectedWindowIDs)
-        }
-        ids.formUnion(ScreenshotPinController.shared.protectedWindowIDs)
-        ids.formUnion(ScreenCaptureService.shared.protectedWindowIDs)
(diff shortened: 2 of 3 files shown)


PathGao:
The split by role is the right cut, and putting the decision in `ScreenshotCapturePolicy` is what makes it testable. I checked the four producers on main — editor, pin, quick preview, selection panels — each returns only its own window, so both buckets are faithful.

`ScreenshotSupport.unifiedCapturePolicy` already answers your recorder question the other way: `hideVorssaintWindows: tool != .recording && screenshotHideVorssaintWindows`. Recording is deliberately exempt from this preference there, and the toggle lives in the Screenshot settings pane. This PR routes it into the recording stream anyway through `protectedWindowIDsForCapture`. That needs the maintainer's call; if the answer is "not recordings", the recorder needs its own accessor that unions both sets.

Second one: the quick preview is in the content bucket, but it's a 12s auto-dismissing toast, not a window anyone parked. With the preference off, back-to-back captures will catch the previous capture's preview panel. That reads as workflow to me.

I read main's source only — no build, no run on my side either.

**Impact:** only with "Hide Vorssaint windows" off (default is on): editors, pins and the preview now appear in screenshots, and in screen recordings.
**Merge:** hold for the maintainer's call on whether the screenshot preference should govern recordings.

ref #780

PathGao:
Sending this back until it has been built and run on a Mac.

Your own description: "This branch was written on a Windows machine, so it was not compiled or run here — not `./build.sh`, not `--selftest`, not `--test`." This is a macOS app; every branch is run on the maintainer's Mac before it merges, and CI on a fork needs approval before it runs. A change to which of the app's own windows get captured, submitted by someone who could not open the app, is a review nobody here can do by reading. Please do not open pull requests against this repository from a machine that cannot build it.

The change itself may be right, and it was read carefully. But it also widens: `protectedWindowIDsForCapture` is what the recorder reads (`ScreenCaptureService.swift:192`, `ScreenRecorderService.swift:442`), and main keeps recording out of the screenshot preference on purpose (`ScreenshotSupport.swift:124`, `tool != .recording`). With the preference off, editors and pinned captures start appearing in recordings; you noted that yourself as "a slightly wider effect". That is a direction question for the maintainer, and it has to be seen on screen, not deduced.

When you have a Mac: build it, run the three commands, record the model and macOS build, try a recording with the preference off, and push. The review picks up from there.

PathGao:
Closing, per the comment above. Redo this from a Mac: build, run, record the model and macOS build, and open a fresh pull request with that in the description. The direction question on the recorder side stays for the maintainer; #780 remains open.

PR #1270, opened 2026-09-01
fix(radial-menu): hand the side-button tap back across a user switch

PathGao:
## Summary

The wheel's mouse tap is a `.defaultTap` on `otherMouseDown` and
`otherMouseUp`, created while any profile is bound to a mouse button. It did
not follow the login session.

the screen, and a filter tap created there keeps its place in the chain. The
window server still routes every extra-button click the account on screen
makes through a process that cannot answer, and waits out the tap timeout on
each one.

Both places that can put the tap into the chain — `syncMouseTap` and the
`tapDisabledByTimeout` re-arm — now ask `SessionActivitySupport.tapShouldRun`,
and `SessionActivity.shared.onChange` brings the tap back with the session.
The callback runs on the main run loop, so the session flag is read on the
thread that writes it, the same way `MouseNavigationService` and the other
main-run-loop taps do; the taps that run a thread of their own hop to the main
queue for the same answer.

Two things travel with the gate because it does not work without them:

- **The open wheel.** This tap is the only thing that sees a held summoner
  released, so a wheel still up on resign would come back stuck in hold phase
  with no release coming for it. The resign handler calls `endSession()`
  before the sync tears the tap down.
- **The feature check.** `syncWithPreferences` guards on
  `AppFeature.radialMenu.isAvailable && radialMenuEnabled` and calls
  `suspend()`, but the resign handler and the capture row call `syncMouseTap`
  directly and never pass that guard. Without it the tap would come back on a
  session switch with the feature switched off.

Accessibility also moves into that gate from the tap-creation guard, and is
asked of the system rather than of `Permissions.shared`, whose answer is a
poll up to `PermissionPollingSupport.interval` old. The re-arm hands a tap it
declines to put back to the sync to be stopped, so a stale grant there would
install it straight back. That matches what the keyboard, Finder, switcher,
snippet, volume, Dock and window taps ask in the same position.

`RadialMenuService.swift` joins the existing tap-owner sweep in
`MetricsTests.swift` rather than carrying its own copy of those checks. One
assertion is its own, because the sweep has no reason to know it: the resign
path has to end the session, not only the tap.

Split out of #1228, which had this and the cheap side-button decode in one
branch. They reference different issues, share no premise, and this half
carries a **Not tested** note the other does not.

## Verification

MacBook Pro, macOS 26.

```
./build.sh --test      TESTS OK (9543 checks)
```

`--test` compiles a whitelist that does not include `RadialMenuService.swift`,

The resign assertion was verified both ways. Dropping `endSession()` from the
`onChange` handler turns it red on its own; restored, the suite is green.

**What was not tested.**

- **The behaviour itself. There is no second logged-in account on this
  machine**, so the fast user switch this change is about was never performed.
  What is verified is that the wiring matches the services that already carry
  it.
- The stuck-hold-phase case. Reaching it needs a wheel held open by a mouse
  button at the moment the session resigns.
- The revoke race. Reaching it needs the window server to disable this tap
  inside the polling interval that follows a revoked Accessibility grant.

## Anything else

Refs #1153

### DIFF
--- a/Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
+++ b/Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift
@@ -72,7 +72,19 @@ final class RadialMenuService: ObservableObject {
     private var activationObserver: NSObjectProtocol?
     private var promptedForAccessibility = false
 
-    private init() {}
+    private init() {
+        // A filter tap owned by a switched-away login session keeps its place
+        // in the chain, so the account on screen waits out this tap's timeout
+        // on every extra-button click it makes (issue #1075). Hand the tap
+        // back on resign and build it again from preferences on the way in.
+        // The open wheel goes with it: this tap is the only thing that sees a
+        // held summoner released, so a wheel still up when the tap is handed
+        // back would come back stuck in hold phase with no release coming.
+        SessionActivity.shared.onChange { [weak self] active in
+            if !active { self?.endSession() }
+            self?.syncMouseTap()
+        }
+    }
 
     var sessionActive: Bool { !stack.isEmpty }
 
@@ -128,7 +140,13 @@ final class RadialMenuService: ObservableObject {
     // app under the pointer; alive only while a button is configured, torn
     // down with the feature)
 
-    private func syncMouseTap(profiles: [RadialMenuProfile]? = nil) {
+    /// Whether the button tap has a job right now. The feature check is here
+    /// rather than only in `syncWithPreferences`, because the resign handler
+    /// and the capture row call the sync directly and so never pass that
+    /// guard. The capture row wants the tap even with the feature switched
+    /// off, so it can tell a button this app cannot see from one that is
+    /// simply set to something else.
+    private func mouseTapWanted(profiles: [RadialMenuProfile]?) -> Bool {
         let defaults = UserDefaults.standard
         // Asked of the stored buttons alone when the caller has no profiles in
         // hand: this only needs to know whether any wheel is bound to a button,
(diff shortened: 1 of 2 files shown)


PathGao:
The resign path ending the session before the tap goes is the easy half to miss, and that assertion is mutation-tested both ways.

**Cleaning Mode brings the tap back.** `CleaningModeManager` calls `suspend()` on this service precisely so a side-button press can't open the wheel over the overlay, but `suspend()` leaves no state behind. A resign+activate while it is up hits the new `onChange` → `syncMouseTap`, and `mouseTapWanted` still sees availability + `radialMenuEnabled` + a bound button, so the tap comes back. `MouseNavigationService` has the same hole today (`suspend()` is `stop()`, `onChange` calls `syncWithPreferences()`), so it doesn't start here and I wouldn't hold the merge for it — but it is your own feature-check argument one step further, and the tap-owner sweep cannot see it. Worth its own issue.

**Most of this diff is already on main**: `claimedMouseButtons`, `RadialMenuProfileButton`, the deferred decode in `handleMouseTap`, the `MouseNavigationService` comment. Net new is the gate and the feature check, and `Depends on #1228` looks discharged.

`mouseTapWanted` reassembles availability + enabled + bound-button by hand; `RadialMenuSupport.claimsMouseButton` already fuses those three. A `claimsAnyMouseButton(defaults:)` plus dropping the `profiles` parameter leaves one answer where there are now two that can disagree on a corrupt blob.

Read against `main`, not run. No second account on my side either.

**Impact:** Cleaning Mode plus a session switch re-arms the summoner; nothing else.
**Merge:** ship it.

ref #1153

PathGao:
@PathGao Rebased onto main now that #1228 is in; the five commits collapse into one. Both conflicts (the tapShouldRun gate in RadialMenuService.swift and the resign assertion in MetricsTests) resolved to the branch side, so the diff against main is the session gate alone: RadialMenuService.swift +53/-5, MetricsTests +15/-2. Full ./build.sh exit 0, TESTS OK (9604 checks). Dropped the Depends on #1228 line from the body.

PathGao:
Rebase checks out. Against main the diff is the gate alone, and the sweep assertion isn't passing vacuously: it splits on `tapDisabledByTimeout` and takes up to the first `return`, and the new block puts `SessionActivity.shared.isActive` inside that span.

Two things.

The body's Verification section didn't come along. It still says the full `./build.sh` was not run, cites 9543 checks, and spends a paragraph explaining why `Sources` was type-checked separately. Exit 0 and 9604 belong there — the body is what gets read at merge time and after.

`MouseClickDebounceService.swift` is a `.defaultTap` owner with `onChange`, an `isActive` re-arm and `CFMachPortInvalidate`, and it is not in the tap-owner list you just added to. It passes every assertion in that loop today; adding the string is what keeps it passing. Not this PR's bug and I wouldn't hold for it, but you're one line away in a file you're already touching.

**Impact:** none — the code is what I reviewed last round.
**Merge:** ship it.

ref #1153

Issue #1258, opened 2026-09-01
Text snippets: place the caret inside expanded text


caikovsky:
### Problem or Motivation

Text snippets always leave the insertion caret at the end of the expanded text. When a reusable message, command, or code fragment needs editing in the middle, the user must move back with the mouse or arrow keys after every expansion.

For example, a greeting template is useful only if the caret can land where the recipient name belongs instead of after the signature.

### Proposed Solution

Recognize one `{cursor}` marker in a snippet replacement. Omit the marker from the expanded text and leave the caret at that position after insertion.

```text
Hello {cursor},

Thanks for reaching out.
```

The behavior should be the same for trigger expansion and the searchable snippet library, including single-line and multiline snippets. Snippets without the marker should remain unchanged.

Keep the scope to one final caret position: no tab stops, selections, or multiple-field navigation.

This fits the existing snippet path: variable expansion is centralized, both entry points share the insertion routine, and multiline paste already exposes a completion callback. It needs no dependency, new permission, setting, or subsystem; it can use the existing Accessibility-backed synthetic key events.

### Feature Area

Mouse and keyboard

### Does another app already do this?

Alfred uses `{cursor}` for this behavior.

### Mockups / References

https://www.alfredapp.com/help/workflows/advanced/placeholders/#cursor

I searched open and closed issues and pull requests for snippet cursor placement before filing this; none covered it.

caikovsky:
@PathGao would it be okay if I propose a PR for this idea? I already have local working version.

PathGao:
Yes, open it. I can't merge or approve — that's vorssaint's call — but a PR is the right way to put it in front of him.

Three things I'd want to see in it, from reading the snippet path:

Use `{{cursor}}`, not `{cursor}`. Every existing token in `TextSnippetSupport.expand` is double-braced; a second brace syntax in the same string is one more shape to remember.

Locate and strip the marker *before* the `{{clipboard}}` substitution. `expand` already puts clipboard last "so pasted text is never re-expanded" — same reason applies here, otherwise a clipboard that happens to contain the marker moves the caret.

Count the left-arrows in `Character`s, not UTF-16 units — arrows move by grapheme, so an emoji after the marker overshoots. The count also has to cover whatever `trailingText` / `trailingKeyCode` adds after the insertion, and on the multiline path it has to fire after `TransientPaste` reports done, not alongside it.

[...]

PR #1275, opened 2026-09-02
feat(snippets): place cursor inside expanded text

caikovsky:
## Summary

Text snippets currently leave the caret at the end of every expansion. This proposal adds one Alfred-compatible {{cursor}} token that is omitted from the inserted text and leaves the caret at that position.

The change stays in the existing snippet path. TextSnippetSupport removes the source marker before dynamic-variable expansion, TextSnippetService applies the caret retreat after typed insertion or transient paste, and SnippetLibraryService reuses the same postExpansion routine. Clipboard text containing {{cursor}} remains literal, Character counting handles composed emoji, and snippets without the marker are unchanged.

For an after-delimiter snippet containing {{cursor}}, the triggering Space, Tab, or Return is not replayed. This prevents the delimiter from adding whitespace, changing focus, or sending a message before the caret reaches its requested position. Markerless snippets retain their existing delimiter behavior.

TransientPaste waits 150 ms after the paste key-up before running completion callbacks, so follow-up caret events do not overtake editors that consume paste asynchronously. Both insertion callers and every postExpansion caller were checked. The Command Bar keeps the existing no-retreat default. The proposal adds no dependency, runtime service, setting, permission, or new subsystem. The editor hint is present in all 13 supported languages.

## Verification

- MacBook Pro MacBookPro18,2 with Apple M1 Max, macOS 26.6.2 build 25G83, locally signed developer build.
- Correct single-line placement was manually verified in the GitHub web comment editor with only Vorssaint Developer running. Running the official and developer builds together initially caused their global event taps to compete.
- Clipboard text containing a literal {{cursor}} and composed emoji placement were manually verified.
- The latest delimiter-suppression change still needs manual testing with Space, Tab, and Return.
- Multiline transient paste still needs real-app testing in Slack, VS Code, and Discord, including while autocomplete is open. App language and display and Spaces layout were not recorded.
- ./build.sh completed without warnings and produced build/stage/Vorssaint.app. ./build/Vorssaint --selftest returned SELFTEST OK. ./build.sh --test returned TESTS OK with 9627 checks. git diff --check is clean.
- Tests cover marker placement, composed emoji Character counting, literal clipboard markers, literal single-braced text, unchanged no-marker snippets, delimiter suppression policy, every localized editor hint, and delayed transient-paste completion.

## Anything else

This follows Alfred's established cursor placeholder behavior: https://www.alfredapp.com/help/workflows/advanced/placeholders/#cursor

Terminal limitation: caret retreat uses Swift Character counts. Terminal editors can move by scalar or cell units, so placement after composed emoji may require additional movement. This proposal does not normalize terminal cursor semantics.

### Demo

<video src="https://github.com/user-attachments/assets/e048b918-ae58-489f-8820-6a0f0354e3b9" />

Refs #1258


PathGao:
Splitting the marker before expansion is the right call — it kills the "clipboard content gets reinterpreted as a marker" class outright. I grepped `postExpansion` on main: three callers, and the Command Bar one only ever gets emoji and transform text, never a snippet replacement, so the no-retreat default there is correct.

Two things:

1. In the paste branch `didPostShortcut` fires right after the ⌘V key-up, so the arrows go out with nothing waiting on the app having inserted. Apps that handle paste off the main thread (Slack, VS Code, Discord) can take the arrows first and leave the caret before the pasted text. Multiline is exactly the branch `{{cursor}}` is for. Which apps did the multiline case run in?

2. `caretRetreat` is a Character count, so it only lands right where the target moves the caret by grapheme cluster. Terminals don't — the emoji case needs more presses there. Worth stating as a limit rather than fixing.

Not run on hardware; read off the diff and main.

**Impact:** marker snippets can land the caret before the pasted text in apps with async paste. The typed single-line path is unaffected.
**Merge:** needs the multiline paste targets named first — your Verification section says the same.

ref #1258

caikovsky:
@PathGao I noticed in my testing that when pressing any expand key because of current behaviors, it feels kind of odd. On the other hand, the "right away" works flawlessly. 

**Example of current behavior (space, tab and return in sequence):**
<video src="https://github.com/user-attachments/assets/bf0f0800-ff58-4acc-8541-159ee43f5a3e" /> 


**Example of PR behavior (space, tab and return in sequence): *
<video src="https://github.com/user-attachments/assets/d7a4535d-1aa6-4f68-8fe4-52141ccf2dc2" />

In a sense, using this feature and expanding with enter key kinda defeats its purpose, as you can see above. Maybe we could limit this to be supported only on `right away` expansion? Or change the behavior of expand keys to be overwritten by this feature instead, not interacting with other elements: spacebar not applying white space, not changing focus with tab or sending message with enter.

**Example with clipboard and emoji:**
<video src="https://github.com/user-attachments/assets/95114514-cfcc-4292-b63a-a6647fe10e0b" />

PathGao:
Option 2, and it's cheaper than it looks. The delimiter never reaches the app — the tap swallows it and replays it afterwards, as `trailingKeyCode` in the typed path or inside `trailingText` in the paste. So when the insertion carries a caret position, just don't replay it:

```swift
let carries = insertion.caretRetreat != nil
return Self.postExpansion(deleteCount: deleteCount,
                          text: insertion.text,
                          caretRetreat: insertion.caretRetreat,
                          trailingKeyCode: carries ? nil : trailingKeyCode,
                          trailingFlags: trailingFlags,
                          trailingText: carries ? "" : trailingText,
                          ...)
```

Tab and Return never fire, so there's nothing to undo. Snippets without the marker go down the same path they do today — no setting, no new branch.

Option 1 is worse: `{{cursor}}` in an after-delimiter snippet would silently do nothing while the editor hint says it works.

I can't watch the videos; this is off your description and the code.

**Impact:** with the guard, an after-delimiter snippet that places the caret loses its trailing delimiter. That's the trade you're asking for.
**Merge:** needs this, plus the multiline paste targets still open from last round.

ref #1258

caikovsky:
Ooh, that's a great suggestion. I will take a look at that when I'm available.

You can't watch the videos because of a something on your side or I need to do something else so I'm able to share them with you?

PathGao:
My side. I only get the text of the PR and the repo, not video attachments — nothing for you to change about how you're sharing them, anyone else on the thread can watch them fine.

Text works. For the clipboard/emoji one: which app, what the snippet text was, and where the caret ended up versus where you wanted it. That's enough to tell whether it's the Character-count limit I mentioned or something else.

caikovsky:
So I took more time to test this extensively and found out a couple of stuff. I tested mostly with Slack.

Tested the Return regression test-first. The cause was an unmatched physical delimiter key-up: the event tap suppressed Return key-down, inserted the replacement, then allowed the physical Return key-up to reach Slack and submit/clear the composer.

The fix consumes exactly one matching physical key-up after a successful after-delimiter cursor expansion. Unrelated, synthetic, and no-pending key-ups pass through. Without any `{{tag}}`, immediate, library-insertion, and failed-expansion behavior remains unchanged.

Manual Slack testing confirmed Return no longer submits/clears the replacement on the first attempt. Stress testing with Tab and Space also exposed intermittent blank insertion with markerless `;t0`snippet, which does not arm the pending key-up state. Event-source, character-sized injection, and outside-event-tap probes did not eliminate it and were reverted. I found out about this possibly being a bigger issue than something localized for this PR only and it points to the broader synthetic-input surface tracked by [#937](https://github.com/vorssaintapp/vorssaint-utils/issues/937). I feel this cannot be fixed on this feature PR. WDYT?

Automated verification: `./build.sh`, `./build/Vorssaint --selftest`, `./build.sh --test` (`9940` checks), and `git diff --check` all pass.

The snippets I used for testing:

- **test1 -** `;t0`
    - Expands: After Space, Tab, or Return
    - Content: `This is a test snippet.`
- **test2** - `;t1`
    - Expands: Right away
    - Content: `Hi {{cursor}}, This is a test.`
- **test3** - `;t2`
    - Expands: After Space, Tab, or Return
    - Content: `Hello {{cursor}}world`
- **test4 -** `;t3`
    - Expands: Right away
    - Content: `Start {{cursor}}{{clipboard}} end`

AllanStepps:
Read this off the diff and the source at `7c15bb5`; not run on hardware.

The design is right. Splitting the marker before variable expansion is what makes the clipboard case safe rather than lucky, and suppressing the triggering delimiter is a better answer than gating the marker to "right away" — `{{cursor}}` silently doing nothing in an after-delimiter snippet would have been the worse bug. Six things, one of which I would not merge without.

**1. The pending key-up arm never expires.** `pendingConsumedDelimiterKeyCode` is written in exactly three places: set at `TextSnippetService.swift:393`, cleared on a matching key-up at `:262`, cleared in `stop()` at `:134`. `resetBuffer()` (`:226`) does not touch it, and every other reset path goes through `resetBuffer()` — app switch, mouse down, secure input, Cmd/Ctrl, arrow keys.

The usual case is fine: the arm lives the few milliseconds until the physical release. The case that isn't: if that key-up never reaches the tap (tap disabled by timeout while the `DispatchQueue.main.sync` at `:401` waits on a busy main thread, feature toggled, session change), the arm survives and the *next* Return the user presses loses its key-up. In Slack, by your own finding, that Return does nothing and the one after it works. That is the same "intermittent, once in a while" shape you were chasing, and in a bug report it would be indistinguishable from #937.

Cheapest fix that keeps the function pure and testable: stamp the arm with the triggering event's timestamp, have `pendingKeyUpDecision` take `now`, expire anything older than about a second. Clearing it in `resetBuffer()` as well costs nothing.

**2. The 150 ms settle also delays `completion`, and PastePlain pays for it.** In `TransientPaste.postPasteShortcut` the delay wraps `completion(true)`, and the caller at `TransientPaste.swift:132-143` runs both `didPost?()` and `completion()` from it. `completion()` is what sets `isPerforming = false` and schedules the restore (`:79-80`). So `PastePlainService.swift:71` now holds `isPerforming` 150 ms longer and restores the clipboard 150 ms later; a second ⌘⇧V inside that window returns `false` at `:28` and does nothing.

Delaying `didPost` is what the caret retreat needs. Delaying `completion` is collateral. Splitting them leaves PastePlain on its old schedule.

**3. `caretRetreat + trailingText.count` at `TextSnippetService.swift:430` is dead.** Every caller that passes a non-nil `caretRetreat` now passes `trailingText: ""` (`:387`), and `SnippetLibraryService.swift:214` and `CommandBarExtras.swift:30` take the default. It reads like a live path. If it stays, it is also wrong for a `trailingText` holding a multi-scalar grapheme.

**4. `{{cursor}}` at the very end gives `caretRetreat == 0`,** and `keepsTriggeringDelimiter` returns false for it (asserted at `MetricsTests.swift:16522-16525`). An after-delimiter snippet ending in the marker therefore loses its space, tab or return and gains nothing in caret position. Deliberate or not, nothing the user reads says so.

**5. A second `{{cursor}}` survives into the inserted text.** `expandedInsertion` takes the first range (`TextSnippetSupport.swift:142`) and `expand` does not know the token, so `a{{cursor}}b{{cursor}}c` inserts `ab{{cursor}}c`. The caret lands correctly; the text has junk in it.

**6. Discoverability.** `variablesHint` is the line people read as the list of variables, and it renders in both places (`TextSnippetsSettings.swift:83` and `:328`). It does not list `{{cursor}}`. The marker appears only at the tail of `editorFormatCaption`, which is a sentence about date-format patterns, and the Features footer's `variablesFormatCaption` (`:85`) does not mention it at all. Separately, #1258 asks for `{cursor}` and this ships `{{cursor}}`. Double braces is the right call for consistency with the other variables, and you test that `{cursor}` stays literal, but someone following the issue verbatim gets literal text with no clue why — worth a line on the issue.

Agree that the markerless intermittent blank insertion belongs to #937 and not here.

I want this feature. Happy to test the multiline paste targets on a MacBook Pro M1 Max / macOS 26.6.2 if that is the last thing holding it, since it is the one open item from PathGao's first round.

Refs #1258

AllanStepps:
Came here to build this and found it already done. Split-before-expansion is the right call, that's what makes the clipboard case safe instead of lucky.

Off the diff at `7c15bb5`, haven't run it. One thing I'd fix, then nits.

**The key-up arm never expires.** Set at `TextSnippetService.swift:393`, cleared only by a matching key-up or `stop()`. `resetBuffer()` doesn't touch it, no timeout. Usually fine, it lives a few ms until you let go. But if that key-up never reaches the tap (tap timed out while `main.sync` at `:401` waits on a busy main thread, feature toggled, session change) it stays armed and eats your *next* Return. In Slack: one Return does nothing, the next one works. Exactly the intermittent thing you were chasing, and indistinguishable from #937 in a report. Timestamp it, pass `now` into `pendingKeyUpDecision`, expire after a second.

**The 150 ms settle leaks into PastePlain.** It wraps `completion(true)`, and `TransientPaste.swift:132-143` fires `didPost?()` and `completion()` off the same closure. `completion()` clears `isPerforming`, so `PastePlainService.swift:71` holds its lock 150 ms longer and a second ⌘⇧V in that window does nothing. Only `didPost` needs the delay.

Nits:

- `caretRetreat + trailingText.count` at `:430` is dead now.
- `{{cursor}}` at the end gives retreat 0 and still kills the delimiter: eats your space, moves the caret nowhere.
- A second `{{cursor}}` comes through literally, `a{{cursor}}b{{cursor}}c` → `ab{{cursor}}c`.
- `variablesHint` is what people read as the variable list and `{{cursor}}` isn't in it. #1258 also says `{cursor}`; yours is better, worth a note there.

#937 isn't yours to fix here. Happy to run the multiline paste targets if that's the last blocker: MacBook Pro 14" M2 Pro, macOS 26.6 (25G5028f), different chip and build from yours.

Refs #1258

caikovsky:
Implemented the two behavior issues in `5b259ff`.

- Pending delimiter key-up state now uses monotonic uptime and expires after one second. A timely match is consumed once; unrelated key-ups retain fresh state; expired matches pass through and clear it.
- The 150 ms settle now applies only to multiline cursor caret movement. Shared `TransientPaste` completion happens immediately again, so Paste Plain keeps its previous timing.

Both changes were added test-first. I also replaced the private source-shape assertion with behavioral coverage.

Verification on the current PR head:

- `./build.sh`
- `./build/Vorssaint --selftest` → `SELFTEST OK`
- `./build.sh --test` → `10772` checks plus preference-cleanup tests
- `git diff --check`

For the remaining notes: cursor-at-end suppressing the delimiter, including retreat `0`, is intentional; only the first `{{cursor}}` is supported, so later markers remain literal. `{{cursor}}` intentionally uses double braces to match the existing snippet tokens.

The multiline real-app gate remains open. @AllanStepps if your offer still stands, the current head is ready for testing, including with autocomplete open.

What do you think about current state @PathGao?

Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
```diff
@@ -395,6 +429,12 @@ final class TextSnippetService {
                 post(event)
             }
         }
+        func postCaretRetreat() {
+            guard let caretRetreat else { return }
+            for _ in 0..<(caretRetreat + trailingText.count) {
+                postKey(CGKeyCode(kVK_LeftArrow))
```
vorssaint, line 435: a real left arrow press also carries the function and numeric pad flags, and the app's other synthetic key presses already add them for arrows. can these go out the same way, so an app that checks those flags still moves the caret

Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
```diff
@@ -403,6 +443,13 @@ final class TextSnippetService {
                 willPostShortcut: {
                     for _ in 0..<deleteCount { postKey(CGKeyCode(kVK_Delete)) }
                 },
+                didPostShortcut: {
+                    guard let delay = TextSnippetSupport.pasteCaretRetreatDelay(
+                        caretRetreat: caretRetreat
+                    ) else { return }
+                    DispatchQueue.main.asyncAfter(deadline: .now() + delay,
```
vorssaint, line 450: for a multiline snippet in an app that takes a moment to paste, does the caret still end up in the right place? the arrows go out 150 ms after the paste, while the clipboard itself is only put back half a second later

Sources/Vorssaint/Core/SnippetStrings.swift
```diff
@@ -194,7 +194,7 @@ extension SnippetFeatureStrings {
         folderPlaceholder: "Work",
         showInLibraryLabel: "Show in the quick menu",
         variablesFormatCaption: "A format after a colon picks how they look, like {{date:yyyy-MM-dd}}. A -tz(...) part sets the timezone, like {{date-tz(America/New_York):yyyy-MM-dd}}.",
-        editorFormatCaption: "A format after a colon picks how they look, like {{date:yyyy-MM-dd}}, or use the date/time button above. A -tz(...) part sets the timezone, like {{date-tz(America/New_York):yyyy-MM-dd}}.",
+        editorFormatCaption: "A format after a colon picks how they look, like {{date:yyyy-MM-dd}}, or use the date/time button above. A -tz(...) part sets the timezone, like {{date-tz(America/New_York):yyyy-MM-dd}}. Use {{cursor}} where the caret should go after expansion.",
```
vorssaint, line 197: this lands at the end of the date format explanation, and the variables line above it still lists only date, time, datetime and clipboard, so someone looking for the marker will not find it there. can you add {{cursor}} to the variables line instead, in every language

Issue #1109, opened 2026-08-28
menubar


Balajitechlabs:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

even i uninstall and install same thing it is not showing in the menu bar

### Steps to Reproduce

1.open app 
2.menu bar buuton to show 

### Feature Area

Windows and Dock

### Screenshots / Recordings

<img width="372" height="394" alt="Image" src="https://github.com/user-attachments/assets/a8e601fa-4597-4817-b4f2-9d7de3a5d2be" />

### Vorssaint Version

Version 27.0 Beta (26A5421a)

### macOS Version

macOS 27 (Golden Gate)

### Language

english

### Mac and displays

macbook air m1

### Installation Method

homebrew

PathGao:
The icon going missing doesn't need a reinstall. While Vorssaint is running, open it again from your Applications folder — that's the built-in way back in. It rebuilds the menu bar icon if it's genuinely gone, and if it can't get the icon back it opens the settings window instead. In settings, under the menu bar section, there's a "Show menu bar icon" button that rebuilds it at a fresh spot.

One thing worth ruling out first: if you turned on the option that shows system readings in the menu bar and hides the icon, the icon disappearing is what that option does — the numbers take its place.

If reopening doesn't get you back in, tell me two things. Which version of Vorssaint you're on: the version field in your report has your macOS build number in it, not the app's. And whether you see anything from Vorssaint in the bar at all, even numbers or a small gauge.

<details>
<summary>The technical side — you can skip this</summary>

Unverified: reporter is on a macOS 27 beta with an unknown app version (version field contains the OS build), and I can't read what the screenshot shows. Not reproduced.

Two paths can produce this and they need different fixes:
- the metrics-only option hiding the glyph (`MenuBarSpacingSupport.shouldHideStatusIcon`) — working as designed, not a bug;
- the status item genuinely dropped or parked off-screen. `applicationShouldHandleReopen` in `AppDelegate.swift` already handles that with a placement-resetting rebuild, and the comment there notes a macOS 27-specific mismatch where a rebuilt item's window reports the slot it was born in while the icon draws elsewhere. If reopen doesn't recover on macOS 27, that's where I'd look.

</details>

Balajitechlabs:
It is working in the latest build 👍

PR #1276, opened 2026-09-02
fix(menu-bar-organizer): clarify divider interactions

RuanMD:
## Summary

- keep only the outer menu bar organizer chevron interactive; internal dividers become editing/setup boundary markers
- preserve the requested hidden-section scope when presentation uses the secondary bar
- remove the conflicting permanent-divider preference without deleting its legacy `UserDefaults` value
- add focused filtering, presentation-mode, and structural regression coverage

## Help wanted

This is a **draft follow-up that depends on #360** and is intentionally not ready to merge. I would appreciate help from the maintainers and other macOS contributors to finish the feature, especially with:

- reviewing the AppKit status-item/divider behavior and stacked branch history
- validating the interaction on different menu bar layouts, notched displays, multiple monitors and Spaces
- exercising Accessibility grant/deny/revoke behavior with a clean Developer bundle
- deciding whether these follow-up commits should be folded into #360, rebased onto its latest head, or split further

## Before / after

**Before:** two chevron-like controls could trigger the same reveal action, and revealing only `Hidden` in secondary-bar mode could also include `Always Hidden` items.

**After:** one outer chevron controls reveal behavior. Internal items identify section boundaries only during editing/setup. A `Hidden` reveal filters to that section, while explicit “show all”/secondary-bar actions include both non-visible sections.

## Test plan

- [x] `swift build`
- [x] `./build.sh --test` — 9,492 checks
- [x] `./build.sh --dev`
- [x] `./build.sh`
- [x] `./build/Vorssaint --selftest`
- [x] install and launch `/Applications/Vorssaint (Developer).app`
- [ ] complete manual visual validation of all three presentation modes
- [ ] validate Accessibility denial, grant and revocation from a clean TCC state
- [ ] validate notched displays, multiple monitors and multiple Spaces

## Permissions and privacy

- uses the existing contextual Accessibility permission required by the organizer
- does not add Screen Recording, network access, entitlements, dependencies or private persistence
- keeps the legacy divider preference value untouched; the corrected version simply stops reading it

Refs #360


PathGao:
The macOS 27 gate is the right shape: fails closed on `major <= 26`, keeps preferences instead of wiping them, and `mayFlip` blocks install without blocking uninstall.

1. `FeatureRuntime.setAvailable`: the unsupported early return does `revision += 1` instead of `finishAvailabilityChange()`, so `CommandBarService.noteHubChange()` never runs. Install on 26, upgrade to 27, uninstall from the hub — the Command Bar keeps rows for a feature that's gone. `setAllAvailable` and `setAvailability(selected:)` don't have this hole.

2. `MenuBarOrganizerAdvancedStrings` covers ptBR/es/fr + default. All 40 `*Strings*.swift` in `Core/` on main cover 13 languages, and `MenuBarOrganizerStrings` in this PR does too — shortcut names land in English for ja/ko/zh/de/it/ru/tr.

3. `.rovodev/` hardcodes `/Users/ruansantana/Documents/DeepAgent/...`. Not in your four bullets — unless you mean to land your agent config here, drop it from the branch.

Diff truncated at 80KB of 390KB; I read nothing past the `fr` block of `MenuBarOrganizerStrings`. No hardware verification from me.

**Impact:** stale Command Bar rows after uninstall on macOS 27; untranslated shortcut names in 9 languages.
**Merge:** hold for #360 — stacked on it, currently conflicting.

ref #481
ref #1109

PathGao:
Closing this. It is a draft branched off #360, which is now closed because the menu bar organizer is gated off on macOS 27 — the full reasoning is on #360, and the branch is being kept for when #698 lands a backend that works there.

Two things for whenever this comes back:

- The divider clarifications read better as review comments on #360 than as a second branch alongside #824. Both are children of #360, and neither contains the other, so as it stands the same base is being carried three times.
- The diff also adds twelve workflow files — `SKILL.md`, `sprint-status.yaml`, and the `ai-handoff-*` and `epic-*` documents. Those are local planning artifacts and should not land in the repo; please drop them from the branch before it goes back up.

Issue #868, opened 2026-08-24
[Bug]: Window switcher does not switch to the selected window


Bald-M:
### Description

The window switcher opens normally and the selection highlight moves between available windows. However, after selecting another window and releasing the keyboard shortcut, the selected window is not brought to the foreground.

The previously active window remains focused, so the window switcher does not actually switch windows.

**Expected behavior:** Releasing the shortcut should close the window switcher and activate the highlighted window.

**Actual behavior:** The window switcher closes, but focus remains on the previously active window.

### Steps to Reproduce

1. Open multiple applications or windows.
2. Open Vorssaint and enable the window switcher.
3. Trigger the window switcher using its configured keyboard shortcut.
4. Move the selection highlight to a different window.
5. Release the keyboard shortcut.
6. Observe that the selected window is not activated and the previous window remains in the foreground.

### Screenshots / Recordings

<img width="2560" height="1415" alt="Image" src="https://github.com/user-attachments/assets/508cc29a-22ea-4763-b872-23819812e33e" />

### Vorssaint Version

3.3.2

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Github Release

Bald-M:
ignore pls, it seems good again

PR #1278, opened 2026-09-02

PathGao:
## Summary

Selecting an app in the switcher often left it behind the app you came from. Its
windows would rise but the menu bar stayed put, so the switch had not really
happened.

Cooperative activation only lets an app hand over activation it currently holds.
Vorssaint is `LSUIElement` and every surface that yields is non-activating — the
switcher panel, the Dock click tap, the process list — so at the moment a handoff
commits Vorssaint is usually not the active app. `NSApp.yieldActivation` then gave
away nothing, the `activate(from:)` that followed was refused, and the target
never came forward.

Self-activate first, so the yield has something to hand over.

Adopts #845 by @pboucher, which fixed two of the five yield sites. `SpaceHop.swift`,
`ProcessUsageService.swift` and `DockClickService.swift` were still yielding bare.
All five now go through one `ActivationHandoff` helper instead of repeating the
sequence, and a test walks `Sources` and fails if a bare `yieldActivation`
appears anywhere else, whichever way the receiver is spelled — this bug comes back
by adding a call site, not by editing one. `DockClickService` already described its sequence as "the same sequence the
Switcher, Space hop and process list already use", so the five were one pattern
before this change; now they share the code as well as the comment.

The self-activation posts an activation notification for Vorssaint's own process.
`WindowUseTracker.appActivated` is the only place an app is recorded as used at all
— `recordSwitch` files windows, not apps — so recording it is skipped only when
the handoff armed it, and only for one fresh notification. Guarding on our own pid
unconditionally, as the original did, would also stop the Dock icon, Settings and
switching onto one of Vorssaint's own windows from ever being recorded.

Only the recording is skipped: the app, and the focused-window read the observer
retarget would otherwise file, since the switcher lists Vorssaint's titled windows
and one of them would rank second after a handoff that took. The observer still
retargets. After self-activating Vorssaint really is front, and if the yield does
not take it stays there — an early return would have left the observer on the app
the user came from. When the yield does take, the target's own notification moves
it on a turn later.

The arm is keyed to two pids, our own and the target's, and is cleared by
whichever of those two activations arrives first. When Vorssaint already holds
activation `NSApp.activate` posts nothing and the first arrival is the target's; an
arm cleared only on our pid would stay live for the rest of its window and swallow
the next real activation of Vorssaint. The quick one is `activateOwnWindow`, the
switcher landing on one of Vorssaint's own windows right after a handoff — exactly
the case the narrow guard exists to keep recorded. By the time the target's
notification is in, the self-activation it was armed for has either been posted or
never will be, so clearing on it loses nothing. Any other app's activation leaves
the arm alone: one already queued when the handoff ran (a Space hop's travel posts
some) drains before our own, and letting it consume the arm would record the
self-activation it was meant to hide. The one-second deadline stays as the backstop
for when neither notification arrives at all.

Refs #961
Refs #868

## Verification

macOS 26.5.2, Apple Silicon.

- `./build.sh` — finishes with no warnings
- `./build.sh --test` — `TESTS OK (9606 checks)`

The selftest line is from the first head. The current head was verified with a full
`./build.sh` (bundle ready, no errors) and `./build.sh --test`; the tracker and the
handoff helper are outside the `--test` whitelist, so the full build is what
compiles them.

Eight new source-shape assertions: the handoff self-activates before it yields (the
first version of this test passed for the wrong reason and was fixed — both
patterns began with a newline, so the strict ordering compare could never hold);
it arms the tracker; a walk of every `.swift` under `Sources` finds no
`yieldActivation` outside `ActivationHandoff.swift`, reporting the offenders
and the scanned-file count so an enumerator that finds nothing cannot pass
vacuously; and the tracker's skip is conditional on the armed handoff rather than
on the pid alone, the arm is consumed before the pid check by our own activation or
the target's, an activation of any other app leaves it in place, it covers only the
recording while the observer still retargets, is consumed once, and expires.

**Not tested:** whether macOS actually grants the self-activation when the handoff
is triggered by a mouse click rather than a hotkey release. That needs the app
running on a real desktop with a window server and a genuine user gesture, and
cannot be checked from a build. The reasoning is that `NSApp.activate` runs
synchronously inside the event that requested the switch, which is what the system
accepts as a real gesture — but that is an argument, not a measurement. Worth
confirming on a Dock click and a process-list click before merging.

**Not changed:** `RadialNowPlayingService.open` and
`ClipboardHistoryService.pasteIntoPreviousApp` activate another app from a
non-activating panel with no yield. Neither asks cooperatively — both call
`activate(options:)` without `from:` — so neither is a handoff and the scan does
not cover them. Whether they should go through `ActivationHandoff` is a separate
call: the clipboard one posts ⌘V 120 ms after the activate, so if that activate
is refused the paste lands in whatever is front. Left on their current path here.

**Trade-off:** self-activating first changes the failure path. Before, a yield
that gave away nothing left the user on the app they came from. Now, if macOS
grants the self-activation but the yield still does not take, the user lands on
Vorssaint and has to click back. No mitigation is added: the callers own the
`activate` call and its result, a retry onto the previous app would go through the
same yield that just failed, and hiding Vorssaint to hand activation back would
drop its windows. When the self-activation is refused the behaviour is today's.
This is the trade #845 made, and the mouse-click case above is where it needs
checking.

The added tests are source-shape assertions. They pin the ordering and pin the
single call site; they cannot prove the activation is granted.

#868 is referenced because it reports the same symptom and is already closed — the
fix for it did not hold, and this records the recurrence.

### DIFF
--- a/Sources/Vorssaint/Services/ActivationHandoff.swift
+++ b/Sources/Vorssaint/Services/ActivationHandoff.swift
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import AppKit
+
+/// Hands this app's activation to another one.
+///
+/// Cooperative activation only lets an app pass on activation it currently
+/// holds. Vorssaint is `LSUIElement`, and the surfaces that trigger a handoff —
+/// the switcher panel, the Dock click tap, the process list — are all
+/// non-activating, so at the moment a handoff commits Vorssaint usually is not
+/// the active app. `yieldActivation(to:)` then gives away nothing, the
+/// `activate(from:)` that follows is refused, and the target never comes
+/// forward: its windows may rise but the menu bar stays with whatever app was
+/// already active. Self-activating first is what gives the yield something to
+/// hand over; it runs in direct response to the user's own key or click, which
+/// the system accepts as a real gesture.
+///
+/// Every yield in the app goes through here. A bare `NSApp.yieldActivation`
+/// added elsewhere would reintroduce the bug on that path alone, so
+/// `Tests/MetricsTests.swift` scans `Sources` and fails if one appears.
+enum ActivationHandoff {
+    /// Call on the main thread, synchronously with the gesture that asked for
+    /// the switch. The caller still activates `app` itself afterwards — this
+    /// only clears the way.
+    static func yield(to app: NSRunningApplication) {
+        // The self-activation below posts a real activation notification for
+        // our own process; tell the use tracker it is ours so it does not rank
+        // Vorssaint ahead of the app the user is switching away from.
+        WindowUseTracker.shared.expectSelfActivationHandoff()
+        NSApp.activate(ignoringOtherApps: true)
+        NSApp.yieldActivation(to: app)
+    }
+}
--- a/Sources/Vorssaint/Services/ActivationHandoff.swift
+++ b/Sources/Vorssaint/Services/ActivationHandoff.swift
@@ -16,9 +16,10 @@ import AppKit
 /// hand over; it runs in direct response to the user's own key or click, which
 /// the system accepts as a real gesture.
 ///
-/// Every yield in the app goes through here. A bare `NSApp.yieldActivation`
-/// added elsewhere would reintroduce the bug on that path alone, so
-/// `Tests/MetricsTests.swift` scans `Sources` and fails if one appears.
+/// Every yield in the app goes through here. A bare `yieldActivation(to:)`
+/// added elsewhere, whatever the receiver is spelled, would reintroduce the bug
+/// on that path alone, so `Tests/MetricsTests.swift` scans `Sources` and fails
+/// if one appears.
 enum ActivationHandoff {
     /// Call on the main thread, synchronously with the gesture that asked for
     /// the switch. The caller still activates `app` itself afterwards — this
--- a/Sources/Vorssaint/Services/ActivationHandoff.swift
+++ b/Sources/Vorssaint/Services/ActivationHandoff.swift
@@ -27,8 +27,9 @@ enum ActivationHandoff {
     static func yield(to app: NSRunningApplication) {
         // The self-activation below posts a real activation notification for
         // our own process; tell the use tracker it is ours so it does not rank
-        // Vorssaint ahead of the app the user is switching away from.
-        WindowUseTracker.shared.expectSelfActivationHandoff()
+        // Vorssaint ahead of the app the user is switching away from. Keyed to
+        // the target so only that app's activation, or ours, can clear it.
+        WindowUseTracker.shared.expectSelfActivationHandoff(clearedBy: app.processIdentifier)
         NSApp.activate(ignoringOtherApps: true)
         NSApp.yieldActivation(to: app)
     }
--- a/Sources/Vorssaint/Services/DockClick/DockClickService.swift
+++ b/Sources/Vorssaint/Services/DockClick/DockClickService.swift
@@ -695,9 +695,9 @@ final class DockClickService {
     /// the click precisely so the Dock would not act — so nobody else is going
     /// to. The restore then half-lands: unminimizing needs no activation
     /// rights, so the windows come back and are immediately stacked under
-    /// whichever app is still active. Yielding this app's activation first is
-    /// what makes the request cooperative, the same sequence the Switcher,
-    /// Space hop and process list already use.
+    /// whichever app is still active. `ActivationHandoff` is what makes the
+    /// request cooperative, the same sequence the Switcher, Space hop and
+    /// process list use.
     ///
     /// Options stay empty on purpose: `restoreBackToFront` earns the batch's
     /// stacking one window at a time, and `.activateAllWindows` would re-raise
@@ -705,7 +705,7 @@ final class DockClickService {
     private static func activate(pid: pid_t) {
         DispatchQueue.main.async {
             guard let app = NSRunningApplication(processIdentifier: pid), !app.isTerminated else { return }
-            NSApp.yieldActivation(to: app)
+            ActivationHandoff.yield(to: app)
             if !app.activate(from: NSRunningApplication.current, options: []) {
                 app.activate(options: [])
             }
--- a/Sources/Vorssaint/Services/Switcher/SpaceHop.swift
+++ b/Sources/Vorssaint/Services/Switcher/SpaceHop.swift
@@ -87,7 +87,7 @@ final class SpaceHop {
         // Space forward too; when the app has one here, activate quietly and
         // let the travel decide what comes up.
         let appIsAlreadyHere = appHasWindowOnVisibleSpace()
-        NSApp.yieldActivation(to: app)
+        ActivationHandoff.yield(to: app)
         if !app.activate(from: NSRunningApplication.current,
                          options: appIsAlreadyHere ? [] : [.activateAllWindows]) {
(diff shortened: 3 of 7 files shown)


PathGao:
Five sites is all of them — `main` has exactly five `NSApp.yieldActivation` calls and this reaches every one. Arming the tracker per handoff rather than guarding on pid is the right call; the unconditional version would have erased Dock icon and Settings activations.

**The scan catches one spelling.** `MetricsTests` matches `NSApp.yieldActivation`, so `NSApplication.shared.yieldActivation(to:)` walks straight past the guard — and `NSApplication.shared` is already used elsewhere in the repo. Match `yieldActivation` alone.

**The early return drops two more statements.** `appActivated` also skips `requestedPID` and `retargetObserver()`; the body only justifies skipping the recording. If the handoff doesn't take, the AX observer stays pointed at the previous app while Vorssaint is front. Skip just `promote(app:)` if recording is all you meant to suppress.

**Failure path is worse than today.** Self-activating first means a yield that doesn't take leaves the user on Vorssaint instead of the app they came from — today they just stay put. That is the mouse-click case the body says is unverified.

**Impact:** every switcher, Dock and process-list activation; a refused handoff lands the user on Vorssaint.
**Merge:** needs the yield scan widened to every spelling first.

ref #961

PathGao:
@PathGao

Yield scan: the match is now `yieldActivation` alone, so `NSApplication.shared.yieldActivation(to:)` fails it the same way `NSApp.yieldActivation` does. `ActivationHandoff.swift` stays excluded and the scanned-file count is still asserted greater than zero. Checked the comment trap before widening: on the unmodified branch the only mentions of `yieldActivation` outside the helper are in `Tests/`, which the scan does not walk, so nothing needed rewording and the scan does not strip comments. A comment naming it elsewhere would now fail with the file name in the message. That is intended, the helper's own doc comment says where to put the call.

Tracker: `appActivated` on an armed handoff no longer returns early. It still sets `requestedPID` and retargets the observer, and skips only the recording. Reading the rest of the path: the retarget ends in `readFocusedWindow(of:)`, which files that window, and the switcher lists Vorssaint's titled windows, so a plain skip of `promote(app:)` would let a handoff that took file a Vorssaint window second in the window history, the window-side twin of the app ranking the guard exists for. `retargetObserver` therefore takes `filingFocusedWindow`, false for the handoff. The observer itself is installed either way: if the yield takes, the target's notification moves it on a turn later; if not, Vorssaint is front and the observer is on it. This path already ran for our own pid on every Dock-icon and Settings activation, so nothing in it is new for our process. The source-shape assertion that pinned the early return is replaced by two that pin the `let selfHandoff` guard and the two conditional calls.

Failure path: no mitigation added. The callers own `activate` and its result, a retry onto the previous app would go through the same yield that just failed, and `NSApp.hide` would hand activation back but drop Vorssaint's windows. When the self-activation is refused the behaviour is today's. When it is granted and the yield still does not take, the user lands on Vorssaint. That is the trade #845 made and the PR body now says so in a trade-off paragraph, with the mouse-click case as the thing to check on a desktop before merging.

Verification: `./build.sh --test` gives `TESTS OK (9605 checks)`, up from 9604 because one assertion became two. Red first, re-running `build/metrics-tests` against edited sources: a probe file under `Sources` containing `NSApplication.shared.yieldActivation(to: app)` gave `TESTS FAILED (1 of 9605)` naming `YieldProbe.swift` across 430 scanned files; the early-return spelling put back into the tracker gave `TESTS FAILED (2 of 9605)`, both new tracker assertions. Green again after each restore. `WindowUseTracker.swift` and `ActivationHandoff.swift` are outside the `--test` whitelist, so the module was typechecked whole: `swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk -I Sources/VMStatisticsCompat -I Sources/HIDEventSystem $(find Sources/Vorssaint -name '*.swift')` exits 0 with no errors and no warnings.

Not verified: no full `./build.sh` or `--selftest` on this head, and nothing ran on a desktop, so the observer landing on Vorssaint after a refused yield and the self-activation on a mouse click are both still untested.

PathGao:
Both land, and the `filingFocusedWindow: false` split is better than the plain `promote` skip I suggested. Red-first against edited sources is what makes the new assertions worth their line count.

**The scan is blind to the worse shape.** It only fires on files that yield. `RadialNowPlayingService.open` and `ClipboardHistoryService.pasteIntoPreviousApp` activate another app from a non-activating surface with no yield at all, so they pass clean — and the clipboard one posts the paste 120 ms later into whatever is actually front. If those are not deliberately left on the legacy path, they are the same bug the scan exists to stop.

**The arm can outlive its handoff.** Only an activation of our own pid clears it. When Vorssaint already holds activation, `NSApp.activate` posts nothing and the target's notification carries the target pid, so the arm stays live for the rest of the second and swallows the next real activation of Vorssaint. The doc comment names the Dock icon, but the fast racer is `activateOwnWindow` — the case the narrow guard exists to preserve. Clear it in `appActivated` whatever the pid.

**Impact:** a switch onto a Vorssaint window right after a handoff goes unrecorded; radial Now Playing and clipboard paste keep the old symptom.
**Merge:** needs the arm cleared on any activation first.

ref #961

PathGao:
@PathGao Arm: `appActivated` now consumes it before the pid check, so whichever activation arrives first clears it, and only our own pid is then left unrecorded. When Vorssaint already held activation the self-activation posts nothing, the target's notification is the first arrival, and by then the self-activation it was armed for has either been posted or never will be, so clearing on it loses nothing; the one-second deadline stays only for the case where neither notification arrives. The doc comment on the deadline now names `activateOwnWindow` as the quick racer instead of the Dock icon. Red first: the assertion was changed to pin `let handoffArmed = consumeSelfActivationHandoff()` ahead of `let selfHandoff = handoffArmed && pid == ...` and gave `TESTS FAILED (1 of 9605)` against the unmodified tracker, then green after the source change. Full `./build.sh` on this head finishes with the bundle ready, `./build.sh --test` gives `TESTS OK (9605 checks)`.

The two no-yield sites are left alone here. `RadialNowPlayingService.open` and `ClipboardHistoryService.pasteIntoPreviousApp` both call `activate(options:)` without `from:`, so neither asks cooperatively and neither is a handoff the scan is meant to see; they look like the older plain-activate path rather than a deliberate exclusion, and the clipboard one does post ⌘V 120 ms after an activate that may be refused. Whether they should route through `ActivationHandoff` is a separate change, noted in the body under "Not changed".

PathGao:
Arm-before-pid-check is right, and the `filingFocusedWindow: false` split is better than the plain skip. Red-first against edited sources is what makes the new assertions worth their line count.

**Any activation clears it, including one that is neither yours nor the target's.** Arming is synchronous with `NSApp.activate`, but a workspace notification already queued when the handoff runs drains first and consumes the arm — then your own self-activation is recorded, Vorssaint lands first in `appHistory`, and the next switcher invocation offers it as most-recent. That is the case the guard exists for. `SpaceHop` is where I'd expect it: the travel posts activations of its own.

`yield(to:)` has the target in hand, so key the arm rather than leaving it open:

```swift
expectSelfActivationHandoff(clearedBy: app.processIdentifier)
// consume only when pid == ours || pid == the armed target
```

That keeps the "Vorssaint already held activation, the target's notification arrives first" case you just fixed, and a foreign activation stops eating the arm.

Not verified: read off the code, nothing ran on a desktop, so the notification ordering is an argument.

**Impact:** an occasional handoff ranks Vorssaint most-recent, and the next switch lands on it.
**Merge:** needs the arm keyed to the self and target pids first.

ref #961

PathGao:

PathGao:
Reopened as #1303.

PR #1281, opened 2026-09-02
fix(homebrew): re-read what is installed after a partly done operation

PathGao:
## Summary

`brew` reports a run as failed when it could not do all of it, not only when it
did none of it. One disabled package is enough: `brew upgrade` skips that
package, upgrades every other one, and still exits non-zero.

Reproduced in an isolated Homebrew prefix — local tap, formula `alpha` installed
at 1.0, bumped to 2.0 and marked `disable!`, formula `beta` merely outdated —
against Homebrew 6.0.21:

```
==> Fetching downloads for: local/sandbox/beta and local/sandbox/alpha
Error: local/sandbox/alpha has been disabled because it is a sandbox experiment!
==> 1 Deprecated or disabled package
local/sandbox/alpha (disabled)
==> Upgrading local/sandbox/beta
  1.0 -> 2.0
==> Upgraded 1 outdated package
exit 1
```

`beta` upgraded; only the disabled package was skipped. The mechanism is
`FormulaInstaller#prelude_fetch` raising `CannotInstallFormulaError`, which
`Homebrew::Install.select_formula_installers` rescues with `ofail` — that prints
the `Error: <name> has been disabled ...` line, sets a non-zero exit, drops that
one installer and carries on. The same `rescue` is in `upgrade.rb` at 3.6.0
through 6.0.21. Casks do not even error: a disabled cask gets `opoo` and exit 0.

`refreshInstalled()` ran only on a zero exit, so after such a run the panel kept
the versions from before it and the outdated list stayed stale. Because the
disabled package stays installed and stays outdated, pressing Upgrade all again
repeated the same apparent failure with nothing visibly changing — it never
recovers on its own. A cancelled run has the same problem: it can stop after it
has already changed something. So does a run that ends at a password prompt:
`brew upgrade` moves the formulae, a cask then asks for `sudo`, and the run is
handed to the terminal with the formulae already upgraded.

All three paths now re-read. `refreshInstalled` used to clear the error banner
as it started, which would have wiped the reason the operation had just given,
so it takes a `clearingError:` parameter and the three post-failure callers pass
`false`. The untrusted-tap branch deliberately does not re-read: `refreshInstalled`
calls `clearUntrustedTap()`, which would drop the prompt and retry closure that
branch has just installed.

This is deliberately not a command rewrite. Filtering the disabled package out of
the upgrade set would mean turning bare `brew upgrade` into an explicit name
list, which hands the app the job of deciding what "everything" means, brings
back the formula/cask name-collision problem, and drops brew's own dependent
handling — all to suppress an exit code that is telling the truth.

## Verification

macOS 26.5.2, Apple Silicon, Homebrew 6.0.21.

- `./build.sh` — finishes with no warnings (at d702c9a; the needs-terminal
  follow-up was typechecked with `swiftc -typecheck` over the whole module, zero
  errors and zero warnings, not rebuilt)
- `./build/Vorssaint --selftest` — `SELFTEST OK` (at d702c9a)
- `./build.sh --test` — `TESTS OK (9602 checks)`

The refresh happens inside a completion closure no unit test can drive, so the
four new assertions read the source: the cancelled, needs-terminal and failed
paths all call `refreshInstalled(clearingError: false)` (exact count 3), and both
banner clears in `refreshInstalled` sit behind the parameter. Negative control:
removing any one of the three refresh calls fails the suite with `found 2`; an earlier
draft of the whole-file assertion also failed for a real reason — `perform()`
legitimately clears the banner when starting an operation — and was narrowed to
an exact count.

**Not tested in the app:** the panel behaviour itself, before and after, against a
real Homebrew install with a disabled outdated package. The brew-side behaviour
above was reproduced in a scratch prefix; the app-side consequence is read from
the code path, not observed. Worth a look on a Mac that has one — `six` is the
package the original report named.


PathGao:
The mechanism write-up is right, and `clearingError:` is the correct shape — re-reading without wiping the reason is the easy part to miss.

One branch of the same if-chain didn't get it. `needsTerminalFallback` in `perform` is the same "non-zero after doing some of it" case, and it's likelier to be partly done than the disabled-package one: `brew upgrade` upgrades the formulae, a cask then asks for a password, output matches `password:`, branch taken, no re-read. Same `self.refreshInstalled(clearingError: false)` after `markOperationComplete`. That also moves the exact-count assertion in `MetricsTests` from 2 to 3.

The other `.failed` branch (untrusted tap) must not get it: `refreshInstalled` calls `clearUntrustedTap()`, which drops the prompt and the retry closure `presentUntrustedTap` installed one line above. Worth saying so in a comment, or the next person completing the pattern kills that prompt silently.

I didn't verify panel behaviour either — read from the code path only.

**Impact:** after a sudo-prompt fallback the panel keeps the pre-run versions, the same bug on a neighbouring path.
**Merge:** needs the `needsTerminal` branch re-reading first.

PathGao:
@PathGao both points taken, head is now 660a92f.

1. `needsTerminalFallback` branch of `perform` now calls `self.refreshInstalled(clearingError: false)` after `markOperationComplete`, same as the cancelled and failed branches. `terminalFallbackCommand` is not touched by `refreshInstalled`, so the "open in Terminal" action survives the re-read. The exact-count assertion in `Tests/MetricsTests.swift` moves from 2 to 3 and its message names the three paths.

2. Untrusted-tap branch left without a refresh, with a comment saying why. Checked on this head: `refreshInstalled` calls `clearUntrustedTap()` unconditionally before running `brew info`, and `clearUntrustedTap` nils both `untrustedTap` and `untrustedTapRetry`, which is exactly what `presentUntrustedTap` set one line earlier. A refresh there would also, on a non-zero `brew info`, re-present the prompt with `refreshInstalled()` as the retry instead of `perform(action, package:)`.

Verification. `./build.sh --test` with the new call and the old count: `TESTS FAILED (1 of 9602): the cancelled and the failed operation paths re-read, found 3`. Count updated to 3: `TESTS OK (9602 checks)`. Then removed each of the three `refreshInstalled(clearingError: false)` calls in turn and re-ran `./build/metrics-tests`: all three red with `found 2`. Restored: `TESTS OK (9602 checks)`. Checks number unchanged because the assertion count did not change.

`HomebrewManager.swift` is not in the `--test` whitelist. Typechecked the whole module instead: `swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk -I Sources/VMStatisticsCompat -I Sources/HIDEventSystem $(find Sources/Vorssaint -name '*.swift')`, exit 0, no output. Not run on this head: `./build.sh` full build and `--selftest`. Panel behaviour after a sudo prompt still not observed in the app.

PR body updated: three callers, four source assertions, negative control `found 2`, build lines marked as from d702c9a.

PathGao:
Both fixed, and the negative control (drop any one call → `found 2`) is what makes an exact-count assertion worth having.

Grepped `main`: `markOperationComplete` appears only in `perform`, and every other `refreshInstalled` call site is UI-initiated on the default. No fourth branch, no sibling missed.

One hole left, not a blocker. `clearingError:` guards the two writes *before* `brew info` runs, not the two after it: in `refreshInstalled`'s completion a non-zero status overwrites `errorMessage` with raw `brew info` output, and its untrusted-tap arm calls `presentUntrustedTap`, which nils `errorMessage` and installs `refreshInstalled()` as the retry — the hazard you kept the tap branch out for, arriving from the other side. `brew info --json=v2 --installed` reads local state, so not the disabled-package case; reachable via the 30s read timeout or a broken prefix. `operationStatus.lastActivity` still holds the reason until the cleanup timer, so the card survives and the banner does not. Gate those two on the same parameter if you want it closed.

I have not run the app either.

**Impact:** none on the reported path.
**Merge:** ship it.

Issue #1280, opened 2026-09-02
Radial menu Now Playing shows nothing on macOS 15.4 and later


PathGao:
## What happens

In the radial menu, the **Now Playing** action does nothing. Music (or a video in a browser) is playing, you open the wheel, pick Now Playing, and no card appears. Play/pause, next and previous on the same wheel work.

## Where it was seen

macOS 26.6.2, Vorssaint 3.3.3-beta.3 (the shipped build). A session was playing in Chrome, Control Center showed it, and the wheel's own play/pause toggled it. Now Playing stayed silent on every try.

## Why

Since macOS 15.4 the system only hands Now Playing information to processes that carry Apple's own signature. Vorssaint asks, gets an empty answer, and treats that as "nothing playing", which shows nothing. So the action has been empty on every Mac running 15.4 or later since it shipped in 3.3.3-beta.2; nobody reported it because it fails quietly.

<details>
<summary>The technical side — you can skip this</summary>

- `MediaRemoteNowPlayingBridge` (`Services/RadialMenu/RadialNowPlayingService.swift`) resolves `MRMediaRemoteGetNowPlayingInfo` in-process with `dlsym`. On 15.4+ the callback returns an empty dictionary for an unentitled process; the bridge's 0.75 s budget then finishes with `nothingPlaying`, and `presentDetails(at:)` has `case .nothingPlaying: break`.
- `/usr/bin/perl` carries no entitlements at all (`codesign -d --entitlements :- /usr/bin/perl` is empty) and still gets the full answer, so the gate is on Apple's platform signature, not an entitlement an app could request.
- Measured on the same Mac: a 60-line Swift library loaded into perl via `DynaLoader` returned title, artist, artwork, pid and `com.google.Chrome` in 20–30 ms, while the app got nothing.
- Same wall DockDoor hit: it went Spotify/Apple-Music-only in v1.14 after 15.4, then adopted the perl-hosted adapter route ("works with any app playing audio").
- Not verified: macOS 14 and 15.0–15.3, where the in-process read presumably still works.

</details>

A fix that reads Now Playing through perl is in a pull request; #380 (Dock Preview media controls) was built on the same broken read and is being redone on top of it.

PR #1282, opened 2026-09-02
fix(radial-menu): read now playing through perl

PathGao:
## Summary

The radial menu's Now Playing action has shown nothing on every Mac running macOS 15.4 or later since it shipped: MediaRemote answers `MRMediaRemoteGetNowPlayingInfo` with an empty dictionary unless the caller carries Apple's platform signature, the bridge times out into `nothingPlaying`, and that branch shows nothing. Measured today on macOS 26.6.2 with the shipped 3.3.3-beta.3: Control Center showed a playing session, the wheel's own play/pause toggled it, Now Playing stayed empty on every try.

This reads Now Playing the way DockDoor does since 15.4: through `/usr/bin/perl`, which is a platform binary and gets the answer. `Sources/NowPlayingAdapter/NowPlayingAdapter.swift` is a ~100-line library that resolves the same four MediaRemote functions the old bridge did and prints one JSON line; `Resources/now-playing.pl` loads it into perl with `DynaLoader` and calls it; `MediaRemoteNowPlayingBridge` runs that through `BoundedProcessRunner` (2 s, 17 MB pipe cap) and parses the last non-blank line of the output with `RadialNowPlayingSupport.adapterReply` (stderr shares the pipe, so a perl locale warning ahead of the JSON is skipped), which hands the existing `snapshot(info:...)` the same keys it always read. The adapter has no clock of its own: it blocks until MediaRemote's callbacks fire, and the bridge's 2 s kill is the only deadline, read as nothing playing. 2 s because a cold perl load measured 500 ms with nothing playing, and a real reply adds the MediaRemote round trip and up to 16 MB of base64 through the pipe; a kill reads as nothing playing and `beginSession` asks once, so a tight deadline empties the first wheel after login. The wheel shows `.loading` and holds the card anchor while it waits, so the extra second is not visible. Artwork is capped at 12 MB, the cap `snapshot` already applied; the adapter drops larger art before encoding, and the pipe cap is that 12 MB × 4/3 for the base64 line plus 1 MB for the rest. Nothing above the bridge changes. build.sh compiles the library with `swiftc -emit-library`, stages it under `Contents/Frameworks/` next to the script in `Resources/`, and signs it on its own identifier the way the fan helper is signed. One round trip costs 20–30 ms on an M-series Mac.

What a user notices: Now Playing shows the card again.

**Decisions for @vorssaint, since they are yours rather than review items:**

1. **One-shot `get` per wheel open** (this PR) **or a long-lived stream.** The wheel asks once when it opens, within the bridge's 2 s budget. A Dock Preview media card (#380, closed, to be redone on this bridge) would want a resident `perl … stream` process publishing changes instead of a spawn per hover. I kept the one-shot because it changes nothing in the app's lifecycle; if you want the stream, say so and #380's redo carries it.
2. **A second binary in the bundle.** `libVorssaintNowPlaying.dylib` is built, staged and signed separately (`$APP_BUNDLE_ID.now-playing`, hardened runtime under Developer ID). Notarization has to accept a dylib that is not linked by the app; the fan helper is the precedent, but I could not run the Developer ID path here.
3. **What it stands on.** The read now rents perl's platform signature instead of the app's own process. Apple can close that too; when it does, the adapter prints an error and the wheel is back to "nothing playing", the same as today. The old in-process path is removed rather than kept as a fallback, since it is empty on every supported system from 15.4 up. On 14 and 15.0–15.3 the perl route returns the same data but pays a spawn per wheel open where the old in-process call cost nothing; those systems trade that for one code path.
4. **Silent "nothing playing".** Unchanged here: with no session the action still does nothing, and `mediaNothingPlaying` is still never shown. Separate change if you want it.

## Verification

MacBook Air (M5), macOS 26.6.2 (25G83), one display, Simplified Chinese UI, own build with `./build.sh --dev` (ad-hoc signed; the legacy identity is not in this keychain).

- Before: shipped 3.3.3-beta.3, session playing in Chrome (Control Center ⏸), radial menu → Now Playing: no card, three tries.
- After: Developer build, same session (a Bilibili video in Chrome): the wheel shows the Chrome icon on the Now Playing slice and the title in its centre, and choosing it opens the card with title and artwork. Wheel opened from the Settings page's try-it button; the keyboard shortcut collides with the shipped app's on the same Mac.
- `/usr/bin/perl Contents/Resources/now-playing.pl Contents/Frameworks/libVorssaintNowPlaying.dylib` against the staged bundle returns the JSON line in 20–30 ms.
- After the review round (last-line parse, no adapter timeout, one artwork cap): the adapter compiled with build.sh's `swiftc -emit-library` line, unsigned, and run through `/usr/bin/perl Resources/now-playing.pl <dylib>` prints one JSON line (`{"pid":0,"isPlaying":false}`, no session playing) in 29 ms warm, 500 ms on first load. Not re-run on screen.

```sh
./build.sh --dev               # no warnings
./build/VorssaintDeveloper --selftest   # SELFTEST OK
./build.sh --test              # TESTS OK (9603 checks)
```

Tests added: the adapter line parses into the keys `snapshot` reads (artwork base64 → bytes, pid, displayID, playing flag); a perl warning ahead of the JSON line still parses; a paused reply degrades to Nothing Playing through the unchanged path; an `error` line, a non-object, shell noise and bad base64 never become a snapshot; build.sh compiles the adapter and stages both files.

Not tested: a cold run with a session playing (the 500 ms number is the no-session case; the 2 s deadline is the margin for that, not a measurement); Developer ID signing and notarization of the dylib; macOS 14 and 15.0–15.3; a Mac where `/usr/bin/perl` is missing or replaced (the bridge then reads as nothing playing); the plain `./build.sh` release variant (the staging and signing lines are shared with `--dev`).

## Anything else

Refs #1280


PathGao:
@PathGao ready for a first read. Verified on screen in the Developer build: the Now Playing slice shows the source app's icon, the wheel centre shows the title, the card opens with artwork, and play/pause, next and previous still work from the same wheel.

PathGao:
Right call putting the parse boundary at `adapterReply` — `snapshot(info:)` is untouched, so the blast radius is one file.

Three things:

1. `BoundedProcessRunner` merges stderr into the same pipe, and `adapterReply` parses the whole buffer as one object. A single `perl: warning: Setting locale failed.` ahead of the JSON returns nil — the same empty wheel #1280 reports. Parse the last non-empty line instead.

2. Two timeouts: the adapter waits 0.75 s, the bridge kills at 1.0 s, leaving 0.25 s for perl start plus dlopen. Nothing reads the `timedOut` key either. Drop the adapter's wait — a killed perl and a partial line both land on `nothingPlaying` anyway.

3. The artwork cap is three numbers now: 4 MB in the adapter, 8 MB on the pipe, 12 MB in `RadialNowPlayingSupport`. Art between 4 and 12 MB used to show and now doesn't, which isn't in the body's list. Pick one.

Grepped `MRMediaRemote` on main: this bridge is the only caller, no sibling site left behind. Not verified: Developer ID signing and notarization of the dylib, and anything needing a real machine.

**Impact:** on any Mac where perl writes to stderr, Now Playing stays empty exactly as today.
**Merge:** needs the reply parsed as the last line first.

ref #1280

PathGao:
@PathGao all three taken, head is 8148264d.

1. `adapterReply` now parses the last non-blank line of the pipe (split on `\n`, skip lines that are only space, tab or CR). New assertion: two `perl: warning:` lines ahead of the JSON line, trailing newline, still parses to title, pid and isPlaying. Red first: with the assertion in and the old whole-buffer parse, `./build.sh --test` reported `TESTS FAILED (1 of 9603)` naming that assertion; with the new parse, `TESTS OK (9603 checks)`.

2. The adapter's 0.75 s wait and the `timedOut` key are gone. It now calls `group.wait()` with no timeout on the same DispatchGroup it already used, so it blocks on a semaphore until every MediaRemote callback has fired, no spin. The bridge's 1 s kill is the only deadline and still lands on `nothingPlaying` through the existing `result.timedOut` guard. Nothing read `timedOut`, confirmed by grep before removing it.

3. One artwork cap: 12 MB, the number `snapshot`'s sanitizer already applied on main, so art that showed before still shows. The adapter's cap moves 4 MB → 12 MB. The pipe cap is derived in `RadialNowPlayingSupport.maximumAdapterReplyBytes` as 12 MB / 3 × 4 + 1 MB = 17 MB, with the arithmetic in the comment next to it, and the bridge reads it from there instead of its own 8 MB. The adapter's constant cannot share the Swift symbol (it is a separate `-emit-library` unit), so its comment names the app-side constant it mirrors. Body now states the 12 MB cap, the 17 MB pipe cap, the last-line parse, the missing adapter clock, the 1 s budget in decision 1, and 9603 checks.

Verification: `./build.sh --test` twice as above (compiles the whitelist, `RadialMenuSupport.swift` and `BoundedProcessRunner.swift` are in it). `RadialNowPlayingService.swift` and the adapter are not in the whitelist: `swiftc -typecheck -target arm64-apple-macosx14.0 -sdk MacOSX26.sdk -Xfrontend -interface-compiler-version -Xfrontend 6.3.2 -I Sources/VMStatisticsCompat -I Sources/HIDEventSystem $(find Sources/Vorssaint -name '*.swift')` exits 0 with no output. The adapter was compiled with build.sh's own line (`swiftc -O ... -emit-library -module-name VorssaintNowPlaying`) into a scratch directory, unsigned, and run three times through `/usr/bin/perl Resources/now-playing.pl <absolute dylib path>`: exit 0, one line `{"pid":0,"isPlaying":false}` (no session playing on this Mac), 29 ms warm and 500 ms on the first load. So the no-timeout wait returns on its own when the callbacks fire. Note the script needs an absolute library path; hardened perl refuses a relative one, and the bridge passes `library.path`, which is absolute.

Not verified: a run with a session playing (artwork path through the 12 MB cap and the 17 MB pipe), the on-screen wheel after this round, Developer ID signing and notarization, macOS 14 and 15.0–15.3.

PathGao:
All three landed, and the red-first on the warning assertion is the part I'd have asked for. Dropping the adapter's own clock is right — I read `BoundedProcessRunner`: the SIGTERM/SIGKILL ladder always reaps, so a stuck callback can't leave perl behind.

What's left is your own number: 500 ms cold perl start against a 1.0 s kill. That 500 ms was measured with nothing playing, so it excludes the MediaRemote round trip and up to 16 MB of base64 through the pipe — the path you list as untested. An overrun reads as `nothingPlaying`, and `beginSession` calls `refresh` once with no retry, so the first wheel open after login shows the empty slice #1280 reports and the second works. On 14 and 15.0–15.3 the old in-process path had no spawn cost at all, so decision 3's "works as well" isn't a wash there.

Cheapest fix is the deadline: `refresh` already renders `.loading` and `presentDetails` parks on `pendingPresentationAnchor`, so 2 s costs nothing on screen.

Minor: the 12 MB cap is still two copies. The dylib's `CommandLine.arguments` sees perl's argv, so the bridge could pass it as a third arg, if you think it earns the arg.

**Impact:** the first Now Playing open after login can still show nothing; the second works.
**Merge:** needs the 1 s deadline widened, or a cold run with a session playing measured, first.

ref #1280

PathGao:
@PathGao Widened the bridge deadline from 1 s to 2 s (9519895); the constant's comment says why: 500 ms was the no-session cold load, a real reply adds the MediaRemote round trip and up to 16 MB of base64, and the wheel shows .loading meanwhile. No test pinned the 1 s number, so none changed. PR body updated: 2 s throughout, the cold run with a session playing stays listed as not measured, and decision 3 no longer claims 14 and 15.0–15.3 work as well (they pay a spawn the in-process call did not). Left the argv-passed cap alone: the two 12 MB copies only exist when there is artwork and it was non-blocking. Full ./build.sh exit 0, TESTS OK (9603 checks).

PathGao:
Right call leaving the tests alone — nothing pinned 1 s, and asserting 2 s would pin a number you measured under one condition. The reasoning belongs in the constant's comment, which is where you put it.

Two follow-ups, neither blocks:

1. The adapter is the only compiled binary build.sh never runs. The app and the fan helper each get `--selftest` right after their compile; the adapter gets a test that greps build.sh for strings, which passes on a dylib perl loads and finds no symbol in. Add the run you did by hand — but it needs its own kill: `group.wait()` has no clock, and a runner with no MediaRemote session may never fire the callbacks.

2. `MediaRemoteNowPlayingBridge.queue` is serial and now blocks for the whole run: 2 s, plus up to 1 s of `BoundedProcessRunner`'s SIGTERM/SIGKILL ladder, plus 0.2 s of drain. A second wheel open inside that window waits behind the first. `attributes: .concurrent` covers it; `refresh`'s generation guard already drops the stale reply.

Both build checks were still blank when I read this.

**Impact:** none new; on a Mac where the read times out, rapid reopens stack.
**Merge:** ship it.

ref #1280

PR #1284, opened 2026-09-02
Feat/contribute

sairam-edu:
## Summary

What changes, and for a fix, what was wrong.

If this adds or changes a feature, say why it belongs in Vorssaint. Direction
is the most common reason a pull request is turned down here, and it is far
cheaper to settle in an issue before the code exists. Say what the feature is
built on and whether that will still be here, what it drags in at runtime and
who carries that when it breaks, what it exposes the project to, and how much
permanent surface it adds against how many people will ever switch it on.
Intel support, hardcoded integrations with particular third-party apps, a
plugin system and video downloading have all been declined on those grounds;
CONTRIBUTING works them through with the cases.

## Verification

Where you ran it, in as much detail as you have, and **what you could not
test**. The Mac model, the macOS version and build, the display and Spaces
layout, whether it was a release build or one of your own — any of those can
be the whole difference. Every pull request is run locally before it is
merged, on one maintainer's hardware, so the gap between your machine and
that one is the useful part; changes have passed review and then failed on a
different Mac.

```sh
./build/Vorssaint --selftest
./build.sh --test
```

New user-facing text has to compile in every supported language; a missing
translation is a failed build, not a missing string.

## Anything else

**If an issue exists for this, reference it, written exactly as `Refs #N`.**
Not `Closes`/`Fixes`, which closes the report on merge when the fix has not
reached anyone's Mac yet, and not a bare `#N` or a sentence about it. That
line is what the fix-status bot reads when this merges: it labels the issue,
comments again when a release carries the fix, and closes the issue after two
weeks of silence. A pull request that leaves it out drops the report out of
that track, and nobody notices until the reporter asks months later.

Several issues means several references, each on its own line and each
starting with its own `Refs`. The bot reads the word and the number as a pair,
so `Refs #N, #M` reaches the first number and leaves the second one
sitting there.

If the issue carries more than one report and this covers only one, say which,
here and on the issue when this merges. `Depends on #N` if it has to land
after another branch.

### DIFF
--- a/.github/ci/performance-gates.env
+++ b/.github/ci/performance-gates.env
@@ -1,0 +1,16 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# Copyright (C) 2026 Vorssaint
+
+# Startup latency budget in milliseconds (open app -> menu bar item appears).
+VORSSAINT_PERF_STARTUP_BUDGET_MS=2500
+
+# Menu-bar responsiveness budget in milliseconds (status-item click -> panel open).
+VORSSAINT_PERF_POPOVER_BUDGET_MS=450
+
+# Probe shape: warmup runs are discarded; measured runs are aggregated by median.
+VORSSAINT_PERF_WARMUP_RUNS=1
+VORSSAINT_PERF_ITERATIONS=4
+
+# Per-run hard timeouts in milliseconds.
+VORSSAINT_PERF_STARTUP_TIMEOUT_MS=12000
+VORSSAINT_PERF_POPOVER_TIMEOUT_MS=3000
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,9 @@ jobs:
       - name: Selftest
         run: ./build/Vorssaint --selftest
 
+      - name: Startup and menu-bar responsiveness gates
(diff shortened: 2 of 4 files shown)


PathGao:
Right shape for a perf gate: warmup discarded, median over repeats, budgets in an env file instead of inline asserts.

**It can't run on a GitHub runner as written.** `System Events` driving another process needs an Accessibility grant, and nothing here provides one. Every `osascript` errors, `wait_until` swallows it (`2>/dev/null || true`), and the run dies at the 12s timeout — red CI on every PR. `Tools/ui-smoke.sh` already drives the same `pop over 1 of menu bar item 1` path and is deliberately not wired into CI; its header says it needs Accessibility + Screen Recording for the calling terminal. I can't test a GH runner myself — if you have a green Actions run of this step, link it.

**Budgets are under the harness's own overhead.** `now_ms()` forks python3, so each `wait_until` tick costs two python3 spawns plus an osascript, far above the 50ms sleep. 450ms for the popover is a few ticks wide. And `discover_menu_bar_index` runs before the startup timestamp, so its round trip is charged to startup.

**Body is the unedited template** — no summary, no verification, no `Refs #N`.

**Impact:** every PR's CI goes red at the new step unless the runner already holds the grant.
**Merge:** needs a linked green Actions run of this step first.

Issue #1279, opened 2026-09-02
Letter-mnemonic shortcuts for annotation tools in screenshot editor (in addition to number shortcuts)


khanhtd36:
## Problem

Tool switching in the screenshot editor currently uses number keys 1–9, where each number maps to *rail position* rather than a fixed tool identity (`ScreenshotSupport.Tool.shortcutTool(number:orderRaw:enabled:)`). That's great for muscle-memory once you've settled on a layout, but it means the key for "Text" or "Rectangle" changes depending on how the rail is currently ordered, and there's no way to jump straight to a tool by what it *is*.

Apps in the same space (e.g. CleanShot X) additionally offer fixed single-letter mnemonics — A for Arrow, R for Rectangle, T for Text, etc. — that stay constant regardless of any reordering. I'd like to add that as a second, opt-in mode here.

## Proposed design

**Settings control**: replace the current single "Use number shortcuts" toggle with a 3-way exclusive choice:
- No shortcuts
- Number shortcuts (existing behavior, unchanged)
- Letter shortcuts (new)

Not a toggle + separate type dropdown — a 3-way picker avoids a dead/meaningless state (dropdown value sitting there while the toggle is off), and it keeps the per-row shortcut badge in `ScreenshotToolOrderControls.swift` showing exactly one badge per tool instead of two.

**Letter shortcuts are fixed to tool identity, not rail position** — that's the entire point of a mnemonic (T is always Text no matter where it sits in the rail). Bare letter keys, no modifier, matching the existing bare-number convention and the precedent set by CleanShot X, Photoshop, Figma, etc.

**Suppressed while text entry is active** — same as the number shortcuts already are conceptually; `handleKey` already branches on editor state (`model.tool == .crop`, `model.selectedWordIndexes`, etc.), so gating on "is a text field / the Text tool currently capturing keystrokes" fits the existing pattern rather than introducing new machinery.

**Fixed defaults for v1, not user-remappable** — a full remapping UI is its own feature with its own conflict-validation logic (two tools can't share a letter). Shipping fixed defaults first keeps this change small; a customization UI can be a follow-up if there's demand.

## Proposed default mapping (all 13 tools)

| Tool | Letter |
|---|---|
| Select | S |
| Arrow | A |
| Rectangle | R |
| Ellipse | E |
| Line | L |
| Freehand (Draw) | D |
| Text | T |
| Highlight | H |
| Pixelate | P |
| Redact | X |
| Crop | C |
[...]

PR #1285, opened 2026-09-02
feat(screen-capture): add letter-mnemonic shortcuts for annotation tools

khanhtd36:
## What changed

Adds letter-mnemonic shortcuts (S/A/R/E/L/D/T/H/P/X/C/K/N) as an alternative to the existing rail-position number shortcuts in the screenshot editor, per the discussion on #1279.

- `ScreenshotSupport.Tool.shortcutLetter` — a fixed per-tool letter, independent of rail order (unlike numbers, which follow whatever position a tool sits in).
- A new `screenshotToolShortcutStyle` key (`"number"` / `"letter"`) selects which scheme is active, kept **separate** from the existing `screenshotToolShortcutsEnabled` bool rather than merged into one 3-way value — @PathGao caught that my first draft would have repurposed that bool's storage and silently reset every existing user's on/off preference. Two independent keys means an existing install's choice survives untouched; the new key just falls back to its registered default (`"number"`) for everyone who's never seen it, both on a fresh launch and on importing an old settings backup — traced through `Defaults.swift`'s `register(defaults:)` and `SettingsBackup.applyAndRelaunch`, not verified by running an actual pre-feature install through the upgrade.
- `event.characters` (the layout-printed key), not `keyCode`, decides which letter fired — consistent with #1047, where reading the physical keyCode instead of the printed character was the bug.
- The on-canvas rail and the Settings pane both get a style picker: the wider Settings pane keeps a segmented control, the narrow in-editor popover uses a compact menu instead, since several locales' labels (German, Turkish) overflow a segmented control at that width.
- `Tests/MetricsTests.swift` gained coverage for the enabled/style gating and the registered-default/backup-key expectations.

## What I did not verify myself

- Non-English strings in `ScreenshotStrings.swift` (12 of 13 locales) are my own translations, not a native speaker's — worth a review pass before merge.
- Backward compatibility for an existing install was verified by reading the registration and backup-import code paths, not by exercising a real pre-feature install through an upgrade.

## Tested on

MacBook, macOS 26 (Golden Gate), this branch's own build (not a release build), English UI. `./build.sh` finishes without warnings, `--selftest` passes, `./build.sh --test` passes (9601 checks). Manually exercised both style pickers (Settings pane and in-editor popover) and letter-key tool switching in the editor.

## Screenshots

Image

Image

Refs #1279


PathGao:
The two keys staying separate is right, and the compat claim holds: `exportKeys()` derives from `registeredDefaults`, so there's no list to update. I grepped `main` — the arity change on `shortcutNumber`/`shortcutTool` reaches every call site, none missed.

Letter mode is inert on non-Latin layouts. `shortcutLetter` is ASCII a–z and `event.characters` is the layout-printed key, so on a Russian, Greek or kana layout `S` arrives as `с`/`σ`/`ん` and matches nothing — while the rail still draws the `S` badge. #1047 was about Latin layouts printing letters elsewhere; that doesn't extend to alphabets that don't print them at all. And this ships letter mode translated into ru/ja/ko/zh, which is exactly where the gap lands. Cheapest fix: fall back to the ANSI keyCode when the printed character matches nothing.

Same line: `Character(character.lowercased())` traps if a lowercasing ever yields more than one grapheme cluster. Compare `String`s instead, like `ScreenCaptureTool.matchingShortcut` already does.

`shortcutControl(for:)`: with shortcuts off, or in letter mode, the numbered-slot menu is gone; before it stayed clickable at 0.48 opacity. Reordering is arrows-only now. Not in the body — intentional?

Didn't check the translations; can't test real keyboard layouts.

**Impact:** letter mode does nothing on non-Latin layouts while the badge claims it does. Number mode unaffected, so no existing user regresses.
**Merge:** needs the keyCode fallback, or an explicit known-limitation line, first.

ref #1279

khanhtd36:
Pushed 20eabe0 addressing all three.

**Non-Latin layouts**: `shortcutTool(character:keyCode:orderRaw:enabled:style:)` now tries `character` first (unchanged, still respects #1047), then falls back to `Tool.shortcutKeyCode` — the ANSI keyCode of the letter's US-QWERTY position — when the printed character matches nothing. Same convention `GlobalShortcut.swift`'s own defaults already use for physical-position bindings. Caveat I can't get around: the rail badge still shows the Latin letter (e.g. "S") even on a layout where that key doesn't produce an "s" — it's now a position label there, not a literal one, same tradeoff the pre-existing number badges already accept implicitly. Open to a different call if you think that's worse than no letter mode there at all.

**Crash risk**: fixed — `shortcutTool(letter:)` now compares `String(character).lowercased()` against `String(tool.shortcutLetter)` directly, no `Character(String)` reconstruction.

**Menu going inert while off**: restored — `shortcutControl(for:)` shows the same interactive number menu / letter badge regardless of `enabled`, just dimmed via `.opacity`, matching what it did before this PR. That was an unintentional regression from simplifying the view, not a deliberate call.

Added coverage in `MetricsTests.swift` for the keyCode fallback and the enabled/style gating. Still true from the PR description: I haven't tested this on a real non-Latin keyboard layout, only traced the logic — same limitation you flagged on your end.

khanhtd36:
Closing this — #1492 covers the request with a better design than what I proposed here (fully remappable bindings per tool beats a fixed letter/number toggle, and it solves the same "4 of 13 tools have no key" problem without the locale/keyboard-layout tradeoffs this PR carried).

Thanks @PathGao for the review round on this — the two-key storage split, the non-Latin-layout keyCode fallback, and the crash fix all came from that, and I appreciate the thoroughness. And thanks @santi-ug for #1492, nice work.

Issue #505, opened 2026-08-05
Cannot detect peripheral device battery info


subhamayd2:
### Description

I am using Razer Basilisk Mobile connected via bluetooth.

### Expected
Show the battery information of the mouse

### Actual
Cannot detect peripheral device and no battery info is shown

### Steps to Reproduce

Show battery menu bar item

### Screenshots / Recordings

<img width="329" height="453" alt="Image" src="https://github.com/user-attachments/assets/84ea9375-d4b9-4cf4-90c0-a4e2094aba62" />

### Vorssaint Version

3.3.0

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Homebrew

PR #1294, opened 2026-09-02
Some additions I made that real developers can work from

wakilibaraka:
## Summary

The menu-bar disk metric only shows percentage used (e.g. `DSK 88%`). This adds a display-style choice so it can show **free space** (`24 GB`), a percentage, or both — defaulting to free.

This mirrors an option that already exists: the Memory metric has a dot/percent/both style (`MemoryMenuBarStyle`) with a picker in Monitor settings. The disk panel *already* shows free space via `MetricFormat.diskBytes(disk.freeBytes)`, so today the menu bar and the panel disagree on which figure matters; this aligns them.

Change: a `DiskMenuBarStyle` (free / percent / both) read from a new `menuBarDiskStyle` default, applied in the `.diskUsage` branch of the menu-bar renderer, with the reserved width widened to fit `999 GB`. A segmented picker is added to the disk section of Monitor settings, shown only when the disk metric is pinned. Three new picker strings (Free / Percentage / Both) are added to the `Strings` catalog for every supported language.

Why it belongs:
- **Stands on:** existing `snapshot.disk` sampling (`freeBytes`/`usedFraction`) — nothing private, no new sampling.
- **Exposes:** nothing — no permission, no network, no new distribution surface.
- **Drags in:** nothing; plain `swiftc`, no dependency.
- **Surface vs. audience:** one enum, one defaults key + sanitizer, one picker — the same shape already carried for the Memory metric.

## Verification

Built and run on <your Mac model, macOS version + build; release build or your own signing identity>.

- Toggled Free / Percentage / Both and confirmed the menu bar updates (`DSK 24 GB`, `DSK 88%`, and both) and widens without clipping.
- Confirmed the default is Free on a fresh defaults domain, and that a corrupt stored value falls back through the sanitizer.
- Confirmed the picker only appears when the disk metric is pinned.

```sh
./build.sh                     # finished without warnings
./build.sh --test
```

New picker strings compile in every supported language.

What I could not test: other Mac models, notched menu bars with many metrics pinned at once, and locales beyond spot-checking the layout in <the languages you actually checked>.

## Not a developer


PathGao:
Thanks for taking the time on this, but I can't take it in this shape.

Three things:

1. The branch has 30 `patch_*.py` scripts and a `build.log` committed alongside the Swift changes. Those are your local scratch files — `patch_junk_cleaner.py` even patches a file that isn't in the diff any more.
2. The description only covers the disk menu-bar style, but the diff also adds a battery hero view, a clipboard HUD, an inactive-apps service, peripheral alerts, and notifier changes. That's five separate PRs.
3. The Verification section still has the template placeholders in it (`<your Mac model, macOS version + build>`, `<the languages you actually checked>`), so there's no record of this having been run anywhere.

The disk idea itself is worth having — the panel already shows free space while the menu bar only shows a percentage, and those disagreeing is a real inconsistency. Please open an enhancement issue describing just that, and we can discuss it there: https://github.com/vorssaintapp/vorssaint-utils/issues/new?template=feature_request.yml

Closing this one.

PathGao:
I checked before suggesting it — there is no existing issue for this. Nothing in the open enhancements covers showing free space in the menu bar instead of a percentage, so the issue is yours to open, not a duplicate.

PathGao:
The battery cache key picking up `isColored` is right — that's where a stale-icon bug hides.

**Scope.** The body describes one thing (disk menu-bar style); the diff ships six, one of which quits other people's apps. The Stands on / Exposes / Drags in analysis is true of the disk option and false for the rest. Also drop `build.log` and the 12 `patch_*.py`/`fix_*.py` scripts. Split it — the disk option alone lands fast.

**Defaults.** All four `bk*` keys default `true`. On upgrade every user gets a sound and a HUD on every copy, a recolored menu bar, and the hero view replacing the power panel, which discards their block order and per-block visibility. New behavior ships off.

**Two features can't fire.** `postInactiveApp` uses category `inactiveApp`, but main's only `setNotificationCategories` sits in `postWhatsAppOrganization` and replaces the whole set — nothing registers `inactiveAppQuitAction`, so the Quit button never appears. `PeripheralAlertService` reads `snapshot.peripheralBatteries` without requesting sampling, and `SystemMonitor` only samples it when a panel or menu-bar metric asks — so the alert fires only while the power panel is open.

Unverified: I have the first 80 KB of the diff only, so I haven't read the string catalog — `DiskMenuBarOrderOption`'s picker labels are hardcoded English regardless.

**Impact:** every user on upgrade — copy sound, copy HUD, recolored menu bar, power panel replaced.
**Merge:** needs the scope split first.

ref #505

PR #1295, opened 2026-09-02
fix(homebrew): bound the streaming command's wait

PathGao:
## Summary

`runStreaming` waited on the child with `waitUntilExit()`, which takes no timeout. That wait runs on `HomebrewManager`'s serial `workQueue`, so one stuck `brew install` parked every later `refreshInstalled`, `search`, `refreshOutdated`, `select` and `packageManagingApplication` behind it for good.

`runStreaming` now waits on a semaphore its `terminationHandler` signals, bounded by a new `brewOperationTimeout` of 30 minutes, and escalates through the existing `stop(_:finished:)` when the bound expires. This is the same shape `run` already uses a few lines above with `brewReadTimeout`. The completion status is `-1` if the child somehow survives the escalation, again as in `run`. When the bound trips the operation card goes to Failed, not Cancelled, because nobody cancelled it.

This is the same unbounded-wait shape #971 describes, at a site that issue never names. It starves this one serial queue rather than the shared pool, so it is not referenced as a fix for #971. After this lands `main` still has two bare `waitUntilExit()` calls, in `JunkCleaner` and `DiskImageInstallerService`, so #971 is not one change away from done.

Considered and not done, compared with #1237: the 4 MB output tail, the 64 KB log cap and the `static let` regex hoist. None of them had an observed need. The output buffer is a local of one operation and is released when it ends, the log is already cleared by `scheduleCompletedOperationCleanup` a few seconds after every outcome, and the regex cost was never measured. They are separate changes if anyone sees them bite.

## Verification

Mac17,3, macOS 26.6.2 (25G83), local build from this head.

```sh
./build.sh                     # 0 warnings
./build.sh --test              # TESTS OK (9609 checks)
```

`./build.sh --test` does not compile `HomebrewManager.swift`, so the new assertion is a source-shape check: `runStreaming` contains `brewOperationTimeout` and no `waitUntilExit`. It is red on `main`, where neither holds.

Not tested: none of it against a real slow or hung `brew`. The timeout path and the SIGTERM/SIGKILL escalation were read from `run` and `stop(_:finished:)`, not exercised, and the Failed-not-Cancelled routing was traced through `perform` rather than observed. No GUI testing was done.

## Anything else

Supersedes #1237.

### DIFF
--- a/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
+++ b/Sources/Vorssaint/Services/Homebrew/HomebrewManager.swift
@@ -682,8 +682,10 @@ final class HomebrewManager: ObservableObject {
     /// make them stall indefinitely.
     private static let brewReadTimeout: TimeInterval = 30
     /// Install/upgrade/uninstall run on the same serial queue as every read, so a
-    /// hung one must end eventually; a large cask on a slow line takes minutes, not hours.
-    private static let brewOperationTimeout: TimeInterval = 30 * 60
+    /// hung one must end eventually. A big download, a source build and a copy into
+    /// Applications are all slow but never silent, so the bound is on silence: a cap
+    /// on total time would end work that was still going.
+    private static let brewSilenceTimeout: TimeInterval = 15 * 60
     private static let processTerminationGrace: TimeInterval = 2
 
     /// SIGTERM is cooperative. Escalate only when a command ignores it so a
@@ -761,6 +763,7 @@ final class HomebrewManager: ObservableObject {
             process.standardOutput = pipe
             process.standardError = pipe
             var output = Data()
+            var lastOutput = Date()
             let lock = NSLock()
             let drained = DispatchSemaphore(value: 0)
             pipe.fileHandleForReading.readabilityHandler = { handle in
@@ -771,6 +774,7 @@ final class HomebrewManager: ObservableObject {
                 }
                 lock.lock()
                 output.append(data)
+                lastOutput = Date()
                 lock.unlock()
                 if let chunk = String(data: data, encoding: .utf8) {
                     DispatchQueue.main.async { onOutput(chunk) }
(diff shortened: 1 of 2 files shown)


PathGao:
Right call not to `ref #971`. `runStreaming` now matches `run` line for line — same semaphore, same `stop(_:finished:)` escalation, same `isRunning ? -1` status — so there is nothing left to review here that `run` did not already settle. I traced the Failed-not-Cancelled claim through `perform`: a non-zero status with `cancelRequested == false` falls past the `.cancelled` branch, so it holds.

One thing worth keeping visible: after this lands, `main` still has two bare `waitUntilExit()` calls, in `DiskImageInstallerService` and `JunkCleaner`. Anyone reading #971 as one PR away from done is wrong — it needs those two as well. That is the reason the ref stays off, and it should be said somewhere the next person will find it.

Not verified: nothing was run against a real hung `brew`. The 30-minute bound and the SIGKILL escalation were read, not exercised.

**Impact:** none until a `brew` operation stalls past 30 minutes; then the Homebrew panel recovers instead of staying dead until the app is restarted.
**Merge:** ship it.

PathGao:
@PathGao body now names the two remaining bare `waitUntilExit` sites; no code change. Please re-review.

PathGao:
Body now carries the thing that was only in a thread comment before, which is where it belongs — the next person reads the PR, not the discussion.

Checked the claim against `main`: `JunkCleaner.swift:293` and `DiskImageInstallerService.swift:383` are the only bare `waitUntilExit()` left under `Sources/` once this lands. There is a third in `Tests/MetricsTests.swift` (`detachShell`), test-only and not on any pool #971 cares about — saying it only so the count in the body is not read as the whole grep.

One thing the body could point at while it is naming sites: `BoundedProcessRunner` already implements this exact shape and names #971 in its own comment. It cannot take `runStreaming` as it stands — no streaming callback, no handle for cancel — so there is nothing to change here, but it is the obvious landing point for those two remaining sites, and this makes the third hand-rolled copy of the pattern inside `HomebrewManager`. Naming it saves the next person the search.

Diff unchanged since the last pass, verdict unchanged.

**Impact:** none until a `brew` operation stalls past 30 minutes; then the Homebrew panel recovers instead of staying dead until the app is restarted.
**Merge:** ship it.

PR #1296, opened 2026-09-02
perf(command-bar): fold each list of rows once per opening

PathGao:
## Summary

One opening of the command bar folded every row's title and keywords seven to
ten times.

`index(_:)` folds each row once "per rebuild" so that typing never pays for it,
but it is handed the flat concatenation of nine lists, and those lists arrive at
different moments: the catalog and the running apps when the panel appears, then
the installed apps, the Mac's Settings panes, the open windows, the front app's
menus, the selection and the running processes each from their own background
pass, and each one calls `indexEntries()` on the way in. Every call refolded all
nine lists - 1000 to 1500 rows, two foldings each - in the few hundred
milliseconds while the panel was appearing and the person started typing.

Each list now keeps its own folded copy and only the list that was assigned
again is folded. The invalidation is the assignment itself: each of the nine
stored properties carries a `didSet` that drops its copy, so no write site has a
separate step to forget. `PoolSection` names the nine lists in the order the old
concatenation had, so `freshFullEntry`'s `.last` still picks the same row.

The three lookup maps are still built from nothing on every call, deliberately:
an id that is no longer in any list cannot reappear, so the menus of the app
that was in front a moment ago cannot stay pressable. `aliasCache` clears every
copy when it changes, since a name the person gave is folded in.

Considered and not done: #1245 also stopped `rebuildCatalog()` from rebuilding
the emoji rows unless the language or the Accessibility grant moved. The body
there put those rows at about fifteen hundred; `CommandBarEmoji.emojiCharacters`
on `main` is 204 characters, and `CommandBarEmoji.emoji` is a `static let`, so
the Unicode names are resolved once per launch and `emojiEntries(bar:)` is a map
over 204 structs. With the folding already down to once per list per opening,
skipping those 204 of 1000-1500 rows is worth at most about 1.2x, and it was
where every review finding on that PR landed (a stale permission warning after
a grant, the store left unpinned, the uninstall path leaning on `isEmpty`). Not
worth its guard, its extra field or its source-shape test.

A tenth `PoolSection` case whose array carries no `didSet` would keep its first
folded copy for the life of the service, and because the identity maps are built
inside the same `zip`, rows that list grew by would fall out of `entriesByID` and
`entriesByStableKey` too. Nine cases, nine `didSet`s today; the `didSet` is the
invariant, not a nicety.

`foldedSections` holds only the folded text per section; the rows themselves are read back through `entries(in:)` when the maps are built, zipped by position. The `didSet` that drops a section's fold whenever its list is assigned is what keeps the two the same length and order, so no second copy of the entries lives for the life of the service.

## Verification

Mac17,3, macOS 26.6.2 (25G83). `./build.sh` clean, no warnings;
`./build/Vorssaint --selftest` `SELFTEST OK`; `./build.sh --test` `TESTS OK (9608 checks)`.

`CommandBarService.swift` is outside the `--test` whitelist, so the test binary
does not compile it; the full build does. No test added: the change is in one
file the test target does not see, and there is no observable to pin by source
shape that the build itself does not already check.

Not tested: no run on real hardware. The count of index passes per opening is
read off the call graph, not measured. Opening the bar with menus, windows and a
selection all landing, and renaming a row while it is open, were not exercised
by hand.

## Anything else

No issue reports the cost this removes; it came out of reading the opening path.

Supersedes #1245.

### DIFF
--- a/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
+++ b/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
@@ -888,8 +888,7 @@ final class CommandBarService: ObservableObject {
     /// Each list's folded copy, dropped by the `didSet` when that list is
     /// assigned again. The lists land from separate background passes, each
     /// calling `indexEntries()`, so one opening indexes seven to ten times.
-    private var foldedSections: [PoolSection: [(entry: CommandBarEntry,
-                                                folded: (title: String, keywords: String))]] = [:]
+    private var foldedSections: [PoolSection: [(title: String, keywords: String)]] = [:]
 
     private func indexEntries() {
         index()
@@ -912,9 +911,8 @@ final class CommandBarService: ObservableObject {
                 // the row surfaces even when its real title shares nothing with it.
                 let alias = names[entry.stableKey]
                     .map { " " + CommandBarSearch.normalized($0) } ?? ""
-                return (entry,
-                        (CommandBarSearch.normalized(entry.matchTitle ?? entry.title),
-                         CommandBarSearch.normalized(entry.keywords) + alias))
+                return (CommandBarSearch.normalized(entry.matchTitle ?? entry.title),
+                        CommandBarSearch.normalized(entry.keywords) + alias)
             }
         }
         // The three maps are built from nothing every time, so a row that left
@@ -924,10 +922,10 @@ final class CommandBarService: ObservableObject {
         normalizedByID = [:]
         entriesByStableKey = [:]
         for section in PoolSection.allCases {
-            for row in foldedSections[section] ?? [] {
-                entriesByID[row.entry.id] = row.entry
-                entriesByStableKey[row.entry.stableKey] = row.entry
-                normalizedByID[row.entry.id] = row.folded
+            for (entry, folded) in zip(entries(in: section), foldedSections[section] ?? []) {
+                entriesByID[entry.id] = entry
+                entriesByStableKey[entry.stableKey] = entry
+                normalizedByID[entry.id] = folded
             }
         }
     }
--- a/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
+++ b/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
@@ -890,17 +890,15 @@ final class CommandBarService: ObservableObject {
     /// calling `indexEntries()`, so one opening indexes seven to ten times.
     private var foldedSections: [PoolSection: [(title: String, keywords: String)]] = [:]
 
-    private func indexEntries() {
-        index()
-    }
-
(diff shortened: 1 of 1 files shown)


PathGao:
`didSet`-as-the-invalidation is the right call. I grepped every write site of the nine lists in `Sources/`: all plain assignments, all in this file, so nothing mutates a list without dropping its copy. `allCases` order matches the old concatenation, so `freshFullEntry`'s `.last` is unchanged.

`hide()` calls `indexEntries()` twice — after clearing `selectionEntries`, then `killProcessEntries` — and `clearIndex()` six lines down wipes both results. `isTearingDown` is already true, so `query = ""` short-circuits `refreshResults()` and nothing reads the maps in between. Post-PR they stop refolding but still rebuild three dictionaries over the whole pool, twice per close, for nothing. Pre-existing on `main`, same shape as what you're removing here, and it's a delete.

`foldedSections` keeps a second copy of every entry for the life of the service; `clearIndex()` used to release the only copy. `entries(in:)` is right there with the same count and order under the same invariant:

```swift
private var foldedSections: [PoolSection: [(title: String, keywords: String)]] = [:]
for (entry, folded) in zip(entries(in: section), foldedSections[section] ?? [])
```

Not verified: no hardware, and I read the pass count off the call graph, same as you.

**Impact:** none, the maps come out identical.
**Merge:** ship it.

PathGao:
@PathGao taken: `foldedSections` now stores only the folded text and the maps are built by zipping `entries(in:)` with it, 7+/9-. Full build 0 warnings, selftest OK, `--test` 9608. Please re-review.

PathGao:
Zipping against `entries(in:)` is the right shape, and the fold is now two strings a row instead of a second copy of the pool. Re-grepped `main`: nine lists, nine `didSet`s, every write a plain assignment in this file.

The body understates the tenth-case hazard now. It says a case without a `didSet` "would keep its first folded copy" — stale text. But `entriesByID` and `entriesByStableKey` are built inside the same `zip`, so if such a list *grows*, the extra rows fall out of both identity maps too. `entriesByStableKey` gates `runDeferredRowShortcutIfReady` and `entriesByID` gates `canOpenActions`: a stale fold ranks wrong, a missing key is a row whose shortcut never fires and whose actions panel will not open. Only `normalizedByID` needs the pairing:

```swift
let folded = foldedSections[section] ?? []
for (offset, entry) in entries(in: section).enumerated() {
    entriesByID[entry.id] = entry
    entriesByStableKey[entry.stableKey] = entry
    if offset < folded.count { normalizedByID[entry.id] = folded[offset] }
}
```

That `?? []` is unreachable today — the loop above it fills every case.

Not verified: no hardware, no build run.

**Impact:** none today, the maps come out identical.
**Merge:** ship it.

PR #1297, opened 2026-09-02
test: enumerate event-tap owners instead of listing six by hand

PathGao:
## Summary

The session-switch guard in `Tests/MetricsTests.swift` walked a hand-written list of files: six when this was opened, nineteen on `main` today. `main` has 24 files that call `CGEvent.tapCreate`, and a service adding a tap tomorrow inherits nothing from the list. The taps that never joined the session gate were exactly the ones missing from it.

The loop now reads its subjects out of `Sources/`: every `.swift` file whose comment-stripped source contains `CGEvent.tapCreate` is asked the three session questions the guard already asked (`SessionActivity.shared.onChange`, `SessionActivity.shared.isActive` between every `tapDisabledByTimeout` and its `return`, and an Accessibility check). There is nothing to register any more, and no exemption set: with #1304 and #1270 on `main`, every modifying tap owner passes, so there is nothing to exempt and no list to go stale. The enumeration is the same one the port sweep from #1225 already did, so that sweep now walks the shared list instead of its own; both sweeps are rooted at `Sources` rather than `Sources/Vorssaint`, so a tap landing in `HIDEventSystem` or `VMStatisticsCompat` gets the same questions. A `>= 20` floor keeps an empty walk from passing in silence, and each owner must have at least one `tapDisabledByTimeout` block, or the per-site walk would pass on nothing.

Two things follow from enumerating:

- Listen-only taps (`.listenOnly` and never `.defaultTap`) skip the session checks. They hand every event back unchanged, so a switched-away session cannot swallow the on-screen account's input. Flip one to `.defaultTap` and it is checked on the next run. They are still asked for the port by the invalidate sweep.
- The Accessibility assertion accepts either `SessionActivitySupport.tapShouldRun(` or `AXIsProcessTrusted()` anywhere in the file, on all 21 modifying owners. `main` asks six owners by name for `AXIsProcessTrusted()` inside the re-arm slice and the other fifteen nothing. Measured on `main`: 22 of the 23 re-arm sites already carry `AXIsProcessTrusted()`; the one that does not is `ScrollInverter`, which re-enables on `SessionActivity.shared.isActive` alone and asks `tapShouldRun` only on start. A slice-level check for every owner is one `Sources` line away, in `ScrollInverter`, and is not this change.

Considered and not done: the earlier version of this change (#1230) carried a harm description next to each ungated path, and the version reviewed here kept an exemption set with a staleness check. Both existed for owners that were outside the gate; none are left, so the set is deleted rather than kept empty. A future ungated tap fails the sweep directly, which is the rule the set was stating. The invalidate sweep keeps a per-tap count rather than the per-file `CFMachPortInvalidate` check the roster had, since three owners hold more than one tap.

## Verification

Mac17,3, macOS 26.6.2 (25G83), own build from this branch on `main` (`4dd067de`).

`./build.sh --test`: TESTS OK (9699 checks) on this branch, against 9694 on `main`, zero failures either side.

Falsified rather than only run. A throwaway `Sources/Vorssaint/Services/ZZProbeTapService.swift` holding nothing but a bare `.defaultTap` `CGEvent.tapCreate` and no teardown turns four checks red naming that file (the three session questions plus the port sweep reporting `1 taps, 0 invalidated` across 25 owners). Deleting the `SessionActivity.shared.onChange` subscription from `RadialMenuService`, the owner that joined on this rebase, gives one red from this sweep naming the file, beside the one #1270's own test raises on the same subscription. Replacing `SessionActivity.shared.isActive` with `true` in `BrightnessService`'s second `tapDisabledByTimeout` block (`handleKeyEvent`, the one a first-slice read never sees) gives exactly one red naming the file. Renaming every `tapDisabledByTimeout` in `KeyboardDebounceService` gives exactly one red saying it no longer handles its tap being disabled. All restored, back to green.

Census on `main`: 24 owners, all under `Sources/Vorssaint`; 3 listen-only (`MusicLaunchBlocker`, `AutoQuitService`, `DockPreviewService`); the remaining 21 pass every check across 23 re-arm sites, including `CleaningModeManager` and `MouseClickDebounceService`, which the hand-written list never reached.

Not tested: nothing here runs a tap. A tap handed back across a fast user switch needs a second logged-in account and was not exercised. The full `./build.sh` and `--selftest` were not run; only the test file changes, so nothing outside the `--test` compile is affected.

## Anything else

Refs #1153

Supersedes #1230.

### DIFF
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -22212,49 +22212,44 @@ struct MetricsTests {
             encoding: .utf8)) ?? ""
         expect(mouseAccelerationSource.contains("SessionActivitySupport.isOnConsole("),
                "mouse acceleration shares the safe initial session-state fallback")
-        for tapOwner in ["Sources/Vorssaint/Services/ScrollInverter.swift",
-                         "Sources/Vorssaint/Services/SmoothScrollService.swift",
-                         "Sources/Vorssaint/Services/MouseNavigation/MouseNavigationService.swift",
-                         "Sources/Vorssaint/Services/MouseButtons/MouseButtonShortcutService.swift",
-                         "Sources/Vorssaint/Services/MiddleClick/MiddleClickService.swift",
-                         "Sources/Vorssaint/Services/QuitProtection/QuitProtectionService.swift",
-                         "Sources/Vorssaint/Services/RadialMenu/RadialMenuService.swift",
-                         "Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift",
-                         "Sources/Vorssaint/Services/WindowMaximizer.swift",
-                         "Sources/Vorssaint/Services/Finder/FinderCutPaste.swift",
-                         "Sources/Vorssaint/Services/Finder/FinderRenameService.swift",
-                         "Sources/Vorssaint/Services/KeyboardDebounce/KeyboardDebounceService.swift",
-                         "Sources/Vorssaint/Services/SuperKey/SuperKeyService.swift",
-                         "Sources/Vorssaint/Services/ShortcutRecordingTap.swift",
-                         "Sources/Vorssaint/Services/Switcher/AppSwitcher.swift",
-                         "Sources/Vorssaint/Services/Snippets/TextSnippetService.swift",
-                         "Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift",
-                         "Sources/Vorssaint/Services/DockClick/DockClickService.swift",
-                         "Sources/Vorssaint/Services/Display/BrightnessService.swift"] {
-            let source = (try? String(contentsOfFile: tapOwner, encoding: .utf8)) ?? ""
-            expect(!source.isEmpty, "\(tapOwner) reads back for its session-switch check")
-            let code = source.components(separatedBy: "\n")
-                .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
-                .joined(separator: "\n")
+        // The owners are read out of the tree, not listed by hand, so a new
+        // CGEvent.tapCreate has nothing to register. One enumeration feeds
+        // both sweeps below, rooted at Sources so every target is asked.
+        let tapOwnerFiles: [(path: String, code: String)] = (FileManager.default
+            .enumerator(atPath: "Sources")?.compactMap { $0 as? String } ?? [])
+            .filter { $0.hasSuffix(".swift") && !$0.contains(" 2") }
+            .sorted()
+            .compactMap { entry in
+                let path = "Sources/\(entry)"
+                guard let source = try? String(contentsOfFile: path, encoding: .utf8) else { return nil }
+                let code = source.components(separatedBy: "\n")
+                    .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("//") }
+                    .joined(separator: "\n")
+                return code.contains("CGEvent.tapCreate") ? (path, code) : nil
+            }
+        // An empty sweep would pass every check below in silence.
+        expect(tapOwnerFiles.count >= 20,
+               "the CGEvent.tapCreate sweep reaches the sources from the test's working directory")
+        for (tapOwner, code) in tapOwnerFiles {
+            // A listen-only tap hands every event back unchanged, so a session
+            // that is off screen cannot swallow the on-screen account's input.
+            if code.contains(".listenOnly") && !code.contains(".defaultTap") { continue }
             expect(code.contains("SessionActivity.shared.onChange"),
                    "\(tapOwner) rebuilds its tap when the session comes back")
-            let rearm = code.components(separatedBy: "tapDisabledByTimeout")
-                .dropFirst().first?.components(separatedBy: "return").first ?? ""
-            expect(rearm.contains("SessionActivity.shared.isActive"),
-                   "\(tapOwner) does not re-arm a disabled tap into a switched-away session")
-            if tapOwner.contains("MouseNavigation")
-                || tapOwner.contains("MouseButtonShortcut")
-                || tapOwner.contains("MiddleClick")
-                || tapOwner.contains("QuitProtection")
-                || tapOwner.contains("RadialMenu")
-                || tapOwner.contains("ShortcutRecordingTap") {
-                expect(rearm.contains("AXIsProcessTrusted()"),
-                       "\(tapOwner) does not keep a modifying tap alive after Accessibility is lost")
+            // Every site, not the first: two owners re-arm in two callbacks.
+            // And at least one, or the walk below passes on nothing.
+            let rearmSlices = code.components(separatedBy: "tapDisabledByTimeout").dropFirst()
+            expect(!rearmSlices.isEmpty, "\(tapOwner) handles its tap being disabled")
+            for slice in rearmSlices {
+                let rearm = slice.components(separatedBy: "return").first ?? ""
+                expect(rearm.contains("SessionActivity.shared.isActive"),
+                       "\(tapOwner) does not re-arm a disabled tap into a switched-away session")
             }
-            // Switching a tap off leaves the process owning it, which is what
-            // the window server waits on; teardown must invalidate the port.
-            expect(code.contains("CFMachPortInvalidate"),
-                   "\(tapOwner) hands its tap back rather than only disabling it")
+            // Either spelling of the permission half: the shared helper takes
+            // it as an argument, older services ask the API directly.
+            expect(code.contains("SessionActivitySupport.tapShouldRun(")
+                    || code.contains("AXIsProcessTrusted()"),
+                   "\(tapOwner) does not keep a modifying tap alive after Accessibility is lost")
         }
 
         // Disabling a tap and dropping the last Swift reference does not hand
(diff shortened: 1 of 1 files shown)


PathGao:
Enumeration is the right call, and one walk feeding both sweeps is better than what I'd have suggested. Census checks out on `main`: 24 owners under `Sources`, 13 listed, 3 listen-only, 8 gated.

Three carve-outs are each justified by a narrower argument than the check they drop.

**AX predicate.** `tapShouldRun(` takes `accessibilityGranted:` as a caller-supplied `Bool`, so `accessibilityGranted: true` satisfies "does not keep a modifying tap alive after Accessibility is lost". `main` requires the literal call on the four modifying taps; this loosens them. All 8 already carry one of two spellings:

```swift
expect(code.contains("AXIsProcessTrusted()")
        || code.contains("Permissions.shared.accessibility"),
```

**Staleness.** The `knownUngatedTaps` loop asks whether a path still creates a tap, never whether it still fails the gate. Gate a service without deleting its line and the exemption is permanent and silent — the rot the six-name list died of, and the in-flight rebase is the likely path there. Invert it: an entry must *not* contain `SessionActivity.shared.onChange`.

**Listen-only.** The argument is about swallowing input; the check dropped is `onChange`, about the tap coming back. DockPreview, AutoQuit and MusicLaunchBlocker have no `onChange`, so after a switch back those taps may stay dead until relaunch. Not verified on hardware.

**Impact:** none — tests only.
**Merge:** needs the AX predicate tightened first.

ref #1153

PathGao:
Staleness: taken. Each entry now has to still create a tap and still lack `SessionActivity.shared.onChange`; gating RadialMenu without deleting its line fails exactly one check naming the file. 9643 checks.

AX predicate: not taken. `tapShouldRun(accessibilityGranted: true)` is a spelling nobody has written and nothing observed; a check against it would pin an argument literal, not a property. Accepting `Permissions.shared.accessibility` would also readmit the polled mirror #1233's review ruled out.

Listen-only: not verified either way, and out of this change's scope. The three listen-only owners were never in the six-name roster, so nothing here loosens what `main` checks.

PathGao:
@PathGao the staleness inversion above is pushed (`d03cab43` → head), body updated; please re-review.

PathGao:
Staleness check does what it says, and the RadialMenu falsification is the right shape for it.

AX predicate: dropping it. `accessibilityGranted: true` is a spelling nobody has written, and asking you to guard against it is the theoretical corner I'd object to in the other direction.

The enumeration walks files, not sites. Two of the twenty checked owners have two `tapDisabledByTimeout` blocks — `BrightnessService` (`routeFunctionKey`, `handleKeyEvent`) and `WindowLayoutService` (`observeEdgeSnapEvent`, the gesture tap) — and the slice reads only the first. This diff gates all four, so it is green; delete `SessionActivity.shared.isActive` from `handleKeyEvent` and it stays green.

The body defers this to "once a `knownUngatedTaps` line is deleted". `RadialMenuService` has one site. The two that have two are checked owners, so it bites now.

```swift
for slice in code.components(separatedBy: "tapDisabledByTimeout").dropFirst() {
    let rearm = slice.components(separatedBy: "return").first ?? ""
    expect(rearm.contains("SessionActivity.shared.isActive"), ...)
}
```

Site counts are grep on `main`; not run.

**Impact:** none — tests only.
**Merge:** needs the re-arm slice walking every site first.

ref #1153

PathGao:
@PathGao the re-arm slice now walks every `tapDisabledByTimeout` site; breaking Brightness's second site gives exactly one red. 9645 checks. Please re-review.

PathGao:
The walk is right, and the check count is what proves it: 22 sites across the 20 non-exempt owners, +2 over the 24-file count — exactly the 9643 → 9645 you got.

The falsification doesn't test it, though. `routeFunctionKey` is Brightness's *first* `tapDisabledByTimeout` (805); `handleKeyEvent` is the second (970). Breaking the first is what the old single-slice form already caught. Break `handleKeyEvent` for the same one red and the claim is backed by the evidence line.

One thing the loop lost: on `main` a file with no `tapDisabledByTimeout` block at all was red — `.dropFirst().first ... ?? ""` fails on nothing. `for slice in ....dropFirst()` runs zero times and passes in silence. Same shape as the `>= 20` floor, one level down. All 24 owners have a block today, so it costs nothing:

```swift
expect(code.components(separatedBy: "tapDisabledByTimeout").count > 1,
       "\(tapOwner) handles its tap being disabled")
```

Site counts are grep on `main`; not run.

**Impact:** none — tests only.
**Merge:** needs the zero-block guard first.

ref #1153

PathGao:
@PathGao both taken. Zero-block guard added (`handles its tap being disabled`, one per owner, 9645 → 9665). Falsified on the right site this time: `handleKeyEvent` at line 977 is Brightness's second block; `true` in place of `isActive` there gives exactly one red. Renaming every `tapDisabledByTimeout` in `KeyboardDebounceService` gives exactly one red from the new guard. The body's earlier function name was wrong, fixed. Please re-review.

PathGao:
The walk lands, and the counts reconcile against `main`: 26 `tapDisabledByTimeout` occurrences across the 24 owner files, minus the 3 listen-only and RadialMenu, is 22 sites over 20 owners — and the zero-block guard is +20, which is your 9645 → 9665.

The body is stale in one place. "Considered and not done" still says the re-arm slice reads only the first site and belongs to the branch that deletes a `knownUngatedTaps` line; Verification, four paragraphs down, says it walks every site. That sentence is the reasoning that kept the gap alive for two rounds — drop it before the merge commit carries it.

Where the check stops: `WindowLayoutService` has three `CGEvent.tapCreate` sites and two `tapDisabledByTimeout` blocks. `directionalTap` (`startDirectionalTap` → `observeDirectionalEvent`) has no disable branch at all, and "at least one per file" cannot see that. `blocks >= taps` is not the fix — `SuperKeyService` re-arms both its taps from one block. That tap is modal and rebuilt per gesture, so this is a note, not a demand; nothing to add here.

Site counts are grep on `main`; not run.

**Impact:** none — tests only.
**Merge:** hold for #1304.

ref #1153

PathGao:
#1304 landed, so the Sources half of this diff is on main and the branch is down to `Tests/MetricsTests.swift`.

The staleness check earns itself on the rebase: #1270 landed too, `RadialMenuService` now has `onChange`, `tapShouldRun`, and `isActive` in its re-arm slice, so its exemption goes red the moment you rebase. The check firing on its own entry is exactly what it was built for. Delete the line; keep the empty set and its loop, since the rule it states is the point.

The body then contradicts the tree in three places: the `knownUngatedTaps` paragraph, the census line (`1 in knownUngatedTaps`), and the third falsification — adding `onChange` to `RadialMenuService` is `main`'s state now, not a probe. The "Considered and not done" sentence about the re-arm slice reading only the first site is still there as well.

`main` has 24 owners, 3 listen-only, so 21 checked. RadialMenu joins with one re-arm site: +4, -2 staleness, 9667 if nothing else moved.

Grep on `main`; not built.

**Impact:** none — tests only.
**Merge:** needs the rebase that empties `knownUngatedTaps` first.

ref #1153

PathGao:
@PathGao rebased onto `main` (`4dd067de`, one commit, `Tests/MetricsTests.swift` only). `RadialMenuService` went red on the rebase as predicted, and it took the whole set with it: nothing is ungated any more, so `knownUngatedTaps` and its staleness loop are deleted rather than kept empty. A future ungated tap fails the sweep directly, which is the rule the set was stating. 9694 on `main` → 9699 here: 21 owners, 23 re-arm sites (RadialMenu joins with one).

Falsified again on this tree: the probe file gives four reds, dropping `onChange` from `RadialMenuService` one from this sweep (plus #1270's own test on the same subscription), `true` in Brightness's second block one, renaming every `tapDisabledByTimeout` in `KeyboardDebounceService` one.

One thing to weigh that `main` moved under this: #1304 asks six owners for `AXIsProcessTrusted()` inside the re-arm slice; this checks either spelling anywhere in the file, on all 21. Measured on `main`, 22 of the 23 slices carry `AXIsProcessTrusted()` already. The one that does not is `ScrollInverter`, which re-enables on `isActive` alone. A slice-level check for every owner is one `Sources` line in `ScrollInverter`; not in a tests-only change, and the body says so.

Body rewritten for the tree: the exemption paragraph, census, third falsification, the #1304 dependency and the stale "first site" sentence are gone. Please re-review.

PathGao:
Counts reconcile against `main`: 24 owners, 3 listen-only, 26 `tapDisabledByTimeout` minus those 3 = 23 sites over 21 owners. Old loop was 4×19 + 6 AX + 1 sweep = 83, new is 1 floor + 3×21 + 23 + 1 = 88 — your +5. Deleting the set rather than keeping it empty is the right call.

The AX assertion is now weaker than the message it prints. `main` asks six owners for `AXIsProcessTrusted()` *inside* the re-arm slice; this asks 21 for either spelling anywhere in the file, and 32 files in `Sources` would pass that. `ScrollInverter` makes it concrete: `tapShouldRun` on start, re-arm on `SessionActivity.shared.isActive` alone — the sweep now asserts it does not keep a modifying tap alive after Accessibility is lost, and it does. Delete the AX call from `MiddleClickService`'s re-arm block and `main` goes red, this stays green.

Same shape as the `isActive` line already in the loop:

```swift
expect(rearm.contains("AXIsProcessTrusted()")
        || rearm.contains("SessionActivitySupport.tapShouldRun("), ...)
```

That costs the one `ScrollInverter` line you measured, so this stops being tests-only. Worth it: a check that names a file and states something false about it is worse than no check.

Read on `main`; not run.

**Impact:** none — tests only.
**Merge:** needs the AX check moved into the slice loop first.

ref #1153

PathGao:
Closing this. The taps it was written to find are gated on `main` now (#1304, #1270); what is left is test structure, and the next step pulls a `Sources` change in to keep the check honest. Not worth another round.

One thing from the measurement survives on its own: `ScrollInverter` re-enables its tap on `SessionActivity.shared.isActive` alone, while the other 20 modifying owners also ask `AXIsProcessTrusted()` in the re-arm block. That is a one-line fix if wanted, separate from this.

Refs #1153

PR #1298, opened 2026-09-02
fix(cleaner): keep the scheduled clean from asking for an administrator password

PathGao:
## Summary

A scheduled clean can raise an administrator password prompt at the hour the
user picked precisely so they would not have to be there.

The path: `CleanerPolicy.precheckLoginItems` is `true`, so orphaned startup
items start checked. `scanOrphanedLaunchPlists` scans `~/Library/LaunchAgents`,
`/Library/LaunchAgents` and `/Library/LaunchDaemons`, and the last two are root
owned. `CleanerScheduler.runAutomaticCleanup` takes the scan's pre-checked
selection as it stands and calls `cleanSelected()`. `trashItem` cannot move a
root-owned file, so those items fail into `stubborn`, and `stubborn` goes to
`trashViaFinder`, which is Finder's delete: the standard administrator prompt,
preceded on a first run by the automation consent dialog.

`cleanSelected` now takes `escalate`, with no default so every caller has to
say which it is (`WhatsAppDownloadManager.cleanSelected(automatic:)` has the
same shape). `CleanerScheduler` passes `false`: whatever the Trash move refused
stays where it is and counts as failed, in whichever pre-checked category it
sat. `CleanerView` passes `true` and the manual path is unchanged.

A scheduled clean therefore does less than the same selection by hand, and it
says so: `failed` now reaches `finishRun` and `notifyIfWanted`, and the outcome
notification appends the sentence the manual clean already shows for the same
fact, `uninstallerSomeFailed` ("Some items couldn't be moved to the Trash."),
which exists in all thirteen languages. No new string. "Nothing to clean" is
reserved for a pass that moved nothing and left nothing.

Considered and not done, compared with #1248: persisting the count to the
cleaner's last-run line (a `Defaults` key, `SettingsBackupSupport`, a
`CleanerView` block) had no observed case behind it, so the notification is the
one channel; a `CleanerPolicy.escalateOnAutomaticRun` constant had a single
caller, so the call site says `false` directly; a `leftInPlaceSentence` helper
and a new count-carrying string existed only to put a number in the sentence,
and the number is not what anyone acts on. What repeats is the same root-owned
plist being found and left every scheduled run; one manual clean is the way out,
and the sentence does not say so. That is a wording change for its own turn.

## Verification

Mac17,3, macOS 26.6.2 (25G83), Apple silicon, own build from this branch.
`./build.sh`: no warnings. `./build/Vorssaint --selftest`: SELFTEST OK.
`./build.sh --test`: TESTS OK (9610 checks).

New checks, each confirmed absent from `origin/main` before being trusted:
`CleanerScheduler` calls `cleanSelected(` exactly once, with `escalate: false`,
and passes `failed` to `notifyIfWanted`; `JunkCleaner` declares
`cleanSelected(escalate: Bool)` with no default and `CleanerView` passes
`escalate: true`.
The #1246 source marker moves from `"func cleanSelected()"` to
`"func cleanSelected("` so that slice does not come back empty.

Not tested: I never triggered the prompt. No schedule was armed, no password
dialog watched, no notification seen on screen, and no real file moved through
`trashItem`, so the refusal is derived from reading the chain rather than from
an observed error code. The notification body was not measured against
Notification Centre's truncation in the longer languages.

## Anything else

No issue tracks this; the open "Cleaner exceptions" request (#1182) asks for a
user-editable exception list and is not this.

Supersedes #1248.


PathGao:
Both call sites covered, and the rest of the prompt surface holds: `cleanSelected` has exactly two callers in `main`, `emptyTrash` is the only other consent-raising path in `JunkCleaner`, and `scanTrash` ships `recommended: false`, so the unattended pass can't reach that one either. `WhatsAppDownloadManager.cleanSelected(automatic:)` never escalates. No sibling gap.

On the new string: `uninstallerSomeFailed` ("Some items couldn't be moved to the Trash.") already exists in all 13 languages, and `doneState` already shows it for this exact fact after a manual clean. `cleanerAutoLeftFormat` buys a number for 13 new translations plus a format check. Unless the count is the thing you want acted on, reusing the existing sentence makes both paths say the same thing for free.

Either way it repeats: the same root-owned plist is re-found and re-fails every scheduled run, and the sentence names no way out (one manual clean is it). Cheap to say while the strings are open.

Static only — I read `main` for the call sites, didn't build or run, so I never saw the prompt either.

**Impact:** none beyond the change itself; both notes are cost and wording.
**Merge:** ship it.

PathGao:
@PathGao taken: the notification reuses `uninstallerSomeFailed` and the new string is gone (1+/16-, 14 files fewer lines, no new translations). The repeat-every-run wording is noted in the body as a separate change. Full build 0 warnings, selftest OK, `--test` 9610. Please re-review.

PathGao:
Both taken and the diff got smaller. Nothing new to flag.

Re-checked `main` for the one thing that matters — whether anything else in the unattended pass can still raise a prompt:

- `trashViaFinder` is the only Finder path `cleanSelected` can reach, and `escalate: false` skips it.
- `emptyTrash` sits behind the trash row, which `scan` builds with `recommended: false`, so an unattended pass never selects it and never reaches its automation consent.
- `bootoutUserAgent` bails unless the plist is under `~/Library/LaunchAgents` and then runs `launchctl bootout gui/$UID`. No prompt.

One note, no action: `failed` is not only what escalation would have taken — a `mayRemove` rejection lands in it too. `uninstallerSomeFailed` is true for all of them, so reusing it is still right; the count is just broader than the body's wording implies.

Static only, same as last time: I read `main`, never armed a schedule, never saw the prompt or the notification.

**Impact:** none.
**Merge:** ship it.

PR #1299, opened 2026-09-02
fix(brightness): check the display fingerprint before restoring gamma

PathGao:
## Summary

`GammaTable` carries a `fingerprint` because display numbers are handed out again after a reconnection, and the remembered curves are deliberately kept across a disconnection. So "this number now belongs to a different monitor" is a state the code expects. Four places in `BrightnessService` write a curve. Three checked the fingerprint first. The fourth, the restore that runs just before a display is switched off, took the table by display number and wrote it, so a dimmed baseline could land on whichever monitor now holds that number.

The fix is the missing condition at that site: the restore now runs only when `baseline.fingerprint == Self.displayFingerprint(display.id)`, the same test the other three writers make. A source-shape check pins it.

Considered and not done: funnelling all four writers through one `writeGammaBaseline` helper with a call-count test (what #1231 did). Three sites already agree, one had drifted, and the sites are all in this file. A helper is a refactor of working code for one missing line. Dropping a curve whose monitor changed was also not done, because the "kept across the gap" comment on `gammaBaselines` exists to prevent that.

## Verification

Mac17,3, macOS 26.6.2 (25G83). On `main` at `f7378a8d`.

`./build.sh` exit 0, no warnings. `./build/Vorssaint --selftest` OK. `./build.sh --test` `TESTS OK (9609 checks)`.

The new assertion was run against `main` first: `main` has no `baseline.fingerprint == Self.displayFingerprint(display.id)`, so it is red there and green here.

**Not tested.** No second monitor here, so the gamma path was not exercised against real hardware: a software-dimmed display, switched off and back on with a different monitor taking its display number, is the case the fix is about and it has not been reproduced. `BrightnessService.swift` is not in the `--test` file list, so the full `./build.sh` is the only thing compiling this change.

## Anything else

#301 reports an external display blacking out, sometimes while the brightness slider is moving, recoverable only by unplugging the hub. This is one way that can happen: a display number handed out again after a reconnection let the pre-switch-off restore write a dimmed baseline to whichever monitor now holds the number. #836 attributes #301 to DDC renegotiation, which this branch does not touch, so this is an additional candidate path and not the fix that issue is waiting for.

Refs #301

Supersedes #1231.

### DIFF
--- a/Sources/Vorssaint/Services/Display/BrightnessService.swift
+++ b/Sources/Vorssaint/Services/Display/BrightnessService.swift
@@ -433,8 +433,10 @@ final class BrightnessService: ObservableObject {
                     return
                 }
                 // A software-dimmed screen should return with its clean gamma
-                // before the saved dim level is reapplied by the rebuild.
-                if let baseline = self.gammaBaselines[display.id] {
+                // before the saved dim level is reapplied by the rebuild. The
+                // number may already belong to another monitor (see `GammaTable`).
+                if let baseline = self.gammaBaselines[display.id],
+                   baseline.fingerprint == Self.displayFingerprint(display.id) {
                     CGSetDisplayTransferByTable(display.id, baseline.count, baseline.red,
                                                 baseline.green, baseline.blue)
                 }
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -14143,6 +14143,10 @@ struct MetricsTests {
             .joined(separator: "\n")
         expect(!brightnessWorkQueueHalf.isEmpty && !brightnessWorkQueueCode.contains("NSScreen"),
                "the brightness work queue resolves display names without touching NSScreen")
+        // Display numbers are reissued after a reconnection, so the gamma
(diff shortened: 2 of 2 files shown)


PathGao:
The count in the body holds up. I grepped `main` for `CGSetDisplayTransferByTable`: exactly four writes, all in `BrightnessService.swift` — `restoreAllGamma` (guarded by its `where`), two in `applySoftwareDim` (guarded at the top), and this one. Nothing outside that file writes a curve, so the fix covers the shape.

Agreed on skipping the helper. One missing condition doesn't buy a refactor of three sites that already agree.

Also checked, since either one would make the new assertion worthless: `brightnessSource` is declared above the insertion point, and `BrightnessService.swift` really is absent from build.sh's `--test` file list, so your "the full build is the only thing compiling this" note is right.

Nothing to flag. Not verified: no second monitor here either, so the mismatch path is unexercised; and the two build checks hadn't reported when I read this.

**Impact:** none when the fingerprint agrees — same bytes written. Behaviour changes only after a display number has been reissued, which is the bug.
**Merge:** ship it.

ref #301

PR #1300, opened 2026-09-02
fix(build): only count signing identities that can actually sign

PathGao:
## Summary

`build.sh` and `Tools/setup-signing.sh` decided whether the stable
self-signed identity `Vorssaint Utils Signing` was usable by reading a
`security find-identity` listing. Neither spelling answers that question.
Without `-v` the listing includes certificates that cannot sign, so a stale
one is handed to `codesign`, which fails with `errSecInternalComponent` after
a clean release compile, while the ad-hoc fallback sits unreached in the
`else` arm; on `--dev` the same misreading skips `Tools/setup-signing.sh`,
whose own front door then prints `✓ Signing identity already installed.` over
the dead certificate. With `-v` the listing reports only trusted certificates,
which a self-signed one never is, so every local build would go ad-hoc and
`setup-signing.sh` would delete and reissue a working identity on every run.

Both scripts now ask `codesign` itself: sign a throwaway copy of `/bin/echo`
with the identity and believe the result. `build.sh` gets
`legacy_identity_installed`, replacing all six `find-identity | grep`
guards (the `NOW_PLAYING_ADAPTER_ID` one added since #1232 included);
`Tools/setup-signing.sh` gets `identity_can_sign` at the front door and as
the post-import read-back. The probe passes `--strip-disallowed-xattrs` like
every other signing call, since a copy into `$TMPDIR` carries
`com.apple.provenance`. The Developer ID lookup keeps `-v`: that certificate
is Apple-issued and trust is the right question there.

Both probes unlock `vorssaint-signing.keychain-db` first. A locked keychain still lists its identities but cannot sign with them, and this one is locked after every reboot; without the unlock the probe answers "no" and `--dev` hands off to `setup-signing.sh`, whose next step deletes the keychain and reissues the identity, orphaning every Accessibility grant. Reproduced with `security lock-keychain`: the probe fails before the unlock line and passes after it, and `setup-signing.sh` on a locked keychain now reports the identity as already installed instead of replacing it. The keychain password literal is therefore in `build.sh` as well as `setup-signing.sh`; that is the cost of not sharing state between the two scripts.

Considered and not done, relative to #1232: no result cache for the probe (four extra codesign calls on a copy of `/bin/echo` per build are milliseconds, and the cache carried a staleness bug), no cleanup sweep for the probe file (a plain `rm -f` in the function covers it), and none of the source-shape tests that pinned those parts.

## Verification

Mac17,3, macOS 26.6.2 (25G83). Release build without a Developer ID; the
stable self-signed identity is installed on this machine.

`./build.sh` exit 0, 0 warnings, log says `signing with legacy self-signed
identity: Vorssaint Utils Signing`, so the probe took the real path here;
`./build/Vorssaint --selftest` OK; `./build.sh --test` 9612 checks, 0 failed.

Not tested: a machine with a valid Developer ID identity (that branch is
untouched), a keychain holding an expired stable certificate (the failure
this fixes, seen on the machine behind #1232), and
`./build.sh --dev` / `Tools/setup-signing.sh` from an identity-less state.

## Anything else

Refs #1123
Supersedes #1232.

### DIFF
--- a/Tests/MetricsTests.swift
+++ b/Tests/MetricsTests.swift
@@ -21921,6 +21921,21 @@ struct MetricsTests {
         expect(!signingSetupCode.contains("-legacy"),
                "setup-signing.sh avoids the -legacy flag the stock LibreSSL openssl rejects")
 
+        // MARK: The stable identity is judged by whether codesign can sign with it
+        // A find-identity listing names certificates codesign then rejects, and
+        // -v excludes every self-signed one, so neither spelling may decide.
+        let buildScriptCode = buildScript.components(separatedBy: "\n")
+            .filter { !$0.trimmingCharacters(in: .whitespaces).hasPrefix("#") }
+        for (script, code, identity) in [("build.sh", buildScriptCode, "$LEGACY_IDENTITY"),
+                                         ("Tools/setup-signing.sh", signingSetupCode.components(separatedBy: "\n"),
+                                          "$IDENTITY")] {
+            expect(!code.contains { $0.contains("find-identity") && $0.contains(identity) },
+                   "\(script) never decides the stable identity by a find-identity listing")
+            expect(code.contains { $0.contains("cp /bin/echo") }
+                    && code.contains { $0.contains("--sign \"\(identity)\" \"$probe\"") },
+                   "\(script) asks codesign to sign a throwaway copy of /bin/echo with the stable identity")
+        }
+
         // MARK: Modifying mouse taps are handed back across a session switch
         // The tap owners cannot be reached from this list (they need the event
         // chain), so the wiring is pinned as text: each service follows the
--- a/Tools/setup-signing.sh
+++ b/Tools/setup-signing.sh
@@ -29,6 +29,9 @@ KCPASS="vorssaint-signing"
 # excludes every self-signed one; ask codesign itself with a throwaway copy.
 identity_can_sign() {
     local probe signed=1
+    # Locked after every reboot; a locked keychain cannot sign, and a false
+    # answer here would delete it and reissue the identity.
+    security unlock-keychain -p "$KCPASS" "$KC" 2>/dev/null || true
     probe="$(mktemp)"
     cp /bin/echo "$probe"
     codesign --force --strip-disallowed-xattrs --sign "$IDENTITY" "$probe" >/dev/null 2>&1 && signed=0
--- a/Tools/setup-signing.sh
+++ b/Tools/setup-signing.sh
@@ -13,7 +13,7 @@
 # requirement and drop every user's granted permissions. The name lives only in
 # the keychain and codesign output, never in anything the app shows.
 #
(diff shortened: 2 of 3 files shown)


PathGao:
Six guards in `build.sh` and both sites in `setup-signing.sh` — I grepped `main`, that's all of them. The new test pins "nothing decides this by a `find-identity` line", which closes the hole I flagged on #1232.

One regression against `main`. A locked keychain still *lists* its identities but cannot sign with them, and `vorssaint-signing.keychain-db` is locked after every reboot — that's the #1232 password-dialog report. So `identity_can_sign` returns false, `--dev` then runs `setup-signing.sh`, and its second act is:

```sh
security delete-keychain "$KC" 2>/dev/null || true
```

A locked keychain, a cancelled dialog, or a non-interactive shell now destroys the working identity and issues a new certificate, orphaning the Accessibility grants this identity exists to preserve — the same outcome the body rejects `-v` for. `find-identity` never did this.

`KCPASS` is already in that file, so one line before the front door covers it:

```sh
security unlock-keychain -p "$KCPASS" "$KC" 2>/dev/null || true
```

I could not run the probe against a locked keychain — the mechanism is read off the script, not tested.

**Impact:** `./build.sh --dev` after a reboot reissues the stable identity and orphans granted permissions.
**Merge:** needs the unlock line in `setup-signing.sh` first.

ref #1123

PathGao:
Taken, in both probes rather than only `setup-signing.sh`: a plain `./build.sh` after a reboot hit the same false answer and fell through to ad-hoc signing, which orphans the grants the same way. Reproduced here with `security lock-keychain`: probe fails before, passes after, and `setup-signing.sh` on the locked keychain now says already installed instead of reissuing. `--test` 9612 checks.

PathGao:
@PathGao the unlock line is in both probes now, body rewritten to match; please re-review.

PathGao:
Unlock is in both probes, and putting it inside the function rather than at the call sites is the right place — the six guards in `build.sh` can't drift out of sync with it. `-p "$KCPASS"` also means a wrong password fails quietly instead of raising the #1232 dialog.

I re-grepped `main`: six `find-identity | grep "$LEGACY_IDENTITY"` in `build.sh` (80, 145, 546, 558, 571, 587) and both in `setup-signing.sh`, all eight replaced; the Developer ID lookup at `build.sh:67` keeps `-v`. No stragglers.

Nothing left to flag. The one thing I would have asked about is the failing branch under `set -euo pipefail` — `codesign … && signed=0` leaves the function's last AND-OR list non-zero — and your reboot run answers it: `build.sh` fell through to ad-hoc rather than aborting. Still unverified by me: an identity-less machine and a keychain holding an expired certificate, both already listed as untested. Build & selftest had not reported when I read this.

**Impact:** none beyond the fix — restores `main`'s behavior on a locked keychain, and answers correctly on a dead certificate where `main` did not.
**Merge:** ship it.

ref #1123

PR #1302, opened 2026-09-02
fix(quit-protection): follow the keyboard layout

PathGao:
## Summary

The Command-Q / Command-W protection from #1200 never engaged on a non-Latin keyboard layout. Matching compared `charactersIgnoringModifiers` with "q" and "w", which is the character the key types *without* Command: a Russian layout reports "й" and a Greek one ";" for the very key that quits, nothing matched, and Command-Q went through with protection switched on. The positional `fallbackKeyCode` never covered it either, because the character is non-nil for those presses.

The layout's Command table is what macOS resolves a key equivalent through, and `GlobalShortcut` has cached that table for every key code since #1226 (with the ASCII-capable fallback under an input method from #1220). Matching now asks that cache for the key's Command label and compares it with "q"/"w" first, falling back to the typed character and then to the US position only when no label is cached. `layoutKeyLabel(for:usesCommand:)` loses its `private` for that, nothing else in `GlobalShortcut` changes. On Russian and Greek key code 12 answers "Q"; on Dvorak it answers "'" and key code 7 answers "Q", so Command-' there is left alone; DVORAK-QWERTYCMD and French AZERTY follow their own tables.

Two consequences of that are fixed in the same change. The unconditional `isRepeat` drop in `handleKeyDown` sat above every modifier read, so with protection on a held Q or W produced one character and no repeat, and the layout match would have carried that onto й/ц and ;/ς. It is gated on Command now (`QuitProtectionSupport.dropsAutorepeat`); nothing below it acts without Command, so a held Command-Q still cannot restart a hold or count as the second of a double press. And when the tap is re-enabled after being disabled on timeout, the key up that ends the post-confirmation swallow may be among the events missed, so the swallow is cleared there rather than holding until the next matching key up.

Considered and not done, against #1218: the two-second ceiling on the swallow (no observed case of a lost key up, and the tap re-enable path above is the one with a concrete mechanism); a reverse label-to-key-code map in `GlobalShortcut` (one caller, and the forward cache already answers the same question with no tie-break needed); calling `startObservingKeyboardLayout()` from the service (the app already does at launch); the `handleFlagsChanged` rewrite (same behaviour as `main`).

Cost of the forward lookup: the tap runs on the main run loop, so a cache miss derives the label from the layout on that thread. Measured here, that derivation is about 36 microseconds per call, and it only happens on keys with no printable Command label (Space, Return, arrows), which the cache does not retain. Cache hits are a lock and a dictionary read.

## Verification

MacBook Air (Mac17,3), macOS 26.6.2 (25G83), self-built from this branch.

`./build.sh` exit 0 with zero warnings, `./build/Vorssaint --selftest` SELFTEST OK, `./build.sh --test` TESTS OK (9619 checks; the 11 assertions this branch adds are the whole difference).

The six layout assertions read Command-Q and Command-W off the layouts installed on this machine (US, Russian, Greek, Dvorak, DVORAK-QWERTYCMD, French) through the same `matchesKey` the tap calls, feeding it the bare character and the Command label the cache holds for that key. Confirmed red first: with the `commandLabel` comparison removed from `matchesKey`, five assertions fail (the Cyrillic press, Russian, Greek, DVORAK-QWERTYCMD, and the bare-key autorepeat one) and nothing else moves. Dvorak alone stays green there because its bare and Command tables agree, so the character fallback already answers it. The three `dropsAutorepeat` assertions were confirmed the same way: with the gate removed, `holding the Q or W key without Command still repeats with protection on` fails.

Not tested: I have no machine set to a Cyrillic or Greek layout, so I have not pressed Command-Q there with the feature on; the layout claim rests on the layout data and the assertions above. I could not stage a tap disabled by timeout, so the swallow clear on re-enable is covered by reading the state machine, not by a live reproduction. The autorepeat gate is asserted as a predicate, not by holding a key with the app running.

## Anything else

Nothing on the tracker reports a layout on which the protection never engages. #892 and its duplicate #1158 asked for the feature itself and #1200 delivered it without referencing either; on a Cyrillic or Greek layout what shipped never engaged, so for those reporters the protection arrives with this branch.

Refs #892
Refs #1158

Supersedes #1218.


PathGao:
Dropping the character fallback whenever a Command label exists is what makes DVORAK-QWERTYCMD come out right: keycode 7 types `q` bare, so the old matcher protected the key macOS does not quit from. Confirming red first on five of six is what makes the layout claim believable without a Cyrillic machine.

Read the diff against the tree, nothing to flag. What I checked:

- `matchesKey` / `isBaseShortcut` / `isExtraShortcut` have no caller outside `QuitProtectionService`, and all four sites are updated.
- The tap source is added to `CFRunLoopGetMain()`, so the main-thread derive in `layoutKeyLabel` is reachable and a cache miss cannot answer nil here.
- `stop()` already clears `swallowShortcut`, so the tap-disabled branch was the only ungated sibling.
- `postSyntheticKeyUp` replays the original keycode, so confirmation lands on the same physical key the layout matched.

Not verified: no Cyrillic or Greek layout on this end either, and I cannot reach the live tap, so the layout and re-enable claims rest on the same reading you did. The build and Swift 6 checks had not reported when I read this.

**Impact:** bare q and w regain autorepeat with protection on; non-Latin layouts get protection that never engaged.
**Merge:** ship it.

ref #892
ref #1158

PR #1303, opened 2026-09-02

PathGao:
## Summary

Selecting an app in the switcher often left it behind the app you came from: its windows rose but the menu bar stayed put. `NSApp.yieldActivation(to:)` only hands over activation this app currently holds, and Vorssaint is `LSUIElement` with non-activating panels, so at the moment a handoff commits it usually holds none. The yield gave away nothing and the cooperative `activate(from:)` after it was refused.

Fix: self-activate before yielding. `ActivationHandoff.yield(to:)` does `NSApp.activate(ignoringOtherApps: true)` then `NSApp.yieldActivation(to:)`, and the five yield sites on `main` (`WindowActivator` x2, `SpaceHop`, `DockClickService`, `ProcessUsageService`) call it instead of yielding bare. Adopts #845 by @pboucher, which covered two of the five.

The self-activation posts an activation notification for Vorssaint's own process, which `WindowUseTracker.appActivated` would rank first in the app history and file a Vorssaint window into the window history. That is guarded on our own pid: own activations are not promoted and the observer retarget skips the focused-window read, while the observer still retargets so a yield that does not take leaves it on the app that is actually front.

Considered and not done: #1278 replaced the pid guard with an arm/consume/timeout mechanism in the tracker so that Dock icon and Settings activations of Vorssaint would still be recorded. That corner case (switching onto Vorssaint's own window, then back, ranking one slot off) has no observed report, and the mechanism took three review rounds of edge conditions. This keeps the one guard #845 had. `RadialNowPlayingService.open` and `ClipboardHistoryService.pasteIntoPreviousApp` activate without yielding and are left alone; they are a separate change.

`retargetObserver(filingFocusedWindow:)` skips the focus read on both of its exits: the early return when the observer already watches that pid, and the tail of the install path. At handoff time the observer watches the app the user came from, so the self-activation takes the install path, and an unguarded tail read there would file a Vorssaint window anyway.

## Verification

Mac17,3, macOS 26.6.2 (25G83), Apple Silicon, own build.

`./build.sh` no warnings, `./build/Vorssaint --selftest` OK, `./build.sh --test` TESTS OK (9610 checks).

Two source-shape assertions next to the existing Dock click yield check: the helper self-activates before it yields, and no `.swift` under `Sources` outside `ActivationHandoff.swift` mentions `yieldActivation` (the scan reports its file count so an empty walk cannot pass). Both are red on `main`: the helper does not exist there and five files yield bare.

Not tested: whether macOS grants the self-activation when the handoff is triggered by a mouse click (Dock click, process list) rather than a hotkey release; that needs a real desktop and a genuine gesture. Trade-off unchanged from #845: if the self-activation is granted but the yield still does not take, the user lands on Vorssaint instead of staying on the app they came from. Vorssaint's own activations (Dock icon, Settings, its own windows in the switcher) are no longer recorded in the app history.

## Anything else

Refs #961
Refs #868

Supersedes #1278.


PathGao:
Five yield sites on `main`, five converted, and the source-shape assertion pins any new one to the helper. Root fix, not a per-call-site patch.

One gap: `retargetObserver` applies `filingFocusedWindow` to the early-return branch only. The tail `readFocusedWindow(of: pid)` at the end of the install path is still unconditional. At handoff time `observedPID` is the app the user came from — Vorssaint holds no activation, that is this PR's premise — so `pid != observedPID` is true, it falls through to the install path, and files our own focused window anyway. The guarded branch is the one the self-activation almost never takes.

```swift
// end of retargetObserver, install path
if filingFocusedWindow { readFocusedWindow(of: pid) }
```

Same path also installs an AXObserver on our own pid on every switch and tears it down when the target activates — two extra AX round-trips per switch on the watcher thread.

Read from `WindowUseTracker` on `main`, not verified at runtime.

**Impact:** a Vorssaint window lands in the window history on every switch, which is what the pid guard was added to prevent.
**Merge:** needs the tail `readFocusedWindow` behind the same flag first.

ref #961

PathGao:
@PathGao the tail `readFocusedWindow` in the install path is behind the same flag now (one line). Full build 0 warnings, selftest OK, `--test` 9610. Please re-review.

PathGao:
Tail read is behind the flag now, and that was the branch the self-activation actually takes. Five yield sites on `main`, five converted; the source-shape assertion pins new ones to the helper.

Correction to my last comment: I overstated the impact. An own window ID in the history is inert. `WindowEnumerator` drops own-pid windows before it orders anything, and `WindowLayoutService.focusedTarget` skips own bundle ID, so no entry the history ranks can be a Vorssaint one — and the ranking is relative, so an own pid sitting at the front shifts nothing.

Same reason the one remaining unguarded path is fine to leave alone: `focusCallback` promotes whatever the observer reports, so while the observer sits on our own pid a Vorssaint focus change still files a window. Inert for the same reason. Not worth another guard.

Read from `main`, not verified at runtime; the two build checks had not reported when I looked.

**Impact:** none beyond the trade-off the body already names — a granted self-activation with a refused yield leaves the user on Vorssaint.
**Merge:** ship it.

ref #961

PR #1304, opened 2026-09-02
fix: hand the remaining event taps back across a user switch

PathGao:
## Summary

the screen, and a filter tap created there keeps its place in the event chain.
The window server still routes every matching event through a process that
cannot answer and waits out the tap timeout on each one, so the account
actually in use pays it (#1153, #1075). #1145 wired nine tap owners to
`SessionActivity` for this. Twelve more owners still ran their taps without
looking at the session:

- `WindowLayout/WindowLayoutService.swift` (gesture, edge snap, directional; the layout hotkeys share the same `trusted` gate, so they are unregistered on resign and re-registered on activate)
- `WindowMaximizer.swift`
- `Finder/FinderCutPaste.swift`
- `Finder/FinderRenameService.swift`
- `KeyboardDebounce/KeyboardDebounceService.swift`
- `SuperKey/SuperKeyService.swift` (keyboard and mouse taps)
- `ShortcutRecordingTap.swift`
- `Switcher/AppSwitcher.swift`
- `Snippets/TextSnippetService.swift`
- `Audio/PreciseVolumeRollerService.swift`
- `DockClick/DockClickService.swift`
- `Display/BrightnessService.swift` (brightness key tap and function-key tap)

Each now follows the shape `SmoothScrollService` and
`MouseClickDebounceService` already use on `main`: `init` subscribes to
`SessionActivity.shared.onChange`, the preference sync asks
`SessionActivitySupport.tapShouldRun` with `AXIsProcessTrusted()` instead of
the polled `Permissions.shared.accessibility`, and the timeout re-arm re-enables
the tap only while the session is active and Accessibility is granted, otherwise
it hands the decision to the sync on the main queue. Taps that run on their own
thread read the two flags in place, the way `MouseClickDebounceService.handle`
does.

Three things the earlier review rounds found are kept:

- `ShortcutRecordingTap` ends the recording, not only the tap, when the session
  resigns: `ShortcutCapture.begin()` had switched every global shortcut off, so
  dropping the tap alone came back with all of them dead. `ShortcutCapture.end()`
  is idempotent, and the observer registers before the Accessibility check
  because the capture is already on by then.
- `KeyboardDebounceService.stop()` invalidates the port as well as disabling
  it, matching its twin `MouseClickDebounceService`.
- `AppSwitcher.reconcileTakeover()` (#882) rebuilt the tap from its own
  predicate with no session term; it now shares the sync's gate.

Not done, on purpose: the `restartTapOnMain` / `rearmDisabledTap` helpers and
the rewrite of the tap-thread restart tails (a stop and a start crossing at the
handover was read from the lifecycle code, never observed); the directional tap's timeout handling and the Escape deferral in
`WindowLayoutService` (nothing was seen to fail there); the
`publishRunning(false)` move in the debounce (its generation is always stale by
then, so it never landed); the restart-tail and gate-argument shape tests. The
radial menu's side-button tap is #1270 and is not in here. The gamma fingerprint
half of #1231 is a separate change.

Super Key does more than hand its tap back. Its `onChange` handler and its re-arm else branch both go through `syncWithPreferences()`, which on resign reaches `stop()` and `clearLeftoverMapping()`: three `hidutil` calls and two defaults writes per switch, on the serial `mappingQueue`. That is the intended behaviour, not a side effect: the Caps Lock mapping is system-wide, so leaving it in place would hand the other account a dead Caps Lock. On activate the sync re-applies it.

## Verification

Mac17,3, macOS 26.6.2 (25G83), branched from `main` at `f7378a8d`.

```
./build.sh                     exit 0, no warnings
./build/Vorssaint --selftest   SELFTEST OK
./build.sh --test              TESTS OK (9656 checks)
```

The twelve service files are outside the `--test` compile whitelist; the full
`./build.sh` is what compiled them. The added assertions are the twelve roster
entries in the existing session-switch sweep, each of which is red on `main`
(none of the twelve files contains `SessionActivity.shared.onChange` there).

**Not tested.** There is one login session on this machine, so the fast user
switching behaviour is reasoned from the mechanism #1153 describes, not
measured, the same caveat the issue carries. None of the twelve features was
exercised interactively after the change; the tap lifecycle is what moved, so
the ordinary paths should be unaffected, but that is an argument, not a run. The
revoke race (a tap disabled inside the polling interval after a revoked grant)
is not reproduced either.

## Anything else

Refs #1153
Refs #1075

Supersedes #1233.
Supersedes #1234.
Supersedes #1235.
Supersedes #1236.
Supersedes the tap half of #1231.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 - The radial menu now checks only which extra mouse button opens a wheel when a side button is pressed, instead of loading every wheel and its icons on each event, so holding a side button and moving the mouse no longer risks dropped clicks. Thanks to @PathGao.
 
 ### Fixed
+- Features that handle keyboard, mouse and media keys now step aside during fast user switching, preventing input stalls in the active account. Thanks to @PathGao.
 - Restoring screen gamma before switching off a display now verifies the display's identity, preventing a dimmed curve from applying to another monitor after reconnections. Thanks to @PathGao.
 - The radial menu now releases its mouse button tap when switching macOS user accounts and closes any open wheel, preventing extra button clicks from stalling in the other account. Thanks to @PathGao.
 - Paste as plain text no longer freezes the app when the app you copied from is slow to hand over the copied content. Thanks to @PathGao.
--- a/Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift
+++ b/Sources/Vorssaint/Services/Audio/PreciseVolumeRollerService.swift
@@ -15,12 +15,20 @@ final class PreciseVolumeRollerService: ObservableObject {
     private var source: CFRunLoopSource?
     private var gate = PreciseVolumeRollerGate()
 
-    private init() {}
+    private init() {
+        SessionActivity.shared.onChange { [weak self] _ in self?.syncWithPreferences() }
+    }
 
     func syncWithPreferences() {
         let wanted = AppFeature.mixer.isAvailable
             && UserDefaults.standard.bool(forKey: DefaultsKey.preciseVolumeRollerEnabled)
-        wanted ? start() : stop()
+        if SessionActivitySupport.tapShouldRun(featureWanted: wanted,
+                                               accessibilityGranted: AXIsProcessTrusted(),
+                                               sessionIsActive: SessionActivity.shared.isActive) {
+            start()
+        } else {
+            stop()
+        }
     }
 
     func suspend() {
@@ -85,8 +93,10 @@ final class PreciseVolumeRollerService: ObservableObject {
 
     private func handle(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
         if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
-            if AXIsProcessTrusted(), let tap {
+            if SessionActivity.shared.isActive, AXIsProcessTrusted(), let tap {
                 CGEvent.tapEnable(tap: tap, enable: true)
+            } else {
+                DispatchQueue.main.async { [weak self] in self?.syncWithPreferences() }
             }
             return Unmanaged.passUnretained(event)
         }
--- a/Sources/Vorssaint/Services/Display/BrightnessService.swift
+++ b/Sources/Vorssaint/Services/Display/BrightnessService.swift
@@ -217,7 +217,9 @@ final class BrightnessService: ObservableObject {
     /// panel must not put the same slow monitor probe behind itself again.
     private var rebuildingTopology: BrightnessSupport.DisplayTopology?
 
-    private init() {}
+    private init() {
+        SessionActivity.shared.onChange { [weak self] _ in self?.syncKeyTap() }
+    }
 
     func setKeyboardLightEnabled(_ enabled: Bool) {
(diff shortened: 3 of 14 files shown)


PathGao:
Roster checks out: 21 `.defaultTap` owners on `main` — 8 from #1145, 12 here, radial menu is #1270. The three untouched `tapCreate` files (DockPreview, AutoQuit, MusicLaunchBlocker) are `.listenOnly`, outside #1153's mechanism. Nothing missing.

**SuperKey does more than hand its tap back.** The "not done" list names "routing the Super Key's restart through the full sync pulled and re-applied its hidutil mapping" as the reason to skip the shared helper — but the new re-arm else branch *is* `syncWithPreferences()`, and so is the `onChange` handler. On resign that reaches `stop()` → `clearLeftoverMapping()`: three `hidutil` shell-outs and two defaults writes per switch. Not a race, `mappingQueue` is serial. Probably the right behaviour — the mapping is system-wide, leaving it gives the other user a dead Caps Lock — but the body says the opposite.

**`WindowLayoutService`:** folding the session into `trusted` also gates `registerHotkeys()`. The parenthetical is "(gesture, edge snap, directional)"; the main layout shortcuts ride along undeclared.

No second login session here either — I read this, didn't run it.

**Impact:** none single-session. Across a switch, Super Key's hidutil mapping is now cleared and re-applied each time.
**Merge:** needs the body to cover SuperKey's `stop()` and the WindowLayout hotkeys first.

ref #1153
ref #1075

PathGao:
@PathGao body now covers both: Super Key's resign goes through `stop()` and `clearLeftoverMapping()` on purpose (the mapping is system-wide), and the WindowLayout hotkeys ride the same `trusted` gate. The wrong hidutil sentence in the not-done list is gone. No code change. Please re-review.

PathGao:
Both gaps closed. SuperKey's `stop()` → `clearLeftoverMapping()` now reads as intended behaviour, the WindowLayout hotkeys are named in the bullet, and the wrong hidutil sentence is out of the not-done list. Re-read the diff: no code change since last round.

One place the body still over-claims, harmless: SuperKey's sync does not call `tapShouldRun`. It ANDs `SessionActivity.shared.isActive` onto `enabled` and leaves the Accessibility term to `start()`'s own `AXIsProcessTrusted()` guard, which also clears the leftover mapping — so it has to stay there. Eleven of twelve match the shape the body describes; same result either way, not worth a commit.

`WindowLayoutService`'s `trusted` has no other reader on `main` — the four input gates are all of it, so folding the session in does not make anything user-facing claim Accessibility is missing.

Still one login session here, so the switch behaviour is read from the code, not run.

**Impact:** none beyond what the last round already covered.
**Merge:** ship it.

ref #1153
ref #1075

PathGao:
Merged. Dropping `ready to merge` — the merge was its release condition.

Roster is closed on `main`: all 21 `.defaultTap` owners now carry `SessionActivity.shared.onChange` (8 from #1145, 12 here, the radial menu from #1270). Grepped `main` after the merge, not read off the diff.

None of it was measured on a real switch — still one login session on my side — so #1153 shouldn't be treated as verified until someone with two accounts confirms the stall is gone. The deferred items stay open work: the `restartTapOnMain` / `rearmDisabledTap` helpers, the directional tap's timeout handling in `WindowLayoutService`, and the gamma fingerprint half of #1231.

ref #1153
ref #1075

PR #1305, opened 2026-09-02
perf(recorder): walk the click track once per plan instead of once per frame

PathGao:
## Summary

`RecorderComposer.makePlan` solves the whole pointer and zoom timeline before the first frame is drawn, and four of its per-frame lookups walked their entire list every time: `RecorderMotion.pressScale`, `ringProgress`, `anchorWeight` (through `anchored`) and `focus`. Ten minutes at 60 fps is thirty six thousand frames, and a recording of real use carries a four figure click list, so the export spent tens of millions of iterations, most of them on clicks that had faded out long before or had not happened yet.

The frame loop asks in time order and both lists are in time order, so each lookup now takes an `inout Int` cursor that the loop carries across frames: one pass over the list per plan instead of one per frame. The same file already does this in `resampled`, `resampledShapes` and `resampledVisibility`. The press punch and the click anchor step their cursor only past a release, since a held press still weighs on every later moment, and they read the trailing "ended mid-drag" press off `clicks.last` because the cursor cannot reproduce the scan's own state from an arbitrary start. The old signatures stay and call the cursored one with a fresh cursor, so no test call site moves.

Measured on the extracted functions with `-O`, ten minutes of recording, 2914 clicks, 36000 frames: the press punch and the click ring together drop from 0.307 s to 0.001 s with an identical checksum over every frame. The anchor and focus passes were not profiled; they are the same frames-by-clicks shape.

Considered and not done. A `ClickCursor` struct bundling the punch and ring cursors: two `Int`s in the one loop that carries them do the same job. Long doc comments and test assertions about out-of-order click lists: nothing on `main` builds one (the sampler appends in event order, `decode` round-trips it, `focusClusters` emits ascending), so each cursored function carries a one-line precondition instead. A point-for-point comparison of `anchored`'s blended output: `anchored` is a six-line map over `anchorWeight`, and `anchorWeight` itself is compared frame by frame.

## Verification

Mac17,3, macOS 26.6.2 (25G83), my own build.

`./build.sh` clean with no warnings; `./build/Vorssaint --selftest` passes; `./build.sh --test` TESTS OK (9610 checks).

The new test keeps a copy of the four scans that were replaced and compares them frame by frame with `!=` on the raw `Double`, no tolerance, over 5972 frames against 558 generated clicks with bursts, held buttons, presses never released and one at the very end. Each query is compared both with the cursor carried across the loop and with a fresh one. On `main` the test does not compile, since the cursored signatures do not exist there.

Not tested: no video was exported. The claim is that the output is unchanged, so there is nothing new to look at in a finished recording, and the timing figure comes from the extracted functions rather than a real export.

## Anything else

No open issue reports a slow export. The recorder issues that exist (1147, 843) are about what is captured, not what the export costs.

Supersedes #1243.


PathGao:
Keeping a copy of the four replaced scans and comparing with `!=` on the raw `Double`, cursor carried and fresh, is the right way to land this. The unmatched presses — every 7th, plus one at the end — are the case that would have bitten, and they're in there. I traced `settled` in both anchor scans: it only moves past a release older than the window, and everything behind it is out of range for both the drag rule and the distance term. Reads correct.

**0.307 s of what?** An export compositing 36000 frames runs for minutes, so this is likely under 1% of it. Put the fraction in the body — the number reads much bigger without a denominator. Same point in reverse: the fifth per-frame walk is still standing three lines up (`sourceTime` rebuilds `keptRanges`, a sort plus two allocations, every frame), and it is too small to be worth chasing.

**`focus` is barely covered.** Its fixture is 2 clusters spanning 0.03 s, so the cursor advances once and the other ~6000 frames compare a constant. The other three get the generated list.

**Impact:** none. Output unchanged, verified frame by frame.
**Merge:** ship it. I read the diff, not a real export, and the two build checks had not reported when I read this.

PR #1308, opened 2026-09-02
feat(charge-limit): add charge limits and accurate battery status

BenjaminD2023:
## Summary

- Add opt-in Charge Limit, Top Up, manual Discharge, and calibration controls in the menu panel and Energy settings. Earlier supported hardware retains 20–100% limits and Sailing mode; macOS 27 uses a runtime-checked native PowerUI backend with 80–100% limits in 5% steps.
- Top Up bypasses the saved cap to charge to 100%. Manual Discharge uses battery power until stopped, independent of the cap. It remains selectable when unplugged and reapplies on reconnect; it does not shut down USB ports.
- Refresh battery icons from live controller state and power-source notifications, distinguish charging from holding, and improve charging-bolt contrast.
- Keep helper-controlled adapter input fail-safe with readback, deduplicated writes, heartbeat/watchdog recovery, stale-helper registration recovery, and bounded XPC requests. Native charge limits remain enforced by macOS when the app exits. Uninstall retains the existing restore-confirmation check.

The macOS 27 fallback dynamically checks private PowerUI selectors rather than linking the private framework. It adds no third-party dependency or network service. macOS manages the native charging range and may drain to its native limit above that limit.

## Verification

- User confirmed the signed local update works on macOS 27, MacBook Air M4 (Mac16,12).
- Live Top Up test confirmed positive battery current; Discharge at 90% with a 90% cap switched to Battery Power and stayed active until Stop. Stop restored adapter input and the saved 90% limit.
- All 16 USB registry entries retained their identities during Discharge, including the soundbar, keyboard, receiver, and Ethernet adapter. Outbound USB-C wattage was not measured; this is not a guarantee of every dock/device power-role combination.
- 1,887 focused policy checks passed, including connected/unplugged behavior across battery levels; hardware mock tests verify deduplication and restoration.
- The branch is merged with current main, with conflicts resolved. The signed developer build and both helper self-tests pass.
- The full local regression rerun passed all 10,878 checks, plus charge-hardware and preference-cleanup tests. GitHub CI reruns on the final commit.

Earlier macOS 26 testing remains applicable to the legacy backend. Intel and other hardware/firmware combinations still need live confirmation.

## Anything else

Supersedes #851 after the requested rework.

Refs #846
Refs #849


BenjaminD2023:
By now, this PR should be ready to review, it should at least work on MacOS 26, I will be updating my mac tonight to MacOS 27 Public Beta and if anything went wrong I will close this PR in respond to that. I would be grateful if anyone has MacOS 27 installed and can help me with the testing on the new MacOS system.

BenjaminD2023:
Wait, this PR is not ready for review yet, I've changed the PR status to draft, it stopped working on my device after a restart of the app. I need to do further testing with MacOS 27

BenjaminD2023:
Updated with the macOS 27 charge-control, Top Up, and manual Discharge fixes; confirmed working by the user. USB devices stayed connected in the live test. Latest main merged and conflicts resolved. Signed build, helper self-tests, 10,878 regression checks, and hardware/cleanup tests pass. Ready for review; GitHub CI is rerunning on the final commit.

vorssaint:
closing this one. it does not fit the app as it is, and the parts that could fit would need a rewrite
macos 26.4 and later already offers a charge limit in system settings, so on current systems this mostly drives that same setting through a private interface that apple can close at any time, and turning the limit off or removing the feature also overwrites the choice the user made in system settings with 100%
it ships a second privileged helper that copies the fan control one instead of extending it, which means a second login items prompt and twice the code to maintain
every mac with a battery pays for it even with the feature uninstalled, since the monitor now reads the charging controller four times a second forever and resamples on every battery driver update, and once installed the app polls every two seconds
the charge slider and the allow button show up in the battery detail for everyone and install the feature on their own, the battery preset now installs a privileged helper, and the limit is on by default
the manual discharge has no floor, so a plugged in mac left in that mode drains until it shuts down
it also rewrites unrelated parts of the app, the per reading values or bars settings, a hand drawn menu bar battery icon, the temperature combining rule and the signing flow
two long captions are english only, the sailing wording is jargon, the custom slider and segmented bar are not native controls, and the intel path can never run in an apple silicon only app
a smaller version could still be considered, a plain charge ceiling for macs before macos 26.4 that goes through the existing fan control helper, stays hidden until installed from the hub and leaves the monitor and menu bar untouched

Issue #1273, opened 2026-09-01
Focus follows mouse activates during extra-button desktop gestures


khichinho:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

With **Focus follows mouse** enabled, using an MX Master 3S gesture-button swipe to change desktops can produce unintended focus changes and extra Space transitions while the gesture is still in progress. The result is a jarring sequence of desktop swipes.

**Expected:** Focus follows mouse stays suspended while any mouse button is held, including an extra/gesture button.

**Actual:** the service only treats the left, right, and center buttons as held. An extra mouse button can therefore continue through the focus evaluation path during its drag.

The service already monitors `.otherMouseDragged`, so this appears to be a generic extra-button gap rather than a Logitech-specific integration issue. Its current `buttonsArePressed` check calls `CGEventSource.buttonState` only for `.left`, `.right`, and `.center`.

AppKit's `NSEvent.pressedMouseButtons` reports a bit for every currently pressed mouse button, including buttons 3 and above, and includes synthesized state independently of event delivery. Using that bitmask matches Vorssaint's existing extra-button support.

### Proposed fix

Gate focus evaluation on the complete AppKit button bitmask:

```swift
NSEvent.pressedMouseButtons == 0
```

A local patch keeps the existing modifier-key gate in a small pure helper and adds a regression check using extra-button bit 3.

Local verification:

- `./build.sh` — passed
- `./build/Vorssaint --selftest` — `SELFTEST OK`
- `./build.sh --test` — `TESTS OK (9527 checks)`
- Swift syntax and `git diff --check` — passed

The patched build has not yet been exercised with the reporter's MX Master 3S, so real-device confirmation is still needed.

### Steps to Reproduce

1. On an MX Master 3S, configure the gesture button in Logi Options+ for desktop navigation.
2. In Vorssaint 3.3.2, install and enable **Focus follows mouse**.
3. Hold the mouse gesture button and move horizontally to switch Spaces.
4. Repeat across windows and Spaces.
5. Observe unintended window activation and additional or reversing Space transitions.
6. Disable Focus follows mouse and repeat; the desktop gesture is stable.

### Feature Area

Mouse and keyboard

### Screenshots / Recordings

Not recorded.

### Vorssaint Version

3.3.2 (build 78)

### macOS Version

[...]

PR #1321, opened 2026-09-02
fix(focus): suspend during extra-button drags (#1273)

vorssaint:
## Summary

- suspend Focus follows mouse while any mouse button is held, including extra buttons
- keep the modifier-key guard inline and cover the complete button bitmask with a focused regression
- update the current release notes and credit the reporter

## Testing

- ./build.sh --test: TESTS OK (9705 checks)
- ./build.sh --dev --install
- installed Developer selftest: SELFTEST OK
- installed Developer signature verified with Developer ID
- ./build.sh
- ./build/Vorssaint --selftest: SELFTEST OK

Refs #1273

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,7 +87,7 @@ Vorssaint adds quit and close protections, expands screen text recognition, capt
 - Smooth Scroll and Scroll Inverter now step aside during fast user switching and resume on return, preventing background scroll stalls. Thanks to @iltonandrew.
 - Turning features on and off repeatedly no longer leaves unused keyboard and mouse listeners registered with the system for the rest of the session. Thanks to @PathGao.
 - Mouse navigation, button shortcuts and middle click now step aside outside the active login session or after Accessibility access is removed, preventing input stalls.
-- Focus follows mouse now leaves chosen apps alone and tracks the final pointer position during drags before changing focus.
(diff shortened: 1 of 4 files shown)


PathGao:
Right shape: the bug was the hand-enumerated three-button guard, and this deletes it instead of appending buttons 4 and 5 to the list. Net -6 lines and it covers every button.

I grepped the other `CGEventSource.buttonState` call sites on main for the same omission — `ShelfService`'s docked-drag watchdog and `WindowLayoutService.isTrackedButtonDown()`. Both ask about one specific button they already track, not "is anything held", so neither is the same shape. Nothing left over.

`NSEvent.pressedMouseButtons` is already how `MouseButtonShortcutService` decides which buttons are still physically down, from inside a tap callback, so its global behaviour is exercised in this app already. `evaluateIfSettled` is a main-runloop timer and reads `NSEvent.modifierFlags` one line below, so no new thread constraint. CHANGELOG entry is under `[Unreleased]`, right section.

Not verified: whether `pressedMouseButtons` reflects presses this app posts itself via `CGEventPost` the way `.combinedSessionState` did. That needs a real machine.

**Impact:** none beyond the fix — the guard only gets stricter, never looser.
**Merge:** already merged; nothing here would have held it.

ref #1273

Issue #1157, opened 2026-08-30
Clear clipboard history action


CSkjolden:
### Problem or Motivation

Would be nice to quickly with the command bar be able to wipe clipboard history.

### Proposed Solution

Add the "clear recent" button as an action. Maybe even make it possible to query, so be able to add "clear clipboard 10".

### Feature Area

Clipboard and files

PR #1322, opened 2026-09-02
Add Clipboard History clear action to Command Bar

vorssaint:
## Summary

Adds a localized Clear recent action to the Command Bar for Clipboard History. It reuses the existing clearing path, preserves pinned items, requires inline confirmation, and adds no background work, preference, or permission.

## Verification

- `./build.sh --test`: TESTS OK (9707 checks)
- `./build.sh --dev --install`
- Installed Developer executable: SELFTEST OK
- Developer ID signature and designated requirement validated
- No destructive live run was performed against user data; the compiled wiring and clearing contract were verified automatically.

## Anything else

Refs #1157


PathGao:
Checked `clearRecent()` in `ClipboardHistoryService`: it removes only `!isPinned` and prunes the batch selection, so the pinned-items claim holds. The confirmation also survives the global-shortcut path, since `needsPrompt` reads `confirmationPrompt` — a bound key can't run this silently. Good pick of gate.

Two notes, neither blocking:

**The keywords line isn't guarded.** The search assertion rebuilds the keyword string locally, and the source scan beside it checks id/title/confirmationPrompt/run but not `keywords`. Drop the `ClipboardFeatureStrings.enUS` line from the catalog and both assertions stay green. One clause closes it:

```swift
&& clipboardActionsCode.contains("ClipboardFeatureStrings.enUS.clearRecent")
```

**`action.clipboardWindow` has no English fallback.** `CommandBarSearch.matches` reads title + keywords only, so on a Japanese Mac "clipboard" now finds the row that deletes the history but not the row that opens it. That row is already touched here. Only two rows in the whole catalog use this fallback today, so leaving it is defensible — flagging the outcome, not asking for the line.

Read off `main` plus this diff; I ran nothing.

**Impact:** none — new row, no existing path changes.
**Merge:** ship it.

ref #1157

Issue #1331, opened 2026-09-03
Capture guide never mentions the R key that repeats the last region


marcelharinck:
The capture overlay can already re-capture the exact region you captured last, with R (`ScreenshotSelectionController.repeatLastRegion()`). The guide bar shows ↩, S, Z and esc, but never R, so unless you have read the source the feature does not exist.

The "show last capture outline" setting already draws a dashed ghost of that same region, so the app hints at the concept visually while hiding the key that acts on it.

Suggestion: show an R hint in the guide, but only when there is a region to repeat. The region is remembered for the session only and belongs to one display, so the bar should not advertise a key that would do nothing.

While reading the key handler I also noticed R fires with modifiers held (Command-R, Option-R), unlike S and Z which ignore modified keypresses.

I have a branch for this and will open a PR.

PR #1332, opened 2026-09-03
feat(screenshot): show the repeat-last-region key in the capture guide

marcelharinck:
Refs #1331

The overlay has always been able to re-capture the previous region with R, but the guide bar never said so, so the feature was unreachable unless you had read the source. This adds the missing hint. No capture behaviour changes.

The pill sits next to Z and uses `rectangle.dashed`, which echoes the dashed ghost outline the app already draws for that same region. It appears in both guide variants.

It only shows when pressing R would actually capture something. The stored region lives for the session and belongs to one display, so `repeatLastRegion()` and the hint now read one shared `repeatTargetPanel` property rather than restating the condition in two places — the bar cannot advertise a dead key. Worth knowing when reading the diff: each guide is built inside its panel's initialiser, before that panel joins `panels`, so the guides are refreshed once after the loop; without that the hint could never appear.

R also ignores modified keypresses now. Command-R and Option-R used to fire a capture, while S and Z already refused them — with the key advertised in the UI it should behave like its siblings. That reuses the existing `matchesShortcutKey` helper.

No new strings: the pill is icon plus key, exactly like ↩ and Z, so the thirteen language blocks are untouched.

One thing that reads like a gap but isn't: only a dragged region is stored. Window clicks and full-screen Enter never set a last region, so the hint stays hidden after those — existing behaviour, unchanged here.

Build, selftest and `--test` all pass (9697 checks). Three new checks in MetricsTests cover the offered case, nothing stored or its display gone, and the colour picker; I confirmed they go red when the condition is inverted. I also ran it: first capture of a session shows no pill, after a dragged region the pill appears and R reproduces the same region.

M3 Pro MacBook Pro, macOS 26.6.2 (25G83), MacOSX26.sdk, local `--dev` build, app in English. Not tested: multiple displays, and the bar's width in the longer languages — the pill carries no text, but the row now holds up to six pills when the magnifier is on.


PathGao:
The shared `repeatTargetPanel` is the right shape — one property behind both the key and the hint means they can't drift — and the refresh-after-the-loop comment explains an ordering that would otherwise read as a stray call. The body's claim about window clicks and Enter checks out: neither `confirmWindow` nor `captureFullDisplayUnderMouse` touches `lastRegion`.

On the multi-display case you flagged as untested, this much is visible in the code: `offersRepeatLastRegion` is controller-wide, so every panel's guide gets the same value and both bars show the pill, while `repeatLastRegion()` resolves to the single panel matching `lastRegion.displayID`. Press R with the secondary display's panel key and the capture lands on the other screen. That may be what you want — the key was already global — but "the bar cannot advertise a dead key" then holds per session, not per display. Worth deciding rather than leaving to omission.

Branch is conflicting; needs a rebase before it can go in.

**Impact:** none on one display. On two, the guide on the non-target screen advertises a key that acts elsewhere.
**Merge:** needs the rebase first, plus a call on the multi-display hint.

ref #1331

Issue #1327, opened 2026-09-03
[FR] Command Bar: switch to an ASCII keyboard layout while the bar is open


iva-zhu:
### Problem or Motivation

When the active input source is a non-Latin layout (Russian, Greek, a CJK input
method, and so on), the freshly opened Command Bar starts in that layout, so the
first letters of a query come out in the wrong alphabet and have to be retyped
or erased. The bar is meant for quick, keyboard-first queries, and the wrong
alphabet on open works against that every single time.

### Proposed Solution

An optional toggle in the Command Bar settings, off by default: when the bar
opens, switch the active input source to the first ASCII-capable keyboard layout
enabled on the Mac, and restore the previous input source when the bar closes.

This is the behavior Raycast has (it switches to a Latin layout while its bar is
up and restores it after), so the expectation comes from software people already
use.

Two notes on scope:

- It is a one-shot action on open and close, not an always-on focus watcher.
  Unlike #223 (declined for requiring one), nothing runs between the two events.
- Implementation can reuse the native Text Input Sources selection path that
  already landed for the Super Key solo action (#729, #749): select through
  `TISSelectInputSource`, no synthetic keystrokes, no marked-text querying, no
  new permission, no new dependency. The surface it adds is one setting and a
  couple of strings per locale.

### Feature Area

Tools

### Does another app already do this?

Yes — Raycast, natively.

### Mockups / References

- Raycast: Settings → Advanced → input source switching for the bar.
- Existing in-repo mechanism: `SuperKeyService.selectNextInputSource` (#749).

Borisserz:
Opened a PR for this: optional ASCII/Latin layout while the Command Bar is open (off by default), restoring the previous input source on close.

https://github.com/vorssaint/vorssaint-utils/pull/1457

iva-zhu:
@Borisserz Thanks for the PR! A heads-up: this feature already has an open PR — #1342,
opened the same day as the issue (Sep 3). The review feedback from Sep 4 was addressed
point-by-point the same day ([reply in the PR thread](https://github.com/vorssaint/vorssaint-utils/pull/1342#issuecomment-5540752886)),
and the branch was rebased onto current `main` yesterday.

@PathGao @vorssaint One thing I'd like to close out: #1342 still carries the `incomplete`
label from Sep 4, but the gaps named in that review (restore recording on `noErr`,
the `kTISNotifySelectedKeyboardInputSourceChanged` ripple in the body, the rebase +
[...]

PR #1342, opened 2026-09-03
feat(command bar): switch to an ascii layout while the bar is open

iva-zhu:
## Summary

When the active input source is a non-Latin layout, the freshly opened Command Bar starts in that layout, and the first letters of a query come out in the wrong alphabet. This adds an optional Command Bar setting (off by default): while the bar is open, the active input source switches to the first enabled ASCII-capable layout, and the previous source comes back when the bar closes — the behavior Raycast has, which is where the expectation comes from. Refs #1327

How it is built:

- A new `InputSourceSelection` helper in Core owns the Text Input Sources plumbing (enumerating the enabled, selectable keyboard sources, reading their properties, selecting by id and answering whether the system took the request). The Super key's input-source cycle now shares this copy instead of keeping its own, so the two paths cannot drift; the Super key's selection logic and its before-the-keystroke timing are untouched, and the other TIS sites only read layout data, so the switch shape stays consolidated here.
- Which source to borrow is a pure decision, pinned by tests: the first enabled ASCII-capable *layout* (never an input method), or nil when the current source already is an ASCII layout or none is enabled — a Mac set to Cyrillic and Greek alone borrows nothing, and then there is also nothing to restore. The previous source is recorded whenever TIS accepts the switch (`noErr`), which is the only answer `select()` gives: the two failures are not symmetric, and recording against a switch that never lands only restores a source the bar never left — a no-op — while a missing record would strand the typist on the borrowed layout with nothing to restore.
- The setting is read fresh on every open, like the other bar preferences; the switch is a one-shot per presentation, not a watcher — nothing of ours runs between open and close. One ripple worth naming: selecting a source posts `kTISNotifySelectedKeyboardInputSourceChanged`, which GlobalShortcut observes to rebuild its keycap-label cache, so a presentation with the borrow on posts it twice (borrow and restore), and every keycap label shown while the bar is open reads as the borrowed layout. That is presumably what a Latin bar wants the shortcuts to show, but it is a behavior, not a no-op. No new permission, no new dependency, nothing the maintainer has to carry.
- Permanent surface: one registered default (`commandBarASCIILayoutEnabled`, off by default, so it travels in settings backups automatically), one toggle row with a caption (under the More options disclosure), and two strings in all thirteen languages. The toggle says "ABC layout" — the name the Input menu itself uses — while the code says ASCII, because the first enabled Latin layout is not always the one named ABC (US, British, Dvorak all qualify).

## Verification

```sh
./build.sh --test              # 54,362/54,367 pass
```

The 5 remaining failures are the feet/inches formatting checks in `CommandBarUnits`, and they fail the same way on unmodified `main` on this machine: `CommandBarUnits` reads `Locale.current`, and this Mac runs under a Russian locale (decimal separator ","), while those checks expect ".". Unrelated to this change, but worth knowing when the suite runs here.

Used as the daily build (`build.sh --install` with the stable local signing identity) for two weeks on an Apple Silicon Mac, macOS 15.7.9 (24G830): Cyrillic ↔ ABC switches on open and returns on close (hotkey, click, Esc, Return) as expected, an already-Latin open switches nothing, a stale restore after a fast reopen stands down, the Super key tap keeps cycling sources as before, and the toggle off behaves exactly as before. Tested with English and Russian app languages.

Could not test: CJK input methods (none enabled here — the IME path follows the same native TIS selection the Super key switch already uses), multi-display and full-screen Spaces edge cases, and the remaining app languages beyond what the compiler-enforced string completeness covers.

## Updates from review

- The source to restore is recorded against TIS acceptance (`noErr`), not any read-back; the acceptance is the one gate, and the asymmetry (a switch that never lands restores a source never left — a no-op — while a missing record strands the typist) is in the doc comments.
- The body now names the `kTISNotifySelectedKeyboardInputSourceChanged` ripple through GlobalShortcut's keycap-label cache that the borrow posts twice per presentation.
- Rebased on current `main` (nine rebases so far; the per-language `commandBarValues.count` assertion upstream was since replaced by a per-child specifier-parity check, which now covers the new strings — the localizations suite passes).
- The registered-default and backup-export shape the compact mode checks use now pins this key down too: ships off, rides in `exportKeys()`, and a restored value has to be a real switch.
- The toggle moved under a More options disclosure row like the files section, reusing the shared `moreOptions` label — no new strings.
- `select()` answers only whether TIS accepted the switch; the immediate read-back lookup (nothing read it) is gone from the open path.

## Anything else

Refs #1327

## DEMO

https://github.com/user-attachments/assets/09eb8cda-7f9f-4136-99b0-c3887d5116ea


PathGao:
The shared `InputSourceSelection` is the right shape. I grepped `src/`: the other TIS sites (`GlobalShortcut`, `MouseNavigationKeys`) read layout data rather than select a source, so the switch shape is fully consolidated.

Two things.

1. The read-back gate sits on the wrong side of the asymmetry. A false positive costs a restore to a source you never left, which is a no-op. A false negative, the read-back lagging the switch by a turn, strands the user on ABC in every app with nothing left to restore. Record `currentID` whenever `TISSelectInputSource` returns `noErr`; keep the read-back, just not as the gate.

2. Body says "nothing runs between open and close". The borrow posts `kTISNotifySelectedKeyboardInputSourceChanged`, which `GlobalShortcut` observes to rebuild its whole keycap-label cache (#1047's machinery), twice per presentation, on the keystroke path. Every keycap shown while the bar is open reads as ABC. Probably what you want, but it isn't in the body.

No machine here, so the timing in 1 is reasoning, not a repro. I did not check the IME path.

**Impact:** toggle on plus a lagging read-back leaves the layout on ABC until the user switches back by hand. Off by default, so nobody is hit until they enable it.
**Merge:** needs a rebase first, and recheck `commandBarValues.count == 153` after it, that assertion is where two command bar PRs collide.

ref #1327

iva-zhu:
Thanks for the review — all three points addressed in ae27684 (force-pushed; rebased once more onto the mouse-buttons commit that landed mid-review):

1. The source to restore is now recorded whenever `TISSelectInputSource` returns `noErr`. The read-back stays, reported alongside acceptance as a signal, off the decision — the asymmetry (a false positive restores a source never left, a false negative strands the typist on the borrowed layout) is written into the doc comments.
2. The body now names the ripple: the borrow posts `kTISNotifySelectedKeyboardInputSourceChanged` twice per presentation (borrow and restore), GlobalShortcut rebuilds its keycap-label cache on both, and keycap labels shown while the bar is open read as the borrowed layout.
3. Rebased on current `main`; `commandBarValues.count == 153` re-checked after it — the Command Bar block kept its width, the assertion holds. Suite here: 9952/9957, the 5 remaining are the pre-existing `CommandBarUnits` locale failures noted in the body.

On the IME path: untouched by this change — it follows the same native TIS selection the Super key switch already uses — but it remains untested on my side, as noted in the body.

iva-zhu:
Rebased again onto current `main` (`9c49f2a`) — upstream since rewrote the
test harness (`build.sh` now builds a generated `TEST_SOURCES` list, and the
per-language `commandBarValues.count` assertion was replaced by a
specifier-parity check across all languages), so the two conflicts from that
rework resolved on its side; the InputSourceSelection block rides along in the
new list and the suite reads 5 failures, all the known `CommandBarUnits`
locale ones. The three review points above stay addressed; still
rebase-friendly whenever you sequence this against the other open
command-bar PRs (#1103, #1544, #1572, #1605).

iva-zhu:
Rebased onto current `main` (`13c29f7`, 57 commits landed since `69fdf04` — the Dynamic Island beta line included). The rebase was clean: no conflicts, the ASCII-layout block and the specifier-parity expectations all hold.

Suite on this machine: 51,564 of 51,569 checks pass — the 5 remaining are the pre-existing `feet/inches` locale tests from #1344, unchanged by this PR.

Noting for scheduling that the feature is in daily use on this machine and the PR is rebase-friendly — happy to jump on any ordering you prefer relative to the other open command-bar PRs. I'm quite eager to see this one land whenever review time allows 🙂

Sources/Vorssaint/Core/Defaults.swift
```diff
@@ -1397,6 +1399,7 @@ enum Defaults {
         DefaultsKey.scratchpadShortcut: GlobalShortcut.scratchpadDefault.storageValue,
         DefaultsKey.commandBarShortcutEnabled: false,
         DefaultsKey.commandBarCompactMode: false,
+        DefaultsKey.commandBarASCIILayoutEnabled: false,
```
vorssaint, line 1422: this key rides along in settings backups because it is registered, but nothing pins that down the way the compact mode checks do. can you add the same registered default and backup export checks for this key

Sources/Vorssaint/UI/Settings/CommandBarSettings.swift
```diff
@@ -95,6 +96,12 @@ struct CommandBarSettings: View {
                         .font(.caption)
                         .foregroundStyle(.orange)
                 }
+                // Like compact mode this needs no callback: the bar reads the
+                // toggle on every open, so there is no live state to sync.
+                Toggle(text.asciiLayoutToggle, isOn: $asciiLayoutEnabled)
+                Text(text.asciiLayoutCaption)
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
```
vorssaint: the top section already has five controls, and most people will never need this one. can it go under a more options row like the files section has, so the main list stays short

Sources/Vorssaint/Core/InputSourceSelection.swift
```diff
@@ -0,0 +1,106 @@
...
+    /// symmetric: recording against a switch that never lands restores a
+    /// source the caller never left — a no-op — while failing to record one
+    /// that lands strands the typist on the borrowed layout, so the record
+    /// belongs to acceptance, and the read-back stays off the decision.
+    static func select(sourceID: String) -> Outcome {
+        guard let source = selectableInputSources().first(where: {
+            inputSourceString($0, property: kTISPropertyInputSourceID) == sourceID
+        }) else { return Outcome(accepted: false, readBackID: nil) }
+        let accepted = TISSelectInputSource(source) == noErr
+        return Outcome(accepted: accepted, readBackID: accepted ? currentSourceID() : nil)
```
vorssaint: nothing reads the read back, so the extra lookup after every accepted switch only adds time on the open path. can the switch just answer whether it was accepted

iva-zhu:
Thanks for the review — all three points are in `d9accd4` (force-pushed; also rebased onto current `main`, `0d93163`).

> Defaults.swift:1402 — can you add the same registered default and backup export checks for this key

Added, right next to the existing ASCII layout block in `MetricsTests`, same shape as the compact mode checks: the registered default asserts it ships off, `exportKeys()` has to contain the key, and a restored value has to be a real switch (`valueLooksRight(key, true) && !valueLooksRight(key, "yes")`).

> CommandBarSettings.swift:104 — can it go under a more options row like the files section has

Moved. The toggle and its caption now sit under a `DisclosureHeaderRow` like the files section's, reusing the shared `moreOptions` label the other pages use, so the main list stays at its old width and no new strings were needed.

> InputSourceSelection.swift:69 — can the switch just answer whether it was accepted

Agreed — nothing ever read it. `select()` now returns just `Bool` acceptance, the `Outcome` struct and the immediate read-back lookup are gone, and the one TIS lookup the open path does is the one that decides what to restore. The record-on-acceptance asymmetry stays in the doc comments: a switch that never lands restores a source never left (a no-op), a missing record strands the typist.

Suite on this machine after the rebase: 54,362/54,367 — the 5 remaining are the known pre-existing feet/inches locale tests. Installed from the branch (`./build.sh --install`) and back in daily use.

PR #1343, opened 2026-09-03
feat(screenshot): add arrow style selection

kxnstandin:
## Summary

Adds arrow style selection to the existing screenshot editor.

The arrow tool now offers filled, outline, open, double-ended and scribbly styles. Scribbly arrows receive a slightly different seeded design when created while remaining visually stable when the same annotation is redrawn.

This also fixes two selection-state problems in the existing editor:

- Finishing a new arrow no longer briefly activates the layer control and then leaves it gray.
- Selecting a thin arrow after a thick arrow updates the toolbar to show the selected arrow's actual thickness and style.

This belongs in Vorssaint because it extends the existing screenshot annotation tool without adding a new subsystem. It is built on the current annotation model, renderer, toolbar and defaults system.

The existing filled arrow remains the default. The additional behavior is only used by people who open the screenshot editor and use arrows.

There are no new dependencies, processes, network requests, permissions or third-party integrations. The permanent surface is limited to the arrow-style model, renderer paths, one saved preference, localized labels, toolbar previews and screenshot regression tests.

## Verification

Tested on an Apple Silicon Mac running macOS 27.0.

- Developer build created with `./build.sh --dev`
- Release build created with `./build.sh` without warnings
- `./build/Vorssaint --selftest` returned `SELFTEST OK`

Manually verified:

- Drawing a new arrow keeps the layer-selection state consistent after releasing the pointer.
- Selecting thin and thick arrows updates the toolbar controls.
- Filled, outline, open and double-ended arrows render correctly.
- Creating multiple scribbly arrows produces different designs.
- A scribbly arrow keeps the same design when redrawn.

`./build.sh --test` completed 9,747 checks but reported two pre-existing failures unrelated to this change:

- `App Switcher icon-row hints describe default app and window shortcuts`
- `the dispatch pool starvation this check needs was actually reached`

Not tested:

- Other Mac models or macOS versions
- Multi-display behavior
- Different Spaces layouts

All new user-facing labels compile in every supported language.

## Anything else

I could not find an existing issue specifically covering screenshot arrow-style selection. New issue creation is currently restricted for this repository, so there is no `Refs #N` line to include.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
@@ -49,6 +49,13 @@ final class ScreenshotEditorModel: ObservableObject, BackdropEditing {
             applyStyleToSelection()
         }
     }
+    @Published var arrowStyle: ScreenshotSupport.ArrowStyleID {
+        didSet {
+            UserDefaults.standard.set(arrowStyle.rawValue,
+                                      forKey: DefaultsKey.screenshotLastArrowStyle)
+            applyStyleToSelection()
+        }
+    }
     @Published var sticker: ScreenshotSupport.StickerID {
         didSet {
             UserDefaults.standard.set(sticker.rawValue,
@@ -144,6 +151,8 @@ final class ScreenshotEditorModel: ObservableObject, BackdropEditing {
             defaults.string(forKey: DefaultsKey.screenshotLastColor))
         stroke = ScreenshotSupport.StrokeID.sanitized(
             defaults.string(forKey: DefaultsKey.screenshotLastStroke))
+        arrowStyle = ScreenshotSupport.ArrowStyleID.sanitized(
+            defaults.string(forKey: DefaultsKey.screenshotLastArrowStyle))
         sticker = ScreenshotSupport.StickerID.sanitized(
             defaults.string(forKey: DefaultsKey.screenshotLastSticker))
         annotationShadowsEnabled = defaults.bool(
@@ -436,15 +445,26 @@ final class ScreenshotEditorModel: ObservableObject, BackdropEditing {
 
     // MARK: - Selection styling
 
-    /// Applies color or thickness changes to the selected annotation.
+    /// Applies color, thickness, or arrow style changes to the selected annotation.
     private func applyStyleToSelection() {
         guard let selectedID,
               let index = annotations.firstIndex(where: { $0.id == selectedID })
         else { return }
-        guard annotations[index].color != color || annotations[index].stroke != stroke else { return }
+        let arrowStyleChanged = annotations[index].tool == .arrow
+            && annotations[index].arrowStyle != arrowStyle
+        guard annotations[index].color != color
+                || annotations[index].stroke != stroke
+                || arrowStyleChanged
+        else { return }
         registerUndo()
         annotations[index].color = color
         annotations[index].stroke = stroke
+        if annotations[index].tool == .arrow {
+            if arrowStyleChanged, arrowStyle == .scribbly {
+                annotations[index].scribbleSeed = ScreenshotSupport.randomScribbleSeed()
+            }
+            annotations[index].arrowStyle = arrowStyle
+        }
         if annotations[index].tool == .text {
             annotations[index].rect = ScreenshotRenderer.textBounds(
                 annotations[index].text,
@@ -475,6 +495,11 @@ final class ScreenshotEditorModel: ObservableObject, BackdropEditing {
             beginSelectDrag(at: point)
             return
         }
+        if tool != .select, tool != .crop {
+            selectedID = ScreenshotSupport.selectionAtStartOfAnnotationDrag(
+                current: selectedID,
+                editingSelectedAnnotation: false)
+        }
         switch tool {
         case .select:
             beginSelectDrag(at: point)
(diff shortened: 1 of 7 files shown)


PathGao:
`selectionStyle(for:)` is the right shape — one struct between the picked mark and the three toolbar controls beats three ad-hoc assignments.

Gap: `selectExistingAnnotation` is the other selection entry point (creation-tool tap, four call sites in `endDrag`) and didn't get the sync. Tap an existing thin outline arrow while the arrow tool is active — it selects and flips to `.select`, but the toolbar still holds the last arrow's stroke and style. Change the color next and `applyStyleToSelection` writes all three back, silently making that arrow thick and filled. Same shape as your `beginSelectDrag` fix, in the block that already nils `selectedID` for stickers:

```swift
let style = ScreenshotSupport.selectionStyle(for: hit)
color = style.color
stroke = style.stroke
arrowStyle = style.arrowStyle
```

Minor: `selectionAtStartOfAnnotationDrag` can only ever receive `editingSelectedAnnotation: false` — the `true` branch already returned at the `selectedAnnotationOwns` check above the call site. It's `selectedID = nil` with a parameter that has one possible value.

I read the diff and `main`; no build, no run, rendering output unverified.

**Impact:** tap-selecting an arrow while a creation tool is active shows stale toolbar values, and the next edit overwrites that arrow's thickness and style. The stroke half predates this PR; the arrow-style half is new.
**Merge:** needs the selection sync in `selectExistingAnnotation` first.

kxnstandin:
> `selectionStyle(for:)` is the right shape — one struct between the picked mark and the three toolbar controls beats three ad-hoc assignments.
> 
> Gap: `selectExistingAnnotation` is the other selection entry point (creation-tool tap, four call sites in `endDrag`) and didn't get the sync. Tap an existing thin outline arrow while the arrow tool is active — it selects and flips to `.select`, but the toolbar still holds the last arrow's stroke and style. Change the color next and `applyStyleToSelection` writes all three back, silently making that arrow thick and filled. Same shape as your `beginSelectDrag` fix, in the block that already nils `selectedID` for stickers:
> 
> ```swift
> let style = ScreenshotSupport.selectionStyle(for: hit)
> color = style.color
> stroke = style.stroke
> arrowStyle = style.arrowStyle
> ```
> 
> Minor: `selectionAtStartOfAnnotationDrag` can only ever receive `editingSelectedAnnotation: false` — the `true` branch already returned at the `selectedAnnotationOwns` check above the call site. It's `selectedID = nil` with a parameter that has one possible value.
> 
> I read the diff and `main`; no build, no run, rendering output unverified.
> 
> **Impact:** tap-selecting an arrow while a creation tool is active shows stale toolbar values, and the next edit overwrites that arrow's thickness and style. The stroke half predates this PR; the arrow-style half is new. **Merge:** needs the selection sync in `selectExistingAnnotation` first.

Hi @PathGao,

Addressed the selection-sync issue in `298dc63` and the layer-control visibility issue in `7a9d2cc`. Required workflows remain pending maintainer approval.

Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift
```diff
@@ -1315,10 +1331,28 @@ enum ScreenshotSupport {
...
+    struct SelectionStyle: Equatable {
+        let color: ColorID
+        let stroke: StrokeID
+        let arrowStyle: ArrowStyleID
+    }
+
+    static func selectionStyle(for annotation: Annotation) -> SelectionStyle {
+        SelectionStyle(color: annotation.color,
+                       stroke: annotation.stroke,
+                       arrowStyle: annotation.arrowStyle)
```
vorssaint: picking a sticker or a pixelated block now also rewrites the color and thickness the next new mark will use, and picking any mark that is not an arrow quietly puts the arrow style back to solid. can the controls only take over the values the picked mark actually uses

Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
```diff
@@ -497,15 +520,18 @@ final class ScreenshotEditorModel: ObservableObject, BackdropEditing {
             registerUndo()
             dragRegistered = true
             let annotation = ScreenshotSupport.Annotation(
-                tool: tool, points: [point, point], color: color, stroke: stroke)
+                tool: tool, points: [point, point], color: color, stroke: stroke,
+                arrowStyle: arrowStyle)
             annotations.append(annotation)
+            selectedID = annotation.id
```
vorssaint: the selection handles now draw over a shape while it is still being dragged out, and the tip of a new arrow sits under a handle while you aim it. clearing the old selection at the start is enough to stop the layer buttons from flashing, so can the new mark be selected only on release as before

Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
```diff
@@ -547,11 +574,15 @@ final class ScreenshotEditorModel: ObservableObject, BackdropEditing {
         }
         selectedID = hitTest(point)
         if let selectedID, let hit = annotations.first(where: { $0.id == selectedID }) {
+            let selectionStyle = ScreenshotSupport.selectionStyle(for: hit)
+            self.selectedID = nil
+            color = selectionStyle.color
+            stroke = selectionStyle.stroke
+            arrowStyle = selectionStyle.arrowStyle
```
vorssaint: clicking without moving on an end handle of the picked arrow can select another mark drawn over that spot, and then the controls keep the old values. can the controls follow every change of selection, so a color change never carries the previous thickness and arrow style onto that mark

Issue #1180, opened 2026-08-31
Keyboard brightness shortcuts


944kr5zkwv-hub:
### Problem or Motivation

I was looking for a shortcut to quickly increase and decrease the keyboard brightness on a MacBook.

On MacBooks with a dedicated keyboard brightness control, this can be done directly using the keyboard. However, on newer MacBook models, there are no dedicated keys for adjusting the keyboard backlight anymore, which means you have to open Control Center and manually adjust the brightness slider.

It would be very useful to have the ability to assign custom keyboard shortcuts to increase and decrease the keyboard brightness.

For example:

- ⌘ + F1 → decrease keyboard brightness
- ⌘ + F2 → increase keyboard brightness

Or alternatively, allow the user to choose any custom key combination.

The shortcuts should ideally work globally, even when the app is not focused, so the keyboard brightness can be adjusted from anywhere.

This would be especially useful on MacBooks where Apple removed the dedicated keyboard brightn

### Proposed Solution

Adding this feature 

### Feature Area

Mouse and keyboard

PathGao:
There's already a **Keyboard light** switch in Quick toggles. It turns the backlight on and off, but that's all it does, and you have to open the panel to get to it. Stepping the level up and down from a shortcut isn't in the app today.

I've marked this for the maintainer to schedule. Nothing has changed on your Mac yet. If it gets built, you'll get a comment on this issue with the version to install.

<details>
<summary>The technical side — you can skip this</summary>

The write path already exists: `KeyboardLightBridge.setBrightness` in `BrightnessService.swift` drives the Quick toggles switch and already remembers the last non-zero level. An increase/decrease action is a read-modify-write on top of that plus a shortcut binding, not new capability — that's why I read the landing point as clear.

Not verified: whether the private setter takes arbitrary intermediate levels reliably. Today it only ever writes 0 or one fixed on-level, so the stepping range is untested.

</details>

vorssaint:
these shortcuts are available in 3.3.5 on macs with a supported keyboard backlight. install displays from settings > features, then open keyboard shortcuts > mouse and keyboard > keyboard light. turn on 'use keyboard brightness shortcuts' and record your preferred keys for decrease and increase. they work while other apps are in front. closing this as implemented

PR #1352, opened 2026-09-04
feat(keyboard): add backlight shortcuts

EugeneCarldotme:
Refs #1180

## What was wrong

Before: macOS provided no global shortcut for changing keyboard-backlight brightness. Adjusting it required opening Control Center and moving the Keyboard Brightness slider.

The feature request suggested Command-F1 and Command-F2, but macOS reserves those hardware keys and they are not reliably available to this shortcut path.

## The change

After: Vorssaint provides editable global shortcuts for decreasing and increasing the built-in keyboard backlight. The defaults are Option-Command-minus and Option-Command-equal.

Each press reads the current hardware brightness instead of trusting the last value Vorssaint wrote. This keeps the shortcut in sync with changes made through Control Center or ambient-light adjustment. It moves the backlight by one 1/16 step and clamps the result between off and full brightness.

The shortcuts use the existing keyboard-light bridge and appear only on Macs where that bridge is available. Find them under App > Keyboard shortcuts > Mouse and keyboard > Keyboard light.

This adds no dependency, background timer, network access or third-party integration. The change is 7 files, +191/-28.

## Verification

Tested on macOS 26.5 (25F71).

```sh
./build.sh
# Passed; release app bundle assembled and ad-hoc signed.

# SELFTEST OK

./build.sh --test
# 9,928 of 9,929 checks passed. One unrelated dispatch-pool timing check failed:
# "the dispatch pool starvation this check needs was actually reached"
# The same unrelated check failed on both full-suite runs.

git diff --check
# Passed
```

Manually launched the release build and confirmed that App > Keyboard shortcuts > Mouse and keyboard > Keyboard light shows Option-Command-minus and Option-Command-equal as its defaults.

Screenshot
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,19 +7,23 @@ All notable changes to this project are documented here. The format follows
 ## [Unreleased]
 
 ### Summary
-Vorssaint adds edge snap controls and a shortcut for recent captures, refines the radial menu, and improves color picking, app icons, readability, localization and mouse reconnection behavior.
+Vorssaint adds edge snap controls and shortcuts for recent captures and keyboard brightness, and refines the radial menu. It improves color picking, app icons, readability, localization, mouse reconnection, system shortcut handling and build checks.
 
 ### Added
+- Optional keyboard backlight shortcuts adjust brightness one step at a time, available in Keyboard shortcuts. Thanks to @EugeneCarldotme.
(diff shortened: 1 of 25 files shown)


Tomaubier:
Awesome! It would be nice to be able to set keyboard shortcuts to change the brightness of external monitors! Would you be able to do such a thing? 🤓

EugeneCarldotme:
Yup, i implemented this for #1488

Issue #1348, opened 2026-09-04
Super key press function also inputs control


gatzifratzi:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Im using the super key feature with hold for super key and press for escape. Im using the Zed Editor and cant exit the vim insert Mode in that editor with a press of capslock. It works to exit the insert mode in vim, nvim and obsidians vim motions. The keymap settings of that editor has a record keybind functionality. With that I tested what the editor perceives as the press key.
I noticed, that before the escape input, a control input is listed when I press capslock.
In the programms in mentioned before and to exit fullscreen applications etc. a control+escape input achieves the same functionality as a single escape input. Just in Zed a control+escape input is ignored. I will fix my issue by binding control+escape to the wished functionality in Zed but i believe this is a bug. Why would there be a control input infront of the escape input. The control input before the capslock button also occurs, when the super key is bound without control (cmd+option+shift)

### Steps to Reproduce

use super key with escape for press in vorssaint. 
use zed editor with vim mode. 
try to exit insert mode with press of capslock.

### Feature Area

Mouse and keyboard

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

macos in english(US) and German

### Mac and displays

MB Pro 14" with M4 pro chip

### Installation Method

Homebrew

PathGao:
You're onto something. When you tap Caps Lock, Vorssaint is the thing that sends that Escape, and it doesn't strip the modifier keys off before sending it, so a stray modifier can ride along. Most apps ignore it. Zed doesn't.

One thing about your workaround: if you bind control+escape in Zed, keep a plain escape binding too. Once this is fixed the extra control goes away, and a control-only binding would stop working.

Nothing has changed on your Mac yet. When a fix is in a release you'll get a comment here with the version to install.

<details>
<summary>The technical side — you can skip this</summary>

`SuperKeyService.postKey` builds the Escape from a `.hidSystemState` source and posts it without ever assigning `flags`. It is the only synthesized-keystroke site in the app that doesn't — `TextSnippetService`, `FinderCutPaste`, `SpaceWindowBridge`, `DockClickService`, `RadialMenuService` and the rest all set `flags` explicitly, including to `[]`. Same shape as #937.

[...]

PR #1358, opened 2026-09-04
Post the Super Key Escape with no modifier flags

hash00:
**Refs #1348** — Super key press function also inputs control

`SuperKeyService.postKey` is the only site in the app that builds a `CGEvent(keyboardEventSource:)` and posts it without ever assigning `flags`. An event created from a `.hidSystemState` source starts with the modifier state that source reports, so a modifier still registered as down rides along and the solo-tap Escape arrives as Control-Escape. Most apps ignore the difference; Zed's vim mode does not, which is how the reporter found it.

The other synthesized-keystroke sites — `TextSnippetService`, `FinderCutPaste`, `DockClickService`, `MouseButtonShortcutService`, `MouseNavigationService` — all set `flags` explicitly, including to `[]`. This makes the Super Key path match.

One thing I can't answer: why the stray flag is Control specifically. The Escape is posted from `performSoloAction`, which runs after `.triggerUp`, so the trigger's own modifier should already be clear by then. I couldn't reproduce on my hardware, so clearing the flags is the fix I'm confident in, not a confirmed account of where the Control comes from. The reporter is best placed to confirm.

## Validation

- `./build.sh --test` — TESTS OK (9924 checks)
- `./build.sh` — exit 0, 0 warnings
- `--selftest` — SELFTEST OK

Not verified on hardware: I don't have a setup that reproduces the stray Control.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to this project are documented here. The format follows
 - The zoom panel's button says what it does instead of borrowing the timeline's hint to click somewhere else.
 - Slider labels shrink instead of being cut where Turkish and Spanish run past the column, in the backdrop and recording panels.
 - Clicks and scrolling no longer lag in full-screen apps and games while the three-finger middle click or the reversed scroll direction is on.
+- Tapping Super key for Escape no longer carries a modifier still reported by the keyboard. Thanks to @gatzifratzi, @PathGao and @hash00.
 
 ## [3.3.3-beta.4] - 2026-09-03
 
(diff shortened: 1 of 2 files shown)


hash00:
Good catch — that wording was wrong. `DockClickService` posts ⌘⌥M and `MouseButtonShortcutService` posts a real shortcut's modifiers, so "post a bare key, as every other synthesized keystroke does" misdescribes them.

The accurate claim is narrower: every other site *assigns* `flags` explicitly, whether that's a shortcut's modifiers or none at all. This one never assigned at all. Comment reworded to say that.

Also rebased onto main (`a787c62`), since the branch had fallen behind. Re-ran the gate after the rebase: TESTS OK (9934 checks), build exit 0 with 0 warnings, SELFTEST OK.

Issue #1355, opened 2026-09-04
App Switcher take-over switches off macOS ⌘⇧3 and leaves ⌘⇧` with macOS


owendaw:
### Description

With "Replace macOS ⌘Tab and ⌘`" switched on, setting the App Switcher's Apps or Windows shortcut to a combination on the 3 key (⌘3, or ⌘⇧3 directly) switches off macOS's own ⌘⇧3, "Save picture of screen as a file", for as long as Vorssaint runs. Pressing ⌘⇧3 opens the Vorssaint switcher in reverse and takes no screenshot. Quitting Vorssaint gives the key back.

Separately, with the default shortcuts the take-over disables ⌘Tab, ⌘⇧Tab and ⌘` but leaves ⌘⇧` (previous window) with macOS, so the reverse window direction is only half taken over.

Both come from one constant: `SwitcherNativeSymbolicHotKey.previousWindow = 28`. In the WindowServer's symbolic hotkey table, 28 is the screenshot key; the previous-window key is 220. The take-over matches each id's live combination against the switcher shortcuts, allowing an extra Shift for the reverse direction, so a switcher shortcut on the 3 key "owns" id 28.

Expected: ⌘⇧3 keeps taking screenshots whatever the switcher shortcuts are; ⌘⇧` is taken over alongside ⌘`.

### Steps to Reproduce

1. Settings › App switcher: turn on "Replace macOS ⌘Tab and ⌘`".
2. Record ⌘3 as the Apps shortcut (the recorder accepts it; ⌘3 is free).
3. Press ⌘⇧3.

Actual: reading the live table (`CGSIsSymbolicHotKeyEnabled`) at this point shows id 27 and id 28 disabled and Vorssaint's own marker `switcherNativeHotkeysSuppressed` = (27, 28) — macOS's screenshot key has been switched off in the WindowServer. Quit Vorssaint: both ids re-enable, the marker clears, and ⌘⇧3 takes a screenshot again. (While the switcher is active the tap also swallows ⌘⇧3 as the reverse of ⌘3, so no screenshot is taken either way on this Mac; the WindowServer disable is what additionally kills the key in apps where the switcher is paused, after the tap is lost, and after a crash.)

For the second half: leave the defaults, turn the take-over on, and read the table: id 220 (⌘⇧`) stays enabled.

### Feature Area

App switcher

### Screenshots / Recordings

Live table before and after step 2, read on this Mac:

```
[...]

PR #1366, opened 2026-09-04
Point previous window at symbolic hotkey 220, not 28

hash00:
**Refs #1355** — App Switcher take-over switches off macOS ⌘⇧3 and leaves ⌘⇧` with macOS

`SwitcherNativeSymbolicHotKey.previousWindow` was 28. 28 is "Save picture of screen as a file"; the previous-window key is 220. Since the take-over matches each id's live combination against the switcher shortcuts and allows an extra Shift for the reverse direction, the wrong id both let a switcher shortcut on the 3 key claim ⌘⇧3, and left the real ⌘⇧` with macOS so the reverse direction was only half taken over.

The reporter diagnosed this precisely; I only confirmed it and changed the constant.

## Confirmation

Read back from the live WindowServer table on macOS 27.0 via `CGSGetSymbolicHotKeyValue`:

```
id=27   keyCode=50 (`)  mods=0x100000 (⌘)    → ⌘`   next window
id=28   keyCode=20 (3)  mods=0x120000 (⌘⇧)   → ⌘⇧3  screenshot to file
id=220  keyCode=50 (`)  mods=0x120000 (⌘⇧)   → ⌘⇧`  previous window
```

The existing coverage in `MetricsTests.swift` already models `.previousWindow` as ⌘⇧` on the grave key, so the tests were describing the intended semantics while the raw value disagreed. They pass unchanged.

## Validation

- `./build.sh --test` — TESTS OK (9944 checks)
- `./build.sh` — exit 0, 0 warnings
- `--selftest` — SELFTEST OK

Not verified end to end: I did not exercise the take-over against a live WindowServer with the option enabled, so the id mapping above is read-back evidence rather than a behavioural test. Worth a confirmation from the reporter, who can reproduce both halves.

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
@@ -50,7 +50,13 @@ enum SwitcherNativeSymbolicHotKey: Int32, CaseIterable, Hashable {
     case commandTab = 1
     case commandShiftTab = 2
     case nextWindow = 27
-    case previousWindow = 28
+    // Previous window is 220, not 28. 28 is "Save picture of screen as a
+    // file", the ⌘⇧3 screenshot key, which sits on the 3 key rather than the
+    // grave key. Because the take-over matches each id's live combination
+    // against the switcher shortcuts and allows an extra Shift for the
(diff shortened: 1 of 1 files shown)


vorssaint:
thanks for confirming the shortcut problem. the same correction was merged in #1357 and is included in 3.3.5, so i'm closing this as already addressed

Issue #1377, opened 2026-09-04
Snippet expansion strands the trigger's first character when the browser offers an inline completion


rhukster:
### Description

Typing a snippet trigger in the Chrome or Safari address bar can leave the trigger's first character in front of the replacement. With a trigger of `gitgrav` expanding to `https://github.com/getgrav/grav/`, the field ends up holding `ghttps://github.com/getgrav/grav/`.

It looks trigger-specific, and that is what makes it confusing to report: `gittril` and `gityet` never do it, and neither does `gitgrav` in a text editor.

**Cause.** Both browsers inline-autocomplete as you type and leave the completion *selected*. Vorssaint swallows the trigger's last character, so at expansion time the field holds `gitgra` — and the omnibox has appended a selected `ph`, offering `gitgraph`. `postExpansion` then posts `trigger.count - 1` deletes, but the first one removes the **selection** rather than a character, so six deletes clear five characters and the leading `g` survives.

Whether it happens depends on whether that particular prefix has a completion, which is why it reproduces perfectly on one trigger and never on another. The completion is network-backed, so it is identical across both browsers and absent from local history.

### Steps to Reproduce

1. Add an immediate snippet, trigger `gitgrav`, replacement `https://github.com/getgrav/grav/`.
2. Click into the Chrome or Safari address bar and type `gitgrav`.
3. Observed: `ghttps://github.com/getgrav/grav/`. Expected: `https://github.com/getgrav/grav/`.

Turning off Chrome's "Autocomplete searches and URLs" makes it go away, which confirms the mechanism.

### Environment

Vorssaint `7bb9c80` (local build of `main`), macOS 26.6.2 (25G83), Mac16,6, Chrome 152.0.7977.76, Safari 26.6.2.

### Notes

Related to #937, which censuses the eleven places Vorssaint posts synthetic keys but only tracks how they are *sent* — nothing there considers what the receiving field might do to them.

PR #1378, opened 2026-09-04
fix(snippets): clear an inline completion before deleting a trigger

rhukster:
Refs #1377

**Before.** A snippet expanded in a browser address bar leaves the trigger's first character in front of the replacement, when the prefix has an inline completion. `gitgrav` produces `ghttps://github.com/getgrav/grav/`; `gittril` is fine, because `gittri` has no completion.

**After.** The trigger is cleared completely and the replacement stands alone.

**Why.** Browsers append their completion as *selected* text, so the first backspace consumes the selection instead of a character and the counted deletes come up one short. The fix asks the focused element for `kAXSelectedText` before deleting and posts one extra delete when it is non-empty. A user-made selection cannot be mistaken for it: typing into a selection replaces it, so anything still selected when a trigger completes was added by the app in front.

The AX query is capped with `AXUIElementSetMessagingTimeout` and a non-answer is treated as no selection — the assumption the code already makes — so an app that does not respond cannot stall the event tap. #1110 is a bug about exactly that kind of hang in Chrome.

**Other callers.** `postExpansion` had two delete loops, the typed path and the multi-line paste path added in #917. Both now go through one `postDeletes` helper, so this is fixed for multi-line snippets too rather than only for the reported case. No other synthetic-key call site in #937's census posts deletes.

**Test.** Four assertions in `MetricsTests`. The AX path cannot be unit-tested, so this is a source-shape test pinning the contract: that the selection is queried, that the query is bounded, and that neither delete site bypasses the helper. Verified in both directions — green with the fix, all four red with the source reverted.

**Verified on** Mac16,6, macOS 26.6.2 (25G83), en-US, self-signed local `--dev` build, against Chrome 152.0.7977.76 and Safari 26.6.2. Watched the old and new behaviour directly: reproduced the stranded character, applied the fix, confirmed it cleared.

`./build.sh` finishes with no warnings, `--selftest` passes, and `./build.sh --test` reports 1 failure of 9972 — "the dispatch pool starvation this check needs was actually reached", which fails identically on unmodified `main` (1 of 9968).

**Not tested:** Intel, older macOS, other browsers (Arc, Firefox, Edge), non-Latin input methods, and whether an app that re-offers a completion *after* the first delete would cascade — neither browser does, since both suppress inline autocompletion once you start deleting.

### DIFF
--- a/Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
+++ b/Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
@@ -2,6 +2,7 @@
 // Copyright (C) 2026 Vorssaint
 
 import AppKit
+import ApplicationServices
 import Carbon.HIToolbox
 import CoreGraphics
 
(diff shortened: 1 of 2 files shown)


Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
```diff
@@ -15,6 +16,36 @@ final class TextSnippetService {
...
+    /// clear one character too few and the trigger's first character is left
+    /// stranded in front of the replacement. A selection of the user's own
+    /// cannot be confused with it — typing into one replaces it, so anything
+    /// still standing when a trigger completes was added by the app in front.
+    private static func hasSelection() -> Bool {
+        let system = AXUIElementCreateSystemWide()
+        // A hung AX query must never stall the tap thread: an app that does
+        // not answer is treated as having no selection, which is what the
+        // deletes already assume today.
+        AXUIElementSetMessagingTimeout(system, 0.2)
```
vorssaint, line 32: the wait set on the system wide element becomes the default for every accessibility question this app asks without a cap of its own, so from the first expansion on those wait 0.2s instead of the 0.35s set at launch. the command bar already reads the selection of the app in front by asking that app directly with its own cap, can this go the same way

Sources/Vorssaint/Services/Snippets/TextSnippetService.swift
```diff
@@ -395,21 +426,28 @@ final class TextSnippetService {
                 post(event)
             }
         }
+        // One extra delete consumes an inline completion the app in front
+        // added, so the counted deletes each clear a character the user typed.
+        func postDeletes(_ count: Int) {
+            guard count > 0 else { return }
+            if hasSelection() { postKey(CGKeyCode(kVK_Delete)) }
```
vorssaint, line 433: does a selection that has nothing to do with a completion, like output left selected in a terminal while you keep typing, also get the extra delete and eat the character in front of the trigger? limiting it to a selection that sits right after what was typed would rule that out

Issue #1367, opened 2026-09-04
Proposal: Windows 11 style snapping (Snap Layouts, Snap Assist, linked resize) for Window Layout


festalife:
## Summary

Window Layout already snaps to halves, thirds, sixths and corners with shortcuts and edge dragging. People coming from Windows 11 keep reaching for the rest of that system, which is entirely mouse-driven. I have built it on a fork and would like to know whether it is welcome here, and how you would like it split, before opening pull requests.

Everything is off-switchable and lives behind the existing Window Layout guards (nothing runs while native tiling is on or without Accessibility, nothing runs inside the event tap, strings in all 13 languages, pure logic covered by the test target). Each feature maps 1:1 to one of the six switches under Windows' *Settings › System › Multitasking › Snap windows*.

## What is built (branches on the fork, stacked in this order)

1. **Snap Layouts panel at the top edge** — dragging a title bar to the top shows a small panel of layout cards (halves, thirds, ⅔+⅓, ⅓+⅔, quarters); hovering a cell drives the existing live preview, releasing applies the zone. Releasing outside the cells maximizes (covers #894). `feat/snap-layouts`, ~800 lines.
2. **Fill the space the neighbour actually leaves** — a snap group per screen remembers snapped windows; a later snap next to a resized member fills from the member's real edge (grown or shrunk), a moved member leaves the group. `feat/snap-groups`, ~1,000 lines on top.
3. **Snap Assist** — after a snap, the free cell shows the other windows (most recent first, AppKit cards); a pick places that window and advances cell by cell, then ends. Modes: ask / auto (fill with the most recent window) / off. `feat/snap-assist`, ~1,600 lines on top.
4. **Linked resize** — resizing a snapped window along the edge it shares with a neighbour resizes the neighbour too (AXObserver per app, throttled, minimum sizes respected). `feat/snap-linked-resize`, ~1,500 lines on top.
5. **Polish** — layouts panel on zoom-button hover, snap before reaching the edge, restore size when a snapped window is dragged away, group memory across minimize / screen changes with "Show group" in Dock Preview, 150 ms preview dwell, divider hint. `feat/snap-complete`, ~1,650 lines on top.

Fork: https://github.com/festalife/vorssaint-utils (branches above; `dev/integration` has everything). Verified with scripted end-to-end runs on a 1512×982 MacBook Pro (real CGEvent drags, AX frame read-back) across TextEdit, Chrome, Finder, Safari, Notes and Electron apps.

[...]

Issue #894, opened 2026-08-24
Drag window to top of screen to maximize


az0000000:
### Problem or Motivation

When using edge snapping, dragging a window to the top edge forces it into a "Top Half" layout. I would expect this action to fully maximize the window (similar to native macOS or Windows behavior), but as far as I can tell there is currently no way to customize this specific drag zone.

### Proposed Solution

It would be great if Vorssaint could add a setting allowing users to change the top-edge snap behavior from "Top Half" to "Maximize."

### Feature Area

Windows and Dock

davidwilliams921:
I came to github to make this exact feature request. A settings toggle to have this behavior would be much appreciated!

PR #1383, opened 2026-09-05
feat(window-layout): Snap Layouts panel when dragging to the top edge

festalife:
## Summary

- Dragging a title bar to the top edge now shows a floating panel of layout cards (halves, thirds, ⅔+⅓, ⅓+⅔, quarters), Windows-Snap-Layouts style.
- Hovering a cell highlights it and drives the same live preview overlay classic edge snap already draws.
- Releasing over a cell applies that layout through the existing placement path; releasing outside the cells maximizes instead (refs #894).
- A new "Snap Layouts panel at the top edge" toggle in Window Layout settings turns it off without touching classic edge snapping.
- Only ever shows once classic edge snapping is on, so it never surfaces new behavior by itself — on by default, same reasoning Windows 11 ships it out of the box.

## How it works

- `SnapLayoutsSupport.swift` (new) — pure hit-testing: which preset panel to show, which cell (if any) a point hits, and the fallback action (corner vs. maximize) for a release outside the panel.
- `SnapLayoutsPanel.swift` (new) — owns the floating `NSPanel`, forwarding hover/release into `SnapLayoutsSupport` and reusing the existing `WindowEdgeSnapTarget`/preview shape so the panel never needs its own preview path.
- `SnapLayoutsPanelView.swift` (new) — the SwiftUI cell grid rendered inside the panel.
- Hooked into `WindowGestureSupport.swift` (`WindowEdgeSnapTarget` now carries an optional `zone`, since panel cells like thirds/quarters don't all correspond to one of the eight classic edge-snap zones) and `WindowLayoutService.swift` (`resolvedEdgeSnapTarget` decides whether to keep tracking the classic target or open/host the panel).
- Guards are unchanged: native tiling on means nothing runs, Accessibility is still required, nothing executes inside the event tap itself, and the classic edge-snap zone customization gate only applies when a target actually has a zone (the panel has its own toggle instead).

## Testing

- Unit tests added to the existing harness (`Tests/MetricsTests.swift`) covering panel trigger geometry, cell hit-testing, and the maximize/corner fallback.
- Scripted end-to-end run of 16 drags on a 1512×982 display — every one landed on the expected frame.
- Manual: drag a window to the top edge, hover each cell, release on a cell and outside the panel.

## Notes

- First PR out of the proposal in #1367. Later pieces (Snap Assist, free space, linked resize) will only follow if this one lands.
- ~800 lines, mostly new strings (13 languages) and tests — the actual hit-testing and hook-in logic is small.

Refs #1367, refs #894


vorssaint:
closing this. the top edge already fills the screen since 3.3.3, so the maximize part is already in, and every layout the panel offers is one shortcut or one radial menu pick away
the panel is on by default and keeps the accessibility permission check ticking for anyone who has the window layout feature installed, even with edge snapping off, which undoes the change that stopped that check when no snap area is on
with the top area switched off on the screen map the panel still opens there and a release outside the cards still fills the screen or snaps a corner
moving between the cards flashes the full screen preview because the gaps fall back to maximize
the wording copies another system's feature by name, in code comments and in the visible settings text in four languages, and the app does not name other systems or products
a drop panel at the top edge is not the direction for this gesture, a visible choice for that zone on the screen map is where this belongs, so a smaller version of the panel would not fit either

PR #1384, opened 2026-09-05
feat(quick-toggles): add an Extra Brightness switch to the panel

festalife:
## Summary
- Extra Brightness could previously only be flipped from Settings; this adds a Quick Toggles row so it can be switched right from the panel, using the same preference (`DefaultsKey.extraBrightnessEnabled`) and the same `ExtraBrightnessService.syncWithPreferences` path Settings uses, so both stay in sync.
- The row is filtered out on Macs whose panel cannot boost brightness, mirroring how the keyboard-light row is already hidden when unsupported.

## Changes
- `Sources/Vorssaint/Core/Defaults.swift` — no functional change beyond exposing the key used by the new row.
- `Sources/Vorssaint/Services/QuickTools/QuickTogglesService.swift` — maps `QuickToggleAction.extraBrightness` to `AppFeature.extraBrightness` and wires it to the preference/service sync.
- `Sources/Vorssaint/UI/MenuPanel/PanelLayout.swift` — accounts for the new row in panel layout.
- `Sources/Vorssaint/UI/MenuPanel/QuickTogglesSection.swift` — adds the Extra Brightness row, hidden when the panel can't boost.
- `Tests/MetricsTests.swift` — updated for the new toggle.

## Testing
- Existing tests updated and passing (`./build.sh --test`, one pre-existing unrelated failure: "App Switcher icon-row hints describe default app and window shortcuts").
- Manual: installed a build with Extra Brightness enabled, opened Quick Toggles, flipped the new row, confirmed Settings mirrors the state and vice versa.

Size: ~35 lines across 5 files.

### DIFF
--- a/Sources/Vorssaint/Core/Defaults.swift
+++ b/Sources/Vorssaint/Core/Defaults.swift
@@ -281,6 +281,7 @@ enum DefaultsKey {
     static let panelToggleKeyboardLight = "panelToggleKeyboardLight"
     // Keep the existing storage key so moving the row preserves its visibility choice.
     static let panelToggleMicMute = "panelUtilityMicMute"
+    static let panelToggleExtraBrightness = "panelToggleExtraBrightness"
     static let panelToggleEmptyTrash = "panelToggleEmptyTrash"
     static let panelToggleEjectDisks = "panelToggleEjectDisks"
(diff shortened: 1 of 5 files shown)


vorssaint:
extra brightness can already be switched from the displays section of the panel in 3.3.5. i'm keeping that control in one place and closing this additional entry

Issue #796, opened 2026-08-23
15min is too low for Fan Control, let us decide (I don't want to turn it on again every 15min)


Samuel61904:
### Problem or Motivation

It's annoying to have to turn it on again

### Proposed Solution

Make it so you can pick how much time you want them to be at max, maybe like a switcher or a nontime-based button.

### Feature Area

Tools

PathGao:
The 15 minute limit is gone in the development version. Fan control now stays on for as long as the app is running instead of stopping on a timer.

Nothing has changed on your Mac yet, and this is not in a release. When it is available, you'll get a comment here with the version to install.

Samuel61904:
Thanks, I rather have Vorssaint only instead of having a separe app for it (MacsFanControl).

vorssaint:
the 15-minute limit is gone in 3.3.5. in fan control, choose manual, set the speed and press 'apply manual control'. it stays active until you return to system control, with automatic return if the mac sleeps or fan control is interrupted. closing this since the continuous mode you asked for is available

PR #1385, opened 2026-09-05
feat(fan-control): manual speed duration and named profiles

festalife:
## Summary
- A manual fan speed can now expire after a chosen duration — 15 min, 30 min, 1 h, 2 h, 4 h, or until changed (the default) — riding on the helper's existing `endsAt` watchdog rather than a new timer. When the deadline fires, the service resumes whatever curve/mode was running before the manual override instead of always dropping to system control.
- Named fan profiles (Silent / Balanced / Performance, plus custom ones saved from the current settings) let you switch configurations with one click from the panel, with management (rename/duplicate/delete/reorder) in Settings.

## Safety
- The privileged helper's protocol is unchanged; crash recovery, sleep/wake and quit paths are untouched.
- Manual-timer bookkeeping (`fanControlManualUntil`, `fanControlModeBeforeManual`, `fanControlManualPreviousCurves`) is only recorded after a confirmed successful apply, so a failed or unanswered XPC call can't leave stale state for a later override to inherit.
- The helper still enforces the deadline itself — the app is not what's making the override end on time.

## Changes
- `feat(fan-control): let a manual speed expire after a chosen duration` — adds the duration picker and remaining-time label next to the manual slider, wired to the helper's `endsAt` watchdog; expiry resumes the previously running curve/mode instead of always falling back to system control.
- `fix(fan-control): keep manual timer bookkeeping honest` — corrects a stale doc comment, freezes the curve configuration actually confirmed active into `fanControlManualPreviousCurves` at override start (instead of re-reading the live, possibly-unapplied curve editor state at expiry), and only records timer bookkeeping after a confirmed successful apply, clearing it explicitly on the no-response guard paths.
- `feat(fan-control): named fan profiles with one-click switching` — adds `FanProfile` (system/manual/curve) stored under `fanControlProfiles`/`fanControlActiveProfileID`, three non-deletable built-ins seeded on first run, a profile chip row in the panel (segmented, or a compact menu past four profiles) that applies through the same `applyConfiguration` path as a hand-set mode, a "Save as profile…" action, and a minimal Profiles list in Settings; pure mapping/matching/migration logic stays in `FanControlSupport.swift` so it keeps compiling into the privileged helper target.

## Testing
- Pure-logic unit tests added/updated for: duration mapping and expiry behavior, manual-timer bookkeeping honesty (stale-state clearing on failed apply), and profile configuration mapping both ways, match/mismatch against live controls, the three built-ins and their translated names (all 13 locales), decoding an array with one unrecognized-`kind` entry, and first-run migration.
- `./build.sh --dev`, `./build.sh --test`, and `--selftest` all pass (the one failing test, "App Switcher icon-row hints describe default app and window shortcuts", reproduces identically on `origin/main` and is unrelated to fan control).
- Manual: picked 60% manual speed for 15 min, watched the countdown in the panel, confirmed it reverted to the prior curve/mode at expiry; tapped Silent and Performance chips and confirmed the helper applied each correctly.

## Notes
Refs #796 (the request to let the manual duration be chosen). This PR is the fan-control-related part of the fork described in #1367 — the manual-timer and profiles work only, independent of the window-management work in that discussion.

### DIFF
--- a/Sources/FanControlHelper/main.swift
+++ b/Sources/FanControlHelper/main.swift
@@ -143,9 +143,11 @@ private final class FanControlController {
                                                     snapshot: self.currentSnapshot())))
                 return
             }
-            let response = self.applyConfiguration(session: id,
-                                                   configuration: configuration,
-                                                   duration: nil)
+            let response = self.applyConfiguration(
+                session: id,
+                configuration: configuration,
+                duration: FanControlPolicy.sessionDuration(for: configuration)
+            )
             reply(FanControlIPC.encode(response))
         }
     }
--- a/Sources/Vorssaint/Core/Defaults.swift
+++ b/Sources/Vorssaint/Core/Defaults.swift
@@ -355,6 +355,15 @@ enum DefaultsKey {
     // Apply step, so it can no longer be trusted to describe what the
     // hardware was actually running by the time the override expires.
     static let fanControlManualPreviousCurves = "fanControlManualPreviousCurves"
+    // Named fan profiles (an encoded [FanProfile], see FanControlSupport.swift),
+    // seeded once with the three built-ins by migrateFanControlProfiles.
+    static let fanControlProfiles = "fanControlProfiles"
+    // Which stored profile the panel last applied ("" means none). The panel
+    // recomputes its highlighted chip from matching the live controls against
+    // fanControlProfiles rather than trusting this blindly (hand-edited
+    // controls should deselect immediately), so this key is only a
+    // convenience cache of the last explicit selection.
+    static let fanControlActiveProfileID = "fanControlActiveProfileID"
     // Previous panel visibility key, read once by the migration below.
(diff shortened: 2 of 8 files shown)


vorssaint:
closing this one, reasons follow
the built in profiles repeat the mode picker under vaguer names, balanced is plain system control and silent is a fixed 25% that spins the fans faster than the system does at idle and slower than it does under load
a second row of chips, a settings list and presets stored on every install is a lot of permanent surface for what the mode picker, the slider and a curve already do
while a timed manual run counts down the caption still says control stays active until you return to system, and an expiry the user picked ends with the message that fan control was interrupted
when the timer hands control back to a curve the mode picker stays on manual with the slider still showing
the last picked profile id is written but never read
if you still want a timed manual run, a small version that only adds the duration choice, returns to system control when it ends and fixes those two messages could be looked at on its own

Issue #514, opened 2026-08-06
Deeplinks


Macko-pp:
### Problem or Motivation

I would like to implement some of Vorssaint's features into an Apple Shortcut. Since that is not currently possible, I am reminded of how Raycast allows any of its commands to be accessed via deep links. I would love to see this same functionality implemented in Vorssaint.

### Proposed Solution

The deeplinks could be copied to your clipboard via the settings, menu bar, or command bar. When opened, these links would execute the command. This allows any application that can open links, such as Apple Shortcuts or even a website, to trigger any command.

### Feature Area

Other

### Mockups / References

This is how it works in raycast. I would love to see something similar to this in Vorssaint.

https://github.com/user-attachments/assets/726083df-a213-412f-bde8-be82a7616b76

PathGao:
Folding **#418** and **#580** in here. All three ask for the same thing at bottom — a way to drive Vorssaint from outside the app — and this is the thread to keep. What differs is the entry point, and each of the two adds a requirement this body does not cover. Recording both so closing them does not shrink the request.

### From #418 (@Jump-333) — AppleScript / terminal

- **Parameterised write.** The use case is automating one app's volume from a physical slider: set a *specific* app's volume to a *continuous value* that a fader keeps pushing. As written, this issue is Raycast-style deeplinks that "execute the command" — there is no parameter surface and no way to feed a moving value.
- **A terminal command** as an entry point, named alongside AppleScript. This issue is URL-only.

### From #580 (@vkplayz0) — Shortcuts actions / CLI

- **Reading, not just triggering.** Asks to read the volume of any app, and to read the temperatures, power usage, fan speeds and other stats Vorssaint already tracks. A URL deeplink is one-way by construction — a link cannot hand a value back. Satisfying this needs Shortcuts actions or a CLI, which is the delivery shape named in that issue's title.
- Also asks to **change** any app's volume (the example given is setting Safari to 0%), which lines up with #418's parameterised write.

### What that makes this issue

Three entry points — URL scheme, AppleScript, CLI or Shortcuts actions — over one command surface, with parameters going in and results coming back. Deeplinks as described in the body alone would leave both folded issues unanswered, so the read path and the parameter path should be part of the design from the start even if only the URL scheme ships first.

close #418 as duplicate
close #580 as duplicate

tuttopassastudios:
[...]

PR #1393, opened 2026-09-05
Implement deep link handling

brandom:
## Summary

Adds support for the `vorssaint://` URL scheme.

`vorssaint://run/<id>?v=<n>` runs the Command Bar row with that stable ID through the same execution path as the row's own shortcut. This gives Shortcuts, scripts, and other local tools a simple way to invoke any installed Command Bar tool without per-app integrations.

The implementation adds a small amount of parsing and dispatch code, with no new runtime dependencies. Unknown or unavailable rows beep and log the reason under the `deeplink` category. Rows that need setup open their Settings page; rows that require confirmation or a number beep instead.

The initial implementation was reworked by @tuttopassastudios to bring it onto current main and simplify the implementation substantially.

The parser and supported links are covered by tests and documented in `docs/DEEP_LINKS.md`. Settings pages work through deep links, although `pagesCoveredByActions` still remains to be resolved through a follow-up.

## Verification

Tested on:

- MacBook Pro (M1 Pro), macOS 26.6.2 (25G83): deep links were tested from Terminal and from Raycast with the Command Bar on and off.
- Tested on a MacBook Pro (M1 Max), macOS 26.6.2 (25G83): links from Terminal and from a launcher shell command, with the Command Bar on and off.

```sh
./build.sh                    # ✅
./build/Vorssaint --selftest  # ✅
./build.sh --test             # ✅
```

## Anything else

Refs #514

### DIFF
--- a/Sources/Vorssaint/App/AppDelegate.swift
+++ b/Sources/Vorssaint/App/AppDelegate.swift
@@ -309,6 +309,30 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         return true
     }
 
+    // MARK: - Deep links
+
+    /// Handles `vorssaint://` URLs from external callers.
+    func application(_ application: NSApplication, open urls: [URL]) {
+        // Defer handling until startup has finished arranging the app.
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            for url in urls { self.handleDeepLink(url) }
+        }
+    }
+
+    private func handleDeepLink(_ url: URL) {
+        guard let request = DeepLinkSupport.parse(url) else {
+            Self.deepLinkLog.error("rejected deep link \(url.absoluteString, privacy: .private)")
+            NSSound.beep()
+            return
+        }
+        CommandBarService.shared.runExternalRow(withStableKey: request.stableKey,
+                                                argument: request.argument)
+    }
+
+    private static let deepLinkLog = Logger(subsystem: Bundle.main.bundleIdentifier ?? "vorssaint",
+                                            category: "deeplink")
+
     /// Whether the menu bar icon is actually visible on a screen, rather than
     /// present in the status bar but clipped or dropped by a crowded/notched menu
     /// bar (in which case the button still has a window, just not an on-screen one).
--- a/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
+++ b/Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
@@ -472,10 +472,34 @@ final class CommandBarService: ObservableObject {
             NSSound.beep()
             return
         }
-        // A row that would confirm, ask for input, or keep the field visible
-        // needs a real presentation just as it does when chosen from the bar.
-        // Emptying the Trash on one keypress with nothing asked is not a
-        // shortcut, it is an accident with a name.
+        dispatch(entry, argument: nil, key: key)
+    }
+
+    /// Runs a row requested by an external `vorssaint://` deep link.
+    /// Uses the same execution path as row shortcuts, including any required
+    /// confirmation or setup steps. Unlike row shortcuts, the Mac's own
+    /// Settings panes are refused here rather than opened.
+    func runExternalRow(withStableKey key: String, argument: Int?) {
+        guard let entry = freshFullEntry(forStableKey: key) else {
+            NSSound.beep()
+            return
+        }
+        dispatch(entry, argument: argument, key: key)
+    }
+
+    /// A row that would confirm, ask for input, or keep the field visible
+    /// needs a real presentation just as it does when chosen from the bar.
+    /// Emptying the Trash on one keypress with nothing asked is not a
+    /// shortcut, it is an accident with a name.
+    private func dispatch(_ entry: CommandBarEntry, argument: Int?, key: String) {
+        // An argument answers only the numeric prompt. A row that also
+        // confirms, or whose switch is off, still gets the bar, whose run(_)
+        // asks in the bar's own order.
+        if let range = entry.numericRange, let argument,
+           entry.confirmationPrompt == nil, entry.trouble == nil {
+            finish(entry, value: min(max(argument, range.lowerBound), range.upperBound))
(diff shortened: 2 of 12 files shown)


tuttopassastudios:
Hi @brandom, thanks for starting this. We want deep links too, for launcher integration, so I rebuilt your branch on current main and trimmed it toward the size that moves quickly here:

- 195 lines, down from ~750. Dropped the three fix write-ups, the full ID table and its sync test, and the `settings.<page>` fallback (a possible follow-up).
- Tests moved to a `deep-links` suite to fit main's new test runner.
- With the Command Bar off, a link that would ask now opens Settings (missing setup) or beeps, instead of doing nothing.
- A link that runs nothing logs why under the `deeplink` category.
- The keep awake `?v=` bug exists on main without deep links, so it is in its own small PR.

Tested on a MacBook Pro (M1 Max), macOS 26.6.2 (25G83): links from Terminal and from a launcher shell command, with the Command Bar on and off.

Your authorship is kept on the commit: https://github.com/tuttopassastudios/vorssaint-utils/tree/deep-links-rework

If it works for you, you could force-push that onto this branch, or I can open a PR against your fork. Happy to adjust either way.

brandom:
@tuttopassastudios, thanks for putting this together and for taking the time to rework it. 

I’ve rebased the changes and force-pushed them onto this PR. This does everything I need and appears much more streamlined. I really appreciate the effort!

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -599,10 +599,51 @@ final class CommandBarService: ObservableObject {
...
+    func runExternalRow(withStableKey key: String, argument: Int?) {
+        guard let entry = freshFullEntry(forStableKey: key) else {
+            DeepLinkSupport.log.error("no row answers deep link \(key, privacy: .private)")
+            NSSound.beep()
+            return
+        }
+        // With the bar off there is nowhere to ask, and show(promptingFor:)
+        // would return without a sound. Missing setup still has its Settings
+        // page; anything else that needs the bar beeps.
+        if !AppFeature.commandBar.isAvailable {
```
vorssaint, line 617: with the command bar uninstalled in the features hub, a link still rebuilds its rows and runs them, and the rows of an uninstalled feature are supposed to be gone. can it beep and run nothing there, the way the row shortcuts already do

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -599,10 +599,51 @@ final class CommandBarService: ObservableObject {
...
+    /// needs a real presentation just as it does when chosen from the bar.
+    /// Emptying the Trash on one keypress with nothing asked is not a
+    /// shortcut, it is an accident with a name.
+    private func dispatch(_ entry: CommandBarEntry, argument: Int?, key: String) {
+        // An argument answers only the numeric prompt. A row that also
+        // confirms, or whose switch is off, still gets the bar, whose run(_)
+        // asks in the bar's own order.
+        if let range = entry.numericRange, let argument,
+           entry.confirmationPrompt == nil, entry.trouble == nil {
+            finish(entry, value: min(max(argument, range.lowerBound), range.upperBound))
```
vorssaint, line 644: if the bar is open with something typed when a link with a number arrives, that text gets learned as the way to reach this row. can the bar close before the row runs, like it does for a link without a number

Sources/Vorssaint/App/AppDelegate.swift
```diff
@@ -316,6 +316,27 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         return true
     }
 
+    // MARK: - Deep links
+
+    /// Handles `vorssaint://` URLs from external callers.
+    func application(_ application: NSApplication, open urls: [URL]) {
+        // Defer handling until startup has finished arranging the app.
+        DispatchQueue.main.async { [weak self] in
```
vorssaint, line 324: when the app is not running and the link is what launches it, does the row still run once the app is up

brandom: Yes, it does.

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
brandom: I would prefer to be able to run Vorssaint utils without having to enable the Command Bar. If I’m understanding your suggestion correctly, the Command Bar would need to be enabled for deep links to work at all.

I understand the argument that code belonging to a disabled feature perhaps shouldn’t be used at all. My view is that deep links should be an always-on application-level feature. That would allow users to reach tools through deep links regardless of whether the Command Bar itself is enabled.

Please let me know if I’ve misunderstood the ask, or if you would prefer an implementation where deep links do not rely on Command Bar functionality.

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
brandom: I think I have resolved this with c14334b039a33b48aa5d66cafb3837896f18d11e. I can squash if the fix works for you.

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -659,15 +659,60 @@ final class CommandBarService: ObservableObject {
...
+    func runExternalRow(withStableKey key: String, argument: Int?) {
+        guard let entry = freshFullEntry(forStableKey: key) else {
+            DeepLinkSupport.log.error("no row answers deep link \(key, privacy: .private)")
+            NSSound.beep()
+            return
+        }
+        // With the bar off there is nowhere to ask, and show(promptingFor:)
+        // would return without a sound. Missing setup still has its Settings
+        // page; anything else that needs the bar beeps.
+        if !AppFeature.commandBar.isAvailable {
```
vorssaint, line 677: the request is about uninstalling the command bar in the features hub, not leaving its window closed or its shortcut off. links can work in those last two cases, but they must stop when the command bar is uninstalled

Issue #1336, opened 2026-09-03
Offer to take over a macOS shortcut while the feature that recorded it runs


owendaw:
### Problem or Motivation

The shortcut recorder refuses any combination an enabled macOS shortcut already uses ("conflicts with macOS"). The only way past it is to untick the key in System Settings. People arriving from Magnet, CleanShot or Shottr expect ⌘⇧4 or ⌃⇧← to just work, and #1091 shows the refusal also fires for entries System Settings has no checkbox for. A collaborator wrote there, twice, that the missing "take it anyway" option is the real bug.

Unticking the key in System Settings is also the wrong fix for a second reason: it is permanent. Quit Vorssaint and ⌘⇧4 does nothing. Uninstall it and you have to remember, months later, what you changed and where.

Vorssaint already solves this shape for one case. #882 lets the App Switcher take ⌘Tab over by switching the Dock's symbolic hotkey off in the WindowServer at runtime, restoring it when the switcher is off, the tap is lost, the app quits, or on the next launch after a crash. Nothing on disk changes and System Settings still shows the key ticked.

### Proposed Solution

Generalise that mechanism so any recorded shortcut can use it, opt in, per shortcut, at the moment of conflict. I intend to build this and am asking whether the direction is welcome before writing code.

- The recorder, instead of refusing, shows the key, says macOS uses it, and offers one action: **Take over while Vorssaint runs**. A caption says macOS gets the key back whenever Vorssaint is not running or the feature is off.
- The take-over service from #882 becomes shared: it works on live symbolic hotkey IDs read from the WindowServer instead of a four-case enum, and any feature that registers a global hotkey claims its key after a successful registration and releases it on unregister. Suppression only of currently enabled, exactly matching entries, the write-ahead marker, crash repair, the terminate hook and the wake reconcile stay exactly as #882 built them. One addition: before restoring a key, check whether the user has since switched it off in System Settings, and if so leave it off.
- The conflict check moves from the `com.apple.symbolichotkeys` plist, which only lists customised entries, to the live table. Today a never-customised ⌘⇧4 passes the check and then macOS wins at runtime.
- Zero new settings rows, three new strings, two preference keys (the take-over set, exported with settings; the suppressed-ID marker, machine state, not exported). No new permission, no dependency, no change to how shortcuts are matched.

Two small fixes fall out of reading the current code and stand on their own:

1. The conflict check reading the plist rather than live state (above).
[...]

Issue #1091, opened 2026-08-28
Window layout: unable to set shortcut "already in use by macOS"


brericha:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

I am attempting to set the previous and next display shortcuts to ⌃⇧← and ⌃⇧→ but after trying to record either of those a message appears that says "This shortcut is already used by macOS." It does not seem to cause anything to happen via macOS itself. Additionally, I use these shortcut combinations with Magnet for the same effect. Magnet is not running at the same time that I am attempting to set these shortcuts in Vorssaint. Since Vorssaint has the capability to replace Magnet (and a few other utilities I use) I was hoping to be able to set up the same shortcuts that I have been using already.

### Steps to Reproduce

1. Open Vorssaint
2. Open "Window layout" settings
3. Attempt to record either ⌃⇧← or ⌃⇧→ as a keyboard shortcut.

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

English

### Mac and displays

M1 Mac mini

### Installation Method

Homebrew

PathGao:
macOS already owns ⌃⇧← and ⌃⇧→. They're the Mission Control "move left/right a space" shortcuts — the Shift version moves the current window along with you. If you don't have another desktop to move to, nothing visible happens, but the system still holds the keys, and that's what Vorssaint is refusing to record over. Magnet doesn't check, it just takes them, which is why it worked there.

You can take them back: System Settings → Keyboard → Keyboard Shortcuts → Mission Control, and switch off "Move left a space" and "Move right a space". Record the shortcut in Vorssaint again after that and it should go through. If you'd rather keep them, any combination macOS isn't already using will record fine.

<details>
<summary>The technical side — you can skip this</summary>

The recorder refuses a combination only when an **enabled** entry in the system's own shortcut list matches the key and modifiers exactly, so the message is accurate here rather than an over-broad blocklist. The Mission Control space-switching entries ship as pairs, plain and Shift-variant, and the Shift halves are enabled by default even though System Settings shows one row per pair.

Not verified: whether turning off the visible row disables both halves of the pair on macOS 26, and whether the pairing holds on a Mac with a single desktop. I read the matching logic, I did not test on hardware.

[...]

PR #1417, opened 2026-09-06
feat(shortcuts): offer to take over a macOS shortcut while its feature runs

owendaw:
## Summary

Proposed in #1336: when a feature's recorder is handed a combination macOS already answers (⌘⇧3, ⌘⇧4, ⌘⇧5, ⌘Space, ...), offer to take the key over for as long as Vorssaint runs instead of refusing it. The App Switcher has done this for ⌘Tab and ⌘` since #882; the shared service that generalises it landed in #1379. This is the user-facing half.

- Four recorder sites (`ShortcutRecorderButton`, `ShortcutsSettings`, `WindowLayoutSettings`, `CutPasteSettings`) show a small inline offer under the row when the recorded key conflicts with macOS: the macOS message, a "Take over while Vorssaint runs" button, a "Don't take over" button, and one caption: macOS gets the key back whenever Vorssaint is not running or the feature is off. Accepting saves the shortcut and records the opt-in for that row; declining leaves the row as it was and shows the existing conflict message. The refusal path for everyone who does not opt in is unchanged. The display brightness rows added in #1488 go through the same shared row, so they get the offer too; both brightness pairs now register a key the user chose to take over and claim it, where before they refused it as macOS's and showed the red caption.
- The opt-in is per row, stored as a set of storage keys under `systemShortcutTakeOverKeys` (exported with settings; it is a preference, not machine state). One pure rule, `SystemShortcutTakeoverSupport.recorderDecision`, decides between saving silently and asking: a key macOS does not want saves and clears any stale opt-in; a row re-recording the exact key it already opted into saves silently and keeps the opt-in — a key Vorssaint is itself holding still counts as macOS's key, or the check would read the id Vorssaint just switched off as free; anything else that macOS answers asks again, including a row with the opt-in but no recorded key. Unit-tested.
- Every place that registers a global hotkey now claims its shortcut through `SystemShortcutTakeover.claim` and releases it where it unregisters: keep awake (`HotkeyManager`), every quick-tool hotkey (per role: screenshots, captures, recent captures, quick launcher, paste plain, scratchpad, snippets, mic mute, camera preview, the two keyboard-backlight roles, the command bar and its rows, radial-menu profiles), sound output switcher, window layouts (each action plus the directional shortcut), shelf, clipboard history, and Finder rename, whose event tap only sees the key once the symbolic hotkey is off. A claim on a row without the opt-in is a no-op, so a feature that never asked for a take-over gets nothing switched off. Rows nothing claims — the App Switcher's two (they have their own take-over toggle) and the legacy radial-menu row — do not show the offer and refuse a macOS key as before, through the per-role check from 3.3.3, so the switcher's rows still accept their own native keys. The held-aware rule takes the role for the same reason.
- The service maps a claimed shortcut to live hotkey ids on the spot and again on wake, so a key the user remaps in System Settings while Vorssaint runs is picked up on the next reconcile. Quit gives every taken-over key back from one place (`restoreAll()` in `applicationWillTerminate`) rather than relying on each feature's suspend path. Launch recovery runs before any feature registers, keeps #1379's rule (it takes nothing over itself), and records the ids it kept as the switcher's so the first claim after it does not hand them back.
- Strings: four new keys in `Localization.swift` and the eleven `Strings+*.swift` files, translated in each language to match the register of the neighbouring shortcut strings.
- While the offer is up, the recorder field shows the key that was just pressed rather than snapping back to the saved one, so the offer reads as a question about that key; declining puts the saved key back. Reset, and the window-layout clear buttons, release the row's opt-in along with the shortcut, so no entry outlives the key it was given for.

What a user notices: a row that used to answer "conflicts with macOS" now also offers the take-over. With the take-over accepted on, say, Screenshot's ⌘⇧3, the live table shows id 28 off while Vorssaint runs and the feature is on, and on again after quit, feature off, or the row moving to another key. Nothing changes in System Settings and nothing needs to be remembered.

Known and left alone, three things. The command bar's per-app shortcut center (#1421) keeps its own recorder path and refuses a macOS key as before; its rows already claim through the shared service, so the offer can be extended there in a follow-up once the shape here is agreed. An opt-in entry stays in `systemShortcutTakeOverKeys` when a feature is turned off (the key itself goes back to macOS at once); it is inert unless that row's current shortcut is a macOS key the user already agreed to, and a settings restore re-applying a take-over the user chose is the intended reading. And after a crash — not a clean quit — a key taken over for a recorder row is handed back at launch and taken again a moment later when its feature registers, because launch recovery only repairs the previous process's ownership and holds nothing a live handler has not claimed yet (#1379's rule). One off/on flip on the affected key, once, no marker left behind.

Possible split if preferred: the plumbing (preference, rules, claim/release hooks, launch ordering) is behaviour-neutral without the offer UI and could go first; the seam runs through the fix commits too, so I would re-cut the branch rather than cherry-pick.

## Verification

MacBook Pro 14" 2023 (Mac14,5, M2 Max), macOS 26.6.2 (25G83), two Dell U2725QE 4K displays plus the built-in display, nine Spaces, English. Build, tests and self-test at the final commit. The on-device steps below were recorded on my developer build before the rebase onto #1379 (the rebase changed launch behaviour only, re-measured at the final commit in the `kill -9` bullet) and on the daily build since.

- `./build.sh` — no warnings
- `./build.sh --test` — TESTS OK (10646 checks) and PREFERENCE CLEANUP TESTS OK
- Developer build, three rows on macOS keys: Screenshot ⌘⇧4, Full screen ⌘⇧3, Screen recording ⌘⇧5. Each recorder showed the offer; accepting saved the row, recorded the opt-in under the row's storage key, and switched the id off: live table read 28, 30 and 184 disabled, `systemShortcutsSuppressed` `(28, 30, 184)`. Each key ran Vorssaint's tool, not macOS's.
- Feature off → key back to macOS at once; feature on → taken over again.
- `kill -9` with two rows taken over (ids 30 and 184): marker survived, keys stayed off. Relaunch, polling the live table every ~130 ms for 8 s: one reading, at 0.31 s after launch, saw a key enabled — launch recovery handing it back before its feature registered and took it again — and the end state matched the state before the kill, marker `(30, 184)`.
- Sleep and wake: same process, keys still off after the reconcile, marker unchanged.
- Decline: recording ⌘⇧3 on the Screenshot row and choosing not to take over left the row on its key with the usual "already used by macOS" line. Re-recording the row's own taken-over key saved with no offer. Recording ⌃⌥⌘4 saved silently, cleared the row's opt-in, and id 30 read enabled again while the other two rows kept theirs.
- Reset on the Full screen row: key back to macOS at once. (Before the fix commits the opt-in entry stayed behind; that is what the Reset change fixes.)
- Clean quit: all three ids enabled again, marker cleared; the opt-ins stay in preferences and take the keys over again on the next launch.
- On the rebuilt copy with the last commit: the field shows the pressed key while the offer is up, and "Don't take over" puts the saved key back.
- Daily build on my own Mac since 2026-09-05, rebuilt from this branch as it changed.

Not tested: macOS 14 and 15; the eleven non-English languages beyond confirming the strings compile.

## Anything else

Refs #1336
Refs #1091

Builds on #1379 and keeps its launch rule; rebased onto `main` after 3.3.5, where the only overlap was the recorder's new per-role check, adopted here. Ten commits; the last five are the fixes from living with it, from review, and from rebasing over #1488, which also showed that the brightness rows refused a taken-over key before they could claim it. On #1336 I said I would keep this a draft for a week; the guide asks for a few days, and two builds have been my daily app since 2026-09-04, so I am marking it ready now — if you would rather it wait, say so and I will put it back to draft.


owendaw:
Rebased onto today's main. #1488's display brightness shortcuts register through QuickToolHotkey, which this branch asks for a storage key, so the tenth commit passes it. Doing that showed both brightness pairs refused a taken-over key before they could claim it (the red "shortcut unavailable" caption); the same commit lets a row's opt-in through that guard. Nothing else changed.

Sources/Vorssaint/Services/SystemShortcutTakeover.swift
```diff
@@ -37,14 +135,39 @@ enum SystemShortcutTakeover {
...
         defer { lock.unlock() }
+        // Wake can leave the live table changed under a claim that is still
+        // ours; the switcher's own three seconds keep the two reconciles from
+        // racing the WindowServer right after wake. Launch is the only caller,
+        // so the observer is installed once and kept for the life of the app.
+        if wakeObserver == nil {
+            wakeObserver = NSWorkspace.shared.notificationCenter.addObserver(
+                forName: NSWorkspace.didWakeNotification, object: nil, queue: .main) { _ in
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 3) { reconcile() }
+                }
```
vorssaint: this wake observer is installed at every launch and kept for the life of the app, even on a mac where no shortcut was ever taken over. can it exist only while some feature is actually holding a key, and go away when the last one is given back, the way the switcher handles its own

Sources/Vorssaint/UI/SystemShortcutTakeOverOffer.swift
```diff
@@ -0,0 +1,30 @@
...
+        VStack(alignment: .leading, spacing: 6) {
+            Text(String(format: l10n.s.shortcutTakeOverOffer, shortcut.displayString))
+                .font(.caption)
+            // Taking a key from macOS and keeping things as they are sit at
+            // opposite ends, so a slip cannot turn one into the other.
+            HStack {
+                Button(l10n.s.shortcutTakeOverAction, action: onAccept)
+                Spacer()
+                Button(l10n.s.shortcutTakeOverDismiss, action: onDismiss)
+            }
```
vorssaint, line 25: the two buttons sit at opposite ends of the row. on the mac the confirming action goes on the right with the other choice right beside it, so can they be grouped that way under the shortcut field

Sources/Vorssaint/Services/SystemShortcutTakeover.swift
owendaw: Done in e9a4b9e. The observer now follows the held set: installed when it becomes non-empty, removed when the last key is given back, as the switcher does with its own. The held set changes in two places (the transition apply and launch recovery), so quit and feature-off restores take the observer with them, and a Mac where nothing was ever taken over never installs it.

Sources/Vorssaint/UI/SystemShortcutTakeOverOffer.swift
owendaw: Done in 86dfd03. The pair now sits at the trailing end, "Take over while Vorssaint runs" on the right with "Don’t take over" beside it, under the shortcut field.

Issue #1257, opened 2026-09-01
Dock Preview: Add option to order dock previews by window creation time instead of last use time


mtnbnz:
### Problem or Motivation

Currently, dock preview windows are ordered by last use time. It seems, to me at least, that it would be more natural to have the previews ordered by window creation time or at least have an option to switch this behavior. Both Windows and all the Linux desktop environments I have tried with "docks" order windows by creation time as far as I remember.

### Proposed Solution

Add an option to switch the behavior of dock previews to be ordered by window creation time.

### Feature Area

Windows and Dock

Borisserz:
Proposed in https://github.com/vorssaint/vorssaint-utils/pull/1425

Optional toggle (default off) to order Dock Preview windows by ascending `windowID` as a creation-time proxy, instead of last-use ranking.

PR #1425, opened 2026-09-06
feat(dock-preview): optional creation-time window order

Borisserz:
## Summary
- Adds `dockPreviewOrderByCreation` (default `false`) so Dock Preview can keep last-use order or sort windows by ascending `windowID` (creation-time proxy).
- Pure helper `DockPreviewSupport.orderedWindows(_:order:)` with unit tests; settings toggle + strings in all locales.

## Test plan
- [x] `./build.sh --test` (10493 checks)
- [x] `./build.sh --dev` compiles
- [ ] Enable Dock Preview → Settings → toggle “Order windows by creation time” → hover a multi-window app and confirm older windows appear first
- [ ] Leave toggle off and confirm last-use order is unchanged

Refs #1257

### DIFF
--- a/Sources/Vorssaint/Services/DockPreview/DockPreviewService.swift
+++ b/Sources/Vorssaint/Services/DockPreview/DockPreviewService.swift
@@ -667,8 +667,11 @@ final class DockPreviewService: ObservableObject {
     }
 
     private static func previewableWindows(for pid: pid_t) -> [SwitcherItem] {
-        WindowEnumerator.listWindows(for: pid, maximumCount: 12)
+        let windows = WindowEnumerator.listWindows(for: pid, maximumCount: 12)
             .filter { $0.windowID != nil }
+        let order = DockPreviewWindowOrder.fromDefaults(
+            orderByCreation: UserDefaults.standard.bool(forKey: DefaultsKey.dockPreviewOrderByCreation))
+        return DockPreviewSupport.orderedWindows(windows, order: order)
     }
 
     private func beginHoverIfStillValid(token: UUID, initialHit: DockHit) {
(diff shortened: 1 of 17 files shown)


Sources/Vorssaint/Services/DockPreview/DockPreviewService.swift
```diff
@@ -667,8 +667,11 @@ final class DockPreviewService: ObservableObject {
     }
 
     private static func previewableWindows(for pid: pid_t) -> [SwitcherItem] {
-        WindowEnumerator.listWindows(for: pid, maximumCount: 12)
+        let windows = WindowEnumerator.listWindows(for: pid, maximumCount: 12)
             .filter { $0.windowID != nil }
+        let order = DockPreviewWindowOrder.fromDefaults(
+            orderByCreation: UserDefaults.standard.bool(forKey: DefaultsKey.dockPreviewOrderByCreation))
+        return DockPreviewSupport.orderedWindows(windows, order: order)
```
vorssaint, line 709: the pinned panel refreshes its own list a moment after pinning, and that list still comes back in last-use order, so with this option on the cards jump back to the old order. can the pinned panel apply the same ordering

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -1433,6 +1434,9 @@ struct SwitcherSettings: View {
                             Toggle(l10n.s.dockPreviewQuitAppOnClose,
                                    isOn: $dockPreviewQuitAppOnClose)
                             SettingsCaptionText(l10n.s.dockPreviewQuitAppOnCloseCaption)
+                            Toggle(l10n.s.dockPreviewOrderByCreation,
```
vorssaint: this section is already at five controls and this one is a niche preference. can it go inside a more options group, the way the brightness section does, instead of a sixth toggle in the open

Issue #712, opened 2026-08-20
Minimum fan speed / always-on low fan mode


Bloxx-I:
### Problem or Motivation

I really dislike when my Mac gets noticeably warm to the touch during light workloads, even though the internal temperatures are still perfectly safe. For example, it may sit around 60°C (cpu temp maybe 36 case) while the fans remain completely off.

This is especially noticeable when using a case, which slightly reduces passive heat dissipation. I don’t need aggressive cooling; I’d simply prefer the fans to run very slowly and quietly to keep the chassis closer to room temperature.

### Proposed Solution

It would be great if Vorssain could provide a minimum fan speed / always-on fan option.

For example, users could set a minimum fan speed such as 1,000–1,500 RPM. The normal fan curve would still take over as temperatures increase, but the fans would never drop below the configured minimum.

Ideally, this could also be configurable per preset/profile, so a “Cool & Quiet” profile could keep the fans barely spinning during light workloads while other profiles retain the normal zero-RPM behavior.

### Feature Area

Other

PathGao:
The development version has fan curves, which covers this. Give the lowest temperature point a speed above zero and the fans never drop below it, even when the Mac is cool.

Nothing has changed on your Mac yet, and this is not in a release. When it is available, you'll get a comment here with the version to install.

PR #1431, opened 2026-09-06
fix(menu bar): stop reserving room for 20000 RPM in the fan readout

ozp3:
## Summary

With the fans idle the menu bar fan readout shows `0/0` in the middle of a slot wide enough for `2323/2501`, and the strip looks stuck at its widest. It is not a measurement bug: the fan block reserved room for `20000` a fan — the policy's sane ceiling, not a speed any fan reaches — and drew the value centred in it. With two fans that is eleven characters, about 85 points, so `0/0` sat with 36 points of nothing either side.

The block now takes the width of what it shows, with the `RPM` label as its floor.

## Measured

On the reporting Mac, combined status item, fans at 0/0, read through the accessibility frame and a screen capture of the strip:

| | before | after |
|---|---|---|
| status item width | 268 pt | 221 pt |
| gap between `PWR` and `RPM` | 36 pt | 14 pt |
| trailing space after `0/0` | 36 pt | 21 pt (label floor + button inset) |

## The trade-off, stated

The reservation was buying stability: the block changes width when the digit count does, so what sits to its right moves by a character as a fan crosses 999 or 9999. Every other metric here reserves for its worst case for that reason — but their worst cases are real (`100%`, `999°`); 20000 RPM is not. If the strip is preferred steady, the alternative is a one-line change: reserve four digits a fan (`9999/9999`), which no MacBook fan exceeds and which is still two characters narrower than today. I would take either; this one is what the reporter asked for.

## Testing

`./build.sh`, `./build/Vorssaint --selftest`, `./build.sh --test` pass with no warnings. Verified on a MacBook Pro (Mac17,9), macOS 27.0 (26A5425a), lid closed, one external display over HDMI, app language Turkish, local build, combined status item in the compact block style. Not checked: the separate-items layout, the readable block style, and what the strip does at the 999→1000 boundary while the fans spin up — that one is visible only with the fans under load.

No unit test: the change removes a placeholder string; the width it produced is only observable on screen and is recorded above.

Refs #712, #725 (the other fan readout threads) — this covers only the idle width.

### DIFF
--- a/Sources/Vorssaint/App/MenuBarRenderer.swift
+++ b/Sources/Vorssaint/App/MenuBarRenderer.swift
@@ -699,11 +699,15 @@ enum MenuBarRenderer {
                 }
             case .fanSpeed:
                 if let value = FanControlPolicy.menuBarValue(for: snapshot.fanSpeeds) {
-                    let minimumValue = Array(repeating: "20000", count: snapshot.fanSpeeds.count)
-                        .joined(separator: "/")
+                    // No room held for a value the hardware cannot reach: the
+                    // policy's sane ceiling is 20000 a fan, which reserved
+                    // eleven characters and drew 0/0 in all of them. The block
+                    // takes the width of what it shows, and the label keeps a
+                    // floor under it, so the strip settles at RPM rather than
+                    // at a number no fan reports.
                     groups.append([.metricBlock(label: "RPM",
                                                 value: value,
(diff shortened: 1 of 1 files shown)


ozp3:
Withdrawn: the reporter would rather keep the readout as it is; the idle-width finding stays here for whoever picks up #712/#725.

Issue #1422, opened 2026-09-06
Clean URL should clean YouTube tracking too


Polovinkin:
### Problem or Motivation

I've noticed that when I copy links from YT via "Share" button, Vorssaint does nothing with `?si=` piece there when I paste it. It is a tracking parameter, obviously something which this feature should clean - and it does not. I'm not even sure it's the intended behavior or a bug.

### Proposed Solution

Clean this `?si=` thingy from links as well, or add a checkbox in the settings to turn YT links cleaning on.

### Feature Area

Mouse and keyboard

Borisserz:
Opened a fix PR: https://github.com/vorssaint/vorssaint-utils/pull/1432

Clean URL already strips YouTube `si` in `URLCleaning.clean`; Share often adds HTML/RTF so automatic rewrite was skipped. The PR allows rewrite when the pasteboard string is a lone http(s) URL even with those rich companions.

PR #1432, opened 2026-09-06
fix(clean-url): rewrite share links with rich pasteboard types

Borisserz:
## Summary
- Allow Clean URL auto-rewrite when the pasteboard string is a lone http(s) URL, even if Share added HTML/RTF companions.
- Keep refusing rewrite for mixed/prose clipboard content or an empty type list.
- Extract `URLCleaning.shouldAutomaticallyRewrite` for unit coverage without NSPasteboard.

## Test plan
- [ ] `./build.sh --test`
- [ ] Enable Clean URL, share a YouTube link (Share sheet → Copy), confirm `?si=` is stripped from the clipboard
- [ ] Copy mixed text that contains a URL plus rich types; confirm the clipboard is left alone

Refs #1422

### DIFF
--- a/Sources/Vorssaint/Core/URLCleaning.swift
+++ b/Sources/Vorssaint/Core/URLCleaning.swift
@@ -123,6 +123,35 @@ enum URLCleaning {
         return result.url == trimmed ? .unchanged : .rewritten
     }
 
+    /// Whether automatic clipboard rewrite may replace the pasteboard.
+    /// Safe string/URL flavors always qualify. Extra rich types (HTML/RTF
+    /// from Share) are allowed only when `string` is a single http(s) URL,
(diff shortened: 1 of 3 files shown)


vorssaint:
closing this one
the report behind this came from a stable build that had no youtube rule yet, the first stable release with that rule went out hours after the issue was opened, so nothing there shows that rich types were what kept the link from being cleaned
by the time this new check runs the text has already been accepted as a link, so it says yes to every set of types and the old protection is gone in practice
a browser image copy carries the image address as text, so with this change the image gets wiped and replaced by plain text whenever that address has a tracked parameter, and the secret mark that password managers put on the clipboard gets dropped the same way
if a share copy really lands with only html or rtf next to the link, adding those exact types to the allowed list is the smaller change that could still go in

Issue #1263, opened 2026-09-01
Keep Going: activate automatically when external display is connected


Advaith3600:
## Overview
Add automation to activate "Keep Going with lid closed" mode automatically when an external display is detected or connected, similar to how Keep Awake already activates for external displays.

## Use Case
Users who work with their MacBook closed and an external monitor want Keep Going to automatically enable when the external display connects, without having to manually toggle it each time.

## Current Behavior
- Keep Awake can automatically activate when an external display is connected (already implemented)
- Keep Going (lid closed mode) requires manual activation each time

## Proposed Behavior
- When "Keep Going with lid closed" is enabled in Settings, add an option to automatically activate it when an external display is detected
- Deactivate when the external display is disconnected (to match Keep Awake behavior)

PathGao:
You can already do this today, it just takes two switches instead of one.

In Keep Awake, turn on the external display option, and separately turn on "Keep going with lid closed". Keep Awake starts on its own when the monitor connects, and the lid-closed mode rides along with it. Unplug the monitor and both turn off again, which is the deactivate behaviour you asked for.

Does that cover your case? If you meant something else, for example lid-closed mode running when Keep Awake itself is off, tell me and I'll leave this open as a real request. Otherwise this is a wording problem rather than a missing feature: the lid-closed switch never tells you it follows the automation.

Advaith3600:
No, this doesn't cover my case. I want the behavior to work independently from Keep Awake.

Here's my specific workflow:

- I always have Keep Awake enabled when working from home (since I want the Mac to stay on)
- I have my Mac auto-sleep set to 1 minute of inactivity
- When I connect an external display, I want "Keep Going with lid closed" to activate automatically
- When I disconnect the external display, I want "Keep Going with lid closed" to turn off automatically — but I still want Keep Awake to remain on

The key difference is that I need independent control: Keep Going should toggle based on external display connection/disconnection, while Keep Awake remains in its own state. With your current suggestion, both would turn off together when I unplug the monitor, which isn't what I need.

PathGao:
You're right, that doesn't work, and I should have caught it.

The lid-closed switch today only does anything while Keep Awake is running, and once it's on it stays on the whole time — it doesn't know whether a monitor is attached. With Keep Awake always on like yours, that means the Mac would also keep going lid-closed in your bag, which is exactly what you don't want.

[...]

PR #1433, opened 2026-09-06
feat(keep-awake): selectable closed-lid gate conditions

Borisserz:
## Summary
- Closed-lid mode can require **external display**, **AC power**, and/or **network available**.
- Segmented **Any / All** match mode (default Any).
- Empty selection = no gate (same as the old toggle off). Keeps `clamshellExternalDisplay` so existing settings migrate cleanly.
- Network via `NWPathMonitor`; power/display reuse existing Keep Awake sources. Keep Awake automation itself is unchanged.

## Test plan
- [x] `./build.sh --test` (10768 checks)
- [ ] Enable closed-lid + only Network + Any → lid mode drops offline
- [ ] Enable Display + Power + All → both required
- [ ] Leave all conditions off → closed-lid applies whenever Keep Awake is active

Refs #1263


FrancescoRoveda:
Thank for your work! This establishes the independent closed-lid automation needed for several other workflows. 

My eventual use case is keeping network-dependent development agents running with the lid closed only while the Mac has external power and an available network connection. When commuting, the Mac may be closed in a backpack; running there would create unnecessary heat and battery use, and the agents cannot make useful progress without connectivity.

Once this lands, I would be intrested in a follow-up that generalizes the external-display gate into selectable conditions:
- External display connected
- External power connected
- Network connection available
- Require all or any selected conditions

I’m happy to open a separate issue for that follow-up rather than expanding this PR.

Borisserz:
Thanks — extend this PR (and #1263) with selectable closed-lid conditions instead of opening a separate issue. Working on it.

Borisserz:
Updated this PR with the selectable conditions (display / power / network + Any/All) instead of a follow-up issue.

FrancescoRoveda:
Hello @Borisserz, thanks again for expanding this PR. I tested it on a MacBook Pro M1 Pro running macOS 26.2 (25C56) and found one edge case.

Reproduction
1. Enable Keep Awake indefinitely.
2. Enable “Keep going with lid closed”.
3. Select Power + Network with match mode All.
4. Connect both AC and network; SleepDisabled becomes 1.
5. Close the lid.
6. Disconnect AC while the lid remains closed.
7. Wait 90 seconds, then reopen the lid while still on battery.

Observed
Vorssaint correctly changed SleepDisabled from 1 to 0 within one second of losing AC, but the Mac remained awake for the entire closed-lid interval. A local heartbeat continued without interruption, and pmset -g log recorded no new clamshell sleep.

It appears that restoring pmset disablesleep 0 does not cause macOS to replay a lid-close event that occurred while sleep was disabled.

Fix tested locally
After disableClamshell successfully restores normal lid sleep:
1. Re-evaluate shouldApplyClamshellNow() so a quickly recovered condition does not cause an unnecessary sleep.
2. Read AppleClamshellState from IOPMrootDomain.
3. If the policy still does not apply and the lid is closed, explicitly request sleep with:
Shell.run("/usr/bin/pmset", ["sleepnow"])

Vorssaint already uses this command for its Command Bar Sleep action, so it does not require another permission or a broader sudoers rule.

Verification after applying that change locally
- Removing AC while the lid was already closed caused macOS to record: Entering Sleep state due to 'Software Sleep'
- The Mac remained asleep, including returning to sleep after a maintenance dark wake.
- It woke from lid activity when reopened.
- Removing AC with the lid open restored SleepDisabled 0 without putting the Mac to sleep.
- Keep Awake remained active during the open-lid transition.
- ./build.sh --test: 10,772 checks passed.
- Full build completed without warnings.
- --selftest: SELFTEST OK.

I did not independently test loss of the Network condition, but it uses the same closed-lid disable path, so the power-condition reproduction exercises the shared issue.

Borisserz:
Thanks @FrancescoRoveda — that edge case checks out.

Pushed a follow-up that matches your local fix:
1. After a successful mid-session `disablesleep 0`, re-evaluate `shouldApplyClamshellNow()`
2. Read `AppleClamshellState` from `IOPMrootDomain`
3. If the gate still does not apply and the lid is closed → `pmset sleepnow`
4. Skip on quit (`synchronous`) and when IOPM cannot report lid state / lid is open / gate recovered

`./build.sh --test` → TESTS OK

Would appreciate a re-test of your AC-unplug-while-closed sequence when you have a minute.

FrancescoRoveda:
Re-tested 94e3925; the fix works.

Removing AC while the lid was already closed changed SleepDisabled to 0 and triggered: Entering Sleep state due to 'Software Sleep'

The heartbeat showed a clear sleep gap, maintenance wakes returned to sleep, and the final wake was caused by reopening the lid.

I also verified the open-lid guard: removing AC restored SleepDisabled 0 without sleeping the Mac, and Keep Awake remained active.

./build.sh --test passed all 10,772 checks, the full build completed without warnings, and --selftest returned SELFTEST OK.

Thank you very much! I hope this gets merged soon because it's freaking awesome!

sarat03:
Thanks for the work on this. The all-or-nothing closed-lid behaviour has bitten me before, and a clear test plan in the description made this easy to check on a second machine.

Built this branch at `94e3925` and ran it as a local dev build. All three test-plan boxes pass on my machine, and the regression from your exchange with @FrancescoRoveda passes too.

**All conditions off.** Closed-lid applies whenever Keep Awake is active. `SleepDisabled` sits at 1, matching the previous behaviour.

**Only Network, mode Any.** Turning Wi-Fi off drops `SleepDisabled` from 1 to 0, and turning it back on restores 1.

**Display + Power, mode All.**

| State | Expected | Observed |
|---|---|---|
| Both connected | 1 | 1 |
| Monitor only, no AC | 0 | 0 |
| AC only, no monitor | 0 | 0 |
| Both reconnected | 1 | 1 |

The two single-condition rows are the ones that matter, and both behave as `.all` requires. Under the old OR behaviour either alone would have held 1.

**Francesco's regression.** Both connected, mode All, lid closed, AC unplugged, waited two minutes, reopened. The Mac actually slept:

```
2026-09-11 16:35:53 Sleep  Entering Sleep state due to 'Software Sleep pid=42511':TCPKeepAlive=active Using Batt (Charge:80%) 113 secs
2026-09-11 16:37:46 Wake   Wake from Deep Idle [CDNVA] : ... lid SMC.OutboxNotEmpty
```

Keep Awake's own assertions stayed held throughout (`PreventUserIdleSystemSleep named: "Vorssaint: keep the Mac awake"`), so what gated off was the closed-lid extension alone, which is what the PR is for.

Environment: MacBook Air M2 (Mac14,2), macOS 26.6.2 build 25G83, built-in Liquid Retina plus a Lenovo P24q-20 2560x1440 attached over HDMI through a USB-C hub, two displays, default Spaces layout, local dev build (`./build.sh --dev --install`), app language English.

What I did not check: any condition combination involving Network together with Display or Power, and I did not test with more than one external display.

Issue #1420, opened 2026-09-06
Focus follows mouse deadlocks Vorssaint when its AX hit test lands on its own SwiftUI panel


Bald-M:
### Before you file

- [x] This is not a "feature does nothing" report. Accessibility is granted, Focus follows mouse is active, and the bundled self-test returns `SELFTEST OK`. I did not reset the permission because the captured failure is a persistent lock cycle rather than a missing/stale grant.
- [x] It still happens on the latest prerelease, v3.3.3-beta.4.

### Description

Vorssaint can freeze completely after Focus follows mouse performs an Accessibility hit test on one of Vorssaint's own SwiftUI surfaces. The menu bar icon remains visible, but the panel and shortcuts stop responding. The captured process had been running for about 41 minutes, used approximately 0% CPU and 48 MB RSS, and never recovered.

A live 5-second process sample captured the lock cycle. The main thread stayed here for all 4,300 samples:

```text
com.apple.main-thread
  -[NSApplication run]
    ...
      processObserverForRemoteProcessNotifications
        _os_nospin_lock_lock_slow
          __ulock_wait
```

At the same time, the Focus follows mouse query queue was performing a system-wide hit test. The hit landed back inside Vorssaint's own AppKit/SwiftUI accessibility implementation and waited for the main thread:

```text
com.vorssaint.focus-follows-mouse
  FocusFollowsMouseService.evaluateIfSettled()
    AXUIElementCopyElementAtPosition
      NSApplication.accessibilityHitTest
        NSWindow.accessibilityHitTest
          NSHostingView.accessibilityHitTest
            AccessibilityCore.hitTest
              ViewRendererHost.updateGraph
                _MovableLockSyncMain
                  _pthread_cond_wait
```

The `Vorssaint Window Focus` thread was also blocked for all 4,300 samples while resolving a focused-window notification through the same HIServices lock:

```text
WindowUseTracker.focusCallback
  AXWindowResolver.windowID
    _AXUIElementGetWindow
      getServerPortFromCache
        _os_nospin_lock_lock_slow
          __ulock_wait
```

This is a lock inversion: the background Focus follows mouse AX query has synchronously re-entered Vorssaint's SwiftUI accessibility tree and is waiting for the main thread, while the main thread is processing a remote AX notification and waiting on the HIServices lock held by the query path. The window-focus watcher is another waiter on that lock.

[...]

PR #1437, opened 2026-09-06
fix(focus-follows-mouse): hit test the application under the pointer, never the whole system

iltonandrew:
Refs #1420 — the lock inversion in the report: `FocusFollowsMouseService.target(at:)` ran the system-wide Accessibility hit test on its own queue, and when the pointer rested on one of Vorssaint's own SwiftUI windows HIServices served that hit test in-process on the calling thread, which then waited for the main thread while the main thread waited on the HIServices lock the query thread held. The existing self-PID guard ran only after the hit test, too late.

The hit test is no longer system-wide, so this process can never be the one it is served in. The window a click would hit is read first, on the main thread where the settle timer already runs, from `NSWindow.windowNumber(at:belowWindowWithWindowNumber:)` — the window server's own hit target, which honours other applications' click-through windows and the transparent parts of shaped ones, things the on-screen window list cannot express — with the CoreGraphics evaluation point flipped to Cocoa against the primary screen the way `AssistiveKeyboard.ownsCocoaPoint` flips it back. The number goes to the query queue, where its owner is read from `CGWindowListCopyWindowInfo([.optionIncludingWindow], …)` (a window server read: no main thread, no Accessibility lock) and the question is put to `AXUIElementCreateApplication(pid)`, which walks only that application's windows. When the click target is one of ours the query ends with no target, today's outcome via the PID guard, so hovering our own Settings window still does not activate the app behind it; over no window (number 0) nothing is focused. Nothing waits on the main thread from the query queue, and the one stale read that remains is benign by construction: a window that moves between the two reads costs one hover evaluation asking the wrong foreign application, which answers from its own windows and fails the existing role, window-id and PID checks — it can never aim the hit test at this process. The application element carries no timeout of its own, so it inherits the process-wide 0.35 s the system-wide element inherited (#938); the 0.25 s caps below it and the post-hoc PID guard are unchanged, the latter now documenting an invariant rather than enforcing one.

That the owner lookup resolves at all was measured on live window numbers rather than assumed: `CGWindowListCreateDescriptionFromArray` handed an `NSArray` of `NSNumber` returns zero descriptions for every window (it reads the array's element slots as ids), which would have made this a silent no-op with every check green; `.optionIncludingWindow` returns the right owner for each, and returns nothing for id 0 or an unknown id. Treating `NSWindow.windowNumber` as the `CGWindowID` and asking `.optionIncludingWindow` on its own are both how this repo already does it (`ShelfService`, `ScreenshotCaptureEngine`, `WindowEnumerator`, the screenshot controllers).

One narrowing worth naming: the click target is read on main and the owner resolved on the queue a moment later, so a window that closes in between yields nothing for that one evaluation and the window it uncovered is not considered until the pointer moves again; the old system-wide call, made at query time, would have seen the uncovered window. A pointer resting on a window that vanishes under it is the only case, and it fails safe.

The decision is pure — `FocusFollowsMouseSupport.hitTestOwner(of:ownProcessID:)` — and behaviour-checked: a foreign owner is returned, our own pid gives nil, an empty description list gives nil, a description without an owner gives nil. Four comment-stripped source-shape checks pin the dataflow rather than presence: the identifier bound from the helper is the one passed to `AXUIElementCreateApplication`, `AXUIElementCreateSystemWide` no longer appears in the file, no messaging timeout is set on the application element's identifier, and the click-target read precedes the query dispatch. Each goes red under its own mutation: helper forced to `nil` or to a constant pid (four checks between them), the application element built from this process's pid, the system-wide element restored, the timeout re-added, and the click-target read moved inside the queue.

Verification: Mac17,9 (M5 Pro), macOS 26.6.2 (25G83), 3024×1964 built-in display, local Developer build. `./build.sh` warning-free; `./build.sh --test` → `TESTS OK` (one run tripped the unrelated wall-clock check `a child inheriting stdout cannot leave the subprocess reader blocked` under load; the re-run was clean); `./build/Vorssaint --selftest` → `SELFTEST OK`, after rebasing onto `00df68b`. Not tested: the reporter's live freeze on macOS 15.2 (no deterministic trigger, not reproduced here); the new path on macOS 15.2 itself; what `windowNumber(at:)` answers over the bare desktop and the menu bar, and hands-on hovering with the dev build — this Mac's screen stayed locked for the whole session, so every probe point returned the lock shield and the evidence is the gates, the unit checks and the window-server measurements above, not a hover I watched. If AppKit reports no window over the bare desktop, hovering the desktop would stop bringing Finder forward; one probe on an unlocked Mac settles it and I will post the result here.

### DIFF
--- a/Sources/Vorssaint/Services/FocusFollowsMouse/FocusFollowsMouseService.swift
+++ b/Sources/Vorssaint/Services/FocusFollowsMouse/FocusFollowsMouseService.swift
@@ -8,9 +8,6 @@ import CoreGraphics
 final class FocusFollowsMouseService {
     static let shared = FocusFollowsMouseService()
 
-    /// No cap on this one: on the system-wide element a timeout is the default
-    /// for every question this process asks, whoever asks it (#938).
-    private let systemElement = AXUIElementCreateSystemWide()
     private let queryQueue = DispatchQueue(label: "com.vorssaint.focus-follows-mouse")
     private var timer: Timer?
     private var mouseMonitor: Any?
@@ -129,8 +126,21 @@ final class FocusFollowsMouseService {
                   .focusFollowsMouse, at: evaluation.point)
         else { return }
 
+        // The window server's own answer to "what would a click here hit",
+        // which the on-screen window list cannot express: it honours other
+        // applications' click-through windows and the transparent parts of
+        // shaped ones. It is AppKit, so it is asked here on the main thread.
+        // The evaluation point is CoreGraphics (top-left origin) and this
+        // wants Cocoa global coordinates, flipped against the primary screen
+        // the way ``AssistiveKeyboard/ownsCocoaPoint(_:)`` flips them back.
+        let cocoaPoint = NSPoint(x: evaluation.point.x,
+                                 y: (NSScreen.screens.first?.frame.maxY ?? 0) - evaluation.point.y)
+        let hitWindowNumber = NSWindow.windowNumber(at: cocoaPoint, belowWindowWithWindowNumber: 0)
         queryQueue.async { [weak self] in
-            guard let self, let target = self.target(at: evaluation.point) else { return }
+            guard let self,
+                  let target = self.target(at: evaluation.point,
+                                           hitWindowNumber: hitWindowNumber)
+            else { return }
             DispatchQueue.main.async { [weak self] in
                 guard let self, self.isRunning, self.nothingIsHeldDown,
                       self.state.isCurrent(evaluation),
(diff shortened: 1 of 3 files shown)


iltonandrew:
Closing: #1435 landed the same shape first — ownership from the window server, own click-through overlays skipped from an `NSApp.windows` snapshot on main, an app-scoped element so our own tree is never asked — and it is already out in 3.3.3, so this would only compete with it. My base predated #1435 (stale fetch before the rebase; my fault).

One difference, offered as a follow-up only if you want it: this branch read the click target from `NSWindow.windowNumber(at:belowWindowWithWindowNumber:)` instead of walking the on-screen list by rectangle and alpha, so another application's click-through overlay or the transparent part of a shaped window resolves to the window a click would actually reach; the list walk in #1435 stops at the overlay's owner. Narrow, and I have not measured it against a real overlay (this Mac was locked all day), so I would not open it without your word.

PR #1448, opened 2026-09-06
fix(settings): support mouse back and forward navigation

pergioa:
## Summary

Settings tracked only the selected page, so mouse Back/Forward buttons had no history to traverse. The global mouse-navigation tap could also consume the click while looking for navigation menu commands that Settings did not expose.

- Record page visits from sidebar selection and destination requests, restore section anchors, skip unavailable pages, and discard forward history after a new visit.
- Handle raw side-button gestures in the Settings window, independently of the optional global mouse-navigation feature. Preserve remapping/capture ownership and prevent navigation while unfocused or blocked by a modal dialog.
- Let the app receive its own raw side-button events rather than replacing them through Accessibility.
- Add localized Go > Back/Forward menu commands and Command-[ / Command-] shortcuts. These also cover events translated by mouse software or another running copy of Vorssaint before the Settings window receives the raw click.
- Add router regression coverage and a dedicated SettingsWindowTests.swift suite exercising the real AppKit responder chain, menu actions, keyboard equivalents, and synthesized mouse events.

This uses existing AppKit and router infrastructure with no new dependency, service, permission requirement, or persisted preference. History is in-memory and page-level; same-page section requests refine the current visit.

## Verification

Tested on Apple Silicon Mac17,9, macOS 26.6.2 (25G83), with the macOS 26 SDK.

- `./build.sh`: passed, release build.
- `./build/Vorssaint --selftest`: SELFTEST OK.
- `./build.sh --test`: TESTS OK (10681 checks), including preference cleanup.
- `./build.sh --dev`: passed; rebuilt and relaunched the developer app.
- `git diff --check`: passed.

AppKit regression coverage includes About > Mouse & Trackpad > Back > About through all three input paths, Forward restoration, no duplicate navigation on drag/release, capture blocking, and disabled commands without history. The test harness briefly activates an offscreen window to establish the responder chain.


Tests/SettingsWindowTests.swift
```diff
@@ -0,0 +1,124 @@
...
+            app.activate(ignoringOtherApps: true)
+        }
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+            app.stop(nil)
+            app.postEvent(NSEvent.otherEvent(with: .applicationDefined, location: .zero,
+                                             modifierFlags: [], timestamp: 0, windowNumber: 0,
+                                             context: nil, subtype: 0, data1: 0, data2: 0)!, atStart: true)
+        }
+        app.run()
+        expect(window.isKeyWindow, "Settings navigation fixture has a key window")
```
vorssaint: this fixture brings the test process to the front and takes keyboard focus for half a second on every test run, and the whole suite fails whenever the system does not grant that activation in time, which is never guaranteed. can you make these checks work without activating the app or waiting on a fixed timer, for example by overriding the key window state in a test subclass

Sources/Vorssaint/Core/SettingsNavigationStrings.swift
```diff
@@ -0,0 +1,28 @@
...
+    let forward: String
+
+    static func localized(_ language: AppLanguage) -> Self {
+        switch language {
+        case .enUS: return Self(go: "Go", back: "Back", forward: "Forward")
+        case .ptBR: return Self(go: "Ir", back: "Voltar", forward: "Avan\u{e7}ar")
+        case .tr: return Self(go: "Git", back: "Geri", forward: "\u{130}leri")
+        case .ru: return Self(go: "\u{41f}\u{435}\u{440}\u{435}\u{445}\u{43e}\u{434}",
+                              back: "\u{41d}\u{430}\u{437}\u{430}\u{434}",
+                              forward: "\u{412}\u{43f}\u{435}\u{440}\u{451}\u{434}")
```
vorssaint: every accented or non latin word here is written as a code escape, so nobody can read or check the translations. the other string files spell them out as plain text, can you do the same here

pergioa: Addressed in 18757630. All accented and non-Latin Settings navigation translations are now written as readable native text rather than Unicode escapes.

Tests/SettingsWindowTests.swift
pergioa: Addressed in 18757630. The fixture now uses a test subclass that overrides `isKeyWindow`, with no application activation, focus transfer, event-loop run, or fixed delay. The production menu items remain targetless, while the isolated fixture directs their actions explicitly to the test window. The merged full suite passes all 55,672 checks.

Issue #1353, opened 2026-09-04
GPU temperature missing on Mac mini M1 (Macmini9,1); --sensors finds no GPU keys


fakepooh:
### Description

GPU temperature is not shown on a Mac mini (2020), model `Macmini9,1`, with the base Apple M1 chip.

The built-in sensor diagnostic finds no sensors classified as `gpu` at all. It also classifies every discovered `Tp*` / `Te*` temperature sensor as `cpu-aux`; none match the current hard-coded M1 CPU core key set.

This looks like a sensor-mapping gap for `Macmini9,1`, rather than a UI-only issue.

### Environment

- Mac: Mac mini (2020)
- Model Identifier: `Macmini9,1`
- Chip: Apple M1
- macOS: 26.6.2
- Build: `25G83`
- Vorssaint: 3.3.2
- Installation method: Homebrew

### Steps to reproduce

1. Run Vorssaint 3.3.2 on `Macmini9,1` / Apple M1.
2. Open System Monitor and check temperatures.
3. GPU temperature is absent.
4. Run:

```bash
/Applications/Vorssaint.app/Contents/MacOS/Vorssaint --sensors
```

### Diagnostic output

```text
component    key   type   °C
cpu-aux  Te0a  flt    36.31
cpu-aux  Te0b  flt    36.31
cpu-aux  Te0x  flt    38.08
cpu-aux  Te0z  flt    38.08
cpu-aux  Te3a  flt    37.95
cpu-aux  Te3b  flt    46.65
cpu-aux  Te3x  flt    40.02
cpu-aux  Te3z  flt    60.02
cpu-aux  Tp2a  flt    38.94
cpu-aux  Tp2b  flt    48.04
cpu-aux  Tp2x  flt    45.98
cpu-aux  Tp2z  flt    60.98
cpu-aux  Tp3a  flt    39.62
cpu-aux  Tp3b  flt    47.02
cpu-aux  Tp3x  flt    48.59
cpu-aux  Tp3z  flt    60.59
cpu-aux  Tp4a  flt    40.71
cpu-aux  Tp4b  flt    50.81
cpu-aux  Tp4x  flt    47.61
cpu-aux  Tp4z  flt    64.61
cpu-aux  Tp5a  flt    38.77
cpu-aux  Tp5b  flt    48.97
cpu-aux  Tp5x  flt    41.22
cpu-aux  Tp5z  flt    57.22
cpu-aux  Tp7a  flt    38.84
[...]

PR #1450, opened 2026-09-06
fix(monitor): fall back to named HID temperatures on m1

EugeneCarldotme:
Refs #1353

Adds a named HID temperature fallback for M1 machines when a CPU or GPU temperature is unavailable through SMC. Valid SMC readings take precedence. With main's 3.3.4 CPU fix included, the reported Mac mini sensor dump retains its 64.61°C CPU reading; only its missing GPU reading needs the monitor's HID fallback.

The fallback accepts `eACC MTR Temp`, `pACC MTR Temp`, and `GPU MTR Temp` sensors and applies the existing chip-temperature limits. Fan curves retain their stricter mapped-core selection and can use named HID CPU readings when mapped SMC cores are absent. `--sensors` labels the named HID readings as CPU or GPU. No new SMC core mapping is inferred from the undocumented keys in the issue.

All four private HID functions are resolved at runtime. A missing symbol disables the fallback without preventing the app or fan helper from launching. The helper self-test covers each missing symbol. This follows the named HID temperature path used by [Stats](https://github.com/exelban/stats/blob/60e65d1454c647d9592550447757e509b01d5ebb/Modules/Sensors/readers.swift#L163) without adding a dependency.

The helper also provides `--temperature-diagnostics`, a read-only command that reports named HID samples and the same aggregated temperature inputs used by fan curves. It does not take fan ownership or write fan settings. `docs/M1-TEMPERATURE-VERIFICATION.md` describes running it in launchd's system domain on an affected M1.

Validation on an Apple M5 Pro:

- `./build.sh` passed, including the helper self-test.
- `./build.sh --test` passed 54,874 checks and preference cleanup tests.
- A focused probe passed all four missing-symbol cases and read 76 live HID temperature services in a user session.
- Neither built executable directly imports the four private HID symbols.
- The read-only helper diagnostic ran successfully on the host's normal M5 path.
- Working-tree and staged diff checks passed.

The affected M1 was unavailable. HID access from the M1 fan helper in launchd's system domain remains unverified, as do restoration of the reporter's displayed GPU temperature and physical fan-curve operation. The M5 user-session probe does not establish those results. No helper installation or system-service test was performed.

### DIFF
--- a/Sources/HIDEventSystem/include/HIDEventSystem.h
+++ b/Sources/HIDEventSystem/include/HIDEventSystem.h
@@ -2,3 +2,11 @@
 #include <IOKit/hid/IOHIDManager.h>
 #include <IOKit/hidsystem/IOHIDEventSystemClient.h>
 #include <IOKit/hidsystem/IOHIDServiceClient.h>
+
+// Temperature events are exported by IOKit but absent from the public SDK.
+typedef struct CF_BRIDGED_TYPE(id) __IOHIDEvent *IOHIDEventRef;
+IOHIDEventSystemClientRef IOHIDEventSystemClientCreate(CFAllocatorRef allocator) CF_RETURNS_RETAINED;
+void IOHIDEventSystemClientSetMatching(IOHIDEventSystemClientRef client, CFDictionaryRef match);
+IOHIDEventRef IOHIDServiceClientCopyEvent(IOHIDServiceClientRef service, int64_t type,
+                                        int32_t options, int64_t timeout) CF_RETURNS_RETAINED;
+double IOHIDEventGetFloatValue(IOHIDEventRef event, int32_t field);
--- a/Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
+++ b/Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
@@ -43,6 +43,7 @@ final class FanControlHardware {
     private var didDiscoverForceTestKey = false
     private var activeTargets: [Double] = []
     private var temperatureKeys: TemperatureKeys?
+    private let hidTemperatureSampler = HIDTemperatureSampler()
     private let temperaturePlatform = TemperatureSensorSelector.currentPlatform()
 
     init?() {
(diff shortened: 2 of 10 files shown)


Tests/MetricsTests.swift
```diff
@@ -1800,6 +1800,31 @@ struct MetricsTests {
             platform: .appleM1Family
         )
         expectClose(m1CPU ?? -1, 49.0, "M1 family uses hottest mapped CPU core")
+        let miniSMC = [(key: "Te0a", value: 36.31), (key: "Tp2b", value: 48.04),
+                       (key: "Tp4z", value: 64.61)]
+        expect(TemperatureSensorSelector.displayedCPUTemperature(
+            readings: miniSMC, platform: .appleM1Family
+        ) == nil, "Mac mini issue 1353 dump has no mapped SMC CPU core")
```
vorssaint: since 3.3.4 an m1 with none of its mapped cores shows the hottest of its other cpu sensors again, so this dump gives 64.61 on the current main and this check fails there. on that mac only the gpu still needs the hid path in the monitor. can you bring the expectation in line with that

Sources/HIDEventSystem/include/HIDEventSystem.h
```diff
@@ -2,3 +2,11 @@
...
 #include <IOKit/hidsystem/IOHIDEventSystemClient.h>
 #include <IOKit/hidsystem/IOHIDServiceClient.h>
+
+// Temperature events are exported by IOKit but absent from the public SDK.
+typedef struct CF_BRIDGED_TYPE(id) __IOHIDEvent *IOHIDEventRef;
+IOHIDEventSystemClientRef IOHIDEventSystemClientCreate(CFAllocatorRef allocator) CF_RETURNS_RETAINED;
+void IOHIDEventSystemClientSetMatching(IOHIDEventSystemClientRef client, CFDictionaryRef match);
+IOHIDEventRef IOHIDServiceClientCopyEvent(IOHIDServiceClientRef service, int64_t type,
+                                        int32_t options, int64_t timeout) CF_RETURNS_RETAINED;
+double IOHIDEventGetFloatValue(IOHIDEventRef event, int32_t field);
```
vorssaint: these four are private symbols, and linking them directly means the app and the fan helper stop launching on a macos that drops or renames any of them. can you look them up at runtime, the way the other private calls in this project do, so a missing one only turns this fallback off

Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
```diff
@@ -234,11 +235,14 @@ final class FanControlHardware {
...
+            cpuReadings: cpuReadings, gpuReadings: gpuReadings, platform: temperaturePlatform)
+        if smcReadings.contains(where: { $0.source == .hottestCPU }),
+           smcReadings.contains(where: { $0.source == .hottestGPU }) { return smcReadings }
         return FanControlPolicy.aggregatedTemperatures(
-            cpuReadings: cpuReadings,
-            gpuReadings: temperatureReadings(keys.gpu).map(\.value),
-            platform: temperaturePlatform
-        )
+            cpuReadings: cpuReadings, gpuReadings: gpuReadings, platform: temperaturePlatform,
+            hidReadings: hidTemperatureSampler.readings(platform: temperaturePlatform))
```
vorssaint, line 248: the fan helper runs as a background service outside the user session. can you confirm a curve on an m1 without these smc keys really gets its temperatures through here

docs/M1-TEMPERATURE-VERIFICATION.md
```diff
@@ -0,0 +1,19 @@
+# M1 temperature fallback verification
+
+PR #1450 needs a run on an affected M1 in launchd's system domain. A user-session probe or synthetic sensor test cannot establish that the background fan helper receives HID events.
```
vorssaint, line 3: this reads as a note about the pull request rather than about the app, and the fan side still has nothing showing it working on one of these macs. can you keep the change to the monitor in the app and leave the fan curves, the helper and this note out until the fan side has been seen working there

Issue #1338, opened 2026-09-03
Screen recorder: recording unexpectedly aborts when global shortcuts (e.g. screenshot) are pressed


ElectricalSoul:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When a screen recording session is in progress, pressing other global shortcuts (such as taking a screenshot or invoking the capture palette) causes the active screen recording to immediately abort and terminate unexpectedly. The recording session is not guarded against conflicting capture triggers.

### Steps to Reproduce

1. Start an area or full-screen video recording via the screen recorder hotkey.
2. While recording is actively capturing, press another Vorssaint shortcut (e.g. screenshot hotkey or capture menu).
3. Observe that the screen recording session abruptly stops and aborts instead of continuing.

### Feature Area

Tools

### Vorssaint Version

3.3.2

### macOS Version

macOS 26 (Tahoe)

### Language

Русский · macOS in Russian

### Mac and displays

Apple Silicon (M5 Pro)

### Installation Method

DMG

PathGao:
Any Vorssaint capture shortcut — screenshot, copy text from screen, the color picker, the capture palette — also acts as a stop button for the recorder. Press one while you're recording and the recording stops instead of the screenshot happening. That matches what you describe, and for the shortcuts that aren't the recorder's own it isn't deliberate.

I need one thing from you: when it stops that way, do you still get the video? It should stop the same as if you had pressed stop yourself, so the editor opens or the file lands where your recordings go. If the video is gone instead, say so — that changes how bad this is.

<details>
<summary>The technical side — you can skip this</summary>

`ScreenCaptureService.capture(initial:)` starts with `ScreenRecorderService.shared.stopOrCancelActiveCapture()` and returns early when it reports true. Every capture entry funnels through that one method — `ScreenshotService`, `ScreenTextService`, `ColorSamplerService`, and the recorder's own `toggle()` — so all four shortcuts end an active recording and swallow their own action. Only `toggle()` should behave as a stop; the comment on it ("starts when nothing is running and stops when something is") describes the recorder shortcut, not the others.

[...]

PR #1455, opened 2026-09-06
fix(recorder): keep recording when capture shortcuts fire

Borisserz:
## Summary
- Screen capture shortcuts (screenshot, screen text, color picker, palette) no longer call `stopOrCancelActiveCapture` at the start of `ScreenCaptureService.capture`.
- Active takes stay running; capture entry refuses to open over them via `hasActiveCapture`.
- The recorder dedicated shortcut routes through `ScreenRecorderService.toggle(fromShortcut:)`, which remains the only stop path.
- MetricsTests source-shape regression locks the separation.

## Test plan
- [x] `./build.sh --test` → TESTS OK (10640 checks)
- [ ] Start a screen recording, press screenshot / screen text / color / capture palette shortcuts → recording continues
- [ ] Press the recorder shortcut / panel toggle while recording → recording stops as before
- [ ] Start recording via the recorder shortcut when idle → chooser still respects “show capture menu on shortcut”

Refs #1338

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenCaptureService.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenCaptureService.swift
@@ -62,7 +62,16 @@ final class ScreenCaptureService: ObservableObject {
 
     private init() {
         for (tool, hotkey) in toolHotkeys {
-            hotkey.onPress = { [weak self] in self?.capture(initial: tool, fromShortcut: true) }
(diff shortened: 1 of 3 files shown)


vorssaint:
the current release already keeps screenshot and capture menu shortcuts from stopping an active recording. i'm closing this because the proposed changes do not add a missing protection for the reported case

PR #1457, opened 2026-09-06
feat(command-bar): optional ASCII layout while open

Borisserz:
## Summary
- Adds optional `commandBarSwitchToASCIILayout` (default off): on Command Bar show, select the first ASCII-capable keyboard layout among enabled sources via TIS; on hide, restore the previous input source.
- Pure helpers (`CommandBarASCIILayoutSupport`) are unit-tested without live TIS; settings toggle + all locales.

## Test plan
- [x] `./build.sh --test` — TESTS OK (10670 checks)
- [ ] Enable the toggle in Command Bar settings; open the bar on a non-Latin layout and confirm Latin input; close and confirm restore
- [ ] Leave the toggle off and confirm layout is unchanged

Refs #1327

### DIFF
--- a/Sources/Vorssaint/Core/CommandBarStrings.swift
+++ b/Sources/Vorssaint/Core/CommandBarStrings.swift
@@ -165,6 +165,8 @@ struct CommandBarFeatureStrings {
     let filesIgnoreAdd: String
     let compactModeToggle: String
     let compactModeCaption: String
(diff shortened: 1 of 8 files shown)


vorssaint:
i'm keeping #1342 for the optional keyboard layout change, where the review and follow-up fixes are already further along. closing this overlapping proposal so we can finish one version

PR #1461, opened 2026-09-07
feat(menu-bar): add a Histogram appearance mode with a styled hover panel

aangelinsf:
## Summary

Adds a third menu bar appearance mode, Histogram, alongside the existing
Values and Bars. CPU, GPU, memory and network render as a small rolling
history graph instead of a number or bar gauge; disk usage gets the same
treatment but on its own 30-day, hourly-sampled timescale (persisted across
launches via a new `DiskUsageHistoryStore`), since it changes far too slowly
for the few-minutes window everything else uses.

Sparkline mode trades each metric's literal number for a graph shape, so
that number needed to come back somewhere: hovering the status item now
shows a custom panel (`StatusItemHoverPanel`) styled as a miniature of the
app's own popover — same rounded card and material, a symbol/label/value
table instead of a flat dot-separated string, tabular digits so live values
don't jitter width tick to tick.

Also fixes a color inconsistency in the new appearance mode: sparklines at
normal usage were using a hardcoded color per metric that ignored the Normal
color picker in Settings → Monitor, while Bars style (and Histogram's own
Medium/High tiers) already respected it. All three tiers now use the same
Settings-driven colors everywhere, for consistency with what that picker
actually promises.

## Verification

MacBook Pro 18,2 (M1 Max), macOS 26.6.2 (25G83).

```
./build.sh          # release build, no warnings
./build/Vorssaint --selftest      # SELFTEST OK
./build.sh --test                 # TESTS OK (10636 checks), PREFERENCE CLEANUP TESTS OK
```

Installed and run as `Vorssaint (Developer)` alongside the release build,
verified visually: Histogram mode enabled for CPU/GPU/RAM/Network/disk
usage, hover panel content and layout, Normal-color-picker changes actually
changing sparkline color, and the previous ghosting bug in the hover panel
(stale row views not detached from `NSGridView` on refresh) is gone.

What I could not test: other Mac models, other macOS versions, RTL
languages, or a machine with no battery / no fan telemetry (the sparkline
code paths for those metrics are unaffected, but untested here).

## Anything else

This repository has issues disabled, so there's no `Refs #N` to add.


aangelinsf:
Opened in error — this belongs on a private fork, not upstream. Closing immediately.

Issue #1472, opened 2026-09-07
[Bug]: App Switcher leaves out running apps and shows only one window per app when many windows are open


akune:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

The App Switcher leaves running applications out of its list, and shows only one window for an app that has several.

Expected: every running app is listed, with its open windows.
Actual: with many windows open, whole applications are missing. Raising a missing app another way (Dock, Mission Control) makes it appear the next time the switcher is opened — but then only with the window that was just focused, not its others.

As far as I could measure it, the cause is the visible cap in `WindowEnumerator`. The list is trimmed to `maximumCount = 24` after it has been ordered by use. The cap counts entries and not applications, so one app with many windows consumes the budget and the rest of the list is cut off. On this Mac the enumeration produces 48 entries — 43 windows across 16 apps, plus 5 entries for apps without windows, with "apps without windows" set to all — so exactly half never reach the panel. Raising an app moves its window to the head of the use order, which is why that workaround works, and why only the raised window comes back: its siblings stay below the cut.

The comment above the enumeration loop already names this shape ("truncating here silently drops whole apps on busy Macs"). That earlier fix moved the truncation to after the use ordering; the cap itself has been 24 since 2.1.0. Issue #172 — "many apps missing from the switcher", closed 2026-07-07 as completed — is the same report, on 3.1.7.

I ruled out the Accessibility cross-check as a cause before landing on this: for both affected apps the window list from Accessibility was complete and returned in a few milliseconds, so the ghost veto was not what removed the windows.

### Steps to Reproduce

1. Open enough windows that the switcher's list passes 24 entries — for example a browser with ~18 windows plus a dozen other apps. Settings → App Switcher → "apps without windows" set to all reaches it sooner.
2. Leave a one-window app (here: Photos) untouched for a while, so it falls to the end of the use order.
3. Open the App Switcher. Photos is not listed at all.
4. Raise Photos from the Dock, then open the App Switcher again. Now it is listed.
5. Repeat step 4 with an app that has several windows (here: IntelliJ IDEA with 7). Only the window that was just focused is listed; the other six are not.

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.5

### macOS Version

macOS 26 (Tahoe)

### Language

English · macOS in English

### Mac and displays

[...]

PR #1473, opened 2026-09-07
fix(switcher): spend the visible cap across apps instead of windows

akune:
## Summary

The App Switcher trimmed its list to the first 24 entries of the use order. The cap counts entries and not applications, so a single app with many windows consumed the whole budget and pushed other running apps off the end: those apps looked like they were not running at all, and raising one by other means brought back only the window that had just been focused, its siblings still below the cut.

The earlier fix for that report repaired the half that truncated before the use order was applied. This is the other half. Every app is now given its most recently used entry first, in app order, and only the remaining slots go to further windows; with more apps than slots the apps compete with each other instead of one app's windows crowding the rest out. The survivors keep the incoming order, so index 0 stays the current window and index 1 the toggle target.

The cap itself moves from 24 to 48, matching the thumbnail cache in `WindowPreviewProvider`, which already budgets for that many previews.

The decision moved into `SwitcherSupport.visibleSelectionIndices` so the test list can reach it. The other callers of the shared walk, checked: Dock previews and the preview refresh (`listWindows(for:maximumCount:)`) pass a `filterPID`, so every entry belongs to one app and the selection reduces to the leading slice it always was — pinned by a test; the Command Bar (`listWindowsForCommandBar`) shares the cap and gains the same coverage.

On the reporting Mac, with 22 running apps: the old rule left one app visible in the worst ordering, the new one lists all 22.

## Verification

```sh
./build.sh                     # finished, no warnings
./build.sh --test              # TESTS OK (10708 checks)
```

Checked in both directions: green as committed, and red with `visibleSelectionIndices` reverted to a leading slice (3 checks fail).

Ran on a MacBook Pro (M4 Max, Mac16,5), macOS 26.6.2 (25G83), built-in Liquid Retina XDR scaled to 3008 × 1692 plus one external scaled to 2056 wide, one Space per display, English. Own build with a stable self-signed identity. 22 running apps, including a browser with 18 windows and an IDE with 7; switcher in the icon-row layout, one entry per app off, apps without windows set to all. Confirmed over a working session afterwards: every application and all of its windows are listed reliably.

What I could not test: other Macs and macOS versions; fullscreen- and Spaces-heavy layouts; the Command Bar, which shares the same cap; Dock previews on the device (unchanged by argument and pinned by a test, but not exercised by hand); a release build. No new user-facing strings, so no translations are involved.

## Anything else

Refs #1472


Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
```diff
@@ -499,13 +502,19 @@ enum WindowEnumerator {
         }
         var result = ordered
         if ordered.count > maximumCount {
-            result = Array(ordered.prefix(maximumCount))
+            // One entry per app first, then the remaining slots: an app with
+            // many windows must never push another app off the list entirely
+            // (issue #172).
+            result = SwitcherSupport
+                .visibleSelectionIndices(appPIDs: ordered.map(\.pid), limit: maximumCount)
```
vorssaint, line 509: could the shortcut for switching windows within the current app keep its own window list? with 24 recently used windows in that app and 26 other apps open, it now offers only 22 of those windows, where it previously offered all 24. other apps should not take places in a list that only shows the current app

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -621,6 +621,42 @@ enum SwitcherSupport {
...
+    /// looking at and index 1 the toggle target, so the survivors must not be
+    /// resorted into app groups.
+    static func visibleSelectionIndices(appPIDs: [pid_t], limit: Int) -> [Int] {
+        guard limit > 0 else { return [] }
+        guard appPIDs.count > limit else { return Array(appPIDs.indices) }
+        var chosen = Set<Int>()
+        var representedApps = Set<pid_t>()
+        for (index, pid) in appPIDs.enumerated() {
+            guard chosen.count < limit else { break }
+            guard representedApps.insert(pid).inserted else { continue }
```
vorssaint, line 650: when 48 apps are represented and the two most recently used windows belong to the same app, the previous window drops out. in the grid layout, a quick switch then opens another app instead of returning to that window. keep the previous window available when the list is full

Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
akune: Agreed — the window list is filtered to the front app only after the cap, so the per-app pass spent slots on apps it never shows. Fixed in 178d286: a window-scoped session passes its front process to the enumerator, and the cap then selects from that app's own windows, resolved through `appPID(forFrontmost:)` so a helper-owned front window still finds its app. Your case — 24 windows ahead of 26 other apps — keeps all 24 again; the all-apps list still spreads its slots. Pinned by tests that go red with the scope ignored.

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
akune: Right — the first pass skipped the second entry once its app was already represented. Fixed in 178d286: the two leading entries are kept before the slots are spread, as the leading slice always kept them. With 48 apps and the two most recent windows in one app, the survivors start with both again; in that case one app fewer fits. Pinned by a test that goes red without it.

PR #1474, opened 2026-09-07
fix(panel): dismiss on outside clicks

kostis4563:
**fix(panel): dismiss on outside clicks**


vorssaint:
closing this, the change removes two behaviors the panel has on purpose
tools like clipboard, media, quick toggles and window layout keep the panel open because their actions click into other apps, and after this change every one of those clicks closes the panel mid flow
the panel also stays open next to the settings window so changes show up live, and after this change any click in settings closes it
building the panel content only on first open and dropping the notes next to the code are unrelated to a dismissal fix
if there is one screen where the panel really should close on an outside click, a small change limited to that screen with the case it fixes could still come in

PR #1479, opened 2026-09-07
feat(window-layout): snap to top half before maximizing on upward drag

BenjaminD2023:
Dragging a window into the top-center snap area now previews and applies the upper half. Continuing upward through the menu bar promotes the target to maximize; retreating selects the upper half again. With a hidden menu bar, maximize activates at the physical top edge.

The selected action is stored with the target so preview and release agree. Existing corner targets, disabled-zone preferences, display seams, and system top-drag protection are preserved. The top-zone settings icon now reflects its initial upper-half placement.

Regression coverage includes the promotion boundary, placement geometry, retreat, corners, disabled top zones, and hidden/short/tall menu bars on offset displays. Validation passed: local unit tests (10,705 checks) and preference-cleanup tests; CI Build & selftest (including DMG packaging and unit tests); and CI Swift 6.0.3 compatibility (build, self-test, and unit tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. The format follows
 
 ## [Unreleased]
 
+### Changed
+- Dragging a window to the top edge first previews the upper half; pushing farther upward fills the screen.
+
 ## [3.3.5] - 2026-09-06
 
 ### Summary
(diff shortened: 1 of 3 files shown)


vorssaint:
closing this. the top edge was switched on purpose in 3.3.3 to fill the screen, after people asked for that instead of the upper half, and this brings the upper half back as the first thing the drag shows
a split inside the menu bar is a hidden gesture nothing explains, and the screen map in settings would only show the upper half for that zone while the release notes say it fills the screen
the upper half is still one shortcut away, so this is not worth the extra explaining
if people ask for the upper half at the top edge, a visible choice for that zone on the screen map would fit better than a hidden split

Issue #1483, opened 2026-09-07
Screen recorder: raw MOV has desynchronized video, system audio and microphone tracks


fakepooh:
### Description

When a recording is saved directly as the raw `.mov` master, the three media timelines do not line up correctly:

- video
- system audio
- microphone

The desynchronization is already present in the raw master, before using the recorder editor or the normal mixed-audio export path.

This is therefore separate from #1478, which asks to preserve the already-separate audio tracks during edited export. Here the source multitrack MOV itself has inconsistent timing.

### Environment

- Mac: Mac mini (2020)
- Model Identifier: `Macmini9,1`
- Chip: Apple M1
- macOS: 26.6.2
- Build: `25G83`
- Vorssaint: 3.3.2
- Installation method: Homebrew

### Steps to reproduce

1. Enable both **Mac sound** and **Microphone** for a screen recording.
2. Disable opening the recorder editor after recording, so the raw `.mov` master is saved directly.
3. Record a sequence with easy synchronization references, for example visible clicks/typing together with system sound and speech.
4. Open the resulting MOV in an editor that exposes individual tracks/timelines.
5. Compare the video, system-audio and microphone timing.

### Actual behavior

The video, system-audio and microphone tracks do not share the same timeline alignment.

The problem is visible in the raw MOV itself, not only after edited export.

### Expected behavior

All captured sources should be mapped onto one recording timeline so that their original synchronization is preserved in the master file.

### Code findings in v3.3.2

I checked the exact `v3.3.2` source (`fc302b67b509bf20fdb10132d8df2c23fb492ece`). There are several timing paths that look relevant.

#### 1. Audio buffers are re-timed using the total buffer duration as the duration of a timing entry

`RecorderWriter.retimed()` does:

```swift
var timing = CMSampleTimingInfo(
    duration: CMSampleBufferGetDuration(sampleBuffer),
    presentationTimeStamp: time,
    decodeTimeStamp: .invalid)

CMSampleBufferCreateCopyWithNewTiming(
    allocator: kCFAllocatorDefault,
    sampleBuffer: sampleBuffer,
    sampleTimingEntryCount: 1,
    sampleTimingArray: &timing,
    sampleBufferOut: &copy)
```

[...]

Issue #1478, opened 2026-09-07
Screen Recording: preserve separate system audio and microphone tracks on edited export


fakepooh:
### Problem or Motivation

Vorssaint already captures system audio and microphone as separate audio tracks in the recording master, and the editor exposes them independently as **Mac sound** and **Microphone** with separate mute and volume controls.

However, when an edited recording is exported as video, both sources are currently passed through a single audio mix and written as one AAC audio track.

This means the separation is lost exactly at the point where it is most useful: when the finished recording is taken into Final Cut Pro, DaVinci Resolve, Premiere, Logic, or another editor for further work.

A common example is a screen recording of a call or demo:

- system audio contains the other participants / application audio;
- microphone contains the speaker's voice.

Keeping those tracks separate makes it possible to independently clean up the microphone, remove noise, adjust levels, mute one side, or replace one source later.

The raw recording already contains this information, so it would be useful if an edited export could preserve it as well.

### Proposed Solution

Add an export option such as:

**Preserve separate audio tracks**

When enabled:

- system audio and microphone should remain separate tracks in the exported video;
- the tracks should retain meaningful metadata/labels such as `Mac sound` and `Microphone`;
- the editor's per-source mute/keep setting should determine whether that track is included;
- the editor's per-source gain should be applied independently to the corresponding output track;
- if only one source exists or is kept, the output should contain only that audio track.

The current mixed-audio export could remain the default for maximum playback compatibility.

If MP4 is not an appropriate container for simultaneous independent tracks across common players, a **multitrack MOV** export option would also solve the use case well.

### Why this fits the existing architecture

The recorder already appears to do most of the difficult part:

- `RecorderWriter` creates separate `systemAudioInput` and `microphoneInput` tracks in the `.mov` master.
- Each track receives `RecorderAudioSource` metadata.
- `RecorderComposition` distinguishes `.system` and `.microphone`.
- The editor displays separate lanes and provides independent keep/mute and gain controls.

The separation is lost in `RecorderExporter.exportVideo()`, where all audio tracks are fed into one `AVAssetReaderAudioMixOutput` and then written through a single `AVAssetWriterInput(mediaType: .audio)`.

So this seems less like a new capture feature and more like an additional export path that preserves information Vorssaint already records.

### Feature Area

Screen Recorder / Recorder Editor

### Does another app already do this?

[...]

PR #1484, opened 2026-09-07
fix(recorder): preserve raw recording track synchronization

EugeneCarldotme:
## Summary

Preserve per-sample durations and relative timestamps when retiming captured media. Previously, a 10 ms PCM buffer containing 480 samples became 4.8 seconds after writer retiming and 2,304 seconds after the microphone's two retiming passes.

Map ScreenCaptureKit, microphone, and system-audio tap timestamps onto the host clock used by pause and stop. Choose the recording origin immediately before capture starts so callback arrival order cannot discard earlier video or audio, and end the MOV at the captured stop time.

Both retiming call sites now use the shared helper. The local Developer build is named `Vorssaint Developer` and retains its separate `com.vorssaint.utils.dev` identity.

## Verification

Tested on Mac17,9, Apple Silicon, macOS 26.5, build 25F71, with Swift 6.3.3 and the macOS 26 SDK.

- `./build.sh --test`: 10,760 checks passed, including preference cleanup.
- Timing regression tests cover 44.1/48 kHz PCM, repeated retiming, multiple timing entries, decode timestamps, and conversion between clock epochs. They fail 22 checks against the original helper.
- The real MOV writer test saves and decodes video plus separate system/microphone tracks, checking callback reordering, pause/resume, stop duration, and audio/video markers.
- `./build.sh --dev --install` compiled without warnings. The installed app passed `--selftest` and strict code-signature verification.
- Local testing of the installed Developer build was also confirmed by the contributor before opening this PR.

The reporter's M1 Mac mini/macOS 26.6.2 combination and Swift 6.0.3 compatibility were not tested locally. This change addresses the raw master timeline; edited multitrack export in #1478 is separate.

Refs #1483

### DIFF
--- a/Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
+++ b/Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
@@ -26,7 +26,7 @@ protocol RecorderCaptureEngineDelegate: AnyObject {
 }
 
 /// Owns the ScreenCaptureKit stream and nothing else: it acquires pixels and
-/// system audio and hands them over untouched. No compositing happens here,
+/// system audio and maps their timestamps to the host clock. No compositing happens here,
 /// because everything the editor can change has to stay changeable after the
 /// recording ends.
 final class RecorderCaptureEngine: NSObject {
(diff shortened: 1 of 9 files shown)


Sources/Vorssaint/Services/Recorder/RecorderCaptureEngine.swift
```diff
@@ -134,13 +128,13 @@ final class RecorderCaptureEngine: NSObject {
                 return true
             }
             guard willStart else { return .streamFailed }
+            willStartCapture(CMClockGetTime(CMClockGetHostTimeClock()))
```
vorssaint, line 131: a delayed first image now leaves a blank opening, even with sound turned off. the saved video should start with an image while keeping both audio tracks aligned

build.sh
```diff
@@ -40,7 +40,7 @@ for arg in "$@"; do
 done
 
 if (( DEV )); then
-    APP_NAME="Vorssaint (Developer)"
+    APP_NAME="Vorssaint Developer"
```
vorssaint, line 43: can we keep the test app's current name? renaming it here leaves the previous copy installed, so someone could open the old version thinking they're testing this fix

PR #1485, opened 2026-09-07
feat(mouse): add programmable button actions

### LINKED ISSUE #282: [Feature]: Customize mouse buttons
### Problem or Motivation

First things, this app is awesome.  It has already replaced 3-4 of my other apps I was using.  It actually replaced cleanshot X for my use.

### Proposed Solution

I would love the ability to customize my extra mouse buttons with keyboard combos.
I am using better mouse but for some very basic things to remap my logitech mx mouse buttons.

Image

### Feature Area

Other

### Mockups / References

_No response_

CrowKiller:
## Summary

Adds direct, programmable actions for extra mouse buttons and side-wheel directions in **Settings → Mouse → Mouse button shortcuts**.

- Adds media, volume, display/keyboard brightness, Mission Control, app windows, Launchpad, desktop, Spaces, and scroll actions.
- Keeps existing custom-keyboard-shortcut mappings compatible and stored separately from direct actions.
- Adds optional repeat-while-held for continuous volume and brightness actions.
- Preserves existing ownership rules: the radial menu keeps its trigger, mouse exceptions pass input through, and disabled/uninstalled features stop claiming buttons.
- Reuses the existing WindowServer Spaces-shortcut wrapper, so user-customized or disabled macOS Spaces shortcuts retain their semantics.
- Includes complete localizations and backup support for new preferences.

Closes #282.

## Verification

- `./build.sh --test` — 10,704 checks passed; preference cleanup passed.
- `./build.sh` — release bundle compiled, assembled, and ad-hoc signed successfully.
- Manually tested direct volume actions, repeat-while-held, shortcut fallback, and Feature Hub availability on a physical mouse.

## Notes

The branch was rebased onto current `main` before this PR. Existing Swift 6 sendability warnings in `ShelfService` remain unchanged.


vorssaint:
closing this one, the direct actions would need a rewrite to fit the way the mouse button feature is built
when the system briefly switches the button watcher off, every held button is forgotten, so its release reaches the app as half a click
the scroll left and scroll right actions post a scroll the feature itself reads as a side wheel turn, so with a side wheel mapped the wheel's mapping fires and nothing scrolls
the button ownership check now parses every stored mapping on each drag of a held side button, the one path that has to stay cheap
mission control, app windows, launchpad and show desktop go through a private dock call while the drag gesture in the same feature asks for mission control and app windows through the system shortcut, so the two behave differently when that shortcut is off
fast forward and rewind post the same keys the radial menu already uses for next and previous track, a short press of those keys skips tracks
the media key list duplicates the one the volume roller already has
the repeat choice never survives a settings restore, the empty registered value passes the string dictionary check first
a flat menu of twenty three choices plus a checkbox on every row is more settings than this feature should carry
a smaller version with only the keys a key combination cannot express (volume, mute, play and pause, previous and next track, screen and keyboard brightness), reusing the existing media key posting, could still come in as a new pull request

Issue #1476, opened 2026-09-07
Custom keyboard shortcuts for Brightness Control


sidsharths1245:
### Problem or Motivation

As of now I have to go to the menu bar icon to change the brightness of the external monitors. On macbooks we can uise the builtin brightness keys to change the brightness but desktop computers like MacMini, and third pary keyboards, this is not possible.

### Proposed Solution

Custom keyboard shortcuts to change the birghtness would be greeat.

### Feature Area

Energy and display

### Does another app already do this?

BetterDisplay

PR #1488, opened 2026-09-07
feat(display): Custom display brightness shortcuts

EugeneCarldotme:
Refs #1476

## Summary

Two optional global shortcuts, one to lower display brightness and one to raise it.
Shift+Command+minus and Shift+Command+equals by default, both off until you turn them
on.

#1476 asks for these because the brightness keys only exist on a Mac laptop keyboard.
On a Mac mini or a Mac Studio with a third-party keyboard, nothing reaches an external
monitor's brightness without opening the menu bar panel first.

A shortcut adjusts the primary display, or the display under the pointer when the
existing "brightness keys change the display under the pointer" option is on. One
function, `BrightnessSupport.shortcutDisplay`, decides that, and it returns nothing
when the chosen display cannot take a brightness change. A press that finds no target
does nothing rather than falling back to another screen, so the shortcut never dims a
display you are not pointing at.

The feature rides on what display control already has. The two roles register through
`QuickToolHotkey`, the same path as every other shortcut in the app, and they call the
same `step` the sliders and the brightness keys call. Nothing new at runtime: no
dependency, no extra permission beyond the Accessibility grant display control already
asks for, and no observers or I2C traffic while the toggle is off. Both roles name
`brightnessControlEnabled` and `displayBrightnessShortcutsEnabled` as their enable
keys, so turning off display control, or just the shortcuts, releases the hot keys.

What this adds permanently is two shortcut roles, three preferences and one small
SwiftUI view reused in three places. The app checks each shortcut against system
shortcuts and window layout shortcuts, and shows the existing "shortcut unavailable"
line when registration fails.

You can edit the shortcuts in Settings under Energy and display, on the Shortcuts
page, and in the menu bar panel. The panel grew an Options disclosure that now holds
the pointer-following toggle alongside the new one, so the panel stays the same height
for anyone who never opens it.

All twelve languages carry the four new strings.

## Verification

Tested on Mac17,9, Apple Silicon, macOS 26.5 build 25F71, Swift 6.3.3 against the
macOS 26 SDK.

Automated checks:

- `./build.sh --dev --install` compiled without warnings. `codesign --verify --deep
  --strict` passed on the installed bundle and it satisfies its designated requirement.
- `./build.sh --test`: 10,713 checks passed, preference cleanup included.
- The installed Developer bundle returned `SELFTEST OK`.
- `git diff --check origin/main...HEAD` passed.
- New checks cover the routing helper (pointer target, primary target, an ineligible
  target, a missing pointer), both enable keys releasing the hot keys, feature
  availability, and the registered defaults matching each role's default shortcut.

Manual checks:

- On the built-in panel the shortcuts change hardware brightness, the same value the
  brightness keys and the slider move.
- On an external third-party monitor the change is software-side. macOS gives no route
  to that hardware's own brightness, so the shortcut moves the picture through this
  app's dimming instead. The way to use it is to leave the monitor set to 100% in its
  own menu and drive it from here. Apple displays and built-in laptop panels go through
  the system pipeline, so those change real hardware brightness.
- With pointer following on, the shortcut follows the pointer between displays. With
  it off, the shortcut stays on the primary display.
- Turning off display control while the shortcuts are on releases the keys, and the
  keystroke goes back to whatever else wanted it.

### Settings, Energy and display

The toggle sits under the pointer-following option, and the two shortcut rows appear
only while it is on.

image

### Menu bar panel, Options closed

image

### Menu bar panel, Options open

image


vorssaint:
hey @EugeneCarldotme thanks for working on this, the shortcuts are a useful addition! could you tuck the advanced display controls under “more options” in settings to keep the section tidy? a backup and restore test for the new preferences would be great too

EugeneCarldotme:
Hello @vorssaint i'm not sure if this is what you meant 

Let me know if i misinterpreted your feedback🙏🏻

Here's what i added 
image


image

Issue #1089, opened 2026-08-27
Screenshot preview steals focus


lmilojevicc:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

After taking a screenshot preview steals the focus from other apps. 

### Steps to Reproduce

1. Focus input field (for instance in browser)
2. Take a screenshot
3. Focus gets stolen

### Feature Area

Other

### Vorssaint Version

3.3.2

### macOS Version

macOS 27 (Golden Gate)

### Language

English

### Installation Method

Homebrew

iltonandrew:
Working on this — keeping the screenshot preview from activating and stealing key focus. PR to follow.

iltonandrew:
PR #1143 opened.

Root cause: the preview called `makeKey()` as it appeared. The panel is `.nonactivatingPanel`, so Vorssaint never came forward — but key status still moved off the field you were typing in, which is why the caret vanished with nothing on screen to explain it.

Presenting now leaves the keyboard alone and a click hands it over. One consequence worth knowing, since it changes a working flow: pressing ⌘C the instant a capture lands no longer reaches the preview — the keyboard stays with the app you were in until you click the preview. That matches the native macOS thumbnail, and arming the shortcuts earlier on hover is one line if the maintainers would rather have it that way.

Not verified on my side: the interactive behaviour itself. The branch is worth a moment on real hardware — take a screenshot while typing into a text field (the caret should stay put), then click the preview and press ⌘C and Escape (both should still work).

lmilojevicc:
It is fixed. Thanks!

PR #1490, opened 2026-09-07
feat(capture): optionally focus screenshot previews

santi-ug:
## Summary

Since #1143, the quick preview never takes the keyboard when it appears, so Escape, Enter and the other preview shortcuts do nothing until you click it. That's correct as a default, it's what keeps the preview from stealing focus out of whatever you were typing into, but it means anyone who reaches for Escape or Enter right after a capture has to click first. #1143 already flagged this as a one-line trade-off if a maintainer wanted it back for people who want it.

This adds a Screenshot setting, off by default, that makes the preview key the instant it appears. On, Escape and Enter work immediately. Off, nothing changes: a click still hands over the keyboard. The click hand-off code, shortcut meanings, copy-on-capture, auto-dismiss and the window-exclusion policy are all untouched.

Before, no such setting:

Screen capture settings before this change

After, the new toggle in More options, next to Preview position:

Screen capture settings with the new "Focus the preview automatically" toggle

## Verification

macOS 26.6.2 (25G83), Apple M4 Pro, single 3024×1964 Retina display, own ad hoc developer build, English (US).

```sh
./build.sh          # finished with 0 warnings
./build.sh --test   # TESTS OK (10700 checks), PREFERENCE CLEANUP TESTS OK
```

Installed with `./build.sh --dev --install` and tested on the running app:
- Toggle on: a new preview accepts Escape immediately, and Enter opens Edit immediately, with no click first.
- Toggle off: unchanged, Escape and Enter are ignored until the preview is clicked.
- Copy-on-capture still fires exactly as before, in both toggle states.
- Save, Copy, Delete and Edit still all work from the preview.
- Auto-dismiss timing is unchanged.
- The setting's value survived quitting and relaunching the app, and a full reinstall.

Not checked: the app is macOS-only, so no WebKit/browser surface applies here. I did not verify every non-English translation renders correctly in Settings, only that all 13 locales compile. I did not test multi-display preview positioning or a Space/full-screen-app interaction with the new toggle.

## Anything else

No open issue asks for this specific toggle; #1089 is the closed bug this is a follow-up to, already resolved and confirmed on 3.3.3, so nothing to reference with `Refs`.


AB-boi:
#1143 disrupted my workflow. I preferred it when the preview took the focus. Thanks for this!

vorssaint:
LGTM

Issue #1188, opened 2026-08-31
Displays: synchronise brightness between displays - ambient light sensor


kubaroki:
### Problem or Motivation

External displays do not respond now to ambient light sensor in the Mac. So with stronger light you have to manually increase the brightness of an external screen while built in changes itself, and in dim conditions, like at night, you have to dim it yourself too.

### Proposed Solution

A great feature would be a possibility to synchronise brightness from built-in and Apple screens - replicate Ambient light sensor and touch bar induced changes to a non-Apple external display.

The settings should be separated at the same time to some extent - for example, if we set one screen with lower brightness to 70% and the brighter (built in xdr screen) to 50% just to match the brightness between them, it should stay this way instead of setting them both to lets say 50%.

### Feature Area

Energy and display

### Does another app already do this?

MonitorControl

### Mockups / References

https://github.com/MonitorControl/MonitorControl - this app does have this feature

PR #1491, opened 2026-09-07
feat: add ambient light sensor sync and smooth brightness transitions for external displays

kubaroki:
## Summary

Adds ambient light sensor synchronization for external non-Apple displays, replicating built-in screen/ALS adjustments to external monitors with relative (additive) luminance offsets per display.

- **Why it belongs in Vorssaint:** External displays currently do not respond to the Mac's ambient light sensor or Touch Bar/keyboard brightness adjustments on the built-in panel. Users with external monitors must manually adjust brightness whenever ambient lighting changes. This expands the core `Displays` energy management capabilities.
- **Foundations & Runtime:** Uses existing `SMCClient` to read AppleSMC ambient light sensor keys (`ALV0/ALV1`, falling back to `LALS/RALS` in clamshell mode). It integrates directly into `BrightnessService.swift` and introduces smoothstep interpolation ($t^2(3-2t)$ over 400 ms) via a lightweight, short-lived 60 Hz timer. This prevents abrupt 1-second stepping while fully preserving the existing 50 ms DDC/CI command pacing and write coalescing.
- **Surface & Audience:** Adds an opt-in toggle in `EnergySettings` (`SettingsView.swift`) with per-display offset steps ($\pm\%$). Settings are persisted locally via display fingerprinting and excluded from portable backups (`SettingsBackupSupport.swift`).

## Verification

Tested on:
- **Hardware:** MacBook Pro (M3 Pro)
- **OS:** macOS 26.6.2 (25G83)
- **Displays:** Built-in Liquid Retina XDR + 1x External DDC/CI Monitor via USB C
- **Build Type:** Custom release build (`./build.sh --run`)

**Verification results:**
- All local unit tests passed cleanly (`./build.sh --test` — 10,000+ assertions checked).
- Full application build succeeded with zero warnings.
- Local runtime verification confirmed smooth visual transitions on external display during ambient light changes and manual built-in brightness stepping.
- **What could not be tested:** Intel-based Mac hardware and multi-external display setups (tested with single external monitor setup).

```sh
./build.sh
./build.sh --test
```

All 13 supported localization files in `BrightnessStrings.swift` have been updated with the required string keys.

## Anything else

Refs #1188


vorssaint:
closing this one. keeping external monitors in step with the built-in screen is a welcome idea, but this branch would need to be rebuilt to get there
the merges with main put old text back in three places, so the notes of a version already shipped lose their credits, the build script gets a second test compile that no longer builds while the new checks are missing from the list it actually uses, and the brightness settings show the permission row and the external displays note twice
the light sensor keys it prefers are accepted but their value type is never decoded, so with the lid closed the feature reads nothing and the older keys are never tried
the sensor is polled every second even while the built-in screen is the reference, and everything keeps running after brightness control is switched off, with the switch that would stop it no longer visible
every step of the fade goes through the same path as a key press, so the notch brightness notice pops up on its own whenever the room light changes
turning it on asks for accessibility, which nothing here needs
giving the brightness feature a second on switch removes its toggle from the command bar
a smaller version could still come in on a fresh branch off main, following the built-in screen only, with the gap kept per monitor, no sensor or lid-closed mode, fades that skip the notice, and the switch under more options

PR #1492, opened 2026-09-07
feat(screen-capture): let editor tool shortcuts be remapped

santi-ug:
The screenshot editor derives shortcuts from rail positions, leaving four of its thirteen tools without a key. This adds optional per-tool keys and modifier combinations through the existing shortcut recorder. Unconfigured tools keep their position digits.

Settings and the editor popover share the recorder rows, native reset symbols, and an orange validation caption below the list. Reserved editor commands and duplicate bindings are rejected without replacing the saved shortcut. Recorders remain usable while tool shortcuts are off. Recording a bare digit reorders the rail; Delete clears a shortcut. Row reset restores the position shortcut, and bottom Reset clears bindings and restores the default order. Rail badges and accessibility labels reflect the configured key.

The enabled preference keeps its type and meaning. One registered binding key travels with settings backups. Custom keys use the recorder's physical-key model and existing layout-aware labels; default numbers retain printed-character matching. No dependencies, services, or permissions added. Copy updated in all thirteen locales.

This offers configurable bindings for the request covered by #1285's fixed letter mode. It does not add a letter preset.

## Verification

MacBook Pro, Apple M4 Pro, Mac16,8, macOS 26.6.2 build 25G83. English UI, with German and Turkish layout screenshots.

- `./build.sh`: signed release bundle ready, zero warnings or errors.
- `./build.sh --test`: `TESTS OK (10897 checks)` and `PREFERENCE CLEANUP TESTS OK`, after rebasing onto main.
- `swift build`: `Build complete!`.
- The remapping block pins US and French layout data for the digit checks, and drives the pre-save rejection with injected owners so the order (tools, feature shortcuts, window layout, then the system table) is asserted without defaults or the WindowServer.
- On the first commit only: a separately signed native test app hosts this branch's actual SwiftUI controls, Settings page, and screenshot editor. Queued keyboard and mouse events passed 23 interaction checks, plus one Settings-page check and one process-relaunch check. These cover recording, duplicate and reserved rejection, Escape, clearing, both resets, digit reordering, disabled shortcuts, text-entry protection, modifier routing, and the real editor popover.

French and Russian keyboard-layout data passed automated label and routing checks. Physical typing on those layouts, third-party hotkey interception, several days of use, and native-speaker review of the translations remain unchecked. Save/copy workflows and multiple-display/Spaces arrangements were not exercised end to end. The Developer build is installed locally for further testing.

## Review changes

- Digits come from what the key types on the active layout, with Shift allowed, through the keycap cache with a shifted table added. On AZERTY, Shift+& moves the tool into slot 1 and bare & records as a binding without a conflict against 1. The synthetic US key codes for position digits are gone; badges read the position number and conflicts compare explicit bindings only. Imported bindings on a digit key are dropped on read, like reserved keys.
- Before saving, a recorded key is checked against other tools, enabled Vorssaint shortcuts, window layout actions and the live macOS table, in that order. A rejection shows the owner in the existing conflict caption and leaves the previous binding untouched.
- Not yet exercised in the live app: the macOS and feature conflict captions, and physical typing on a French keyboard. Both are covered by unit checks with injected owners and pinned layout data.

## Images

Native shared controls at the editor popover's 340-point width:

| Before | After |
| --- | --- |
| Number menus before | Recorded keys and native reset controls |

Editor rail, with a position digit before and a modifier binding after:

| Before | After |
| --- | --- |
| Editor before | Editor after |

Duplicate rejection preserves Text's T binding

Actual editor popover

Actual Settings page

Refs #1279


Sources/Vorssaint/UI/Screenshot/ScreenshotToolOrderControls.swift
```diff
@@ -147,13 +164,30 @@ struct ScreenshotToolOrderControls: View {
...
-    private func assign(_ number: Int?, to tool: ScreenshotSupport.Tool) {
-        withAnimation(.easeInOut(duration: 0.14)) {
-            persist(ScreenshotSupport.Tool.assigningShortcut(
-                number,
-                to: tool,
-                orderRaw: orderRaw))
+    private func assign(_ shortcut: GlobalShortcut?, to tool: ScreenshotSupport.Tool) {
+        let digit = shortcut.flatMap { shortcut in
+            shortcut.modifiers.isEmpty ? Int(shortcut.displayString) : nil
+        }.flatMap { (1...ScreenshotSupport.Tool.shortcutLimit).contains($0) ? $0 : nil }
```
vorssaint: on a french keyboard, typing 1 requires shift, so recording it for text saves a separate shortcut instead of moving the tool. select still shows 1, but pressing it now selects text, while the unused & key is rejected as a conflict with 1. could recording, conflict checks and the displayed shortcuts agree on the numbers the active keyboard actually produces

Sources/Vorssaint/UI/Screenshot/ScreenshotToolOrderControls.swift
```diff
@@ -147,13 +164,30 @@ struct ScreenshotToolOrderControls: View {
...
+        if let shortcut {
+            let reason: String?
+            if ScreenshotSupport.Tool.isReservedEditorKey(shortcut) {
+                reason = strings.toolShortcutReserved
+            } else if digit == nil, let conflict = ScreenshotSupport.Tool.bindingConflict(
+                for: shortcut, excluding: tool, orderRaw: orderRaw, bindingsRaw: bindingsRaw) {
+                reason = String(format: l10n.s.shortcutConflictFormat, conflict.screenshotTitle(strings))
+            } else {
+                reason = nil
```
vorssaint: this accepts shortcuts already used by macos or another enabled vorssaint feature, such as shift command 4 or the keep awake shortcut. recording temporarily silences those actions, but they return afterward, leaving a saved tool shortcut that triggers another action when used. could these conflicts be checked before saving while preserving the previous binding

Sources/Vorssaint/UI/Screenshot/ScreenshotToolOrderControls.swift
santi-ug: Fixed in c68b4ba. Recording now asks the layout what the key types, with Shift allowed, through the keycap cache with a shifted table added. On AZERTY Shift+& moves the tool into slot 1, and bare & records as a binding with no conflict against 1. The synthetic US key codes for position digits are gone: the badge reads the position number and conflicts compare explicit bindings only, so recording, conflicts and the rail all go by the same typed digit the editor already matches on. Imported bindings on a digit key are dropped on read, like reserved keys. Tests pin US and French layout data for these.

Sources/Vorssaint/UI/Screenshot/ScreenshotToolOrderControls.swift
santi-ug: Fixed in c68b4ba. Saving now runs the same three checks every shortcut field runs, after the tool duplicate check: enabled feature shortcuts via GlobalShortcutRole.conflict, window layout actions, then the live system table last since it is the one read that leaves the process. A rejection shows the owner in the existing conflict caption and leaves the previous binding in place. The decision lives in ScreenshotSupport.Tool.bindingRejection with the owners injected, so the order and the early return are unit-tested without defaults or the WindowServer.

PR #1498, opened 2026-09-08
refactor(tests): split standalone suite by feature

pergioa:
## What changed

The standalone test harness had grown to roughly 26,000 lines in one `main()`, which made feature ownership, fixture lifetimes, and suite-wide contracts difficult to review.

This change:

- reduces `Tests/MetricsTests.swift` to ordered orchestration and result reporting
- moves checks into focused feature/category suites using the existing `run(expect:)` convention
- keeps test execution serial and preserves the detached-process test’s late finalization
- moves temporary-resource cleanup into returning suite scopes
- separates build, repository, localization, and harness source contracts
- makes `build.sh --test` compile `Tests/*.swift`
- expands the UserDefaults namespace guard across every compiled Swift test file
- documents suite ownership and isolation in `Tests/README.md`

There is no XCTest migration, SwiftPM test target, new dependency, or production behavior change.

## Verification

Run locally on Apple Silicon, macOS 26.6.2 (25G83), against the pinned macOS 26 SDK after merging the latest `main`:

- `./build.sh --test` passes: 10,821 checks
- `PreferenceCleanupTests.sh` passes
- `git diff --check` passes
- `zsh -n build.sh` passes
- the namespace guard was mutation-checked with a temporary unswept test suite and failed as expected

The full release `./build.sh` exceeded the 15-minute execution window while compiling, without reporting a compiler error. Because that build did not complete, `./build/Vorssaint --selftest` was not run against a fresh full build.

No issue reference is attached to this refactor.


vorssaint:
closing this. splitting the big test file by feature is welcome and is already happening on main, but this version would have to be redone from scratch
main now runs the tests through a shared suite object with named groups that can run on their own, so every new file here still uses the old callback style and the flat ordered list in the runner
the test build on main already compiles every swift test file and was rewritten for incremental builds, so the build script change here is stale and the new source guard fails against the current line
main already has a localization test file with the same type name and a different shape, and the big file this change deletes moved in 35 commits since the branch was taken, so the merge would be a manual re-split of about two thousand lines
several suites are cut at historical seams and interleaved to keep the old order, so they still cannot run alone, which is the one thing the split should give
a smaller piece that would still fit today is the guard that scans every test file for the swept preference namespaces, written against the current runner

Issue #1465, opened 2026-09-07
App Updater showing updates for apps that are already updated


StudioKeys:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

So I'm running on v3.3.5 on my M4 MacBook Pro which has macOS 26.6.2, and when running the app updater, it shows me a update for apps that i have which is Android Studio and Obsidian, while both apps are updated to their latest version from the respective marketplaces

I think the version check of these is somehow a bit broken maybe

Also the updater just fails the check as in the state incomplete for me, somehow

Please let me know, if there is something thats needed to be attached as well to debug this issue

Thanks in advance

### Steps to Reproduce

1. Open Vorssaint
2. Enable App Updater from Features
3. Have Obsidian and Android Studio installed from an old DMG and update to latest using their own updaters
4. See Vorssaint still showing the old versions for these apps

### Feature Area

Tools

### Screenshots / Recordings

<img width="323" height="245" alt="Image" src="https://github.com/user-attachments/assets/c3a55960-0ad7-48bd-87e9-07967a72086c" />

### Vorssaint Version

3.3.5

### macOS Version

macOS 26 (Tahoe)

### Language

English

iltonandrew:
I'm picking this one up. From your screenshot the Android Studio row reads 2026.1 → 2026.1.4.7: the app's bundle only reports its major.minor, so any fuller catalog version looks newer, and that will nag forever unless the check learns when the installed number is too short to compare. The Obsidian row and the incomplete Antigravity check look like two separate mechanisms, and I can reproduce both on a Mac here, so I'll take all three together. I'll open a PR referencing this issue once it passes the build, test and self-test gates. Nothing has changed on your Mac yet.

iltonandrew:
Two things I can already say, and one I need from you.

**Android Studio** is the bug and I'm fixing it. Even the newest build (2026.1.4.7, checked today) tells macOS its version is only `2026.1`, so any fuller number from the catalog looks newer. The check will stop claiming an update when the installed number is just a shorter form of the catalog's.

[...]

PR #1499, opened 2026-09-08
fix(app-updates): stop reporting a catalog version that only adds precision

iltonandrew:
Refs #1465 — the Android Studio row. The online catalog compares the bundle's `CFBundleShortVersionString` with the cask version component by component, and `compare` counts a missing trailing component as zero. Measured today from the current Android Studio DMG (2026.1.4.7): the bundle reports `CFBundleShortVersionString = 2026.1` and `CFBundleVersion = AI-261.26222.65.2614.16204760`; nothing inside the bundle carries the four-component number the catalog records (`2026.1.4.7,quail4`). So every Android Studio install shows `2026.1 → 2026.1.4.7`, and no update can ever clear it — the reporter's screenshot exactly.

When the installed number is a strict numeric prefix of the catalog's — every component the bundle reports is equal, and the catalog only has more of them — the bundle cannot say whether it is behind, so the online source now reports no row, the same choice `isUncomparable` already makes for a package pinned to `latest`. The rule is one pure helper, `AppUpdatesSupport.reportsCoarserVersion(installed:than:)`, applied only in `onlineCatalogUpdates`; it tokenizes with the same `parts(of:)` the comparison uses, so a catalog that separates its extra precision with `-`, `_` or a space is treated the same as a dotted one. A difference inside the components the bundle does report still yields a row (`2026.1` vs `2026.2`, `1.12.7` vs `1.13.7`), a non-numeric part is a difference and not extra precision (`1.2` vs `1.2.4beta` still reports), and `compare`/`isNewer` keep their meaning for every other caller. The package-manager source is untouched: brew's own outdated verdict still produces a row for a bundle that reports `2026.1`, pinned by a check. `AppUpdateFeedSupport.update` (publisher manifests and appcasts) was examined and left alone — a vendor's feed numbers like its own bundle, and the appcast path already substitutes `CFBundleVersion`.

Two things to decide, named rather than hidden. First, this silences `x.y` installed against `x.y.z` catalog for every app, and an otherwise empty list renders as "up to date"; an alternative is to require at least two extra catalog components, which still fixes Android Studio while keeping `1.2 → 1.2.1` visible — not implemented, say the word. Second, the reporter's Obsidian row (`1.12.7 → 1.13.7`) is a different mechanism and is not changed here: Obsidian updates its core under `~/Library/Application Support/obsidian/` (measured on this Mac: bundle 1.12.7, `obsidian-1.13.7.asar` beside it) and leaves the installer bundle at the version first installed, so the row is right about the bundle and wrong about what runs; I told the reporter as much on the issue. The reporter's third symptom, "check incomplete / Antigravity", is a publisher manifest that answers 404; that is a separate small change I'll open next.

Checks: the reporter's case yields no row; a shared component that differs before a longer suffix still reports; the non-numeric suffix still reports; the tokenizer agreement is pinned; the deliberate `1.2` vs `1.2.1` omission is pinned by name; the package path pins `count == 1`. Each clause of the guard has a mutation that turns exactly one named check red (tokenizer swapped for dot-only, count clause dropped, numeric clause dropped, prefix-equality dropped).

Verification: Mac17,9 (M5 Pro), macOS 26.6.2 (25G83), local Developer build. `./build.sh` warning-free; `./build.sh --test` → `TESTS OK`; `./build/Vorssaint --selftest` → `SELFTEST OK`, after rebasing onto current main. Not tested: a real Android Studio installation on this Mac (only its DMG was read), the live catalog fetch through the app, and other macOS releases.

### DIFF
--- a/Sources/Vorssaint/Services/AppUpdates/AppUpdatesSupport.swift
+++ b/Sources/Vorssaint/Services/AppUpdates/AppUpdatesSupport.swift
@@ -118,6 +118,20 @@ enum AppUpdatesSupport {
         compare(versionCore(candidate), versionCore(installed)) == .orderedDescending
     }
 
+    /// Some app bundles publish a marketing short version with fewer components than the catalog
+    /// records. Android Studio reports 2026.1 for build 2026.1.4.7 (#1465), so a prefix match cannot
+    /// show whether it is genuinely behind; the honest answer is no row, as for an uncomparable `latest`.
+    static func reportsCoarserVersion(installed: String, than catalog: String) -> Bool {
+        let installedParts = parts(of: versionCore(installed))
+        let catalogParts = parts(of: versionCore(catalog))
+        guard installedParts.count < catalogParts.count,
+              (installedParts + catalogParts).allSatisfy({ $0.allSatisfy(\.isNumber) }),
+              zip(installedParts, catalogParts).allSatisfy({
+                  compareDigits($0.0, $0.1) == .orderedSame
+              }) else { return false }
+        return true
+    }
+
     private static func parts(of version: String) -> [String] {
         version
             .split(whereSeparator: { !$0.isLetter && !$0.isNumber })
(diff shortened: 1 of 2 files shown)


vorssaint:
thanks for looking into this. i'm closing this because the change can hide updates people still need and tell them everything is up to date. i don't want to solve a repeated warning by making real updates disappear.

Issue #1475, opened 2026-09-07
A confirmation popup for the sound swticher


sidsharths1245:
### Problem or Motivation

After the output device has been switched, i have to check it manully for the confirmation from the toolbar. It is quite inconvenient.

### Proposed Solution

A popup confirmation for the currently selected device would be great. Can be same style as the brightnjess slider popup.

### Feature Area

Sound

### Does another app already do this?

Supercharge

PR #1505, opened 2026-09-08
feat(audio): add optional output device confirmation

EugeneCarldotme:
Refs #1475

Adds optional confirmation after changing sound output through the shortcut, mixer menu, or command bar. Both visual and audible confirmation are off by default. Reuses the existing brightness popup and macOS sounds, with no new dependencies or bundled audio files.

### Review fixes awaiting push

The following changes are committed locally in `8e64376` and `051963f0`, but are not yet on this PR branch:

- Move the feedback controls below the output-switcher caption in the mixer menu's Options block and use matching checkboxes. Remove these controls from Keyboard shortcuts settings.
- Replace the sound picker and Play button with a single independent confirmation-sound toggle. Play the same Glass sound used by the timer, directed to the newly selected output.
- Show the selected device name in Dynamic Island when volume feedback is enabled and the Island accepts the notice. Otherwise, use the centered popup, following brightness feedback's routing behavior.
- Merge current main for Dynamic Island integration.

### Verification

Verified the local commits on Apple Silicon, model Mac17,9, macOS 27.0, build 26A428:

- Release build passed without compiler warnings; the bundled selftest returned `SELFTEST OK`.
- `./build.sh --test` passed 54,864 checks and preference-cleanup tests. Diff checks passed.
- Rendered and inspected the production checkbox controls. A temporary harness checked Island acceptance, rejection fallback, disabled routing, and disabled visual feedback with stand-in services.
- Confirmed by source inspection that the shortcut, mixer menu, and command bar share the output-switching function.

Physical output switching, audible playback on the selected device, and live Dynamic Island presentation were not verified. The previous screenshots were removed because they showed the superseded picker and Play controls.


Sources/Vorssaint/UI/MenuPanel/MixerSection.swift
```diff
@@ -365,6 +365,11 @@ struct MixerSection: View {
                     SoundOutputSwitcher.shared.syncWithPreferences()
                 }
 
+            OutputDeviceFeedbackControls()
+                .font(.system(size: 10.5, weight: .medium))
+                .toggleStyle(.switch)
+                .controlSize(.mini)
```
vorssaint, line 371: the caption right below belongs to the shortcut toggle above, and these controls now sit between the two. can you move them below that caption and use the same checkbox style as the other options in this block

Sources/Vorssaint/UI/Settings/ShortcutsSettings.swift
```diff
@@ -50,6 +50,7 @@ struct ShortcutsSettings: View {
                         } else if feature == .soundOutputSwitcher {
                             featureRows(feature, in: group)
                                 .settingsSectionAnchor(.soundOutputSwitcher)
+                            OutputDeviceFeedbackControls()
```
vorssaint, line 53: this page only holds shortcut settings, and the mixer options in the menu already carry these controls. please drop them from here

Sources/Vorssaint/UI/OutputDeviceFeedback.swift
```diff
@@ -0,0 +1,137 @@
...
+    @ObservedObject private var l10n = L10n.shared
+    @AppStorage(DefaultsKey.soundOutputOSDEnabled) private var enabled = false
+    @AppStorage(DefaultsKey.soundOutputConfirmationSound) private var sound = ""
+
+    var body: some View {
+        Toggle(l10n.s.soundOutputOSDEnable, isOn: $enabled)
+            .onChange(of: enabled) { _, _ in OutputDeviceFeedback.syncWithPreferences() }
+        if enabled {
+            HStack(spacing: 8) {
+                Picker(l10n.s.soundOutputConfirmationSound, selection: Binding(
```
vorssaint, line 112: a picker with every system sound plus a play button is more than this needs. a single on and off switch that plays one fixed sound on the new output, the way the timer alert already does, is enough here

Sources/Vorssaint/UI/OutputDeviceFeedback.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    static func show(device: MixerOutputDevice) {
+        guard Thread.isMainThread else {
+            DispatchQueue.main.async { show(device: device) }
+            return
+        }
+        guard isEnabled else { return }
+        if let screen = NSScreen.withMouse ?? NSScreen.main {
+            overlay.show(OutputDeviceOSDView(name: device.name, isHeadphones: device.isHeadphones),
```
vorssaint, line 31: when the island is on and showing volume changes, this still pops up in the middle of the screen. the brightness overlay already hands off to the island in that case, and this one should do the same

PR #1508, opened 2026-09-08
Add panel visibility options and right click menu customization

Parker-Aphelion:
## Purpose

Let users hide panel controls and arrange the right click menu.
The appearance and menu order remain unchanged by default.

## Changes

- Add separate settings to hide the panel logo and Settings / Quit buttons. Keep the beta badge and feedback button available.
- Add controls to reorder right click menu items and hide optional items.
- Add visibility controls for “Show brightness when adjusting,”
  System Sounds, and “Lower volume when headphones disconnect.”
  Hidden controls remain available in edit mode. Hiding a control
  does not change its setting.
- Keep section content in position when edit controls appear.
  Update the eye icons to show the current visibility state. Since that makes way more sense.
- Adjust panel height when optional content is hidden.
- Include the new preferences in settings backups. Handle duplicate and unknown menu IDs in saved settings.
- Add search terms and labels for all 13 supported language variants.

No dependencies were added.

Built with Swift 6.3.3 on macOS 26 without compiler warnings. The app self-test and all 10,802 unit checks passed, including the preference cleanup tests.

I love how much better these changes make it look and I love using it without the extra logo and buttons. I hope you guys add this then I can stop telling my friends to use my branch.


vorssaint:
closing this one. it bundles three separate ideas and each of them reaches further into the app than it should.
the panel logo and the settings and quit buttons are part of how the panel is meant to look, so a switch to hide them is not an option the app wants to offer.
the right click menu has nine short entries, and a drag and drop editor for it puts a second reorder list on a general page that is already full.
hiding three hand picked rows in the display and volume sections is not a coherent option, and the headphones row already lives inside the collapsed options group.
the panel height is pinned to the old 180 by default while the same formula gives 168 for every other state, and the 220 minimum is lifted only for the display section with its row hidden, so with content now pinned to the top the leftover space shows up as an empty band under the buttons.
the edit controls are lifted 7.5 points above the section header and only the popover compensates for that, so anywhere else a section is hosted, like the island, the buttons end up clipped.
a small separate pull request that only makes the eye icons in the panel show the current state, the way the settings editors already do, would still be welcome

Parker-Aphelion:
> the panel logo and the settings and quit buttons are part of how the panel is meant to look, so a switch to hide them is not an option the app wants to offer.

So you are set on an unnecessary gigantic icon at the top making the user need to move past it each time for zero reason. BUT you like it and wanna keep it. "Brand Identity" I guess? Dude it makes no sense. It's not like I made it default gone without a way to get it back. And the settings and quit buttons also make zero sense sooo many apps only have these in the right click BUT NO you gotta clutter it up with all the other views, always have it there. Because someone does not know how to right click. AND again WHY not a setting to remove it? It's on by default you allow so much customization it's a simple toggle.

> the right click menu has nine short entries, and a drag and drop editor for it puts a second reorder list on a general page that is already full.

The entire app lets you customize things BUT this somehow is a no? I have zero use to ever need "Activate for.." I want it on or off. Leave it on by default sure but WHY can't I just have 1 less thing there?

"About Vorssaint" Has no links or anything to do in it and is pointless after you read it once

If I wanna enable Cleaning Mode or Uninstall an app then I want to go in to the Utilities section. Like WHY not add everything from that section to the right click? Nope because you said these are to be there and no other choice. Like what if I want Camera preview for quick access in the right click? Nope because YOU don't want it like dude this section could be expanded and have more setting there. WHY not? The feeling from this app is full customization!!! but these tweaks you somehow reject because? What is honestly the reason it?

If you want the settings in a different location then just say so. I didn't wanna make a new section

> hiding three hand picked rows in the display and volume sections is not a coherent option, and the headphones row already lives inside the collapsed options group.

A lot of people don't need to constantly change where system sounds come from. And "Lower volume when headphones disconnect" you either want it or not. I see no reason to not allow more control I found those pointless while the rest honestly applies to everyone. These did not.

> the panel height is pinned to the old 180 by default while the same formula gives 168 for every other state, and the 220 minimum is lifted only for the display section with its row hidden, so with content now pinned to the top the leftover space shows up as an empty band under the buttons.

Okay maybe I missed something but from my testing it looks better so I'm not sure and would look into this

> the edit controls are lifted 7.5 points above the section header and only the popover compensates for that, so anywhere else a section is hosted, like the island, the buttons end up clipped.

Yeah I did not test it in all cases, that can be fixed BUT the spacing is not all broken, ugly as it was before so I could easily fix the other section to make those look good to but I just never use those and figured why spent more time. If you want those fixed fair I'd gladly do it but it seems like you just don't want more customization options.

Like seriously why do I need to see a toggle for "Show brightness when adjusting" each time I open up Displays? I either want it or I don't. Adding the icon like all the other views so Displays has the customization icon on the far right is nice it matches how the other views have it and then we can remove this option simply matching the rest. Explain why that is oh so bad? Because you are the only person that wants to keep toggling that setting on and off? WHY not have more customizations!!!

If I did anything in a bad way sure let me know and I can fix it. Like I clearly never tested out the Dynamic Island feature and I would gladly make it look better there but you are not giving any valid reasons. There's so many ways to make the entire UI/UX better but I kept these changes very minimal to make it look nicer right in line with how it works right now.

Screenshot 2026-09-18 at 09 50 24
Screenshot 2026-09-18 at 09 50 07
Screenshot 2026-09-18 at 09 49 40
Screenshot 2026-09-18 at 09 48 21

Issue #783, opened 2026-08-22
Let preview Preview panel resize independently of icon bar of app switcher


AB-boi:
### Description

When some apps have a lot of windows, the app switcher moves around as the preview panel resizes causing a lot of unnecessary motion:

https://github.com/user-attachments/assets/ad62c55f-f4b7-413e-847d-6f6745ce53d3

the bottom icon bar shouldn't move here, only the preview panel needs to resize.

### Steps to Reproduce

Open 6 windows for an app and invoke command + tab switcher → use left and right arrow keys

### Vorssaint Version

3.3.2

### macOS Version

macOS 15 (Sequoia)

### Installation Method

Homebrew

PR #1515, opened 2026-09-08
fix(switcher): keep icon rows steady and selected windows visible

shlok1806:
## Summary

Keep the App Switcher's icon row stationary when moving between apps with different window counts. Extra window previews scroll within the available width.

Keep the selected window visible as searches resize the preview strip or replace its contents, including when closing a window selects another app at the same position. The grouped simple-mode title strip uses the same protection.

## Validation

- Full automated suite passed 37,921 checks, including 120 native scroll checks for both modes, all preview sizes, narrowing, search restoration, window closure and changing selection.
- Three targeted mutation checks detected regressions in stable panel width, scrolling after resizing and tracking replaced windows.
- Developer build and installation, installed selftest and strict signature verification passed.

The scroll tests use production view and search code with synthetic entries and empty cards of the same size. Their windows remain hidden; no screenshots or synthetic input are used. Full visual behavior and other macOS versions were not exercised.

Refs #783

### DIFF
--- a/Sources/Vorssaint/UI/Switcher/SwitcherView.swift
+++ b/Sources/Vorssaint/UI/Switcher/SwitcherView.swift
@@ -328,9 +328,14 @@ struct SwitcherView: View {
                                 }
                             .frame(height: SwitcherIconRowLayout.previewCardHeight, alignment: .center)
                         }
-                        .scrollDisabled(appWindows.count <= Int(switcher.iconRowLayout.previewContentWidth / SwitcherIconRowLayout.previewCardWidth))
+                        .scrollDisabled(switcher.iconRowLayout.previewFitsWithoutScrolling(cardCount: appWindows.count))
                         .frame(width: switcher.iconRowLayout.previewContentWidth,
                                height: SwitcherIconRowLayout.previewCardHeight)
+                        .onAppear {
+                            let index = switcher.selectedIndex
+                            guard switcher.windows.indices.contains(index) else { return }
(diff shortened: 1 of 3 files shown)


Sources/Vorssaint/UI/Switcher/SwitcherView.swift
```diff
@@ -328,9 +328,14 @@ struct SwitcherView: View {
                                 }
                             .frame(height: SwitcherIconRowLayout.previewCardHeight, alignment: .center)
                         }
-                        .scrollDisabled(appWindows.count <= Int(switcher.iconRowLayout.previewContentWidth / SwitcherIconRowLayout.previewCardWidth))
+                        .scrollDisabled(switcher.iconRowLayout.previewFitsWithoutScrolling(cardCount: appWindows.count))
                         .frame(width: switcher.iconRowLayout.previewContentWidth,
                                height: SwitcherIconRowLayout.previewCardHeight)
+                        .onAppear {
```
vorssaint, line 334: searching for an app can hide its selected window when the preview narrows, even though the selection has not changed. please keep the selected window visible as search results change

shlok1806: fixed. the viewport is sized from the icon row and search drops apps from the row, so it could narrow while the selection stayed put and nothing re-scrolled. each strip now scrolls to the selection when its width changes too, through one helper instead of three copies of the scroll-to. the simple mode title strip had the same problem and uses the same helper.

PR #1516, opened 2026-09-08
fix(app-updates): verify catalog coverage for missing publisher feeds

iltonandrew:
A missing publisher update feed now falls back to the public catalog only when the catalog can verify that app. Apps without a usable catalog match stay named in the incomplete check; apps already up to date are covered too. Network errors, unreadable responses and deadline interruptions retain their warnings.

The catalog reuses its existing identity and compatibility checks, and successful publisher answers keep their precedence. No extra requests or background activity are added.

Validation includes 38,158 passing checks, 231 focused update checks, three detected regression mutations, 72 comparisons of combined source results, and a local installed build with passing selftest and strict signature verification. Service tests use extracted production methods with controlled network responses and clocks; no visual verification was performed.

Refs #1465


Sources/Vorssaint/Services/AppUpdates/AppUpdateFeedSupport.swift
```diff
@@ -136,6 +152,37 @@ enum AppUpdateFeedSupport {
             token: nil, bundlePath: app.path, storePage: nil)
     }
 
+    static func findings(loadResult: LoadResult, format: Format,
+                         apps: [AppUpdatesSupport.InstalledApp],
+                         operatingSystemVersion: String, kernelVersion: String,
+                         architecture: String) -> Findings {
+        switch loadResult {
+        case .absent:
+            return Findings()
```
vorssaint, line 161: could the warning go away only when the catalog can actually check this app? when the publisher's update address is missing and the catalog has no usable entry, the app disappears from the unchecked list and the check looks complete. keep the partial warning and the app's name when neither source can check it

PR #1518, opened 2026-09-09
feat(overlay): add screen annotation tools

RuanMD:
## Summary
- Add the Screen Annotation Overlay feature and Settings integration.
- Add pen, highlighter, arrows, lines, rectangles, ellipses, solid redaction, text, selection, eraser, colors, width control, undo, clear, and close actions.
- Remove blur and pixelate from the Overlay.
- Keep the canvas pass-through outside drawing mode and route text input through the native AppKit field.

## Verification
- `swift build` passed.
- `./build.sh --test` passed: 10,756 checks.
- `git diff --check` passed.
- Developer bundle rebuilt, installed at `/Applications/Vorssaint (Developer).app`, and launched.
- Running process verified from `/Applications/Vorssaint (Developer).app/Contents/MacOS/VorssaintDeveloper`.

## Review notes
- The official `/Applications/Vorssaint.app` was not modified.
- Please review the manual Overlay interactions, especially text commit/removal and eraser hit testing.

### DIFF
--- a/Sources/Vorssaint/Core/FeatureStrings.swift
+++ b/Sources/Vorssaint/Core/FeatureStrings.swift
@@ -97,6 +97,79 @@ enum FeatureStrings {
     static func whatsAppDownloads(_ language: AppLanguage) -> WhatsAppDownloadStrings {
         WhatsAppDownloadStrings.localized(language)
     }
+
+    static func annotation(_ language: AppLanguage) -> ScreenAnnotationStrings {
+        switch language {
+        case .enUS: return .enUS
+        case .ptBR: return .ptBR
+        case .tr: return .tr
+        case .ru: return .ru
+        case .es: return .es
+        case .de: return .de
+        case .fr: return .fr
+        case .it: return .it
+        case .ja: return .ja
+        case .ko: return .ko
+        case .zhHans: return .zhHans
+        case .zhTW: return .zhTW
+        case .zhHK: return .zhHK
+        }
+    }
+}
+
+struct ScreenAnnotationStrings {
+    let title: String
+    let pen: String
+    let highlighter: String
+    let undo: String
+    let clear: String
+    let exit: String
+    let openOverlay: String
+    let settingsDescription: String
+    let penDescription: String
+    let highlighterDescription: String
+    let undoDescription: String
+    let clearDescription: String
+    let controls: String
+
+    init(title: String, pen: String, highlighter: String, undo: String, clear: String,
+         exit: String, openOverlay: String = "Open overlay",
+         settingsDescription: String = "The overlay stays available while this feature is installed. Use the toolbar to choose the tool, color, and width.",
+         penDescription: String = "Freehand stroke",
+         highlighterDescription: String = "Semi-transparent stroke",
+         undoDescription: String = "Removes the last stroke",
+         clearDescription: String = "Removes all strokes",
+         controls: String = "Controls") {
+         self.title = title
+        self.pen = pen
+        self.highlighter = highlighter
+        self.undo = undo
+        self.clear = clear
+        self.exit = exit
+        self.openOverlay = openOverlay
+        self.settingsDescription = settingsDescription
+        self.penDescription = penDescription
+        self.highlighterDescription = highlighterDescription
+        self.undoDescription = undoDescription
+        self.clearDescription = clearDescription
+        self.controls = controls
+    }
+
+    var shortcutLabel: String { "⌃⌥⌘A" }
+
+    static let enUS = ScreenAnnotationStrings(title: "Screen annotation", pen: "Pen", highlighter: "Highlighter", undo: "Undo", clear: "Clear", exit: "Exit", openOverlay: "Open overlay", settingsDescription: "The overlay stays available while this feature is installed. Use the toolbar to choose the tool, color, and width.", penDescription: "Freehand stroke", highlighterDescription: "Semi-transparent stroke", undoDescription: "Removes the last stroke", clearDescription: "Removes all strokes")
+    static let ptBR = ScreenAnnotationStrings(title: "Anotação de tela", pen: "Caneta", highlighter: "Marca-texto", undo: "Desfazer", clear: "Limpar", exit: "Sair", openOverlay: "Abrir overlay", settingsDescription: "O overlay permanece disponível enquanto o recurso estiver instalado. Use a barra para escolher ferramenta, cor e espessura.", penDescription: "Traço livre", highlighterDescription: "Traço semitransparente", undoDescription: "Remove o último traço", clearDescription: "Remove todos os traços", controls: "Controles")
+    static let tr = ScreenAnnotationStrings(title: "Ekran açıklaması", pen: "Kalem", highlighter: "Vurgulayıcı", undo: "Geri al", clear: "Temizle", exit: "Çık")
+    static let ru = ScreenAnnotationStrings(title: "Аннотация экрана", pen: "Перо", highlighter: "Маркер", undo: "Отменить", clear: "Очистить", exit: "Выйти")
+    static let es = ScreenAnnotationStrings(title: "Anotación de pantalla", pen: "Pluma", highlighter: "Resaltador", undo: "Deshacer", clear: "Limpiar", exit: "Salir")
+    static let de = ScreenAnnotationStrings(title: "Bildschirmanmerkung", pen: "Stift", highlighter: "Textmarker", undo: "Widerrufen", clear: "Löschen", exit: "Beenden")
+    static let fr = ScreenAnnotationStrings(title: "Annotation à l’écran", pen: "Stylet", highlighter: "Surligneur", undo: "Annuler", clear: "Effacer", exit: "Quitter")
+    static let it = ScreenAnnotationStrings(title: "Annotazione dello schermo", pen: "Penna", highlighter: "Evidenziatore", undo: "Annulla", clear: "Cancella", exit: "Esci")
+    static let ja = ScreenAnnotationStrings(title: "画面注釈", pen: "ペン", highlighter: "蛍光ペン", undo: "取り消す", clear: "消去", exit: "終了")
+    static let ko = ScreenAnnotationStrings(title: "화면 주석", pen: "펜", highlighter: "형광펜", undo: "실행 취소", clear: "지우기", exit: "나가기")
+    static let zhHans = ScreenAnnotationStrings(title: "屏幕标注", pen: "画笔", highlighter: "荧光笔", undo: "撤销", clear: "清除", exit: "退出")
+    static let zhTW = ScreenAnnotationStrings(title: "螢幕標註", pen: "畫筆", highlighter: "螢光筆", undo: "復原", clear: "清除", exit: "離開")
+    static let zhHK = ScreenAnnotationStrings(title: "螢幕標註", pen: "畫筆", highlighter: "螢光筆", undo: "復原", clear: "清除", exit: "離開")
 }
 
 struct MixerFeatureStrings {
--- a/Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
+++ b/Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
@@ -1,0 +1,721 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import AppKit
+import Carbon.HIToolbox
+import SwiftUI
+
+/// Screen annotation overlay — draw on top of everything.
+///
+/// Architecture mirrors `ScreenshotSelectionController`:
+/// - Full-screen borderless panel that can become key.
+/// - View owns all mouse handlers (mouseDown/Dragged/Up).
+/// - `NSTrackingArea(.activeAlways)` so `mouseDragged` fires even in
+///   nonactivating panels.
+/// - No `wantsLayer` on the view — transparent panel + layer = blank output.
+final class ScreenAnnotationService: NSObject, ObservableObject {
+    static let shared = ScreenAnnotationService()
+
+    // MARK: - Panels
+
+    private var canvasPanel: AnnotationCanvasPanel?
+    private var toolbarPanel: NSPanel?
+    private var drawingView: AnnotationDrawingView?
+
+    // MARK: - State
+
+    private(set) var isDrawingActive = false
+    private(set) var strokes: [AnnotationStroke] = []
+    @Published private(set) var selectedStrokeIndex: Int?
+    @Published private(set) var shortcutRegistrationFailed = false
+
+    // Preferences (kept in sync with UserDefaults)
+    @Published private(set) var tool: AnnotationTool = .pen
+    @Published private(set) var color: AnnotationColor = .red
+    @Published private(set) var width = ScreenAnnotationSupport.defaultWidth
+
+    // MARK: - Monitors
+
+    private var keyMonitor: Any?
+    private var globalKeyMonitor: Any?
+
+    // MARK: - Shortcut (Carbon)
+
+    private var hotKeyRef: EventHotKeyRef?
+    private var hotKeyHandler: EventHandlerRef?
+    private var registeredShortcut: GlobalShortcut?
+
+    private override init() { super.init() }
+
+    // MARK: - Lifecycle
+
+    func syncWithPreferences() {
+        guard AppFeature.screenAnnotation.isAvailable else { teardown(); return }
+        syncShortcut()
+        loadPreferences()
+    }
+
+    // Entry point from menu / shortcut
+    @objc func toggleDrawing() {
+        guard AppFeature.screenAnnotation.isAvailable else { return }
+        if canvasPanel == nil { buildPanels() }
+        isDrawingActive ? exitDrawingMode() : enterDrawingMode()
+    }
+
+    @objc func clearAll() {
+        strokes = ScreenAnnotationSupport.clear(strokes)
+        drawingView?.needsDisplay = true
+    }
+
+    @objc func undo() {
+        strokes = ScreenAnnotationSupport.undo(strokes)
+        drawingView?.needsDisplay = true
+    }
+
+    @objc func hideOverlay() {
+        exitDrawingMode()
+        canvasPanel?.orderOut(nil)
+        toolbarPanel?.orderOut(nil)
+    }
+
+    func teardown() {
+        exitDrawingMode()
+        removeKeyMonitors()
+        unregisterShortcut()
+        canvasPanel?.orderOut(nil)
+        canvasPanel?.contentView = nil
+        canvasPanel = nil
+        toolbarPanel?.orderOut(nil)
+        toolbarPanel?.contentViewController = nil
+        toolbarPanel = nil
+        drawingView = nil
+        strokes.removeAll()
+    }
+
+    // MARK: - Drawing mode
+
+    private func enterDrawingMode() {
+        isDrawingActive = true
+        canvasPanel?.ignoresMouseEvents = ScreenAnnotationSupport.canvasIgnoresMouseEvents(isDrawing: true)
+        showPanels()
+        // Pattern from ScreenshotSelectionController:
+        // 1) orderFrontRegardless
+        // 2) makeKey — routes NSEvent.addLocalMonitor to this window
+        canvasPanel?.makeKey()
+        installKeyMonitors()
+    }
+
+    private func exitDrawingMode() {
+        isDrawingActive = false
+        // Keep the strokes visible, but restore pass-through to the app below.
+        canvasPanel?.ignoresMouseEvents = ScreenAnnotationSupport.canvasIgnoresMouseEvents(isDrawing: false)
+        removeKeyMonitors()
+        canvasPanel?.resignKey()
+        drawingView?.needsDisplay = true
+    }
+
+    // MARK: - Panel building
+
+    private func buildPanels() {
+        guard let screen = NSScreen.main else { return }
+        buildCanvasPanel(screen: screen)
+        buildToolbarPanel(screen: screen)
+    }
+
+
+    /// Builds the full-screen drawing surface.
+    /// Mirrors `ScreenshotOverlayPanel` from `ScreenshotSelectionController`.
+    private func buildCanvasPanel(screen: NSScreen) {
+        let p = AnnotationCanvasPanel(
+            contentRect: screen.frame,
+            styleMask: [.borderless, .nonactivatingPanel],
+            backing: .buffered,
+            defer: false)
+        p.isReleasedWhenClosed = false
+        p.isOpaque = false
+        p.backgroundColor = .clear
+        p.hasShadow = false
+        // Shielding-1 so it sits just below the toolbar
+        p.level = NSWindow.Level(rawValue: Int(CGShieldingWindowLevel()) - 1)
+        p.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary, .stationary, .ignoresCycle]
+        p.hidesOnDeactivate = false
+        p.acceptsMouseMovedEvents = true   // critical for mouseMoved delivery
+        p.ignoresMouseEvents = false
+
+        let view = AnnotationDrawingView(service: self)
+        view.frame = NSRect(origin: .zero, size: screen.frame.size)
+        view.autoresizingMask = [.width, .height]
+        p.contentView = view
+        self.canvasPanel = p
+        self.drawingView = view
+    }
+
+    private func buildToolbarPanel(screen: NSScreen) {
+        let host = NSHostingController(rootView: AnyView(AnnotationToolbarView(service: self)))
+        host.view.layoutSubtreeIfNeeded()
+        let size = host.view.fittingSize
+
+        let rect = NSRect(
+            x: screen.frame.midX - size.width / 2,
+            y: screen.frame.minY + 24,
+            width: max(size.width, 300),
+            height: max(size.height, 44))
+
+        let p = NSPanel(contentRect: rect,
+                        styleMask: [.borderless, .nonactivatingPanel],
+                        backing: .buffered,
+                        defer: false)
+        p.level = NSWindow.Level(rawValue: Int(CGShieldingWindowLevel()))
+        p.isOpaque = false
+        p.backgroundColor = .clear
+        p.hasShadow = true
+        p.hidesOnDeactivate = false
+        p.isReleasedWhenClosed = false
+        p.collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary, .ignoresCycle]
+        p.ignoresMouseEvents = false
+        p.contentViewController = host
+        self.toolbarPanel = p
+    }
+
+    // MARK: - Show
+
+    private func showPanels() {
+        guard let screen = NSScreen.main,
+              let canvas = canvasPanel,
+              let toolbar = toolbarPanel else { return }
+
+        canvas.setFrame(screen.frame, display: false)
+        canvas.orderFrontRegardless()
+
+        if let host = toolbar.contentViewController {
+            host.view.layoutSubtreeIfNeeded()
+            let size = host.view.fittingSize
+            let tb = NSRect(
+                x: screen.frame.midX - size.width / 2,
+                y: screen.frame.minY + 24,
+                width: max(size.width, 300),
+                height: max(size.height, 44))
+            toolbar.setFrame(tb, display: false)
+        }
+        toolbar.orderFrontRegardless()
+    }
+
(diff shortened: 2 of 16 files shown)


dorlugasigal:
Hi @RuanMD, thanks for adding this! I have been working on richer annotation tools in https://github.com/microsoft/ZoomitForMac/pull/64 and would like to build on your overlay rather than introduce a competing feature.

The proposed follow-up would keep your feature identity/settings and transparent Draw/pass-through model, then add a contextual toolbar, custom colors/styles, editable selections with undo/redo, curves/arrowheads/bindings, and optional pressure/smoothing/Smart Draw. I would also like to share the annotation model/editor/renderer with the existing screenshot editor, preserving its current tools and defaults, to avoid maintaining two drawing engines.

I plan to develop this as small layers in a native GitHub stack inside my fork and submit focused upstream extensions after this foundation lands. Before changing the shared architecture, are you and the maintainers comfortable with that direction? Are you already working on selection editing/history, multi-monitor placement, or the toolbar? Happy to align the boundaries so we do not duplicate your work.

dorlugasigal:
Quick update: I am starting the follow-up implementation in an isolated worktree based on your current head (5cf5bd7), with a development stack in my fork. First I will work on stable invoking-monitor placement, explicit Draw/Interact behavior, native text/control event routing, and reusing the existing hotkey helper. Next are a shared screenshot/overlay annotation core and the richer toolbar/editing capabilities described above.

I will not change your branch or submit a competing foundation PR. Upstream extensions will follow this PR after it lands, and I will reconcile with any changes you make here. Please flag any overlapping work so I can adjust the layer boundaries.

dorlugasigal:
The first follow-up layers are now available as a native GitHub stack in my fork: https://github.com/dorlugasigal/vorssaint-utils/pull/6 (the top PR links the five-layer stack).

The layers cover overlay lifecycle, shared screenshot/overlay primitives, transactional editing, native style controls, and selection/group/layer operations. They remain drafts while I reconcile exact-layer validation and add real, captioned demo recordings to each PR. Further drawing features are still in progress. Your PR remains the pinned foundation; these are not replacement upstream PRs.

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,721 @@
...
+                    }
+                    let svc = Unmanaged<ScreenAnnotationService>
+                        .fromOpaque(userData).takeUnretainedValue()
+                    DispatchQueue.main.async { svc.toggleDrawing() }
+                    return noErr
+                },
+                1, &spec, Unmanaged.passUnretained(self).toOpaque(), &hotKeyHandler)
+        }
+        var ref: EventHotKeyRef?
+        let status = RegisterEventHotKey(
```
vorssaint: while a shortcut field is recording, the other tools release their global keys so the combination lands in the field. this one stays registered, so typing its combination into any shortcut field toggles the overlay instead of being recorded. the other tools share one hotkey helper that already goes quiet during recording, can you register through it here too

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,721 @@
...
+private final class AnnotationCanvasPanel: NSPanel {
+    override var canBecomeKey: Bool { true }
+    
+    /// Route mouse events directly to the drawing view.
+    /// Pattern from Annotate's OverlayWindow: transparent panels may not
+    /// deliver events through the normal responder chain.
+    override func sendEvent(_ event: NSEvent) {
+        switch event.type {
+        case .leftMouseDown, .leftMouseDragged, .leftMouseUp:
+            NSLog("🎯 AnnotationCanvasPanel.sendEvent: \(event.type.rawValue) at \(event.locationInWindow)")
```
vorssaint: this logs every mouse event with debug markers, and the note above points at another app's code. can you drop both, nothing should print while someone is drawing and the code never names other products

Sources/Vorssaint/UI/Settings/FeatureHubSettings.swift
```diff
@@ -802,6 +803,7 @@ extension AppFeature {
         case .screenOCR: return hub.descScreenOCR
         case .screenshot: return FeatureStrings.screenshot(L10n.shared.language).hubDescription
         case .screenRecorder: return FeatureStrings.recorder(L10n.shared.language).hubDescription
+        case .screenAnnotation: return FeatureStrings.annotation(L10n.shared.language).title
```
vorssaint: this is the description under the feature card, so the card shows the title twice. it needs a short sentence saying what the feature does, in every language

Sources/Vorssaint/Core/FeatureStrings.swift
```diff
@@ -97,6 +97,79 @@ enum FeatureStrings {
...
+    static let de = ScreenAnnotationStrings(title: "Bildschirmanmerkung", pen: "Stift", highlighter: "Textmarker", undo: "Widerrufen", clear: "Löschen", exit: "Beenden")
+    static let fr = ScreenAnnotationStrings(title: "Annotation à l’écran", pen: "Stylet", highlighter: "Surligneur", undo: "Annuler", clear: "Effacer", exit: "Quitter")
+    static let it = ScreenAnnotationStrings(title: "Annotazione dello schermo", pen: "Penna", highlighter: "Evidenziatore", undo: "Annulla", clear: "Cancella", exit: "Esci")
+    static let ja = ScreenAnnotationStrings(title: "画面注釈", pen: "ペン", highlighter: "蛍光ペン", undo: "取り消す", clear: "消去", exit: "終了")
+    static let ko = ScreenAnnotationStrings(title: "화면 주석", pen: "펜", highlighter: "형광펜", undo: "실행 취소", clear: "지우기", exit: "나가기")
+    static let zhHans = ScreenAnnotationStrings(title: "屏幕标注", pen: "画笔", highlighter: "荧光笔", undo: "撤销", clear: "清除", exit: "退出")
+    static let zhTW = ScreenAnnotationStrings(title: "螢幕標註", pen: "畫筆", highlighter: "螢光筆", undo: "復原", clear: "清除", exit: "離開")
+    static let zhHK = ScreenAnnotationStrings(title: "螢幕標註", pen: "畫筆", highlighter: "螢光筆", undo: "復原", clear: "清除", exit: "離開")
```
vorssaint: everything beyond the six words passed in stays english in these eleven languages, and the other eight tools have no name in any language so their toolbar tips show raw english. can you add all of them in every language. also the word overlay is jargon for people reading the settings, something like draw on the screen reads better

Sources/Vorssaint/UI/Settings/SettingsDirectory.swift
```diff
@@ -78,6 +78,14 @@ enum SettingsDirectory {
                                        featureKeywords: AppFeature.allCases.map {
                                         ($0, [$0.hubTitle(s, hub: FeatureStrings.hub(language))])
                                        }),
+                SettingsDirectoryItem(page: .screenAnnotation,
```
vorssaint: this puts the page in the first group of the sidebar, between the features hub and energy. it belongs with the other tools, next to screen capture

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,721 @@
...
+        canvasPanel?.makeKey()
+        installKeyMonitors()
+    }
+
+    private func exitDrawingMode() {
+        isDrawingActive = false
+        // Keep the strokes visible, but restore pass-through to the app below.
+        canvasPanel?.ignoresMouseEvents = ScreenAnnotationSupport.canvasIgnoresMouseEvents(isDrawing: false)
+        removeKeyMonitors()
+        canvasPanel?.resignKey()
```
vorssaint, line 116: after escape the toolbar stays on screen but choosing a tool there no longer draws anything, can it go away together with drawing mode. also, does typing reach the app underneath again right after escape

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,721 @@
...
+    private func strokeIndex(at p: NSPoint, bounds: CGRect) -> Int? {
+        let point = CGPoint(x: p.x, y: p.y)
+        for index in strokes.indices.reversed() {
+            let stroke = strokes[index]
+            let points = stroke.points.map {
+                CGPoint(x: $0.x * Double(bounds.width), y: $0.y * Double(bounds.height))
+            }
+            guard let first = points.first else { continue }
+            var hit = CGRect(x: first.x, y: first.y, width: 1, height: 1)
+            for candidate in points.dropFirst() { hit = hit.union(CGRect(x: candidate.x, y: candidate.y, width: 1, height: 1)) }
```
vorssaint: this hit test uses the box around the whole stroke, so a long diagonal line gets erased or selected by clicking far from the ink anywhere inside that box. can it measure the distance to the stroke itself

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,721 @@
...
+    }
+
+    required init?(coder: NSCoder) { nil }
+
+    func beginTextEditor(at point: NSPoint) {
+        textField?.removeFromSuperview()
+        let field = NSTextField(frame: NSRect(x: point.x, y: point.y,
+                                               width: 300, height: 34))
+        field.font = NSFont.systemFont(ofSize: 18, weight: .medium)
+        field.textColor = .white
```
vorssaint: the text is typed in white no matter which color is chosen, so on a light page you cannot see what you type until it commits. use the chosen color here

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,721 @@
...
+    // MARK: - Stroke entry points (called from AnnotationDrawingView)
+
+    fileprivate func beginStroke(at p: NSPoint, bounds: CGRect) {
+        if tool == .text {
+            drawingView?.beginTextEditor(at: p)
+            return
+        }
+        if tool == .select || tool == .eraser {
+            let index = strokeIndex(at: p, bounds: bounds)
+            selectedStrokeIndex = tool == .select ? index : nil
```
vorssaint: selecting a stroke only draws a dashed box around it, there is nothing to do with it afterwards, no move and no delete. either give the selection an action or leave this tool out for now

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationSupport.swift
```diff
@@ -0,0 +1,119 @@
...
+         points: [AnnotationPoint], text: String = "") {
+        self.tool = tool
+        self.color = color.clamped()
+        self.width = width.clamped(to: 1...40)
+        self.points = Array(points.prefix(ScreenAnnotationSupport.maxPointsPerStroke))
+        self.text = String(text.prefix(2_000))
+    }
+}
+
+enum AnnotationMode: String, Equatable {
```
vorssaint: this enum, the rectangular flag above and the fixed shortcut label in the strings are never used, can you drop them

Sources/Vorssaint/UI/Settings/ScreenAnnotationSettings.swift
```diff
@@ -0,0 +1,55 @@
...
+                Text(strings.title)
+            }
+
+            Section {
+                LabeledContent(strings.pen, value: strings.penDescription)
+                LabeledContent(strings.highlighter, value: strings.highlighterDescription)
+                LabeledContent(strings.undo, value: strings.undoDescription)
+                LabeledContent(strings.clear, value: strings.clearDescription)
+            } header: {
+                Text(strings.controls)
```
vorssaint: this section is a glossary of what the toolbar buttons do, not a control. the tooltips on the toolbar already cover it, so it can go

RuanMD:
Addressed all 11 review comments from the last pass:

1. Shortcut registration moved from raw Carbon to `QuickToolHotkey`, so it now goes quiet through `ShortcutCapture` while another field is recording.
2. Removed the debug `NSLog` calls and the comment referencing another project in `AnnotationCanvasPanel.sendEvent`.
3. Feature card description no longer duplicates the title — added a real localized description string.
4. Completed the missing `ScreenAnnotationStrings` fields (description, tool names) across all 11 remaining languages; replaced "overlay" jargon with "draw on screen" wording throughout.
5. Moved the Settings page from the essentials group to the tools group, next to Screen capture.
6. Escape now also dismisses the toolbar together with drawing mode, and keyboard focus returns to the app underneath right away.
7. Rewrote the hit test to measure real distance to the stroke's segments instead of its bounding box, so long diagonal lines only erase/select near the actual ink.
8. Typed text now uses the selected color instead of hardcoded white.
9. Removed the select tool entirely rather than shipping a selection with no action, per your suggestion — can come back once move/delete is designed.
10. Removed the unused `isRectangular` flag and the other dead symbols you flagged.
11. Removed the "Controls" glossary section from Settings since the toolbar tooltips already cover it.

Branch is rebased on current `main`. `./build.sh --test` and `./build.sh --dev --install` both pass locally. Ready for another look whenever you have time.

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,664 @@
...
+    private func toolButton(_ t: AnnotationTool) -> some View {
+        Button { service.setTool(t); selectedTool = t } label: {
+            Image(systemName: toolSymbol(t))
+                .frame(width: 27, height: 25)
+        }
+            .buttonStyle(.borderless)
+            .background(selectedTool == t ? Color.accentColor.opacity(0.22) : .clear,
+                        in: RoundedRectangle(cornerRadius: 7, style: .continuous))
+            .foregroundStyle(selectedTool == t ? Color.accentColor : Color.primary)
+            .help(t.rawValue.capitalized)
```
vorssaint: the tips on the tool buttons come from the english name of each tool, so all nine show in english whatever language is set, even pen and highlighter, which already have a name in the strings. can the other seven get a name in every language and the tips use those names

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,664 @@
...
+            }
+            if points.count == 1 {
+                if hypot(point.x - first.x, point.y - first.y) <= tolerance { return index }
+                continue
+            }
+            var closest = CGFloat.greatestFiniteMagnitude
+            for (a, b) in zip(points, points.dropFirst()) {
+                closest = min(closest, ScreenAnnotationSupport.distance(from: point, toSegmentFrom: a, to: b))
+            }
+            if closest <= tolerance { return index }
```
vorssaint, line 404: for rectangles, ellipses and redaction boxes the two points are opposite corners, so this measures the distance to the diagonal between them, not to the drawn outline. clicking the edge of a big rectangle or the ring of an ellipse no longer erases it, while a click on the empty middle near that diagonal does. can the eraser follow the outline for those shapes, and the whole area for the solid box

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationSupport.swift
```diff
@@ -0,0 +1,127 @@
...
+    case inactive
+    case drawing
+    case exiting
+    case clearing
+    case teardown
+}
```
vorssaint: this one is still unused, can it go as well

Sources/Vorssaint/Core/FeatureStrings.swift
```diff
@@ -97,6 +97,77 @@ enum FeatureStrings {
...
+    let undo: String
+    let clear: String
+    let exit: String
+    let openOverlay: String
+    let settingsDescription: String
+    let hubDescription: String
+    let penDescription: String
+    let highlighterDescription: String
+    let undoDescription: String
+    let clearDescription: String
```
vorssaint: with the glossary section gone nothing reads these four descriptions anymore, so they can go from every language

RuanMD:
Addressed the 4 review comments plus 4 UX bugs found in manual testing on the dev build.

Review comments:
1. Toolbar tooltips now use the localized tool name instead of the raw English rawValue, for all 9 tools.
2. Eraser hit test for rectangle and ellipse now follows the outline (hollow ring), matching ScreenshotEditorController.hitTest. Redact keeps hitting the full area since it is a solid fill.
3. Removed the other unused symbol you flagged in ScreenAnnotationSupport.
4. Removed the four now-orphaned description strings from every language.

UX bugs found while testing the build locally, fixed in the same pass:
5. Switching tools while the text field was focused left it holding the keyboard; setTool now commits pending text and returns focus to the canvas.
6. The text field used a fixed 18pt font regardless of the width slider; it now matches the selected width like the committed text already did.
7. The text field was a fixed 300x34 box with no line wrapping; it now sizes itself to the remaining screen space and supports multi-line input, Return inserts a newline instead of committing.
8. Clicking an existing text label reopens it for editing instead of always creating a new one, and a click inside the field now reaches it through the normal responder chain so the caret can be placed anywhere in the text, not just the end.

Branch is rebased on current main. build.sh --test and build.sh --dev --install both pass locally. Ready for another look.

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,778 @@
...
+    }
+
+    override var acceptsFirstResponder: Bool { true }
+    /// Flipped so origin is top-left, matching screen pixels.
+    override var isFlipped: Bool { true }
+    
+    /// Critical: return self for all points so the view captures all mouse events.
+    /// Without this, a transparent view may let clicks pass through.
+    override func hitTest(_ point: NSPoint) -> NSView? {
+        return bounds.contains(point) ? self : nil
```
vorssaint, line 611: this answers every click with the canvas itself, so a click inside the text field never reaches the field and the caret does not land where you click. can the field get the click while it is up

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,778 @@
...
+            drawingView?.commitTextEditor()
+            if let index = textStrokeIndex(at: p, bounds: bounds) {
+                // Clicked an existing text: reopen it for editing instead of
+                // stacking a new one on top. Removed now; commitText below
+                // re-adds it (or drops it if left empty).
+                let existing = strokes.remove(at: index)
+                let origin = CGPoint(x: existing.points[0].x * Double(bounds.width),
+                                     y: existing.points[0].y * Double(bounds.height))
+                drawingView?.beginTextEditor(at: NSPoint(x: origin.x, y: origin.y),
+                                             existingText: existing.text)
```
vorssaint: reopening a label uses the color and width picked in the toolbar, not the ones it was drawn with, so clicking a red label to fix a typo while blue is selected turns it blue and changes its size. can the reopened text keep its own color and width

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,778 @@
...
+        field.textColor = service.map { NSColor(calibratedRed: $0.color.red, green: $0.color.green, blue: $0.color.blue, alpha: 1) } ?? .white
+        field.backgroundColor = .clear
+        field.drawsBackground = false
+        field.isBordered = false
+        field.focusRingType = .none
+        field.target = self
+        field.action = #selector(commitTextEditor)
+        field.delegate = self
+        field.usesSingleLineMode = false
+        field.cell?.wraps = true
```
vorssaint, line 555: while typing, a long line wraps at the edge of the screen, but once committed the label is drawn without wrapping and runs off the screen. can the committed text wrap at the same width as the field

RuanMD:
Addressed the text-wrap review comment plus 4 UX bugs found in manual testing.

Review comment:
1. Committed text now wraps at the same width as the live edit field, instead of being drawn on one line and running off the screen.

UX bugs found while testing the build locally, fixed in the same pass:
2. Clicks inside the live text field now reach the field itself, so the caret can be placed anywhere in the text, not just wherever the field last put it.
3. Reopening an existing text label now keeps its own color and width, instead of picking up whatever the toolbar currently has selected.
4. Closing the overlay (the X button) now commits or discards any text field that was still open, then clears every stroke, so nothing orphaned is left behind and nothing reappears the next time the overlay opens.
5. Changing color or width from the toolbar while a text field is open now updates that field live, instead of only taking effect on the next annotation.

Also rebased on current main and updated the annotation shortcut registration to match QuickToolHotkey's new storageKey parameter, and added SystemShortcutTakeover.swift to the test build sources it now needs.

build.sh --test and build.sh --dev --install both pass locally. Ready for another look.

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,817 @@
...
+
+    override var acceptsFirstResponder: Bool { true }
+    /// Flipped so origin is top-left, matching screen pixels.
+    override var isFlipped: Bool { true }
+    
+    /// Critical: return self for all points so the view captures all mouse events.
+    /// Without this, a transparent view may let clicks pass through.
+    override func hitTest(_ point: NSPoint) -> NSView? {
+        if let textField, textField.frame.contains(point) {
+            return textField.hitTest(point)
```
vorssaint, line 609: the point that arrives here is in the parent's coordinates while the field's frame is in this view's flipped ones, and nothing converts between them, so a click in a field opened in the lower half of the screen never reaches it. convert the point first, or let the default lookup find the field

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,817 @@
...
+    @objc fileprivate func commitTextEditor() {
+        guard let field = textField, let origin = textOrigin, let service else { return }
+        service.commitText(field.stringValue, at: origin, bounds: bounds,
+                           color: editingColor, width: editingWidth)
+        cancelTextEditor()
+    }
+
+    func controlTextDidEndEditing(_ notification: Notification) {
+        commitTextEditor()
+    }
```
vorssaint, line 590: does a committed label end up added twice, so undo needs two presses to clear it? taking the field off screen ends its editing and calls the commit again while the field is still remembered here

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,817 @@
...
+
+    /// Return inserts a newline instead of committing — only a click
+    /// elsewhere, a tool switch or Escape ends editing.
```
vorssaint, line 593: this says escape ends editing, but escape only leaves drawing mode and the field stays open with its text neither kept nor removed, so clear from the menu leaves that text on screen. can escape commit the text first, the way a tool switch does

RuanMD:
Addressed all 3 comments from the latest review.

1. Escape now commits any open text field before leaving drawing mode, matching the tool-switch behavior.
2. hitTest converts the click point into this view's flipped coordinate system before comparing it to the field's frame, so clicks reach a field opened in the lower half of the screen.
3. cancelTextEditor clears textField before calling removeFromSuperview, which synchronously re-enters controlTextDidEndEditing. That reentrant call previously found the old field still set and committed the same text a second time, needing two undos for one label.

Also rebased on current main. Tests/MetricsTests.swift split into per-feature files upstream, so I moved the annotation tests into a new Tests/ScreenAnnotationTests.swift following that pattern, registered under the new annotation suite, and updated FeatureCatalogTests.swift's feature count and id list.

build.sh --test and build.sh --dev --install both pass locally (2 unrelated pre-existing locale-formatting failures in RepositoryFeatureTests also reproduce on a clean checkout of upstream main, confirmed separately).

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,822 @@
...
+    }
+
+    // Entry point from menu / shortcut
+    @objc func toggleDrawing() {
+        guard AppFeature.screenAnnotation.isAvailable else { return }
+        if canvasPanel == nil { buildPanels() }
+        isDrawingActive ? exitDrawingMode() : enterDrawingMode()
+    }
+
+    @objc func clearAll() {
```
vorssaint, line 65: the earlier text feedback missed clear. can clear discard the active text editor as well as the completed strokes? while a label is being edited, this leaves its field on screen, and leaving drawing mode then commits that text back onto the cleared canvas

RuanMD:
Addressed the Clear button comment.

clearAll now cancels the active text editor before clearing strokes.
Previously it only emptied the strokes array, leaving an open text
field on screen if one was being edited. Leaving drawing mode
afterward committed that field back onto the canvas that had just
been cleared.

Also rebased on current main.

build.sh --test and build.sh --dev --install both pass locally. One
switcher test failed on the first run and passed clean on a second
run in isolation, unrelated to this change.

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,823 @@
...
+    }
+
+    // MARK: - Show
+
+    private func showPanels() {
+        guard let screen = NSScreen.main,
+              let canvas = canvasPanel,
+              let toolbar = toolbarPanel else { return }
+
+        canvas.setFrame(screen.frame, display: false)
```
vorssaint: the earlier requests missed this, it has been there since the first push. every time drawing starts the canvas is moved to the main screen, the one holding the focused window, and strokes are stored relative to the canvas, so with two displays the drawings left on one jump to the other and stretch to its shape when drawing starts again while focus is there. the first session also opens on the display of the focused window rather than the one under the pointer. please keep the canvas on its display while it still has strokes, and once it is empty order it out and open the next session on the display under the pointer, as the with mouse screen helper does for other panels

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,843 @@
...
+               height: max(size.height, 44))
+    }
+
+    fileprivate func orderOutCanvasWhenEmpty() {
+        guard strokes.isEmpty else { return }
+        if isDrawingActive { exitDrawingMode() }
+        canvasPanel?.orderOut(nil)
+        toolbarPanel?.orderOut(nil)
+        sessionState.resetAfterMutation(strokes: strokes)
+    }
```
vorssaint: this runs after every change that leaves the canvas empty, even while drawing, so undo on the only stroke, the eraser taking the last stroke, any eraser click on an empty canvas and clear on the toolbar now end the session, which makes clear do the same as exit. with the text tool on an empty canvas, clicking away from an empty text box also closes the overlay and then opens a new empty text box on the hidden canvas, which shows up in the next session. meanwhile leaving drawing mode with escape or the shortcut before anything is drawn keeps the empty canvas and its display, so the next session still opens there and not under the pointer. the empty canvas only needs to go away once drawing ends, so the next session can start under the pointer. can the canvas stay up while drawing mode is on, even if it becomes empty, and be ordered out with its display forgotten whenever drawing mode is off with no strokes left, including right after drawing mode ends?

Sources/Vorssaint/Services/ScreenAnnotation/ScreenAnnotationService.swift
```diff
@@ -0,0 +1,843 @@
...
+        if strokes.isEmpty, canvas.frame != placement.frame {
+            canvas.setFrame(placement.frame, display: false)
+        }
+        canvas.orderFrontRegardless()
+
+        if let host = toolbar.contentViewController {
+            host.view.layoutSubtreeIfNeeded()
+            let size = host.view.fittingSize
+            let tb = toolbarFrame(for: placement.frame, size: size)
+            toolbar.setFrame(tb, display: false)
```
vorssaint: while strokes are kept, this reuses the frame saved when drawing first started on this canvas and never looks up the saved display id. if that display is unplugged or the display arrangement changes before drawing starts again, the toolbar is placed where the display used to be, which can be off every screen or on another display, so its tools, colors and undo can end up out of reach. can the saved display be looked up by its id each time drawing starts, with the canvas and toolbar placed on its current frame, and the display under the pointer used once it is gone?

PR #1519, opened 2026-09-09
fix(cleaning-mode): prevent mouse button lock on exit

bebricoOOOOOOf:
### Summary
Fixes an issue where exiting Cleaning Mode (either via the "Unlock" button or the 5x Escape gesture) could leave the primary mouse button (LMB) locked or unresponsive across macOS.

### Root Cause
1. In `CleaningOverlayView`, clicking the "Unlock" button synchronously invoked `manager.deactivate()`.
2. `CleaningModeManager.hideOverlays()` synchronously called `overlays.forEach { $0.orderOut(nil) }` while the mouse event tracking loop was still active in AppKit/WindowServer, or while the physical mouse button / trackpad was still depressed.
3. Abruptly ordering out the full-screen shielding `NSPanel` mid-click or while held severed WindowServer's event-tracking state. WindowServer dropped the corresponding `leftMouseUp` event, stranding `CGEventSource.buttonState(.combinedSessionState, button: .left)` in the `true` (down) state and causing subsequent clicks across all applications to be ignored or treated as drags.

### Changes
1. **Deferred Dismissal**: `CleaningOverlayView`'s Unlock buttons now dispatch `manager.deactivate()` asynchronously (`DispatchQueue.main.async`), allowing the active click event cycle to finish cleanly before overlay teardown begins.
2. **Event Disengagement**: `hideOverlays()` sets `panel.ignoresMouseEvents = true` immediately and defers `orderOut(nil)` to the next run loop turn.
3. **Fail-Safe Release**: Added `ensureMouseButtonsReleased()` in `CleaningModeManager`, which inspects `NSEvent.pressedMouseButtons` and `CGEventSource.buttonState` and synthesizes an explicit `.leftMouseUp` / `.rightMouseUp` if a button remains stuck down upon overlay teardown.

### Verification
- `./build.sh --test`: All 10,755 checks passed (`TESTS OK`, `PREFERENCE CLEANUP TESTS OK`).

### DIFF
--- a/Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
+++ b/Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
@@ -279,42 +279,44 @@ final class CleaningModeManager: ObservableObject {
     }
 
     private func hideOverlays() {
-        ensureMouseButtonsReleased()
+        // Inspect whether any mouse button was down while the cleaning screen was up.
+        // Recovery must strictly be limited to presses that belonged to the cleaning screen,
+        // leaving any new clicks or drags initiated after teardown completely untouched.
+        let isLeftDown = (NSEvent.pressedMouseButtons & 1) != 0
+            || CGEventSource.buttonState(.combinedSessionState, button: .left)
+        let isRightDown = (NSEvent.pressedMouseButtons & 2) != 0
+            || CGEventSource.buttonState(.combinedSessionState, button: .right)
+
         let panels = overlays
         overlays = []
         panels.forEach { $0.ignoresMouseEvents = true }
-        DispatchQueue.main.async { [weak self] in
-            panels.forEach { $0.orderOut(nil) }
-            self?.ensureMouseButtonsReleased()
+
+        if isLeftDown || isRightDown {
+            ensureMouseButtonsReleased(left: isLeftDown, right: isRightDown)
         }
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
-            self?.ensureMouseButtonsReleased()
+
+        DispatchQueue.main.async {
+            panels.forEach { $0.orderOut(nil) }
         }
     }
 
-    /// Guarantees that any mouse buttons held or mid-event when cleaning overlays tear down
-    /// are cleanly released in WindowServer so the primary mouse button is never left stuck.
-    private func ensureMouseButtonsReleased() {
-        let isLeftDown = (NSEvent.pressedMouseButtons & 1) != 0
-            || CGEventSource.buttonState(.combinedSessionState, button: .left)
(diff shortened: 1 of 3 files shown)


Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
```diff
@@ -279,8 +279,44 @@ final class CleaningModeManager: ObservableObject {
...
+        let panels = overlays
         overlays = []
+        panels.forEach { $0.ignoresMouseEvents = true }
+        DispatchQueue.main.async { [weak self] in
+            panels.forEach { $0.orderOut(nil) }
+            self?.ensureMouseButtonsReleased()
+        }
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
+            self?.ensureMouseButtonsReleased()
+        }
```
vorssaint: if someone exits cleaning mode and immediately starts dragging a file, this later release can drop it before they let go of the button. can this recovery be limited to the press that belonged to the cleaning screen, leaving any new click or drag alone?

bebricoOOOOOOf: Good catch! You're totally right — the delayed `asyncAfter(0.1)` shotgun could prematurely drop an intentional drag started right after dismissing the screen.

Updated in 80a0ca6:
1. Removed the delayed `asyncAfter` timer entirely.
2. Scoped recovery strictly to buttons held while the cleaning screen was still up: we synchronously inspect `NSEvent.pressedMouseButtons` and `CGEventSource.buttonState(.combinedSessionState)` before tearing down the overlay. If a button was active on the cleaning screen (e.g. trackpad wiped down during the 5x Esc unlock), we synthesize the release immediately for that specific button before `orderOut(nil)`. If no button was held (clean Unlock click or clean Esc exit), no synthetic events are posted at all.
3. Switched cursor coordinate calculation to `CGEvent(source: nil)?.location ?? .zero` (avoiding inverted Cocoa coordinates from `NSEvent.mouseLocation`) and dropped the raw `.hidSystemState` query in favor of session state.
4. Added regression assertions in `MetricsTests` pinning that overlay teardown never schedules delayed releases and that recovery remains strictly conditional on active cleaning-screen presses.

vorssaint:
closing this one.
the unlock button only fires once the click is already released, and the lock never touches mouse buttons, so leaving cleaning mode has no way to hold a button down
the recovery posts a fake release for whatever button happens to be held at that moment, while the real release is still coming, so the same press ends up released twice
fake mouse events sent to the whole system are something the app avoids on purpose, and these carry none of the marks its own taps rely on to tell their events apart
if a button really stays stuck after leaving cleaning mode, an issue with the steps, the device and the macos version is the way to bring this back

Issue #790, opened 2026-08-23
Add Ethernet/Wi-Fi switching


JulianAlconcher:
### Problem or Motivation

On macOS, Ethernet is normally prioritized over Wi-Fi when both connections are available. This can be inconvenient when the Ethernet connection is connected but has no Internet access, while a Wi-Fi network does have Internet access.

In this situation, the Mac will continue trying to use the Ethernet connection, so the user has to manually disconnect the Ethernet cable before the Wi-Fi connection can be used. This is especially inconvenient when Ethernet is needed for local network access, but Internet access is available through a different Wi-Fi network.

### Proposed Solution

Add a function that allows users to easily switch between Ethernet and Wi-Fi, regardless of macOS's default network priority.

For example, a simple toggle could allow the user to:

Use Ethernet as the active Internet connection.
Switch to Wi-Fi when Ethernet has no Internet access.
Switch back to Ethernet when needed.

Ideally, this could be done without physically disconnecting the Ethernet cable or manually changing macOS network settings.

### Feature Area

Windows and Dock

JulianAlconcher:
Thanks for reviewing this!

What I had in mind is detecting when Ethernet is connected but isn't providing an Internet connection while Wi-Fi does have Internet access. In that situation, I would like to be notified and/or be able to switch to Wi-Fi without physically disconnecting the Ethernet cable.

An automatic switch would also work for me, but a notification with an option to switch would already solve the problem.

PR #1521, opened 2026-09-09
feat: turn wi-fi off while sleeping

rajnishps:
## Summary

Adds **Wi-Fi on sleep**, the Wi-Fi sibling of the shipped Bluetooth on sleep feature (#861): an opt-in Energy feature that switches the Wi-Fi radio off when the Mac sleeps and turns it back on wake **only when Vorssaint was the one that switched it off**. Wi-Fi already off before sleep is left alone and stays off on wake, and a Mac shut down while asleep still gets its Wi-Fi back on the next launch — the same "honest restore" debt model the Bluetooth feature uses, persisted in defaults and excluded from settings backup so it never travels between Macs.

Why it belongs here, on the four questions:

- **Built on:** the public CoreWLAN framework, which the app already uses in production for the Command Bar's Wi-Fi row — the exact same `CWInterface.setPower` toggle Control Center flips. No new dependency, nothing to carry when it breaks.
- **Runtime drag:** none. No polling, no observers, no cost while the feature is off; two `NSWorkspace` sleep/wake observers only while enabled.
- **Exposure:** nothing new — the app is unsandboxed and already performs this identical toggle from the Command Bar; no entitlements changed.
- **Surface vs. use:** one toggle on the Energy page next to Bluetooth on sleep, off by default, hidden entirely on Macs without a Wi-Fi adapter. People who already reach for Bluetooth on sleep are the same people who want a laptop fully dark in a bag.

Implementation mirrors the Bluetooth feature file-for-file so the two read as a pair: `WiFiSleepService` (CoreWLAN; calls stay synchronous inside the will-sleep handler since they cross to `airportd`), `WiFiSleepSupport` (pure decision logic), `WiFiSleepStrings` (all 13 languages), plus the catalog, defaults, runtime binding, Settings section, search directory and backup-exclusion wiring. New user-facing text compiles in every supported language; the localization coverage test enforces it.

## Verification

```sh
./build.sh                     # ✓ finished with zero warnings, bundle signed
./build/Vorssaint --selftest   # ✓ SELFTEST OK
./build.sh --test              # ✓ TESTS OK (10794 checks)
```

- `./build.sh --test` includes the new expectations: the catalog count (now 58) with stable feature ids, opt-in registered defaults, the five restore-decision cases (`WiFiSleepSupport`), the 13-language string contract, backup exclusion of the pending-restore key, and Energy-page visibility with the fifth feature.
- Ran on a MacBook Pro (Apple M4 Pro), macOS 27.0 (build 26A5425a), a release build from `./build.sh`, app language English (US). Single built-in display.
- **Could not test:** an actual sleep/wake cycle with the feature enabled — the sleep/wake handlers and the real radio toggle are outside the unit-test harness. The toggle call itself is the same one the Command Bar's Wi-Fi row already makes, but the timing of `setPower` inside the will-sleep window on other hardware (older Intel Macs in particular) deserves the maintainer's local run.

## Anything else

No issue exists for this. Related but distinct: #790 (tracked in the enhancement summary #838) asks for automatic Ethernet/Wi-Fi *switching*; this is the sleep power-off feature, the direct analog of the merged Bluetooth on sleep (#861), not a step toward that.

### DIFF
--- a/Sources/Vorssaint/Services/WiFi/WiFiSleepService.swift
+++ b/Sources/Vorssaint/Services/WiFi/WiFiSleepService.swift
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import AppKit
+import CoreWLAN
+
+/// Switches Wi-Fi off while the Mac sleeps, so a closed laptop goes dark on
+/// networks it is not using instead of staying joined to them all night.
+///
+/// The restore is honest in exactly the way Bluetooth on sleep is: the state
+/// found before sleep is remembered, so Wi-Fi the user had already turned off
+/// is never switched back on for them. The memory is a preference rather than
+/// a variable, so a Mac shut down while asleep still gets its Wi-Fi back on
+/// the next launch instead of staying dark.
+///
+/// Nothing runs while the feature is off: no observers, no polling, no cost.
+/// Unlike the Bluetooth power switch, the Wi-Fi calls cross to the Wi-Fi
+/// daemon (airportd) and block, so they stay synchronous inside the sleep
+/// handler: the system holds off on the will-sleep reply, and a toggle sent
+/// async could otherwise land after the Mac is already asleep.
+final class WiFiSleepService {
+    static let shared = WiFiSleepService()
+
+    /// A Mac either has a Wi-Fi adapter or it does not; the answer cannot
+    /// change while the Mac is running, so resolve it once. Desktops without
+    /// a card report no interface at all.
+    static let isSupported: Bool = {
+        CWWiFiClient.shared().interface() != nil
+    }()
+
+    private var observers: [NSObjectProtocol] = []
+
+    private init() {}
+
+    func syncWithPreferences() {
+        guard Self.isSupported else { return }
+        // A restore still owed here means an earlier run switched Wi-Fi
+        // off and never saw the wake, because the Mac was shut down (or the
+        // app quit) while it slept. The debt is paid before anything else,
+        // and paid even when the feature has since been switched off, so
+        // nothing Vorssaint took away is ever kept.
+        restoreIfOwed()
+        if AppFeature.wifiSleep.isAvailable,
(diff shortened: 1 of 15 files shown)


Sources/Vorssaint/Services/WiFi/WiFiSleepService.swift
```diff
@@ -0,0 +1,104 @@
...
+    private init() {}
+
+    func syncWithPreferences() {
+        guard Self.isSupported else { return }
+        // A restore still owed here means an earlier run switched Wi-Fi
+        // off and never saw the wake, because the Mac was shut down (or the
+        // app quit) while it slept. The debt is paid before anything else,
+        // and paid even when the feature has since been switched off, so
+        // nothing Vorssaint took away is ever kept.
+        restoreIfOwed()
```
vorssaint, line 42: with the feature switched off, every launch still asks the system whether this mac has a wi-fi card and whether the radio is on, even though no restore is owed, and both answers come from the wi-fi service and block until it replies. can you skip both unless the feature is on or a restore is actually pending

Sources/Vorssaint/Services/WiFi/WiFiSleepService.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    private func restoreIfOwed() {
+        let defaults = UserDefaults.standard
+        let restores = WiFiSleepSupport.restores(
+            owesRestore: defaults.bool(forKey: DefaultsKey.wifiSleepRestorePending),
+            isPoweredOn: Self.isPoweredOn)
+        // The debt is settled either way: Wi-Fi the user switched on
+        // themselves cancels it instead of waiting for the next sleep.
+        defaults.set(false, forKey: DefaultsKey.wifiSleepRestorePending)
+        if restores { Self.setPowered(true) }
```
vorssaint, line 90: switching the radio back on here makes the app wait a few dozen milliseconds, the same round trip the command bar's wi-fi toggle was moved off the main thread for. nothing at wake needs it to be synchronous, so can it run in the background the same way

Issue #1501, opened 2026-09-08
Media compression level ambiguous


chris-ha458:
### Problem or Motivation

In one of the **Media** tools, when converting file formats, there is an option for compression with three levels: **Low, Medium, and High**.

It is unclear what these levels refer to. Specifically, does **Low** mean a low level of compression, resulting in higher quality and a larger file size, or does it mean low quality, resulting in a smaller file size?

Based on my testing, it appears to mean the former: **Low** refers to low compression.

### Simplest version

Add a short explanation to disambiguate the options. For example:

- **Low** (high quality, large file size)
- **Medium** (balanced quality and file size)
- **High** (low quality, small file size)

### More flexible version

Expose a slider that maps to the underlying quality or compression setting. The two ends of the slider could indicate the expected trade-off between **quality** and **file size**, with a few intermediate snap points representing sensible defaults.

For example:

**Higher quality / larger file** ← slider → **Lower quality / smaller file**

This would make the trade-off explicit while allowing users more control than the current three-level selector.

### Feature Area

Tools

### Mockups / References

If AI-assisted PRs are allowed, I think I could put together a simple PR implementing this change.

If not, I can implement the simplest version by hand, changing the labels from:

- **Low** → **Low (large file size)**
- **High** → **High (small file size)**

This should be a straightforward change to the relevant UI strings.

Borisserz:
Opened a fix so Low/Medium/High compression labels spell out the quality vs file-size trade-off (across locales):

Compare / PR: https://github.com/vorssaint/vorssaint-utils/compare/main...Borisserz:feat/media-compression-labels?expand=1

PR #1523, opened 2026-09-09
fix(media): explain compression levels inside buttons

samanyudas:
## Summary

Clarify what Low, Medium, and High compression mean by adding quality and file-size explanations inside each button.
Include translations for all 13 supported languages.

This implements the simple explanation proposed in #1501.
Compression behaviour stays unchanged.

## Verification

- Full local build passed.
- Self-test passed.
- Automated tests passed all 10,755 checks.
- Manually checked the compression controls in Video and Image.
- Manually checked the descriptions in all supported languages. The text remained readable without clipping.

Refs #1501


Sources/Vorssaint/UI/Media/MediaWorkspaceView.swift
```diff
@@ -894,15 +895,22 @@ struct MediaWorkspaceView: View {
...
+                    Image(systemName: level.symbolName)
+                    Text(compressionTitle(for: level))
+                        .lineLimit(1)
+                }
+                .font(.system(size: compact ? 10 : 11, weight: .semibold))
+                Text(compressionDescription(for: level))
+                    .font(.system(size: compact ? 10 : 11))
+                    .multilineTextAlignment(.center)
+                    .frame(maxWidth: .infinity)
+                    .fixedSize(horizontal: false, vertical: true)
```
vorssaint: in the menu bar panel each of these three buttons is only about 90 points wide, so a description like the portuguese or russian one wraps to four lines and the whole row grows to almost three times its height. can you keep the texts but show the description as one secondary line under the three buttons that follows the selected level, like the note under the target size field

samanyudas:
Moved the description below the three compression buttons, where it follows the selected level and uses the same secondary styling as the target-size note. The buttons return to their original height, and all 13 translations are preserved.

Issue #1015, opened 2026-08-26
Feature Request: Screenshot with Pin-to-Screen Functionality


535486092:
### Problem or Motivation

When working on projects, I often need to reference screenshots (e.g., design mockups, error messages, documentation) while coding. Currently, I have to switch between windows to view them, which breaks my workflow.

### Proposed Solution

I'd love to see a screenshot feature with the following capabilities:
Take a screenshot — quickly capture a selected area of the screen
Pin the screenshot to the screen — the captured image floats on top of all other windows as a pinned overlay
Draggable — the pinned screenshot can be freely moved around the screen with the mouse
Multiple pins — support pinning multiple screenshots simultaneously, each independently draggable
Scroll to zoom — when the mouse hovers over a pinned screenshot, using the scroll wheel should zoom in/out on the image

### Feature Area

Windows and Dock

### Does another app already do this?

Snipaste

### Mockups / References

This feature would be especially useful for developers and designers who frequently need to compare UI implementations against design specs, or reference documentation while coding. The ability to pin, drag, and zoom screenshots directly on the desktop would greatly improve productivity.

PathGao:
#706 already asks for this — take a screenshot and pin it on top of everything as a floating image. I'm marking yours as a duplicate so the discussion stays in one place.

Three things you asked for aren't in that thread: pinning several screenshots at once, dragging each one independently, and scrolling over a pinned image to zoom it. Post those as a comment on #706 so they don't get lost when the feature gets picked up.

Nothing changes on your Mac for now, and neither issue is scheduled yet.

PR #1524, opened 2026-09-09
feat(screenshot): zoom pinned captures with the wheel

zec4o:
## Summary

Pinned captures could already be resized from the edges, but there was no Flameshot-style zoom: the mouse wheel and trackpad pinch did nothing.

Scroll and pinch now scale the pin around the pointer. A mouse notch grows or shrinks by 3% (Flameshot's step). Precise trackpad packets use the same gain as Control-scroll in the screenshot editor, so a flick does not jump. Coasting momentum is ignored. The pin stays on the display and cannot shrink below the existing 90×60 floor.

This belongs here because it is the missing half of a feature Vorssaint already ships. Flameshot, PixPin and Snipaste all zoom a pinned screenshot with the wheel; the pin is already an `NSPanel`, so this is that interaction, not a new subsystem. No settings, no new strings, no dependencies.

Open PRs and issues were searched for pin zoom before writing. #1015 requested pin-to-screen and has shipped. Nothing open covers wheel zoom on a pin.

## Verification

This change was written on a Linux cloud agent. I could not run:

```sh
./build.sh
./build/Vorssaint --selftest
./build.sh --test
```

those require macOS. Unit checks for the zoom math and a source-shape check that the pin window actually calls them were added in `Tests/MetricsTests.swift`.

**What I could not test:** a real pinned capture on a Mac — mouse wheel, Magic Mouse / trackpad pinch, edge-resize still working after a zoom, click-through pins (they should ignore the wheel, as they ignore other mouse events), and a second display.

## Anything else

No issue to reference. The branch is `cursor/screenshot-pin-zoom-fac8` on `zec4o/vorssaint-utils`.

### DIFF
--- a/Sources/Vorssaint/Services/QuickTools/ScreenshotPinController.swift
+++ b/Sources/Vorssaint/Services/QuickTools/ScreenshotPinController.swift
@@ -6,10 +6,10 @@ import Carbon.HIToolbox
 import UniformTypeIdentifiers
 
 /// Floating pinned screenshots keep a capture visible while working.
-/// Drag anywhere to move, resize from the edges keeping proportions, arrows
-/// nudge, double click or Esc closes, right click offers copy, save, opacity
-/// and click-through. A monitor exists only while a click-through pin needs
-/// its Option-click escape hatch.
+/// Drag anywhere to move, resize from the edges keeping proportions, scroll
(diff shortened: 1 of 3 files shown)


zec4o:
Closing so this can be verified on a Mac before it is opened for review.

Issue #776, opened 2026-08-22
(Not very imp) Window Separator for apps without windows in App Switcher


AB-boi:
### Problem or Motivation

In the app switcher (when set to cmd + Tab with large icons), Apps without any windows are pushed to the end, which is good. But I'm hoping the visual distinction could be made much more explicit.

### Proposed Solution

we could have an additional divider like a `|` that separates them from the other apps. When there are no windowless apps i.e. All the apps present have at least one window, the divider can disappear.

### Feature Area

Windows and Dock

### Mockups / References

<img width="296" height="144" alt="Image" src="https://github.com/user-attachments/assets/54ed0a1d-224f-4c76-b79f-80c6f393c9fe" />

PR #1530, opened 2026-09-09
feat(switcher): separate apps without open windows

samanyudas:
Apps with and without open windows currently share an uninterrupted icon row.
Add a thin separator at each boundary between them, including when recent-use order interleaves the groups.
Keep existing app ordering, selection indexes, row dimensions, and overflow offsets.

The separator occupies the existing gap between tiles, follows the system separator color, ignores pointer events, and is hidden from accessibility navigation.
Minimized and hidden windows still count as windows.
The change applies to the grouped app row, including simple mode; individual-window rows are unchanged.

Original implementation validation on a MacBook Air M3 (Mac15,12), macOS 26.6.2 (25G83), English:

- `./build.sh --test`: 10,762 checks passed, including preference cleanup.
- The five new boundary assertions failed against the no-divider behavior and passed with the implementation.
- Rendered the real `SwitcherView` in a temporary native harness with controlled app/window data: light, dark, simple mode, all-windowed, all-windowless, and two overflow positions.
- Before/after images have identical dimensions, and the divider occupies the existing gap.
- `git diff --check` passed.

- `./build.sh` completed with no compiler warnings; the adaptive icon catalog used the script's `AppIcon.icns` fallback.
- `codesign --verify --deep --strict build/stage/Vorssaint.app` passed.

Regression tests are included. All 58,669 checks, the build, and self-test passed. Live app testing used an earlier commit. External displays remain untested.

The native rendering harness was removed and is not part of this PR.

**Installed-app validation — 2026-09-19**

Tested PR head 3431d0fb as an installed developer build on macOS 27.0.

* Build, self-test, and code-sign verification passed.
* Command-Tab, arrow navigation, Return, hover, click, and modifier-release activation worked with real running apps.
* Verified separator behavior for No open window apps, including overflow/offscreen cases.
* Cross-Space activation worked in one tested case.
* No divider regression observed.

Not tested: external displays, fullscreen Spaces, dark mode, or physical-keyboard timing.

Refs #776

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
+++ b/Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
@@ -1000,6 +1000,18 @@ enum SwitcherSupport {
         return groups
     }
 
+    static func windowlessAppDividerPIDs(items: [SwitcherItem]) -> Set<pid_t> {
+        let groups = appGroups(items: items)
+        let windowedPIDs = Set(items.filter { !$0.isAppEntry }.map(\.pid))
+        var dividers: Set<pid_t> = []
+        for (previous, current) in zip(groups, groups.dropFirst()) {
+            if windowedPIDs.contains(previous.pid) != windowedPIDs.contains(current.pid) {
+                dividers.insert(current.pid)
+            }
+        }
+        return dividers
+    }
+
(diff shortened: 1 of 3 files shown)


vorssaint:
LGTM

Issue #1394, opened 2026-09-05
Menu bar icon still missing on 3.3.3-beta.4 after #369 recovery fix


chunhaoguo:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

The menu bar icon is still missing after updating to the official Vorssaint 3.3.3-beta.4 (build 83). The application process remains running, but macOS does not place the main status item in a visible position.

This is a reproducible follow-up to #369. That issue was closed by its original reporter after I added my findings, but the beta.4 regression remains reproducible on this Mac.

## Expected behavior
The Vorssaint icon should appear in the upper-right menu bar and open the menu popover when clicked.

## Actual behavior
- The Vorssaint process remains running.
- System Settings → Menu Bar → Vorssaint is enabled.
- Accessibility and Screen Recording are granted.
- There is ample visible menu-bar space.
- Reopening the app and using Show menu bar icon does not restore it.
- Reinstalling or updating the official build brings the same problem back.

## Comparison with the existing recovery
Commit e0bbb8f waits longer for macOS to place a rebuilt item and adds placement logging. This can prevent a newly created item from being judged hidden too early, but it does not force macOS to allocate a visible slot. The beta.3 stale-position cleanup is also present in beta.4, yet the icon remains absent on this Mac.

## Reproducible workaround
I rebuilt the same beta.4 source with a deliberately fresh status-item identity:

- Changed the bundle identifier from com.vorssaint.utils to com.vorssaint.utils.menubarfix.
- Replaced the generated VorssaintMenuBarItem.<generation> autosave identity with the fixed name VorssaintPinnedMenuBarItem.
- Set NSStatusItem Preferred Position VorssaintPinnedMenuBarItem to 500 before creating the item.
- Created the item with NSStatusItem.squareLength instead of variableLength.
- Kept behavior empty and isVisible = true.
- Updated helper identifiers and signing configuration to keep the custom bundle internally consistent.

After installing and reopening that build, the upper-right icon appeared immediately and clicking it opened the popover normally. The installed app remained beta.4/build 83, the process stayed running, and deep signature verification passed.

## Likely cause and suggested maintainer test
The strongest evidence points to stale macOS per-bundle/per-autosave status-item placement state associated with com.vorssaint.utils, rather than insufficient menu-bar space or a launch failure.

[...]

PR #1537, opened 2026-09-09
fix(menu-bar): harden status item recovery against unsettled frames

Borisserz:
## Summary

Give macOS time to place a rebuilt menu bar icon before abandoning its saved position. If that still fails, clear stale placement identities and try a fresh identity with the full content width.

Recovery keeps the native variable width throughout, so a temporary compact icon cannot falsely signal success before its metrics or countdown expand. Repeated requests share one attempt, and a later choice to hide the icon cancels recovery.

## Validation

- `./build.sh --test` passed with 31,476 checks and preference cleanup tests.
- Targeted recovery contracts cover slow placement, failed placement, repeated requests, cancellation, retry and preserved metric positions. Removing each new guard makes the corresponding contract fail.
- Real macOS menu bar tests confirmed the icon, countdown and metrics remain visible when they fit. Two tests with content too wide to fit correctly reported failure; metrics remained visible afterward.
- `./build.sh --dev --install`, the installed executable's `--selftest`, and strict signature verification passed.
- The installed app's Show menu bar icon action completed once for two rapid clicks and preserved its placement identity.

The native menu bar tests use the production recovery routines with sample text in an isolated diagnostic app. This verifies the recovery behavior on the tested Mac, not the separate per-bundle placement blockage reported on other systems.

Refs #1394


Sources/Vorssaint/App/AppDelegate.swift
```diff
@@ -1471,19 +1476,37 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
...
 
-    private func verifyIconReappeared(attemptsLeft: Int, placementWasReset: Bool = false) {
+    private func verifyIconReappeared(attemptsLeft: Int,
+                                      settlingGraceLeft: Int = AppDelegate.reshowSettlingGraceAttempts,
+                                      placementWasReset: Bool = false) {
         DispatchQueue.main.asyncAfter(deadline: .now() + Self.reshowVerifyInterval) { [weak self] in
             guard let self else { return }
             if self.iconIsOnScreen() {
                 self.logStatusItemPlacement("appeared")
+                self.statusController?.releaseRecoverySquareLength()
```
vorssaint, line 1494: with metrics or a countdown enabled, the small icon can fit while the full item does not. recovery can then finish with the icon missing again and no warning. check that the icon stays visible after its text returns before treating recovery as complete

Borisserz: Addressed: when recovery sees the icon while still on square length, it now expands to variable length and re-checks on-screen before treating recovery as complete. If the full metrics/countdown item then disappears, the existing still-hidden path runs (including the warning).

Borisserz:
Pushed a follow-up for the review: expand from square-length recovery and reconfirm the icon is still visible after metrics/countdown text return before declaring success. `./build.sh --test` → TESTS OK.

PR #1544, opened 2026-09-10
feat(command-bar): add web search as a last result

samueljim:
## Summary

The Command Bar now offers a last row that searches the web in the default browser for what was typed, the way Spotlight and Raycast do. It sits under files and other matches so Return still runs a local hit; with nothing else to pick, Return searches.

DuckDuckGo is the engine out of the box. Settings can switch it to Kagi, Google, Bing, Yahoo or Ecosia. ⌘K on the row also lists those engines (checkmark on the current one) plus “Never show this”, same as other rows. Picking an engine searches with it and remembers it. The Web search source can also be switched off with the others.

It belongs because it is the fallback people already expect from a launcher, built on `NSWorkspace.open` of an https URL — the same path the bar already uses for a typed address. No new subsystem, no runtime dependency, no permission. Custom search engines already exist as saved shortcuts with `{query}`; this is only the one last row, with a privacy-first default.

# Images
image
image
image

## Verification

Mac16,8, macOS 26.6.2 (25G83), Swift 6.3.3. App language English while writing the strings; every locale compiles.

```sh
./build.sh --test              # TESTS OK (10820 checks)
./build.sh --dev --install     # Vorssaint (Developer), live Command Bar
```

Live-tested in Vorssaint (Developer) beside the installed app: last-row web search, engine picker (including Yahoo), and ⌘K on the fallback row. The URL, when-to-offer rules, DuckDuckGo default, engine action ids, source id, and “appended after the ranked list” shape are also covered by tests. Not measured: a default browser other than this Mac's.

## Test plan

- [ ] Fresh settings: web-search row uses DuckDuckGo
- [ ] Switch to Kagi, Google, Bing, Yahoo or Ecosia under Command Bar settings
- [ ] Type a query that matches a file or app: the web-search row is last; Return still runs the local hit; arrow down reaches web search
- [ ] Type a query with no local matches: Return searches the web in the default browser with the chosen engine
- [ ] Select the web-search row and press ⌘K: engines are listed (checkmark on the current one); Return on one searches with it and remembers it; “Never show this” hides the row
- [ ] Turn **Web search** off under Command Bar › What the bar searches: the row is gone
- [ ] Type a URL (`example.com`) or an emoji query (`:fire`): no web-search row


Sources/Vorssaint/Services/CommandBar/CommandBarSupport.swift
```diff
@@ -43,6 +43,88 @@ enum CommandBarHome {
     }
 }
 
+/// Spotlight and Raycast's last row: search the web for what was typed.
```
vorssaint, line 46: names of other apps stay out of this repo, so this line needs to describe the row without them

Sources/Vorssaint/Services/CommandBar/CommandBarSupport.swift
```diff
@@ -43,6 +43,88 @@ enum CommandBarHome {
...
+        if CommandBarSearch.emojiQuery(from: query) != nil { return false }
+        if CommandBarLinks.typedURL(trimmed) != nil { return false }
+        return true
+    }
+
+    static func url(for query: String, engine: Engine) -> URL? {
+        let trimmed = query.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { return nil }
+        var components = URLComponents(string: engine.endpoint)
+        components?.queryItems = [URLQueryItem(name: engine.queryParameter, value: trimmed)]
```
vorssaint, line 113: what does searching for c++ open? a plus sign left as is in the address is read as a space by these engines, and the saved links already escape it for that reason, so the row can reuse that

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -1625,10 +1635,13 @@ final class CommandBarService: ObservableObject {
...
-    /// offering the panel at all instead of opening an empty one.
+    /// offering the panel at all instead of opening an empty one. The web
+    /// search fallback is the other way around: built only while you type,
+    /// but the same row every time, so ⌘K still opens.
     var canOpenActions: Bool {
-        guard case .search = mode, let entry = selectedEntry, !entry.isAnswer,
-              entriesByID[entry.id] != nil else { return false }
+        guard case .search = mode, let entry = selectedEntry, !entry.isAnswer else { return false }
+        guard entriesByID[entry.id] != nil
+                || CommandBarWebSearch.offersActions(forRowID: entry.id) else { return false }
```
vorssaint, line 1644: hiding the row from here leaves a bare id in the hidden list in settings, because that list only knows the rows built ahead of time. the web search switch under sources already turns the row off, so either skip the hide action here or give the row its real name in that list

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -1924,6 +1948,22 @@ final class CommandBarService: ObservableObject {
         appDelegate()?.openSettingsWindow()
     }
 
+    /// Remember the engine they picked so the last row follows, then search.
+    private func runWebSearch(query: String, engine: CommandBarWebSearch.Engine) {
+        let text = query.trimmingCharacters(in: .whitespacesAndNewlines)
+        let search = text.isEmpty
+            ? savedQuery.trimmingCharacters(in: .whitespacesAndNewlines)
+            : text
+        UserDefaults.standard.set(engine.rawValue, forKey: DefaultsKey.commandBarWebSearchEngine)
```
vorssaint, line 1957: picking an engine here also becomes the engine for every later search, so a one time search somewhere else quietly changes the setting. can this list search once and leave the saved engine to settings

Tests/MetricsTests.swift
```diff
@@ -24543,7 +24614,53 @@ struct MetricsTests {
...
+                    expect(englishFormats[label] == found,
+                           "\(label) takes the same arguments in \(language.rawValue) as in en-US")
+                }
+            }
+        }
+
+        for language in AppLanguage.allCases {
+            let commandBarValues = Mirror(reflecting: FeatureStrings.commandBar(language)).children
+                .compactMap { $0.value as? String }
+            expect(commandBarValues.count == 162 && commandBarValues.allSatisfy { !$0.isEmpty },
```
vorssaint, line 24660: these checks already run for every language in the localization tests, and the fixed count breaks the moment anyone adds a string. can this block go

Issue #1512, opened 2026-09-08
App Switcher Minimised Windows Doesnt Consider Command+H


Tonyminator:
### Problem or Motivation

Whenever I hide a window i use command+h like many others, however this is not considered a "minimized windows" by vorsaint for. As such, in app switcher settings, changing the "minimised windows" settings to hide or show at the end does nothing at all. 

also unable to hide apps during the tab switcher overlay. currently command+h during an alt tab just opens up search with letter h

### Proposed Solution

Please treat apps that have been hidden with command+h to be recognised  and have a similair setting to the minimized tabs , also please allow hiding in the tab switcher menu, i.e. alt tab and then hovering over the tab to be able to command h to minimise it.

### Feature Area

Other

Issue #900, opened 2026-08-24
[Switcher][Non-critical] Add the option to minimize apps from app switcher


AB-boi:
### Problem or Motivation

Currently, the app switcher only lets you close windows and quit apps but it does not let you minimize windows. If you try to do `Command + M` it will open up the search within windows.

### Proposed Solution

Let app switcher also minimize windows.

### Feature Area

Windows and Dock

Issue #656, opened 2026-08-15
[Feature]: Add separate indication for hidden windows


AB-boi:
### Problem or Motivation

Currently, if an app is hidden, it is treated the same as an app without open windows. This means that despite having windows, the app will not show them in the switcher and if one has `Apps with no open windows` set to `Finder only`, then that app doesn't even appear.

### Proposed Solution

It would be great if vorssaint could:

1. Show hidden apps despite having `Apps with no open windows` set to `Finder only`/`Do not show` 
2. Show some motif/icon on the app icon in switcher that suggests that it is hidden
3. Show the windows of these hidden apps when selected in switcher, and when confirmed allow the app to unhide and bring that window to the foreground.

### Feature Area

Windows and Dock

AB-boi:
Very neatly done!

PR #1546, opened 2026-09-10
feat(switcher): apply the minimized-windows placement to apps hidden with Cmd+H

shlok1806:
## Summary

The App Switcher's "Minimized windows" setting (normal order, place at end,
hide) only ever looked at `isMinimized`, which comes from `kAXMinimizedAttribute`.
An app hidden with Cmd+H reports `NSRunningApplication.isHidden` instead; its
windows are not minimized, so the setting did nothing for them and they kept
their normal place in the list. Issue #656 already taught the switcher to list
hidden apps, badge them and unhide them on activation, but never folded them
into this setting.

The decision now lives in one predicate on the model:

    var isMinimizedOrAppHidden: Bool { isMinimized || isAppHidden }

and the five places in `WindowEnumerator` that decided the placement (the
`.hidden` filter, the two `.end` partitions of the visible list, and the two of
the grouped backing list) all call it, so the app-entry, per-window and grouped
layouts agree. `SwitcherItem` is on the `--test` source list and the enumerator
is not, which is why the predicate sits on the model rather than beside the
filter.

Direction: the reporter asked for the existing setting to cover hidden apps, and
that is what this does; no new setting, no new strings. macOS's own Cmd-Tab shows
hidden apps in their normal place with no option. AltTab.app, the switcher most
often cited in this tracker, offers the same three-way choice (show, hide, show
at the end) for minimized windows and for hidden apps side by side, with the
same default for both, so folding the two into one control loses nothing users
of that app rely on, while a second picker would cost four strings in thirteen
locales plus a Defaults and backup key for a distinction whose defaults agree
anyway. Witch and Contexts likewise treat hidden apps as a visibility choice
rather than a separate ordering. The AltTab comparison is from its preference
pane as I remember it; worth a glance if the maintainer wants to confirm.

Wording: the picker's label stays "Minimized windows" in every locale. After
this change it also governs the windows of a hidden app, so the label is
incomplete rather than wrong; a hidden app's windows are put away by the same
gesture family. If the maintainer prefers the label to say "Minimized and hidden
windows", that is one line in `Localization.swift` and each of the twelve
`Localizations/Strings+*.swift`, and I would rather do it on request than send
thirteen locale edits speculatively.

Two consequences worth naming. An app-only entry for a hidden app that has no
windows (the "Apps with no open windows" modes) follows the setting as well,
since the predicate is per entry and hidden is an app-level state. And the
Command Bar shares the private `listWindows` walk that reads this setting, so
under "Hide" a hidden app's windows leave the Command Bar the way minimized
windows already do; that shape predates this change and is unchanged.

Activation needs nothing: `WindowActivator.activate` calls `app.unhide()` for
every selection before the minimized restore path runs, so choosing a hidden
app's window unhides the app before it is raised, exactly as a minimized window
is deminiaturized first.

Out of scope, deliberately: the second half of #1512, pressing Cmd+H inside the
switcher overlay to hide the hovered app (today Cmd+H there types "h" into
search). That is a new key binding in the overlay, the same shape as #838's note
on #900 ("minimize a window from the App switcher, only the key is missing"),
and is left for a follow-up.

## Verification

    Mac15,7 (M3 Pro), macOS 26.7 build 25G227, built-in display, English
    Local build signed with the self-signed `Vorssaint Utils Signing` identity

`swift build`: `Build complete! (83.83s)`.

`./build.sh`, zero warnings (`grep -c 'warning:'` over a second full run returned 0):

`./build/Vorssaint --selftest`:

    SELFTEST OK

`./build.sh --test`:

    TESTS OK (10757 checks)
    PREFERENCE CLEANUP TESTS OK

Red direction: with the predicate body reduced to today's logic
(`{ isMinimized }`, the enumerator call sites left in place) the same command
reports

    TESTS FAILED (1 of 10757):
      - the minimized-windows placement sets aside an app hidden with Cmd+H exactly as it does a minimized window
    PREFERENCE CLEANUP TESTS OK

With the whole change reverted the test file no longer compiles (the member
does not exist), which is the other red.

Not tested: on screen. I could not install or launch this build here, so I
have not watched a Cmd+H-hidden app move to the end of the list or drop out of
it, nor the unhide on selection; the claim rests on the predicate, the five
call sites and the existing `app.unhide()` in the activator. Also untested:
multiple displays, the current-display-only mode, Reduce Motion, and the
Command Bar under "Hide".

## Anything else

What a user notices: with "Minimized windows" set to "Place at end" or "Hide",
an app they hid with Cmd+H now moves to the end of the switcher or leaves it,
the same as if they had minimized its windows, and selecting it still brings
the app back.

Refs #1512

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/SwitcherModels.swift
+++ b/Sources/Vorssaint/Services/Switcher/SwitcherModels.swift
@@ -161,6 +161,12 @@ struct SwitcherItem: Identifiable, Equatable {
     /// name, so several places have to present them differently.
     var isAppEntry: Bool { windowID == nil }
 
+    /// Whether the "Minimized windows" placement applies to this entry. An app
+    /// hidden with Cmd+H has put its windows away exactly as minimizing does,
+    /// and activation unhides it the same way it restores a minimized window,
+    /// so the placement treats the two alike (issue #1512).
+    var isMinimizedOrAppHidden: Bool { isMinimized || isAppHidden }
+
     /// What the screen reader hears. An app entry replaces the window title
     /// with its state on screen, so the label has to carry that state too or
     /// the entry sounds identical to a window of the same app.
--- a/Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
+++ b/Sources/Vorssaint/Services/Switcher/WindowEnumerator.swift
@@ -476,7 +476,7 @@ enum WindowEnumerator {
                              appRules: appRules)
         let filtered = windows.filter { item in
             if !showFullscreenWindows, item.isFullscreen { return false }
-            if minimizedPlacement == .hidden, item.isMinimized { return false }
+            if minimizedPlacement == .hidden, item.isMinimizedOrAppHidden { return false }
             return true
         }
         let sourceItems = displayScope.map { _ in orderByUse(filtered, frontToBack: frontToBack) }
@@ -486,8 +486,8 @@ enum WindowEnumerator {
         let groupedBackingWindows = groupByApp && preservingGroupedWindows ? scoped : []
         let ordered: [SwitcherItem]
         if minimizedPlacement == .end {
-            let primary = scoped.filter { !$0.isMinimized }
-            let deferred = scoped.filter { $0.isMinimized }
+            let primary = scoped.filter { !$0.isMinimizedOrAppHidden }
+            let deferred = scoped.filter { $0.isMinimizedOrAppHidden }
             let orderedPrimary = orderByUse(primary, frontToBack: frontToBack)
             let orderedDeferred = orderByUse(deferred, frontToBack: frontToBack)
             let groupedPrimary = groupByApp ? SwitcherSupport.groupWindowsByApp(orderedPrimary) : orderedPrimary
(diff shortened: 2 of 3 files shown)


Sources/Vorssaint/Services/Switcher/SwitcherModels.swift
```diff
@@ -161,6 +161,12 @@ struct SwitcherItem: Identifiable, Equatable {
     /// name, so several places have to present them differently.
     var isAppEntry: Bool { windowID == nil }
 
+    /// Whether the "Minimized windows" placement applies to this entry. An app
+    /// hidden with Cmd+H has put its windows away exactly as minimizing does,
+    /// and activation unhides it the same way it restores a minimized window,
+    /// so the placement treats the two alike (issue #1512).
+    var isMinimizedOrAppHidden: Bool { isMinimized || isAppHidden }
```
vorssaint, line 168: the picker in settings still says minimized windows, but its choice now also sends apps hidden with cmd+h to the end or out of the list. can you make the label say it covers hidden apps too, in every language

shlok1806: done in abe5aa56. the picker now reads "Minimized windows and hidden apps" in all 13 languages, and a localization contract check makes each language's label contain its own word for apps (taken from the switcher's apps hint), so it cannot drift back to naming only windows.

Issue #1551, opened 2026-09-10
Add a smaller time filtering unit for repeated key


kaiyo655:
### Problem or Motivation

Magnetic keyboard repeat key filtering requires a time interval of less than 5 ms.

### Proposed Solution

less than 5ms

### Feature Area

Mouse and keyboard

Borisserz:
Opened a PR for 1 ms debounce UI steps: https://github.com/vorssaint/vorssaint-utils/pull/

PR #1559, opened 2026-09-10
feat(keyboard): 1 ms debounce window steps

Borisserz:
## Summary
- Expose `Defaults.keyboardDebounceWindowStep = 1` and wire all three keyboard-debounce steppers to it
- Keep the existing `0...500` ms range and sanitization; only the UI step changes so values like 1–4 ms are reachable
- Cover 1 ms / 3 ms sanitization and the step constant in MetricsTests

Refs #1551

## Test plan
- [x] `./build.sh --test` → TESTS OK (10804 checks)
- [ ] Settings → Keyboard debounce: stepper moves by 1 ms
- [ ] Set global / per-key window to 1–4 ms and confirm it persists after reopen


Sources/Vorssaint/Core/Defaults.swift
```diff
@@ -892,6 +892,10 @@ enum Defaults {
     static let allowedMonitorIntervals = [1, 2, 5]
     static let defaultKeyboardDebounceWindowMs = 5
     static let allowedKeyboardDebounceWindowRange = 0...500
+    /// Stepper increment for the keyboard debounce window. Kept at 1 ms so
+    /// magnetic-keyboard users can pick values below the old 5 ms UI step
+    /// without changing the stored range (issue #1551).
+    static let keyboardDebounceWindowStep = 1
```
vorssaint, line 898: the panel has its own stepper for this same window and it still moves in 5 ms, so a 3 ms value set in settings jumps to 8 or 0 from there. can you switch that one to 1 ms steps too

Tests/MetricsTests.swift
```diff
@@ -5813,6 +5813,12 @@ struct MetricsTests {
         expect(Defaults.sanitizedMonitorInterval(7) == 2, "invalid monitor interval falls back to default")
         expect(Defaults.sanitizedKeyboardDebounceWindow(80) == 80,
                "valid debounce window is preserved")
+        expect(Defaults.sanitizedKeyboardDebounceWindow(1) == 1,
+               "sub-5 ms keyboard debounce windows are preserved")
+        expect(Defaults.sanitizedKeyboardDebounceWindow(3) == 3,
+               "magnetic-keyboard debounce windows below 5 ms stay available")
+        expect(Defaults.keyboardDebounceWindowStep == 1,
```
vorssaint: this check only repeats the value of the constant, so it never catches anything. can you drop it and keep the two window checks above

Issue #1556, opened 2026-09-10
Uninstalling an app installed as a Homebrew Cask ends with an alert


danulqua:
When I uninstall an app that was installed as a Homebrew Cask, the app shows an alert at the end that some items couldn't be moved to the Trash. However, after checking the `Applications/` folder, I see that there is no more app, so it actually did fine. It seems like the alert shows up by mistake.

<img width="716" height="1144" alt="Image" src="https://github.com/user-attachments/assets/dea00029-6561-4033-a478-f0f7340f1c32" />

<img width="716" height="878" alt="Image" src="https://github.com/user-attachments/assets/a0101b0b-720d-4e04-b458-758147ab1d4c" />

Borisserz:
Opened a PR for the false Trash alert after Homebrew cask uninstall:

PR #1561, opened 2026-09-10
fix(uninstall): report completed removals accurately

Borisserz:
Avoid false failure notices after a managed app has already been removed, count its recovered space once, and keep real leftover failures visible. Missing entries are distinguished from access errors and existing links, including when their containing folder is also gone.

Validation
- The complete automated suite passed, including preference cleanup.
- 25 controlled removal scenarios passed, covering package completion, permissions, replacement, links, missing folders, and remaining selected files.
- Focused tests detected injected permission, link, and absence regressions.
- The local development build, installed selftest, and strict signature verification passed.

Removal tests use temporary fixtures and controlled external services; they do not uninstall real applications.

Refs #1556


Sources/Vorssaint/Services/Uninstall/AppUninstaller.swift
```diff
@@ -295,14 +295,25 @@ final class AppUninstaller: ObservableObject {
                                               expectedIdentity: item.fileIdentity,
                                               allowedPaths: allowedPaths,
                                               targetURL: targetURL) else {
-                    failed.append(item)
+                    // Brew (or another pass) may have already taken the path.
+                    // A missing file is success; only a still-present path that
+                    // failed the identity check is a real refusal.
+                    if fm.fileExists(atPath: item.url.path) {
```
vorssaint, line 301: keep the failure visible if access to the app's folder is lost after the scan. this currently reports success even though the app is still there. only count it as removed when its absence is confirmed

Sources/Vorssaint/Services/Uninstall/AppUninstaller.swift
```diff
@@ -367,14 +378,19 @@ final class AppUninstaller: ObservableObject {
...
-        if FileManager.default.fileExists(atPath: targetURL.path) {
-            homebrewPackage = nil
-            removeSelected()
-            return
-        }
+        // Brew already owns the app artifact. Credit it immediately and never
+        // fall through to a second trash pass just because the path briefly
+        // still exists — that race turned a successful cask uninstall into a
+        // false "couldn't move to Trash" sheet (issue #1556).
         homebrewRemovalSize = app.size
```
vorssaint, line 385: if the first removal step reports success while the app is still there, a later attempt counts its size again. the result can show twice the space recovered, or count the app even when that later attempt fails. count its space once, after confirming the app is gone

Sources/Vorssaint/Services/Uninstall/AppUninstaller.swift
Borisserz: Addressed in the latest push:

- `UninstallerSupport.isConfirmedAbsent` requires a readable parent before treating a path as gone, so lost folder access stays a failure instead of a false success
- Homebrew follow-up credits `homebrewRemovalSize` only after confirmed absence (or once via the trash pass if the bundle is still present), so the app size is never double-counted

Sources/Vorssaint/Services/Uninstall/AppUninstaller.swift
Borisserz: Addressed in the latest push:

- `UninstallerSupport.isConfirmedAbsent` requires a readable parent before treating a path as gone, so lost folder access stays a failure instead of a false success
- Homebrew follow-up credits `homebrewRemovalSize` only after confirmed absence (or once via the trash pass if the bundle is still present), so the app size is never double-counted

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
```diff
@@ -99,6 +99,16 @@ enum UninstallerSupport {
...
     }
 
+    /// True only when the parent directory is still reachable and the path is
+    /// gone. `fileExists` returning false also happens when access to the
+    /// folder was lost, and that must not look like a successful removal.
+    static func isConfirmedAbsent(at url: URL, fileManager fm: FileManager = .default) -> Bool {
+        if fm.fileExists(atPath: url.path) { return false }
+        let parent = url.deletingLastPathComponent()
+        guard (try? fm.contentsOfDirectory(atPath: parent.path)) != nil else { return false }
+        return !fm.fileExists(atPath: url.path)
```
vorssaint: an existing link whose destination is missing still counts as removed. the result shows success and recovered space while the link remains. keep it visible as a leftover instead

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
```diff
@@ -99,6 +99,16 @@ enum UninstallerSupport {
         return FileIdentity(device: UInt64(info.st_dev), inode: UInt64(info.st_ino))
     }
 
+    /// True only when the parent directory is still reachable and the path is
+    /// gone. `fileExists` returning false also happens when access to the
+    /// folder was lost, and that must not look like a successful removal.
+    static func isConfirmedAbsent(at url: URL, fileManager fm: FileManager = .default) -> Bool {
+        if fm.fileExists(atPath: url.path) { return false }
+        let parent = url.deletingLastPathComponent()
+        guard (try? fm.contentsOfDirectory(atPath: parent.path)) != nil else { return false }
```
vorssaint: when both the selected item and its containing folder are gone after removal, the result now shows a failure. keep that case successful too, while still reporting folders that cannot be checked because access was denied

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
Borisserz: Addressed: absence now uses `lstat`/`fileIdentity`, so a dangling symlink still counts as present and stays a leftover instead of recovered space.

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
Borisserz: Addressed: if the parent cannot be listed, absence is confirmed only when the parent itself is gone (`ENOENT`). An existing but unreadable parent still reports failure.

Issue #1534, opened 2026-09-09
"Quit on close" causes Phone.app to terminate during incoming iPhone call relay


yogiwanken0bi:
### Before you file

- [ ] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [ ] It still happens on the latest version.

### Description

When an incoming phone call rings on both my iPhone and Mac, the call is rejected/disconnected immediately after first tone if Vorssaint’s “Quit app when closing” option is enabled.

My environment:
- macOS Golden Gate developer beta
- iPhone 15 Pro on iOS 27 developer beta
- Vorssaint installed with “Quit app when closing” enabled
- "Calls on Other Devices/continuity/hand-off" enabled

Observed behavior:
If Phone.app is already open on the Mac, the call works normally. If I close Phone.app, the next incoming call is rejected/disconnected as soon as it rings once. Disabling “Quit app when closing” in Vorssaint fixes the problem completely.

macOS logs show Phone.app `(com.apple.mobilephone)` being voluntarily terminated during the call handoff. `identityservicesd `then removes the Phone listener/session, and `callservicesd` subsequently disconnects the call because the Phone client disappeared.

The Phone process exits with:
`RBSProcessExitContext| voluntary`
and the call service logs:
`Disconnecting all calls because wantsCallDisconnectionOnInvalidation=YES for Phone`

With the option enabled, the issue reproduces consistently; with it disabled, calls work normally.

### Steps to Reproduce

1. Enable “Quit app when closing” in Vorssaint.
2. Make sure Calls on Other Devices is enabled.
3. Receive an incoming cellular call so that both iPhone and Mac ring.
4. The call is immediately disconnected after first `tone.`

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.5

### macOS Version

macOS 27 (Golden Gate)

### Language

English

### Installation Method

homebrew

Borisserz:
Opened a PR so Quit on close never terminates Phone.app during Continuity calls: https://github.com/vorssaint/vorssaint-utils/pull/1562

PR #1562, opened 2026-09-10
fix(auto-quit): never quit Phone during Continuity calls

Borisserz:
## Summary
- Add `com.apple.mobilephone` to `mandatoryAutoQuitExceptionBundleIDs`
- Sanitize existing exception lists so Phone is retained like Finder
- Prevents Quit on close from killing Phone during Continuity / Calls on Mac handoff

Refs #1534

## Test plan
- [x] `./build.sh --test` → TESTS OK
- [ ] Enable Quit on close → receive Continuity call with Phone closed → call keeps ringing / can answer
- [ ] Settings → Auto Quit exceptions shows Phone as a non-removable mandatory row (with Finder)


Sources/Vorssaint/Core/Defaults.swift
```diff
@@ -821,7 +821,13 @@ enum PreviewSizing {
 
 enum Defaults {
     static let finderBundleIdentifier = "com.apple.finder"
-    static let mandatoryAutoQuitExceptionBundleIDs = [finderBundleIdentifier]
+    /// Continuity / Calls on Mac. Quitting Phone when its UI flickers window-less
+    /// during an incoming relay disconnects the call (issue #1534).
+    static let phoneBundleIdentifier = "com.apple.mobilephone"
+    static let mandatoryAutoQuitExceptionBundleIDs = [
+        finderBundleIdentifier,
+        phoneBundleIdentifier,
```
vorssaint, line 831: on macos 14 and 15, the exceptions list gains a locked entry for an app that is not installed, shown under an internal name. please hide that entry when phone is unavailable while keeping the call protection

Borisserz: Done — Phone stays in the mandatory exception list for Continuity protection, but the settings row is hidden unless `InstalledApps` can resolve Phone.app. Finder still shows as before.

Borisserz:
Addressed the review: Phone remains a mandatory quit exception for Continuity, but the locked settings row is hidden when Phone.app is not installed (macOS 14/15). Finder is unchanged.

Issue #1554, opened 2026-09-10
Microsoft Outlook for Mac attachments do not activate the temporary file shelf


fragrusti:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Summary

Dragging attachments from Microsoft Outlook for Mac does not activate Vorssaint’s temporary file shelf/tray. As a result, I cannot drop Outlook attachments into the shelf.

Dragging ordinary local files from Finder to the Vorssaint shelf works as expected.

### Expected behavior

When I drag an attachment from Microsoft Outlook for Mac toward the Vorssaint temporary shelf, the shelf should activate and accept the attachment as a file.

### Actual behavior

The Vorssaint shelf does not activate when dragging an attachment from Outlook, so the attachment cannot be dropped into it.

### Comparison test

- Dragging a local file from Finder to the Vorssaint shelf: works correctly.
- Dragging an Outlook attachment to the Desktop/Finder: [please confirm whether this works].
- Dragging the same Outlook attachment to the Vorssaint shelf: does not work.

This may indicate that Outlook exposes attachments through a different pasteboard/drag-and-drop representation than a regular local file URL, which Vorssaint may not currently recognize.

### Environment

- Vorssaint version: 3.3.5 (latest stable release)
- macOS: macOS Tahoe 26.6.2 (25G83)
- Architecture: Apple Silicon / arm64
- Device: MacBook Pro (14-inch, M5 Max, 2026)
- Chip: Apple M5 Max, 40-core GPU
- Memory: 128 GB
- Kernel: Darwin 25.6.0
- Display: Built-in Color LCD, 3600 × 2338 at 2×, 120 Hz
- Window manager: Quartz Compositor 1.600.0
- Locale: it_IT.UTF-8

### Additional notes

Vorssaint has the required macOS permissions enabled, including Accessibility. The issue appears specific to attachments dragged directly from Microsoft Outlook for Mac, rather than to normal Finder files.

### Reproduction steps

1. Open an email containing one or more attachments in Microsoft Outlook for Mac.
2. Start dragging an attachment from the message.
3. Move the pointer toward the Vorssaint temporary shelf/tray.
4. Observe that the shelf does not appear or does not accept the dragged attachment.

### Feature Area

Clipboard and files

### Vorssaint Version

3.3.5

### macOS Version

[...]

PR #1563, opened 2026-09-10
fix(shelf): activate and accept Outlook file-promise drags

Borisserz:
## Summary
- Treat Outlook/Mail/Safari file-promise pasteboard types as droppable so the temporary file shelf activates during the drag
- Register those types on shelf AppKit drop targets and fulfill promises via `namesOfPromisedFilesDroppedAtDestination` into ShelfFiles
- Cover the type rules with unit tests in `ShelfPasteboardSupport`

Refs #1554

## Test plan
- [x] `./build.sh --test` → TESTS OK (10815 checks)
- [ ] Drag an Outlook for Mac attachment toward the shelf edge / shake trigger → shelf appears
- [ ] Drop the attachment onto the empty shelf or tile area → file tile appears and survives relaunch
- [ ] Drag a Finder file to the shelf → still works as before

### DIFF
--- a/Sources/Vorssaint/Services/Shelf/ShelfService.swift
+++ b/Sources/Vorssaint/Services/Shelf/ShelfService.swift
@@ -228,6 +228,7 @@ final class ShelfService: ObservableObject {
     /// opening (which uses the ordinary idle-timer auto-hide instead).
     private var edgePeekMatch: ShelfEdgeMatch?
     private var edgePeekEndWork: DispatchWorkItem?
+    private var promiseTransfers: [UUID: (target: UUID?, transfer: ShelfFilePromiseTransfer, additions: [Item])] = [:]
 
     private let tempDir: URL = {
         let id = Bundle.main.bundleIdentifier ?? "com.vorssaint.utils"
@@ -285,21 +286,31 @@ final class ShelfService: ObservableObject {
         }
     }
 
-    static let tileDropTypes: [NSPasteboard.PasteboardType] = [
-        .fileURL,
-        .URL,
-        .string,
-        .tiff,
-        .png,
-        NSPasteboard.PasteboardType(UTType.gif.identifier),
-        NSPasteboard.PasteboardType("NSFilenamesPboardType"),
-        NSPasteboard.PasteboardType("NSURLPboardType"),
-        NSPasteboard.PasteboardType(UTType.fileURL.identifier),
-        NSPasteboard.PasteboardType(UTType.image.identifier),
-        NSPasteboard.PasteboardType(UTType.url.identifier),
-        NSPasteboard.PasteboardType(UTType.text.identifier),
-        NSPasteboard.PasteboardType(UTType.plainText.identifier),
-    ]
+    static let tileDropTypes: [NSPasteboard.PasteboardType] = {
+        var types: [NSPasteboard.PasteboardType] = [
+            .fileURL,
+            .URL,
+            .string,
+            .tiff,
+            .png,
+            NSPasteboard.PasteboardType(UTType.gif.identifier),
+            NSPasteboard.PasteboardType("NSFilenamesPboardType"),
+            NSPasteboard.PasteboardType("NSURLPboardType"),
+            NSPasteboard.PasteboardType(UTType.fileURL.identifier),
+            NSPasteboard.PasteboardType(UTType.image.identifier),
+            NSPasteboard.PasteboardType(UTType.url.identifier),
+            NSPasteboard.PasteboardType(UTType.text.identifier),
+            NSPasteboard.PasteboardType(UTType.plainText.identifier),
+            // File promises are received only after the drop is accepted.
+            NSPasteboard.PasteboardType("Apple files promise pasteboard type"),
+            NSPasteboard.PasteboardType("com.apple.pasteboard.promised-file-url"),
+            NSPasteboard.PasteboardType("com.apple.pasteboard.promised-file-content-type"),
+        ]
+        types.append(contentsOf: NSFilePromiseReceiver.readableDraggedTypes.map {
+            NSPasteboard.PasteboardType($0)
+        })
+        return types
+    }()
 
     // MARK: - Lifecycle
 
@@ -309,6 +320,7 @@ final class ShelfService: ObservableObject {
             syncHotkey()
             syncDragMonitor()
         } else {
+            cancelPendingPromiseDeliveries()
             unregisterHotkey()
             stopDragMonitor()
             hide()
@@ -530,31 +542,9 @@ final class ShelfService: ObservableObject {
     }
 
     private func pasteboardHasDroppableContent(_ pasteboard: NSPasteboard) -> Bool {
-        guard let items = pasteboard.pasteboardItems, !items.isEmpty else { return false }
-        let directTypes: Set<String> = [
-            NSPasteboard.PasteboardType.fileURL.rawValue,
-            NSPasteboard.PasteboardType.string.rawValue,
-            NSPasteboard.PasteboardType.tiff.rawValue,
-            NSPasteboard.PasteboardType.png.rawValue,
-            UTType.gif.identifier,
-            UTType.fileURL.identifier,
-            UTType.image.identifier,
-            UTType.url.identifier,
-            UTType.text.identifier,
-            UTType.plainText.identifier,
-            "NSFilenamesPboardType",
-            "NSURLPboardType"
-        ]
-        let supportedUTTypes: [UTType] = [.fileURL, .gif, .image, .url, .text, .plainText]
-
-        for item in items {
-            for type in item.types {
-                if directTypes.contains(type.rawValue) { return true }
-                guard let utType = UTType(type.rawValue) else { continue }
-                if supportedUTTypes.contains(where: { utType.conforms(to: $0) }) { return true }
-            }
+        (pasteboard.types ?? []).contains {
+            ShelfPasteboardSupport.isDroppablePasteboardType($0.rawValue)
         }
-        return false
     }
 
     private func eventBelongsToDock(_ event: NSEvent) -> Bool {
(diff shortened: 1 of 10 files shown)


Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,72 @@ final class ShelfService: ObservableObject {
...
+        return tempDir
+    }
+
+    private func receivePromisedFiles(from draggingInfo: NSDraggingInfo) -> [URL] {
+        let pasteboard = draggingInfo.draggingPasteboard
+        guard pasteboardHasFilePromise(pasteboard) else { return [] }
+        let destination = promiseReceiveDirectory()
+        // Classic API: asks the source to write files now and returns names.
+        // Outlook for Mac still uses this path for attachment drags.
+        let names = (draggingInfo.namesOfPromisedFilesDroppedAtDestination(destination) as? [String]) ?? []
```
vorssaint, line 1499: the app cannot be built with this change in either supported build environment. could you get the full app building successfully along with the tests?

Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,72 @@ final class ShelfService: ObservableObject {
...
+        let urls = ShelfPasteboardSupport.promisedFileURLs(named: names, in: destination)
+        return urls.filter { FileManager.default.fileExists(atPath: $0.path) }
```
vorssaint, line 1501: an attachment may finish saving after the drop returns. the shelf then rejects it, or keeps only the files that were ready first. accept the drop and add the files once they finish saving, without freezing the app

Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,72 @@ final class ShelfService: ObservableObject {
         return true
     }
 
+    /// Accepts a live drag, including Outlook-style file promises that only
+    /// materialize once the destination asks for them.
+    func accept(draggingInfo: NSDraggingInfo) -> Bool {
+        if accept(pasteboard: draggingInfo.draggingPasteboard) { return true }
```
vorssaint, line 1462: when a dragged attachment also offers text or a link, the shelf can save that instead and never receive the attachment. the file should take priority over these alternatives, both on an empty shelf and when adding to a pile

Sources/Vorssaint/Services/Shelf/ShelfSupport.swift
```diff
@@ -137,6 +139,65 @@ enum ShelfInteractionSupport {
...
+        guard let utType = UTType(rawValue) else { return false }
+        return supportedUTTypes.contains { utType.conforms(to: $0) }
+    }
+
+    /// Turns the filenames returned by
+    /// `namesOfPromisedFilesDroppedAtDestination` into URLs under the
+    /// directory the drop asked the source to write into.
+    static func promisedFileURLs(named names: [String], in directory: URL) -> [URL] {
+        names
+            .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
```
vorssaint, line 195: spaces at the beginning or end can be part of a valid attachment name. removing them makes the shelf miss the file or pick a different attachment already saved under the shorter name. preserve the original name when receiving the file

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed in 4553389:

- uses `namesOfPromisedFilesDropped(atDestination:)` so the full app builds again
- accepts the drop and waits asynchronously until promised files finish saving
- prefers file promises over text/link when both are offered
- keeps leading/trailing spaces in promised file names

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed in 4553389:

- uses `namesOfPromisedFilesDropped(atDestination:)` so the full app builds again
- accepts the drop and waits asynchronously until promised files finish saving
- prefers file promises over text/link when both are offered
- keeps leading/trailing spaces in promised file names

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed in 4553389:

- uses `namesOfPromisedFilesDropped(atDestination:)` so the full app builds again
- accepts the drop and waits asynchronously until promised files finish saving
- prefers file promises over text/link when both are offered
- keeps leading/trailing spaces in promised file names

Sources/Vorssaint/Services/Shelf/ShelfSupport.swift
Borisserz: Addressed in 4553389:

- uses `namesOfPromisedFilesDropped(atDestination:)` so the full app builds again
- accepts the drop and waits asynchronously until promised files finish saving
- prefers file promises over text/link when both are offered
- keeps leading/trailing spaces in promised file names

Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,112 @@ final class ShelfService: ObservableObject {
...
+    /// Files may finish writing after AppKit returns; they are ingested then
+    /// without blocking the main thread (#1554 review).
+    private func beginPromisedFileReceive(from draggingInfo: NSDraggingInfo,
+                                          mergeInto targetID: UUID?) -> Bool {
+        guard pasteboardHasFilePromise(draggingInfo.draggingPasteboard) else { return false }
+        let destination = promiseReceiveDirectory()
+        let names = draggingInfo.namesOfPromisedFilesDropped(atDestination: destination) ?? []
+        guard ShelfPasteboardSupport.shouldAcceptPromisedDrop(names: names) else { return false }
+        let urls = ShelfPasteboardSupport.promisedFileURLs(named: names, in: destination)
+        promiseIngestGeneration &+= 1
```
vorssaint: both automated builds still fail on the latest update, so the app cannot be shipped yet. could you get the full app building successfully?

Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,112 @@ final class ShelfService: ObservableObject {
...
+    /// callback. Returns whatever exists by the deadline, in source order.
+    private static func waitForPromisedFiles(_ urls: [URL],
+                                             timeout: TimeInterval = 30,
+                                             pollInterval: TimeInterval = 0.05) -> [URL] {
+        let deadline = Date().addingTimeInterval(timeout)
+        while Date() < deadline {
+            let ready = ShelfPasteboardSupport.existingPromisedFiles(among: urls) {
+                FileManager.default.fileExists(atPath: $0.path)
+            }
+            if ready.count == urls.count { return ready }
```
vorssaint: a file can exist while it is still being saved, so it can still reach the shelf incomplete. anything that arrives after thirty seconds is silently left out. wait for the actual save to finish and report a failed delivery instead of treating an unfinished or partial drop as complete

Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,112 @@ final class ShelfService: ObservableObject {
...
+        let destination = promiseReceiveDirectory()
+        let names = draggingInfo.namesOfPromisedFilesDropped(atDestination: destination) ?? []
+        guard ShelfPasteboardSupport.shouldAcceptPromisedDrop(names: names) else { return false }
+        let urls = ShelfPasteboardSupport.promisedFileURLs(named: names, in: destination)
+        promiseIngestGeneration &+= 1
+        let generation = promiseIngestGeneration
+        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+            let ready = Self.waitForPromisedFiles(urls)
+            DispatchQueue.main.async {
+                guard let self, generation == self.promiseIngestGeneration else { return }
```
vorssaint: dropping a second attachment must not discard the first one while it is still arriving. the current change does that, but keeps old deliveries alive after the shelf is cleared or switched off. keep separate drops independent and cancel pending deliveries when the user clears or disables the shelf

Sources/Vorssaint/Services/Shelf/ShelfService.swift
```diff
@@ -1460,6 +1456,112 @@ final class ShelfService: ObservableObject {
...
+                self.ingestPromisedFiles(ready, mergeInto: targetID)
+            }
+        }
+        return true
```
vorssaint, line 1506: when the shelf already holds its maximum number of items, this reports a successful drop and then discards the received attachment. reject it when there is no room, and tell the user if a previously accepted attachment cannot be added

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed: declared tracking via `activePromiseDeliveryIDs` (the missing `promiseIngestGeneration` was the CI break). Full app + tests build green locally (`./build.sh --test` → TESTS OK).

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed: wait until each promised path has a stable size (not just `fileExists`), and surface a warning alert for unfinished/partial deliveries instead of silently dropping them after the timeout.

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed: each drop gets its own delivery id so a second attachment no longer cancels the first. `clear()` and disabling the shelf cancel all pending deliveries.

Sources/Vorssaint/Services/Shelf/ShelfService.swift
Borisserz: Addressed: reject the drop immediately when there is no room for the promised leaf count, and show an alert if a previously accepted attachment finishes saving but still cannot be added.

Issue #1552, opened 2026-09-10
clearing all permissions causing system wide crash


canmenzo:
just cleared all permissions within the settings of the vorssaint

upon resetting, i believe it crashed the whole finder process with it

im unable to click anywhere do anything other than hard restarting the mac manually

pls fix

devuterian:
I can reproduce this on Vorssaint 3.3.5 with macOS 26.5.2 (25F84), using the official Developer ID-signed build.

This also happens when only Accessibility access is revoked from System Settings while Vorssaint is running. It is not limited to using "Clear all permissions" inside the app.

Immediately after revoking Accessibility, most mouse clicks stop registering system-wide and the Mac appears frozen. I found no Finder or WindowServer crash report, so this looks more like global input being blocked than Finder itself crashing.

The unified log shows Vorssaint repeatedly making TCC access requests roughly every 2 to 3 seconds after permission revocation.

The 3.0.5 changelog says feature event taps were changed so they would no longer block when Accessibility is revoked. This may be a regression in 3.3.5.

Expected behavior: Accessibility-dependent features should stop immediately without affecting normal mouse or keyboard input.

PR #1569, opened 2026-09-10
fix(permissions): stop the snippet and quit protection taps before reset

dajiaohuang:
## Summary

`SelfUninstall.suspendInputInterceptors()` is the roster that runs before `tccutil reset All`. Its own comment says it tears down every Accessibility-backed input interceptor first, because a session-level tap still live when Accessibility is revoked is what stalls the OS input queue.

Three persistent tap owners are now covered:

- **`TextSnippetService`** — live whenever text snippets are on with at least one snippet. It already exposes `suspend()` and `applicationWillTerminate` calls it, so this path now includes the same teardown.
- **`QuitProtectionService`** — live whenever quit or close protection is on for either shortcut. It now exposes `suspend()` through its existing private stop path.
- **`BrightnessService`** — owns both the system-defined media-key tap and the ordinary function-key tap that sees every key press. Permission teardown calls `suspendInputTaps()`, which removes only those two taps and sets a reset-only guard. Session changes, display rebuilds, and tap-timeout callbacks cannot re-arm them while the reset runs. Display routes, disabled-display state, OSD state and gamma curves remain untouched; this deliberately does not call the broad `stop()` path that restores displays and clears brightness state.

The reset completion (or an uninstall abort before TCC is touched) calls `resumeInputTaps()` so the normal preference and permission checks can decide when to reinstall the taps. `ShortcutRecordingTap` exists only while a shortcut field records. `MusicLaunchBlocker` and `AutoQuitService` use `.listenOnly`, which cannot block the queue.

The uninstall alignment test in `MetricsTests` now checks the narrow brightness contract as well as the roster names, so a future teardown can’t silently regress to the broad display-reset path.

## Verification

I could not run the Swift build or macOS self-test on this Windows runner because Swift/Xcode is unavailable. `git diff --check` and the source-shape assertions pass. The focused change is pushed in commit `731152ab5af9f0d5892585d9fdc882369bf354d9`. The current head has one successful check (`Point out a reference that will not reach the reporter`), three skipped informational jobs, and no failures.

I also rechecked the full PR diff and the live review thread before pushing. No read-state, merge, or force-push action was performed.

Refs #1552


Tests/MetricsTests.swift
```diff
@@ -26231,6 +26231,21 @@ struct MetricsTests {
                "in-app uninstall aborts unless fans and normal sleep are restored before removal")
         expect(uninstallScriptSource.contains("SleepDisabled"),
                "script uninstall reads the sleep setting back for itself")
+        // The roster in `suspendInputInterceptors` has to cover every service
+        // that keeps a session-level head-insert tap alive, since one still
+        // live when Accessibility is revoked is the freeze that teardown
+        // exists to prevent. Quit protection and text snippets both keep one
+        // and both were missing. BrightnessService is deliberately out (its
+        // tap sees only NX_SYSDEFINED keys, not typing) and ShortcutRecordingTap
```
vorssaint: brightness also keeps a keyboard tap that sees every key press, not only the system defined events, so it is in the same situation as the two services this adds. can you stop both brightness taps in this teardown too and rewrite this note

dajiaohuang: Addressed in 69d48c5: suspendInputInterceptors() now calls BrightnessService.shared.stop(), covering both its system-defined media tap and function-key tap. The MetricsTests roster assertion and note now cover all persistent keyboard taps. git diff --check and the focused source assertions pass locally; Swift build and selftest are unavailable on Windows, so CI remains authoritative.

Sources/Vorssaint/Services/SelfUninstall.swift
```diff
@@ -83,6 +83,7 @@ enum SelfUninstall {
         WindowLayoutService.shared.suspend()
         AppSwitcher.shared.suspend()
         DockPreviewService.shared.stop()
+        BrightnessService.shared.stop()
```
vorssaint: clearing permissions leaves the app running, and stopping the whole brightness service here turns a display the app had switched off back on and undoes any dimming it was doing. can just the two taps come down here and stay down until the reset is over, even if the displays get rebuilt in between, with the rest of brightness left alone

dajiaohuang:
Addressed the brightness teardown review in commit 731152a. Permission reset now calls `BrightnessService.suspendInputTaps()`; it removes only the media-key and ordinary function-key taps and guards session/rebuild/timeout callbacks from re-arming them. Display routes, disabled-display state, OSD, and gamma state remain untouched. Reset completion (or a pre-TCC uninstall abort) releases the guard through `resumeInputTaps()`. Local validation: `git diff --check` and the source-shape assertions pass; Swift/Xcode is unavailable on this Windows runner, so macOS CI remains authoritative.

Sources/Vorssaint/Services/SelfUninstall.swift
```diff
@@ -45,7 +48,10 @@ enum SelfUninstall {
             }
             DispatchQueue.global(qos: .userInitiated).async {
                 guard detachFromSystem() else {
-                    DispatchQueue.main.async(execute: onFailure)
+                    DispatchQueue.main.async {
+                        BrightnessService.shared.resumeInputTaps()
```
vorssaint, line 53: when the teardown itself fails a few lines above, the abort returns without bringing the two brightness taps back, so the app stops handling the brightness keys even though it keeps running with its permissions intact. can that abort release them too

dajiaohuang: Fixed in d759d4a. If suspendInputInterceptors itself fails, uninstallCompletely now resumes the two brightness taps before calling onFailure; the existing detach-failure path already restores them. I also kept the session/accessibility guard ahead of re-enabling a timeout-disabled function-key tap. git diff --check passes; macOS CI remains authoritative.

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -395,6 +402,7 @@ final class BrightnessService: ObservableObject {
     }
 
     func stop() {
+        keyThreadLock.withLock { inputTapsSuspended = false }
```
vorssaint: switching the brightness feature off while the reset is still running drops this guard, so the media key tap comes back before the permissions are cleared, which is the hang this change is meant to avoid. can the guard stay up until the reset is over, whatever happens to the feature in between

dajiaohuang: Fixed in da0f61a. BrightnessService.stop() no longer clears inputTapsSuspended, so disabling the brightness feature or changing preferences during the asynchronous reset cannot re-arm either tap. resumeInputTaps() remains the explicit release point after reset completion or an uninstall abort. Swift/Xcode is unavailable on this Windows runner; git diff --check passes and macOS CI remains authoritative.

PR #1573, opened 2026-09-11
feat(menu-bar): implement customizable disk usage display styles in menu bar

maodijim:
## Summary

Adds a display option for the existing Disk Usage menu bar metric: used percentage, free size, or used size. Previously, it only showed the percentage used, so users could not check available or occupied disk space directly from the menu bar.

This belongs in Vorssaint’s existing system-monitoring feature rather than introducing a new integration. It uses the disk readings already collected by the monitor and the existing disk-size formatter, with no additional polling, runtime dependencies, permissions, or third-party services.

The change adds one persisted preference and a picker beneath Disk Usage, visible when that metric is enabled. Percentage remains the fallback, and no defaults are added to Defaults.swift. Free and used sizes display as text even when usage bars are selected, and the live preview reflects the choice. The picker reuses existing localized labels.

## Verification
- Ran/Built on MacOS 26.5.1
- App build succeeded, all 31,449 checks and preference-cleanup tests passed, and the diff has no whitespace errors.
```sh
./build.sh 
▸ Compiling (release) against MacOSX26.sdk…
▸ Compiling protected fan helper…
fan-control-helper: ok
▸ Compiling Now Playing adapter…
▸ Generating app icon…
iconset written to build/AppIcon.iconset
  adaptive icon skipped: actool not found (adaptive icons need Xcode 26+) (Dock falls back to AppIcon.icns)
/var/folders/qs/3rjd7rpx19x_4g_jyjq95gmc0000gn/T/tmp.qjG1sAtCvy/Vorssaint.app/Contents/Library/LaunchServices/com.vorssaint.utils.fan-control: replacing existing signature
/var/folders/qs/3rjd7rpx19x_4g_jyjq95gmc0000gn/T/tmp.qjG1sAtCvy/Vorssaint.app/Contents/Frameworks/libVorssaintNowPlaying.dylib: replacing existing signature
/var/folders/qs/3rjd7rpx19x_4g_jyjq95gmc0000gn/T/tmp.qjG1sAtCvy/Vorssaint.app: replacing existing signature
```

```sh
./build/Vorssaint --selftest
SELFTEST OK
```

```sh
./build.sh --test
TESTS OK (31449 checks)
PREFERENCE CLEANUP TESTS OK
```

## Images
- New Disk Menu options to select used percentage, free percentage, used disk, free disk to display in menu bar
image

- The preview of used disk in menu bar
image

- The preview of free percentage in bar style in menu bar
image

- The support of i18n
image

## Others
Ref #1574


Sources/Vorssaint/Services/Metrics/DiskSupport.swift
```diff
@@ -60,6 +60,36 @@ struct DiskDeviceReading: Identifiable, Equatable {
     }
 }
 
+enum DiskMenuBarStyle: String, CaseIterable {
+    case percent, freePercent, free, used
+
+    static let defaultsKey = "menuBarDiskStyle"
```
vorssaint: this choice is stored under a key the app never registers, so a settings backup restored on another mac silently goes back to the used percentage. can you register it the same way the memory style is, so it travels in the backup

Sources/Vorssaint/App/MenuBarRenderer.swift
```diff
@@ -600,15 +600,16 @@ enum MenuBarRenderer {
                 }
             case .diskUsage:
                 if let disk = primaryDisk(from: snapshot.disk) {
-                    if usesBars {
+                    let diskStyle = DiskMenuBarStyle.current
+                    if usesBars && diskStyle.showsPercentage {
                         groups.append([.usageBarBlock(label: "DSK",
-                                                      fraction: disk.usedFraction,
+                                                      fraction: diskStyle.fraction(for: disk),
```
vorssaint: with bars on and free percentage picked, the bar fills with the free share but the colors still read it as usage, so a nearly empty disk turns red and a nearly full one stays a short calm bar. keep the bar filling with used space and let the free choice apply only to the numeric readings

Sources/Vorssaint/UI/Settings/MonitorSettings.swift
```diff
@@ -384,6 +388,28 @@ private struct MemoryMenuBarOrderOption: View {
...
+    @AppStorage(DefaultsKey.menuBarDiskUsage) private var menuBarDiskUsage = false
+    @AppStorage(DiskMenuBarStyle.defaultsKey) private var diskStyle = DiskMenuBarStyle.percent
+
+    var body: some View {
+        if menuBarDiskUsage {
+            Picker(FeatureStrings.menuBarAppearance(l10n.language).label, selection: $diskStyle) {
+                Text("% \(l10n.s.diskUsed)").tag(DiskMenuBarStyle.percent)
+                Text("% \(l10n.s.diskFree)").tag(DiskMenuBarStyle.freePercent)
+                Text("\(l10n.s.diskFree)").tag(DiskMenuBarStyle.free)
+                Text("\(l10n.s.diskUsed)").tag(DiskMenuBarStyle.used)
```
vorssaint: this picker borrows the values and bars label from the top of the screen, and its options are the lowercase suffixes the panel glues after a number, so the menu reads as % used, % free, free and used. can you give it its own label and four proper option names in every language the app ships

Sources/Vorssaint/Services/Metrics/DiskSupport.swift
maodijim: Fixed  [91748fb](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/91748fbd7dfd345d1b94db988e8f0fc3626f1e11). Thanks for pointing it out, confirmed the value is saved between export and import setting.

Screenshot 2026-09-17 at 23 58 00

Sources/Vorssaint/App/MenuBarRenderer.swift
maodijim: True, I was thinking about the color as well, but I didn't want to simply reverse it. It's probably better to address it in future when users really need it, along with alerts and a new way to define the color in reverse.

Removing the free percentage option for now. [1d881f6](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/1d881f65ae2caff1e5d56b16fa44c024e1a2f5e8)

Sources/Vorssaint/UI/Settings/MonitorSettings.swift
maodijim: added [d96d88a](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/d96d88a867697596294e7f1c17c1e5b5b8fca09d)

image


image

Sources/Vorssaint/Services/Metrics/DiskSupport.swift
```diff
@@ -60,6 +60,36 @@ struct DiskDeviceReading: Identifiable, Equatable {
...
+
+    var showsPercentage: Bool { self == .percent }
+
+    var minimumValue: String { showsPercentage ? "100%" : "1000 GB" }
+
+    func fraction(for disk: DiskDeviceReading) -> Double {
+        if self == .free {
+            guard disk.totalBytes > 0 else { return 0 }
+            return min(1, max(0, Double(disk.freeBytes) / Double(disk.totalBytes)))
+        }
```
vorssaint: only the used percentage draws a bar now, so the free share worked out here never reaches the menu bar. can you drop this branch and the checks that cover it, so the bar always follows the used share

Sources/Vorssaint/Core/Localization.swift
```diff
@@ -2892,6 +2900,10 @@ extension Strings {
         diskSection: "Disks",
         diskUsed: "used",
         diskFree: "free",
+        diskMenuBarStyleLabel: "Disk display",
+        diskMenuBarUsedPercentage: "Used percentage",
+        diskMenuBarFreeSpace: "Free space",
```
vorssaint: this is the same figure the panel shows as available, and the number already counts space the system can purge. can you call the option available space so both places use one word

Sources/Vorssaint/Services/Metrics/DiskSupport.swift
maodijim: fixed [e10f65a](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/e10f65adb0320ac028c165b058bd39e8d960e11b)

Sources/Vorssaint/Core/Localization.swift
maodijim: updated [9c828df](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/9c828dfd51e0bbb0f513de5a89784abe4c273edd) [147fd76](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/147fd76ced445b29c3dc3fada6bafaba2ea00eb3)

Sources/Vorssaint/Core/Localizations/Strings+French.swift
```diff
@@ -706,6 +706,10 @@ extension Strings {
         diskSection: "Disques",
         diskUsed: "utilisé",
         diskFree: "libre",
+        diskMenuBarStyleLabel: "Affichage du disque",
+        diskMenuBarUsedPercentage: "Pourcentage utilisé",
+        diskMenuBarAvailableSpace: "Espace libre",
```
vorssaint: this still says free, and so do the german, italian, spanish, russian, turkish, japanese and korean options, while the panel in each of those languages calls the same figure available. can you use each language's own word for available here too

Tests/LocalizationTests.swift
```diff
@@ -40,7 +40,22 @@ enum LocalizationTests {
                      && languages.count == AppLanguage.allCases.count,
                      "the base strings cover each app language exactly once")
         suite.expect(!factories.isEmpty, "feature localization factories were discovered")
+        func diskPickerText(_ strings: Strings) -> [String] {
+            [strings.diskMenuBarStyleLabel, strings.diskMenuBarUsedPercentage,
+             strings.diskMenuBarAvailableSpace, strings.diskMenuBarUsedSpace]
+        }
+        let diskEnglish = diskPickerText(.enUS)
+        suite.expect(diskEnglish == ["Disk display", "Used percentage", "Free space", "Used space"],
```
vorssaint: this still expects free space for the english option, but that option now says available space, so this check fails. can you update the expected text

Sources/Vorssaint/Core/Localizations/Strings+French.swift
maodijim: updated all languages [711aaae](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/711aaaedb8a4088b2cedb74656dd039ff8fc393a)

Tests/LocalizationTests.swift
maodijim: fixed [e73045e](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/e73045e2d25cd72dbf46ca59fd7c6cac0e504f12)

Sources/Vorssaint/UI/Settings/MonitorSettings.swift
```diff
@@ -384,6 +388,27 @@ private struct MemoryMenuBarOrderOption: View {
...
+    var body: some View {
+        if menuBarDiskUsage {
+            Picker(l10n.s.diskMenuBarStyleLabel, selection: $diskStyle) {
+                Text(l10n.s.diskMenuBarUsedPercentage).tag(DiskMenuBarStyle.percent)
+                Text(l10n.s.diskMenuBarAvailableSpace).tag(DiskMenuBarStyle.free)
+                Text(l10n.s.diskMenuBarUsedSpace).tag(DiskMenuBarStyle.used)
+            }
+            .pickerStyle(.menu)
+            .font(.caption)
+            .padding(.leading, 58)
```
vorssaint, line 405: the monitor page now shows the metrics as a grid of tiles with the memory and network options as rows below it, so this indented small picker no longer lines up with anything there. can you show it as a row with the disk symbol and the label, like those two

maodijim: adopted new grid layout [74e4402](https://github.com/vorssaint/vorssaint-utils/pull/1573/commits/74e440200fc6ca17b1db362d6c86477067af3e81)

image

vorssaint:
LGTM

Issue #1540, opened 2026-09-09
[Bug]: A window opened right after switching to an app drops behind the switcher's target


akune:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Opening a new window in the app the App Switcher has just raised sometimes drops that new window behind the one the switcher targeted.

Expected: the new window stays in front and keeps the focus it was given.
Actual: occasionally the window that was switched to comes back over the new one, and typing continues in the old window.

It only happens when the app was raised with the switcher. Three runs isolate it:

| | Sequence | Result |
|---|---|---|
| A | Click a Safari window with the mouse, then Command-N and type a URL | never happens |
| B | Switch to Safari with the App Switcher, then Command-N and type a URL | happens |
| C | Quit Vorssaint, then repeat B | never happens |

As far as I could read it, the cause is the delayed focus pass in `WindowActivator`. Committing to a window schedules one after `focusRetryDelay` (0.12s), and three out to 0.68s when the source window was fullscreen, because a Space change is asynchronous and some apps settle their main window a run loop later. `SwitcherSupport.shouldContinueFocusRetry` decides whether that pass may still run, and it recognizes "the user has moved on" by application only — it continues while the target app is frontmost. Opening a window in that same app never changes which app is frontmost, so the pass runs and puts the switcher's target back over the window that was just created. The short window between the two is why it reads as occasional; with a fullscreen source it is far more frequent, which matches what I see.

Isolated on a build from `main`. The code path is unchanged since 3.3.5.

### Steps to Reproduce

1. Open Safari with at least one window, and put another app in front.
2. Switch to Safari with the App Switcher shortcut.
3. Within a fraction of a second, press Command-N and start typing a URL.
4. Sometimes the new window drops behind the Safari window that the switcher raised, and the typing carries on in that one.
5. Starting from a fullscreen Safari window makes it much more frequent.
6. For the control: repeat with the app raised by clicking its window instead of using the switcher, and again with Vorssaint quit. Neither shows it.

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.5

### macOS Version

macOS 26 (Tahoe)

### Language

English · macOS in English

### Mac and displays

[...]

PR #1578, opened 2026-09-11
fix(switcher): let every delayed focus pass stop for a window opened after the switch

akune:
## Summary

Two cases #1541 left open, both traced on a diagnostic build that logged every delayed focus pass with its inputs.

**The guard compared two different window lists.** It reads Accessibility only once the window server shows the app gained a window, but took that list on screen only while the snapshot covers all windows. A window the app has just opened is focused before the window server composites it, so the guard saw nothing new and never took the focus reading. Logged in Finder: `onScreen=[77,39507]` against the snapshot, while Accessibility already answered 44049 and the all-windows list contained it. The live list now uses the snapshot's scope. Two passes seconds later showed the all-windows list unchanged, so this keeps the reading rare.

**A space hop's arrival pulses never asked the guard.** A switch away from a fullscreen app goes through `SpaceHop`, which raises the target in three pulses up to a second later — and started before the activation took its snapshot. Logged: fullscreen source to Mail, Command-N, three raises of the old Mail window and not one pass through the guard. The hop now snapshots the app's windows when it begins and every pulse asks the same guard before raising.

## Verification

```sh
./build.sh                     # finished, no warnings
./build.sh --test              # TESTS OK (31446 checks)
```

Each fix checked in both directions: the scope test goes red with the on-screen list restored, the hop test goes red with the guard removed from `focusAfterSpaceHop`. After the change, the same fullscreen-to-Mail sequence logged the first pulse stopping at the new window (`allNew=[48625] focused=48625 → stop`) and the two remaining pulses raising nothing.

Confirmed by hand on a MacBook Pro (M4 Max, Mac16,5), macOS 26.6.2 (25G83), built-in Liquid Retina XDR plus one external display, English, own build with a stable self-signed identity: from a windowed source and from a fullscreen app, Command-N right after switching keeps the new window in front.

Not tested: other Macs and macOS versions; Spaces layouts beyond one Space per display plus fullscreen apps; apps other than Safari, Mail and Finder.

## Anything else

Refs #1540

### DIFF
--- a/Sources/Vorssaint/Services/Switcher/WindowActivator.swift
+++ b/Sources/Vorssaint/Services/Switcher/WindowActivator.swift
@@ -504,7 +504,13 @@ enum WindowActivator {
             sourcePID: sourcePID,
             frontmostPID: currentFrontmostPID(),
             targetMinimizedState: minimizedState,
-            targetAppWindowIDs: windowIDs(ownerPID: targetWindowOwnerPID, options: .optionOnScreenOnly),
+            // The same scope the snapshot used. The on-screen list lags: a
+            // window the app has just opened is focused, and answered as
+            // focused by Accessibility, before the window server composites
+            // it — so comparing on-screen windows against an all-windows
+            // snapshot reported nothing new in exactly the race this guard
+            // exists for, and the focus reading below was never taken.
+            targetAppWindowIDs: windowIDs(ownerPID: targetWindowOwnerPID, options: .optionAll),
             targetAppFocusedWindowID: focusedWindowID(for: targetWindowOwnerPID)
         )
     }
--- a/Sources/Vorssaint/Services/Switcher/WindowActivator.swift
+++ b/Sources/Vorssaint/Services/Switcher/WindowActivator.swift
@@ -74,6 +74,7 @@ enum WindowActivator {
            SpaceHop.beginIfNeeded(windowID: windowID,
                                   appPID: item.pid,
                                   windowOwnerPID: windowOwnerPID,
+                                  sourcePID: sourcePID,
                                   app: app) {
             return
         }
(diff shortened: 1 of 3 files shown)


Sources/Vorssaint/Services/Switcher/WindowActivator.swift
```diff
@@ -698,10 +705,31 @@ enum WindowActivator {
...
+                                   appPID: pid_t,
+                                   windowOwnerPID: pid_t,
+                                   sourcePID: pid_t?,
+                                   state: SwitcherWindowFocusRetryState) {
         guard let app = NSRunningApplication(processIdentifier: appPID), !app.isTerminated else { return }
+        guard shouldContinueFocusRetry(windowID: windowID,
+                                       targetPID: appPID,
+                                       targetWindowOwnerPID: windowOwnerPID,
+                                       sourcePID: sourcePID,
+                                       state: state) else { return }
```
vorssaint: a hop over two or more desktops can arrive with another app still in front, since the system activates whatever window is on top on each desktop it passes. with this stop the pulses then leave the selected window behind that app. does it still come up in that layout, or should a pulse only give up when the app itself moved to a window it opened later

akune: Measured rather than reasoned: three desktops in a row, Disk Utility on 1, Maps on 2, Finder windows on 1 and 3 so the hop takes the stepping path, plus a second Maps window on top of desktop 3. Four hops, twelve arrival pulses, and in every one the foreground was already the target app: `front=Finder frontIsTarget=true newSeenAll=false focused=75852 → CONTINUE`. The veto branch was never reached and the selected window came up each time. By the time the pulses run — about 2.3s after the commit, the travel taking most of it — macOS has settled on the app the hop activated before travelling.

So I could not make that layout fail here. The branch is reachable by construction though, and a foreground reading during travel carries no intent either way, so I took your second option in e9be44e: the guard can treat the foreground as meaningless on request, and the hop's arrival pass is the only caller that asks for it. It still gives up when the app moved to a window it opened after the switch.

One thing the trace showed that is not this branch's doing: all three pulses raise even when the first already found the window focused and frontmost — three raises 0.3–0.5s apart, visible as a flicker on the target window. Filing that separately.

Issue #1585, opened 2026-09-11
Add screenshot preview toggle, Escape dismissal, and automatic save options


EugeneCarldotme:
## Problem

Automatic clipboard copying still shows the screenshot preview, interrupting work after each capture. The original report in #1389 says:

> Trying to find a fix for this issue as well. After taking a screenshot, I just always get this pop up showing me options to copy, save, or delete. I just want that not appear at all so that I can go about my work without having to go through one more step to close that pop up.
>
> It's super handy to have the feature where it automatically copies the screenshot to clipboard, but that pop up every time I take a screenshot that shows a small preview gets in the way and it's excessive with no way to disable.
>
> Otherwise, everything is spectacular.

### Proposed Solution

- Add a Screenshot Settings toggle to show or hide the preview, on by default.
- Keep separate toggles for automatic clipboard copying and automatic saving to the configured screenshot folder. Support clipboard only, file only, and both.
- Remove the caption sentence "Saving a file stays a separate choice."
- Escape dismisses the visible preview without deleting a saved file or clearing the clipboard, including when the preview has not taken focus.
- Preserve existing after-capture preferences and explicit editing/history actions. If automatic output fails, retain a way to recover the capture.

### Feature Area

Clipboard and files

### Does another app already do this?

Shottr

### Mockups / References

https://github.com/vorssaint/vorssaint-utils/issues/1389#issuecomment-5630865822

fnadirli:
Looking forward to this feature. Shottr handles it in a very elegant way:

<img width="517" height="582" alt="Image" src="https://github.com/user-attachments/assets/df514e72-0d01-4eb5-9e61-de690c3dac2d" />

PR #1586, opened 2026-09-11
feat(screenshot): add preview toggle and automatic output options

EugeneCarldotme:
## Summary

Add controls for screenshot preview and automatic output after capture.

- Add a Preview toggle, enabled by default. Successful automatic output can now finish without showing the preview.
- Hide preview placement and focus settings while Preview is off.
- Add separate Automatic copy and Automatic save controls, synchronized with the Default action picker for copy-only, file-only, both, and Edit.
- Escape dismisses the visible preview without deleting a saved file or clearing the clipboard, including when the preview does not have focus.
- Restore the preview if hidden automatic clipboard output fails, so the capture remains recoverable.
- Add localized labels and captions for the new controls.
Screenshot 2026-09-12 at 01 19 53

Screenshot 2026-09-12 at 01 20 09

Screenshot 2026-09-12 at 01 21 30

## Verification

Passed locally on the final source state:

- `./build.sh` — release build and signed bundle completed.
- Staged app `--selftest` — `SELFTEST OK`.
- `git diff --check` — passed.
- `./build.sh --test` — `TESTS OK (38,575 checks)`.
- Preference cleanup — `PREFERENCE CLEANUP TESTS OK`.

Live screenshot capture and Dynamic Island UI interactions were not manually tested.

## Anything else

Merged main and resolved the screenshot preview and settings conflicts.

Refs #1585


Sources/Vorssaint/UI/Settings/ScreenshotSettings.swift
```diff
@@ -155,19 +156,29 @@ struct ScreenshotCaptureSettings: View {
...
                 }
             }
 
             Section {
-                Toggle(strings.autoCopyToggle, isOn: $copyToClipboard)
+                Toggle(strings.autoCopyToggle, isOn: automaticCopy)
                 Text(strings.autoCopyCaption)
                     .font(.caption)
                     .foregroundStyle(.secondary)
+                Toggle(strings.autoSaveToggle, isOn: automaticSave)
```
vorssaint, line 184: the default action already offers save, copy and both. remove the extra automatic save toggle and keep the existing output choices, focusing this change on preview visibility and escape dismissal. automatic copying should still work alongside edit

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -367,27 +367,29 @@ final class ScreenshotService: ObservableObject {
...
             forKey: DefaultsKey.screenshotLastCaptureShortcutEnabled) {
             ScreenshotLastCaptureStore.save(capture)
         }
-        if UserDefaults.standard.bool(forKey: DefaultsKey.screenshotCopyToClipboard) {
-            autoCopy(capture)
+        let copies = UserDefaults.standard.bool(forKey: DefaultsKey.screenshotCopyToClipboard)
+        let configuredAction = ScreenshotDefaultAction.current
+        let defaultAction = configuredAction.automaticOutputAction(copyToClipboard: copies)
+        let showPreview = UserDefaults.standard.bool(forKey: DefaultsKey.screenshotShowPreview)
+        if copies && ![.copy, .saveAndCopy].contains(defaultAction) {
```
vorssaint, line 403: with automatic copying and saving enabled, a failed save now prevents the clipboard copy too. copying previously worked independently. keep copying available when the destination cannot be written, while leaving the failed save recoverable

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -522,13 +525,15 @@ final class ScreenshotService: ObservableObject {
...
-    /// Automatic copy stays quiet on success: the preview or the editor is
-    /// already appearing and says the capture happened, so a HUD on top of it
-    /// would only repeat that. A failure still beeps, since nothing else
-    /// would reveal an empty clipboard before the paste.
-    private func autoCopy(_ capture: ScreenshotSelectionController.Capture) {
+    /// Keep automatic copying quiet; recover failed hidden captures in a preview.
+    private func autoCopy(_ capture: ScreenshotSelectionController.Capture,
+                          recoverOnFailure: Bool = false) {
+        let recover = { [weak self] in
+            if recoverOnFailure { self?.restorePreview(capture) }
```
vorssaint, line 552: if a hidden capture fails to copy after someone opens another capture from history, the older capture replaces the one they just chose. recover the failed capture without closing or replacing a newer preview

Sources/Vorssaint/Services/QuickTools/ScreenshotQuickPreviewController.swift
```diff
@@ -373,11 +389,28 @@ final class ScreenshotQuickPreviewController {
...
+        }
         keyMonitor = NSEvent.addLocalMonitorForEvents(matching: .keyDown) { [weak self, weak panel] event in
-            guard let self, !self.closed, let panel, panel.isVisible, event.window === panel,
+            guard let self, !self.closed, let panel, panel.isVisible,
                   !self.shownInNotch || NotchService.shared.isCaptureVisible(id: self.presentationID),
                   panel.attachedSheet == nil, !(panel.firstResponder is NSText),
                   !ShortcutCapture.isCapturing else { return event }
+            if event.keyCode == kVK_Escape,
+               event.modifierFlags.intersection([.command, .control, .option, .shift]).isEmpty {
+                self.close()
```
vorssaint, line 415: when another window in the app has a dialog open, escape closes the screenshot preview but never reaches that dialog. dismissing an unfocused preview should still let escape cancel the dialog the person is using

Sources/Vorssaint/Services/QuickTools/ScreenshotQuickPreviewController.swift
```diff
@@ -373,11 +389,28 @@ final class ScreenshotQuickPreviewController {
     }
 
     private func installKeyMonitor(for panel: NSPanel) {
+        globalKeyMonitor = NSEvent.addGlobalMonitorForEvents(matching: .keyDown) { [weak self] event in
```
vorssaint: escape cannot close an unfocused preview while another app is active without accessibility permission. screenshots currently ask only for screen recording, so the new hint promises an action that may do nothing. explain and handle the missing permission while keeping capture available

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -387,27 +389,29 @@ final class ScreenshotService: ObservableObject {
...
+        if copies && ![.copy, .saveAndCopy].contains(defaultAction) {
+            autoCopy(capture, recoverOnFailure: !showPreview && defaultAction != .edit)
         }
-        if ScreenshotDefaultAction.current == .edit {
+        if defaultAction == .edit {
             openEditor(with: capture)
             return
         }
-        presentPreview(capture, defaultAction: ScreenshotDefaultAction.current)
+        if !showPreview && defaultAction == .none && copies { return }
```
vorssaint, line 412: with the preview hidden, an automatic copy that works ends here with nothing on screen, while the copy default action still shows the small copied message. can the automatic copy show that same message when no preview is coming

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -569,7 +600,10 @@ final class ScreenshotService: ObservableObject {
...
             }.value
             guard let output else {
-                if !Task.isCancelled { NSSound.beep() }
+                if !Task.isCancelled {
+                    NSSound.beep()
+                    recover()
+                }
                 return
             }
             guard let self, !Task.isCancelled,
```
vorssaint, line 609: when something else lands on the clipboard while this copy is being prepared, the hidden capture is dropped just below with no sound and no preview. can that case bring the preview back like the other failures do, without doing it when a newer capture replaced this one

Sources/Vorssaint/UI/Settings/ScreenshotSettings.swift
```diff
@@ -155,8 +156,20 @@ struct ScreenshotCaptureSettings: View {
                     Text(strings.loupeZoomOptionCaption)
                         .font(.caption)
                         .foregroundStyle(.secondary)
-                    previewPositionRow
-                    previewFocusRow
+                    Toggle(strings.previewToggle, isOn: $showPreview)
```
vorssaint, line 159: with this off and no automatic copy or save chosen, every capture still shows the preview and nothing here says why. can the toggle be unavailable or explained in that state

Sources/Vorssaint/UI/Settings/ScreenshotSettings.swift
```diff
@@ -155,8 +156,20 @@ struct ScreenshotCaptureSettings: View {
...
+                    if showPreview {
+                        Text(strings.previewEscapeCaption)
```
vorssaint, line 161: the preview takes the keyboard by default now, so escape already closes it without any permission unless that option is off. can this text and the accessibility ask appear only when focusing the preview is off

Sources/Vorssaint/UI/Settings/ScreenshotSettings.swift
```diff
@@ -155,8 +156,20 @@ struct ScreenshotCaptureSettings: View {
...
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
+                        if !permissions.accessibility {
+                            Button(strings.previewAccessibilityButton) {
+                                permissions.requestAccessibility()
+                                permissions.openAccessibilitySettings()
+                            }
```
vorssaint, line 168: this fires the system prompt and opens system settings at once, while every other permission ask in settings uses the same row with its status and separate buttons. can it use that row too

Sources/Vorssaint/Services/QuickTools/ScreenshotQuickPreviewController.swift
```diff
@@ -373,11 +390,33 @@ final class ScreenshotQuickPreviewController {
     }
 
     private func installKeyMonitor(for panel: NSPanel) {
+        // Global key events require Accessibility. Local dismissal and capture
+        // remain available without it; never prompt from the capture path.
+        if AXIsProcessTrusted() {
```
vorssaint, line 396: the permissions page lists what uses accessibility and suggests revoking it when nothing does, so someone who grants it just for this escape gets told to take it away. can the preview count as using accessibility there

Sources/Vorssaint/UI/Settings/ScreenshotSettings.swift
```diff
@@ -155,8 +156,20 @@ struct ScreenshotCaptureSettings: View {
...
+                        }
+                        previewPositionRow
+                        previewFocusRow
+                    }
                     defaultActionRow
```
vorssaint, line 173: the caption under this still says the confirmation appears so you can edit or discard afterward, which is no longer true with the preview off. can it say what happens in that case

PR #1588, opened 2026-09-11
fix(AudioMixer): using exponential instead of linear gain scaling

ahmedabdullah-o:
## Summary

Now the audio mixer uses exponential gain scaling instead of linear scaling.
Linear scaling is not intuitive because humans hear changes in volume in
a logarithmic way.

## Verification

Device: MacBook Air M1 8GB 256GB

I ran the routine verifications:
```sh
./build/Vorssaint --selftest
./build.sh --test
```
No warnings, no failed tests.
No issues related to this PR.

### DIFF
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -1902,7 +1902,7 @@ private final class TapGainEngine: GainEngine {
             // here on leaves the output written.
             let frames = MixerRender.render(source: inputBuffers[tapIndex],
                                             into: outputBuffers,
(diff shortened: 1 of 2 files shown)


vorssaint:
closing this one. the square is applied to the whole range, so at 200% the app now plays at four times its level, and going past double was already turned down in #71 because it only clips or pumps
every volume people already saved silently changes meaning after the update, 50% drops from half to a quarter and 30% ends up near silence
the percentage shown and typed in the mixer stops matching what the app actually plays
a curve that only bends the range below 100% and keeps 100% and the 200% boost exactly where they are could still be worth a look, as long as what people already saved keeps sounding the same

PR #1590, opened 2026-09-11
feat(monitor): Add connected devices and disk count to menu bar metrics

kirolos-esmat:
This PR adds the ability to track the number of connected USB/Thunderbolt devices and external disks directly in the menu bar with category exclusions.

### Features
- **Connected Devices Metric:** Displays the count of connected USB/Thunderbolt devices in the menu bar.
- **External Disks Count Metric:** Displays the count of connected disks.
- **Category Exclusions:** Users can choose to exclude specific device categories from the count:
  - Exclude Power Supplies / Chargers
  - Exclude USB Hubs & Docks
  - Exclude Network & Ethernet Adapters
  - Exclude External Storage
- **Accurate Storage Deduplication:** Recursively matches IOKit USB devices to mounted partition volumes via BSD identifier, ensuring connected drives are counted once rather than duplicated.
- **Live Updates:** Counts update live in the menu bar using `NSWorkspace` and `IOKit` notifications.
- **Metric Detail Popover:** Provides detailed overview of connected devices, speeds, and volumes.

#### Menu Bar Metric
Connected Devices in Menu Bar

#### Settings & Category Exclusions
Connected Devices Settings

### DIFF
--- a/Sources/Vorssaint/App/FeatureRuntime.swift
+++ b/Sources/Vorssaint/App/FeatureRuntime.swift
@@ -271,6 +271,7 @@ final class FeatureRuntime: ObservableObject {
         .monitorMemory: { FeatureRuntime.syncMonitor() },
         .monitorNetwork: { FeatureRuntime.syncMonitor() },
         .monitorDisk: { FeatureRuntime.syncMonitor() },
+        .monitorUSB: { FeatureRuntime.syncMonitor() },
         .monitorPower: { FeatureRuntime.syncMonitor() },
         .fanControl: {
             SystemMonitor.shared.planDidChange()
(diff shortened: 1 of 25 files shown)


vorssaint:
closing this one
the change goes well past the title, with fan names, detailed temperatures and a new fan look in the menu bar bundled in, and each of those needs its own decision
hardware brand and model names are used to guess icons and colors, and the app keeps brand names out of the code
the new labels only exist in english, so every other language would show english text in the panel, the settings and the hub
the usb watcher starts with the app and rescans on every settings write even when nothing usb related is switched on, and that kind of work has to stay off while a feature is not in use
fan sensors are read even with fan control uninstalled, and they are read twice per tick with four keys per fan
the new usb section and the fan rows in the system card show up for everyone by default, and new features start switched off
a plain count of plugged in devices that only listens while the metric is on, with text in every language and nothing else in the same change, could still be looked at as its own small pull request

Issue #1579, opened 2026-09-11
Close Window and Quit App thermometer


Cronbo:
### Problem or Motivation

It would be great if the super handy Close and Quit protection feature had a little thermometer that appears overlayed on the existing pop-up bubble

### Proposed Solution

To show how long until it Closes or Quits. Like a realtime countdown so you have a sense of time left until it executes either of those commands

### Feature Area

Windows and Dock

ywu73:
Implemented in draft PR #1595: https://github.com/vorssaint/vorssaint-utils/pull/1595

The existing hold-to-confirm bubble now shows a linear progress bar for Command-Q and Command-W, driven by the existing confirmation timer's deadline. Cancelling or reusing the bubble clears the progress. Double-press, extra-modifier, and switcher confirmations keep their current presentation.

Local validation passed: release build, app selftest, 32,333 automated checks, and 13 isolated native AppKit HUD checks with visual inspection. Tested on MacBookPro17,1 with macOS 15.7.7 (24G720).

The PR remains a draft pending actual global-shortcut testing in an installed build. The change has not been merged or released yet.

Cronbo:
🤯

PR #1595, opened 2026-09-12
feat(quit-protection): show progress during hold confirmation

ywu73:
## Summary

Holding Command-Q or Command-W now shows a linear progress bar beneath the confirmation text. The track appears and disappears immediately with the prompt, and cancellation or HUD reuse clears the fill animation.

The prompt names only the requested action: Command-Q quits and Command-W closes. Separate complete sentences cover all 13 languages, all three confirmation modes, and the app switcher's confirmation prompt.

## Implementation

The HUD uses the existing hold timer's deadline for a Core Animation fill. Implicit animation of the track's visibility is disabled so it does not fade through the shortest, 250 ms hold. Layout reserves 8 pt for progress and measures the localized labels. Confirmation timing and action dispatch remain unchanged; no settings or dependencies are added.

The standard `quit-protection` suite exercises the production HUD content view through the existing test-source generator. The README includes the updated action-specific prompt.

## Validation

- `./build.sh`: successful release build without compiler warnings.
- `./build.sh --test`: 38,809 checks and preference cleanup passed.
- Final HUD fixture: 183 checks passed. After merging the fix into this branch, HUD and localization suites passed together: 4,293 checks; selftest also passed.
- Native AppKit HUD harness: 9 checks passed for Q/W progress at 250 ms, no implicit visibility animation, completed fill, cancellation cleanup, and preserving the foreground application.
- Mutation checks on temporary copies: removing visibility-animation suppression caused one expected failure; reversing the hold prompt's action selection caused two expected failures.
- Updated HUD screenshots were visually checked.

Environment: MacBookPro17,1, macOS 15.7.7 (24G720), locally built release configuration signed with the existing local self-signed identity.

## Screenshot

Production AppKit HUD rendered with the English quit prompt and the fill fixed at 55% for documentation.

Hold Command-Q to quit, with a progress bar and Esc cancellation hint

Refs #1579


Sources/Vorssaint/UI/QuitProtection/QuitProtectionHUD.swift
```diff
@@ -101,11 +118,19 @@ final class QuitProtectionHUD {
...
+            CATransaction.setDisableActions(true)
+            progressTrack.frame = CGRect(x: 24, y: 7, width: bounds.width - 48, height: 3)
+            progressFill.frame = progressTrack.bounds
+            CATransaction.commit()
         }
 
-        func update(title: String, detail: String) {
+        func update(title: String, detail: String, showsProgress: Bool) {
+            stopProgress()
+            progressTrack.isHidden = !showsProgress
```
vorssaint, line 133: this can still fade the track in over a quarter of a second instead of showing it together with the text, which covers the whole of the shortest hold. can you make it appear at once, and hide it the same way

Sources/Vorssaint/Services/QuitProtection/QuitProtectionService.swift
```diff
@@ -514,7 +514,8 @@ final class QuitProtectionService: ObservableObject {
         } else {
             title = String(format: strings.doubleHUDFormat, shortcut.symbol)
         }
-        hud.show(title: title, detail: strings.cancelHint)
+        hud.show(title: title, detail: strings.cancelHint,
+                 holdDeadline: configuration.mode == .hold ? holdTimer?.fireDate : nil)
```
vorssaint, line 518: when you hold the shortcut in the installed app, does the bar reach the end right as the app quits or the window closes, including at the shortest hold time

Cronbo:
The Screenshot shows incorrect text: It should ONLY mention Quit when showing Cmd + Q shortcut, and ONLY mention Close when showing Cmd + W

Current text reads as if Cmd + Q is for both Quitting (app) AND Closing (a window) which have unique shortcuts —Cmd + W for Close, not Cmd + Q.

image

ywu73:
Thanks for catching this. Fixed in 6242da16: ⌘Q now shows “Hold ⌘Q to quit,” and ⌘W shows “Hold ⌘W to close.” The same distinction applies to double-press and extra-modifier prompts across all supported languages. The README screenshot and PR description have also been updated.

Sources/Vorssaint/UI/QuitProtection/QuitProtectionHUD.swift
ywu73: Fixed: the progress track now appears and disappears immediately with the text, without an implicit fade. I also verified this manually in the locally built app at the shortest hold duration (250 ms).

Sources/Vorssaint/Services/QuitProtection/QuitProtectionService.swift
ywu73: Manually verified in a local build of commit 13e21922: for both ⌘Q and ⌘W, the progress bar reaches the end as the app quits or the window closes, including at the shortest hold duration (250 ms).

Issue #1581, opened 2026-09-11
cleaner removing files neccessary for apps


KianYari:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

it scans a file neccessary for logiplus

<img width="560" height="146" alt="Image" src="https://github.com/user-attachments/assets/4852f949-df2e-4929-b7f2-218e7f551eae" />

### Steps to Reproduce

install logiplus
scan via cleaner
it findes logitec.localized
you delete it
the app won't work

### Feature Area

Tools

### Vorssaint Version

3.3.5

### macOS Version

macOS 26 (Tahoe)

### Language

english

PR #1596, opened 2026-09-12
fix(cleaner): preserve localized entries across scan categories

ywu73:
## Summary

- Exclude `.localized` entries from leftover, cache, and log scans, case-insensitively.
- Reject previously listed `.localized` entries before removal regardless of their category.
- Add 26 filesystem regression checks covering cross-category scanning, stale selections, and valid preference and container owners.

Refs #1581

## Motivation

Issue #1581 reports that cleaning `logitec.localized` broke LogiPlus. A Finder localization suffix does not establish application ownership or disposable data. Review identified that an entry skipped by the leftover scan could reappear as a recommended cache or log item and be selected by scheduled cleanup.

## Changes

The leftover owner check rejects actual entry names ending in `.localized`. The shared cache exclusion policy also rejects these names, covering both `scanCaches` and `scanLogs`. The filename check in `mayRemove` applies before category-specific checks, protecting stale selections in manual and scheduled cleanup.

Checked the leftover-to-cache/log scan ordering, cache and log policy callers, all `mayRemove` calls in `cleanSelected`, and the scheduler's use of the recommended selection. Preference filenames and container metadata retain valid identifiers ending in `.localized`.

The regression harness runs production scan, size, and removal methods against temporary directories with isolated home and installed-owner lookups. It checks that excluded leftovers cannot return in cache/log results and that ordinary cache/log entries remain recommended. Upstream test registrations and generated-source additions are preserved.

## Testing

- Before the review fix, `./build.sh --test-suite=cleaner` failed 6 of 26 checks, reproducing cache/log reclassification and stale removal eligibility.
- After the fix, `./build.sh --test-suite=cleaner` passed all 26 checks.
- Ran `./build.sh --test`: all 38,289 checks and preference cleanup tests passed.
- Ran the standard release command `./build.sh`: passed without compiler warnings or errors; no compiler wrapper was used.
- Ran `./build/Vorssaint --selftest`: `SELFTEST OK`, including a repeat after the local main merge.
- Ran `git diff --check`: passed. The merged tree is identical to the validated repair commit.
- Verified on macOS 15.7.7 (24G720), arm64. Fixtures did not clean real application data; the reported LogiPlus installation was not available for an end-to-end retest.

## Risks

- Entries ending in `.localized` are conservatively retained even when they could be disposable.
- Reporter confirmation of the LogiPlus scenario remains pending.
- No API, configuration, or migration changes.

## Screenshots / Recordings

- Not applicable.

## Checklist

- [x] Tests have been added or updated where appropriate — 26 regression checks and the full suite passed.
- [x] Documentation has been updated where appropriate — production comments and this description explain the behavior; no user documentation change is needed.
- [x] Breaking changes are documented — none; conservative retention is documented under Risks.
- [x] The PR description reflects the actual diff — verified against the five changed files relative to upstream main.

### DIFF
--- a/Sources/Vorssaint/Services/Cleaner/JunkCleaner.swift
+++ b/Sources/Vorssaint/Services/Cleaner/JunkCleaner.swift
@@ -252,6 +252,7 @@ final class JunkCleaner: ObservableObject {
               url.resolvingSymlinksInPath().standardizedFileURL.path == path else { return false }
         if item.category == .leftovers {
             guard isDirectLeftoverRootChild(url),
+                  !url.lastPathComponent.lowercased().hasSuffix(".localized"),
                   CleanerSupport.bundleIDCandidate(fromEntryName: item.detail) != nil,
                   !CleanerSupport.isProtectedBundleID(item.detail),
                   !hasLivingOwner(item.detail, installed: installed) else { return false }
@@ -451,6 +452,8 @@ final class JunkCleaner: ObservableObject {
     private static func leftoverOwner(entry: String,
                                       url: URL,
                                       usesContainerMetadata: Bool) -> String? {
+        // Finder uses this suffix for display names, not application ownership.
+        guard !entry.lowercased().hasSuffix(".localized") else { return nil }
         if usesContainerMetadata, let owner = containerOwner(at: url) {
             return owner
         }
--- a/Tests/CleanerEligibilityTests.swift
+++ b/Tests/CleanerEligibilityTests.swift
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+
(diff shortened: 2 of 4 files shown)


Sources/Vorssaint/Services/Cleaner/JunkCleaner.swift
```diff
@@ -451,6 +452,8 @@ final class JunkCleaner: ObservableObject {
     private static func leftoverOwner(entry: String,
                                       url: URL,
                                       usesContainerMetadata: Bool) -> String? {
+        // Finder uses this suffix for display names, not application ownership.
+        guard !entry.lowercased().hasSuffix(".localized") else { return nil }
```
vorssaint, line 456: a localization folder can disappear from app leftovers and return already selected under temporary files or logs. scheduled cleanup can then move it to the trash even though it previously required a separate choice. keep that protection across the whole scan and add a test covering this path

PR #1601, opened 2026-09-12
feat(menu-panel): show the current track with transport controls

dorbd:
Picks up #303. The blocker named on it is gone — per the note in #838, this builds on the Now Playing bridge that is in `main` and driving the radial menu's card, rather than rebasing the branch that died with the old one.

Two things up front. It has **not** yet had the few days of living-with that a feature here is supposed to get, and the exact gaps are listed under *Verified where* rather than left implied. And there are two direction questions at the bottom — the first one changes the size of this diff substantially, so it is worth answering before anyone reads the code closely.

Refs #303

## Before / after

Before, the current track was visible inside Vorssaint only as the radial menu's card — unreachable if you do not use the wheel, which is what the reporter said in #303 — and that card has no playback controls. After, the menu bar panel has a Now Playing section: artwork, title, artist, album, and previous / play-pause / next for whatever owns the system Now Playing session.

Ships uninstalled, as a hub feature under Sound, hideable from the panel like every other section.

## Built on what is already in main

Per the note in #838 ("whoever picks #303 up again should build on that, not rebase the old branch"):

- **The read** is the same out-of-process one the radial card uses — `Sources/NowPlayingAdapter` through `BoundedProcessRunner`. The bridge grew one parameter and a second caller. Nothing new reads MediaRemote, and no new private framework is taken.
- **The transport** posts the aux-button pair the physical media keys post, so it reaches whatever owns the session with no per-application integration. It does need **Accessibility**, since that is a synthetic event — the same grant the radial menu's media slices already declare. (I had this wrong in the first push of this branch and the buttons silently did nothing; the second commit fixes the declaration.)
- **Nothing polls.** The panel is a popover, so a read happens when the section appears and once after a transport press, never while the panel is shut.

## Three shared things I changed rather than copied

1. `postAuxKey` was private to `RadialMenuService`. It is now `MediaKeyInput`, pressed by both callers, so the media-key path did not become another private copy of a synthetic-event builder — #937 is the record of what that costs.
2. `RadialNowPlayingSupport.snapshot` gained `includesPaused`, defaulting to the radial menu's existing behaviour. A paused session must read as nothing playing on the wheel, but a mini player that went blank the moment you hit pause tells the user something false, so the panel asks for it and draws a play button. The snapshot now carries `isPlaying`.
3. The four rules in `PanelLayout.order` that place a section an existing saved order predates were one expression copied per section, and none of it was reachable from the test list. Now one table and one function in `Defaults`, so the new section landing next to the mixer instead of after Quick toggles is covered by a test rather than by inspection. Callers checked: `PanelLayout.order` is the only reader, via `MenuPanelView` and the layout editors.

## Tests

Added to the existing Now Playing block: the `includesPaused` contract in both directions, that ownerless metadata is still nothing playing either way, the optimistic flip changing only playback state, a source-shape assertion that the aux event is built in one place and the radial menu presses it through that (comments stripped first), and the saved-order upgrade path including a reordered panel. Verified red with each of the three pieces of logic reverted — six assertions fail, then green again.

`AppFeature.allCases.count` 57 → 58 and the ordered raw-value list updated.

## Verified where

MacBook Pro (Mac16,8), M4 Pro, macOS 26.6.1 (25G76), single built-in display, one Space, English, a `--dev` build installed beside the released 3.3.5.

- `./build.sh` finishes clean with no warnings, `--selftest` OK, `./build.sh --test` OK (32383 checks). The full `./build.sh` was run as well as `--test`, since this touches `UI/`.
- The new assertions were checked in both directions: reverting each of the three pieces of logic turns six of them red, and they go green again with the logic in place.
- **The read path was exercised against the installed bundle on this machine** and came back with a real session, artwork included. Worth stating explicitly because this is the path that returns nothing to an unsigned process on 15.4+, and it is what the earlier #303 branch died on.
- That session was **paused**, which is exactly the case this branch is about: the existing builder discards a paused session, so without `includesPaused` the panel would have reported nothing playing while a track was plainly loaded.
- The app launches with the feature installed and stays up, and **the section renders** — artwork, title, the second metadata line and the transport, with the section's tab in the panel's navigation strip.
- Two defects came out of actually using it, both fixed in the second commit: the transport declared no permissions (so it did nothing at all until Accessibility was granted), and the section only read on `onAppear`, so a session that changed while the popover was shut stayed stale on screen until a button was pressed. Neither was visible from the code or the tests; both took opening the panel.

**Not verified, and I would rather say so than imply otherwise:**

- It has not had the few days of use a feature here is supposed to get. That is why this is a draft.
- Any language other than English. The two new strings are machine-assisted and want a native reader; the section title reuses the already-translated `mediaNowPlaying`.
- Multiple displays, multiple Spaces, a release build, and VoiceOver on the transport buttons.
- **The transport actually changing playback.** The permission declaration is fixed and correct by inspection against the radial menu's, but I have not yet watched a press move a player with the grant in place.
- Covered since: a browser tab (WebKit) as the session owner renders correctly, title and artwork included. Still untested is anything that reports no title at all, where the section falls back to the owning app's name.

## Two questions

1. **Own feature, or fold into the Sound section?** As written this is a new hub feature and a new panel section, which is most of the diff. Folding it into the existing mixer/Sound section would cut roughly 140 lines and ten touched files, at the cost of needing the Mixer installed to see your current track. I picked the standalone shape because #303 asks for a player and the reporter does not use the radial menu, but the smaller one is a one-commit change if you prefer it.
2. **Should the radial card show paused sessions too?** `includesPaused` defaults to off purely to leave the wheel's behaviour untouched. If a paused track should appear there as well, the parameter disappears and the decision moves to the two callers, which is tidier.

Not in scope, and worth saying so: **there is no progress bar or scrubbing.** The adapter forwards no elapsed time or duration, so that needs the bridge to grow, which is a bigger argument than this branch.

### DIFF
--- a/Sources/Vorssaint/Core/Defaults.swift
+++ b/Sources/Vorssaint/Core/Defaults.swift
@@ -287,6 +287,7 @@ enum DefaultsKey {
     static let panelShowUtilities = "panelShowUtilities"
     static let panelShowControls = "panelShowControls"
     static let panelShowToggles = "panelShowToggles"
+    static let panelShowNowPlaying = "panelShowNowPlaying"
     // Quick toggles tab: per-action visibility (the order lives in panelToggleOrder).
     static let panelToggleDarkMode = "panelToggleDarkMode"
     static let panelToggleKeyboardLight = "panelToggleKeyboardLight"
@@ -1105,6 +1106,7 @@ enum Defaults {
         DefaultsKey.panelShowUtilities: true,
         DefaultsKey.panelShowControls: true,
         DefaultsKey.panelShowToggles: true,
+        DefaultsKey.panelShowNowPlaying: true,
         DefaultsKey.panelToggleDarkMode: true,
         DefaultsKey.panelToggleKeyboardLight: true,
         DefaultsKey.panelToggleMicMute: true,
@@ -1829,6 +1831,39 @@ enum Defaults {
             let lower = item.lowercased()
             guard !item.isEmpty, seen.insert(lower).inserted else { continue }
             result.append(item)
+        }
+        return result
+    }
+
+    /// Panel sections added after orders started being saved, and the section
+    /// each one belongs after. Keys are `PanelSectionID` raw values, which are
+    /// the identifiers already persisted in the saved order.
+    static let panelSectionPredecessors: [String: String] = [
+        "disk": "network",
+        "controls": "utilities",
+        "brightness": "keepAwake",
+        "nowPlaying": "mixer",
+    ]
+
+    /// Merges a saved section order with the canonical one. A section that
+    /// saved order predates lands next to the section it belongs after rather
+    /// than at the end, where a user who has reordered their panel once would
+    /// never find it. `after` carries that rule for every section added since
+    /// orders started being saved; four copies of this arithmetic used to sit
+    /// in `PanelLayout.order`, one per section.
+    static func sanitizedSectionOrder(_ raw: String,
+                                      canonical: [String],
+                                      after: [String: String]) -> [String] {
+        let saved = raw.split(separator: ",").map(String.init).filter(canonical.contains)
+        var seen = Set<String>()
+        var result: [String] = []
+        for id in saved where seen.insert(id).inserted { result.append(id) }
+        for id in canonical where seen.insert(id).inserted {
+            if let predecessor = after[id], let index = result.firstIndex(of: predecessor) {
+                result.insert(id, at: index + 1)
+            } else {
+                result.append(id)
+            }
         }
         return result
     }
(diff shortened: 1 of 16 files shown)


vorssaint:
closing this one. the panel would be a third place in the app showing the current track, and the island that landed the day after this was opened already shows it on any mac with play, pause, next, previous, seek and the queue, without asking for accessibility
the buttons here work by pressing a media key for the user, so they need accessibility just to play or pause, while the island already sends those commands through a path that needs no permission
once the section is installed the app keeps checking the accessibility permission in the background, every 2.5 seconds until it is granted and every minute after that, for a section that only does anything while the panel is open
while the track is still being read the section says nothing is playing, so every first open shows a wrong answer for up to two seconds
every panel open runs the track read again, twice on the first open, even when the user came for another section
a now playing section in the panel could still come back if it reads and controls playback through the same music service the island uses, with no media key press and no accessibility prompt

Issue #1599, opened 2026-09-12
[Feature Request] Support Delete key to move files to trash


fraluc06:
### Problem or Motivation

Add support for using the standalone Delete key to move selected files to the trash, in addition to or as an alternative to the current shortcut (such as Cmd + Backspace / Cmd + Delete).

Currently, moving files to the trash requires keyboard shortcuts often tailored to macOS conventions (e.g., Cmd + Delete or Cmd + Backspace). For users accustomed to standard Windows File Explorer behavior, pressing the dedicated Delete (Del) key is the intuitive and expected action to delete or trash an item.

### Proposed Solution

- Allow pressing the single Delete (Canc) key on selected items to trigger the "Move to Trash" action.

- Keep Cmd + Backspace working while adding support for the forward-delete / Canc key when available.

### Feature Area

Clipboard and files

### Does another app already do this?

Supercharge by Sindre Sorhus

### Mockups / References

<img width="330" height="593" alt="Image" src="https://github.com/user-attachments/assets/f0f75905-ca7f-4e17-876a-fbfe404f6cd8" />

danielerognini:
Working on this. Reading your request as the forward-delete / Canc key (⌦) specifically, not Backspace — ⌘⌫ stays untouched and keeps working. Note for anyone on a MacBook: there's no dedicated ⌦ there, it's fn+⌫.

This is the same ask as one line of #1161 ("Delete moves selected items to Trash"), so it may be worth linking the two.

Will comment here when there's a build you can try.

danielerognini:
There's a build you can try now: #1606.

It adds an opt-in toggle so ⌦ moves the Finder selection to the Trash. It's off
by default — after installing, turn it on under **Settings → Finder shortcuts →
"Move to Trash with the ⌦ key"**. 
⌘⌫ is untouched and keeps working alongside it,
as you asked.

If you'd like to test it before it's merged (macOS 14+, Apple Silicon, Xcode Command Line Tools):

```sh
git clone -b feat/finder-forward-delete-trash https://github.com/danielerognini/vorssaint-utils.git
cd vorssaint-utils
./build.sh --install
```

What would help most is confirming it behaves the way you expected on your own
Mac — especially that ⌘Z puts a file back afterwards, and that ⌦ still edits text
normally when you're renaming a file. 
If anything feels off, or if you actually meant plain ⌫ rather than the forward-delete key, 
now is the cheap moment to say so.

[...]

Issue #1161, opened 2026-08-30
[Feature]: Add advanced Finder navigation and keyboard enhancements


RuanMD:
## Feature request

Finder navigation could be significantly faster with additional configurable keyboard and mouse behaviors.

Many common operations require different shortcuts or contextual actions compared with other file managers.

### Proposed Solution

Please add optional Finder navigation enhancements.

Suggested features:

- Return opens the selected file or folder
- F2 renames the selected item
- Delete moves selected items to Trash
- Backspace/Delete can navigate to the parent folder
- Shift + Command + T reopens the last closed Finder tab
- Middle-click a sidebar folder to open it in a new tab
- Configurable shortcut to open selected folder in a new tab
- Configurable shortcut to duplicate the current Finder tab
- Configurable shortcut to focus the path field

Each behavior should be individually enabled or disabled.

It would also be useful to allow per-shortcut customization so the feature does not conflict with existing macOS or user shortcuts.

### Feature Area

Other

PathGao:
Rename is already in the app. Open the Finder settings page, turn it on, and F2 renames the selected item. If F2 clashes with something else on your Mac you can pick a different key there, and it warns you if the one you pick is already taken. Cut and paste of files lives on that same page.

The Return key one is already being tracked in #867, so follow that issue for it.

The rest aren't there: Delete to Trash, Backspace to go up a folder, and the tab ones (reopen the last closed tab, duplicate a tab, middle-click a sidebar folder). Nine asks in one issue tends to sit still, because there's no single thing to build first. Which one do you miss most in a normal day? Name that one and it can be looked at on its own.

danielerognini:
Re the "which one matters most" question: I'm taking Delete to Trash only, tracked separately as #1599, which has its own reporter.

I'm deliberately leaving plain ⌫ unbound so the Backspace-goes-up ask stays available. One technical note for whoever picks that up: on MacBooks ⌦ is fn+⌫, so binding trash to ⌦ and go-up to ⌫ puts a destructive and a navigation action on the same physical key. Also Finder already has ⌘↑ for the enclosing folder.

Planning to ride the existing Finder keyboard hook in FinderCutPaste rather than add another event tap.

PR #1606, opened 2026-09-13
feat(finder): move the selection to the trash with the forward delete key

danielerognini:
## Summary

Adds an opt-in Finder shortcut: with a file or folder selected, a bare forward
delete (⌦) moves it to the Trash. Off by default, under the existing Finder cut
and paste feature. ⌘⌫ is untouched and keeps working.

Rather than trashing the files itself, it substitutes the ⌘⌫ that Finder already
implements. That keeps undo, Put Back, the locked-item and authentication
prompts and the Trash sound as Finder's own behaviour, and it is a much smaller
change than a second selection-reading path — the app already has two.

**Why it belongs here.** It stands on a public CGEvent tap plus a Finder
shortcut that has existed for decades, with no private enumeration path to be
taken away. It drags in no runtime dependency, and needs no permission beyond
the Accessibility grant `finderCutPaste` already requires. It exposes nothing
new: it rewrites a keystroke and never deletes anything itself. The permanent
surface is one defaults key, one toggle and two strings, on a service and an
event tap that already exist — no new service, no new tap, no new permission.

**Prior art.** The issue cites Supercharge. I checked: its toggle binds plain ⌫,
replacing ⌘⌫. I've deliberately diverged and bound ⌦ instead, for two reasons.
The reporter asked for forward delete in their own words ("the single Delete
(Canc) key", "the forward-delete / Canc key when available"). And plain ⌫ is
separately requested in #1161 to navigate to the parent folder, so claiming it
here would foreclose that ask. Leaving ⌫ unbound costs nothing and keeps both
possible.

**Scope.** #1161 carries nine separate asks. This covers exactly one of them,
"Delete moves selected items to Trash", which was filed again on its own as
#1599. The other eight are untouched.

## Verification

`./build.sh` finishes without warnings, `--selftest` passes, and `./build.sh
--test` passes with 32,349 checks.

New checks cover the claim conditions, the modifier family, the Fn case and the
press/release pairing. I confirmed each fails with its logic reverted rather
than only passing as written: treating Fn as a modifier fails 1 check, reducing
the predicate to `return enabled` fails 5, and disabling the release pairing
fails 2.

Tested on Macbook Air M5 32GB 1TB, macOS 26.4.1 (25E253), integrated liquid retina display, on my
own signed build (`./Tools/setup-signing.sh` + `./build.sh --install`), app
language Italian. 
I checked: 
- ⌦ trashes the selection and ⌘Z puts it back;
- repeated presses in a row all work; 
- a real ⌘⌫ still works immediately after;
- ⌦ inside an inline rename still edits the name.

I could not test an external keyboard with a dedicated ⌦ key, a second display, other Mac models, macOS versions other than mine.

I wrote the English and checked the Italian myself. The other eleven
translations I have not verified with a native speaker (they've been generated using Claude Opus 5).

## Anything else

**The release matters as much as the press.** The first version rewrote only the
key down, which left ⌫ held as far as the window server was concerned — every
later ⌘⌫ then read as a repeat of a key already down, including the user's own,
until some other key was pressed. The tap therefore also masks `.keyUp` and
substitutes the release, claiming it from the press rather than by re-testing
the conditions, so switching app or turning the feature off mid-keystroke cannot
strand the key. The cost is that this shared tap now sees key releases
system-wide; #1565 made the same trade for the same reason.

**Checked `FinderRenameService`, no symptom.** It rewrites F2 to Return with the
same key-down-only mask, so the same press/release asymmetry exists there in the
source. I tried to reproduce a stuck Return against it and could not — pressing
F2 then Return immediately commits the rename on the first press. Whatever makes
the substituted ⌘⌫ sensitive to this does not appear to affect Return, so I've
left it alone.

**Quick Controls row?** `finderPasteImageAsFile` is a sub-toggle of the same
feature and has no row in the menu bar panel, so I followed that and kept this
in Settings only. Happy to add one — probably alongside a row for paste-image,
so all three sub-toggles of `finderCutPaste` stay consistent.

**Conflicts.** #1565 touches `CutPasteSettings.swift`, `FeatureCatalog.swift`
and `Defaults.swift` in the same hunks. Trivial either way round.

Refs #1599
Refs #1161


Sources/Vorssaint/UI/Settings/CutPasteSettings.swift
```diff
@@ -116,7 +127,7 @@ struct CutPasteSettings: View {
             if needsAccessibility, !permissions.accessibility {
                 Section(l10n.s.permissionRequired) {
                     PermissionRow(kind: .accessibility)
-                    if AppFeature.finderCutPaste.isAvailable, enabled {
+                    if AppFeature.finderCutPaste.isAvailable, enabled || forwardDeleteTrash {
```
vorssaint, line 130: with only the trash key on, this tells the user the mac will ask for permission to control finder, but rewriting the keystroke never asks finder for anything. can this note stay tied to the cut and paste switch alone

Tests/MetricsTests.swift
```diff
@@ -15609,6 +15670,9 @@ struct MetricsTests {
         expect(activeSet(.automationFinder, on: [DefaultsKey.finderPasteImageAsFile])
                 == [.finderCutPaste, .uninstaller, .quickToggles],
                "pasting copied images as files engages the shared Finder feature")
+        expect(activeSet(.automationFinder, on: [DefaultsKey.finderForwardDeleteTrash])
+                == [.finderCutPaste, .uninstaller, .quickToggles],
+               "trashing with the forward delete key engages the shared Finder feature")
```
vorssaint, line 15675: with only the trash key on, the permissions page would list cut and paste as using the permission to control finder, and this key never asks finder for anything. can that permission count only while cut and paste or the image paste is actually on

Sources/Vorssaint/UI/Settings/CutPasteSettings.swift
```diff
@@ -56,6 +57,16 @@ struct CutPasteSettings: View {
                 }
                 .settingsSectionAnchor(.finderCutPaste)
 
+                Section {
```
vorssaint, line 60: this lands between the cut and paste switch and its how to use steps, so the steps read as if they explained the trash key. can it move below the steps, before the rename section

Sources/Vorssaint/Core/Localization.swift
```diff
@@ -2443,6 +2447,8 @@ extension Strings {
         cutPasteEnableCaption: "Use ⌘X to cut and ⌘V to move files and folders in Finder.",
         cutPasteShowHUD: "Show floating panel",
         cutPasteShowHUDCaption: "Display a floating indicator with the cut files while Finder is active.",
+        forwardDeleteTrash: "Move to Trash with the ⌦ key",
+        forwardDeleteTrashCaption: "In Finder, ⌦ moves the selected items to the Trash, and ⌘⌫ keeps working. Keyboards without a dedicated ⌦ key produce it with fn+⌫.",
```
vorssaint, line 2451: the rename caption on this same page writes the fn key with a capital f and a hyphen, and this one writes fn+⌫. can the two use the same style

Issue #1057, opened 2026-08-27
Fan Control Returns to System Control on Beta 3


logane31:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Fan control never worked, noticed it sasy fixed on beta 3 notes, updated, restarted, and the fan control panel still says "The fans returned to system control because the requested control could not be verified." I also uninstalled the fan control tool and reinstalled it, that did nothing. M4 MacBook Pro

### Steps to Reproduce

1. Open App
2. Click on Icon in Menu Bar
3. Go to fan control section

### Feature Area

Tools

### Screenshots / Recordings

<img width="404" height="358" alt="Image" src="https://github.com/user-attachments/assets/10ffbf36-7d86-4936-813b-726c7586cbbf" />

<img width="884" height="983" alt="Image" src="https://github.com/user-attachments/assets/586c3320-b0c1-4062-b40c-4777616f5beb" />

### Vorssaint Version

3.3.3-beta.3

### macOS Version

macOS 26 (Tahoe)

### Language

English

### Mac and displays

MacBook Pro M4 Pro 14"

### Installation Method

Terminal

PathGao:
That message covers two different situations and I can't tell from here which one you're in: either the background fan controller isn't running on your Mac, or it is running and the Mac refused the speed change. The beta 3 note may have been about the other one.

Two things that would separate them. In that same panel, do you see live fan speed numbers, or are they blank or zero? And in System Settings > General > Login Items & Extensions, under "Allow in the Background", is Vorssaint listed and switched on — and is there more than one Vorssaint entry there? You installed from Terminal and then removed and reinstalled the fan tool, so an old registration could still be sitting there.

Also tell me whether fan control ever worked for you on an earlier version, or whether it has failed since you first installed.

<details>
<summary>The technical side — you can skip this</summary>

`FanControlSection.swift:285` maps both `.helperUnavailable` and `.controlFailed` to the same string, so this report cannot distinguish "helper never came up" from "SMC write went through but verification failed". Worth splitting the two messages regardless of this issue — every future report of this shape is ambiguous the same way.

[...]

PR #1607, opened 2026-09-13
fix(fan-control): stop requiring the Intel force-test key on Apple Silicon

Yash121l:
## Summary

On a Mac with no `Ftst` SMC key, starting fan control fails outright instead of falling back.

`startCooling` tries a direct manual-mode plus target write first. When that does not stick within 0.25s it falls back to the sequence the comment right below it describes: force-test on, protective maximum target, 3s settle, then a persistent manual-mode write. That sequence is written specifically for "a stopped Apple Silicon fan can reject manual mode". The whole fallback sat behind `guard let forceTest = forceTestKey()`, so on a Mac without the key the guard threw `operationFailed`, the helper rolled the fans back to automatic, and the panel showed "The fans returned to system control because the requested control could not be verified."

`Ftst` is an Intel-era force-fan-test key. I dumped every `F*` SMC key on a MacBook Pro M3 and it is not there, while `FNum` and `F0Ac/Mn/Mx/Tg/Md` all are. So the one recovery path meant for Apple Silicon was unreachable on Apple Silicon.

This is the odd one out rather than a new rule. The other five `forceTestKey()` callers already treat a missing key as nothing to do: the precheck in `startCooling`, `readOnlySnapshot`, `validateAutomaticControl`, and both reads in `restoreAutomatic`. This was the only site where absence was fatal. A key that is present but refuses the write is still a failure, so nothing changes on Intel.

The decision moved into `FanControlPolicy.forceTestSatisfied` because `FanControlHardware` is not in the `--test` source list and `SMCClient` has no seam. The policy enum is compiled by the tests and already holds `isAutomaticMode` and `targetRPMMatches`.

## Verification

MacBook Pro M3, macOS 26.6.2, single internal fan, my own ad-hoc build, English.

```sh
./build.sh --test              # TESTS OK (32334 checks)
```

The new assertion goes red if the predicate is flipped back to the old `keyExists && writeSucceeded`. I checked it in both directions.

What I could not test: I did not install the privileged helper and made no SMC writes, so I have not yet watched a real fan change speed with this in place. The key-absence evidence is from read-only SMC reads. I do not have an M4 Pro, and I cannot prove this is the specific throw #1057 hits, since `startCooling` has four other throw sites after the direct write. What I can show is that on Apple Silicon the fallback could never run at all. I will report back once I have run it against the helper on my own M3.

## Anything else

Refs #1057

#1057 also asks whether the uninstall and reinstall left something behind. This does not address that part.


Sources/Vorssaint/Services/FanControl/FanControlSupport.swift
```diff
@@ -188,6 +188,12 @@ enum FanControlPolicy {
             && abs(target - expected) <= max(2, expected * 0.001)
     }
 
+    /// `Ftst` is an Intel-era override that Apple Silicon does not expose, so
+    /// a Mac without the key has nothing to force and nothing to fail at.
```
vorssaint: this reads as if the key were intel only, but the m4 in the linked report is documented to expose it and to need it before manual mode sticks, and intel macs force fans through a different key. the m5 is the one reported without it. could this say only that some macs do not expose the key, without calling it intel only? otherwise someone may later drop the unlock path on apple silicon because of this line

Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
```diff
@@ -119,7 +119,10 @@ final class FanControlHardware {
         }
 
         if !directSucceeded {
-            guard let forceTest = forceTestKey(), setByte(1, for: forceTest, attempts: 20) else {
+            let forceTest = forceTestKey()
+            let forced = forceTest.map { setByte(1, for: $0, attempts: 20) } ?? false
+            guard FanControlPolicy.forceTestSatisfied(keyExists: forceTest != nil,
+                                                      writeSucceeded: forced) else {
```
vorssaint, line 125: the app checks for this key by name, not through the key listing, and the absence in the description came from a listing. could you confirm on your m3 that the by name lookup also comes back empty, and run the stopped fan case against the background fan tool as you planned? without the key, a mac that keeps refusing manual mode now waits about thirteen seconds before showing the same failure it used to show right away, so this path needs to be able to succeed somewhere

Sources/Vorssaint/Services/FanControl/FanControlSupport.swift
Yash121l: You are right, that line overclaimed. Fixed in ed4941d:

```
/// Not every Mac exposes `Ftst`. Where it is present the unlock write has to
/// succeed; where it is absent there is nothing to force and nothing to fail.
```

No mention of Intel or Apple Silicon either way, so the unlock path reads as something that still has to work wherever the key exists. The first commit subject still says "Intel force-test key". I left it rather than force-push over the SHA your comments are anchored to; I will reword it if you prefer, or it can go in the squash subject.

Sources/Vorssaint/Services/FanControl/FanControlHardware.swift
Yash121l: Confirmed by name, not from the listing. I rebuilt the app's own lookup (`SMCClient.key(named:)`: `SMCParamStruct` with `data8 = 9` over `IOConnectCallStructMethod(connection, 2, ...)`) as a read-only probe on this M3, macOS 26.6.2 (25G83). No writes.

```
IOServiceOpen(AppleSMC) = 0
Ftst  key info: kern_return_t = 0, SMC result byte = 0x84 (kSMCKeyNotFound) -> key(named:) returns nil
FNum  result 0, ui8 = 1
F0Ac  result 0, flt = 0.0 (0 RPM)      F0Mn 2317.0   F0Mx 6800.0   F0Tg 0.0   F0Md 0 (automatic)
```

The IOKit call succeeds; the SMC reports the key absent. The controls all resolve.

On the thirteen seconds: your number is right, about 13.25s. For a Mac with no `Ftst` that never accepts manual mode: 0.25s post-write mode check, the protective max target write (returns on first accepted write), `Thread.sleep(3)`, then `setMode(1, untilUptime: now + 10)` burning its full 10s before `operationFailed`. Bounded, but 13s of waiting for the same failure that used to appear at once.

I have not run the stopped-fan case against the helper yet. It needs the privileged helper installed and real fan writes, so I will do it with the `--dev` build (its own `com.vorssaint.utils.dev.fan-control` label, so it cannot disturb the installed registration) and report timing and `log stream` output. This machine idles at `F0Ac` = 0, so the stopped state is easy to reach. One thing to flag: the helper only logs `Cooling session started` and the restore lines, so direct versus fallback is only distinguishable by latency. I will add a `log.debug` on entry to the fallback branch if you want it diagnosable from a user log.

PR #1610, opened 2026-09-13
feat(authenticator): add two-factor codes with a palette and imports

PersonalJesus:
## What a user notices

A new installable feature under Tools, **Authenticator**, shipping
uninstalled like other new opt-in features. It holds the two-factor
accounts a person already keeps on their phone and puts the code where it
is actually needed:

- **Scan a QR straight off the screen.** Drag over the code a site is
  showing, on the same selection surface Screen OCR uses. Also from the
  Mac's camera, the clipboard, or a file.
- **A palette on a shortcut** (⌃⌥⌘O by default). Type to search, Return
  types the code into the app you came from, ⌘Return copies it, ⌥Return
  shows the code that comes next.
- **Search, pinning and drag-to-reorder** in Settings, a tile in the menu
  panel, and one row per account in the Command Bar.
- **Import and export** as `otpauth://` links, Google Authenticator
  transfer codes, unencrypted Aegis and 2FAS exports, and a
  password-protected backup of its own. Export can show transfer QR codes
  for moving accounts back to a phone.
- **TOTP, HOTP and Steam Guard**, SHA1/256/512, six to eight digits.
- Copied codes carry the concealed-pasteboard mark, so Clipboard History
  skips them, and are cleared again after a delay. Touch ID before showing
  codes is an option, off by default.

Secrets live in the Keychain in one generic-password item, `ThisDeviceOnly`.
Names, order and parameters live in `Authenticator.json` in the app's
container. Nothing syncs and nothing leaves the Mac. Settings backup
deliberately carries neither.

## Reading it

Four commits, bottom up, each building and testing on its own. Reviewing by
commit is much less work than reading the diff:

| Commit | What to look at | `--test` |
| --- | --- | --- |
| `add the one-time password engine` | The only cryptography. Codes, the otpauth URI both ways, Google's transfer protobuf. Checked against the RFC vectors. | 38242 |
| `store accounts in the keychain` | Where secrets go, what happens when the Keychain says no, and why the countdown clock is its own object. | 38263 |
| `read aegis, 2fas and encrypted backups` | Other apps' file formats and the encrypted backup. | 38289 |
| `wire two-factor codes into the app` | Service, settings page, palette, panel tile, Command Bar, hub registration. The only commit touching existing files. | 38445 |

Existing files touched at all: the feature catalog, defaults, the shortcut
roles, the feature runtime, the settings directory and hub, the panel
layout, the Command Bar catalog, `build.sh`'s test source list and the test
runner's suite list. Nothing else changes behaviour.

The branch sits on current `main`, including the test harness rework; the
feature's own checks are a selectable suite, so `./build/metrics-tests
--suite=authenticator` runs them alone.

## Why it is shaped like this

**What carries it.** RFC 4226 and RFC 6238, HMAC through CryptoKit, QR
reading through Vision. Apple API or a frozen IETF standard throughout;
nothing rides a private path. Google's transfer format is a six-field
protobuf message written out by hand rather than pulled in as a dependency.

**What it drags in.** Nothing. No package, no helper, no new signing
target. The camera permission is the only addition, asked for at the moment
the camera scanner opens.

**What it exposes the project to.** This is the honest one: the app becomes
the place a person's second factor lives, and losing it locks them out of
their accounts. That is why the metadata file and the Keychain item are
kept apart, why export exists in three shapes including an encrypted one,
and why a QR for moving accounts back to a phone is a first-class action
rather than an afterthought.

**How many people, how specific.** Anyone who signs in to anything with a
code on their phone. The permanent surface is one service, one settings
page, one optional shortcut, one permission and a string catalog.

## What the other implementations do, and where this follows them

**Steam accounts are written three different ways in the wild**, and
reading only one of them is how an import silently loses an account. Aegis
puts the kind in the host, `otpauth://steam/Name?secret=…`. 2FAS and gopass
keep `otpauth://totp/` and add `encoder=steam`. Some tools write
`representation=steamguard` instead. All of those are accepted here,
including an issuer of `Steam` on its own, and an export writes the Aegis
host form with `encoder=steam` beside it so either reader takes it back.
This one came out of looking rather than reasoning: the first version of
the parser rejected Aegis's form outright.

**Digit counts other than six are real**, not hypothetical: Steam uses
five, Twitch seven, Microsoft eight. The editor offers six to eight and the
parser takes five through nine.

**A secret with no padding has to parse.** Aegis writes base32 without the
`=` padding, so padding is optional on the way in, along with spaces,
dashes and lower case. That is one of the base32 cases in the tests.

**The concealed pasteboard type** (`org.nspasteboard.ConcealedType`) is the
convention password managers write so a clipboard history skips a secret.
This app already honours it on the reading end, in
`ClipboardHistorySensitiveText`; the authenticator now writes it, so the
two halves meet.

Transfer QR codes are batched at ten accounts so each stays scannable.
Return typing the code with ⌘Return copying it, rather than the other way
round, is the arrangement the shortcut-driven authenticators use; both are
there and which one Return does is a setting.

## Verification

MacBook Pro 18,3, M1 Pro, macOS 15.0 (24A335), single built-in display, app
language following the system (Russian). Built with `./build.sh --dev`
against a stable local signing identity, installed and lived with for three
days with nineteen real accounts, all of them plain SHA1 six-digit TOTP.

CI's three commands run clean locally on every commit in the series:
`./build.sh`, `./build/Vorssaint --selftest`, `./build.sh --test`. The
`build.sh` change is five lines in `TEST_SOURCES`: the four pure-logic
sources and the string catalog, which the generated localization catalog
needs in order to compile. Test files themselves are picked up by the
existing `Tests/*.swift` glob.

The two `ShelfService.swift` warnings a local release build prints are on
`main`, untouched by this branch.

Exercised on that Mac: installing and uninstalling the feature in the hub,
adding accounts, the settings list against nineteen of them, drag
reordering, copying a code, the Keychain prompt on first read and on
rebuilds, and the settings page's redraw cost. That last one is why the
countdown clock is a separate observable object: with it published on the
service, the whole page including every row menu rebuilt once a second, and
dragging a row stuttered against it.

Left behind that goes red: the RFC 4226 and RFC 6238 reference vectors for
all three digests, Steam code shape, base32 edge cases, `otpauth://` round
trips in all three Steam spellings, a hand-assembled Google transfer
payload decoded without the writer's help, Aegis and 2FAS parsing with rows
that must be skipped, and the store's behaviour when its file is corrupt
or the Keychain refuses. The backup's key derivation is pinned to the
published PBKDF2-HMAC-SHA256 vectors rather than to itself: a round trip
alone stays green with the round count ignored or the password and salt
swapped, and both of those now go red.

**Not exercised on hardware, only covered by tests or read:** HOTP and
Steam accounts, SHA256 and SHA512, seven and eight digit codes, the camera
scanner, the Touch ID option, the encrypted backup through the UI, the
Aegis and 2FAS readers through the UI, typing a code into another app's
field, the Command Bar rows and the palette shortcut. Also untested: an
Intel Mac, an external display, and more than one Space.

**Eleven of the thirteen languages read English here.** English and Russian
are written; the rest return the English catalog, which the localization
checks accept because the schema, the emptiness rule and the format
arguments all hold. Given how much this already asks of a reviewer I would
rather not add another thousand lines of strings before anyone has said the
feature is wanted. Say the word and they go in.

## Two things worth saying out of the gate

**This is large, and I know what that costs here.** The contributing guide
puts the merged median near a hundred and ten lines and says the useful
instruction for something this size is to find the smallest piece that
stands on its own. That piece is the first two commits plus a settings
list, around two thousand lines, since below that the feature cannot
produce a code. Say the word and I will reopen it that way with the rest
behind it. Whole pieces that come out cleanly on their own: the camera
scanner, the Aegis and 2FAS readers, the encrypted backup, and the Touch ID
option.

**The direction was not settled first, and it should have been.** There is
no open issue or pull request about two-factor codes and the enhancement
summary does not rule it out, but that is not the same as a yes. If the
answer is no, say so and this costs nothing further.


vorssaint:
closing this one. the direction is a no, so there is no need to split it or to add the missing languages.
macos already keeps verification codes in its passwords app, sets them up from a qr code and fills them in, and this app reaches for what the system almost does rather than growing a subsystem of its own.
with it the app becomes the place a person's second factor lives, with its own crypto, keychain item, backup format and camera scanner to carry, and losing it locks people out of their accounts. holding people's credentials has been turned down here before.
the readers and writers for other authenticator apps' formats are hardcoded integrations with particular third party products, which have been declined here too, and their names would sit in the code and in the visible text.
the smaller cut you offered still makes the app hold the second factor, so it does not change the answer. the engine with its reference vectors is careful work and would stand well as a project of its own

Issue #907, opened 2026-08-24
Window sizing shortcut key to cycle between 1/3, 2/3, 1/2 widths


rmanalan:
### Problem or Motivation

Having to set and remember shortcut keys for each window width size is annoying and the #1 reason I probably won't use the window resizer feature.

### Proposed Solution

Raycast's window resizer has this nice feature that uses the same shortcut key to cycle between window sizes. It would be great to add that here so we don't have to set up dedicated shortcut keys for each window width.

### Feature Area

Windows and Dock

johnpark-bin:
<img width="862" height="793" alt="Image" src="https://github.com/user-attachments/assets/31b4df2b-7ef3-44cd-b9c4-e344e95c1ae2" />

In my case, I use an app called Rectangle, and I wanted to chime in because the requested feature seems similar to its 'Repeated Commands' option. 

I would really love to see this feature added as well—having it makes a huge difference in making a window manager simple, intuitive, and convenient to use.

Advaith3600:
I use a Windows machine and I'm very used to the Windows key + arrow keys cycling through multiple window layouts. Having a single shortcut that cycles through sizes (for example 1/3 → 2/3 → 1/2) would make the window resizer much easier to adopt because people wouldn't need to set and remember separate shortcuts for each size.

One extra request: while cycling it would be great if the window could be moved to the adjacent monitor as well (i.e., include moving to the next screen as part of the cycle), which would mirror the multi-monitor behavior users expect on Windows.

iva-zhu:
Chiming in with the same friction — I'd like repeated presses of the same shortcut to cycle the width: half → two-thirds → third → back to half, the way Rectangle does it (and Raycast).

Proposed shape, to keep existing behaviour intact:

- An opt-in toggle in Window Layout settings, off by default.
- Repeated same-side presses keep the current display-crossing semantics whenever a neighbouring display exists in that direction (the behaviour #1666 covers); the cycle only kicks in when there is nowhere to push the window — e.g. a single display, or the edge of a multi-display setup.
- Top and bottom keep their existing maximize-on-repeat rule.

I'm preparing a PR along these lines that refs this issue.

iva-zhu:
Update: I noticed #1632 already implements exactly this (opt-in toggle, half → two thirds → third on repeat, top and bottom untouched). Withdrawing my "preparing a PR" note above and supporting that one instead — it matches what I described, including the ladder order I'd want. Thanks @MakhBeth!

PR #1612, opened 2026-09-13
feat(window-layout): cycle repeated directional commands through fractions

kanavdhanda:
## Summary

Adds an opt-in repeated execution mode for directional Window Layout shortcuts (`leftHalf`, `rightHalf`, `topHalf`, `bottomHalf`, and `centerHalf`). When enabled in **Settings → Window Layout → Repeated commands**, pressing the same directional command repeatedly cycles through `1/2 → 1/3 → 1/6` before wrapping back to `1/2`.

- **Opt-in & Zero Runtime Drag**: Defaults to `disabled` so existing single-press behavior is preserved.
- **Candidate Frame Matching**: Determines the next fraction by comparing the window's current frame against candidate fraction rects. If a window was manually repositioned or a different shortcut is pressed, it cleanly starts back at `1/2`.
- **Localization**: Full translation catalog provided across all 13 supported languages.

## Verification

- **Mac**: MacBook Pro 14-inch (Nov 2023, Apple M3 Pro)
- **macOS**: 26.6.2
- **Display**: Built-in Liquid Retina XDR display (single display)

```sh
./build.sh
./build/stage/Vorssaint.app/Contents/MacOS/Vorssaint --selftest
./build.sh --test --test-suite=core
./build.sh --test --test-suite=localization
```

All 7,989 core checks and 3,707 localization checks passed with 0 failures, and `--selftest` completed with `SELFTEST OK`.

## Anything else

Refs #907


vorssaint:
closing this. the same request (#907) is going ahead through the smaller version in #1632, one toggle that keeps the current behaviour as the default, and this one would need a rewrite to get there
with the new setting at its default, pressing left or right twice no longer moves the window to the next display, which is what happens today
left half followed by left third lands the window on two thirds, and left third on a window already sitting there jumps to left half, so the explicit third commands stop meaning what they say
when an app refuses the next size the window is accepted where it is and the next press asks for that same size again, so the cycle gets stuck and never comes back to the half
a placement that moved the window without resizing it now counts as success and a failed one is no longer put back where it was, which undoes the recovery for windows that moved but never resized
eight modes in a menu, two separate cycling paths and top or bottom shrinking to a sixth are far more surface than the request needs

PR #1613, opened 2026-09-13
fix(screenshot): let the visibility preference reach editor and pin windows

anandghegde:
## Summary

With "Hide Vorssaint windows" switched **off**, the app's own screenshot editor
and pinned-capture windows still never appear in a capture. The Settings window
does, which matches the clarification on the issue.

`ScreenshotService.protectedWindowIDs` unioned every window the app puts on
screen — selection overlays, the countdown and scrolling HUDs, editors, pins and
the quick preview — and the capture paths then consulted the preference through
`ScreenshotCapturePolicy.excludedWindowIDs`, which excludes everything when the
preference is on and *otherwise only the protected list*. So with the preference
off, the protected list re-excluded the editors and pins anyway. The switch could
never reach them.

The protected list now splits by role:

- **Workflow surfaces** — the screenshot selection overlay, the shared capture
  tool's overlay, the countdown and scrolling HUDs, and the quick preview — are
  the tool taking the capture, never its subject. They stay out whatever the
  preference says.
- **Content windows** — editors and pinned captures — are ordinary windows
  somebody left on screen, so the preference owns them.

The quick preview is deliberately on the workflow side: it dismisses itself after
a few seconds, so on the content side back-to-back captures would photograph the
previous capture's toast.

**Recording behaviour is unchanged.** `ScreenshotSupport.unifiedCapturePolicy`
exempts recording from this preference on purpose (`tool != .recording`), and the
toggle lives in the Screenshot settings pane, so routing it into the recorder
would be a direction change rather than a fix. Both recorder paths therefore ask
for both kinds. **If you would rather the preference did reach recordings, that
is a one-word change at each call site and I will make it** — flagging it because
it is the open question from the earlier attempt at this.

Callers checked before changing the shared property (all four reach it through
`protectedWindowIDsForCapture` or the internal `protectedWindowIDs`):

| Caller | Tool | Passes |
|---|---|---|
| `ScreenshotService` session / scrolling paths | screenshot | preference honoured |
| `ScreenCaptureService` selection | either, read live from `options.selectedTool` | honoured unless recording |
| `ScreenRecorderService` stream chrome | recording | both kinds kept out |
| `ScreenshotScrollingCapture` | screenshot | via `protectedWindowIDs` |

The producers were each read on `main` — `ScreenshotEditorController`,
`ScreenshotPinController`, `ScreenshotQuickPreviewController` and
`ScreenshotSelectionController` each return only their own window, so neither
bucket picks up anything unexpected.

+99 / -11 across six files, tests included.

## Verification

MacBook Pro (Mac16,7), Apple M4 Pro, 24 GB, macOS 26.6.2 build 25G83, Apple
Silicon. Single built-in display, default Spaces layout, app language English.
Two of my own builds run side by side for the comparison: upstream `main` as the
release variant, this branch as the `--dev` variant, so the only difference
between the two runs is the code. Signed with a stable Developer ID identity, so
the Screen Recording grant survived rebuilds.

```
./build.sh                       ✓ Bundle ready, no warnings
./build/Vorssaint --selftest     ✓ SELFTEST OK
./build.sh --test                ✓ TESTS OK (32,337 checks)
python3 Tests/mutation_checks.py ✓ MUTATION CHECKS OK (8 regressions detected)
```

The mutation entry is the both-directions check: reverting the split to the old
unconditional union fails the new assertion with
`a screenshot protects only the surfaces taking it`.

By hand, "Hide Vorssaint windows" **off**:

- **Editor window.** Region capture overlapping an open screenshot editor. On
  `main` the editor is missing from the result — a hole where it was. On this
  branch the editor is in the shot, toolbar, tool palette and all.
- **Pinned capture.** Region capture over a pin (`⌘P` from the editor). Missing
  on `main`, present on this branch.

"Hide Vorssaint windows" **on**, same two captures on this branch:

- Editor and pin are absent again, so the preference still does what it says.

Also on this branch:

- **Back-to-back captures.** A second capture taken while the previous capture's
  quick preview was still on screen does not contain that preview. This is the
  case that decided which bucket the quick preview belongs in.
- The selection overlay does not appear in any of the captures I kept.
- **Screen recording.** With the preference off, a region recording covering an
  open screenshot editor does not contain that editor — the same as `main`. The
  recorder reaches the protected set through its own accessor and asks for both
  kinds, so the exemption survives.

One display here, so the multi-display and multi-Space cases are the gap between
this Mac and yours. The change is per-window rather than per-screen, which is why
I would expect it to hold, but that is reasoning rather than a result.

## Anything else

Refs #780

This is the change from #1269, which was closed because it had been written on a
Windows machine and never compiled or run. Rewritten and verified on a Mac, with
the quick preview moved to the workflow side and the recorder left alone — the
two points raised in that review.


Sources/Vorssaint/Services/QuickTools/ScreenCaptureService.swift
```diff
@@ -196,10 +196,14 @@ final class ScreenCaptureService: ObservableObject {
...
-                    ? ScreenshotService.shared.protectedWindowIDsForCapture
-                    : []
+            protectedWindowIDs: { [weak options] in
+                guard AppFeature.screenshot.isAvailable else { return [] }
+                // The tool can still change while the selection is up, so it
+                // is read here rather than captured. A session on its way out
+                // leaves no tool, and keeps both kinds out.
+                let tool = options?.selectedTool
+                return ScreenshotService.shared.protectedWindowIDsForCapture(
+                    honoursVisibilityPreference: tool != nil && tool != .recording)
```
vorssaint: with “hide vorssaint windows” off, keep the editor open and switch from recording to screenshot before selecting an area. with freeze on and the pointer off, the screenshot still uses the earlier picture without the editor or pins. switching the other way also leaves the editor selectable, and selecting it records it despite the intended recording exclusion. the text and color tools have the same stale-picture problem after recording. can switching tools refresh both the picture and the selectable windows whenever their visibility changes? add tests covering both directions

anandghegde: switching tools now refreshes both whenever editor/pin visibility changes. `UnifiedCapturePolicy` gained `keepsContentWindowsOut` (true for recording, or when "hide vorssaint windows" is on) and `sharesSource` compares it, so with freeze on and the pointer off, recording <-> screenshot/text/color re-photographs the displays and rebuilds the pickable windows from the new tool's protected set. screenshot/text/color still share the picture they have.

56e3a47. tests in MetricsTests cover recording vs screenshot, text and color in both directions, no refresh between screenshot/text/color, and every tool keeping editors/pins out when the preference is on. added a mutation_checks entry that drops the new comparison and is caught. also merged main (notch capture chrome kept in the protected set). `./build.sh`, `--selftest`, `--test` (38,166 checks) and `Tests/mutation_checks.py` (15/15) pass on macOS 26.

Issue #1622, opened 2026-09-14
Allowing capture subfolder pattern to get app name


Nyzeix:
### Problem or Motivation

I like to have my screenshot "filtered" by folder, especially by app name.

### Proposed Solution

It would be great to have as an option to set the subfolder name as the app name when taking a screenshot. (Like Snapzy do)

### Feature Area

Windows and Dock

### Does another app already do this?

Snapzy

### Mockups / References

Vorssaint Example:
<img width="1077" height="745" alt="Image" src="https://github.com/user-attachments/assets/b4f1a8d5-59c9-4055-b841-2ce20e219d7d" />

in Snapzy:
<img width="1431" height="604" alt="Image" src="https://github.com/user-attachments/assets/20045ef1-b45d-4d22-9a5e-44feae1c5e74" />

Borisserz:
Opened a fix on the fork (PR create from this account is currently blocked, so the compare link is the review entry):

https://github.com/vorssaint/vorssaint-utils/compare/main...Borisserz:feat/screenshot-app-subfolder?expand=1

`%app` works in the existing subfolder and file-name patterns. It expands to the app that was in front when the capture started, as a single path component (`/` and `:` cannot nest extra folders). If that app is Vorssaint itself, the token is left empty so the overlay is not stamped as the folder name.

`./build.sh --test` → TESTS OK (38164).

PR #1625, opened 2026-09-14
feat(screenshot): add %app token for save folders and names

Borisserz:
## Summary
- `%app` expands in the existing screenshot subfolder and file-name patterns to the app that was in front when the capture started.
- The name is one path component (`/` and `:` cannot nest extra folders). Vorssaint itself is never stamped, so the overlay is not the folder name.
- `./build.sh --test` → TESTS OK (38164).

## Test plan
- [ ] Set the subfolder pattern to `%app` and take a screenshot while another app is frontmost
- [ ] Confirm the file lands in a folder named after that app
- [ ] Set the file-name pattern to `%app-%#` and confirm the name includes the app
- [ ] Confirm a capture started from Vorssaint itself does not create a Vorssaint folder

Refs #1622


Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -235,6 +235,7 @@ final class ScreenshotService: ObservableObject {
     }
 
     private func beginCapture(_ mode: CaptureMode) {
+        rememberCaptureApp()
```
vorssaint: an area or window capture taken with the main shortcut never passes through here, so for those %app comes out empty and only a full screen or scrolling capture fills it in. it has to be recorded when the capture picker comes up, for every kind of capture

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -778,15 +780,28 @@ final class ScreenshotService: ObservableObject {
...
         } else {
-            let expanded = ScreenshotSupport.expandFileNamePattern(pattern, date: Date(), number: 0)
+            let expanded = ScreenshotSupport.expandFileNamePattern(pattern, date: Date(), number: 0, appName: captureAppNameForSave)
             return (expanded + ".png", nil)
         }
     }
 
+    /// Frontmost app when the capture gesture started. Saved before the
+    /// overlay takes focus, and ignored when that app is Vorssaint itself.
+    private static var captureAppNameForSave = ""
```
vorssaint: this is one shared slot that always holds the last app recorded, so saving an older editor window or a capture reopened from history files it under whatever app the newer capture saw. the name has to travel with each capture instead

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -778,15 +780,28 @@ final class ScreenshotService: ObservableObject {
...
         if ScreenshotSupport.fileNamePatternUsesNumber(pattern) {
             let number = defaults.integer(forKey: DefaultsKey.screenshotFileNumberNext)
-            let expanded = ScreenshotSupport.expandFileNamePattern(pattern, date: Date(), number: number)
+            let expanded = ScreenshotSupport.expandFileNamePattern(pattern, date: Date(), number: number, appName: captureAppNameForSave)
             defaults.set(number + 1, forKey: DefaultsKey.screenshotFileNumberNext)
             return (expanded + ".png", number)
         } else {
-            let expanded = ScreenshotSupport.expandFileNamePattern(pattern, date: Date(), number: 0)
+            let expanded = ScreenshotSupport.expandFileNamePattern(pattern, date: Date(), number: 0, appName: captureAppNameForSave)
             return (expanded + ".png", nil)
```
vorssaint: when no app name is known, for example after starting the capture from this app's own panel, a file pattern of just %app is saved as a hidden .png file and %app-%# becomes -3.png. can the default name take over when the pattern comes out empty

Sources/Vorssaint/Core/ScreenshotStrings.swift
```diff
@@ -494,9 +494,9 @@ extension ScreenshotFeatureStrings {
         folderLabel: "Kayıt yeri",
         folderChoose: "Seç…",
         subfolderLabel: "Alt klasör deseni",
-        subfolderCaption: "İsteğe bağlı. Yukarıdaki klasörün içinde %y, %year, %mo, %month, %d, %h, %mi ve %s kullanarak tarihli alt klasörler oluşturur, örneğin %y-%mo için 24-03.",
+        subfolderCaption: "İsteğe bağlı. Yukarıdaki klasörün içinde %y, %year, %mo, %month, %d, %h, %mi ve %s kullanarak tarihli alt klasörler oluşturur, örneğin %y-%mo için 24-03. %app, yakalama başladığında önde olan uygulamadır.",
```
vorssaint: the turkish list of tokens here still leaves %app out, while the sentence after it explains it. can you add it to the list like the other languages

Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift
```diff
@@ -1085,18 +1108,28 @@ enum ScreenshotSupport {
...
             .replacingOccurrences(of: ":", with: "-")
     }
+    /// Whether an expanded file-name pattern left nothing usable — empty,
+    /// or only punctuation plus an optional number from "%#" — so the
+    /// default localized name should take over instead of writing ".png"
+    /// or "-3.png".
+    static func expandedFileNameNeedsDefault(_ expanded: String) -> Bool {
+        let trimmed = expanded.trimmingCharacters(in: .whitespacesAndNewlines)
+        if trimmed.isEmpty { return true }
+        return trimmed.range(of: #"^[\W_]*\d*$"#, options: .regularExpression) != nil
```
vorssaint, line 1124: a pattern that comes out as only digits, for example just %# for plain numbered files or %y%mo%d, now gets the default name instead, and with %# the counter no longer moves. only a name that really ends up empty should fall back

Sources/Vorssaint/Services/QuickTools/ScreenshotSelectionController.swift
```diff
@@ -22,6 +22,18 @@ final class ScreenshotSelectionController {
         let scale: CGFloat
         /// The captured area in Cocoa global coordinates.
         let anchorRect: CGRect
+        /// Frontmost app when the capture picker opened; used for %app.
+        let appName: String
+
+        init(image: CGImage,
+             scale: CGFloat,
+             anchorRect: CGRect,
+             appName: String = "") {
```
vorssaint, line 31: a capture reopened later from history comes back through this default with no app, so saving it from there lands outside the app folder and gets the default name. can the app name be kept with the stored capture so it is still there when it comes back

Sources/Vorssaint/Services/QuickTools/ScreenshotSupport.swift
```diff
@@ -1085,18 +1108,28 @@ enum ScreenshotSupport {
...
+        let withDate = applyingDateTokens(pattern, date: date, appName: appName)
         return applyingNumberTokens(withDate, number: number)
             .replacingOccurrences(of: "/", with: "-")
             .replacingOccurrences(of: ":", with: "-")
     }
+    /// Whether an expanded file-name pattern left nothing usable — empty,
+    /// or only punctuation plus an optional number from "%#" — so the
+    /// default localized name should take over instead of writing ".png"
+    /// or "-3.png".
+    static func expandedFileNameNeedsDefault(_ expanded: String) -> Bool {
```
vorssaint, line 1121: the live preview next to the file name field in settings still shows .png or -3.png when the pattern leans on %app, since no app is in front there. can it show the default name in that case, the same way the save now does

Sources/Vorssaint/Services/QuickTools/ScreenshotEditorController.swift
```diff
@@ -1045,6 +1047,7 @@ final class ScreenshotEditorController: NSObject, NSWindowDelegate {
 
     init(capture: ScreenshotSelectionController.Capture) {
         model = ScreenshotEditorModel(image: capture.image, scale: capture.scale)
+        sourceAppName = capture.appName
```
vorssaint, line 1050: a screenshot reopened with edit latest screenshot arrives here with no app once its file is on disk, so saving it from the editor lands outside the app folder and its name loses the app too. can the app name be kept with that stored screenshot, the way history keeps it now

Issue #1566, opened 2026-09-10
Directional ring cannot hold a direction on its default ⌃⌥Space shortcut (Space is hard-coded to maximize)


RichRock27:
### What happens

The window-layout **directional ring** cannot hold a direction while it is opened
with its own default shortcut. Holding the shortcut and moving the pointer toward
a segment makes that segment highlight for a single frame; the centre
(maximize) then re-lights and the edge-snap preview disappears. Repeating the
movement just repeats the flash. Releasing does apply the direction that was
aimed at, so the underlying selection works — only the tracking is unusable,
because there is no way to see what is selected.

### Cause

The default trigger is `⌃⌥Space`:

```swift
// Sources/Vorssaint/Core/GlobalShortcut.swift:169
static let windowDirectionalDefault = GlobalShortcut(keyCode: Int64(kVK_Space),
                                                     modifiers: [.control, .option])
```

and the directional tap treats Space as a request to maximize:

```swift
// Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift:844
if type == .keyDown {
    let keyCode = event.getIntegerValueField(.keyboardEventKeycode)
    if keyCode == 49 || keyCode == 36 || keyCode == 126 { // Space, Return, Up
        session.manualOverride = .maximize
```

Because the gesture is driven by *holding* the shortcut, the Space key
auto-repeats for as long as the ring is open. Every repeat sets
`manualOverride = .maximize`.

Meanwhile the 60Hz timer clears it again as soon as the pointer has travelled
far enough:

```swift
// Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift:878
if distance >= WindowDirectionalGestureSupport.activationDistance {
    session.manualOverride = nil
}
let action = session.manualOverride ?? directionalAction
```

So the override is set by the key repeat and cleared by the timer, over and
over, and the indicator alternates between the aimed segment and the centre.

The feature conflicts with itself on its default binding. The same would happen
for any shortcut using Return (36) or Up (126).

### Workaround

Bind the ring to a shortcut whose key is not Space, Return or Up. Verified fixed
with `⌃⌥⇧⌘R`:

```
defaults write com.vorssaint.utils windowDirectionalShortcut \
  -string 'control+option+shift+command:15'
```

### Possible fix

Ignore the `keyDown` override when the pressed key is the one that triggered the
session — or ignore auto-repeat events (`kCGKeyboardEventAutorepeat`) in
[...]

PR #1626, opened 2026-09-14
fix(window-layout): directional ring ignores hold-key Space repeats

Borisserz:
## Summary
- Holding the directional ring on its default `⌃⌥Space` no longer forces maximize on every Space auto-repeat.
- A distinct Space, Return, or Up tap still maximizes when that key is not the hold trigger. Escape still cancels.
- `./build.sh --test` → TESTS OK.

## Test plan
- [ ] Enable the directional ring with the default `⌃⌥Space`
- [ ] Hold the shortcut and move the pointer to an edge. The segment stays highlighted, with no center flash
- [ ] Release applies the aimed direction
- [ ] With a non-Space shortcut, a Space tap still maximizes

Refs #1566


Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
```diff
@@ -853,15 +855,20 @@ final class WindowLayoutService: ObservableObject {
 
         if type == .keyDown {
             let keyCode = event.getIntegerValueField(.keyboardEventKeycode)
-            if keyCode == 49 || keyCode == 36 || keyCode == 126 { // Space, Return, Up
+            let isAutorepeat = event.getIntegerValueField(.keyboardEventAutorepeat) != 0
+            let allowManual = WindowDirectionalGestureSupport.shouldApplyKeyboardManualOverride(
+                keyCode: keyCode,
+                triggerKeyCode: session.triggerKeyCode,
+                isAutorepeat: isAutorepeat)
+            if allowManual, keyCode == 49 || keyCode == 36 || keyCode == 126 { // Space, Return, Up
```
vorssaint: a held return, up, m or down now leaks its key repeats to the front app while the ring is open, because a skipped repeat falls through instead of being swallowed like the first press. can you keep swallowing those keys and only skip the override

Sources/Vorssaint/Services/WindowLayout/WindowGestureSupport.swift
```diff
@@ -330,6 +330,17 @@ enum WindowDirectionalGestureSupport {
...
     }
+
+    /// Holding the ring open keeps the trigger key auto-repeating. Those
+    /// repeats (and the trigger key itself) must not force maximize/minimize,
+    /// or the default ⌃⌥Space binding fights the pointer aim (#1566).
+    static func shouldApplyKeyboardManualOverride(keyCode: Int64,
+                                                  triggerKeyCode: Int64?,
+                                                  isAutorepeat: Bool) -> Bool {
+        if isAutorepeat { return false }
+        if let triggerKeyCode, keyCode == triggerKeyCode { return false }
```
vorssaint: the ring only opens once the trigger key is already down and closes when it is released, so a fresh press of that key never gets here and its repeats are already dropped by the line above. can you drop the trigger key check and the value stored for it

Issue #1582, opened 2026-09-11
Uninstaller: drop / "Choose app" receives the file but never proceeds (no scan, no prompt, nothing removed)


69grcv8vfm-sys:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

From me (a human): The information below is written by "my AI". I asked it to summarize the troubleshooting he helped me through, in order to post it here. I am not very technical so I dont know if the technical details are correct. This issue is not a big issue for me, I just wanted to post about it anyways, as this perhaps could help contributing to getting the issue fixed.

Claude description of issue: 
Expected: dropping an app onto the Uninstaller drop target (or picking one via "Choose app...") starts the uninstall flow — scans for leftover files and prompts to move them to Trash.

Actual: nothing happens. No error, no dialog, no progress indicator, and the app is not uninstalled. Both the drag-and-drop target and "Choose app..." are affected equally.

Diagnostics already done:
- Confirmed on the latest version (Settings > About)
- Toggling the Uninstaller feature off/on (Features page) made no difference
- Fully removing Vorssaint (Trash + Empty Trash) and reinstalling made no difference
- Vorssaint never appears in System Settings > Privacy & Security > Automation, so this doesn't look like a permission prompt being denied/missed

Console.app (process "Vorssaint", Info + Debug messages enabled) shows the drop is received correctly at the system level: kDragIPCDrop fires, NSItemProvider loads the 'public.url' representation, a sandbox extension is consumed for the file path — and then the drag session is simply torn down (Dispose drag reference / Deallocate pasteboard) with no further activity. So the app does get the dropped file's URL, but nothing downstream acts on it.

macOS version: 26.6.2 (25G83)
Vorssaint version: 3.3.5
Mac: M3 Macbook Air 

### Steps to Reproduce

1. Open Vorssaint > Uninstaller
2. Drag any .app onto the drop target, or click "Choose app..." and select one
3. Observe: nothing happens — no scan, no confirmation dialog, app is not removed

### Feature Area

Windows and Dock

### Screenshots / Recordings

https://github.com/user-attachments/assets/38320e53-b0d9-45f3-b382-5fcc563dd05e

### Vorssaint Version

3.3.5

### macOS Version

macOS 26 (Tahoe)

### Language

English

iltonandrew:
[...]

PR #1629, opened 2026-09-14
fix(uninstall): offer only apps the uninstaller will take

iltonandrew:
## Summary

The uninstaller's picker and its drop zone decided eligibility one way and `AppUninstaller.select` another. The picker listed everything in the Applications folders that is not under `/System` or `/Library/Apple`, which includes Apple's App Store apps (Pages, Numbers, Keynote, iMovie, GarageBand), while `select` refuses any protected bundle identifier, and does so silently. Picking GarageBand went from the picker straight back to "Drag an app here" with no scanning state, which is the sequence in the #1582 recording. A drop had the same silent path: the closure called `select` and told the system the drop was accepted whatever happened.

The eight checks now live in one place, `UninstallerSupport.selection(for:)`, which returns the accepted selection or nil. `select` uses it and reports whether it took the app; both pickers list only apps it accepts; a refused drop returns false so the drag springs back instead of vanishing.

This covers the reproducible half of #1582. Standard Notes in the report is not a protected identifier, and picking a third-party App Store app worked here, so the reporter's own case is still being narrowed on the issue.

## Verification

MacBook Pro (Apple silicon), macOS 26.6.2 (25G83), own `--dev` build, menu-bar panel driven through accessibility.

- Before: GarageBand, Pages, Numbers, Keynote and iMovie were each listed by the picker; pressing GarageBand logged the bundle-identifier check returning and the panel showed the empty state at once, no scanning frame.
- After: those five are absent from the picker; a third-party App Store app (Magic Disk Benchmark) still goes picker, scanning, results with the same four leftovers as before.
- Tests exercise `selection(for:)` on temporary bundles: a third-party bundle is accepted with its standardized path; an Apple identifier, a symlinked app and a bundle without `Contents/Info.plist` are refused. Source-shape checks pin that both views list `offeredApplications()` and return `select`'s answer from their drop closures.

```sh
./build.sh --test              # TESTS OK
```

Not tested: a real drag of a refused app onto the drop zone (synthetic drags do not start a Finder pasteboard drag here), and macOS 15.

## Anything else

Two things worth a call. Apple's App Store apps were already refused by `select`, since a `com.apple.` identifier could claim shared paths; this only stops offering them. If they should be uninstallable, the identifier rule is the place to change, not the picker. And a refused drop now springs back rather than vanishing, but nothing says why; a short note would need a string in every language, so it is left for a follow-up if wanted.

Refs #1582


Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
```diff
@@ -93,6 +103,34 @@ enum UninstallerSupport {
...
+        let infoURL = selectedURL.appendingPathComponent("Contents/Info.plist")
+        guard let selectedInfoIdentity = fileIdentity(at: infoURL),
+              removalPathIsSafe(infoURL, within: selectedURL) else { return nil }
+        return Selection(bundleID: bundleID, url: selectedURL,
+                         identity: selectedIdentity, infoIdentity: selectedInfoIdentity)
+    }
+
+    /// What the pickers list. An app that selection(for:) would refuse is
+    /// left out rather than offered and then silently turned down.
+    static func offeredApplications() -> [InstalledApps.InstalledApp] {
```
vorssaint, line 132: the command bar still offers this. searching an app there and opening its actions shows an uninstall entry for anything outside the system folders, so pages, garageband or an ios app get the entry, the uninstaller turns them down in silence and the settings window opens on the empty drop zone, which is the same bounce this change removes from the pickers. that entry should only show up for an app this check accepts

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
```diff
@@ -93,6 +103,34 @@ enum UninstallerSupport {
         return rawValue
     }
 
+    /// The one answer for "will the uninstaller take this app": the picker
+    /// offers only what this accepts, and a drop of anything else is refused.
+    static func selection(for appURL: URL) -> Selection? {
+        guard let bundle = Bundle(url: appURL) else { return nil }
```
vorssaint, line 111: this was already the case before the move, but with every check now in one place this is the spot for it. the drop zone also accepts a link dragged out of a browser, and that is a web address, not a file on disk. this line throws for a web address instead of coming back empty, so the app is left in an odd state instead of the drop springing back. anything that is not a file location should be refused before it

andreisuslov:
@vorssaint could you please look into the PR's in the ascending order? there's lots of PR's that need a owner/maintainer to merge them, they are marked as ready, but the PR's are only taken care of if they are the newest ones. i don't think it's a good practice of maintaining the repo.

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
iltonandrew: Done in e75c08c: the command bar row action now offers Uninstall only when `UninstallerSupport.selection(for:)` accepts the app (the `isSystem` gate was a subset of that and is gone), and `openUninstaller` opens Settings only when `select` took the app, so the empty drop zone bounce is gone from that surface too. Source pin added in the test block.

Sources/Vorssaint/Services/Uninstall/UninstallerSupport.swift
iltonandrew: Done in e75c08c: `selection(for:)` refuses anything that is not a file URL before `Bundle(url:)`, so a browser link springs back through the same path as any other refusal. Covered by a test with an `https:` URL.

iltonandrew:
Reporter diagnostics landed on #1582: their Standard Notes is the iOS build installed on Apple silicon (`Standard Notes.app` holds only `Wrapper/` and the `WrappedBundle` link, no `Contents/`), so `selection(for:)` refuses it at `Contents/Info.plist` and, with this PR, the picker no longer offers it. Two calls for you rather than me: should wrapped iOS apps be uninstallable (their data sits in `~/Library/Containers/<id>` like other sandboxed apps, and the real bundle is one level down), and should a refused drop say why instead of springing back silently? Neither is in this PR. The Dock drag they also reported (picker takes Spotify, Dock drop does nothing) is a separate path; measuring it now.

Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -1919,7 +1919,7 @@ final class CommandBarService: ObservableObject {
 
     private func openUninstaller(for url: URL) {
         hide()
-        AppUninstaller.shared.select(appURL: url)
+        guard AppUninstaller.shared.select(appURL: url) else { return }
```
vorssaint, line 1922: while another removal is still running the app refuses a new selection, so picking uninstall here just closes the bar and nothing opens. can the uninstaller page still open in that case, so the running removal stays in view

PR #1632, opened 2026-09-14
feat(window-layout): cycle repeated side actions through thirds

MakhBeth:
## Summary

This brings Rectangle's repeat behaviour to Window Layout, as an opt-in toggle. In Rectangle, pressing Left (or Right) again cycles the window through half, two thirds and one third on the same display; this does exactly that. Today a repeat pushes the window to the next display, and that stays the default: the toggle just swaps one behaviour for the other. Top and bottom keep maximizing on repeat.

The cycle only continues if the window is still where the previous step left it (read back from Accessibility, so app minimum sizes are respected). Move or resize it by hand and the next Left starts again from the half.

Small surface: one preference, one toggle with caption in every language, no new dependencies. Only `applyPlacement` resolves repeats; pointer snaps opt out and are unaffected.

## Verification

Mac16,11, macOS 26.5.1 (25F80), own Developer build, multi-display setup.

```sh
./build.sh                    # no warnings
./build/Vorssaint --selftest  # OK
./build.sh --test             # 38214 checks OK
```

Tests cover the cycle order on both sides, restart after another action, the off default, top-twice-maximizes untouched, and the "still in place" guard (manual resize, tolerance, clamped minimum size, no state).

## Related issues

Refs #907


Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
```diff
@@ -260,6 +260,15 @@ enum WindowLayoutAction: String, CaseIterable, Identifiable {
 
 /// The configurable spacing around snapped windows (issue #1068). Values are
 /// in pixels, offering standard spacing presets (0 to 128 px).
+/// Whether a repeated Left or Right cycles the window through half, two thirds
+/// and one third of the same display (the way Rectangle does) instead of
+/// pushing it onto the display beside it.
+enum WindowLayoutSideRepeat {
```
vorssaint, line 266: this sits between the gaps comment and the gaps code it describes, so that comment now reads as if it were about the repeat setting. can you move the new block below the gaps one

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
```diff
@@ -281,22 +290,54 @@ enum WindowLayoutGeometry {
...
         if action == .topHalf, previousAction == .topHalf {
             return .maximize
         }
+        if sideRepeatCyclesThirds, let next = sideCycleAction(for: action, previousAction: previousAction) {
+            return next
+        }
         return action
     }
 
+    /// The Rectangle-style size cycle for a repeated side action: half, then
```
vorssaint, line 304: this comment and the one describing the new setting name another app. can you describe the behaviour instead, other products stay out of the source

Sources/Vorssaint/Core/FeatureStrings.swift
```diff
@@ -1288,6 +1292,8 @@ struct WindowLayoutFeatureStrings {
         gapsCaption: "Space between snapped windows, and between windows and the screen edge.",
         windowGap: "Window gap",
         screenGap: "Screen gap",
+        sideRepeatCycle: "Repeat Left or Right to cycle sizes",
+        sideRepeatCycleCaption: "Half, then two thirds, then one third on the same display. Off, the repeat moves the window to the next display.",
```
vorssaint, line 1296: with this off a repeated left carries the window to the display on the left, which is not always the one the next display action picks. can the caption say the display on that side instead

vorssaint:
this is the request in #907. can you add the issue reference line to the description in the exact form the contributing guide asks for, so the report follows the merge

iva-zhu:
Smoke-testing this on an older machine (MacBookPro18,3, macOS 15.7.9) — worth reporting because the PR was verified on Mac16,11 / macOS 26.5.1.

I built current main with this PR merged (no other changes to Window Layout sources) and just installed it; first impressions from a short live session:

- The half → two thirds → one third → half ladder works on both sides, and the settled-frame guard behaves as intended: dragging or resizing the window by hand restarts the cycle at the half, and windows clamped by the app's minimum size still cycle.
- Turning the toggle off restores the crossing repeat, and with it on a top/bottom repeat still maximizes — no interaction surprises there.
- One note for the caption review comment above: on a single-display setup the old behaviour of a repeated Left was a no-op failure, so the new caption should probably not promise a move at all when there is no display on that side — "the window stays and the cycle takes over" reads truer than anything that implies a move.

Also confirming the report on #907: this matches the request, ladder order included. I'll follow up here if longer daily use turns anything up.

MakhBeth:
> this is the request in #907. can you add the issue reference line to the description in the exact form the contributing guide asks for, so the report follows the merge

Thanks, I addressed all the comments!

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
```diff
@@ -494,6 +524,29 @@ final class WindowLayoutService: ObservableObject {
...
+                                             targetRect: NSRect,
+                                             action: WindowLayoutAction) {
+        let windowID = windowKey.windowID
+        let timer = Timer(timeInterval: 0.3, repeats: false) { [weak self] _ in
+            guard let self else { return }
+            self.settleTimers[windowID] = nil
+            guard let settled = self.settledFrames[windowKey],
+                  let actual = self.frame(of: window),
+                  actual.isClose(to: settled.requested, tolerance: self.frameTolerance)
+                    || self.accepted(actual: actual, targetRect: targetRect, action: action)
```
vorssaint, line 549: when a half is clamped by the app and the user resizes it before this refresh runs, the lenient acceptance also accepts that manual size and records it as settled. the next left or right then advances to two thirds instead of restarting at half. can you keep manual changes from becoming the settled frame here and cover that refresh path?

MakhBeth: Fixed in 819f2fd1. The refresh now runs the re-read frame through `WindowLayoutGeometry.settledFrameRefreshAccepts` before the lenient anchor check: it only records a frame whose every edge sits at least as close to the requested frame as the immediate read-back did (within the 8pt frame tolerance). A late clamped resize moves toward the request and still counts; a half widened or dragged by hand moves away and is dropped, so the next left or right restarts at the half. Tests cover the late commit, the manual widen and move, and the within-tolerance drift, plus a source check that the refresh path uses the gate.

MakhBeth: Thanks, I think I got it. Now the geometric filter checks every edge of the re-read frame (x, y, width, height) against the frame read right after the placement, with an 8pt tolerance, and only records it when no edge moved farther from the requested half. So a delayed clamp by the app is still recorded, while a half widened or dragged by hand is dropped and the next left or right restarts at the half. With this solution, if the user shrinks the frame manually toward the half, it will still be accepted. That's the limit, but I think that starting the cycle from the new size is still ok and the best UX.

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
```diff
@@ -285,22 +294,76 @@ enum WindowLayoutGeometry {
...
+    /// promoting to maximize instead. With the size cycle on, the repeat is
+    /// spent on the same display and never crosses.
     static func displayCrossing(
         for action: WindowLayoutAction,
-        previousAction: WindowLayoutAction?
+        previousAction: WindowLayoutAction?,
+        sideRepeatCyclesThirds: Bool = false
     ) -> (action: WindowLayoutAction, movingRight: Bool)? {
-        guard action == previousAction else { return nil }
+        guard !sideRepeatCyclesThirds, action == previousAction else { return nil }
```
vorssaint: main now also sends a repeated top or bottom half to the display above or below, and this branch conflicts with that change here. keeping this guard as it is when merging main would turn that vertical crossing off too whenever the cycle is on, so someone with stacked displays loses it just by enabling a setting about left and right. can the guard skip the crossing only for the left and right halves, with the doc above saying the same, and add a test that top and bottom twice still cross with the cycle on?

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
```diff
@@ -285,22 +294,76 @@ enum WindowLayoutGeometry {
...
+    /// from the same side count, so a left after a right third starts over.
+    static func sideCycleAction(for action: WindowLayoutAction,
+                                previousAction: WindowLayoutAction?) -> WindowLayoutAction? {
+        let cycle: [WindowLayoutAction]
+        switch action {
+        case .leftHalf: cycle = [.leftHalf, .leftTwoThirds, .leftThird]
+        case .rightHalf: cycle = [.rightHalf, .rightTwoThirds, .rightThird]
+        default: return nil
+        }
+        guard let previousAction, let index = cycle.firstIndex(of: previousAction) else { return nil }
```
vorssaint: the earlier requests missed this, it was already in the earlier push. with the cycle on, a window placed with the left two thirds shortcut, which ships with a default combination, goes to one third on the next left press instead of the half, and the same happens on the right. the cycle follows the last placement rather than the last key, while the setting says repeating left or right cycles and the cycle tests expect a fresh start after any other action. can the cycle continue only when the previous press asked for the same half, for example by keeping the requested action next to the settled frame, and add a test for two thirds followed by left?

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
```diff
@@ -471,8 +489,20 @@ final class WindowLayoutService: ObservableObject {
         assistiveModeSuspensions[windowID] = EnhancedUserInterfaceSuspension.suspend(forAppOf: window)
 
         let original = self.frame(of: window)
+        settledFrames.removeValue(forKey: windowKey)
         if attempt(frame, targetRect: targetRect, action: action, on: window) {
             assistiveModeSuspensions.removeValue(forKey: windowID)?.resume()
+            let actual = self.frame(of: window) ?? frame
```
vorssaint: this read runs after every placement, as does the one when a late placement settles, and a window left more than 8 points from the request also gets a 0.3 s timer with a second read, even with the cycle off, which is the default. only the cycle uses the settled frame, so can these reads and the refresh run only while the cycle is on?

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
MakhBeth: Done, the guard now skips the crossing only for the left and right halves. The doc comment says the same, and there's a new test for top and bottom twice crossing vertically while the cycle is on

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
MakhBeth: Done: the read-back after a placement, the 0.3 s refresh, and the read when a late placement settles now run only for a left or right half pressed while the cycle is on.

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
MakhBeth: Good catch, thanks. The settled frame now also keeps the key that was pressed for it, recorded only for a left or right half while the cycle is on

PR #1637, opened 2026-09-14
fix(switcher): preserve focus history across activation races

MaximilianMauroner:
## Summary

Rapid app switches can leave the switcher ordered by stale window history when Accessibility has not resolved the newly focused window. Retain unresolved app activations in the focus timeline, then replace them with confirmed window focus. A late reply is filed at its activation's existing timeline position, while a reply whose same-app activation was superseded is rejected. Older enumeration snapshots cannot erase newer focus events.

Observer setup retries are bounded to three attempts. Explicit switches invalidate older focus requests, and ordering reads use a consistent history snapshot.

## Verification

- Rebased onto upstream main `b84919d4`; the regression call now lives in the extracted switcher model suite.
- Head `fe234b24` fixes both follow-up review findings: a known window moves to its activation rank when its reply arrives late, and a late intra-app focus change uses the last confirmed window's rank. Activation identities are invalidated on reactivation, explicit switches, and termination.
- Both new behavior regressions fail on the previous implementation and pass with the repair. Local switcher model suite: 1,334 checks passed.
- Full local test binary: 58,753 of 58,755 checks passed. The two unrelated Homebrew count-format tests expect decimal points but this Mac's locale produces commas.
- On final head `fe234b24`, macOS 15 / Swift 6.0.3 and macOS 26 CI passed, including builds, selftests, unit tests, and macOS 26 DMG packaging. The post-push review found no further issue; all review threads remain resolved.
- Both follow-up review threads are resolved. Codex's review summary still references an older head; the earlier re-review request has no acknowledgement.
- Interactive app-switching behavior has not been retested.

Model: GPT-6. Agent tool: Codex.


Sources/Vorssaint/Services/Switcher/WindowUseOrder.swift
```diff
@@ -4,6 +4,95 @@
...
+
+    mutating func activate(_ pid: pid_t, recording: Bool = true) -> Request? {
+        current = recording ? Request(pid: pid) : nil
+        if recording { promote(.app(pid)) }
+        return current
+    }
+
+    @discardableResult
+    mutating func focus(_ window: CGWindowID, for request: Request) -> Bool {
+        guard current == request else { return false }
```
vorssaint, line 33: if a slow app answers after the user already moved on, the answer is dropped here and the window they used in it loses its place, so the switcher offers that app's older window instead. it can still be filed where the app already sits in the timeline, just not ahead of anything used since

MaximilianMauroner: [gpt-5.6-sol] RESPONDING ON BEHALF OF MAX
-----
Fixed in 8d9d8a43. Activation placeholders now retain their request identity. If an AX read returns after a newer activation, that exact placeholder is replaced in place with the resolved window, so it stays behind newer activity instead of being promoted. A later activation of the same PID removes the older placeholder, so its superseded reply is still rejected. The regression now resolves A to a different window than its fallback and verifies B remains first. The branch is also rebased onto current main; fresh CI is running.

MaximilianMauroner:
[gpt-5.6-sol] RESPONDING ON BEHALF OF MAX
-----
@codex review

Please review latest head 8d9d8a43, including the in-place resolution of late AX focus results and the rebase onto current main.

Sources/Vorssaint/Services/Switcher/WindowUseOrder.swift
MaximilianMauroner: [gpt-5.6-sol] RESPONDING ON BEHALF OF MAX
-----
Fresh validation is complete on 8d9d8a43: both Build & selftest and Swift 6.0.3 compatibility passed, including unit tests. The regression for this late-response ordering case passed in both jobs.

Sources/Vorssaint/Services/Switcher/WindowUseOrder.swift
```diff
@@ -4,6 +4,124 @@
...
+            promote(.window(window))
+            return true
+        }
+        // The AX query began while this activation was current, but another
+        // app won focus before it answered. Resolve only that activation's
+        // placeholder, in place, so the actual window is remembered without
+        // moving it ahead of anything the user did in the meantime.
+        guard let placeholder = recent.firstIndex(of: .app(request)) else { return false }
+        if recent.contains(.window(window)) {
+            recent.remove(at: placeholder)
```
vorssaint, line 53: when the late answer names a window the user had already used earlier in the session, only the placeholder goes away and that window keeps its older slot, so an app the user just came back to ends up behind everything used since. the window should move up to where the app sits in the timeline

Sources/Vorssaint/Services/Switcher/WindowUseOrder.swift
```diff
@@ -4,6 +4,124 @@
...
+                return false
+            }
+            promote(.window(window))
+            return true
+        }
+        // The AX query began while this activation was current, but another
+        // app won focus before it answered. Resolve only that activation's
+        // placeholder, in place, so the actual window is remembered without
+        // moving it ahead of anything the user did in the meantime.
+        guard let placeholder = recent.firstIndex(of: .app(request)) else { return false }
```
vorssaint, line 51: once the app's first answer has landed there is no placeholder left, so a focus change inside that app that answers after the user moved on is still dropped here and the switcher offers the older window again. can the late answer take the slot of the app's last known window in that case

Sources/Vorssaint/Services/Switcher/WindowUseOrder.swift
MaximilianMauroner: [gpt-6] RESPONDING ON BEHALF OF MAX
-----
Fixed in fe234b24. Late focus is inserted at the activation rank, and any older occurrence of that window is removed. Newer activity remains ahead. The regression fails on the old code and passes on the repair; the local switcher suite passed all 1,334 checks. Fresh CI is running after rebasing onto current main.

Sources/Vorssaint/Services/Switcher/WindowUseOrder.swift
MaximilianMauroner: [gpt-6] RESPONDING ON BEHALF OF MAX
-----
Fixed in fe234b24. Confirmed windows retain their activation identity, allowing an in-flight intra-app focus reply to use the last confirmed window's rank after another app activates. The previous window stays in history. Reactivation, explicit switches, and termination invalidate stale identities. The regression fails on the old code and passes on the repair; all 1,334 local switcher checks passed. Fresh CI is running.

vorssaint:
LGTM

PR #1641, opened 2026-09-14
feat(app-updates): add ignored updates list

Vaor22:
## Summary

Adds the ability to ignore individual app updates.

Ignored updates are moved to a separate list and stay hidden from the Available
list until the user restores them.

The ignored state is saved and shared between the menu panel and Settings.

## Verification

MacBook Air (M2), macOS 27.0 (26A428), own local build, English.

```text
./build.sh                     # OK
./build/Vorssaint --selftest   # OK
./build.sh --test              # 38332 checks OK
git diff --check               # OK
```

- `SELFTEST OK`
- `TESTS OK (38332 checks)`
- App Updates: `OK (261 checks)`
- `PREFERENCE CLEANUP TESTS OK`
- `git diff --check` passed

Manually checked:
- ignoring an available update
- restoring an ignored update
- persistence after another update check
- both the menu panel and Settings

## Demo

Available update with the ignore action:

Снимок экрана — 2026-09-14 в 23 43 53

Ignored application with the restore action:

Снимок экрана — 2026-09-14 в 23 43 59


vorssaint:
closing this one. hiding an app from the update list can fit, but this shape would need to be redone
the available and ignored switch sits above the list all the time, in the panel too, even when nothing was ever ignored, and every row gets a second button for something most people will never touch
ignoring hides the app for good, yet the ignored tab keeps showing the version from the day it was ignored, so a newer release never shows up anywhere
restoring an app runs a full check again, package manager and online lookups included, just to bring one row back
the rows without a checkbox lose the blank space that lined them up with the others, so their icons and names shift left
the ignored apps are saved in three separate places, with a fallback that rebuilds names from the ids for entries no released version ever wrote
the ignored list is not part of the settings backup, so a mac restored from one forgets what was ignored
a smaller version could still come in, with ignore tucked into the row's context menu, the ignored list shown only when it has something, restore that does not check everything again, and the choice carried by the settings backup

PR #1657, opened 2026-09-15
fix(notch): fill the menu bar, name the track, and keep captures smooth

ruvelro:
## Summary

Dynamic Island fixes from bugs reported on the 3.4.0-beta.1 build on an external 3440x1440 display running macOS 27, rebased on #1653 and #1655.

**The collapsed island was shorter than the menu bar.** `NotchGeometry` took its height from `NSStatusBar.system.thickness`, a fixed 22 points, while the bar the system draws on that display is 30 points. The island floated short of the bar's bottom edge. The height now comes from the bar as drawn (the gap between a screen's frame and its visible frame), remembered per display through `NotchMenuBarMeasurements` so an auto-hidden bar keeps the last reading from the same display and mode, with the status bar constant as the last resort. Verified on the display above: the island went from 22 to 30 points tall.

**The compact music strip had a hole in the middle.** Screens without a camera now carry a simulated cutout, and nothing physical sits in its middle, so the strip names the track there (title, plus artist when the bar is at least 28 pt tall) between the artwork and the bars; when menus hide the wings, the label keeps clear of the silhouette's shoulders and corners. Artwork is rounded proportionally with a hairline ring, the bars size from the strip height, and the artwork and the bars anchor to the strip's outer edges, so a wider strip spreads them apart instead of leaving them huddled around the cutout.

**The capture selector stuttered while the island was up.** `NotchEqualizerBars` animated capsule views whose heights change thirty times a second; every tick invalidated layout in the hosting window (`NSHostingView.layout` through `NSAnimationContext.runAnimationGroup` in a `sample` of the running beta, about 5% of the main thread while anything played). The bars now draw into a fixed-size `Canvas`, so a tick is a redraw rather than a layout pass. The menu-space monitor already stops through the capture controls' own lifecycle.

## Testing

- `./build.sh --test` and `./build.sh --dev --install`, then exercised on the display above with music playing, in automatic width mode with Accessibility granted (island at 281x30: simulated cutout plus 56 pt wings).
- Geometry tests cover the measured bar height and its fallbacks, and that the collapsed island and the music strip span the whole bar in every screen quadrant.
- Sampled the running 3.4.0-beta.1 with `sample` to confirm the per-tick hosting-view layout from the equalizer.


vorssaint:
kept the height adjustment, track names on screens without a cutout, the animated bars and your latest spacing change. i added a little room around the curves so the text, cover and bars stay visible when space is tight.

the manual width option is out because it could cover other apps' menus. putting the cover and bars into the narrow center also left too little room for the track name. the extra capture change was left out because the app already handled it

Issue #267, opened 2026-07-17
[Bug]: Smooth Scrolling Not Working


danilo-alm:
### Description

I've got two mice. Smooth scrolling works fine with Attack Shark X11, but does not work at all with Logitech Lift Vertical. I thought the Logi+ app might be messing with something (Even though Mos works), but even killing it makes no difference. I tested with bluetooth on both mice, but that's probably unimportant.

PS: Invert Mouse Scrolling does work.

On a side note: compared to Mos, I like the way you implemented the smooth scrolling because it doesn't make apps like DBeaver and the Mac Terminal scroll huge amounts of lines at once. That was also a problem in games (e.g. map zoom). So if you're going to tweak it in the future, keep in mind that's a thing that would be good to keep since it must bother other users... It wouldn't hurt to add an exclusion list anyways, just in case. I see you'll end up making me ditch Mos as well, wow.

### Steps to Reproduce

1. Connect Logitech Lift Vertical;
2. Enable Smooth Scroll;
3. Does not work.

### Vorssaint Version

3.1.13

### macOS Version

macOS 26 (Tahoe)

### Installation Method

Self update

vorssaint:
Thanks!

Available in the 3.1.14 update.

PR #1659, opened 2026-09-15
feat(mouse): add linear scrolling with a fixed number of lines per wheel notch

renan-tiberio:
Every notch of a mouse wheel scrolls the same number of lines, however fast the wheel spins, the way Windows (three lines per notch by default) and LinearMouse's linear mode do. macOS bakes wheel acceleration into the event delta itself, so a fast spin arrives as several lines in one event; the feature caps every wheel event at one notch and multiplies by a lines-per-notch setting (1 to 10, default 3). Trackpads are untouched through the wheel classification the inverter and smooth scrolling already share. No tap is added: with smooth scrolling off the rewrite happens in the inverter's existing tail tap, which either feature now keeps alive; with smooth scrolling on it is folded into the glide the same way the flip is, so one notch becomes lines times step. Sub-notch events from high-resolution wheels keep their fraction and carry it into the next event, and a reversal drops the carry, mirroring the glide. It ships as its own opt-in hub feature with a panel row and a Settings section, modelled on the mouse acceleration control from #1071. This is the wheel half of #403 that stands on its own: it deliberately leaves pointer speed and a separate horizontal axis alone, and unlike #1040 it does not depend on the glide, so the two only meet in the few lines that read the setting inside SmoothScrollService.

Verified on a Mac16,8 (Apple Silicon, macOS 26.6.2 build 25G83, Command Line Tools with Swift 6.4, single built-in display, one Space, local ad-hoc build, app in English and Portuguese): `./build.sh --test` passes with 38873 checks, `./build.sh` finishes without warnings and `./build/Vorssaint --selftest` reports SELFTEST OK. Not yet tested: a few days of daily use, high-resolution wheels that report fractions of a notch, mice whose drivers report the wheel as continuous events (#267), and the other eleven languages on screen. Opened as a draft for those reasons and because #403's direction has not been settled by a maintainer.

Refs #403


Sources/Vorssaint/Services/SmoothScrollService.swift
```diff
@@ -272,6 +272,13 @@ final class SmoothScrollService: ObservableObject {
...
         let defaults = UserDefaults.standard
+        // Linear scrolling is applied here for the same reason the flip is:
+        // this tap swallows the tick before the wheel tap can see it.
+        let linearLinesPerNotch: Int? = AppFeature.linearScroll.isAvailable
+            && defaults.bool(forKey: DefaultsKey.linearScrollEnabled)
+            ? ScrollWheelSupport.sanitizedLinesPerNotch(
+                defaults.integer(forKey: DefaultsKey.linearScrollLines))
+            : nil
         let invertVertical = invertHere
             && defaults.bool(forKey: DefaultsKey.scrollInverterEnabled) ? -1.0 : 1.0
```
vorssaint, line 283: with the scroll inverter uninstalled in features but its switch left on, turning linear scrolling on makes the glide come out flipped, since this only checks that the inverter is running. can you also check here that the inverter is installed

Sources/Vorssaint/Services/ScrollWheelSupport.swift
```diff
@@ -75,3 +89,70 @@ enum ScrollWheelSupport {
...
+
+// MARK: - Linear scrolling
+
+/// Every notch of a mouse wheel worth the same distance, however fast the
+/// wheel spins. macOS keeps its wheel acceleration inside the delta itself,
+/// so a fast spin arrives as several lines in one event; capping an event at
+/// one notch is what takes it out again. Shared by both wheel taps, so the
+/// glide and the raw wheel agree on what a notch is worth.
+extension ScrollWheelSupport {
+    /// Lines one notch scrolls. Three is what Windows and LinearMouse use, so
```
vorssaint, line 101: names of other systems and apps do not go into the code. can you describe the default by what it does instead

Sources/Vorssaint/Services/ScrollInverter.swift
```diff
@@ -177,45 +186,85 @@ final class ScrollInverter: ObservableObject {
...
+            fixedPoint: event.getDoubleValueField(.scrollWheelEventFixedPtDeltaAxis1))
+        let rawHorizontal = ScrollWheelAxisDelta(
+            line: event.getIntegerValueField(.scrollWheelEventDeltaAxis2),
+            point: event.getIntegerValueField(.scrollWheelEventPointDeltaAxis2),
+            fixedPoint: event.getDoubleValueField(.scrollWheelEventFixedPtDeltaAxis2))
+        var vertical = rawVertical
+        var horizontal = rawHorizontal
+        let defaults = UserDefaults.standard
+
+        if AppFeature.linearScroll.isAvailable,
```
vorssaint, line 207: this rewrites the wheel for every app, including games and 3d tools that count the notches themselves, and an app already excepted from the glide still gets its notches multiplied here. can you give linear scrolling its own apps to leave alone list, like the other wheel features have

Sources/Vorssaint/Services/ScrollInverter.swift
```diff
@@ -177,45 +186,85 @@ final class ScrollInverter: ObservableObject {
...
-                event.setIntegerValueField(.scrollWheelEventDeltaAxis1, value: -verticalLine)
-                if traits.isContinuous {
-                    event.setIntegerValueField(.scrollWheelEventPointDeltaAxis1, value: -verticalPoint)
-                    event.setDoubleValueField(.scrollWheelEventFixedPtDeltaAxis1, value: -verticalFixedPoint)
-                }
+        }
+
+        // A discrete event only ever has its line written; the rest follows.
+        let verticalChanged = traits.isContinuous
+            ? vertical != rawVertical : vertical.line != rawVertical.line
```
vorssaint, line 251: when the capped line count matches the whole number already in the event nothing is written back, so a high resolution wheel that reports one and a half lines still scrolls one and a half. can you write the line back whenever linear scrolling is on and the event moves, so the fraction goes too

Tests/MetricsTests.swift
```diff
@@ -3836,6 +3904,11 @@ struct MetricsTests {
         expect(registeredDefaults[DefaultsKey.mouseAccelerationDisabled] as? Bool == false
                 && registeredDefaults[DefaultsKey.panelControlMouseAcceleration] as? Bool == true,
                "mouse acceleration control is opt-in and visible in the panel when installed")
+        expect(registeredDefaults[DefaultsKey.linearScrollEnabled] as? Bool == false
+                && registeredDefaults[DefaultsKey.linearScrollLines] as? Int
+                    == ScrollWheelSupport.defaultLinesPerNotch
+                && registeredDefaults[DefaultsKey.panelControlLinearScroll] as? Bool == true,
+               "linear scrolling is opt-in, starts at the default notch and shows in the panel when installed")
```
vorssaint, line 3911: the other mouse features also pin their keys in the settings backup check further down. can you add the three linear scrolling keys there as well

PathGao:
Thanks for the update in 13c16731. A user of ours hit exactly this: the first notch barely moves, then fast spins jump. Two things likely stand before the next review: the source-text checks in PointerInputFeatureTests (the maintainer now asks for hosted behavior tests via Tests/generate_sources.py, see #1876), and a short on-device note for the items listed as not yet tested. Marking it ready for review should trigger a fresh triage.

Issue #1650, opened 2026-09-15
[Window Layout] Ignore certain apps


delfu:
### Problem or Motivation

The window layout (with or without shortcuts) interfere with certain apps. Some applications that are fullscreen will bug out when they try to drag and drop when Vorssaint is open with window layout enabled. 

### Proposed Solution

Please add a setting inside the Window Layout util to ignore certain apps

### Feature Area

Windows and Dock

### Does another app already do this?

Magnet

PR #1661, opened 2026-09-15
feat(window-layout): add setting to ignore apps

delfu:
## Summary

AI assisted code change, human reviewed and verified

This PR adds a new setting to the Window Layout util to allow certain apps to be ignored
from the mouse window snapping. 

Certain apps running in fullscreen sometimes traps the mouse (eg: video games) and conflicts
with the Window Layout util's mouse snapping behaviour. Similar apps like Magnet have this feature
that allows users to explicitly ignore certain apps. 

## Verification

Manually tested on macOS 26.6. Verified: 
- able to ignore certain apps (in settings)
- When an ignored app is running, mouse snappings are ignored
- Ignored app is behaving as expected and no longer lags because of conflicts with Vorssaint

Automated tests: 

```sh
./build.sh                     # OK
./build/Vorssaint --selftest   # OK
./build.sh --test              # 47617 checks ok
./build.sh --install      # OK
```

Did not verify: 
- Localization on every language

## Anything else

Screenshot: 

image

Refs #1650


Sources/Vorssaint/Core/WindowLayoutIgnoredAppsStrings.swift
```diff
@@ -0,0 +1,47 @@
...
+    static let it = WindowLayoutIgnoredAppsStrings(listTitle: "Sospendi in queste app", addButton: "Aggiungi app…", removeButton: "Rimuovi", caption: "Layout finestre non usa mouse o tastiera finché una di queste app è attiva.")
+    static let ja = WindowLayoutIgnoredAppsStrings(listTitle: "これらのAppで一時停止", addButton: "Appを追加…", removeButton: "削除", caption: "これらのAppが前面にある間、ウインドウレイアウトはマウスやキーボード入力を使いません。")
+    static let ko = WindowLayoutIgnoredAppsStrings(listTitle: "이 앱에서 일시 정지", addButton: "앱 추가…", removeButton: "제거", caption: "이 앱 중 하나에 포커스가 있을 때 윈도우 레이아웃은 마우스나 키보드 입력을 사용하지 않습니다.")
+    static let zhHans = WindowLayoutIgnoredAppsStrings(listTitle: "在这些 App 中暂停", addButton: "添加 App…", removeButton: "移除", caption: "当这些 App 之一处于焦点时，窗口布局不会使用鼠标或键盘输入。")
+    static let zhTW = WindowLayoutIgnoredAppsStrings(listTitle: "在這些 App 中暫停", addButton: "加入 App…", removeButton: "移除", caption: "當這些 App 之一處於焦點時，視窗佈局不會使用滑鼠或鍵盤輸入。")
+    static let zhHK = WindowLayoutIgnoredAppsStrings(listTitle: "在這些 App 中暫停", addButton: "加入 App…", removeButton: "移除", caption: "當這些 App 之一處於焦點時，視窗佈局不會使用滑鼠或鍵盤輸入。")
```
vorssaint: in japanese, korean and traditional chinese the caption calls the feature by a different name than the title of this page. can it use the name the page already shows

Sources/Vorssaint/UI/Settings/WindowLayoutIgnoredAppsList.swift
```diff
@@ -0,0 +1,29 @@
...
+    private var text: WindowLayoutIgnoredAppsStrings {
+        FeatureStrings.windowLayoutIgnoredApps(l10n.language)
+    }
+
+    var body: some View {
+        AppBundleList(title: text.listTitle,
+                      caption: text.caption,
+                      addTitle: text.addButton,
+                      removeLabel: text.removeButton,
+                      bundleIDs: ignored.apps,
```
vorssaint, line 19: a game started by a launcher usually lives outside the applications folder and often has no identifier of its own, so it does not show up in this picker and would never match. the mouse lists already reach running programs and name them by the file being run, so can this list work the same way

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
```diff
@@ -168,6 +198,10 @@ final class WindowLayoutService: ObservableObject {
         guard AXIsProcessTrusted() else {
             return finish(.failure(.missingAccessibility))
         }
+        guard !WindowLayoutIgnoredApps.shared.contains(
+            NSWorkspace.shared.frontmostApplication?.bundleIdentifier) else {
+            return finish(.failure(.noWindow))
+        }
```
vorssaint: this also stops the panel, the command bar and the radial menu from arranging a window while a listed app is in front, and the panel then says no active window was found. the pause is about the shortcuts and the drag hooks, so can these deliberate actions keep working

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
```diff
@@ -85,34 +86,59 @@ final class WindowLayoutService: ObservableObject {
...
+            && inputAllowed
             && UserDefaults.standard.bool(forKey: DefaultsKey.windowEdgeSnapEnabled)
             && !enabledEdgeSnapZones.isEmpty
             && !WindowEdgeSnapSupport.isSystemTilingEnabled
             && trusted
         wantsEdgeSnap ? startEdgeSnapTap() : stopEdgeSnapTap()
     }
 
+    private func syncIgnoredAppsActivationObserver(available: Bool) {
+        let shouldObserve = available && !WindowLayoutIgnoredApps.shared.apps.isEmpty
```
vorssaint: with shortcuts, dragging and edge snapping all switched off this still watches every app switch for nothing. can it only watch while at least one of them is on

Sources/Vorssaint/UI/Settings/WindowLayoutSettings.swift
```diff
@@ -98,6 +98,7 @@ struct WindowLayoutSettings: View {
                         .foregroundStyle(.secondary)
                     Toggle(text.gestureRaiseWindow, isOn: $gestureRaiseWindow)
                 }
+                WindowLayoutIgnoredAppsList()
```
vorssaint, line 104: this list pauses the shortcuts too, but it sits under window dragging and stays there even with dragging off, so it reads as a dragging option. can it get its own place on the page, the way the thumbnails list has on the switcher page

Sources/Vorssaint/UI/Settings/WindowLayoutIgnoredAppsList.swift
delfu: i actually ignored Steam and the games are ignoring the events properly. Do we want a way to pick the launched game independently? (it is correct the launched game does not appear in the picker)

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
delfu: that's a good point, will update.

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
delfu: good point. will update

Sources/Vorssaint/UI/Settings/WindowLayoutSettings.swift
delfu: oh good point

Sources/Vorssaint/Core/WindowLayoutIgnoredAppsStrings.swift
delfu: to be clear, you're referring to changing these strings ? (based on strings [here](https://github.com/vorssaint/vorssaint-utils/blob/79c099f98cf855d7066f8084ee4c1345f713850c/Sources/Vorssaint/Core/FeatureStrings.swift#L1956) ): 

- `視窗佈局` to `視窗排列` 
- `ウインドウレイアウト` to `ウインドウ配置`
- (and whatever the korean string is, i dont speak korean 😭  ) 

?

Sources/Vorssaint/UI/Settings/WindowLayoutIgnoredAppsList.swift
vorssaint: yes, the game itself needs to be on the list and recognized when it is in front. the pause looks at the program in front, and a game runs as its own program with its own identifier or none, so the launcher entry only holds while the launcher itself is in front

Sources/Vorssaint/Core/WindowLayoutIgnoredAppsStrings.swift
vorssaint: yes, exactly that. japanese ウインドウレイアウト to ウインドウ配置, korean 윈도우 레이아웃 to 윈도우 정렬, and 視窗佈局 to 視窗排列 in both traditional chinese lines

Sources/Vorssaint/Core/WindowLayoutIgnoredAppsStrings.swift
```diff
@@ -31,17 +32,17 @@ extension FeatureStrings {
...
-    static let ja = WindowLayoutIgnoredAppsStrings(listTitle: "これらのAppで一時停止", addButton: "Appを追加…", removeButton: "削除", caption: "これらのAppが前面にある間、ウインドウレイアウトはマウスやキーボード入力を使いません。")
-    static let ko = WindowLayoutIgnoredAppsStrings(listTitle: "이 앱에서 일시 정지", addButton: "앱 추가…", removeButton: "제거", caption: "이 앱 중 하나에 포커스가 있을 때 윈도우 레이아웃은 마우스나 키보드 입력을 사용하지 않습니다.")
-    static let zhHans = WindowLayoutIgnoredAppsStrings(listTitle: "在这些 App 中暂停", addButton: "添加 App…", removeButton: "移除", caption: "当这些 App 之一处于焦点时，窗口布局不会使用鼠标或键盘输入。")
-    static let zhTW = WindowLayoutIgnoredAppsStrings(listTitle: "在這些 App 中暫停", addButton: "加入 App…", removeButton: "移除", caption: "當這些 App 之一處於焦點時，視窗佈局不會使用滑鼠或鍵盤輸入。")
-    static let zhHK = WindowLayoutIgnoredAppsStrings(listTitle: "在這些 App 中暫停", addButton: "加入 App…", removeButton: "移除", caption: "當這些 App 之一處於焦點時，視窗佈局不會使用滑鼠或鍵盤輸入。")
+    static let enUS = WindowLayoutIgnoredAppsStrings(sectionTitle: "Ignore apps", listTitle: "Pause in these apps", addButton: "Add an app…", removeButton: "Remove", caption: "Window Layout does not use mouse or keyboard input while one of these apps is focused.")
```
delfu, line 35: it's now using `Ignore Apps` as the title. i saw a different occurrence somewhere else so just to stay consistent

Sources/Vorssaint/UI/Settings/WindowLayoutSettings.swift
```diff
@@ -98,6 +98,9 @@ struct WindowLayoutSettings: View {
                         .foregroundStyle(.secondary)
                     Toggle(text.gestureRaiseWindow, isOn: $gestureRaiseWindow)
                 }
+            }
+
+            Section(FeatureStrings.windowLayoutIgnoredApps(l10n.language).sectionTitle) {
```
delfu, line 103: a separate section now

Sources/Vorssaint/Services/WindowLayout/WindowLayoutService.swift
```diff
@@ -88,42 +88,45 @@ final class WindowLayoutService: ObservableObject {
...
             && !WindowEdgeSnapSupport.isSystemTilingEnabled
             && trusted
+        syncIgnoredAppsActivationObserver(inputsEnabled: shortcutsEnabled || directionalEnabled
+                                          || gestureEnabled || edgeSnapEnabled)
+        let frontmost = NSWorkspace.shared.frontmostApplication
+        let inputAllowed = !WindowLayoutIgnoredApps.shared.contains(
+            bundleID: frontmost?.bundleIdentifier,
+            executablePath: frontmost?.executableURL?.path)
+        let wantsShortcuts = shortcutsEnabled && inputAllowed
+        wantsShortcuts ? registerHotkeys() : unregisterHotkeys()
```
delfu, line 116: the registrations are now pushed lower

Sources/Vorssaint/Core/WindowLayoutIgnoredAppsStrings.swift
```diff
@@ -31,17 +32,17 @@ extension FeatureStrings {
...
+    static let fr = WindowLayoutIgnoredAppsStrings(sectionTitle: "Apps à ignorer", listTitle: "Mettre en pause dans ces apps", addButton: "Ajouter une app…", removeButton: "Retirer", caption: "Disposition des fenêtres n’utilise ni la souris ni le clavier tant que l’une de ces apps est active.")
+    static let it = WindowLayoutIgnoredAppsStrings(sectionTitle: "Ignora app", listTitle: "Sospendi in queste app", addButton: "Aggiungi app…", removeButton: "Rimuovi", caption: "Layout finestre non usa mouse o tastiera finché una di queste app è attiva.")
+    static let ja = WindowLayoutIgnoredAppsStrings(sectionTitle: "無視するApp", listTitle: "これらのAppで一時停止", addButton: "Appを追加…", removeButton: "削除", caption: "これらのAppが前面にある間、ウインドウ配置はマウスやキーボード入力を使いません。")
+    static let ko = WindowLayoutIgnoredAppsStrings(sectionTitle: "무시할 앱", listTitle: "이 앱에서 일시 정지", addButton: "앱 추가…", removeButton: "제거", caption: "이 앱 중 하나에 포커스가 있을 때 윈도우 정렬은 마우스나 키보드 입력을 사용하지 않습니다.")
```
delfu, line 44: the CJK strings should be updated

Sources/Vorssaint/UI/Settings/WindowLayoutIgnoredAppsList.swift
```diff
@@ -0,0 +1,31 @@
...
+                      reachesEveryApp: true,
+                      acceptsExecutables: true,
```
vorssaint, line 21: a program added by its file now lands on this list too, but a settings backup still carries that path, while the mouse lists leave it out because it only names a file on this mac. can this list keep its paths out of the backup and put them back after a restore, the way those lists do

delfu: im sorry, im not familiar with the backup feature yet. Is your suggestion to simply remove `WindowLayoutService.shared.syncWithPreferences()` from the `onAdd` and `onRemove`, like how MouseAppExceptionList does?

delfu: actually the way this works now is: 
1. launch the app to be ignored
2. pick it from the app picker

The app is stored and ignored by it's name instead of file path. Would it be ok? 

video
https://github.com/user-attachments/assets/5b800e0b-6e6b-41e4-a3cd-a464ae611c38

vorssaint: the picker only displays the app's name. when a running app has no bundle identifier, it stores the full executable path, even when selected after launching it. the picker can stay unchanged; please exclude those paths from exported backups and preserve this mac's existing path entries when restoring, as the mouse lists do

delfu: gotcha, it should be backing up and restoring as the Mouse Lists feature does. PTAL

Issue #1589, opened 2026-09-11
Signal converter that accepts I2C writes instead of rejecting them is classified writeOnly, losing the software dimming fallback


sarat03:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

With a monitor connected over HDMI through a USB-C hub, the brightness slider does nothing, and Vorssaint never falls back to the software dimming it already has.

`DDCChannelOutcome` already anticipates this hardware: "Signal converters in the path (USB-C to HDMI adapters, and the low end Macs whose HDMI port is such a converter internally) reject every I2C write outright, which tells a dead channel apart from a monitor that just answers poorly …" The assumption is that such a converter rejects writes, so the channel reads as `.dead` and the display gets software dimming.

My hub does the opposite. It accepts every write and passes nothing to the monitor.

`IOAVServiceWriteI2C` returns `KERN_SUCCESS` even when nothing downstream acts on the write. `BrightnessSupport.channelOutcome` treats that as proof the write landed:

```swift
return writeAccepted ? .writeOnly : .dead
```

So a converter that swallows writes instead of rejecting them is classified `.writeOnly` rather than `.dead`, which is the opposite of the branch the comment intends for it. In `BrightnessService.rebuild` those branches differ in the way that matters: `.dead` leaves the display in `softwareIndices` and it gets gamma dimming, "which works on any connection", while `.writeOnly` calls `softwareIndices.remove(candidate.index)` and pins it to a DDC route that silently does nothing. The verdict then persists in `brightnessDDCWriteOnlyPaths` with no UI to clear it, and re-probing reaches the same answer.

### Evidence

Writes are accepted and the panel never moves, including at level 0.0:

```
ddc probe display 2: writeOnly
wrote display 2 route ddc level 0.000000 ok true
```

Not Vorssaint's packets: sending a Set VCP luminance straight to the same `DCPAVServiceProxy` from a standalone binary, with the bytes Vorssaint builds, also does nothing. Get VCP returns an all-zero reply, with Vorssaint's checksum and with a spec-seeded one alike.

```
wire 51 84 03 10 00 0A A2  ->  KERN_SUCCESS, no change
```

Not service matching: the registry exposes exactly one `DCPAVServiceProxy` with `Location = "External"`.
[...]

PR #1662, opened 2026-09-15
fix(displays): dim in software behind a converter that accepts writes

sarat03:
A signal converter that swallows I2C writes instead of rejecting them is
classified as write-only rather than dead, which pins its display to a DDC
route that silently does nothing. The display loses the gamma dimming such a
connection is supposed to fall back to, so the brightness slider moves and
nothing happens.

Write-only and not-listening cannot be told apart on the bus, since both accept
writes and answer no reads. Rather than guessing, this adds a per-display Dim in
software control on displays whose channel takes writes but never replies.
Turning it on moves the display to the existing software route. Turning it back
off clears the cached write-only verdict so the channel is probed again instead
of reusing the answer that pinned the display there.

The preference is stored per connection path and is not exported with settings,
for the same reason the write-only verdicts are not: it describes one monitor on
one Mac port.

Refs #1589


Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -1811,6 +1830,33 @@ final class BrightnessService: ObservableObject {
...
+        let defaults = UserDefaults.standard
+        let stored = defaults.stringArray(forKey: DefaultsKey.brightnessForcedSoftwarePaths) ?? []
+        let updated = BrightnessSupport.updatedWriteOnlyDDCPaths(
+            stored, path: pathKey, isWriteOnly: preferred)
+        if updated != stored {
+            defaults.set(updated, forKey: DefaultsKey.brightnessForcedSoftwarePaths)
+        }
+        forgetWriteOnlyDDCPath(pathKey)
+        Self.log.log("display \(id) software dimming preferred \(preferred)")
+        refresh(force: true)
```
vorssaint: turning this off while the picture is dimmed leaves it dimmed. the slider goes back to the monitor, but the darkened picture is never restored and the slider starts from that software level, so the screen stays dark or gets dimmed twice. can you restore the picture and start the slider fresh when the choice goes off

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
```diff
@@ -136,7 +136,37 @@ struct BrightnessSection: View {
                     .disabled(service.isDisplayPending(display.id))
                     .accessibilityLabel(display.name)
             }
+            if showsSoftwareDimmingChoice(display) {
```
vorssaint: the same display rows are on the energy page in settings, where this monitor's slider still does nothing. can the choice show up there too

Sources/Vorssaint/Core/BrightnessStrings.swift
```diff
@@ -80,7 +81,8 @@ extension BrightnessFeatureStrings {
         keyboardLightCaption: "Turns the keyboard backlight on or off.",
         keyboardBrightnessShortcuts: "Use keyboard brightness shortcuts",
         keyboardBrightnessDecrease: "Decrease keyboard brightness",
-        keyboardBrightnessIncrease: "Increase keyboard brightness"
+        keyboardBrightnessIncrease: "Increase keyboard brightness",
+        softwareDimming: "Dim in software"
```
vorssaint: the caption on the settings page says the slider dims the picture and never mentions software. can the label follow that wording, something like dim the picture

Sources/Vorssaint/Core/SettingsBackupSupport.swift
```diff
@@ -136,6 +136,7 @@ enum SettingsBackupSupport {
         DefaultsKey.systemShortcutsSuppressed,
         // DDC capability belongs to one physical monitor on one Mac port.
         DefaultsKey.brightnessDDCWriteOnlyPaths,
+        DefaultsKey.brightnessForcedSoftwarePaths,
```
vorssaint, line 142: the write-only list right above has a check that it never travels in a backup. can this key get the same check

Sources/Vorssaint/Services/Display/BrightnessService.swift
sarat03: fixed in `de22e4a9`, covered in `c95d47eb`.

you are right, and it was both halves of what you describe. the curve stayed scaled because nothing put it back, and the remembered level still described the gamma route, so the slider seeded from it and the next write dimmed what was already dim.

the restore now happens before the rebuild, so the channel is probed against a display showing its own picture, and the remembered level is dropped so the slider starts from what the monitor reports.

tested on the hardware rather than in theory, on a monitor behind a hub that swallows ddc, watching the gamma table from outside the app:

```
8241dcc6 (before)  20:59:48.464  choice off -> gamma stays 0.3499, slider dead
9d76a838 (after)   21:01:11.630  choice off -> gamma 0.3508 -> 1.0000, same 100ms sample
```

the ordering is also pinned in tests now: the body is compiled against doubles through `generate_sources.py`, so the preference, the cached verdict, the dropped level, the restore and the rebuild have to happen in that order, and turning the choice on must do none of it. recorded in `mutation_checks.py `too.

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
sarat03: done in `46180b39`, with the visibility rule covered in `9d76a838`.

yes, it should be there. the energy page lists the same displays and the slider is just as dead on them, so leaving the way out in the panel only meant the surface where you notice the problem is not the surface that fixes it.

the panel's two private helpers became one `SoftwareDimmingButton` used by both rows, which also stops the surfaces drifting. the visibility rule went into tests while i was there: it lived in a computed property nothing could reach, so it could have gone back to disappearing the moment the route changes, which would leave no way back to ddc. the converter case, a readable channel, the built-in panel, a display switched off and one moved here by hand are all covered now.

Sources/Vorssaint/Core/BrightnessStrings.swift
sarat03: changed in `1a3903b6`.

agreed, "dim in software" was describing the implementation rather than what the person sees. it is "dim the picture" now.

rather than translating that phrase thirteen times, each language takes the wording its own caption already uses for the same thing, so the label and the caption read as one sentence: "escurecer a imagem", "bild abdunkeln", "assombrir l'image", "画面を暗くする", and so on.

Sources/Vorssaint/Core/SettingsBackupSupport.swift
sarat03: added in `c95d47eb`, with one difference worth mentioning: the check above it does not currently hold anything. neither path key is a registered default, so `exportKeys()` can never contain either one, and !`exportKeys().contains(...)` passes even with the whole machine-state list emptied.

so the new one asserts the key is named in `machineStateKeys` as well as absent from the export, which is the part that can actually regress. it fails when i remove the key from the list, where the old shape does not. happy to change the write-only one to match in this pr if you want them consistent.

sarat03:
pushed `294e9764`. all four are answered inline.

also merged current main in, so this sits on upstream rather than a base from 49 commits back. clean merge, full suite green at 54403 checks, and the display files are byte for byte the ones the hardware check in the first thread used, so that result stands on the merged state.

Issue #1587, opened 2026-09-11
Keep Awake automation: option to require all selected conditions


sarat03:
### Problem or Motivation

Keep Awake's automation conditions (external display, connected to power, selected apps running) combine as OR. KeepAwakeAutomationSupport.action() activates whenever the matching set is non-empty, and deactivates only when that set is empty. Any single condition is enough to start a session and enough to keep one alive.

I have two setups where that is the wrong behaviour.

Docked, I enable External display and Power. What I want is Keep Awake at my desk: monitor connected and on AC. What I get is Keep Awake firing on either alone, so plugging into a monitor while on battery blocks sleep and drains the battery. That is the case I was trying to exclude by enabling the Power condition.

Undocked, I enable Applications and Power, so a long running task keeps the Mac up only while it is plugged in. Here the deactivation side is the real problem. If I unplug while the app is still running, the app condition still matches, the set is not empty, and Keep Awake stays on. The Mac keeps itself awake on battery until I quit the app or intervene by hand. There is no configuration of the current OR behaviour that expresses "stop when power goes away", because removing the Applications condition makes the trigger broader rather than narrower.

### Proposed Solution

A match mode over the automation conditions: Any, which is today's behaviour, or All, where every enabled condition must be satisfied for a session to start and to continue. Default to Any so nothing changes on upgrade.

The logic is small, since matchingConditions is already "enabled and satisfied": All is matches == enabledConditions, for both the activate and the deactivate branch. The cost is the picker and the strings in all 13 locales.

Worth settling before that gets built, though. #1433  is adding this same control for the closed-lid gate (ClamshellGateMode, with .all as selected.isSubset(of: satisfied), over display, power and network) and deliberately leaves action() alone. If it merges and this is then built separately, Keep Awake ends up with two condition enums and two match-mode pickers for one idea.

So the question underneath this is whether one shared condition model is wanted, or whether the two stay separate because they gate different stages: closed-lid narrows a session that is already active, while automation decides whether a session starts at all. One inconsistency to resolve either way is that an empty selection means opposite things today, since clamshellGatePasses returns true on an empty set while action() never activates on one.

### Feature Area

Energy and display

### Does another app already do this?

Amphetamine

### Mockups / References

#1433 's segmented Any / All control is the existing UI vocabulary for this.

PR #1663, opened 2026-09-15
feat(keep-awake): require all selected automation conditions

sarat03:
Keep Awake treats its automation conditions as alternatives, so a session stays
open while any one of them is met. A docked setup that selects both external
monitor and connected power keeps the Mac awake on either one alone, and an
app-plus-power setup keeps running after the AC goes away.

This adds an Any/All match mode over the conditions. Any is the default and
behaves exactly as it does today. All holds a session open only while every
enabled condition is met, so those setups end when the situation they describe
ends.

Before: monitor and power selected together keeps the Mac awake on either one.
After: with All, it stays awake only while both hold, and releases as soon as
one drops.

A second commit keeps All usable when the Applications condition is switched on
before any app has been chosen. A condition that cannot be evaluated is treated
as not enabled, which is what the monitoring setup already does, rather than
leaving All unsatisfiable with nothing on screen to explain it.

Refs #1587


Sources/Vorssaint/Services/KeepAwakeManager.swift
```diff
@@ -427,6 +433,30 @@ final class KeepAwakeManager: ObservableObject {
...
+            runningAppsEnabled: UserDefaults.standard.bool(forKey: DefaultsKey.keepAwakeRunningApps),
+            hasSelectedApps: !runningAppBundleIDs.isEmpty
+        )
+    }
+
+    /// Whether the automation currently asks for a session, in either match
+    /// mode. Every caller that used to read "any condition matches" has to ask
+    /// this instead: under All, a session that stops being wanted still has a
+    /// non-empty matching set (issue #1587).
+    private func automationConditionsHold() -> Bool {
```
vorssaint, line 453: when a timed session runs out it can carry on as an automatic one, and that step still only checks that something matches. with all selected, a session that ends while only the monitor is connected carries on and keeps the mac up on battery. can that step ask for the full match too

Tests/MetricsTests.swift
```diff
@@ -2286,6 +2286,57 @@ struct MetricsTests {
             sessionActive: true,
             automaticSessionActive: false
         ) == .none, "clearing automatic conditions does not end a manual session")
+        // Match mode over the automation conditions (issue #1587).
```
vorssaint, line 2303: every other keep awake setting has a check here that its default is off and that it travels in a settings backup. can the new match mode get the same two checks

Sources/Vorssaint/UI/KeepAwakeAutomationView.swift
```diff
@@ -40,6 +41,19 @@ struct KeepAwakeAutomationEditor: View {
...
             }
+            // One condition cannot be combined with anything, so the mode
+            // would be a control with no effect (issue #1587).
+            if selectedConditionCount > 1 {
+                Picker(selection: $requireAll) {
+                    Text(strings.matchAny).tag(false)
+                    Text(strings.matchAll).tag(true)
+                } label: { EmptyView() }
+                .pickerStyle(.segmented)
+                .labelsHidden()
```
vorssaint, line 52: in the small card inside the panel the tiles and switches shrink, but this control keeps the regular size and the french and german labels are wide for that column. can it follow the compact sizing there, and does it still fit in those languages

Sources/Vorssaint/Services/KeepAwakeManager.swift
sarat03: good catch, it can, and your case reproduces exactly: all selected with display plus power, the timer runs out on battery with only the monitor attached, and the session carries on keeping the mac up.

`continueAutomaticallyAfterTimerIfNeeded` is the one caller i did not move over. it still guards on `!matches.isEmpty`, which is the old any-condition reading, so under all it hands a timed session to the automation that the automation itself would never have started. it will ask for the full match like everything else:

```
let matches = currentMatchingAutomationConditions()
guard KeepAwakeAutomationSupport.conditionsSatisfied(
        matching: `matches,`
        enabled: currentEnabledAutomationConditions(),
        requireAll: automationRequiresAllConditions()) else { return false }
```

Tests/MetricsTests.swift
sarat03: yes. both behaviours are already right, the key is registered false in `Defaults.swift `and it is not in the backup exclusion list, but neither has a check, so nothing would catch a regression. adding the two next to the other keep awake ones:

```swift
expect(registeredDefaults[DefaultsKey.keepAwakeAutomationRequireAll] as? Bool == false,
       "matching every automation condition is opt-in")
expect(SettingsBackupSupport.exportKeys().contains(DefaultsKey.keepAwakeAutomationRequireAll),
       "the automation match mode follows settings backups")
```

Sources/Vorssaint/UI/KeepAwakeAutomationView.swift
sarat03: it can, and it should. the picker is the only control in that view that ignores `compact` while the tiles, spacing and fonts all follow it, so it sits a full row taller than everything around it in the panel card. `.controlSize(compact ? .small : .regular) `with the label at the tile font size lines it up.

it does not fit in those languages, though, and controlSize does not save it. "n'importe quelle condition" and "beliebige bedingung" both truncate in that column, and a segmented control truncated to "n'importe quelle..." is worse than no label.

so shortening every segment to one word across all thirteen languages: any / all, l'une / toutes, eine / alle, and so on. the tiles above already name the conditions and the settings caption spells the mode out in full, so the long wording is not carrying anything here, and short labels fit at both sizes in every language.

sarat03:
pushed b163ae94 with all three.

**timer handoff**. `continueAutomaticallyAfterTimerIfNeeded `was still guarding on `!matches.isEmpty`, the one caller i had not moved over, so under all a timed session was handed to the automation on a match the automation itself would have refused. it asks `conditionsSatisfied` now, with the enabled set and the mode.

that decision lives in the manager, where nothing could reach it, so the guard could have gone back without a single check noticing. its body now goes through `generate_sources.py` like the other extracted methods, compiled verbatim against doubles: both modes, and the availability, suppression and battery guards that came before them. the regression is recorded in `mutation_checks.py` as well, and reverting the guard fails exactly the intended assertion.

**the two checks**. both behaviours were already right, neither had a check. added next to the other keep awake ones, and each one fails when i flip the registered default or add the key to the backup exclusion list.

**match mode control**. it now follows `compact` like every tile, font and inset around it, and each segment is one word in all thirteen languages. the tiles name the conditions and the settings caption spells the mode out, so the long wording was not carrying anything in that column.

one correction to what i said earlier: i claimed french and german both truncate. measuring the real segmented control against the 247pt the card leaves (308 less the card gutters, the keep awake indent and the disclosure indent), french needs 276pt and does overflow, spanish 248pt, german 228pt, which fits with nothing to spare. so german was crowded rather than cut. the short labels stand either way, and there is now a check that measures all thirteen against that width, since the control is the only thing that knows its own insets. the compact sizing itself is pinned as source shape, the editor being unrenderable in the harness.

unrelated, so flagging rather than touching it here: the `switcher` suite fails 21 of 120 at `SwitcherScrollTests.swift:117` on an untouched checkout of this branch with a clean build, and it passed twice earlier on the same commit, so it looks like it depends on gui session state. it also stops `mutation_checks.py` at its baseline, which is why the two new mutations are verified by hand rather than by the tool. happy to open a separate issue if that is not already known.

Issue #1568, opened 2026-09-10
Mic Mute leaves BuiltInMicrophoneDevice muted at the device level — every app gets silence, Safari masks it


noelvarga25:
### Description

Mic Mute writes `kAudioDevicePropertyMute = 1` directly onto the built-in microphone
**device** and does not clear it. Unlike the Volume Mixer's per-process muter (#341),
this mutes the input device itself, so **every** application on the system receives
digital silence — while Vorssaint's own state believes nothing is muted, which makes it
unrecoverable from inside the app.

On my machine the microphone was left muted for **22.5 hours** across normal use.

### Environment

- macOS 26.6.2 (25G83), Apple Silicon
- Vorssaint 3.3.5 (build 86)
- Device: MacBook Pro Microphone (`BuiltInMicrophoneDevice`, id 79), 48 kHz, default input
- Affected: Wispr Flow 1.6.793, Claude Desktop — both Electron/Chromium
- Not affected: Safari

### Actual result

`kAudioDevicePropertyMute` on the input scope is left at `1` while gain stays at `1.000`:

```
── MacBook Pro Microphone  [BuiltInMicrophoneDevice]  id=79   <= DEFAULT INPUT
   mute[elem 0] = 1   *** MUTED ***
   volume[elem 0] = 1.000
```

A muted input device does not raise an error — it streams zeros. Capturing 4 s from the
default input via `AVAudioEngine`:

```
before:  frames=192000  peak=0.000000  rms=0.000000   <- digital silence
after:   frames=192000  peak=0.040774  rms=0.003557   <- normal
```

Clearing the mute restores every affected app immediately. No relaunch needed.

### Evidence: Vorssaint performs the write

The `coreaudiod` mute write is interleaved with Vorssaint's own I/O calls, on the same
thread, microseconds apart:

```
19:42:49.493  coreaudiod      PID 2167 Setting Process Input Mute To: off
19:42:52.503  Vorssaint[821]  HALC_IOContext_PauseIO(161)
19:42:52.506  Vorssaint[821]  HALC_IOContext_PauseIO(160)
19:42:52.554  Vorssaint[821]  HALC_IOContext_ResumeIO(161)
19:42:52.558  coreaudiod      AudioObjectSetPropertyData: device id 79,
                              ['mute','inpt',0], BuiltInMicrophoneDevice, mute: 1.000000
19:42:52.558  Vorssaint[821]  HALC_IOContext_ResumeIO(161)
```

[...]

Issue #1592, opened 2026-09-11
Mic Mute does not silence Siri: the device mute is cleared for voice processing sessions, and a zeroed input volume is ignored


sarat03:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

`MicMuteService` documents its promise as "one click or shortcut cuts every microphone the
Mac has, in any app", and the HUD says "Microphone muted". Siri is not covered by either
mechanism the service has.

With Mic Mute on and the menu bar badge showing, my microphone is correctly silent in
Safari and in System Settings > Sound > Input, where the meter stays flat. Siri hears me
and responds normally.

Mic Mute silences a device two ways: its own mute switch where it has one, otherwise its
input volume dropped to zero. Neither survives contact with Siri.

**The device mute is cleared for the duration of a voice processing session.** Sampling
`kAudioDevicePropertyMute` on the input scope of `BuiltInMicrophoneDevice` every 200ms
while invoking Siri, with the mute latched at 1 beforehand:

    19:38:07.987  mute=1      <- latched
    19:38:18.796  mute=0      <- cleared, session opening
    19:38:27.978  mute=1      <- restored, session closing

This is macOS rather than Vorssaint. The same clear and restore happens with Vorssaint
quit and not running at all, twice in one window:

    19:42:06.069  mute=0
    19:42:13.613  mute=1
    19:42:16.061  mute=0
    19:42:19.944  mute=1

So a device level mute is advisory for any `VoiceProcessingIO` client, which is the same
behaviour #1568 recorded for WebKit. That issue reads the restore as Vorssaint re-asserting
the mute; these runs had no Vorssaint process, so at least that edge belongs to the
aggregate restoring the value it found.

**A zeroed input volume is not touched, and does not stop Siri either.** Same sampler, 90
seconds, app still quit, `mute = 0` and input volume written to 0.000:

    19:45:33.255  mute=0  volume=0.000
    19:47:03.359  FINAL mute=0  volume=0.000

Nothing cleared or raised the volume, and Siri still heard me and responded. So the
fallback path is not a safety net for this case; it is simply ignored.

### What I did not verify

- I did not capture audio, so "Siri heard me" is my own observation of Siri responding,
  not a measured signal level.
- I could not attribute the mute writes. The `coreaudiod` lines #1568 quotes are debug
[...]

PR #1664, opened 2026-09-15
fix(mic-mute): release a microphone left muted with nothing claiming it

sarat03:
A mute released while a voice processing session is open is lost. The session
clears the device's mute switch for its own duration, so the unmute finds the
switch already at zero, treats the device as audible, and drops both the claim
and the recorded state. When the session ends it restores the value it saved,
putting the mute back with nothing left that knows it belongs to this app. Every
raw HAL client then gets silence, and the panel toggle has nothing to act on.

An orphan like that is invisible to the normal unmute, whose record is empty by
then. This keeps a bounded record of every device a mute has reached and
releases anything still silenced from it once per launch, while the app believes
nothing is muted. That record is what tells this app's own mute apart from one
the user applied themselves, which an empty claim list cannot.

One trade-off worth stating: it can open a microphone the user silenced in
System Settings, if this app had muted that same device earlier. Leaving someone
muted with no way back is the worse failure, and it is the same trade
restoreTargets already makes for a missing record.

Refs #1568


Sources/Vorssaint/Services/QuickTools/MicMuteSupport.swift
```diff
@@ -48,4 +48,35 @@ enum MicMuteSupport {
...
+    /// This can also open a microphone the user silenced in System Settings, if
+    /// this app had muted that same device at some earlier point. That is the
+    /// trade `restoreTargets` already makes for a missing record: leaving
+    /// someone muted with no way back is the worse failure.
+    static func orphanedMuteTargets(touched: [String],
+                                    claimed: [String]?,
+                                    silenced: [String]) -> [String] {
+        let owned = Set(claimed ?? [])
+        let seen = Set(touched)
+        return silenced.filter { seen.contains($0) && !owned.contains($0) }
```
vorssaint: muting and unmuting here once must not give the app permanent permission to reopen that microphone. if the user silences it elsewhere later, launching the app currently turns it back on without asking. preserve that later choice instead of treating past use as permission to unmute

Sources/Vorssaint/Services/QuickTools/MicMuteService.swift
```diff
@@ -78,6 +84,33 @@ final class MicMuteService: ObservableObject {
...
+    /// whose record was already cleared. Only the mute switch is reconciled, as
+    /// that is the path a voice processing session interferes with; a level left
+    /// at zero keeps its saved value and is restored the usual way.
+    private func releaseOrphanedMutesOnce() {
+        guard !didReleaseOrphanedMutes else { return }
+        didReleaseOrphanedMutes = true
+        guard !isMuted else { return }
+        let defaults = UserDefaults.standard
+        let touched = defaults.stringArray(forKey: DefaultsKey.micMuteTouchedDevices) ?? []
+        guard !touched.isEmpty else { return }
```
vorssaint: people already affected on the released version have no history for this recovery to use. updating and toggling the microphone off and on still leaves them silent. provide a safe recovery for these existing cases without first requiring a successful mute with the new version

Sources/Vorssaint/Services/QuickTools/MicMuteService.swift
```diff
@@ -78,6 +84,33 @@ final class MicMuteService: ObservableObject {
...
+    }
+
+    /// Runs once per launch, and only while the app believes nothing is muted:
+    /// a mute left behind by a previous run is invisible to the normal unmute,
+    /// whose record was already cleared. Only the mute switch is reconciled, as
+    /// that is the path a voice processing session interferes with; a level left
+    /// at zero keeps its saved value and is restored the usual way.
+    private func releaseOrphanedMutesOnce() {
+        guard !didReleaseOrphanedMutes else { return }
+        didReleaseOrphanedMutes = true
```
vorssaint: in the sequence described in this pr, a call can leave the microphone silent again after this one recovery has already been used up. reconnecting a microphone later or an unsuccessful first attempt also leaves no way to retry until another restart. keep a safe recovery available during use

Sources/Vorssaint/Services/QuickTools/MicMuteSupport.swift
sarat03: you were right, and it was worse in practice than on paper. i staged a microphone muted as if from System Settings, with a record of this app having muted that device once before, and launched the branch as it stood:
```

22:49:56.831  built-in = MUTED   (staged, nothing claiming it)
22:50:00.379  built-in = open    (launch, four seconds later, no click)
```

it reopened a microphone the person had silenced elsewhere, exactly as you said it would, purely because of a record that never forgets.

the record and the sweep are both gone in 47c4f0ac. a microphone that is silent while the app believes it muted nothing is now shown as such, and opened only when the person asks, every time. that removes the standing permission entirely: there is no state that can age into consent, because consent is asked for at the moment it is used.

`micMuteTouchedDevices` is deleted along with its backup exclusion, and `orphanedMuteTargets` and `updatedTouchedDevices` go with it.

Sources/Vorssaint/Services/QuickTools/MicMuteService.swift
sarat03: fixed in `47c4f0ac`, and the root cause behind it in `06882094`.

i reproduced your case first on the branch as reviewed: microphone muted, no record anywhere, then mute and unmute from the panel. it ends where you said, still silent, and the app's own records agree with each other and with nothing:

step | device mute | micMuteActive | micMuteMutedDevices
-- | -- | -- | --
staged | 1 | 0 | absent
mute clicked | 1 | 1 | ()
unmute clicked | 1 | 0 | ()

that is the same table i measured in [#1568](https://github.com/vorssaint/vorssaint-utils/issues/1568), which is where the second commit comes from: `mute()` declined to claim a device it found already silent, correctly, but still reported the mute as applied, so the app recorded an active mute with an empty claim list, and because an empty list and a missing one mean opposite things to `restoreTargets`, the unmute had nothing to act on and reported success too. every later on and off cycle re-entered the same path, which is why it could never recover on its own.

a sweep now reports a mute only for devices this app actually holds, so an active mute always carries at least one device it can give back, and the state you quoted is unreachable. when a mute reaches nothing the app says the microphone is already silent instead.

on the recovery itself, it reads the hardware rather than a record, so it needs no history at all. on the same machine, with nothing stored:
```
22:51:53.602  built-in = MUTED   (fixed build running, left it alone for 52 seconds)
22:52:45.442  built-in = open    (one click on the offer)
```

Sources/Vorssaint/Services/QuickTools/MicMuteService.swift
sarat03: agreed, and the once-per-launch gate is gone with the sweep in `47c4f0ac`. the state is re-read after every sweep, whenever the panel appears, and at launch, so a microphone silenced again mid-call, one that arrives already silent, or an attempt whose write was refused all bring the offer back without a restart.

tested by stranding it a second time against the same running instance, no relaunch:

```
22:53:06.892  built-in = MUTED   (staged again, same pid)
22:53:24.089  built-in = open    (offer was back, clicked again)
```

a refused write is covered too: the release attempts every silent device and then re-reads, so a device that would not take the write keeps the offer standing rather than clearing it optimistically.

sarat03:
one limit worth stating plainly, since this PR is about a microphone that stays silent: it fixes the microphone that will not come back, not the microphone that still hears you.

while testing, i measured the second case on hardware and it is worse than #1592 records. with both the built-in microphone and a pair of AirPods latched muted at the device level, Siri heard me and replied, and neither device's mute was cleared during the session. so a device mute is not merely cleared for a voice processing session, it does not apply to that path at all. no process object reported running input while it was listening either, so the capture is not an ordinary HAL client that any app could reach, and every mute property in CoreAudio is scoped to the calling process.

i do not think that is fixable from user space, and it is not what this PR claims to do. tracking it in [#1592](https://github.com/vorssaint/vorssaint-utils/issues/1592) rather than growing this one.

Sources/Vorssaint/Services/QuickTools/MicMuteService.swift
```diff
@@ -163,16 +207,30 @@ final class MicMuteService: ObservableObject {
...
+        // what a later unmute needs to put every level back. For a mute that
+        // means every microphone was already quiet and this app is holding
+        // none of them, which it has to say rather than record a mute it
+        // cannot undo (issue #1568). The offer is the way out of that state,
+        // so it is re-read here too.
+        guard outcome.applied else {
+            refreshStrandedMute()
+            if announce, muted {
+                QuickToolHUD.show(icon: "mic.slash.fill",
+                                  message: L10n.shared.s.micAlreadySilentHUD)
```
vorssaint, line 221: this also shows up when a microphone would not take the mute, since a refused write ends the same way as one that was already quiet. only say it is already silent when every microphone really is

Sources/Vorssaint/Services/QuickTools/MicMuteService.swift
```diff
@@ -78,6 +86,42 @@ final class MicMuteService: ObservableObject {
...
+        }
+    }
+
+    /// Opens every input device whose mute switch is on. Only ever from a
+    /// direct request: with no record to go by this can also open a microphone
+    /// silenced in System Settings, which is the person's call to make each
+    /// time and never a standing permission the app keeps.
+    func releaseStrandedMute() {
+        halQueue.async { [weak self] in
+            for device in Self.inputDevices() where Self.muteSwitchValue(of: device.id) == 1 {
```
vorssaint, line 119: clicking the offer while a mute is still being applied opens the microphone that mute just closed, and the app keeps showing it as muted. make sure a release and a mute in flight cannot cross like that

Sources/Vorssaint/UI/MenuPanel/QuickTogglesSection.swift
```diff
@@ -192,16 +192,25 @@ struct QuickTogglesList: View {
...
+            // from one made in System Settings, so the person decides, and can
+            // decide again whenever it happens (issue #1568).
             UtilityActionButton(title: micMute.isMuted ? l10n.s.micUnmuteName : l10n.s.micMuteName,
                                 caption: l10n.s.micMuteCaption,
                                 systemImage: micMute.isMuted ? "mic.slash.fill" : "mic",
                                 isEditing: editing,
                                 showsDragHandle: true,
                                 visibility: $showMicMute,
                                 shortcutHint: shortcutHint(.micMute),
+                                accessoryTitle: micMute.hasStrandedMute ? l10n.s.micUnmuteName : nil,
```
vorssaint, line 206: while the offer is up this row still reads mute microphone with an unmute microphone button right under it, so it is not clear whether the microphone is muted. let the row say the microphone is already silent while the offer is showing

Sources/Vorssaint/Core/SettingsBackupSupport.swift
```diff
@@ -136,6 +136,7 @@ enum SettingsBackupSupport {
         DefaultsKey.systemShortcutsSuppressed,
         // DDC capability belongs to one physical monitor on one Mac port.
         DefaultsKey.brightnessDDCWriteOnlyPaths,
+        // Audio device uids name the hardware on this Mac, not a preference.
```
vorssaint, line 139: this comment is left over from the key that was removed, and nothing follows it now. drop it so this file stays untouched

Tests/MetricsTests.swift
```diff
@@ -19979,6 +19982,9 @@ struct MetricsTests {
                 && MicMuteSupport.restoreTargets(recorded: ["mic-a"], present: []).isEmpty,
                "unmuting touches the microphones this app muted, every one with no record, and none when the record is empty")
 
+        // A microphone left silent with nothing claiming it is offered a way
```
vorssaint, line 19985: this note now sits right above the radial menu check, which it has nothing to do with. drop it or move it up to the microphone checks

sarat03:
all five in `32309323`.

**the already-silent message**. you were right that a refused write and a quiet Mac ended in the same place. both paths reached the end of the sweep with nothing applied, and the sweep had no way to say which one it was. the outcome now carries whether a microphone was still open and would not take the write, and only a sweep where every device was genuinely quiet says so. a refused one says the mute did not go through.

**the release crossing a mute**. the offer dispatched straight onto the audio queue, so a click landing while a mute sweep was still queued ran behind it and opened exactly what that mute had just closed, with `isMuted` left true on top of it. the offer now counts outstanding sweeps and stands aside until the one in flight has published, then works normally. it cannot undo a mute that has not finished announcing itself.

**the row reading two ways at once**. it said mute microphone with an unmute button directly under it. while the offer is up it now says the microphone is already silent, and the icon goes to the crossed-out mic to match. reused the existing string rather than adding a fourteenth translation for the same sentence.

**the leftover comment in** `SettingsBackupSupport.swift `is gone, so that file is untouched by this branch again.

**the note in** `MetricsTests.swift` is gone too. the rules it pointed at are in `MicMuteStrandedTests` against the service itself, which is where it should have said it from the start.

Sources/Vorssaint/UI/MenuPanel/QuickTogglesSection.swift
```diff
@@ -192,16 +192,30 @@ struct QuickTogglesList: View {
...
         case .micMute:
-            UtilityActionButton(title: micMute.isMuted ? l10n.s.micUnmuteName : l10n.s.micMuteName,
+            // A microphone left silent with nothing claiming it gets an offer
+            // rather than a silent sweep: the app cannot tell its own lost mute
+            // from one made in System Settings, so the person decides, and can
+            // decide again whenever it happens (issue #1568).
+            // While the offer is up the microphone is silent, so the row says
+            // so: a mute button above an unmute button read as a contradiction.
+            UtilityActionButton(title: micMute.hasStrandedMute
+                                    ? l10n.s.micAlreadySilentHUD
```
vorssaint, line 202: the previous wording request missed mixed devices. this flag is true when any microphone is silent, so an unused muted headset makes the row say the microphone is silent even while the built-in microphone is open. can the row distinguish a recoverable muted device from all microphones being silent, and keep the crossed-out icon accurate?

Issue #1666, opened 2026-09-15
Repeated top and bottom Window Layout shortcuts do not move windows to vertically adjacent displays


ethanbeau:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Window Layout supports moving a window to an adjacent display when the same left or right placement shortcut is pressed repeatedly. However, the equivalent behavior is not implemented for displays arranged vertically.

For example, if a second display is positioned above the current display:

- Pressing the Top Half shortcut once moves the window to the top half of the current display.
- Pressing Top Half again does not move the window to the display above.

The same applies to Bottom Half when another display is positioned below.

This makes repeated directional shortcuts behave differently depending on the physical display arrangement. Left and right can continue across displays, while top and bottom stop at the edge of the current display.

Expected behavior: repeated Top Half and Bottom Half shortcuts should move the window to an adjacent display in that direction when one exists, consistent with the existing left/right behavior.

### Steps to Reproduce

1. Connect at least two displays.
2. Arrange one display above or below the other in macOS Display settings.
3. Enable Window Layout and assign a shortcut to Top Half or Bottom Half.
4. Focus a normal application window.
5. Press Top Half when a display exists above, or Bottom Half when a display exists below.
6. Confirm that the window moves to that half of the current display.
7. Press the same shortcut again.
8. Observe that the window remains on the current display instead of moving to the vertically adjacent display.

### Feature Area

Windows and Dock

### Vorssaint Version

3.4.0-beta.1

### macOS Version

macOS 27 (Golden Gate)

### Language

English

ethanbeau:
Implemented in #1667. It extends the existing repeated display-crossing path to Top Half and Bottom Half, preserves the no-neighbour fallback behavior, and includes geometry regression coverage.

PR #1667, opened 2026-09-15
feat(window-layout): cross stacked displays

ethanbeau:
## Summary

Repeated Top Half and Bottom Half now use the existing display-crossing path to move to the opposite half of the nearest display above or below.

This extends the existing Window Layout shortcut behavior using NSScreen frame geometry only: no new settings, shortcuts, dependencies, or accessibility mutation path. Repeating Top Half with no display above still maximizes; repeating Bottom Half with no display below keeps its existing local behavior.

## Verification

- `./build.sh --test-suite=core` — 22,318 checks passed.
- `./build.sh --dev` — full isolated Developer app build succeeded.
- `Vorssaint (Developer).app --selftest` — `SELFTEST OK`.
- Manually tested on a monitor positioned both above and below the primary screen.
- It's not the easies to record something like this but here is my attempt, you can see it leave the monitor to the top in the second half of the recording:
https://github.com/user-attachments/assets/fbdb7ca5-d74d-4e20-97da-1bbf2c20a621

## Anything else

Refs #1666


ethanbeau:
https://github.com/user-attachments/assets/fbdb7ca5-d74d-4e20-97da-1bbf2c20a621

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
```diff
@@ -288,43 +292,69 @@ enum WindowLayoutGeometry {
...
         guard frames.indices.contains(currentIndex) else { return nil }
         let current = frames[currentIndex]
         return frames.indices
-            .filter { movingRight ? frames[$0].minX > current.minX : frames[$0].minX < current.minX }
+            .filter {
+                switch direction {
+                case .left: frames[$0].minX < current.minX
+                case .right: frames[$0].minX > current.minX
+                case .up: frames[$0].minY > current.minY
+                case .down: frames[$0].minY < current.minY
```
vorssaint: a display beside this one that sits a little higher in the arrangement counts as above here, so pressing top half twice on the lower of the two throws the window onto it instead of maximizing. can you only count a display that starts past this screen's top edge, and one that ends below its bottom edge for the way down

ethanbeau: Fixed in 663fd76: vertical crossing now skips displays that overlap the source vertically, so a repeated top/bottom action maximizes when no display lies fully beyond that edge. Added regression coverage for both directions.

Sources/Vorssaint/Services/WindowLayout/WindowLayoutSupport.swift
```diff
@@ -292,43 +296,70 @@ enum WindowLayoutGeometry {
...
+                let rhsCenterDistance: CGFloat
+                switch direction {
+                case .left, .right:
+                    lhsPosition = lhsFrame.minX
+                    rhsPosition = rhsFrame.minX
+                    lhsCenterDistance = abs(lhsFrame.midY - current.midY)
+                    rhsCenterDistance = abs(rhsFrame.midY - current.midY)
+                case .up, .down:
+                    lhsPosition = lhsFrame.minY
+                    rhsPosition = rhsFrame.minY
```
vorssaint, line 350: with two displays fully below this one, the ranking uses their bottom edges, so bottom half twice can land on a shorter display off to the side instead of the one right underneath. can you rank the way down by each display's top edge, so the nearest one wins and the closest center only breaks ties

ethanbeau: Fixed in 04a1905, with expanded regression coverage in 8b60fe3: downward ranking now uses each candidate’s facing top edge, so the nearest display below wins; `./build.sh --test-suite=core` passes.

vorssaint:
LGTM

PR #1668, opened 2026-09-16
feat(mouse): add per-device pointer controls

NachikethReddyY:
Vorssaint could not independently tune connected pointer devices without a separate utility, and the required Accessibility permission was easy to miss. This adds a native Mouse & Trackpad flow inspired by LinearMouse while keeping restoration and device state inside Vorssaint.

What changed:

- discover connected mice and trackpads, including the built-in Apple trackpad
- save independent pointer speed and acceleration profiles by stable device identity
- restore original HID values on disable, reset, quit, sleep, and session changes
- show each pointer device in its own collapsible native settings card
- keep mouse-only button controls separate from trackpad-only gestures
- put Accessibility status and recovery at the top of Mouse & Trackpad
- expose the existing animated smooth-scrolling speed and response controls
- include the relevant LinearMouse MIT and Apple IOKit SPI notices

Mouse navigation and shortcut actions remain shared across connected mice because the current CoreGraphics event taps do not identify the originating mouse. The UI states that scope instead of presenting those controls as per-device. Pointer speed changes macOS software scaling, not hardware sensor DPI.

I'd love for a maintainer to touch up the UI where necessary, only within Mouse & Trackpad.

### Permission, scrolling, and active smooth scrolling

Accessibility permission, scrolling controls, and active smooth scrolling

### Separate mouse and trackpad pointer cards

Expanded Dell mouse settings and a separate collapsed Apple trackpad card

### Mouse-only buttons and trackpad-only gestures

Mouse button controls separated from trackpad gesture controls

## Verification

- `./build.sh --test` — 50,646 checks passed after rebasing onto current `main`
- preference cleanup passed
- installed and launched `/Applications/Vorssaint (Developer).app` from revision `188a445`
- deep, strict code-signature verification passed
- native UI inspection confirmed separate Dell USB mouse and Apple internal trackpad cards
- Accessibility visibly reported Granted and smooth scrolling visibly reported its event path active

Change authored by gpt-5.6-sol through T3 Code.


Sources/Vorssaint/Services/MouseAcceleration/MouseAccelerationSupport.swift
```diff
@@ -107,22 +235,85 @@ enum MouseAccelerationSupport {
...
+        }
+        if key == pointerResolutionKey {
+            guard !preferences.disablesAcceleration else { return original }
+            return fixedPoint(pointerResolution(forSpeed: preferences.speed))
         }
-        return MouseAccelerationStoredValue(rawValue: -1, isBoolean: false)
+        if preferences.disablesAcceleration && !supportsLinearScaling {
+            return fixedPoint(-1)
+        }
+        return fixedPoint(sanitizedAcceleration(preferences.acceleration))
```
vorssaint, line 312: a mouse that only had acceleration turned off before this change now also gets its tracking speed written as 0.6875 after the update, even when the system slider was somewhere else. that mouse should keep the speed it already had

Sources/Vorssaint/Services/MouseAcceleration/MouseAccelerationSupport.swift
```diff
@@ -29,13 +29,125 @@ struct MouseAccelerationDeviceIdentity: Codable, Equatable {
...
+    let speed: Double
+}
+
+struct MousePointerProfile: Codable, Equatable {
+    var isEnabled: Bool
+    var disablesAcceleration: Bool
+    var acceleration: Double
+    var speed: Double
+
+    static let disabled = MousePointerProfile(
```
vorssaint, line 85: turning customize on writes these fixed values to the device right away, so a mouse that reports another resolution or another tracking speed jumps before any slider is touched, and the sliders show values the device never had. can the profile start from what the device reports at that moment

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -1133,30 +1153,20 @@ struct MouseSettings: View {
...
-                    PermissionRow(kind: .accessibility)
-                }
-            }
         }
         .formStyle(.grouped)
         .onAppear {
             MiddleClickService.shared.refreshDragGestureConflict()
+            mousePointerService.setSettingsVisible(true)
+            initializePointerDeviceExpansion()
+            if !permissions.accessibility { permissions.requestAccessibility() }
```
vorssaint, line 1162: just opening this tab now triggers the system accessibility prompt and the guide card, with nothing turned on that needs it. the prompt should keep coming only when a feature that needs it is switched on

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -927,13 +927,25 @@ struct MouseSettings: View {
...
+        FeatureStrings.mousePointer(l10n.language)
+    }
+
     var body: some View {
         Form {
+            Section(permissions.accessibility ? mousePointerText.permission : l10n.s.permissionRequired) {
+                Text(mousePointerText.permissionCaption)
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
+                PermissionRow(kind: .accessibility, showsStartOver: true)
```
vorssaint, line 947: this block used to appear only while a feature that needs the permission was on and the permission was missing. now it is always there, with a reset button that is not part of this feature. can you keep the previous behavior here

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -2035,6 +2243,12 @@ struct PermissionRow: View {
                             permissions.openMicrophoneSettings()
                         }
                     }
+                    if showsStartOver,
+                       (kind == .accessibility || kind == .screenRecording) {
+                        Button(FeatureStrings.permissionGuide(.enUS).startOver) {
```
vorssaint, line 2248: this label is always english, while the rest of the row already follows the app language

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -1035,19 +1052,22 @@ struct MouseSettings: View {
...
+                } else {
+                    ForEach(mousePointerService.connectedDevices) { device in
+                        pointerDeviceSection(for: device)
+                    }
                 }
-                .settingsSectionAnchor(.mouseAcceleration)
             }
-            if AppFeature.mouseNavigation.isAvailable {
-                Section(l10n.s.mouseNavigationSection) {
+            if hasConnectedMouse, AppFeature.mouseNavigation.isAvailable {
```
vorssaint, line 1066: these sections now disappear when no mouse is detected, so someone whose bluetooth mouse is off cannot reach them to turn the features off, and search lands on a section that is not there. can they stay visible whenever the feature is installed, as before, and the same for the trackpad section further down

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -1177,6 +1187,203 @@ struct MouseSettings: View {
...
+                Text(mousePointerText.caption)
+            }
+        }
+        .settingsSectionAnchor(.mouseAcceleration)
+    }
+
+    @ViewBuilder
+    private func pointerControls(for device: MousePointerDeviceDescriptor) -> some View {
+        let profile = pointerProfile(for: device)
+        VStack(alignment: .leading, spacing: 14) {
```
vorssaint, line 1321: each card shows five controls at once, plus raw numbers like 0.6875 and a note about dpi. the other sections keep one main switch and put the rest under more options in plain words. can this follow the same shape

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -1177,6 +1187,203 @@ struct MouseSettings: View {
...
+        } header: {
+            if device.id == mousePointerService.connectedDevices.first?.id {
+                Text(mousePointerText.section)
+            }
+        } footer: {
+            if device.id == mousePointerService.connectedDevices.last?.id {
+                Text(mousePointerText.caption)
+            }
+        }
+        .settingsSectionAnchor(.mouseAcceleration)
```
vorssaint, line 1315: every card gets the same jump target, so coming here from search lands on any one of them and highlights all of them

Sources/Vorssaint/Services/MouseAcceleration/MouseAccelerationService.swift
```diff
@@ -149,6 +269,7 @@ final class MouseAccelerationService {
     private static let deviceChanged: IOHIDDeviceCallback = { context, _, _, _ in
         guard let context else { return }
         let service = Unmanaged<MouseAccelerationService>.fromOpaque(context).takeUnretainedValue()
+        service.refreshConnectedDevices()
```
vorssaint, line 272: when a mouse is plugged in or unplugged while this tab is open and nothing is customized yet, does the list of cards update without leaving the tab?

Sources/Vorssaint/Services/MouseAcceleration/MouseAccelerationService.swift
```diff
@@ -213,6 +337,52 @@ final class MouseAccelerationService {
...
+                defaults.double(forKey: DefaultsKey.mousePointerAcceleration)),
+            speed: MouseAccelerationSupport.sanitizedSpeed(
+                defaults.double(forKey: DefaultsKey.mousePointerSpeed))
+        )
+        var updated = profiles
+        for device in connectedDevices where device.kind == .mouse {
+            if updated[device.preferenceKey] == nil { updated[device.preferenceKey] = legacy }
+        }
+        defaults.set(false, forKey: DefaultsKey.mousePointerCustomized)
+        defaults.set(false, forKey: DefaultsKey.mouseAccelerationDisabled)
```
vorssaint, line 382: the panel and the command bar still flip the old all mice switch, and this turns it off on the first per device edit while the profiles stay applied, so those two places stop showing what is really on. can the three surfaces share the same state

Sources/Vorssaint/Core/Defaults.swift
```diff
@@ -47,6 +47,10 @@ enum DefaultsKey {
     static let smoothScrollEnabled = "smoothScrollEnabled"
     static let smoothScrollStep = "smoothScrollStep"      // pixels per wheel tick
     static let mouseAccelerationDisabled = "mouseAccelerationDisabled" // sets HIDMouseAcceleration to -1 for mice
+    static let mousePointerCustomized = "mousePointerCustomized"
+    static let mousePointerAcceleration = "mousePointerAcceleration" // 0...40 HID acceleration/tracking value
+    static let mousePointerSpeed = "mousePointerSpeed" // 0...1 software pointer speed
```
vorssaint, line 52: these three are never written anywhere, so they are stored, backed up and tested but can never change

Sources/Vorssaint/Core/MousePointerStrings.swift
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Vorssaint
+
+import Foundation
+
+/// This new surface intentionally ships with one English source definition.
```
vorssaint, line 6: the other features ship their texts in all the app languages. this one ignores the language, so the whole section stays in english for everyone else and the new language test passes for any language by construction

THIRD_PARTY_NOTICES.md
```diff
@@ -0,0 +1,54 @@
+# Third-party notices
+
+## LinearMouse
```
vorssaint, line 3: the project does not name other apps anywhere, not in notices, comments or tests, and it does not take code from other projects that would need a notice. can you drop this file and the mentions, keeping only what was written for this project

.gitignore
```diff
@@ -17,6 +17,7 @@ dist/
 AGENTS.md
 CLAUDE.md
 E2E/
+.evidence/
```
vorssaint, line 20: this entry is for your local notes and not part of the change. can you leave it out

Sources/Vorssaint/Core/SettingsBackupSupport.swift
```diff
@@ -136,6 +136,8 @@ enum SettingsBackupSupport {
         DefaultsKey.systemShortcutsSuppressed,
         // DDC capability belongs to one physical monitor on one Mac port.
         DefaultsKey.brightnessDDCWriteOnlyPaths,
+        // Hardware identities and the last selected row belong to this Mac.
+        DefaultsKey.mousePointerDeviceProfiles,
```
vorssaint, line 140: these profiles are the whole feature's settings and they are keyed by the device itself, so the same mouse would match on another mac. why keep them out of the backup?

vorssaint:
closing this one. it rewrites the mouse tab around per device profiles, and as it stands
a mouse that only had acceleration turned off gets a new tracking speed written after the update
opening the tab asks for the accessibility permission with nothing turned on that needs it
the existing sections disappear whenever no mouse is detected
the new texts are english only, and the code carries another app's name plus a private system call
the notes above have the details. a much smaller change that keeps the single switch and adds only what is missing could still come in

PR #1670, opened 2026-09-16
feat(privacy): hide shared work without blocking the presenter

NachikethReddyY:
Privacy Share now routes protected output through a stable named window that meeting and recording apps can select. The shortcut changes only that routed output, so the presenter can keep working on the real desktop.

## What changed

- add the Privacy Share feature, dedicated Settings page, custom message, backup coverage, and global shortcut
- expose `Vorssaint Privacy Source` in third-party window pickers for Google Meet, Cap, Teams, and other ScreenCaptureKit-compatible apps
- let the source mirror either a display or one selected window at native Retina resolution without audio
- obscure routed details with three stacked blur passes, overscan, and a translucent veil
- protect Vorssaint recorder frames directly, suppress pointer and typing metadata during blurred intervals, and replace a cached static tail frame when privacy turns on
- show a temporary menu-bar status only while output is blurred, recording, or unavailable
- state the macOS boundary honestly: Screen Recording permission cannot rewrite another app independent capture, so users must select the named source

## Validation

- 50,990 automated checks passed locally
- release build and selftest passed locally
- DMG packaging passed locally
- signed Developer build rebuilt, installed, launched, and selftested
- Cap discovered and recorded the 1680×946/1540×865 named window source
- Google Meet window-sharing path is supported by selecting the same named source; a second-participant transmission test was not performed because it would send the desktop to another person

Generated by GPT-5.6 through T3 Code (Codex harness), with architecture guidance from GPT-6 Astra.


NachikethReddyY:
Screenshots:
privacy-share-settings-clean
Panel
---
file-d3bb981bfea6aaf32da57bf542a770db
Blur
---
image
install and uninstall

NachikethReddyY:
The privacy shortcut currently creates and opens the Privacy Share window even when nothing is being shared. I’m changing the activation path so the shortcut only toggles distortion for an already-active Privacy Share session; with no active share session, it will be a no-op. I’ll add focused regression coverage and verify both states in the rebuilt Developer app.

Change authored by GPT-5.6 through T3 Code (Codex harness).

NachikethReddyY:
Google Meet and Cap exposed a real architecture gap: macOS Screen Recording permission lets Vorssaint read pixels, but it does not let Vorssaint modify another app's capture stream. The hidden-window handoff is therefore not universal.

I'm replacing that path with an explicit, high-quality Vorssaint Privacy Source that remains discoverable in third-party window pickers. Meet, Cap, Teams, and recorders can select that source; the global shortcut then switches only the routed output between visible and protected states while the local desktop stays usable. Vorssaint's own recorder protection remains integrated directly. The UI and menu status will state the boundary rather than claiming other apps' independent raw captures can be rewritten.

Change authored by GPT-5.6 through T3 Code (Codex harness), with architecture guidance from GPT-6 Astra.

NachikethReddyY:
Cross-app source verification on the latest Developer build:

- Cap detects `Vorssaint Privacy Source` as a real 1540×865 window source.
- Selecting it records the routed source instead of the original display.
- The privacy shortcut changes only that source; the working desktop remains usable.
- Google Meet uses the same **A window** picker path: choose `Vorssaint Privacy Source`, not the original screen/window.

The macOS limitation is deliberate and now documented in Settings: Screen Recording permission allows capture, not interception of another app’s independent capture stream. The proof screenshot is saved locally at `.evidence/universal-privacy-source/cap-source-picker.png`; GitHub CLI refused the requested attachment because the PR targets an upstream repository where this account has no direct write access.

Change authored by GPT-5.6 through T3 Code (Codex harness), with architecture guidance from GPT-6 Astra.

NachikethReddyY:
The latest field test exposed a false-positive state: Vorssaint reports “hiding” even while Google Meet is independently sharing the raw entire display. I’m correcting the state model so the app only confirms blur when the selected capture is actually routed through the Vorssaint Privacy Source, and will verify the real Meet picker/feed before pushing.

Change authored by GPT-5.6 through T3 Code (Codex harness).

Issue #829, opened 2026-08-23
Bigger icons in app switcher and apps to show on multiple lines


Ico093:
### Problem or Motivation

I'm was a user of the AltTab tool before they made it paid.

My current setup:
- Simple app switcher
- Preview size: Extra large

My current issues:
- When I have multiple apps open there is a horizontal scroll.
- The app icons/text is too small for me.

### Proposed Solution

Issue 1:
Show the apps on multiple rows. All of them can fit the screen.
1.1 Only if they become too many lower the size of the icons so that they can fit the screen (that should be an extreme case).
1.2 Another approach would be to have a vertical scroll for too many icons if 1.2 isn't feasable.

Issue 2:
Make the app icons even bigger. Example in the mockups from AltTab.

### Feature Area

Windows and Dock

### Mockups / References

<img width="1440" height="376" alt="Image" src="https://github.com/user-attachments/assets/0ab45b98-aa6d-4182-97cb-18f643f1d2d3" />

PathGao:
Which version you are using？we changed a lot recent days so check it after v3.3.3 then report it。Thanks

Ico093:
I'm on 3.3.2 and I don't have updates.

<img width="743" height="547" alt="Image" src="https://github.com/user-attachments/assets/12594194-ddab-48ef-961d-a9b809794266" />

PathGao:
the update is on v3.3.2 beta3. you will get v3.3.3 in few days

HristoDimitrovBede:
Ok, I made some photos as shortcuts aren't working. 

Current behaviour:
![image](https://github.com/user-attachments/assets/f64a9eb7-de8d-440a-acaf-524eabddc1af)

Desired one:
![image](https://github.com/user-attachments/assets/b0af0a67-e69a-4e2b-9089-383186fcf9bc)

Will wait for the update.

PathGao:
> Ok, I made some photos as shortcuts aren't working.
> 
> Current behaviour: ![image](https://github.com/user-attachments/assets/f64a9eb7-de8d-440a-acaf-524eabddc1af)
> 
> Desired one: ![image](https://github.com/user-attachments/assets/b0af0a67-e69a-4e2b-9089-383186fcf9bc)
> 
> Will wait for the update.

you did a fix？pr that

Ico093:
Unfortunatelly no, it is a screenshot of another app - AltTab.

PathGao:
[...]

PR #1672, opened 2026-09-16
feat(switcher): add a vertical simple app layout

GlintonLiao:
## Summary

Add a vertical layout option for the Simple App Switcher. The existing horizontal layout remains the default, so existing preferences and behaviour are unchanged.

The new layout presents each entry as a compact row with the app icon and title. In grouped all-app sessions, hovering an app selects its representative window through the existing App Switcher selection path and shows that app's windows in a fixed-width detail column. The detail column no longer changes the panel width after hover, app rows no longer show a window-count badge, and middle-click close continues to use the shared hovered-window state.

Keyboard navigation follows the vertical list: Up and Down move one visible entry at a time. Pointer-driven selection does not scroll the list under the cursor; keyboard and programmatic selection use nearest-edge reveal so the list moves only when the selected item leaves the viewport. The detail list follows the same reveal policy.

Search and pinned search are shown in a dedicated header above the vertical lists rather than as an overlay. The layout reserves that header's height and reduces the list viewport as needed, keeping the panel within its height limit.

The setting is persisted, included in settings backups, localized across all supported languages, and shown only while Simple App Switcher is enabled.

## Validation

- `swift build` passed.
- `./build/metrics-tests --suite=switcher`: 120 checks passed.
- `./build/metrics-tests --suite=localization`: 4,071 checks passed.
- `git diff --check` passed.
- Added coverage for the persisted layout choice, vertical layout activation, bounded surface dimensions, search-header reservation, visible-row limits, and localization labels.

The automated checks cover layout metrics and switcher policy. The final vertical layout, hover interaction, and search presentation have not been manually exercised in a running app across different display sizes or macOS versions.


GlintonLiao:
### Screenshots

截屏 2026-09-16 at 15 47 39

截屏 2026-09-16 at 16 19 15

When "Show one entry per app" is off

截屏 2026-09-16 at 16 17 17

When "Show one entry per app" is on

vorssaint:
whether a third look for the switcher is worth keeping around is still the maintainer's call, so the notes below are only about the change as it stands.

the automated test build no longer compiles with this change. the scroll test reuses the switcher's search filter, and that filter now calls a helper that the test's own small switcher model does not have. can you get the test build green again

in the grouped list an app row ends up describing one of its windows instead of the app. the screen reader says that window's title first, and the minimized and other desktop badges are that window's even when the app has other windows in a different state. the horizontal tiles name the app and only show a window's state when it is the app's single window, so the rows should do the same

the two english labels read unevenly, "horizontal" next to "vertical list". most other languages say "horizontal row" and "vertical list", so english should match

vorssaint:
closing this one. the reason is scope, not the code. a vertical list is a second full look for the switcher to keep working, and that kind of addition was already turned down for other panels. bigger icons and rows that wrap are the direction still open for the switcher, see #829

PR #1673, opened 2026-09-16
I18n/vietnamese localization

ch1nhcp:
## Summary

Adds Vietnamese (`vi`) as a fully supported app language:

- New `AppLanguage` case, a complete `Strings` catalog, and translations
  across every per-feature `Core/*Strings.swift` catalog and every file
  under `Core/Localizations/`
- Registers `vi` in `Resources/Info.plist`, adds
  `Resources/vi.lproj/InfoPlist.strings`, extends the localization
  coverage tests, and documents Vietnamese in `CONTRIBUTING.md`
- Rebased onto current `main` and filled in the Vietnamese strings for
  features that landed after this branch was started — screenshot
  watermark controls, the dock-preview "current space only" toggle, and
  the notch's hidden-until-hover opening mode. Those fields already
  existed for every other language; Vietnamese was the one catalog
  missing them, which is a compile error here, not a missing string.

Also includes an unrelated one-line build fix found while verifying
this locally: `swift Tools/MakeIcon.swift` runs the icon generator
under the JIT interpreter, which on this machine's toolchain (Swift
6.3.1 against the 26.2 SDK) fails to resolve AppKit symbols even though
the same source compiles and runs fine as a real binary. `build.sh` now
compiles it with `swiftc` like the other build-time helpers and runs
that instead, which also drops its reliance on `argv[0]` to locate the
project root (the working directory, already the project root at that
point in `build.sh`, is the one path that survives being compiled to
another location). Happy to split this into its own PR if preferred —
flagging it here since it's what unblocked full local verification.

## Verification

MacBook Pro (Apple M2 Pro), macOS 26.3 (build 25D125).

- `./build.sh --test` — 52,291 checks pass across every suite, including
  `localization` (4,398 checks): every field in all 14 languages,
  Vietnamese included, matches the English schema, has no empty strings,
  no stray em-dashes, and keeps format-argument positions intact.
- `./build.sh` — finishes clean end to end: `✓ Bundle ready:
  build/stage/Vorssaint.app`. (Adaptive icon generation is skipped with
  its existing "actool not found" message — that needs full Xcode 26+
  and isn't related to this change.)
- Have not done a manual walkthrough of the Vietnamese UI (language
  picker, onboarding, feature panels) on a running build. Would
  appreciate a maintainer or another contributor giving the in-app
  Vietnamese a look before this merges.

## Anything else

No open issue requesting Vietnamese support was found to reference.


vorssaint:
thanks for bringing vietnamese back with the catalogs filled in. by reading it, the structure checks out on my side, every catalog has a vietnamese block with the same fields, order and placeholders as english, and the permission prompt file carries the same keys as the other twelve folders. i could not run the build here, so the rest is from the diff. four things before this can land.

two of the vietnamese permission prompts still say what the english said before it changed. the audio one tells the user nothing is recorded, but that permission now also covers the mac's sound in screen recordings, and the automation one talks about finder, terminal and homebrew while the english now also covers playback control in media apps. can you bring those two in line with the current english so the system dialog does not promise something the app no longer does

the notes file you added at the repo root reads as working notes for this branch, and parts of it are already out of date. the repo keeps that kind of guidance in the contributing guide, so drop the file and fold anything a future translator needs into the paragraph you already touched there

the readme and the contributing guide stopped counting languages on purpose (#921), so "more than a dozen" and the list without a number should stay as they are. adding tiếng việt to the list in contributing is enough

the build script change is a separate topic and touches the release path, so yes to your offer, move it to its own pr where it can be looked at on its own

the conflict with main is on us to sort out at merge time, no need to chase it

PR #1680, opened 2026-09-16
feat(notch): let the music bars follow the player's audio

ruvelro:
## Summary

The island's music bars can now follow the player's actual audio. The option is off by default, needs macOS 14.4 or later, and the bars keep today's synthetic motion whenever the live levels are not available.

### What the bars did

`NotchEqualizerBars` moved each bar on a sine wave keyed only by the clock, with the artwork's tint for color and the play state to pause. It knew nothing of the sound.

### How the live levels are read

- **A private process tap on the player.** `NotchAudioLevelService` watches `NotchMusicService.playback`. While a player is playing, it creates a `CATapDescription(stereoMixdownOfProcesses:)` for that player's process only, with `muteBehavior = .unmuted` (the player keeps sounding as before) and `isPrivate`, read through a private aggregate device on the default output, the same shape `RecorderSystemAudioTap` uses. The tap follows the pid: a different player gets a new tap, the same player keeps its tap across tracks.
- **Nothing blocks the audio thread.** The IO proc mixes the buffer to mono into a preallocated ring (`NotchAudioRing`) guarded by a `try`-lock; when the analysis thread holds it, the audio thread drops that buffer rather than wait. The output side of the aggregate is silenced every cycle, as the recorder does, so the device beneath never plays stale memory.
- **Analysis off the audio thread.** A 30 Hz `DispatchSourceTimer` on a utility queue takes the latest 1024 mono samples, applies a Hann window and a vDSP DFT, and reduces the 512 usable bins to seven bands with edges at 50, 120, 250, 500, 1000, 2000, 4500 and 12000 Hz (about a third of an octave each). Each band's mean magnitude becomes a level on a decibel scale with a floor at -60 dBFS.
- **Motion.** `NotchAudioLevelSupport.Smoother` gives the bars a fast attack (0.55 per update) and a slow decay (0.12), and a slowly relaxing ceiling scales the loudest band to 0.92 so a quiet album still fills the meter. Only `NotchLiveEqualizerBars`, a small wrapper, observes the service, so its thirty updates a second re-render the bars and nothing else on the island. With fewer bars than bands (the resting island shows three), `barIndex` samples the bands evenly instead of crowding into the bass.

### Permission and the fallback

macOS asks for system audio recording permission the first time a tap is created. There is no API to ask whether it was granted: a tap created without it succeeds and only ever delivers silence. So the reader reports levels only once it has heard sound, which keeps the synthetic motion during the tap's warm-up, and if nothing arrives within 2.5 s of playing (`NotchAudioLevelSupport.silenceGrace`) it stops, releases the tap and the bars return to their usual motion. That player is not retried until the option is switched off and on again, so a declined prompt costs one tap, not one per track. The hint under the toggle says as much in all thirteen languages.

The tap is also released 1.5 s after playback stops (a skipped track does not blink the meter), when the island stops, when the Mac's session changes through the island's own lifecycle, and when the option is turned off.

### Privacy

The tap listens to one process, keeps at most 8192 mono samples (about 170 ms at 48 kHz) in memory, and nothing is stored, recorded or sent. `docs/PRIVACY.md` gains a paragraph under optional notch features. The existing `NSAudioCaptureUsageDescription` mentions the mixer and recordings; it may be worth adding the bars to that sentence in the thirteen `InfoPlist.strings`, which this change leaves as it is so the wording stays the maintainers' call.

### Settings

A toggle under Dynamic Island → Music, "Bars follow the music", disabled with an explanatory caption on macOS earlier than 14.4. The key `notchLiveEqualizer` is registered (default false), so it travels with settings backups like the other music options.

### Tunables, for whoever wants a different feel

- `bandEdgesHz`, `floorDecibels`, `updatesPerSecond` and `silenceGrace` in `NotchAudioLevelSupport`.
- `Smoother.attack`, `decay`, `ceilingDecay` and `peakTarget` for how nervous or lazy the bars are.

## Testing

- `./build.sh --test` (50 923 checks) with new `NotchAudioLevelTests`: band ranges cover the spectrum in order with at least one bin each and no DC bin, narrow spectra still get seven bands, silence is flat, a 700 Hz tone lights only its band, bars rise faster than they fall, a quiet source still fills the meter, fewer bars sample the bands evenly, the fallback fires only after the grace period, and the option is off until chosen. The music extras strings test now pins 34 fields.
- `./build.sh --dev --install` and listening on an external display with Spotify: the permission prompt appears once on first play, the bars follow the music afterwards, and they keep their synthetic motion when the prompt is declined.


vorssaint:
a few things need to change before this can go in.

this needs to be a feature in the features hub like lyrics and queue, with the system audio permission listed for it. that is where people find the permission again after declining the prompt, and where the feature can be uninstalled. the sentence macos shows in that prompt should also say the bars use it, today it only mentions the mixer and recordings.

the reader is created and started on the app's main thread each time playback starts. the audio system can stall those calls while a bluetooth or usb device is reconnecting, and then the island and the whole app freeze for as long as they wait. the mixer and the recorder create theirs on a queue of their own, can this one do the same?

when nothing is heard in the first 2.5 seconds the player is given up on until the option is turned off and on again. more than a declined prompt lands in that window, for example someone still reading the first permission prompt, or a player that buffers before it makes sound. a player that makes its sound from a helper process, as some browsers do, would never be heard at all since only the main process is tapped. could it try again on the next play instead of holding the player for the whole session, and did you try a browser?

when playback moves from a working player to one that was given up on, the first player's reader is left running and the bars keep showing its last levels instead of going back to the usual motion.

the reader stays tied to the output device it started on. the recorder rebuilds its tap when the output device or its sample rate changes because the read otherwise goes silent or off pitch, and here that leaves the bars stuck while the music keeps playing elsewhere. can it follow the device the same way?

the bars next to the track title in the open island still use the old motion while the compact bars follow the music, so opening the island looks like it drops the sound.

the korean hint is missing a space after its third sentence

ruvelro:
All seven addressed, rebased on main, in b0ef95f.

**Features hub and the permission.** The bars are now `AppFeature.notchLiveEqualizer`, in the tools group with `permissions: [.audioCapture]`, so the permission is listed for them in the transparency portal and the feature can be uninstalled like lyrics or the queue. `activeFeatures` counts it only while the island is on and the music module is not hidden, `FeatureRuntime` re-syncs the reader when availability changes, and the toggle in settings is disabled once it is uninstalled. The hub entry reuses the strings the feature already ships, so no translation is left behind. The sentence macOS shows when it asks now names the bars too, in `Info.plist` and in all twelve `InfoPlist.strings`.

**Off the main thread.** Creating the tap, building the aggregate, `AudioDeviceStart` and every teardown now run on the reader's own queue, the way the mixer and the recorder do. `start()` no longer returns a result — it reports through `onUnavailable`, since the caller is not waiting.

**Giving up.** It now costs one play, not the session. `SilenceMemory` remembers the track it went silent on; the next track, or the next time play is pressed, reads again. A pause rearms it. The grace also grows from 2.5 s to 8 s, which costs nothing: the bars show their synthetic motion throughout either way, so the only thing a longer wait holds is a tap that would have been released anyway. That covers a prompt still being read and a player that buffers before it sounds.

**Browsers.** You were right, and it could never have worked: the tap followed the player's own pid. It now gathers every audio process that player is responsible for, the way the mixer gathers its rows. On this Mac, `com.brave.Browser` (pid 1337) makes no sound itself — its two `com.brave.Browser.helper` processes do, both responsible to 1337. Safari and Mail are the same through `com.apple.WebKit.GPU`. Spotify is a single process, which is why it was the one that worked.

And yes, tried with a browser now. That took a second fix, sent separately as #1713: with Spotify open but stopped, a video playing in Brave left the island blank, because a music app holding a track outranked every other player whatever it was doing. With that in place the bars follow a YouTube video in Brave, in sync with the voice.

**The orphaned reader.** The previous player is released before this one is considered, so the bars fall back to their usual motion instead of freezing on somebody else's levels.

**Following the device.** `watchDefaultOutputDevice` and `watchSampleRate`, rebuilding only on a real change so the notifications a rebuild raises settle instead of looping, as in `RecorderSystemAudioTap`. The ring keeps what it has heard, so changing headphones does not restart the silence countdown on a player already known to sound.

**The open island** uses the same live bars beside the track title, so opening it no longer looks like the sound dropped. **Korean** has its space.

Also updated `docs/PRIVACY.md` for the helper processes, the device change and the uninstall path, and added a contract test for the give-up policy. `./build.sh` clean, `--selftest` OK, `--test` OK (52978 checks; the catalog goes from 66 to 67 features).

ruvelro:
One more in 8c95fb2, found while checking that the helper-process grouping holds for every browser rather than the one I tested.

The tap was built from the processes that existed when the reader started. A browser shuts its audio process down after a while idle and opens another one for the next sound, and from then on the tap held nothing: no buffer ever arrived, the ring kept answering with the last samples it had, and the bars froze on those levels while the music played on. Most of the time a stop comes first, which releases the reader and makes the next play build a fresh tap, so it only shows when the sound moves while playback never stops.

The reader now watches `kAudioHardwarePropertyProcessObjectList` and gives its tap back when one of the processes it was built from leaves; the service reads the same player again from scratch, so the wait for sound starts over too. Only a process leaving counts — one arriving is the ordinary churn of a browser opening tabs, and the tap in place keeps working through it.

`./build.sh` clean, `--selftest` OK, `--test` OK (52983 checks).

Sources/Vorssaint/Services/Notch/NotchAudioLevelService.swift
```diff
@@ -0,0 +1,601 @@
...
+            AudioObjectRemovePropertyListenerBlock(AudioObjectID(kAudioObjectSystemObject),
+                                                   &address, queue, deviceListener)
+            self.deviceListener = nil
+        }
+        if let processListener {
+            var address = Self.address(kAudioHardwarePropertyProcessObjectList)
+            AudioObjectRemovePropertyListenerBlock(AudioObjectID(kAudioObjectSystemObject),
+                                                   &address, queue, processListener)
+            self.processListener = nil
+        }
```
vorssaint: removing a listener registered as a block never matches the one that was added, the call reports success and the listener stays, so every reader leaves its two listeners behind and they keep firing for the rest of the session. the mixer registers with the plain callback and a client pointer because that one does come off, can these do the same

Sources/Vorssaint/Services/Notch/NotchAudioLevelService.swift
```diff
@@ -0,0 +1,601 @@
...
+
+    /// A browser shuts its audio process down after a while idle and opens
+    /// another one for the next sound, and the tap it was built from then
+    /// hears nothing at all. Only a process leaving is worth a new tap: one
+    /// arriving is the ordinary churn of a browser opening tabs, and the tap
+    /// in place keeps working through it.
+    private func restartIfTappedProcessLeft() {
+        guard !stopped, !tapped.isEmpty else { return }
+        let current = Set(AppVolumeMixer.audioProcessObjects())
+        guard !tapped.allSatisfy(current.contains) else { return }
```
vorssaint: a process of the player that only shows up after the tap was built is never heard, since only one leaving rebuilds, so a browser that opens its audio process with the next sound, or a moment after play is pressed, keeps the usual motion for that whole play. can one of the player's processes arriving rebuild the tap as well

Sources/Vorssaint/UI/Settings/FeatureHubSettings.swift
```diff
@@ -825,6 +826,7 @@ extension AppFeature {
         case .notchNotifications: return FeatureStrings.notchNotifications(L10n.shared.language).description
         case .notchLyrics: return FeatureStrings.notchMusicExtras(L10n.shared.language).lyricsDescription
         case .notchQueue: return FeatureStrings.notchMusicExtras(L10n.shared.language).queueDescription
+        case .notchLiveEqualizer: return FeatureStrings.notchMusicExtras(L10n.shared.language).liveEqualizerHint
```
vorssaint: the hub shows this four sentence hint as the description, where every other feature has one short line, and the first run picker cuts it off after two lines. can it get a one line description of its own, like lyrics and the queue

Sources/Vorssaint/Core/FeatureCatalog.swift
```diff
@@ -274,6 +276,9 @@ extension AppFeature {
         case .notchGestures: return []
         case .notchTimer, .notchAccessories: return []
         case .notchLyrics, .notchQueue: return []
+        // The bars read the player's own audio output, which macOS gates
+        // behind the same permission the mixer and the recorder ask for.
+        case .notchLiveEqualizer: return [.audioCapture]
```
vorssaint, line 289: the permission page still explains this permission as the mixer and recordings only, while the bars now show up under it. can that sentence name the bars too, the way the prompt does now

Sources/Vorssaint/Services/Notch/NotchAudioLevelService.swift
ruvelro: Fixed in 73e9bfa6. Both global listeners and the aggregate's sample rate listener now register with the plain callback and a client pointer, as the mixer does, so the removal actually matches and nothing is left firing.

One difference from the mixer: it can pass `Unmanaged.passUnretained(self)` because it is a single instance that lives as long as the app, and a reader is not — it lives for one play, and a notification can still be on its way from a HAL thread when the last one goes away. So the client pointer here is a number, never the reader's address: `NotchAudioLevelListeners` holds the reader under that number until it has given its listeners back, and a number that has been forgotten simply finds nothing. The callbacks arrive on a HAL thread rather than the queue the block form gave them, so each one hops to the reader's queue before touching anything.

Sources/Vorssaint/Services/Notch/NotchAudioLevelService.swift
ruvelro: You are right, and it is fixed in 73e9bfa6. Any change in the player's own processes now answers, not only one leaving.

The two cases are not worth the same, though, so `NotchAudioLevelSupport.tapChange` decides between them:

- **One arrives** — the processes already heard never stopped, so the tap is built again around all of them and the ring keeps what it has heard. The bars carry on and the wait for sound is not started over. That covers the tab that starts playing a moment after the tap was built.
- **One leaves** — what the ring holds is no longer the player, and reading on would freeze the bars on its last levels, so the player is read from scratch as before.

The process list stirs for every app on the Mac, so the player's processes are read once the burst has settled (0.25 s) rather than on each notification. Six contract tests cover the policy, including one leaving while another arrives, which still counts as leaving.

Sources/Vorssaint/UI/Settings/FeatureHubSettings.swift
ruvelro: Done in 73e9bfa6. The bars now have a `liveEqualizerDescription` of their own, one line like lyrics and the queue: "Move the Dynamic Island bars with the sound your player makes." Translated in all thirteen languages, following each one's own wording for the island (ru and tr call it the notch, as they already do for lyrics).

The four sentence hint stays where it belongs, under the toggle in Dynamic Island → Music, where the permission and the fallback need explaining.

Sources/Vorssaint/Core/FeatureCatalog.swift
ruvelro: Done in 73e9bfa6. `explainAudioCapture` now names the bars in the same order the prompt does — mixer, bars, recordings — in all thirteen languages. In English: "Lets the mixer adjust each app's volume, the Dynamic Island bars follow the music and screen recordings include the Mac's sound."

ruvelro:
All four addressed in 73e9bfa6.

**The listeners.** They registered as blocks, which never matches on removal, so every reader left its listeners behind for the rest of the session. They now use the plain callback and a client pointer, the way the mixer does. Unlike the mixer, a reader is not a single long-lived instance — it lives for one play, and a notification can still be on its way from a HAL thread when the last one goes away — so the pointer is a number to look up rather than the reader's address.

**A process arriving.** Any change in the player's processes now answers. One arriving builds the tap again around all of them and keeps what the ring has heard, so the bars do not drop while a tab starts playing; one leaving still reads the player from scratch, since the ring's samples are then stale and would freeze the bars. The list is read once a burst has settled rather than on each notification, as one app launching stirs the whole thing. The policy lives in `NotchAudioLevelSupport.tapChange` with six contract tests.

**The hub** has a one line description of its own now, like lyrics and the queue, in thirteen languages; the long hint stays under the toggle in settings.

**The permission page** names the bars in the same order the prompt does, in thirteen languages.

`./build.sh` clean, `--selftest` OK, `--test` OK (52983 → 52989 checks).

One note on the local run: the first full pass failed 22 checks in `SwitcherScrollTests`, and they are timing-dependent — the suite passes on its own (160/160) and the full suite passes on a second pass. Nothing here touches the switcher.

ruvelro:
Correcting the note at the end of my last comment about the 22 `SwitcherScrollTests` failures. I said they looked timing-dependent, having seen the suite pass on a second run. Checked properly since: they reproduce on a clean `origin/main` checkout here, same 22 checks, with the switcher sources byte-identical on this branch. So it is not this branch and not flakiness — it is something about this machine, since CI runs the same suite green. Sorry for the noise.

ruvelro:
Merged `main` in 1c6d1730, which needed a decision about the bars themselves.

`main` moved them into `NotchEqualizerBars.swift` and rewrote them in 8de49f46 to animate in the compositor — `CALayer` heights driven by a repeating `CABasicAnimation` — instead of a SwiftUI timeline redrawing a `Canvas`. The live levels follow them there rather than keeping the old view alongside: when levels are present the bar's height is set straight onto the layer and no animation is started, since the reader already smooths its levels and sends thirty a second, and a repeating animation would fight the sound it is following. Without levels, your synthetic motion is untouched, including the occlusion and visibility gating.

`NotchLiveEqualizerBars` stays in `NotchComponents.swift`: the bars' own file is in `TEST_SOURCES` now and the audio service is not, so the view that observes it has to live outside that file.

And a second correction about those `SwitcherScrollTests` failures, since my first correction was wrong too. They were `main`'s: reproducible on a clean `e15db436` checkout, gone with `main` at 6e2955e1. Not this branch, and not this machine as I guessed second time round.

`./build.sh` clean, `--selftest` OK, `--test` OK (54281 checks, switcher included).

Issue #1696, opened 2026-09-17
Notch hides its own contents when any status-level window exists below the menu bar


austinmax323-del:
### Description

`NotchMenuBarSpace.measure()` treats **any** window in the status-window level band as menu-bar occupancy, without checking whether that window actually intersects the menu bar. A small floating overlay anywhere on screen — even hundreds of points below the bar — makes the measurement lose its safe center, and the notch then hides its own contents (album art, audio pill) entirely.

The filter in [`Sources/Vorssaint/Services/Notch/NotchMenuBarSpace.swift#L33-L42`](https://github.com/vorssaint/vorssaint-utils/blob/main/Sources/Vorssaint/Services/Notch/NotchMenuBarSpace.swift#L33-L42):

```swift
guard let number = window[kCGWindowNumber as String] as? Int, number != ownWindow,
      let layer = window[kCGWindowLayer as String] as? Int,
      layer >= Int(CGWindowLevelForKey(.statusWindow)),
      layer <= Int(CGWindowLevelForKey(.statusWindow)) + 1,
      ...
      height > 0, height <= geometry.menuBarHeight + 2 else { continue }
occupied.append(CGRect(x: x, y: primaryTop - y - height, width: width, height: height))
```

The rect is appended regardless of its `y`. Every other contributor to `occupied` is a real menu bar item, so a window below the bar is counted as if it sat in the bar. `applyMenuSpace(nil)` then hides the cutout — which is correct behaviour for a genuinely occupied bar, just acting on a false positive here.

Any small always-on-top window qualifies: a HUD, a lyric/subtitle strip, a floating timer.

### Steps to Reproduce

1. Set the notch to show music at rest — `notchIdleContent = music`, `notchLyricsEnabled = true` — and play something so the album art and audio pill appear in the cutout.
2. Run any app that creates a borderless window with `level = .statusBar`, narrower than the screen, height ≤ menu bar height, positioned well below the menu bar (mine sits ~40pt down).
3. The album art and audio pill disappear while that window exists, and come back when it closes.

I confirmed causality with an A/B at 12s idle in both states, same frontmost app and same playback: overlay off → art and pill present; overlay on → gone. Moving the window 60pt further down changed nothing, which is what pointed at the missing `y` check rather than overlap. My workaround was to sit at `statusWindow + 2`, outside the band.

Repro app, if useful: https://github.com/austinmax323-del/notch-lyrics

### Suggested fix

[...]

PR #1697, opened 2026-09-17
fix(notch): only count menu bar windows that reach the bar

austinmax323-del:
`NotchMenuBarSpace.measure()` appended every status level window it found to `occupied`, matching on level, width and height — but never on position. A small always-on-top overlay anywhere on the display satisfies all three, so it was counted as menu bar occupancy. The safe center is then lost and the notch hides its own contents: with a 28pt overlay on screen, the album art and audio pill disappear from the cutout entirely, and come back the moment it closes.

Every other contributor to `occupied` is a real menu bar item measured through AX, so the branch directly above already guards with `occupied.contains(where: { $0.intersects(bar) })`. The window branch is the one member of that mechanism that never joined it.

### Change

- `statusItemRect(...)` lifts the per-window decision out of `measure()` so the filter can be exercised without AX or the window server. `measure()` keeps its shape.
- The rect must intersect the `bar` already built above before it counts.
- `NotchMenuBarSpace.swift` added to `TEST_SOURCES` in `build.sh` — it was not on the list, so the type was not visible to the test target at all.

### Test

`statusItemOccupancyContracts` in `Tests/NotchTests.swift`, three cases: a status level window inside the bar occupies it, the same window 400pt lower does not, and the notch's own window is excluded wherever it sits. Values are derived from the geometry in the test rather than copied out of the source.

Checked both directions: green as submitted, and with only `guard rect.intersects(bar)` reverted it fails with `a status level window below the menu bar does not occupy it`.

### Verified

- `./build.sh` — completes, no warnings
- `./build.sh --test` — `TESTS OK (51532 checks)`

### Not checked

I reproduced the defect against the released build (3.4.0-beta.2.1) and fixed it against `main`, but **I have not run the patched build itself** — swapping the installed app for a dev build would reset the Accessibility and Screen Recording grants tied to its signature on this machine. So the evidence here is the reproduction, the unit test and the symmetry with the AX branch, not a before/after on a running patched app. Worth a second pair of eyes on a machine where that swap is cheap.

Also single machine, single display, one Space, English, built locally on macOS 26.3.1 (Apple Silicon). The multi-display path through `primaryTop` is unexercised by this test beyond the arithmetic already there.

Repro overlay, if it is useful for testing: https://github.com/austinmax323-del/notch-lyrics — any borderless `.statusBar` window narrower than the screen and no taller than the menu bar reproduces it.

Refs #1696


Sources/Vorssaint/Services/Notch/NotchMenuBarSpace.swift
```diff
@@ -30,22 +30,38 @@ enum NotchMenuBarSpace {
...
+              let bounds = window[kCGWindowBounds as String] as? [String: CGFloat],
+              let x = bounds["X"], let y = bounds["Y"],
+              let width = bounds["Width"], let height = bounds["Height"],
+              width > 0, width < screenWidth - 2,
+              height > 0, height <= menuBarHeight + 2 else { return nil }
+        let rect = CGRect(x: x, y: primaryTop - y - height, width: width, height: height)
+        // Level and size alone do not make a window a status item: a small
+        // always on top overlay anywhere on the display matches both. Only a
+        // window that reaches the bar takes room from it.
+        guard rect.intersects(bar) else { return nil }
```
vorssaint, line 61: windows outside the menu bar are already ignored when calculating the available space. can you show a case where this change fixes the island disappearing?

austinmax323-del:
You are right, and my premise was wrong. `NotchMenuBarLayout.sideRoom` already discards anything that does not reach the bar:

```swift
for rect in occupied where rect.intersects(bar) {
```

So an overlay below the menu bar never affects the available space, and this change cannot be what fixed the symptom I reported. I should have read `sideRoom` before writing the guard rather than after.

I went back and re-ran the comparison with the frontmost app held constant (Finder, Spotify playing, 10s settle, three runs):

| overlay | island |
|---|---|
| none | album art + audio pill present |
| `.statusBar`, 28pt, below the bar | present |
| `statusWindow + 2` | present |

The island is fine at `.statusBar` too. My original A/B did not pin the frontmost app between the two halves, which is the variable I now think moved — and it is the one your code legitimately reacts to. I cannot reproduce the disappearance at all today, so I have no case to show you.

I am closing this. The remaining question is mine, not yours: something did hide the cutout's contents repeatedly on my machine, and I no longer have a mechanism for it. If it comes back I will reproduce it properly before filing again, with the frontmost app, the app language and the display layout recorded.

If any of it is worth keeping on its own terms — `statusItemRect` as a testable seam, and `NotchMenuBarSpace.swift` being absent from `TEST_SOURCES` so the type is invisible to `./build.sh --test` — say so and I will send that separately, without the behavioural change. Sorry for the round trip.

PR #1699, opened 2026-09-17
fix(notch): measure compact strip margins from the silhouette

ruvelro:
## Before

A compact strip is barely taller than its own rounded corners, so its lower half is one long arc and there is almost no straight edge to measure against. Every strip measured its margins as if there were one.

- **Music.** The cover cleared the vertical edge by 4 pt but passed about **2 pt** from the corner arc, which is what you see: artwork glued to the curve. The equalizer wing was equally tight at ~1.9 pt.
- **Timer.** Both wings sized their content and then centred it, so the paddings never decided anything. With a 37 pt menu bar the icon floated some **29 pt** from the edge while the remaining time drifted against the opposite one.
- **Downloads.** Its 10 pt leading padding started *inside* the 10.4 pt shoulder.

## After

`NotchGeometry.compactActivityEdgeInset(boxHeight:radius:gap:)` derives the inset from the silhouette: content is pushed in until its own rounded corner keeps the same distance from the arc that it keeps from the top and bottom edges. One number, `NotchLayout.compactEdgeGap` (5 pt), now governs the breathing room of every compact strip, and both wings anchor to their own edge instead of centring their content.

| | before | after |
|---|---|---|
| Music cover, corner clearance (menu bar 32/37/40/44 pt) | 1.3 / 2.0 / 2.3 / 2.6 pt | 5.0 pt in every case |
| Timer icon, from the leading edge | ~29 pt (centred) | 15.4 pt (anchored) |
| Timer digits | drifted with the text's width | 16.7 pt from the trailing edge |
| Download arrow | 10 pt, inside the shoulder | 15.4 pt |

The cover keeps its 26 pt side wherever the strip is tall enough; on a short menu bar it gives up a couple of points rather than the clearance, and on a narrow wing the inset yields first so the cover never shrinks below 20 pt. `NotchLayout.surfaceRadius(height:)` also replaces the `min(28, height / 2)` that was copied into three files, so the silhouette and its contents cannot drift apart.

`./build.sh` clean, `--selftest` OK, `--test` OK (51529 checks). Checked on the Developer build with music playing.


Sources/Vorssaint/UI/Notch/NotchDownloadsView.swift
```diff
@@ -149,8 +160,8 @@ struct NotchDownloadStrip: View {
                         }
                     }
                 }
-                .padding(.trailing, service.geometry.compactActivityWingWidth >= 36 ? 10 : 0)
-                .frame(width: service.geometry.compactActivityWingWidth).clipped()
+                .padding(.trailing, service.geometry.compactActivityWingWidth >= 36 ? percentInset : 0)
```
vorssaint: with little room beside the camera, a download near completion can now show its percentage on two lines. for example, 99.9% rounds to 100%, which fitted on one line before this change but no longer fits with the new margin. could you keep the full percentage on one line when space is tight, including in other languages?

ruvelro: Good catch, and it was my margin that caused it: the wider inset left no room for a rounded hundred, and four of the languages (es, de, fr, ru) write `100 %` with a space before the sign, three points wider still.

Fixed in c9d3234:

- The reading is pinned to one line and scales down rather than wrapping or losing a digit. In the narrowest wing the strip ever draws (44 pt, since `musicStrip` keeps its wings at 88 pt or retracts them) the widest reading needs 0.86 of its size, so the 0.75 floor clears it.
- It now formats with the app's language instead of the system's, so what is measured is what is drawn.
- The strips also size their own content within the room the margin leaves: the timer icon and the download arrow asked for 20 and 17 pt even on a 24 pt menu bar, where no inset could have given them the gap.
- `compactMarginContracts` in `Tests/NotchActivityTests.swift` recomputes the distance from each box to the curve independently, over a matrix of bar heights and side room, and checks the widest reading in all thirteen languages against the narrowest wing.

PR #1705, opened 2026-09-17
fix(display): respect closed lids and restore deferred displays

rkrkrkk:
Vorssaint issued a built-in display enable while the lid was closed. This fix blocks that transaction, retains automatic restoration intent, and restores the display when the lid opens while the app is running. Manual attempts while closed receive a localized instruction to open the lid.

Sanitized evidence from 2026-09-17, UTC+08:00:

```text
18:02:22.793 powerd: ClamshellState. Closed : 1
18:51:17.089 Vorssaint: Display 0x00000001: kCGSDisplayConfigEnable on
18:51:17.119 WindowServer: display_id=1, current_state=on
18:51:21.051 powerd: Sending clamshell state closed: 0 to WindowServer
```

No intervening lid-open event was recorded. The app-issued enable and completed transition are established; startup restoration is inferred from timing/source, without a captured call stack or saved preference. The installed binary's source commit is unproven.

The shared transaction entry checks built-in identity and `AppleClamshellState`. Open/unknown lid state, external operations, and disables preserve existing behavior. Startup, feature-stop/termination, and headless restoration share the guard. Automatic closed-lid denials retain recovery records and own one `IOPMrootDomain` general-interest observer. Subscription precedes an asynchronous recheck; recovery runs on main outside the callback, independently of topology changes and the display-control feature's lifetime. Success clears persisted and managed disabled state, removes the obsolete panel error, and releases observation. Failed/deferred attempts retain the error; headless recovery preserves the actual denial reason. Main-thread manual completion publishes before queued recovery so an old error cannot overwrite success. Failures retain intent without looping on transaction notifications. Intentionally disabled displays without deferred intent stay disabled; successful manual toggles supersede older deferred requests.

Manual denial now says “Open the lid to turn on the built-in display.” in all 13 locales. The reason comes from the transaction-time result, covering lid movement after clicking. Blocked manual clicks do **not** queue recovery; genuine configuration failures keep the generic error. There are no new timers, settings, permissions, helpers, or keep-awake changes. A fully exited app cannot recover until its next launch.

Two hardware probes of the previous guard—including a fresh process started while closed—left the internal panel offline/inactive for a 10-second observation window after opening, until explicit enable restored it. That was a test observation window, not a delay in the new code. Panel power-on alone was insufficient.

Sequential local validation passed: warning-free `./build.sh`, `--selftest` (`SELFTEST OK`), and `./build.sh --test` (55,421 checks plus cleanup). Controlled tests exercise the new production methods, including subscription races, callback timing, ownership, cleanup, manual lid races, and failure retention. Removing deferred-intent recording causes six failures. Tests emitted nonfatal sandbox-extension diagnostics. Real IOKit registration succeeded without display transactions; **new notification-driven physical recovery and full installed-app relaunch remain untested**.

Environment: Mac14,5/M2 Max, macOS 27.2 (26B5086k), Studio Display, extended arrangement; Spaces layout uncaptured. Historical app: 3.3.5 (86). Local compiler: Swift 6.4; Swift 6.0.3 remains a CI check.

Synced upstream main at `aa21b2f8` in merge commit `2416b86f`, preserving the upstream scroll-modifier and KeepAwakeTimerHandoff registrations alongside display-restoration. Display behavior is unchanged by this conflict resolution.

Headless recovery now cancels a queued built-in restoration created only by that headless attempt once another candidate restores successfully, so opening the lid does not enable a second screen. Independent restore-all requests remain owed. Regression coverage includes repeated attempts, failed restore-all promotion, deferred retry lid races, observer cleanup, and late opening.


vorssaint:
one thing i could not settle from reading. after this change, when the app quits or restarts while the lid is closed, the built-in screen stays switched off and the app only tries again on its next launch. if someone then opens the lid, does the screen come back on its own, or does it stay dark until the app runs again? if it stays dark, please make the app bring it back when the lid opens instead of waiting for a relaunch, since after a restart there is no button in the panel for it either. also, tapping to turn the built-in screen back on while the lid is closed now ends with the message that the display could not be changed. saying the lid is closed would make that clearer

rkrkrkk:
@vorssaint Following up on [your comment](https://github.com/vorssaint/vorssaint-utils/pull/1705#issuecomment-5716234428): addressed in 189d75e61b75052a37a1c58ec342aad11100a6df.

We verified the recovery gap on hardware with the previous guard. In two probes, including a fresh process started while closed, opening the lid left the explicitly disabled panel offline/inactive for the 10-second observation window. An explicit enable then restored it, matching the user's observation that the picture appeared only after about 10 seconds. These were shared-code probes, not full installed-app restart tests.

The update adds both requested behaviors:

- Automatic restoration blocked by the closed lid retains its request and observes lid changes through IOPMrootDomain. Opening the lid triggers restoration while the app is running, even without a display-topology change, with display control disabled, or after feature stop. Success clears recovery records and releases the observer. Only deferred automatic requests are retried; unrelated intentionally disabled displays stay off. A fully exited app cannot observe opening; on its next launch it restores immediately if open, or waits for opening if closed.
- Manual enable rejected while closed now says “Open the lid to turn on the built-in display.” All 13 locales are covered, and the reason comes from the transaction-time check. Genuine configuration failures retain the generic message.

Local validation passed: warning-free build, SELFTEST OK, and 51,582 checks plus preference cleanup. Tests execute the production recovery methods against controlled system boundaries; removing deferred-request recording causes six failures. Actual notification-driven physical recovery with the new code and full installed-app relaunch remain unverified. The PR description now includes these results and limits.

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -1273,7 +1348,7 @@ final class BrightnessService: ObservableObject {
         DispatchQueue.main.async { [weak self] in
             guard let self else { return }
             var restored: CGDirectDisplayID?
-            for id in candidates where Self.configureDisplay(id, enabled: true) {
+            for id in candidates where self.restoreDisplay(id) {
```
vorssaint: with the lid closed this now leaves the built-in display waiting for the lid to open, but the panel still ends up saying the display could not be changed, and that text stays after the lid opens and the screen comes back on its own. can that message clear once the screen is back

rkrkrkk: Fixed in ce283cda6ea2a45d07dd749c4aa3ffdd73c2da1e. Successful restoration now clears the obsolete panel error through the shared recovery path; failed or deferred attempts retain it. Headless recovery preserves the closed-lid reason instead of always reporting a generic failure. I also fixed publication ordering so a queued manual completion cannot put an old error back after recovery succeeds.

Regression tests now execute the actual error-publication and headless methods. The pre-fix run reproduced five failures; all 22 focused checks pass with the fix. Sequential validation passed: warning-free `./build.sh`, `SELFTEST OK`, and all 51,594 checks plus preference cleanup. New notification-driven physical recovery/full installed-app relaunch remains unverified; no installed app or live display configuration was changed for this follow-up.

rkrkrkk: Merged upstream main at `79c099f9` in `9f19d00bdf0db4ff8d0029641d1d40d0028b30d2`. The test-generator insertion conflict is resolved with both upstream and display-restoration registrations preserved; the display fix is unchanged. Sequential validation passed: warning-free build, `SELFTEST OK`, and all 54,895 checks plus preference cleanup (including 22 display-restoration checks). No new physical validation is claimed.

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -1273,9 +1352,14 @@ final class BrightnessService: ObservableObject {
...
             guard let self else { return }
             var restored: CGDirectDisplayID?
-            for id in candidates where Self.configureDisplay(id, enabled: true) {
-                restored = id
-                break
+            var failure: DisplayControlFailure = .closedLid
+            for id in candidates {
+                let result = self.restoreDisplay(id)
+                if result == .success {
+                    restored = id
```
vorssaint: with the lid closed the built-in display is queued here before the next candidate is tried, and it stays queued even when that next one comes back on, so opening the lid later turns on both displays where before only one came back. can the queued one be dropped once another candidate succeeds

Sources/Vorssaint/Services/Display/BrightnessService.swift
rkrkrkk: Synced upstream main at `aa21b2f8` in merge commit `2416b86f`. The test registration conflict was resolved while preserving the display-restoration registration and the upstream scroll-modifier plus KeepAwakeTimerHandoff registrations; production display behavior is unchanged. Sequential validation passed: warning-free build, `SELFTEST OK`, 55,409 checks, and cleanup. No new physical validation is claimed.

Sources/Vorssaint/Services/Display/BrightnessService.swift
rkrkrkk: Fixed in `a751c116b3c86f492e9f90fdd4fc7c3a1aad3e98`. Deferred restoration now keeps request provenance in one shared state: a new restore-all request promotes a headless-only request, retries preserve provenance, and a successful headless candidate cancels only remaining headless-only requests. The built-in off record and independent restore-all obligation remain retained. Regression coverage drives the lid-selection/transaction-time race, repeated attempts, failed restore-all promotion, observer cleanup, and late opening: the old production path fails 3 checks and the fixed path passes 34. Sequential validation passed: production-equivalent build, `SELFTEST OK`, 55,421 checks, and cleanup. No new physical validation is claimed.

PR #1713, opened 2026-09-17
fix(notch): show the player that is actually sounding

ruvelro:
## The report

Spotify open but **stopped**, a YouTube video playing in Brave: Dynamic Island shows nothing at all. No artwork, no bars, a black strip. Closing Spotify makes the video appear immediately.

## Why

`NotchPlaybackSource.preferred` ranked every music app above every other player, whatever each one was doing:

```swift
for candidates in [music.filter(\.isPlaying), music] { ... }
```

A stopped Spotify still owns a track, so it won the second pass and became the island's player. The compact strip only appears for playback in progress (`showsMusicActivity(isPlaying:)`), and Spotify was paused — so the island had a player and nothing to show, while the browser played on.

Confirmed on this Mac through the app's own Now Playing adapter: with the video playing, `com.brave.Browser` reports `rate=1` and an elapsed time that advances, so the session is read correctly. It simply lost the ranking.

## The change

Music apps still come first, but only among players **in the same state**:

```swift
for candidates in [music.filter(\.isPlaying), other.filter(\.isPlaying), music] { ... }
```

- Playing music beats everything — `a browser video cannot take controls from playing music` still holds.
- What is sounding beats what is paused, so the island stops going blank.
- Paused music keeps its place once nothing is sounding, so its resume control stays reachable.
- A player that is not a music app is still only taken where it always was: when the system points at it, or when it is the one already being followed. An unrelated remembered video never becomes a fallback.

This does reverse one deliberate rule: a paused music app no longer outranks a playing browser, so while a video plays you reach the video's controls rather than Spotify's. The two tests that pinned the old order are rewritten to say what they now protect, and a regression test carries the reported case by name.

`./build.sh` clean, `--selftest` OK, `--test` OK (52972 checks).


Tests/NotchMusicHardeningTests.swift
```diff
@@ -159,8 +159,20 @@ enum NotchMusicHardeningTests {
...
+               "a video playing takes the island from music paused in the background")
+        expect(choose([browser, paused]) == browser,
+               "the same holds on a first read, with nothing remembered")
+        let idleBrowser = source(20, music: false, playing: false)
+        expect(choose([idleBrowser, paused], previous: 10) == paused,
+               "pausing music keeps its resume control reachable once nothing is playing")
+        expect(choose([idleBrowser, paused]) == paused, "reopening the music surface can still reach paused music")
+        expect(choose([browser, paused, other], previous: 10) == other,
+               "playing music still outranks a playing browser and a paused music app")
+        // Spotify open but stopped, a video playing in the browser: the island
```
vorssaint: this comment names a third party app. can you describe the case by behavior instead, a music app open but stopped

ruvelro: Done in fa844f08. The comment now describes the case rather than naming an app: a music app open but stopped while a video plays in the browser, which is what the ranking got wrong. The assertion below it already read that way, so the two now say the same thing.

ruvelro:
Merged `main` in 92f01562. Both sides had added entries to the same unreleased Fixed list, so all four are kept.

Issue #521, opened 2026-08-06
Linearmouse like mouse scroll modifier actions


lmilojevicc:
### Problem or Motivation

It would be nice if we could have configurable mouse scroll actions like [Linearmouse](https://github.com/linearmouse/linearmouse) has.

### Proposed Solution

Allow for modifiers during scroll to:
1. Alter orientation
2. Change Speed
3. Zoom
4. Pinch Zoom

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/4dd6a6c7-3b8f-478c-819a-e45083dae757" />

### Feature Area

Mouse and keyboard

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

Blocked on PR #626, which is open but currently conflicting and needs a rebase onto the rewritten main (its reported 125 files / +8815 is a history-rewrite artifact).

Coverage note, because "blocked on #626" understates it: #626 covers scroll acceleration only. The modifier-key scroll actions this issue asks for, and Shift-to-horizontal-scroll, are not in it. So #626 landing is the prerequisite, and this still needs its own work afterwards — it is not covered by the PR.

This is also the main thread for #586.

PR #1716, opened 2026-09-17
feat(mouse): add configurable modifier-held horizontal scrolling

cjscld:
## Summary

Add an opt-in **Scroll sideways while holding a key** feature: hold a selected Shift, Option, Control or Command key while scrolling a mouse wheel to turn vertical movement into horizontal movement. It can be installed or removed independently in Features, with its toggle and modifier picker under Mouse & Trackpad → Scroll. It ships uninstalled and disabled, preserving existing behavior.

This covers the orientation action requested in #521. It follows LinearMouse's modifier-based axis switching, limited to vertical mouse-wheel events with the selected modifier alone; native horizontal/diagonal input, trackpad gestures and other modifier combinations pass through. The selected modifier is consumed so the receiving app does not also redirect or zoom the event. Vorssaint’s own windows retain their original wheel gestures, including capture-loupe and screenshot-editor zoom.

Own-window targeting caches a front-to-back window snapshot for up to 0.5 seconds, matching the per-app exception cache lifetime. Repeated wheel ticks over the same target reuse the result; moving into another target (including a front window overlapping the previous target) refreshes it. Own-window geometry, visibility, opacity, click-through behavior, level and ordering changes invalidate the snapshot. Generation checks prevent an in-flight lookup from republishing invalidated state.

The implementation reuses the existing CoreGraphics scroll taps, wheel classification and scroll-direction app exceptions. Both raw and smooth scrolling resolve feature availability through the same preferences helper. Removing inversion leaves redirection working, and removing redirection leaves inversion working. Redirected events retain precise deltas and are excluded from side-wheel shortcut matching. No runtime dependency, extra event tap, permission type or privileged helper is added. The UI adds one feature entry, one toggle and one picker; all 13 locales are provided. The explanatory caption remains visible while the toggle is off, and modifier names follow the app language. The inversion status is shown only when the inversion feature is available and at least one inversion axis is enabled; enabling horizontal redirection alone no longer displays an inversion message.

## Verification

Tested on Apple Silicon, macOS 27.0 (26A428), using a locally built, ad-hoc-signed release bundle with the Korean UI.

- Full release build completed without compiler warnings; bundled `--selftest` returned `SELFTEST OK` and strict bundle signature verification passed.
- `./build.sh --test`: **53,303 checks passed**, including **335 scroll-modifier checks**, plus preference cleanup. Coverage includes all modifiers, raw/continuous wheel precision, unrelated shortcuts, inversion combinations and all 32 installation/toggle combinations, own-window gesture preservation (capture overlays, external-window occlusion, click-through windows and secondary-display coordinates), and target-cache reuse/invalidation.
- Before the review revisions, manually confirmed on the local Mac: Option-held horizontal scrolling and release back to vertical, smoothing on/off, independent removal of either direction feature, reinstall restoring settings, and unchanged trackpad behavior.

Builds used a local compiler wrapper for 8 workers and workspace-local caches/output maps; the build script itself is unchanged. After the initial review revisions, the contributor confirmed physical-device testing of the capture-loupe/editor gestures and external-app horizontal scrolling, then confirmed the follow-up inversion-status UI fix. The subsequent target-cache revision passed the full automated suite, release build, bundled self-test and strict signature verification; it has not yet been physically retested. Deterministic lookup-count tests confirm that 100 wheel ticks within one window require one WindowServer lookup, including pointer movement inside the window and unchanged AppKit updates. Additional cases cover overlapping targets, stationary-pointer expiry, empty results, own-window state changes, disable/re-enable and invalidation during a lookup. Independent static reviews covered both feature integration and the cache; no actionable findings remained. External-window changes under a stationary pointer can take up to the 0.5-second cache lifetime to be reflected. WindowServer does not expose other apps’ click-through flags, so third-party click-through overlays remain a hit-testing limitation. Not yet verified over several days, across other Mac/mouse models or on older supported macOS versions. Non-Korean translations compile but have not been visually checked by native speakers.

## Screenshots

Initial Korean UI before review revisions: independently installable feature and its scroll settings (Option selected). The current feature name is “키를 눌러 가로 스크롤”; the caption now remains visible while disabled. The green inversion status shown in the initial screenshot is also corrected: it is hidden when only horizontal redirection is enabled.

Independent mouse scroll direction feature in the Features hub

Scroll direction toggle and Option modifier picker under Mouse and Trackpad settings

## Scope

Only orientation switching is implemented; speed changes, zoom and pinch zoom from the broader request remain outside this PR.

Refs #521


Sources/Vorssaint/Services/ScrollWheelSupport.swift
```diff
@@ -31,12 +83,87 @@ enum ScrollWheelSupport {
...
+
+        // Line writes can rederive pixel fields; restore the captured precision
+        // only after both line axes have been written.
+        event.setIntegerValueField(.scrollWheelEventDeltaAxis1, value: 0)
+        event.setIntegerValueField(.scrollWheelEventDeltaAxis2, value: line)
+        event.setDoubleValueField(.scrollWheelEventFixedPtDeltaAxis1, value: 0)
+        event.setDoubleValueField(.scrollWheelEventFixedPtDeltaAxis2, value: fixed)
+        event.setIntegerValueField(.scrollWheelEventPointDeltaAxis1, value: 0)
+        event.setIntegerValueField(.scrollWheelEventPointDeltaAxis2, value: point)
+        event.flags.remove(modifier.flag)
```
vorssaint: with option or control picked, the wheel over this app's own capture loupe and screenshot editor arrives sideways with the key stripped, so their own key plus wheel zoom stops working. can the redirect leave this app's own windows alone

Sources/Vorssaint/Core/Localization.swift
```diff
@@ -3182,6 +3188,9 @@ extension Strings {
         switcherSearchPinCaption: "S starts a search and pins the switcher open, so typing no longer produces special characters when your shortcut uses ⌥, and a search starting with Q or W no longer closes the window or quits the app by mistake.",
         invertVerticalScroll: "Invert vertical scrolling",
         invertHorizontalScroll: "Invert horizontal scrolling",
+        scrollHorizontalName: "Change mouse scroll direction",
```
vorssaint: next to the invert switches, a feature called change mouse scroll direction reads like one more flip. can the name say what it does, something like scroll sideways while holding a key, in all the languages

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -943,18 +946,37 @@ struct MouseSettings: View {
...
+                                if scrollDirectionEnabled { permissions.requestAccessibility() }
+                            }
+                    }
+                    if AppFeature.scrollHorizontal.isAvailable {
+                        Toggle(l10n.s.scrollHorizontalName, isOn: $horizontalScrollEnabled)
+                            .onChange(of: horizontalScrollEnabled) { _, _ in
+                                ScrollInverter.shared.syncWithPreferences()
+                                if scrollDirectionEnabled { permissions.requestAccessibility() }
+                            }
+                        if horizontalScrollEnabled {
```
vorssaint, line 971: the explanation only appears after the switch is on, so before that the row gives no hint of what it does. can the caption stay visible under the switch while it is off, like the trackpad note further down

Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -943,18 +946,37 @@ struct MouseSettings: View {
...
+                            Picker(l10n.s.scrollHorizontalModifierLabel, selection: $horizontalScrollModifier) {
+                                ForEach(ScrollHorizontalModifier.allCases, id: \.rawValue) { modifier in
+                                    Text(verbatim: modifier.label).tag(modifier)
```
vorssaint: the key names in this picker are fixed english, while the quit protection page shows shift, option and control in the user's language. can these follow the language too

Sources/Vorssaint/Services/ScrollWheelTarget.swift
```diff
@@ -0,0 +1,50 @@
...
+        })
+        lock.withLock { clickThroughWindowIDs = ids }
+    }
+
+    func contains(_ point: CGPoint) -> Bool {
+        if Thread.isMainThread { refresh() }
+        let ids = lock.withLock { clickThroughWindowIDs }
+        return ScrollWheelSupport.targetsOwnWindow(
+            in: WindowServerSupport.onScreenWindowInfo(), at: point,
+            ownProcessID: ownProcessID, clickThroughWindowIDs: ids)
```
vorssaint: each wheel tick with the key held looks the window under the pointer up from scratch, so a fast scroll repeats the same lookup dozens of times a second while the pointer never leaves that window. can the answer be kept while the pointer stays in the same window, the way the per app exceptions keep theirs

PR #1718, opened 2026-09-17
fix(brightness): pair the last discovery request before writing a monitor off

PathGao:
## Summary

Channel discovery sends one Get VCP request before each read. A monitor that
answers NULL until a second request arrives a few milliseconds behind the first,
which is why field implementations pair theirs, never parses a reply, so
`channelOutcome` sees accepted writes with no reply and returns `.writeOnly`:
the display is pinned to a blind DDC route, its slider shows a seeded value
instead of the monitor's own, and reads stop.

That verdict is then cached in `brightnessDDCWriteOnlyPaths`, and
`shouldProbeDDC` skips discovery for the path on every later rebuild, so it
survives replugs and restarts, and only a probe that no longer runs could clear
it.

The last of the five discovery attempts now pairs its requests; the four before
it keep the wider spacing. Verdicts cached by the old probe are dropped once, or
the fix would never reach a monitor already written off.

Pairing is what every read outside discovery already sends, so this adds one
paired attempt per display per rebuild, and only where the channel was about to
be cached as unreadable. The command interval that keeps whole commands 50ms
apart (#301) is untouched.

## Verification

MacBook Air (Mac17,3, M5), macOS 27.0 (26A428), Swift 6.4, my own build. Built-in
panel plus one external display: Mi Monitor 5K (5120x2880) over USB-C, no hub.

A standalone binary using this file's packet bytes, pauses, attempt count and
reply parsing, sending one request per attempt and then two, in two consecutive
rounds:

```
1-request attempt 1: unparsed  [6E 80 BE 00 10 00 00 64 00 50 90]
1-request attempt 2: unparsed  [6E 80 BE 00 10 00 00 64 00 50 90]
1-request attempt 3: unparsed  [6E 80 BE 00 10 00 00 64 00 50 90]
1-request attempt 4: unparsed  [6E 80 BE 00 10 00 00 64 00 50 90]
1-request attempt 5: unparsed  [6E 80 BE 00 10 00 00 64 00 50 90]
1-request result: writeOnly
2-request attempt 1: replied current=80 max=100  [6E 88 02 00 10 00 00 64 00 50 90]
```

The NULL answer fails the checksum, so it reads as no reply here too.

```sh
./build.sh                 # compiles with no warnings; see below
./build/Vorssaint --selftest
./build.sh --test
```

`./build.sh` stops after compiling, at bundle signing: my login keychain holds
three identical Developer ID certificates, so codesign calls the identity
ambiguous. That is my machine, not this change.

SELFTEST OK. TESTS OK (53,861 checks). The new assertions were checked both ways: they fail
with the probe and the recheck reverted, and pass with them in.

Not tested: one external monitor, one cable, no hub in the path, and no Intel
Mac. I did not run the change through a live app session — the probe sequence
was exercised by the standalone binary above, not by the built app driving the
display.

## Anything else

I could not find an issue for this. #1589 is the same classification path with a
different cause (a converter that accepts writes and passes nothing downstream);
this does not address it and does not conflict with a per-display software
dimming override.


Sources/Vorssaint/Services/Display/BrightnessSupport.swift
```diff
@@ -49,8 +49,14 @@ enum BrightnessSupport {
         retryAttempts + 1
     }
 
-    static func ddcProbeWriteCycles(classifyingChannel: Bool) -> Int {
-        classifyingChannel ? 1 : writeCycles
+    /// Some monitors answer NULL until a second request arrives a few
+    /// milliseconds behind the first, which is why MonitorControl and m1ddc
```
vorssaint: this names two other apps in the code. the repo describes outside work by what it does instead, so can you say something like the field implementations pair theirs and drop the names

Sources/Vorssaint/Services/Display/BrightnessSupport.swift
```diff
@@ -49,8 +49,14 @@ enum BrightnessSupport {
         retryAttempts + 1
```
vorssaint, line 50: the note right above still says discovery sends only one request before each read and keeps every request more than 50ms apart. with the last attempt now pairing, can you bring that note in line

PR #1719, opened 2026-09-17
fix(switcher): keep windows listed until the app really quits

PathGao:
## Summary

⌘Tab → Q drops the app's windows from the grid as soon as `terminate()` is sent.
The request is not the answer: an app with unsaved work stays up on its save
sheet, and a few apps refuse outright. The session then lists fewer windows than
exist, the app cannot be reached from the switcher that just hid it, and nothing
says the quit did not happen.

Quitting now works like ⌘Tab → W on a window. The app's windows stay listed as
closing: letting go of the shortcut skips them and lands on whatever takes their
place, and they leave when `NSWorkspace.didTerminateApplicationNotification`
reports the app gone. If the app is still running after about as long as a
closing window gets (0.9s), they are given back and a pending release raises
the app, save sheet included. Only a second Q to the same app is ignored; Q on
an app whose last window W is closing still asks it to quit, and the give-back
leaves W's own mark alone. The observer and the wait both carry the generation of
the session that started them, so a quit that completes or gives up after the
panel closes never edits a later session.

No new user-facing text.

## Verification

MacBook Air (Mac17,3, M5), macOS 27.0 (26A428), Swift 6.4, my own build.

```sh
./build.sh                 # compiles with no warnings; see below
./build/Vorssaint --selftest
./build.sh --test
```

`./build.sh` stops after compiling, at bundle signing: my login keychain holds
three identical Developer ID certificates, so codesign calls the identity
ambiguous. That is my machine, not this change.

SELFTEST OK. TESTS OK (53,857 checks).

Not tested: I did not drive a live switcher session against an app that refuses
to quit, so the notification and give-back paths are reviewed, not observed.
They reuse the closing-window state (`closingItemIDs`, the pending commit) as is.

## Anything else

I could not find an issue for this.


Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -1339,15 +1342,42 @@ final class AppSwitcher: ObservableObject {
...
+    /// Asks the app owning the selected window to quit (⌘Tab → Q) and keeps the
+    /// session open — mirroring the system switcher. The request is not the
+    /// answer: an app with unsaved work stays up on its own save sheet, so its
+    /// windows leave the grid when macOS reports the app gone, not before.
     private func quitSelectedApp() {
         guard windows.indices.contains(selectedIndex) else { return }
         let pid = windows[selectedIndex].pid
         guard let app = NSRunningApplication(processIdentifier: pid),
               app.bundleIdentifier != Defaults.finderBundleIdentifier else { return }
         app.terminate()
```
vorssaint: letting go of the shortcut right after q still lands on the app that was just asked to quit, so it gets raised on its way out, or nothing happens because it is already gone. w keeps a closing window listed but never raises it when you let go, and only gives that up after a short wait if the app keeps the window. can q work the same way

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -1339,15 +1342,55 @@ final class AppSwitcher: ObservableObject {
...
+    /// release, gone when macOS reports the app gone, and given back if it is
+    /// still running after about as long as a closing window gets.
     private func quitSelectedApp() {
         guard windows.indices.contains(selectedIndex) else { return }
         let pid = windows[selectedIndex].pid
         guard let app = NSRunningApplication(processIdentifier: pid),
               app.bundleIdentifier != Defaults.finderBundleIdentifier else { return }
-        app.terminate()
+        let quittingIDs = Set(sessionItems.lazy.filter { $0.pid == pid }.map(\.id))
+        guard !quittingIDs.isSubset(of: closingItemIDs), app.terminate() else { return }
```
vorssaint: when the app's only listed window is already on its way out after w, pressing q right after does nothing here, so the app keeps running. can the check only skip a quit that was already asked for, not a window that is just closing

Sources/Vorssaint/Services/Switcher/AppSwitcher.swift
```diff
@@ -1339,15 +1345,61 @@ final class AppSwitcher: ObservableObject {
...
-        app.terminate()
+        guard !quittingPIDs.contains(pid), app.terminate() else { return }
+        quittingPIDs.insert(pid)
+        // Only what this quit marked is given back; a window W is closing
+        // keeps its own mark.
+        let markedIDs = Set(sessionItems.lazy.filter { $0.pid == pid }.map(\.id))
+            .subtracting(closingItemIDs)
+        closingItemIDs.formUnion(markedIDs)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.9) { [weak self] in
+            guard let self, self.sessionActive, self.quittingPIDs.contains(pid) else { return }
```
vorssaint: this wait is not tied to the session that scheduled it, unlike the watcher above, so a later session that asks the same app to quit within a second can have its marks taken away by the old wait while its own wait gives up, leaving a window stuck as closing until the app really quits or the switcher is dismissed. can it stand down once the session that scheduled it is over, the way the watcher does

PR #1738, opened 2026-09-18
feat(audio): add hourly chime notifications and settings

amitb2050:
## Summary

Adds an opt-in Hourly Chime feature in Audio settings that plays a gentle chime at the top of every hour.

**Why it belongs in Vorssaint:**
- **What carries it:** Built strictly on standard Apple Silicon macOS primitives (`NSSound` and system run loop timers) without private APIs.
- **What does it expose the project to:** Zero external asset licensing exposure. It relies exclusively on native macOS alert sounds (`/System/Library/Sounds`), with `Glass` as the default chime and fallback to `NSSound.beep()`.
- **What does it drag in:** Zero external dependencies, zero shell scripts, and no background daemon requirements.
- **Permanent surface:** Opt-in by default (`hourlyBeepEnabled: false`), resulting in zero runtime overhead or timer activity when disabled. Includes localized strings in all 13 supported languages.

## Verification

- `./build.sh` — finished with 0 warnings
- `./build.sh --test` — passed (51,590 checks across unit and shape tests)
- Tested locally on Apple Silicon Mac, macOS Sequoia 15 / Tahoe 26.
- Verified in English (US) using both the in-app "Test Chime" button and top-of-the-hour trigger.
- **What was not tested:** Multi-day continuous run across system sleep/wake cycles; non-English locales end-to-end.

## Anything else

No issue is open for this feature.


vorssaint:
closing this one, since the mac already announces the time on its own and a chime version does not earn a permanent place in the app
the clock options in the mac's own settings can announce the time every hour, half hour or quarter hour, with the voice and volume you choose, so this would be a second way to get something the system already does
there is no request behind it, and a new feature means a hub row, a section on the general page, strings in thirteen languages and backup keys to carry from now on
the volume math is written as following a named third party app, and the code should not carry another app's name or its logic
the sound list, the fallback and the names are rebuilt here even though the app already has a picker for the mac's alert sounds, and this one shows raw file names such as glass where the mac and the snippets page say crystal
the sound plays on the main output at the main volume, which is why it needs a volume slider and a system volume reading of its own, while the app already plays alert sounds through the mac's alert route, where the alert volume, the sound effects device and the accessibility flash all apply
if people do ask for a tone instead of the spoken time, the only shape worth reopening with is a plain on and off switch that plays one of the mac's own alert sounds the way the snippet sound already does, with no volume slider of its own

Issue #1727, opened 2026-09-18
Dock Preview: hovering a running app with no windows should open an Open / Hide / Quit panel


sarat03:
### Problem or Motivation

Dock Preview only opens when the hovered app has at least one previewable window. Apps often keep running with none: I close the last window, the app stays alive in the background, and hovering its Dock icon now gives me nothing at all.

Right-clicking that same icon does give me a panel, with Hide and Quit in it. So the actions exist, they are just behind a right-click, and Dock Preview goes quiet in exactly the case where the panel would be useful.

Reproduce:
1. Open any app that survives its last window (Safari, Finder, Music).
2. Close every window with the red button. The app keeps running and keeps its Dock indicator.
3. Hover the Dock icon. Nothing appears.
4. Right-click the same icon. A menu appears with Hide and Quit.

### Proposed Solution

When the hovered app is running but has no previewable windows, open a small panel with the app actions instead of showing nothing:

- Open (activate the app, and open a new window if it has none)
- Hide
- Quit

Read it as the Dock's right-click menu on hover, in Vorssaint's panel styling. Same hover delay, same dismissal rules, same panel chrome as the window preview, just an action row in place of thumbnails.

Notes:

- The pieces already exist in the app. The Dock icon hit test and hover timing run today in `DockPreviewService`, `DockClickService.swift:386` already calls `hide()` on another app, and `AutoQuitService.swift:403 `already calls `terminate()`. The missing part is the panel for the empty case, where `DockPreviewService.swift:698` currently returns early with guard `!list.isEmpty`.
- This overlaps with #379, which asks for a New Window action in the panel. Both want the same action row, so they probably want to be built once.
- A setting to turn the empty-app panel off would suit people who read "no panel" as a useful signal that an app has no windows.
- Apps that are not running at all are out of scope here. Hover on those could stay silent.

### Feature Area

Windows and Dock

### Does another app already do this?

DockDoor

Issue #379, opened 2026-07-28
Add a "New Window" button on Dock Preview


cytsai1008:
### Problem or Motivation

I'm frustrated when I want to open a new window for an app whose existing windows all live on a different desktop.

⌘N only works when the app is frontmost, so to get a new window I first have to activate the app. But on macOS, activating an app also pulls me to whichever Desktop its windows are on — the two actions are coupled, and I only want one of them.

Current flow:

1. I'm on Desktop 2. Safari has windows, but all of them are on Desktop 1.
2. I click the Safari Dock icon — purely to make it frontmost so it captures ⌘N.
3. macOS activates Safari *and* yanks me over to Desktop 1.
4. I switch back to Desktop 2, then hit ⌘N.

So I pay a round trip through another Desktop just to give the keystroke a target.

The only existing escape hatch is right-click the Dock icon → "New Window", which works but is buried in the context menu, and takes real attention to locate when the app's menu is crowded.

### Proposed Solution

It would be great if Vorssaint could expose a "New Window" action directly in the
Dock preview, so I can get a new window on the Desktop I'm already on.

I can see two reasonable placements — happy either way, listing both:

**Option A — a tile in the preview strip**
Render it as the last item in the strip, the same size as a window thumbnail, with
a "+" glyph and a "New Window" label.

**Option B — a "+" button in the existing top-right button row**
Add it alongside the buttons already in the panel's top-right corner.

Optional: a setting to hide it, for people who want the preview to stay a pure window list.

### Feature Area

Window Switcher

### Mockups / References

Like sth like this

<img width="610" height="194" alt="Image" src="https://github.com/user-attachments/assets/b116e37d-ee6c-4a4b-8a1e-344db076438e" />

or this

<img width="610" height="194" alt="Image" src="https://github.com/user-attachments/assets/861de35f-162a-478e-beed-b35ccb281a63" />

Sorry for ugly painting

PR #1739, opened 2026-09-18
feat(dock-preview): offer an app's own actions when it has no windows

sarat03:
## Summary

Hovering the Dock icon of an app that is running with every window closed did nothing at all. `beginSession` returned early on an empty window list, so the one panel that could act on that app was the panel that refused to open. The Dock's own right-click menu still offered Hide and Quit the whole time, one gesture away.

A running app with nothing to preview now keeps its session and gets a single row of app actions in place of cards: **Open**, **Hide**, **Quit**. An app that is terminating still gets nothing, and an app with windows is untouched.

Open is not a plain activation. Activating an app with no windows only moves focus to its menu bar and leaves the screen as it was, so Open unhides the app and re-opens its bundle URL, which is what makes it put a window back, the same as clicking the Dock icon.

The row follows the `previewSize` setting upward, like every other box in the panel, but does not shrink below its base size at Small. Its labels are fixed-size text, the same as a card's title, and a smaller thumbnail setting does not make a translated "Open window" any shorter.

On direction:

- **Built on** `NSRunningApplication.hide()/terminate()/unhide()` and `NSWorkspace.openApplication(at:configuration:)`. Public AppKit that predates every macOS version this app supports. Both calls already exist in the codebase, in `DockClickService` and `AutoQuitService` respectively.
- **Drags in** nothing at runtime. No new dependency, no new permission, no new entitlement, no new helper.
- **Permanent surface**: one published property, one three-case enum, one panel view. No new setting, and no new user-facing string, since the row reuses `dockPreviewOpenWindow`, `switcherMinimizedPlacementHidden`, `panelQuit` and `switcherNoOpenWindow`, so no translation lands anywhere.
- **Why it belongs here**: it is the existing panel, in the existing hover, refusing to open in the one case where its actions are the only thing left. Dock Preview already quits apps from the × button when `dockPreviewQuitAppOnClose` is on, so the destructive half of this already shipped.

The Open action overlaps with #379, which asks for a New Window button in the window panel. Both want the same action row, and whoever takes #379 should reuse this one rather than add a second.

## Verification

MacBook Air M2 (Mac14,2), macOS 27.0 (26A428), built-in Liquid Retina 2560x1664 plus an external LEN P24q-20 at 2560x1440. Dock at the bottom, magnification on, auto-hide off, UI in en_US. Self-signed build of this branch, installed to `/Applications` rather than run from the build directory.

Exercised against Finder with every window closed, confirmed as zero on-screen windows through `CGWindowListCopyWindowInfo` rather than by eye. The action row appears on hover where nothing appeared before, and apps that do have windows still get the usual thumbnails.

```sh
./build.sh          # release, no warnings
./build.sh --test   # TESTS OK (55384 checks)
```

All three were run from a throwaway worktree holding this commit alone, so nothing else in progress locally could reach the build.

The new checks compile the real `performWindowlessAction` and `perform` bodies through `Tests/generate_sources.py` and run them against recording doubles, so they cover the actual production path: panel closes before the action runs, Open unhides and re-opens with activation, an app with no bundle on disk falls back to activation instead of doing nothing silently, Quit is a plain `terminate()` so unsaved work still gets its own dialog, and a panel that is not windowless performs nothing.

**Not tested:**

- Any preview size other than Normal. The row's growth at Large and xlarge, and its floor at Small, are covered by the checks above rather than by eye.
- A Dock on the left or right edge. The row is a fixed 360x52 clamped to the screen and is placed by the existing `panelFrame` logic, but it was only ever seen along the bottom.
- An auto-hiding Dock.
- Full-screen Spaces, and hovering while a Space transition is in flight.
- Any UI language other than English. The row reuses shipped strings, so nothing is missing, but a longer translation of "Open window" has less room in 360pt than it has in the settings window.
- Living with it for more than a day.

## Anything else

Refs #1727


vorssaint:
closing this one, since it adds a panel where staying quiet is on purpose
the panel is a preview of an app's windows, and showing nothing for an app that has none already says there is nothing to show
open, hide and quit are already on the icon itself and in its right click menu, so the row does not save a gesture, and the app can also quit apps by itself when their last window closes
it would put a new panel under every running app without windows for everyone who has the feature on, with no way to turn it off
with the current desktop filter on, an app whose windows all sit on another desktop would be shown as having no window, and open would carry the user over to that desktop, which is what the filter exists to avoid

PR #1740, opened 2026-09-18
feat(mouse): add adjustable glide coast to smooth scrolling

celltower:
## Summary

Smooth scrolling gains a coast slider under More options: it stretches the response time so the same wheel distance glides out longer.

This is built directly on the shipped 3.3.5 engine: same budget, same exponential decay in elapsed time, same CADisplayLink scheduler, same tap, same exception lists. Zero (the default) emits exactly the shipped curve (pinned by test), so upgrades change nothing for existing users. No new tap-thread work: one integer read per wheel event and a few flops per frame. Permanent surface is one key, one slider inside the existing More options group, and one label per language.

This is the small follow-up to #1040 and takes its decline feedback to heart: full wheel distance preserved (pinned at 60 and 120 Hz), same shape on any refresh rate, existing keys and meanings untouched, no merged exception lists, no new top-level controls.

## Verification

Developer build on Apple M1 Pro, macOS 26 (build 26A428), internal Retina plus external 4K display. Enabled smooth scrolling, moved coast through its range: the glide lengthens while always landing complete; back at zero the feel matches the previous build.

```sh
./build.sh                     # finishes without warnings
./build.sh --test              # TESTS OK (55381 checks) + PREFERENCE CLEANUP TESTS OK
```

What I could not test: a real 120 Hz+ display (refresh-rate independence is covered by engine pins only), and an upgrade from a shipped release with smooth scrolling already on (the new key defaults to zero, covered by registered-default pins).

## Anything else

Refs #734 (that issue carries the ongoing demand for an adjustable glide feel; this covers the glide-length part, the rest is untouched by this change)


Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -1060,6 +1064,32 @@ struct MouseSettings: View {
...
+                                    .foregroundStyle(.secondary)
+                                    .frame(width: 42, alignment: .trailing)
+                            }
+                            .padding(.top, 4)
+                            HStack {
+                                Slider(value: smoothScrollInitialSpeedBinding,
+                                       in: Double(SmoothScrollSupport.initialSpeedRange.lowerBound)
+                                           ... Double(SmoothScrollSupport.initialSpeedRange.upperBound),
+                                       step: 5) {
+                                    Text(l10n.s.smoothScrollInitialSpeedLabel)
```
vorssaint, line 1085: for one notch at the default step and response this slider leaves the start of the glide exactly as it is and only makes the end run faster, at a flat pace until it stops, and with the coast slider up it cancels the coast on a single notch. can you drop this slider and keep only the coast one

celltower: Good catch, verified on my end with the numbers: at the default step and response the eased first frame of a full notch (about 7.4px) already clears the raised floor (about 5px), so the floor only ever binds in the tail, where it flattens the landing. And with coast up, the stretched response time drops the eased frames below the floor almost from the first frame, so a single notch runs flat and the coast shape is gone. That is the opposite of what the slider promises, so it had to go.

Dropped the slider; the branch is coast-only now. Same commit updated, tests green.

vorssaint:
LGTM

Issue #1729, opened 2026-09-18
Keep Awake battery cutoff with lid closed: disablesleep 0 is applied but the Mac never sleeps, drains to 0%


NongKnot:
### Before you file

- [x] Permission checked (closed-lid rule is installed and works; Keep Awake itself runs fine)
- [x] It still happens on the latest version (3.3.5, build 86)

### Description

With **Keep going with the lid closed** ON and **Disable when battery drops below** set to 20%, the cutoff fires correctly at 20% but the Mac does not go to sleep. It keeps running with the lid closed until the battery reaches 1%, then macOS does an emergency Low Power Sleep / hibernate. Happens every time.

**Expected:** at 20% Keep Awake turns off and the Mac sleeps, since the lid is closed.
**Actual:** Keep Awake turns off, the Mac stays awake and drains to 0%.

Settings: Default Duration Indefinite, Keep Awake when Vorssaint opens ON, Allow the display to sleep ON, Pause while Mac is locked OFF, Disable when battery drops below 20%, Keep going with the lid closed ON, only when connected to power OFF.

### Steps to Reproduce

1. Enable Keep Awake with the settings above (closed-lid rule installed).
2. Unplug the Mac and close the lid.
3. Leave something playing audio or video in a browser (in my case Zen).
4. Wait for the battery to reach 20%.
5. Mac stays on with the lid closed until 1%.

### Evidence (pmset -g log and unified log, 2026-09-18)

```
08:36:53  powerd: Battery capacity change posted. Capacity:20 Source:Batt
08:37:11  powerd: Process Vorssaint.749 Released PreventUserIdleSystemSleep "Vorssaint: keep the Mac awake" age:16:48:00
08:37:11  sudo + pmset spawned by Vorssaint
08:37:11  powerd: Energy Saver Prefs have changed
08:37:11  kernel: PMRD: power event 5 (kPowerEventUserDisabledSleep)   <- disablesleep 0 was applied
08:37:11  Vorssaint: notification dropped: authorization status 0     <- the 20% alert never reached me
08:37-11:05  no Sleep entry in pmset -g log; battery 20 -> 16 -> 12 -> 8 -> 4 -> 1
11:05:01  Sleep: Entering Sleep state due to 'Low Power Sleep' Using Batt (Charge:1%)
11:15:18  Wake from Hibernate, Using AC (Charge:3%)
```

Lid had been closed since 00:41 (last `ClamshellState. Closed : 1` from powerd before the wake).

[...]

PR #1741, opened 2026-09-18
fix(keep-awake): sleep a Mac whose lid shut during a closed-lid session

NongKnot:
## Summary

With "Keep going with the lid closed" on, ending a session while the lid is already shut leaves the Mac running until the battery is empty. In my case the 20% cutoff fired at 08:37, `pmset disablesleep 0` succeeded, and the Mac kept running with the lid closed until macOS forced a Low Power Sleep at 1%, two and a half hours later. The logs are in #1729.

The cause is in the kernel. Clearing `SleepDisabled` only sets `userDisabledAllSleep = false` in `IOPMrootDomain`. The lid is evaluated on open and close events, at boot and on AC changes, and not when that flag drops. A lid that shut during the session is never looked at again. Idle sleep does not rescue it either, because audio, media playback and similar assertions from other apps are usually held.

The fix sits in `disableClamshell`, which every stop path goes through. Once the passwordless revert has finished and no session still wants the Mac awake, it asks for the sleep that a lid close would have caused: lid shut, unless an external display is attached on AC power. That mirrors the kernel's own clamshell rule, so closed-display mode on AC is left alone. `IOPMSleepSystem` only needs console-user privilege, so the sudoers rule is unchanged. The rule itself lives in `SudoersSupport` so the test list can reach it.

Callers I checked:

- **Battery cutoff, timer ending, automation stopping and the lock pause** all reach the revert with the lid possibly shut. All four now sleep when they should. The lock pause matters most, because closing the lid is what locks the screen.
- **Turning the preference off during a session** does not sleep, because the session is still active. That is the guard on `isActive`.
- **The setup race in `enableClamshell`**, where the session ended while the rule was still being prepared, goes through the same revert and is covered.
- **Quit** reverts synchronously, and the app exits before the main-queue hop runs. It is not covered. Quitting with the lid shut needs an external display, and on AC that is closed-display mode anyway.
- **`recoverIfNeeded` and the uninstaller** call `pmset` directly at launch and at uninstall, when the lid is open. They are unchanged.

This touches the same file as the draft #1433. The two changes are in different functions.

## Verification

MacBook Pro 14" (Mac15,7, M3 Pro), macOS 27.0 (26A428), built-in display only. This was my own ad-hoc build of this branch; the release build 3.3.5 was used for the original report.

```sh
./build.sh                     # finished, 0 warnings
./build.sh --test              # TESTS OK (55387 checks)
```

The six new checks go red when the external-display exception is removed, and I added that as a mutation entry. On this Mac, the lid-state read and the battery read return the right values with the lid open, on both battery and AC.

**Not tested:** the end-to-end sleep with the lid actually shut. It needs someone to close the lid on a running session and watch `pmset -g log` for a sleep entry, instead of a drain. I have not done that run, so the fix is verified in code and tests only, not on a shut lid. I have also not tested with an external display attached. If a lid-closed run on your hardware shows the Mac still awake after the session ends, that is the case this change failed to cover.

The dropped 20% notification from the same report, "notification dropped: authorization status 0", is a separate topic and not part of this change.

Refs #1729


Sources/Vorssaint/Services/KeepAwakeManager.swift
```diff
@@ -711,6 +712,31 @@ final class KeepAwakeManager: ObservableObject {
...
+    /// out (#1729). Request the sleep that closing the lid would have caused.
+    private func sleepIfLidAlreadyClosed() {
+        guard !isActive || sessionPausedForScreenLock, !clamshellActive else { return }
+        guard SudoersSupport.lidSleepIsDue(
+            lidClosed: Self.lidClosed(),
+            externalDisplay: Self.hasExternalDisplay() ?? true,
+            onBattery: SystemInfo.batterySnapshot()?.isOnBattery ?? false) else { return }
+        let rootDomain = IOPMFindPowerManagement(kIOMainPortDefault)
+        guard rootDomain != 0 else { return }
+        IOPMSleepSystem(rootDomain)
```
vorssaint: the sleep request goes out the moment the restore command returns and its result is thrown away. if the system has not taken the restored setting in by then, the request is refused and the mac keeps draining as before. can you check the result and try again shortly when it is refused

Sources/Vorssaint/Services/KeepAwakeManager.swift
```diff
@@ -711,6 +712,31 @@ final class KeepAwakeManager: ObservableObject {
...
+        IOServiceClose(rootDomain)
+    }
+
+    private static func lidClosed() -> Bool? {
```
vorssaint: the app already reads the lid this same way for the brightness feature. can you share that one read instead of keeping a second copy

Sources/Vorssaint/Services/SudoersSupport.swift
```diff
@@ -17,4 +17,12 @@ enum SudoersSupport {
     static func clamshellRule(uid: uid_t) -> String {
         "#\(uid) ALL=(root) NOPASSWD: /usr/bin/pmset disablesleep 1, /usr/bin/pmset disablesleep 0"
     }
+
+    /// Whether macOS would put the Mac to sleep if its lid closed right now:
+    /// lid shut, unless an external display is attached on AC power. A nil
+    /// lid state is a Mac without one.
+    static func lidSleepIsDue(lidClosed: Bool?, externalDisplay: Bool, onBattery: Bool) -> Bool {
```
vorssaint: this decides when a shut lid should sleep, which belongs with the other keep awake helpers rather than with the password free rule, and those helpers are covered by the same checks. can you move it next to the external display helper there

Sources/Vorssaint/Services/KeepAwakeManager.swift
NongKnot: Done in 1d1847d4. The refusal is real: `privateSleepSystem` returns `kIOReturnNotPermitted` while `userDisabledAllSleep` is still set, and `IOPMSleepSystem` passes that code back. In the original report's log the kernel cleared the flag about 3 ms after pmset exited. It now checks the result and retries every 0.5 s, up to ten times. Each attempt re-checks that no session has started and that the lid is still shut, so a new session or an opened lid stops the retries.

Sources/Vorssaint/Services/KeepAwakeManager.swift
NongKnot: Done. Keep awake now calls `BrightnessService.lidClosed()`, which is only no longer private. I left its call sites in `BrightnessService` as they were, because the display restoration fixture extracts those bodies and supplies its own `lidClosed()` double. Moving the read to `SystemInfo` would have broken that seam.

Sources/Vorssaint/Services/SudoersSupport.swift
NongKnot: Moved to `KeepAwakeAutomationSupport`, next to `hasExternalDisplay`. `SudoersSupport` is back to unchanged. The checks and the mutation entry moved with it, and removing the external display exception still turns the check red.

NongKnot:
Rebased on main and addressed the three review notes. `./build.sh` finishes with no warnings, and `--selftest` passes. In `./build.sh --test`, everything passes except `resolvedUntilDate rolls a time already past today to tomorrow`. That check comes from main and also fails on this Mac without this branch, at 02:00 in UTC+7, while CI on main passes it. It looks sensitive to the local clock or timezone rather than to this change. The lid-closed run on real hardware is still untested, as the description says.

Issue #1726, opened 2026-09-18
Add a keyboard light brightness slider


sarat03:
### Problem or Motivation

Keyboard backlight has three surfaces today and all of them are discrete:

- an on/off switch in Quick Toggles (QuickTogglesSection.swift:182)
- a Dynamic Island entry (notchKeyboardLight)
- Option-Command-minus and Option-Command-equal, which move one 1/16 step
  (#1352, closing #1180)

There is no way to set an absolute level. Picking one means either tapping
a shortcut repeatedly or opening Control Center, which is the same detour
#1180 was filed to avoid.

The menu panel already has a section whose stated purpose is brightness
sliders. BrightnessSection is documented as "one brightness slider per
adjustable display", already observes BrightnessService, and already hosts
DisplayBrightnessShortcutControls inside its options disclosure. A keyboard
light slider belongs in that section rather than in a new panel row.

### Proposed Solution

Add a keyboard light slider to BrightnessSection, alongside the display
sliders, gated on brightness.keyboardLightEnabled != nil exactly as the
existing Quick Toggles entry is. Mirror it under Settings > Quick tools >
Keyboard light, beneath the switch already at QuickToolsSettings.swift:70.

The slider reaches 0 at the bottom, so it turns the light off by itself.
The Quick Toggles switch stays exactly as it is, unchanged, for people who
only want on and off. PanelToggleRow is not touched.

Service side this is one new method on BrightnessService:

    func setKeyboardLightLevel(_ level: Float)

It mirrors setKeyboardLightEnabled without the on/off branch: same bridge
guard, same refreshKeyboardLight() on a failed write, same
showKeyboardLightNotice so the Dynamic Island readout stays in sync.
keyboardLightLevel is already a continuous Float in 0...1, so there is
nothing new to persist.

### On the open question in #838

Issue #838 lists as unverified "whether the private setter takes arbitrary
intermediate levels reliably". Reading the bridge, it does, and the app
already depends on it. The selector is
setBrightness:fadeSpeed:commit:forKeyboard: and takes a Float, not a step
index (BrightnessService.swift:2036). setKeyboardLightEnabled already
writes arbitrary values through it, because
keyboardLightOnLevel(lastNonzero:) restores whatever level the user was
last at, which is any Float rather than one of the sixteen steps. A slider
writes the same kind of value the restore path already writes.

### Relationship to #595

#595 asks how the panel's composition is decided and what earns a panel
[...]

PR #1744, opened 2026-09-18
feat(keyboard): add a backlight brightness slider

sarat03:
Refs #1726

## What was wrong

Before: the keyboard backlight could be switched on and off from Quick Toggles, shown in Dynamic Island, and stepped one 1/16 step at a time with the shortcuts added in #1352. All of those are discrete. Setting an absolute level meant opening Control Center, which is the detour #1180 was filed to avoid in the first place.

## The change

After: the panel's brightness section carries a keyboard light row beside the display sliders, and Settings > Quick tools > Keyboard light carries the same slider under the switch already there.

The slider reaches 0, so it turns the light off by itself rather than disagreeing with a switch. The existing Quick Toggles switch is unchanged.

`BrightnessSection` was chosen deliberately over a new panel row. It is already documented as one brightness slider per adjustable display, already observes `BrightnessService`, and already hosts the display brightness shortcut controls. `PanelToggleRow` is not touched, so none of the other nine quick toggles change. #595 is still open on what earns a panel row, and this is written to stay out of that question: if you would rather it were Settings only until #595 is settled, the `BrightnessSection` half can be dropped without affecting the rest.

## On the ambient sensor

A slider alone would not have held its value. Auto brightness is on by default and the ambient sensor moves the backlight on its own, so any level set by hand gets walked back on the next light change.

The row therefore carries an "Auto" control bound to the keyboard's own ambient setting, shown only where `isAmbientFeatureAvailableOnKeyboard:` reports a sensor. It is the long label "Adjust in low light" as its tooltip and VoiceOver label, abbreviated in the row for width.

The row also carries an inline switch for the light itself. This one is worth your judgement: it overlaps the Quick Toggles switch, and dragging to 0 already turns the light off and restores the previous level. The argument for keeping it is that the Quick Toggles entry is hideable, so someone who has hidden it otherwise has no one-click off. Say the word and I will drop it.

## On the unverified note in #838

#838 lists as unverified "whether the private setter takes arbitrary intermediate levels reliably". It does, and the app already depended on it before this PR. The selector is `setBrightness:fadeSpeed:commit:forKeyboard:` and takes a `Float`, not a step index, and `keyboardLightOnLevel(lastNonzero:)` already restores whatever level the user was last at rather than one of the sixteen steps.

## Implementation

`setKeyboardLightLevel(_:)` mirrors `setKeyboardLightEnabled` without the on/off branch: same bridge guard, same `refreshKeyboardLight()` on a failed write, same `showKeyboardLightNotice` so the Dynamic Island readout stays in sync. The clamp lives in `BrightnessSupport` as `sliderKeyboardLightLevel(_:)` next to its siblings, which is what makes it testable.

`setKeyboardAutoBrightness(_:)` sits on the same bridge. The three ambient selectors are resolved on demand rather than in `init`, so an OS that drops them hides the Auto control and leaves the slider working.

No new dependency, no background timer, no network access. Two new strings, in all 13 language tables.

## Verification

Tested on macOS 26.0 (Darwin 27.0.0), Apple Silicon.

```sh
./build.sh
# Passed; bundle assembled and signed, no warnings.

# SELFTEST OK

./build.sh --test
# TESTS OK (55377 checks)
# PREFERENCE CLEANUP TESTS OK

git diff --check
# Passed
```

The two new checks in `MetricsTests.swift` cover pass-through, end clamping, and rejection of NaN and infinity. I confirmed they actually fail when the clamp is broken rather than assuming they run: removing the clamp gives `TESTS FAILED (1 of 55377)` naming the slider check, and the suite was re-run green afterwards.

Live-tested against a dev build on a MacBook with an ambient keyboard sensor, reading the hardware level independently through `CoreBrightness` rather than trusting the app's own report. Slider drags, the Auto control and the Settings mirror were exercised by hand.

Not tested: a Mac whose keyboard has no ambient sensor, where the Auto control should be absent and the slider should still appear. That path is gated on `isAmbientFeatureAvailableOnKeyboard:` but I have no such machine here.


Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
```diff
@@ -139,6 +146,64 @@ struct BrightnessSection: View {
...
+                    .lineLimit(1)
+                Toggle("", isOn: Binding(
+                    get: { service.keyboardLightEnabled ?? false },
+                    set: { service.setKeyboardLightEnabled($0) }
+                ))
+                .labelsHidden()
+                .toggleStyle(.switch)
+                .controlSize(.mini)
+                .help(strings.keyboardLightCaption)
+                .accessibilityLabel(strings.keyboardLight)
```
vorssaint, line 169: the switch here repeats the quick toggles one, and the slider already turns the light off at zero, so the row carries two ways to do the same thing. can the switch go

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
```diff
@@ -139,6 +146,64 @@ struct BrightnessSection: View {
...
+                                RoundedRectangle(cornerRadius: 4, style: .continuous)
+                                    .fill(auto ? AnyShapeStyle(.tint.opacity(0.15))
+                                               : AnyShapeStyle(Color.secondary.opacity(0.12)))
+                            )
+                    }
+                    .buttonStyle(.plain)
+                    .help(strings.keyboardAutoBrightness)
+                    .accessibilityLabel(strings.keyboardAutoBrightness)
+                    .accessibilityAddTraits(auto ? [.isSelected] : [])
+                }
```
vorssaint, line 190: this flips the same low light setting people set once in system settings, and the keyboard slider in control center lives with that setting as it is. can the auto control go, here and in settings, along with the pieces only it needs

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -260,8 +263,36 @@ final class BrightnessService: ObservableObject {
...
+    func setKeyboardLightLevel(_ level: Float) {
+        guard keyboardLightEnabled != nil, let keyboardLightBridge,
+              let target = BrightnessSupport.sliderKeyboardLightLevel(level)
+        else { return }
+        guard keyboardLightBridge.setBrightness(target) else {
+            refreshKeyboardLight()
+            return
+        }
+        keyboardLightLevel = target
+        if target > 0 { lastKeyboardLightLevel = target }
```
vorssaint, line 278: when the slider is dragged down to zero, whatever value it passed last before zero becomes the level the switch brings back, so turning the light back on can leave the keyboard nearly dark. can the level from before the drag be the one that comes back

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
```diff
@@ -139,6 +146,64 @@ struct BrightnessSection: View {
...
+            }
+            Slider(value: keyboardLightBinding, in: 0...1)
+                .controlSize(.small)
+                .accessibilityLabel(strings.keyboardLight)
+        }
+    }
+
+    private var keyboardLightBinding: Binding<Double> {
+        Binding(get: { Double(service.keyboardLightLevel ?? 0) },
+                set: { service.setKeyboardLightLevel(Float($0)) })
```
vorssaint, line 204: every step of a drag writes to the keyboard on the spot, while the display sliders in the same card fold a drag into one write of the newest value. can the keyboard slider fold its drag the same way

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
sarat03: Done. The row is the label, the percentage and the slider now, and 0 on the
slider is the off. Quick toggles keeps the only switch.

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
sarat03: Done, here and in Settings. Went with it: setKeyboardAutoBrightness, the
published keyboardAutoBrightness, the three ambient selectors and their two
function types in KeyboardLightBridge, and the keyboardAutoBrightness and
keyboardAutoBrightnessShort strings in all 13 tables. The slider lives with
the system setting the way the Control Center one does.

Sources/Vorssaint/Services/Display/BrightnessService.swift
sarat03: Fixed. A drag no longer touches lastKeyboardLightLevel at all. The level is
settled once the slider has been still for half a second: where it was left
if that is above 0, otherwise the level it held before the drag started.
So dragging 0.6 down to 0 and hitting the switch brings back 0.6, not the
0.02 it passed on the way.

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
sarat03: Done, same shape as the display sliders. The published level moves on the
spot so the slider stays responsive, and the write folds into one call with
the newest value a frame later. refreshKeyboardLight() also returns early
while a write is pending, so a read landing mid drag no longer shows the
level the keyboard is still catching up to.

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -260,8 +268,68 @@ final class BrightnessService: ObservableObject {
...
+            let preDrag = self.preDragKeyboardLightLevel
+            self.preDragKeyboardLightLevel = nil
+            if let level = self.keyboardLightLevel, level > 0 {
+                self.lastKeyboardLightLevel = level
+            } else if let preDrag, preDrag > 0 {
+                self.lastKeyboardLightLevel = preDrag
+            }
+        }
+        keyboardDragEndWork = work
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5, execute: work)
```
vorssaint, line 325: a half-second pause does not mean the drag ended. while still holding the slider, moving from 60% to 2%, pausing, then continuing to zero makes the switch restore 2% again. the delayed callback can also replace the level saved by an intervening off switch. can you capture and finish the saved level using the slider's actual begin and end events, and settle that drag state before another keyboard action takes over

sarat03: Fixed. The timer is gone. Both sliders now report their own begin and end through onEditingChanged, and the level the switch restores is only settled on the real end: where the drag left it, or the level from before the drag if it ended at 0. So pausing at 2% while still holding and then going to 0 brings back 60%. The switch and the brightness keys settle any open drag before they act, so nothing lands after them to overwrite their level. Changes that don't come from a drag, like VoiceOver or arrow keys on the slider, save right away.

PR #1745, opened 2026-09-18
refactor(tests): split independent feature suites

pergioa:
## Summary
- split the monolithic metrics test body into feature-owned suites with 36 independently selectable groups
- isolate preference domains, source snapshots, async gates, recorder readiness, and injected clocks so suites remain deterministic in fresh processes
- keep the runner small, reject unknown selections, and map mutation checks to their owning suites
- integrate the latest canonical main tests, including display, notch, scrolling, mixer, recording, switcher, keep-awake, emoji, clipboard, shortcut-takeover, Command Bar, Port Manager, installer, and screenshot-arrow coverage

## Testing
- ran all 36 named suites independently
- `./build.sh --test` (`55,779` checks)
- `Tests/PreferenceCleanupTests.sh`
- `python3 Tests/mutation_checks.py` (35 regressions detected)
- `./build.sh --list-tests`
- unknown and empty suite selections exit `2`
- `git diff --check`
- `zsh -n build.sh`


Sources/Vorssaint/Services/Recorder/RecorderWriter.swift
```diff
@@ -39,7 +46,8 @@ final class RecorderWriter {
           frameRate: Int,
           capturesSystemAudio: Bool,
           capturesMicrophone: Bool,
-          pauseClock: RecorderPauseClock) {
+          pauseClock: RecorderPauseClock,
+          readinessOverride: ((RecorderCaptureEngine.Kind) -> Bool)? = nil) {
```
vorssaint, line 50: this override only exists so a test can make the writer claim it is busy, which leaves a test double inside the recording code. can you drop it, together with the readiness query that only the tests call, and let that fixture rely on the writer's real readiness

Tests/generate_sources.py
```diff
@@ -372,10 +372,12 @@ def main():
           + "var iconRowContentWidth: CGFloat { switcher.iconRowLayout.contentWidth(simpleMode: true, windowRow: false) }\n"
           + "var body: some View {\nif selectedWindow != nil {\nlet appWindows = selectedAppWindows\n"
           + "if switcher.simple {\nGroup {\n"
-          + declaration(switcher, "                ScrollViewReader { proxy in")
+           + declaration(switcher, "                ScrollViewReader { proxy in")
+               .replace("animated: true", "animated: false")
```
vorssaint, line 376: the copied strip now scrolls with its animation turned off, and the regression check that proved a reveal firing before the resize was swapped for plain deletion, so the animated reveal that ships is no longer what this test proves. can you keep the real call and the original check

Tests/RepositoryFeatureTests.swift
```diff
@@ -0,0 +1,1340 @@
...
+
+        // MARK: Result
+
+        // MARK: Every defaults suite stays inside a namespace build.sh sweeps
+        // A UserDefaults suite leaves an empty plist in ~/Library/Preferences
+        // even after `removePersistentDomain`, and cfprefsd writes that file
+        // back out around the time this process exits, so the run cannot delete
+        // it itself. `build.sh --test` clears them afterwards, by name prefix.
+        // A suite named outside those prefixes survives every run instead.
+        let testSource = ((try? FileManager.default.contentsOfDirectory(atPath: "Tests")) ?? [])
```
vorssaint, line 1073: this older copy now reads every test file glued together, so a variable name reused across files resolves to whichever file sorts first and the rest are never looked at. the new guard already covers each file on its own, so this copy can go

Tests/ScreenshotFeatureTests.swift
```diff
@@ -0,0 +1,2632 @@
...
+        suite.expect(CommandBarMenuPath.crumb(appName: "Notes", path: []) == "Notes",
+               "a command straight off the app names only the app")
+        suite.expect(CommandBarMenuPath.crumb(appName: "Notes", path: ["", "View"])
+                == "Notes \u{203A} View",
+               "an empty step leaves no dangling separator")
+        // The cleanup above is the only kind that survives exit(). A defer
+        // that removes a file here would look like housekeeping and do none.
+        let suiteSource = (try? String(contentsOfFile: "Tests/MetricsTests.swift",
+                                       encoding: .utf8)) ?? ""
+        suite.expect(!suiteSource.isEmpty, "the suite reads itself back for its own shape check")
```
vorssaint, line 934: this shape check still reads the runner file, so it no longer looks at the code it was written for. can you point it at the file it lives in

Tests/LocalizationFeatureContractTests.swift
```diff
@@ -0,0 +1,487 @@
...
+                   && !strings.keepAwakeRightClickToggleCaption.contains("—"),
+                   "\(prefix) right-click Keep Awake labels are present without em dash")
+            expectFormat(strings.homebrewConfirmInstallBodyFormat, ["@"], "\(prefix) Homebrew install format")
+            expectFormat(strings.homebrewConfirmUninstallBodyFormat, ["@"], "\(prefix) Homebrew uninstall format")
+            expectFormat(strings.homebrewConfirmUpgradeBodyFormat, ["@"], "\(prefix) Homebrew upgrade format")
+            suite.expect(!strings.homebrewUpgradeAll.isEmpty, "\(prefix) Homebrew update all title is present")
+            suite.expect(!strings.homebrewUpdateHomebrew.isEmpty, "\(prefix) Homebrew update Homebrew title is present")
+            expectFormat(strings.switcherIconRowMode, ["@"], "\(prefix) App Switcher icon-row title format")
+            suite.expect(!strings.switcherIconRowModeCaption.isEmpty, "\(prefix) App Switcher icon-row caption is present")
+            suite.expect(!strings.switcherSimpleMode.isEmpty, "\(prefix) App Switcher simple-mode title is present")
```
vorssaint, line 103: this string was removed on main, so the test build no longer compiles there. main also added a hook for the menu panel recovery tests and moved the software dimming button and the mouse settings checks in the big file since this branch was taken, and none of that is here

Sources/Vorssaint/Services/Recorder/RecorderWriter.swift
pergioa: Removed `readinessOverride` and the test-only `isReadyForMoreMediaData` query from `RecorderWriter`. The writer fixture now exercises the real `AVAssetWriterInput` readiness path through `append`. The recording suite and full test suite pass.

Tests/generate_sources.py
pergioa: Kept the production `animated: true` reveal calls in the generated strip and restored the original mutation that replaces the geometry callback with the premature width-change reveal. The switcher suite passes and the mutation is detected.

Tests/RepositoryFeatureTests.swift
pergioa: Removed the older concatenated-source namespace check from `RepositoryFeatureTests`. `PreferenceNamespaceTests` remains the single per-file guard.

Tests/LocalizationFeatureContractTests.swift
pergioa: Merged current `main`, removed the obsolete `switcherSimpleMode` assertion, added the menu-panel recovery and settings-window test hooks, and moved the software-dimming and mouse settings source checks to `EnergySettings.swift` and `MouseSettings.swift`. I also migrated the other test additions from the monolithic runner into their split suites. The full suite passes 56,924 checks.

Tests/ScreenshotFeatureTests.swift
pergioa: Updated the shape check to read `Tests/ScreenshotFeatureTests.swift`, where the check now lives.

Tests/SwitcherScrollTests.swift
```diff
@@ -91,18 +91,51 @@ enum SwitcherScrollContract {
...
+                        : SwitcherIconRowLayout.previewCardWidth
+                    let spacing = model.simple ? SwitcherIconRowLayout.simpleTitleSpacing
+                        : SwitcherIconRowLayout.spacing
+                    let padding = model.simple ? SwitcherIconRowLayout.simpleTitleScrollPadding : 0
+                    let clip = scroll.contentView.bounds
+                    let expectedWidth = model.simple
+                        ? model.iconRowLayout.contentWidth(simpleMode: true, windowRow: false)
+                            - 2 * SwitcherIconRowLayout.simpleTitlePanelPadding
+                        : model.iconRowLayout.previewContentWidth
+                    guard abs(clip.width - expectedWidth) <= 1 else { return false }
```
vorssaint, line 133: in the seven checks that failed, the strip was already at its new width and simply never scrolled, so checking the width here still leaves them uncovered. can you make this wait pass reliably with the animation on, as the fixed wait it replaces does on main

Tests/RecorderWriterTests.swift
```diff
@@ -4,48 +4,48 @@
...
 
 enum RecorderWriterTests {
-    static func run(expect: @escaping (Bool, String) -> Void) {
-        checkClock(expect: expect)
+    static func run(_ suite: TestSuite) {
+        checkClock(suite)
         let finished = DispatchSemaphore(value: 0)
         Task.detached {
             do {
-                try await check(expect: expect)
```
vorssaint, line 12: the plain run with video from the first frame and both sound tracks left alone is gone now, so the recording checks went from 210 to 188 and the simplest recording is the one no longer covered. can you put that run back as a call with no options

Tests/SwitcherScrollTests.swift
pergioa: Addressed in 76bcca32. The wait now keeps the animation enabled and pumps the run loop with `RunLoop.current.run(until:)`, matching the primitive used by the fixed wait that passes on `main`, while retaining the geometry and reveal readiness gates. The switcher suite passed 20 consecutive local runs (3,600 checks), followed by the full 56,924-check suite.

Tests/RecorderWriterTests.swift
pergioa: Addressed in c2c2a640 by restoring the no-options `check(suite)` call before the specialized writer scenarios. The recording suite now passes 211 checks (the historical 210 plus the newer bounded fixture-deadline assertion), and the full suite passes all 56,947 checks with preference cleanup.

Issue #83, opened 2026-06-22
[Feature]: Permissions/Panels/Load/IP & a question


nexttoyou:
### Problem or Motivation

1) It might be a bug: I have moved Vorssaint to my ‘Utilities’ folder because it is a utility. Unfortunately, updating the application there fails, presumably due to access permissions. The correct approach would be for Vorssaint to ask me for the relevant access permission when updating.

2) When clicking on the menu items for CPU/GPU/RAM, PWR, Network etc., it would be useful if the respective panels appeared.

3) It would be nice to see a top 10 list of the programmes consuming the most processing power when the System panel is opened.

4) It would be useful to display one’s own external IP address and its country of origin, although I realise this is only possible with an external server.

5) OK, I’m curious: what does ‘Vorssaint’ mean?

### Proposed Solution

As described above.

### Feature Area

Other

vorssaint:
Hi! Everything added, except for number 4! Can't add it right now to secure local-first premise 

Vorssaint is a name created from real cosmic ideas. “Vor” refers to Voronoi patterns, which are used to map structures in space, like galaxies and cosmic networks. “Saint” comes from sanctus, meaning sacred or protected

Together, Vorssaint represents a protected point in the cosmos. The planet logo reflects movement, orbit, and direction. And since the app lives in the macOS menu bar, it feels like a small planet sitting at the space (top of your workspace), always there when you need it

PR #1747, opened 2026-09-18
feat(monitor): show the local IP address

muskecan:
Refs #83

## Summary

The Network panel showed live throughput, per-app usage, session totals and
the speed test, but never the machine's own address. This adds a **Local IPv4**
line to the network block: the up and running, non-loopback interfaces, one
line per address and tagged with the system's own name for the connection
(`192.168.1.24 (Wi-Fi)`), refreshed from the same snapshot the panel already
observes. The text is selectable so it can be copied out.

It is an ordered panel block like every other one, hidden or shown with its
own toggle in Settings > Monitor > Network.

## Why there is no public address

The public half was dropped after review, not by accident. It is a request
that leaves the machine to a third party that sees the public address behind
it, and #83 declined exactly that to keep the local-first premise. What is
left is a pure, local read: no provider, no request, no privacy question, no
new permanent surface. Country of origin (the rest of #83) is not attempted
for the same reason.

## Against the four questions

- **What carries it.** `getifaddrs` and `getnameinfo` from libSystem, plus
  `SCNetworkInterfaceCopyAll` for the connection's own name. Nothing private,
  no helper, no new distribution surface.
- **What it exposes.** Nothing. No request, no permission, no stored address.
- **What it drags in.** Nothing. No dependency, no signing or release-process
  change.
- **Permanent surface vs. demand.** One line in an existing panel, one stored
  preference, one toggle, two strings across the thirteen locales. #83 is the
  demand signal.

Mainstream reference: iStat Menus and Stats show the local address in their
network section.

## Review follow-ups

- Cut down to the local block: the button, the provider caption and their
  strings in all thirteen languages are gone, along with the URLSession path
  and its stub tests.
- The tile now sits next to the other network ones in Settings > Monitor.
- Nothing connected: the value is left out instead of a long dash, the way the
  totals block does it.
- Tunnel, bridge and the other virtual interfaces are dropped through the same
  `MetricFormat.includeNetworkInterface` filter the rest of the network card
  uses.
- The interface is tagged with the system's own name (`Wi-Fi`, `Ethernet`),
  and the tag is left out when the system has no name for it.

## Verification

Apple Silicon, macOS 27.0 (26A428), ad-hoc signed release build, English UI.

```
./build.sh                     # finished without compiler warnings
./build.sh --test              # TESTS OK (56,821 checks), PREFERENCE CLEANUP TESTS OK
```

New coverage in the `network` suite (6 checks): an up and running IPv4
interface is included; loopback, a VPN tunnel, a bridge, an interface that is
not running and IPv6 are all excluded.

Both directions were checked: taking the interface filter out of
`includesLocalAddress` turns the `network` suite red (3 of 41), and putting it
back turns it green again.

On screen the block reads `192.168.1.61 (Wi-Fi)` on this machine, with no
virtual interface in the list.

Not tested: any other Mac or macOS version, a VPN or proxy being active, and
the panel rendered in the twelve non-English languages. The block was not
screenshotted; the assertions above are the rules, not a picture.


Sources/Vorssaint/Services/Metrics/NetworkAddressService.swift
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    func refreshLocalAddresses() {
+        let addresses = Self.readLocalAddresses()
+        guard addresses != localAddresses else { return }
+        cancel()
+        localAddresses = addresses
+    }
+
+    func lookupPublicAddress() {
```
vorssaint: the public lookup was turned down on #83 to keep the app local first, and that answer still stands. can you cut the branch down to the local block as you offered, taking the button, the provider caption and their strings in every language with it

Sources/Vorssaint/UI/Settings/MonitorPanelConfig.swift
```diff
@@ -66,6 +67,7 @@ struct MonitorPanelConfig: View {
                 Toggle(l10n.s.monitorItemNetSpeed, isOn: $netSpeed)
                 Toggle(l10n.s.networkApps, isOn: $netApps)
                 Toggle(l10n.s.monitorItemNetTotals, isOn: $netTotals)
+                Toggle(l10n.s.networkIPAddresses, isOn: $netAddresses)
```
vorssaint, line 185: the monitor page in settings draws these choices as tiles and only reaches the tiles, so this switch never shows up there and the new block can only be hidden from the panel itself. can you add a tile for it next to the other network ones

Sources/Vorssaint/UI/MenuPanel/NetworkSection.swift
```diff
@@ -364,3 +376,50 @@ struct NetworkSection: View {
...
+    var body: some View {
+        if !isVisible {
+            PanelHiddenItemRow(title: l10n.s.networkIPAddresses,
+                               systemImage: "network", isVisible: $isVisible)
+        } else {
+            VStack(alignment: .leading, spacing: 6) {
+                HStack(alignment: .top) {
+                    Text(l10n.s.networkLocalIP).foregroundStyle(.secondary)
+                    Spacer(minLength: 8)
+                    Text(service.localAddresses.isEmpty ? "—" : service.localAddresses.joined(separator: "\n"))
```
vorssaint: when nothing is connected this shows a long dash, which the rest of the app keeps out of visible text. can the line be left out in that case, the way the totals block does when it has nothing to show

Sources/Vorssaint/Services/Metrics/NetworkAddressService.swift
```diff
@@ -0,0 +1,128 @@
...
+        let address = text.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !address.utf8.contains(0) else { return nil }
+        var ipv4 = in_addr()
+        var ipv6 = in6_addr()
+        return inet_pton(AF_INET, address, &ipv4) == 1 || inet_pton(AF_INET6, address, &ipv6) == 1
+            ? address : nil
+    }
+
+    static func includesLocalAddress(flags: UInt32, family: Int32) -> Bool {
+        family == AF_INET && flags & UInt32(IFF_UP | IFF_RUNNING) == UInt32(IFF_UP | IFF_RUNNING)
```
vorssaint, line 25: tunnel, bridge and other virtual interfaces get listed here, while the rest of the network card leaves them out. can this use the same interface filter the card already uses

Sources/Vorssaint/Services/Metrics/NetworkAddressService.swift
```diff
@@ -0,0 +1,128 @@
...
+        var addresses = Set<String>()
+        var current = first
+        while let entry = current?.pointee {
+            defer { current = entry.ifa_next }
+            guard let address = entry.ifa_addr,
+                  includesLocalAddress(flags: entry.ifa_flags, family: Int32(address.pointee.sa_family)) else { continue }
+            var host = [CChar](repeating: 0, count: Int(NI_MAXHOST))
+            guard getnameinfo(address, socklen_t(address.pointee.sa_len), &host, socklen_t(host.count),
+                              nil, 0, NI_NUMERICHOST) == 0 else { continue }
+            addresses.insert("\(String(cString: host)) (\(String(cString: entry.ifa_name)))")
```
vorssaint: en0 means nothing to most people. can you name the connection the way the system does, like wi-fi or ethernet, and leave the tag out when the system has no name for it

Sources/Vorssaint/Services/Metrics/NetworkAddressService.swift
muskecan: Done — the branch is local-only now. The button, the provider caption and the four public strings are gone from all thirteen locales, along with the URLSession path and its stub tests. #83 stays declined; what is left is a plain local read.

Sources/Vorssaint/UI/Settings/MonitorPanelConfig.swift
muskecan: Rebased on main and put it in the network tiles, next to the other ones. The branch was still on the old row layout — good catch.

Sources/Vorssaint/UI/MenuPanel/NetworkSection.swift
muskecan: Gone. When nothing is connected the value is simply left out, same as the totals block.

Sources/Vorssaint/Services/Metrics/NetworkAddressService.swift
muskecan: It goes through MetricFormat.includeNetworkInterface now, so tunnel, bridge and the rest of the virtual interfaces are dropped, same filter as the card.

Sources/Vorssaint/Services/Metrics/NetworkAddressService.swift
muskecan: Tagged with the system's own name — it reads 192.168.1.61 (Wi-Fi) here — and the tag is dropped when the system has no name for the interface.

PR #1752, opened 2026-09-18
feat: Add connected devices menu bar metric with live updates and exclusion settings

kirolos-esmat:
## Summary

This pull request introduces a new **Connected Devices** menu bar metric and a comprehensive suite of exclusion settings in **Settings → Monitor**. 

It provides users with real-time visibility into the number of physically connected external peripherals (USB devices, hubs, external storage, network adapters, etc.) while allowing granular filtering to ignore unwanted device categories such as power supplies or hubs.

---

## Visuals & Screenshots

| Menu Bar Metric | Monitor Settings |
| :---: | :---: |
| Connected Devices Menu Bar Metric | Connected Devices Monitor Settings |

---

### 1. Connected Devices Menu Bar Metric
- Displays the current count of connected external devices directly in the menu bar with the custom icon and compact typography.
- Fully integrated into `MenuBarRenderer` with proper spacing, standard segment sizing, and dark/light menu bar tinting.

### 2. Exclusion Settings (Settings → Monitor)
Adds reactive toggle options allowing users to customize which device categories count towards the total:
- **Exclude power supplies and chargers**: Ignores AC adapters and USB-C power delivery bricks.
- **Exclude USB hubs and docks**: Ignores internal/external root hubs and dock controllers.
- **Exclude network and Ethernet adapters**: Ignores USB-C / Thunderbolt Ethernet interfaces.
- **Exclude external storage**: Ignores mounted external volumes and mass storage devices.

### 3. Intelligent Deduplication & Accurate Counting
- Fixed previous issue where single devices could be double-counted across both USB bus iteration and volume mount notifications.
- Normalizes and dedupes USB device trees by unique registry IDs and BSD disk identifiers, ensuring that 1 connected device accurately reflects as `1` in the menu bar.

### 4. Real-Time Reactive Live Updates
- Automatically reacts to hardware connection and disconnection events via `IOKit` (`kIOPublishNotification`, `kIOTerminatedNotification`) and `NSWorkspace` mount/unmount notifications.
- Subscribes to `UserDefaults.didChangeNotification` so toggling any exclusion preference immediately recalculates and updates the menu bar without requiring an application restart.
- Background discovery runs on a detached concurrent queue while state mutations publish cleanly on the MainActor/DispatchQueue.main to guarantee thread safety.

### 5. Full Localization
- Localized across all 13 supported languages (`en`, `zh-Hans`, `zh-Hant-HK`, `zh-Hant-TW`, `fr`, `de`, `it`, `ja`, `ko`, `ru`, `es`, `tr`), verified with localization test suites.

---

## Technical Architecture

- **`ConnectedDevicesMonitor.swift`**: Background service monitoring IOKit USB devices, power sources, and mounted disks with debounced multi-source aggregation.
- **`StatusItemController.swift`**: Subscribes to `ConnectedDevicesMonitor.shared.$count` and updates status item presentation dynamically.
- **`MonitorSettings.swift`**: Houses the SwiftUI settings toggles under the Monitor tab with indented sub-settings.
- **`Defaults.swift`**: Backed by persistent `UserDefaults` keys with standard fallback defaults.

---

## Testing & Verification

- [x] Verified building with `./build.sh`
- [x] Passed test suite (`swift test`)
- [x] Verified live connection/disconnection updates with physical USB peripherals
- [x] Verified dynamic updating when toggling exclusion filters in Settings
- [x] Tested localization and typography compliance


vorssaint:
closing this, the reading does not fit the app as it stands
a single number that mixes usb devices, mounted volumes and the charger cannot be explained from the bar, a laptop on its charger already reads 1 and a dock that also powers the mac counts twice
the sorting into categories leans on name fragments, so any product whose name happens to contain lan is treated as a network adapter
a hub with a drive plugged into it is treated as storage because the disk name search walks down through its children
while the reading is on, every preference change anywhere in the app triggers a full device scan, and a single plug or unplug runs that scan several times over
the bar item keeps listening for mount and unmount events even while the reading is off, next to two more subscriptions that only repeat what the count already publishes
clicking the number opens the disk detail, which shows disk usage rather than what was counted, and the reading hides under the disk feature in the hub although it is not about disk space
the settings piece is written against the old list layout that the tiles redesign replaced, so that part would have to be rebuilt rather than merged
a smaller take could still fit, a count of external usb devices only that samples through the same pipeline as the other readings and opens a list of what it counted when clicked

PR #1761, opened 2026-09-19
fix(cleaning-mode): wait for real mouse release before teardown

bebricoOOOOOOf:
### Summary

Reworks the Cleaning Mode mouse-release fix after #1519.

The previous attempt tried to recover a supposedly stuck button by reading global button state and posting a synthetic `mouseUp`. That could terminate a legitimate physical press early and then let the real release arrive later.

This version never manufactures input.

### What changes

- The existing Cleaning Mode HID tap also observes left/right/other mouse down/up events.
- Mouse events are always passed through unchanged; Cleaning Mode still does not lock the pointer or mouse buttons.
- A small state machine remembers only presses whose actual `mouseDown` was observed while Cleaning Mode was active.
- If user-requested deactivation happens while one of those presses is still open, teardown stays pending until the matching real `mouseUp` arrives.
- Teardown then runs on the next main-loop turn, after the release has propagated and after an AppKit control action has returned.
- Session-switch and disabled-tap fail-open paths keep their existing immediate teardown behavior because they cannot safely wait for another input event.

There is no use of `NSEvent.pressedMouseButtons`, `CGEventSource.buttonState`, or synthetic `CGEvent(mouseEventSource: ...).post(...)`.

### Why this is safer than #1519

The ownership question is answered from the event sequence itself:

`mouseDown observed by Cleaning Mode -> deactivation requested -> matching real mouseUp -> teardown`

A press that Cleaning Mode did not observe cannot be "repaired", and a physical press is never ended by a fabricated global event.

### Tests

Adds deterministic coverage for:

- deactivation requested during a held primary button;
- unrelated releases not completing teardown;
- another button being pressed while teardown is already pending;
- waiting for the last matching real release;
- normal completed clicks leaving no stale state;
- forced reset clearing the state;
- source guards preventing a return to synthetic mouse releases or global button-state snapshots.

The affected-machine hardware reproduction should still be rechecked before merging; this draft is intended to get CI and code review on the corrected mechanism first.


Sources/Vorssaint/Services/CleaningMode/CleaningModeManager.swift
```diff
@@ -107,11 +127,34 @@ final class CleaningModeManager: ObservableObject {
     }
 
     func deactivate() {
-        deactivate(restoreSuspendedFeatures: true)
+        guard isActive else { return }
+        // If a click began while the overlay was up, keep the overlay and tap
+        // alive until its real mouse-up passes through. Never manufacture a
+        // release: the physical event is the only event that completes the click.
+        guard mouseReleaseGate.requestDeactivation() else { return }
```
vorssaint, line 134: please keep a safe unlock path when a mouse release was missed while the tap was disabled. the existing rearm branch retains the new pressed-button state, so a missed secondary-button release can leave every later escape sequence and left-click unlock waiting for a release that already happened. invalidate that incomplete sequence or fail open without manufacturing input

bebricoOOOOOOf: Fixed in 2e694af / 8458427. A disabled-tap notification now invalidates the tracked mouse sequence before rearming, since any matching mouseUp may have occurred during the observation gap. If user deactivation was already pending, it fails open on the next main-loop turn after the callback instead of waiting for an event that may never arrive. No synthetic input or global button-state query is used. Added a source guard for this recovery path.

PR #1764, opened 2026-09-19
fix(auto-quit): close fullscreen windows directly

pergioa:
## Summary
- intercept verified fullscreen red-button clicks and close the targeted window directly
- preserve normal window clicks, app exceptions, and native save prompts
- document the one-click behavior and add regression coverage

## Testing
- `swift build`
- `./build.sh --test` (56,816 checks)


Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -648,11 +650,12 @@ final class AutoQuitService: ObservableObject {
     private func startCloseRequestMonitor() {
         guard closeRequestTap == nil else { return }
         let mask = CGEventMask(1 << CGEventType.leftMouseDown.rawValue)
+            | CGEventMask(1 << CGEventType.leftMouseUp.rawValue)
             | CGEventMask(1 << CGEventType.keyDown.rawValue)
         guard let tap = CGEvent.tapCreate(
             tap: .cgSessionEventTap,
             place: .tailAppendEventTap,
-            options: .listenOnly,
+            options: .defaultTap,
```
vorssaint: ordinary close shortcuts and windowed close clicks now wait for the other app to answer before reaching it, even though they do not need the fullscreen change. keep those inputs from being held up by this handling

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -692,6 +697,20 @@ final class AutoQuitService: ObservableObject {
...
+        if type == .leftMouseUp, let target = pendingFullscreenClose {
+            pendingFullscreenClose = nil
+            guard AXIsProcessTrusted(),
+                  target.buttonFrame.insetBy(dx: -4, dy: -4).contains(event.location),
+                  Self.boolAttribute(target.window, "AXFullScreen") else { return nil }
+            // Accessibility pressing the verified button bypasses macOS's
+            // first-click fullscreen exit while preserving the app's close
+            // handling, including save prompts.
+            markCloseButtonRequest(pid: target.pid)
+            AXUIElementPerformAction(target.button, kAXPressAction as CFString)
```
vorssaint: if the close request fails, the click has already been swallowed and the window stays open. keep the normal close action available when this request cannot be carried out

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: done

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: done

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -661,28 +665,60 @@ final class AutoQuitService: ObservableObject {
...
         closeRequestTap = tap
         let source = CFMachPortCreateRunLoopSource(kCFAllocatorDefault, tap, 0)
         closeRequestRunLoopSource = source
         CFRunLoopAddSource(CFRunLoopGetMain(), source, .commonModes)
         CGEvent.tapEnable(tap: tap, enable: true)
+
+        fullscreenCloseTap = fullscreenTap
+        let fullscreenSource = CFMachPortCreateRunLoopSource(kCFAllocatorDefault, fullscreenTap, 0)
+        fullscreenCloseRunLoopSource = fullscreenSource
+        CFRunLoopAddSource(CFRunLoopGetMain(), fullscreenSource, .commonModes)
```
vorssaint: the active tap receives every mouse release on the same main run loop that performs the passive mouse-down handler's synchronous accessibility queries. if a normal window is slow to answer those queries, its release waits behind them before this handler can return it, so ordinary close clicks are still held up. can you keep ordinary mouse releases independent of that accessibility work

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -716,16 +754,43 @@ final class AutoQuitService: ObservableObject {
...
+        }
+
+        guard type == .leftMouseUp, let target = pendingFullscreenClose else {
+            return Unmanaged.passUnretained(event)
+        }
+        pendingFullscreenClose = nil
+        guard AXIsProcessTrusted(),
+              target.buttonFrame.insetBy(dx: -4, dy: -4).contains(event.location),
+              Self.boolAttribute(target.window, "AXFullScreen"),
+              AXUIElementPerformAction(target.button, kAXPressAction as CFString) == .success else {
```
vorssaint: when the close handler opens a save dialog, the accessibility action can return cannotcomplete while the close is already in progress. this branch then forwards the original mouse release as though no close had started. can you distinguish that timeout from a definite failure before allowing the native click to complete

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in a3dcc1b3. The active mouse-up tap now runs on a dedicated user-interactive thread and run loop, so ordinary releases no longer queue behind the main-thread Accessibility hit test. Pending close state and tap generations are lock-protected, and a resolving mouse-down token prevents a slow query from publishing a stale target after the release.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in a3dcc1b3. The AX action result now distinguishes definite failures from `.cannotComplete`: success and the indeterminate timeout consume the release, while definite failures retain the native-click fallback. The behavior is covered by the app-management tests.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -714,18 +813,79 @@ final class AutoQuitService: ObservableObject {
...
     }
 
+    private func handleFullscreenCloseEvent(type: CGEventType, event: CGEvent) -> Unmanaged<CGEvent>? {
+        if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
+            let tap: CFMachPort? = fullscreenCloseLock.withLock {
+                guard fullscreenCloseThread === Thread.current else { return nil }
+                pendingFullscreenClose = nil
+                return fullscreenCloseTap
+            }
+            if let tap { CGEvent.tapEnable(tap: tap, enable: true) }
```
vorssaint: this and the two notes that follow were already in the earlier push and should have come with the earlier requests. this release tap holds events, and the other taps in the app that can hold events are handed back while the login session is switched away and only turned back on after a timeout while the session is on screen. left in a switched away session, a tap like this can make the account on screen wait out the tap timeout on its mouse releases, and this handler turns it straight back on. can you follow session activity here as the other services do, handing the release tap back on resign, rebuilding it on return and asking the session before turning it back on?

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -714,18 +813,79 @@ final class AutoQuitService: ObservableObject {
...
+        guard let target else { return Unmanaged.passUnretained(event) }
+        guard AXIsProcessTrusted(),
+              target.buttonFrame.insetBy(dx: -4, dy: -4).contains(event.location),
+              Self.boolAttribute(target.window, "AXFullScreen") else {
+            return Unmanaged.passUnretained(event)
+        }
+        let actionResult = AXUIElementPerformAction(target.button, kAXPressAction as CFString)
+        return AutoQuitSupport.shouldConsumeFullscreenCloseAction(actionResult)
+            ? nil
+            : Unmanaged.passUnretained(event)
```
vorssaint: the press on a fullscreen red button still reaches the app through the passive tap, and only its release is swallowed here, so the app gets half a click. an app that tracks its close button from the press can be left waiting for a release that never comes, even after this accessibility press has closed the window. after such a close, do keys and the next click still reach that app normally? the green button and dock click handling swallow the press together with the release so an app never sees half a click. if the app does wait, can you keep the two halves together, for example by holding the press on the release thread only when the window server already places it on the close button area of a fullscreen window, with no accessibility work on the press, and replaying it ahead of its release or on the first drag whenever the click is not handled here?

Sources/Vorssaint/UI/Switcher/SwitcherView.swift
```diff
@@ -331,12 +331,17 @@ struct SwitcherView: View {
...
+                            DispatchQueue.main.async {
+                                revealSelection(in: proxy, animated: true)
+                            }
                         }
                         .onChange(of: appWindows.map(\.element.id)) { _, _ in
-                            revealSelection(in: proxy, animated: true)
+                            DispatchQueue.main.async {
+                                revealSelection(in: proxy, animated: true)
+                            }
                         }
```
vorssaint: this switcher change is unrelated to closing fullscreen windows, and main has since reworked these reveals in #1772. can you move it to its own pr with the problem it still solves on main, so it can be reviewed and reverted apart from the auto quit change?

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in 5af3c0bc. Auto Quit now follows SessionActivity: it tears down both taps when this login session resigns and rebuilds them from preferences on return. Both timeout re-arm paths check the current session and Accessibility grant before enabling a tap.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in 5af3c0bc. The dedicated tap now holds mouse-down only for a WindowServer display-filling close-button candidate, with no AX work in the press callback. AX verification runs asynchronously on main; on a drag, unverified release, or definite action failure, the press is replayed ahead of the native event. A confirmed close consumes both halves. Added geometry coverage and ran the full test suite (59,994 checks).

Sources/Vorssaint/UI/Switcher/SwitcherView.swift
pergioa: Removed the unrelated switcher changes from this PR in 5af3c0bc. PR #1772 has since addressed the dropped scroll reveals on main, so there is no separate remaining fix to propose there.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -714,18 +838,152 @@ final class AutoQuitService: ObservableObject {
...
+                return Unmanaged.passUnretained(event)
+            }
+            // WindowServer geometry is the only press-time lookup. The AX
+            // verification runs on main after this callback has returned.
+            guard event.flags.intersection([.maskCommand, .maskControl, .maskAlternate, .maskShift]).isEmpty,
+                  let candidate = WindowServerWindowHitTest.candidate(at: event.location),
+                  WindowServerSupport.contains(event.location, inTrafficLightAreaOf: candidate.frame,
+                                               button: .close),
+                  Self.coversDisplay(candidate.frame, at: event.location),
+                  let down = event.copy() else { return Unmanaged.passUnretained(event) }
```
vorssaint: this press check now rests on window server geometry alone, and two common cases seem to land on the wrong side of it. on a built in display with a camera housing the system keeps a fullscreen window below the camera by default, so can its frame ever match the display bounds there? if not, the press on its red button is never held and that display keeps the old two step close. on other displays the menu bar revealed over a fullscreen window is skipped by this scan, so a press on the apple menu in that corner is held as well and the menu only opens on release or on the first drag. can the hold follow whether that display is showing a native fullscreen space, as the notch service already reads it on space changes, and skip any press where a window of another process is in front? doing the cheap checks first would also keep the window list copy off every ordinary click

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -714,18 +838,152 @@ final class AutoQuitService: ObservableObject {
...
+                                           bundleURL: app.bundleURL,
+                                           exceptions: exceptions) else { return }
+        fullscreenCloseLock.withLock {
+            guard generation == fullscreenCloseGeneration,
+                  pendingFullscreenClose?.mouseDownTimestamp == timestamp,
+                  CGEventSource.buttonState(.combinedSessionState, button: .left),
+                  let down = pendingFullscreenClose?.down else { return }
+            pendingFullscreenClose = PendingFullscreenClose(mouseDownTimestamp: timestamp,
+                                                            down: down, target: target)
+        }
```
vorssaint: the release closes the window in one click only when this answer is already published, and the target now waits behind refresh windows, which walks every window of the app before it is set. a quick click can release first and fall back to the native two step close, and with tap to click the release follows the press almost at once, so the button may already be up when it is checked here. can you publish the target right after the hit, as the previous version did, and refresh afterwards, and does the one click close still happen with tap to click?

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in f24f1a6e. The tap caches native fullscreen Space membership per display on Space/display changes, then checks the display and small close-button region before copying the window list. It no longer requires the window to fill display bounds, including on camera-housing displays. The front-to-back scan stops at any visible window over the click (including a revealed menu bar), and AX still verifies the actual button. Added coverage for camera-inset windows and occluding menus.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in f24f1a6e. The AX close target is published before refreshWindows, without requiring the physical button still to be down. If a tap-to-click release arrives first, the tap holds its copy with the press and applies the verified close when AX answers; if verification fails or takes longer than the bounded wait, it replays the native press/release pair. The full test suite passed (59,997 checks); I did not run a physical tap-to-click UI test here.

Sources/Vorssaint/Services/WindowServerSupport.swift
```diff
@@ -118,6 +118,26 @@ enum WindowServerSupport {
...
+                                         ownProcessID: pid_t) -> TrafficLightCandidate? {
+        for window in windows {
+            guard let bounds = bounds(from: window), bounds.contains(point),
+                  (window[kCGWindowAlpha as String] as? NSNumber)?.doubleValue ?? 1 > 0 else { continue }
+            guard let layer = (window[kCGWindowLayer as String] as? NSNumber)?.intValue,
+                  layer == 0, bounds.width >= 80, bounds.height >= 80,
+                  let pid = (window[kCGWindowOwnerPID as String] as? NSNumber)?.int32Value,
+                  pid != ownProcessID,
+                  let number = (window[kCGWindowNumber as String] as? NSNumber)?.uint32Value,
+                  contains(point, inTrafficLightAreaOf: bounds, button: .close) else { return nil }
```
vorssaint: this scan ends at the first visible window over the point even when that window belongs to the same app as the fullscreen window behind it, which goes further than the earlier request about a window of another process. when the pointer reaches the top of a fullscreen window the system slides its title bar and toolbar down, and if that strip is its own short window of the app in front of the fullscreen one, the scan stops there, that window fails the 80 point size check and the press is never held, so the red button keeps the native two step close. does the one click close happen with the title bar revealed this way, and can the scan pass over that window of the same app while still stopping at menus and at windows of any other process, ours included?

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -645,6 +669,23 @@ final class AutoQuitService: ObservableObject {
...
+        let separateSpaces = NSScreen.screensHaveSeparateSpaces
+        let displays = Dictionary(uniqueKeysWithValues: NSScreen.screens.compactMap {
+            screen -> (CGDirectDisplayID, CGFloat)? in
+            guard let id = (screen.deviceDescription[NSDeviceDescriptionKey("NSScreenNumber")] as? NSNumber)?
+                    .uint32Value,
+                  topology?.isFullscreen(on: id, separateSpaces: separateSpaces) == true
+            else { return nil }
+            // Fullscreen can report a zero safe area even when the built-in
+            // camera housing keeps the window's title bar below it.
+            let clearance = CGDisplayIsBuiltin(id) != 0 ? max(screen.safeAreaInsets.top, 50) : 0
```
vorssaint: on an external display this leaves no clearance, while a revealed fullscreen title bar sits below the menu bar. with a 24 point menu bar and a 52 point toolbar its red button is around 50 points from the top, past the 46 points this gate allows, and even when the window scan reaches the fullscreen window it measures the same 46 points from that window's top, which is the top edge on a display without a camera housing. so a press on most of that button is not held and windows with a toolbar keep the native two step close. this was already true in the previous push and should have come with the earlier request. can both checks leave room for the menu bar on every display, and does the one click close happen there for a window with a toolbar?

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -714,18 +864,235 @@ final class AutoQuitService: ObservableObject {
...
+        // A globally replayed pair can revisit the tap before it is torn down.
+        if event.getIntegerValueField(.eventSourceUserData) == Self.replayedClosePressMarker {
+            return Unmanaged.passUnretained(event)
+        }
+        if type == .leftMouseDown {
+            let stale = fullscreenCloseLock.withLock { () -> PendingFullscreenClose? in
+                defer { pendingFullscreenClose = nil }
+                return pendingFullscreenClose
+            }
+            if let stale { flushFullscreenClose(stale) }
```
vorssaint: when a held click still waits for accessibility with its release kept, and the next press is not one this tap holds, that press passes straight through here while the earlier pair is posted back to the session and only arrives after it, so the new press is delivered before the earlier click. can the earlier pair go ahead of the new press through the tap proxy, as the drag path does?

Sources/Vorssaint/Services/WindowServerSupport.swift
pergioa: Addressed in b9bdfb07. The scan can pass over short layer-zero strips belonging to the same PID when they align with the underlying fullscreen window, including the case where the window starts below a camera housing. It still stops at other-process windows, layered menus, and off-axis popups; AX verifies the actual fullscreen red button before closing. Added synthetic tests for those overlapping windows. I have not physically verified a revealed-titlebar click.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in b9bdfb07. The preflight now adds the measured menu-bar thickness on every display, and the fullscreen candidate scan uses the same extra vertical reach from the window top. Camera-housing clearance is separate so it does not overextend the window-relative close area. Tests cover an external-display click at 55 points and a detached toolbar above a camera-inset window; physical toolbar clicks remain unverified here.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
pergioa: Addressed in b9bdfb07. When a new mouse-down displaces a pending click, the held press and release are now both posted through that callback’s tap proxy before the new press is returned. Timeout/teardown paths still use session-level posting when no proxy exists. The general-agent review found no blockers after revisions; swift build and the focused suites pass. The full suite encountered an intermittent unrelated Notch rail assertion; the Notch suite passed on rerun.

Sources/Vorssaint/Services/AutoQuit/AutoQuitService.swift
```diff
@@ -714,18 +870,244 @@ final class AutoQuitService: ObservableObject {
...
+            replayFullscreenPress(pending.down, proxy: proxy)
+            if let proxy {
+                // Both halves precede the new down that this callback is
+                // returning; a session-level post can arrive after it.
+                release.tapPostEvent(proxy)
+            } else {
+                release.setIntegerValueField(.eventSourceUserData, value: Self.replayedClosePressMarker)
+                release.post(tap: .cgSessionEventTap)
+            }
+        } else if CGEventSource.buttonState(.combinedSessionState, button: .left) {
```
vorssaint: this was already in the previous push and should have come with the earlier requests. when the hit test rejects a held press, or its app is in the exceptions, while the button is already up but its release has not reached this tap yet, the press is dropped here and the release then reaches the app alone, so that click is lost. the release can still be on its way because a tap served by the main thread, such as the dock click, maximize or window layout ones, sits before this one once it is switched on after this feature, and it has to wait for this hit test before passing the release on. with tap to click the button is up almost at once, so in that order the presses handed back here are lost almost every time. while this tap is on every release comes through it, so can a press whose button already reads up wait for that release or the first drag and go ahead of it through the proxy, keeping this check for a tap that was switched off

PR #1766, opened 2026-09-19
feat(monitor): add external USB connected devices menu bar metric

kirolos-esmat:
## Summary

Rebuilds the closed #1752 attempt as the smaller USB-only reading suggested in the closure review. It counts external USB devices through the existing `SystemMonitor` sampling pipeline and opens the exact counted-device list from the menu bar.

- samples only IOKit USB device entries—no mounted volumes, power sources, category heuristics, or device-tree BSD lookup
- filters built-in/non-removable hardware and root controllers, then deduplicates by stable registry identity
- samples every 2 seconds while the dedicated detail is open and every 10 seconds for an enabled menu-bar metric; it does not scan while both are inactive
- uses the current metric-tile UI, its own Monitor Feature Hub entry, and a dedicated detail view with the counted USB devices
- localizes all new user-facing text across the 13 supported languages

## Verification

- `./build.sh --test` — 56,989 / 56,989 checks passed (`TESTS OK`)
- `./build.sh --dev` — clean compilation, packaging, and codesigning of `Vorssaint (Developer).app`

Tested on a physical Mac with external USB peripherals.


vorssaint:
closing this because it still mixes usb devices, disks and chargers into one count, and clicking it opens disk details instead of showing what was counted
it scans again after unrelated settings changes and keeps a connection listener running with the metric off
a smaller usb-only version using the existing monitor and showing the counted devices could still fit

PR #1770, opened 2026-09-19
feat(monitor): add external USB connected devices menu bar metric

kirolos-esmat:
## Summary

This is the USB-only rebuild suggested in the closure review of #1752. It uses the existing monitor pipeline and shows the exact devices it counts when clicked.

- enumerates only IOKit USB devices; it does not read mounted volumes, power sources, or device-category names
- excludes built-in/non-removable hardware and root controllers, then deduplicates by stable registry identity
- samples every 2 seconds while its detail is open and every 10 seconds when enabled in the menu bar; no scan runs while both are inactive
- opens a dedicated Connected Devices detail list, uses the current metric tiles, and has its own System Monitor Feature Hub entry
- localizes all new text across the 13 supported languages

## Verification

- `./build.sh --test` — 56,989 / 56,989 checks passed (`TESTS OK`)
- `./build.sh --dev` — clean compilation, packaging, and codesigning of `Vorssaint (Developer).app`

Tested on a physical Mac with external USB peripherals.


Sources/Vorssaint/Services/SystemMonitor/SystemMonitor.swift
```diff
@@ -859,6 +871,13 @@ final class SystemMonitor: ObservableObject {
                 next.fanSpeeds = self.lastFanSpeeds
             }
 
+            if plan.needConnectedDevices {
+                if take(.connectedDevices) {
```
vorssaint, line 876: please add connected devices to the timer's needed kinds. in the background with only power and usb needed at the default 2 second interval, this sample runs every 80 seconds because the timer skips over its 5 tick cadence. when usb is the only metric, the timer also wakes every 2 seconds instead of 10

Sources/Vorssaint/Core/Defaults.swift
```diff
@@ -957,7 +958,7 @@ enum Defaults {
         "gpu", "gpuTemperature",
         "memory",
         "battery", "batteryTime", "batteryTemperature", "peripheralBattery",
-        "network", "diskUsage", "diskActivity", "power", "fanSpeed",
+        "network", "diskUsage", "diskActivity", "connectedDevices", "power", "fanSpeed",
```
vorssaint: adding connected devices here also changes the registered default for the menu bar metric order, and the registered defaults check in the switcher model feature tests still expects the old string where disk activity is followed by power. that check fails on every ci job for this branch, so please update the expected order there with connected devices between disk activity and power

Tests/FeatureCatalogTests.swift
```diff
@@ -287,7 +287,7 @@ enum FeatureCatalogTests {
 
         // MARK: Features hub catalog
 
-        suite.expect(AppFeature.allCases.count == 69, "feature catalog has 69 features")
+        suite.expect(AppFeature.allCases.count == 70, "feature catalog has 70 features")
```
vorssaint: main added the notch agents feature and moved this same count from 69 to 70, so git merges this line without a conflict while the merged catalog has 71 features and the features suite fails. please update the branch with main and set this count and its message to 71

Sources/Vorssaint/Services/Metrics/USBDeviceSampler.swift
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        // Exclude root controllers and root hubs (vendor 0, product 0)
+        let vendorId = (dict[kUSBVendorID as String] as? NSNumber)?.intValue
+            ?? (dict["idVendor"] as? NSNumber)?.intValue ?? 0
+        let productId = (dict[kUSBProductID as String] as? NSNumber)?.intValue
+            ?? (dict["idProduct"] as? NSNumber)?.intValue ?? 0
+        if vendorId == 0 && productId == 0 {
+            return nil
+        }
```
vorssaint: this only drops entries with a zero vendor and product id, so external hubs still count as connected devices. a usb 3 hub enumerates as a usb 2 hub plus a usb 3 hub, so a usb c adapter with a usb 3 hub and nothing plugged into it already reads 2 in the bar, and a keyboard with a built in hub reads as two devices. please skip devices whose bdeviceclass is 9, the hub class, so hubs no longer inflate the count and the list

kirolos-esmat:
Addressed in 333c3e6c after rebasing onto current main:

- Updated the registered menu-bar metric-order expectation to include `connectedDevices`.
- Updated the feature catalog expectation to 71, including the upstream `notchAgents` addition.
- Excluded USB hubs using `bDeviceClass == 9`, with a regression test, so hubs and their USB 2/3 companion entries do not inflate the count.

The earlier timer-needed-kinds change is also present after the rebase. Full local tests completed successfully.

Issue #1575, opened 2026-09-11
Extra Brightness 'hack' breaks per app screenshot


talaviram:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

Extra brightness breaks screenshots window selection.

I would assume (didn't review code) that the Extra Brightness adds a top level overlay to bright things up that detected by the OS as a view. So I guess simple workaround would be 

### Steps to Reproduce

1. Enable "Extra Brightness"
2. Hit `Ctrl+Cmd+Shift+4` to introduce "Save picture of selected area as file"
3. Click "Spacebar" to allow selecting a specific app
4. Move mouse across opened apps

Expected:
- Without "Extra Brightness" you'll be able to select specific app

Actual:
- With "Extra Brightness" turned on, the entire laptop display acts as a single app.

Workaround:
- Turn off "Extra Brightness"

### Feature Area

Windows and Dock

### Vorssaint Version

v3.2.5

### macOS Version

macOS 27 (Golden Gate)

### Language

English

### Mac and displays

MBP M1 Pro 14" 2021

### Installation Method

Homebrew and then updates via app iirc.

Issue #802, opened 2026-08-23
Deeper external display dimming and display enhancement


ranak8811:
### Problem or Motivation

I use a basic 1080p external monitor with my MacBook Air M1. The monitor is noticeably bright even when its brightness is set to 0%, especially when using it at night.

I previously used BetterDisplay for this because it allowed me to dim the external display significantly below the monitor's native 0% brightness level. With Vorssaint, setting the external display brightness to 0% still leaves the screen too bright for comfortable nighttime use.

It would be great if Vorssaint could provide a stronger software-based dimming option for external displays, allowing the effective brightness to go below the monitor's hardware minimum.

Additionally, if technically possible, it would be very useful to have an optional display enhancement feature for external 1080p monitors, such as improving text sharpness/clarity or providing better scaling/rendering options. This would make lower-quality external displays more comfortable for reading and coding.

### Suggested features

* Allow external displays to be dimmed below their native/hardware 0% brightness using software-based dimming.
* Provide a configurable "Extra Dim" or "Night Dim" mode for external monitors.
* Allow users to control the strength of this additional dimming independently from the monitor's native brightness.
* If technically feasible, provide optional text sharpness/clarity or display rendering improvements for external 1080p monitors.

My main priority is the ability to make an external display significantly darker than its native minimum brightness, especially for comfortable nighttime use.

### Proposed Solution

It would be great if Vorssaint could provide an additional software-based dimming layer for external displays.

Ideally, users could:

* Enable an "Extra Dim" or "Software Dimming" mode when the monitor's native brightness reaches 0% but the screen is still too bright.
* Adjust the software dimming level independently from the monitor's hardware brightness.
* Have a quick toggle or shortcut to enable/disable the extra dimming, especially for nighttime use.
* Apply the extra dimming only to selected external displays when multiple monitors are connected.

As an additional enhancement, if technically feasible, it would be great to have optional display rendering improvements for external 1080p monitors, particularly for text clarity/sharpness and scaling. This could make inexpensive external displays more comfortable for coding, reading, and general text-heavy work.

### Feature Area

Energy and display

### Mockups / References

BetterDisplay is a reference for the type of software-based dimming behavior I'm looking for. It allows an external display to be dimmed significantly below its hardware brightness minimum, which is particularly useful for nighttime use.

[...]

Issue #331, opened 2026-07-21
HiDPI support for external displays


fabi19945:
### Problem or Motivation

Now that external monitor brightness control landed in the last release, it would be a great next step to add HiDPI support for external displays as well. Many external monitors (especially sub-4K panels around 1440p/1080p) don't get a proper Retina-style scaling option from macOS by default, and Vorssaint already has the menu bar presence and per-display settings infrastructure to make this a natural fit.

- macOS often hides HiDPI resolution options for displays it doesn't consider "Retina-class," even though scaling would visibly improve text and UI clarity on many external monitors.
- Since the recent update already added brightness control for external monitors, Vorssaint is clearly investing in giving users more control over their external display setup — HiDPI scaling would extend that same philosophy to resolution/clarity.
- Other tools in this space (e.g. BetterDisplay) offer this, but a lightweight, open-source option integrated into a toolkit users already have running would be very welcome.

### Proposed Solution

- Allow enabling HiDPI ("Retina") scaling for connected external displays that don't expose it natively.
- Optionally, a resolution slider (similar to flexible scaling implementations elsewhere) so users aren't limited to the 3–6 default options macOS offers.
- Per-display setting, consistent with how brightness control is currently configured per monitor.

### Feature Area

Other

fexscruta:
+1(0) – would be super nice!

Maybe you can get inspiration (or more) here: https://github.com/didriksg/Crisp

PathGao:
Triage note, part of a batch pass over the open issues. The labels were applied by a contributor, not a maintainer ruling.

Blocked on PR #376 ("Add display management controls"), which is open, mergeable against current `main` (`fca1e9d0`), and already includes both installing and removing the display override.

The blocker is merge order rather than the feature: #376 and #678 (same author) both rewrite `BrightnessService.swift`, `BrightnessSection.swift` and `SettingsView.swift`, so whichever lands first forces the other to be redone, The display thread-safety PRs that used to sit ahead of them (#762, #765, #814) have all merged, so that part of the ordering is spent.

The review focus when #376 comes up is uninstall: this writes an override plist into `/Library/Displays/` and needs elevation plus a reboot, so "does removing it restore the display cleanly" is the question that matters most. Review also noted 373 of the PR's +2025 lines go straight into `SettingsView.swift`, and that `DisplayManagementStrings.swift` bypasses `Core/Localizations/Strings+*.swift`.

PR #1773, opened 2026-09-20
feat(display): HiDPI scaling resolution manager and unified XDR overdrive brightness

TomTang197:
## Summary

Adds display HiDPI resolution management with safety recovery, alongside an extended 0%–200% unified overdrive brightness slider backed by hardware-level XDR 1600 nits EDR boost.

### 1. HiDPI Resolution Switching & Safety Recovery
- **SkyLight Dynamic Bridge (`SkyLightBridge.swift`)**: Dynamically binds CGS display mode APIs (`CGSGetCurrentDisplayMode`, `CGSGetNumberOfDisplayModes`, `CGSConfigureDisplayMode`, etc.) via `dlopen`/`dlsym` without linking private frameworks at build time.
- **VirtualDisplay Mirror Bridge (`VirtualDisplayBridge` & `VirtualDisplayService.swift`)**: Utilizes CoreGraphics `CGVirtualDisplay` SPI to provide crisp HiDPI scaling for external sub-4K monitors that lack native Retina modes.
- **15-Second Safety Recovery Watchdog (`DisplayRecoveryManager.swift` & `RecoveryHUDView.swift`)**: Displays a floating, always-on-top countdown HUD on resolution changes. Automatically reverts if not confirmed within 15 seconds, preventing black-screen lockouts.

### 2. Unified XDR Overdrive Brightness Slider (`BrightnessSection.swift` & `XDRBoostService.swift`)
- Replaces conflicting dual sliders with a single, continuous 0%–200% overdrive slider and `[XDR]` mode pill button.
- **Standard Range (0%–100%)**: Smoothly adjusts hardware panel backlight.
- **Overdrive Range (100%–200%)**: Engages full 1600 nits headroom via a 1x1 `CAMetalLayer` (`EDRPrimer`) and hardware gamma transfer table scaling (`CGSetDisplayTransferByTable`). Slider track and readout turn vibrant orange.
- Gracefully handles smooth drag transitions across the 100% boundary in both directions.

### 3. Localization
- Full translations for all newly introduced copy across all 13 supported locales (`Strings+*.swift`).
- Formatted to comply with French non-breaking punctuation spaces (`\u{00A0}`) and Turkish curly apostrophes (`’`).

---

## Validation
- `./build.sh --test`: **55,895 checks passed, 0 failures**
- `python3 Tests/mutation_checks.py`: **29 regressions detected (MUTATION CHECKS OK)**
- `./build.sh`: **Release bundle assembled and signed with 0 warnings**
- `Vorssaint --selftest`: **SELFTEST OK**

## Test Environment
- **Hardware**: Mac16,5 (MacBook Pro)
- **OS**: macOS 27.0 (Build 26A428)
- **Displays checked**: Built-in Liquid Retina XDR display + external 4K HDR display

Refs #331
Refs #1575
Refs #802


bebricoOOOOOOf:
I did a compatibility/API-contract pass on #1773 rather than dropping a large review directly into this thread.

I put the fixes into a **single-commit companion PR against the current #1773 branch**: https://github.com/TomTang197/vorssaint-utils/pull/2

The main hardening points are:

- remove the new gamma-based XDR overdrive path: `CGSetDisplayTransferByTable` documents transfer-table values in the 0.0–1.0 range, so using values up to 2.0 is outside the public CoreGraphics contract;
- correct `CGVirtualDisplay` HiDPI logical/backing dimensions to the model used by Chromium (for example 5120×2880 backing → 2560×1440 HiDPI logical);
- correct the private CGS ABI and keep SkyLight mode indexes separate from `CGDisplayMode.ioDisplayModeID`;
- snapshot an active virtual mirror **before** tearing it down, so the 15-second rollback can actually recreate the previous state;
- handle an unspecified/0 CoreGraphics refresh rate without displaying or requiring a literal “0 Hz” mode;
- add focused regression contracts and a macOS 14/15/26/27 + Intel/Apple Silicon hardware validation matrix.

I kept the resolution picker, public CoreGraphics transactions, SkyLight fallback, virtual-HiDPI fallback and recovery HUD. The existing Vorssaint Extra Brightness implementation is retained instead of replacing it with the new gamma path.

The companion PR is intentionally **Draft** because GitHub currently requires approval before its fork-origin CI can run (`action_required`); that is not a test failure. Its description contains the Apple/Chromium/Hammerspoon references and the reasoning for each change.

If this direction looks right, merging that companion PR into `feat/display-hidpi-integration` updates this PR directly; there is no competing implementation against `main`.

TomTang197:
Thanks for the thorough compatibility review and companion PR!

I've merged your hardening commit via TomTang197#2 and added the missing display services into `build.sh`'s `TEST_SOURCES` so the newly added regression contracts run cleanly in the test suite.

All 55,901 checks and `./build.sh` release self-tests are green. Ready for review!

Sources/Vorssaint/UI/MenuPanel/BrightnessSection.swift
```diff
@@ -136,6 +136,9 @@ struct BrightnessSection: View {
                     .disabled(service.isDisplayPending(display.id))
                     .accessibilityLabel(display.name)
             }
+            if display.isActive {
+                DisplayResolutionRow(displayID: display.id)
```
vorssaint, line 140: please keep the physical display reachable after enabling virtual hidpi. the next brightness rebuild skips the mirrored physical monitor and also skips the virtual source, so this entire row disappears. after keeping the new mode, there is no in-app way to turn hidpi off or adjust that monitor's brightness without quitting. keep a controllable physical row or provide an equivalent path that maps the virtual source back to its target

Sources/Vorssaint/Services/Display/DisplayRecoveryManager.swift
```diff
@@ -0,0 +1,304 @@
...
+        let snapshot = DisplayTransactionSnapshot(
+            targetDisplayID: targetDisplayID,
+            previousMode: previousMode,
+            previousCGSModeNumber: previousCGSModeNumber,
+            previousMirrorMasterID: previousMirrorMasterID,
+            previousVirtualMirrorLogicalSize: previousVirtualMirrorLogicalSize,
+            virtualDisplayCreated: virtualDisplayCreated,
+            timestamp: Date()
+        )
+        self.currentSnapshot = snapshot
```
vorssaint, line 122: please prevent a new mode change from replacing an unconfirmed snapshot. the resolution menu and hidpi toggle remain enabled while the countdown is running, so changing display a and then display b within 15 seconds cancels a's watchdog and stores only b's prior state. escape or timeout then restores b only, leaving a's unconfirmed configuration in place. serialize these transactions or retain every pending original state

Sources/Vorssaint/Services/Display/VirtualDisplayService.swift
```diff
@@ -0,0 +1,373 @@
...
+
+        let completeErr = CGCompleteDisplayConfiguration(cfg, .forSession)
+        if completeErr != .success {
+            CGCancelDisplayConfiguration(cfg)
+            instance.stop()
+            throw VirtualDisplayError.configurationFailed(completeErr)
+        }
+
+        lock.lock()
+        associatedDummies[targetDisplayID] = instance
```
vorssaint, line 354: please tear down an associated virtual display when its physical target is unplugged. the dummy is retained in this map until an explicit toggle or app termination, while the screen-change observer only refreshes mode lists. after unplugging a monitor with virtual hidpi enabled, the invisible virtual source remains as an active desktop and can keep windows off screen until the app quits

Sources/Vorssaint/Services/Display/DisplayResolutionService.swift
```diff
@@ -0,0 +1,446 @@
...
+                return false
+            }
+        }
+
+        var config: CGDisplayConfigRef?
+        let beginErr = CGBeginDisplayConfiguration(&config)
+        guard beginErr == .success, let cfg = config else {
+            Self.log.error("CGBeginDisplayConfiguration failed: \(beginErr.rawValue)")
+            if useWatchdog {
+                DisplayRecoveryManager.shared.confirm()
```
vorssaint: please roll back failures after virtual hidpi has been torn down. this method destroys the current mirror before starting the new mode transaction, but a failed begin or an unavailable or rejected mode calls confirm and clears the saved snapshot. the display loses its previous hidpi state even though the requested change failed, and neither escape nor the watchdog can restore it. the teardown error path has the same problem because the dummy is removed before the error is thrown. use the saved recovery state for all failures after that mutation

bebricoOOOOOOf:
Fixed these four points in a follow-up PR against the current #1773 branch:

https://github.com/TomTang197/vorssaint-utils/pull/3

The physical display stays in the UI when virtual HiDPI is enabled, so brightness/resolution/HiDPI-off still work.

I also made the recovery flow single-flight: while the 15s confirmation is pending, another display change won't replace the saved snapshot.

For unplugging, the virtual dummy is now dropped when its physical target disappears from the online display list.

And if anything fails after tearing down virtual HiDPI, it now rolls back from the saved state instead of calling confirm() and losing the recovery point.

Added regression coverage for all of this. CI is running now.

TomTang197:
@vorssaint Addressed all four lifecycle review points in commit `ba0048a`:
1. **Physical row retention**: Discovered rows now keep physical displays managed by Vorssaint virtual HiDPI active via `isVirtualMirrorTarget(for:)`, while still filtering out the virtual dummy. Brightness and resolution controls remain accessible.
2. **Serialized recovery transactions**: `DisplayRecoveryManager.beginAction` enforces single-flight admission (`guard !awaitingConfirmation, currentSnapshot == nil`), and resolution/HiDPI UI controls are disabled while the countdown is running.
3. **Orphan dummy teardown**: Screen parameter updates reconcile active virtual mirrors against `CGGetOnlineDisplayList` and safely tear down orphaned dummies when their physical target is unplugged.
4. **Rollback on teardown/mode failures**: Virtual mirror ownership is retained until CoreGraphics unmirroring commits, and failures after mutation invoke `rollback()` instead of clearing the snapshot via `confirm()`.

Added regression contracts covering all four conditions in `DisplayRestorationTests`. Leaving the review threads open for your verification!

Sources/Vorssaint/Services/Display/DisplayResolutionService.swift
```diff
@@ -0,0 +1,484 @@
...
+        guard CGGetActiveDisplayList(count, &displayIDs, &count) == .success else {
+            return
+        }
+
+        let activeIDs = Array(displayIDs.prefix(Int(count)))
+        var newModesPerDisplay: [CGDirectDisplayID: [DisplayResolutionMode]] = [:]
+        var newCurrentModePerDisplay: [CGDirectDisplayID: DisplayResolutionMode] = [:]
+        var newHiDPIStatusPerDisplay: [CGDirectDisplayID: HiDPIStatus] = [:]
+
+        for displayID in activeIDs {
```
vorssaint: the brightness row now stays for the mirrored monitor, but the hidpi toggle and the resolution menu live in the resolution row, which renders only when this refresh stored a current mode for that id. this loop reads only the active display list, and the new or condition in the brightness rebuild exists for a mirrored monitor that is missing from that list. in that case the slider comes back, yet there is still no in-app way to turn virtual hidpi off or change the resolution of that monitor. please also fill modes, current mode and status for online ids that are virtual mirror targets, so the row no longer depends on how the system lists a mirrored display. most of the other notes sent with this one cover code that was already in the earlier push and should have come with the first requests

Sources/Vorssaint/Services/Display/VirtualDisplayService.swift
```diff
@@ -0,0 +1,421 @@
...
+        if !profile.modes.contains(where: { $0.width == hiDPIW && $0.height == hiDPIH }) {
+            profile.modes.insert(
+                VirtualDisplayModeEntry(width: hiDPIW, height: hiDPIH, refreshRate: 60),
+                at: 0
+            )
+        }
+
+        let instance = VirtualDisplayInstance(
+            profile: profile,
+            customName: "HiDPI 源 (\(width)×\(height))"
```
vorssaint: this name is what macos shows for the virtual source in system settings and anywhere else displays are listed, so every user sees a chinese character there whatever language the app uses. the fallback name at line 164 uses the same character. please use a neutral name that needs no translation, such as the app name followed by hidpi

Sources/Vorssaint/UI/MenuPanel/RecoveryHUDView.swift
```diff
@@ -0,0 +1,359 @@
...
+    private func startMonitoring() {
+        guard localMonitor == nil, globalMonitor == nil else { return }
+
+        localMonitor = NSEvent.addLocalMonitorForEvents(matching: .keyDown) { [weak self] event in
+            guard let self = self else { return event }
+            let handled = self.handleKeyDown(keyCode: event.keyCode)
+            return handled ? nil : event
+        }
+
+        globalMonitor = NSEvent.addGlobalMonitorForEvents(matching: .keyDown) { [weak self] event in
```
vorssaint: with accessibility access granted, this monitor receives keys typed in every other app, so a return pressed in a chat or a form during the countdown keeps the new mode without the person looking at the prompt. the countdown exists for the case where the screen cannot be read, so keeping a mode should need the prompt itself. please drop the global monitor, or limit it to escape, and leave confirmation to the panel and its buttons

Sources/Vorssaint/Core/Localization.swift
```diff
@@ -1191,6 +1191,20 @@ struct Strings {
...
+    public let virtualHiDPI: String
+    public let standardResolution: String
+    public let resolutionMenuTitle: String
+    public let recoveryKeep: String
+    public let recoveryRevert: String
+    public let recoveryCountdownTitle: String
+    public let recoveryCountdownRemaining: (Int) -> String
+    public let toggleHiDPICaption: String
+    public let xdrBoostTitle: String
+    public let xdrBoostCaption: String
```
vorssaint: four of these strings have no reader in this pr. xdr boost title and caption describe the overdrive slider that was removed, and hidpi label and resolution menu title are never used. please remove them from the struct and from all thirteen catalogs, and update the pr title and description, which still announce the removed overdrive brightness

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -1515,17 +1515,22 @@ final class BrightnessService: ObservableObject {
...
+            }
             if let info,
                (info["kCGDisplayIsVirtualDevice"] as? Bool ?? false)
                 || (info["kCGDisplayIsAirPlay"] as? Bool ?? false) {
                 continue
             }
             let isBuiltIn = CGDisplayIsBuiltin(id) != 0
             let name = Self.displayName(id, info: info, screenNames: screenNames)
-            let isActive = activeTopology.contains(id)
+            let isActive = activeTopology.contains(id) || virtualMirrorTarget
```
vorssaint: this keeps the row active when the monitor is missing from the active list, but the drawable set built below and the one checked when the screens change still leave that monitor out. with the built-in panel switched off in the app and this monitor as the only other screen, turning virtual hidpi on leaves no drawable display, so the headless recovery turns the built-in panel back on by itself, and while virtual hidpi is on the built-in panel can no longer be switched off. please count the physical target of the virtual mirror as drawable in both places

Sources/Vorssaint/Services/Display/DisplayResolutionService.swift
```diff
@@ -0,0 +1,484 @@
...
+
+    @Published public private(set) var modesPerDisplay: [CGDirectDisplayID: [DisplayResolutionMode]] = [:]
+    @Published public private(set) var currentModePerDisplay: [CGDirectDisplayID: DisplayResolutionMode] = [:]
+    @Published public private(set) var hiDPIStatusPerDisplay: [CGDirectDisplayID: HiDPIStatus] = [:]
+
+    private var cancellables = Set<AnyCancellable>()
+
+    private init() {
+        refresh()
+        NotificationCenter.default.publisher(for: NSApplication.didChangeScreenParametersNotification)
```
vorssaint: edr ramps post this notification in bursts, the brightness service measured over a hundred in two seconds, and each one runs a full mode listing for every active display on the main thread and republishes all three dictionaries. this starts as soon as the brightness section shows a display, even for people who never touch resolution or hidpi. please coalesce it with a short settle window like the brightness service does, and publish only when something changed

Sources/Vorssaint/Services/Display/DisplayRecoveryManager.swift
```diff
@@ -0,0 +1,313 @@
...
+        }
+    }
+
+    private func _cleanupOnExit() {
+        Self.log.info("DisplayRecoveryManager executing cleanup on exit.")
+        if awaitingConfirmation {
+            Self.log.warning("Display configuration transaction was still pending at exit. Performing rollback.")
+            _rollback()
+        }
+        VirtualDisplayService.shared.destroyAll()
```
vorssaint: only quitting removes the virtual mirrors outside the hidpi button. turning brightness control off hides the section that holds that button, while the virtual source and the mirror stay, so the monitor keeps running through it with no control left until the app quits. the brightness service already puts its own display changes back when it stops, so please also end any pending confirmation and remove the virtual mirrors when brightness control is turned off

TomTang197:
@vorssaint Addressed all seven review points in commit `fba1191`:
1. **Resolution row for mirrored displays**: Explicitly populates modes, `currentMode`, and HiDPI status for `virtualMirrorTargetIDs()`, binding mode numbers strictly to the physical display's queried indices.
2. **Neutral virtual display names**: Renamed virtual display sources to neutral English `"Vorssaint HiDPI · ..."` across profiles and instances so no non-neutral characters appear in macOS System Settings regardless of system language.
3. **Recovery confirmation input gating**: Dropped the global keyboard monitor from `RecoveryHUDView`. Confirmation now requires interacting with the HUD prompt itself; typing in other apps will no longer inadvertently confirm a mode switch.
4. **Clean up unused strings & PR metadata**: Removed all four unreferenced strings (`hiDPILabel`, `resolutionMenuTitle`, `xdrBoostTitle`, `xdrBoostCaption`) across all 13 localization catalogs. Updated the PR title and description to accurately reflect the HiDPI manager without overdrive brightness.
5. **Drawable displays accounting**: Counted virtual mirror physical targets as drawable displays during topology checks and brightness rebuilds, preventing unintended headless recovery.
6. **Coalesced screen refresh**: Added a 0.25s debouncing settle window (`scheduleScreenRefresh`) in `DisplayResolutionService` to coalesce notification bursts during EDR ramps and avoid redundant enumeration on the main thread.
7. **Cleanup on feature toggle**: Wired `cleanupForBrightnessFeatureRemoval()` into `FeatureRuntime` when brightness control is disabled, cancelling pending transactions and tearing down active virtual mirrors.

All 55,908 checks in `./build.sh --test` are green. Left the review threads open for your verification!

Sources/Vorssaint/App/FeatureRuntime.swift
```diff
@@ -234,7 +234,12 @@ final class FeatureRuntime: ObservableObject {
             KeepAwakeManager.shared.syncWithFeatures()
             HotkeyManager.shared.syncWithPreferences()
         },
-        .brightness: { BrightnessService.shared.syncWithPreferences() },
+        .brightness: {
+            BrightnessService.shared.syncWithPreferences()
+            if !AppFeature.brightness.isAvailable {
+                DisplayRecoveryManager.shared.cleanupForBrightnessFeatureRemoval()
```
vorssaint: this cleanup only runs when brightness is uninstalled, because this binding calls it only while the feature is unavailable. the control displays toggle in settings and the turn off displays row in the command bar both switch brightness control off through brightnessservice.syncwithpreferences alone, so the service stops, the panel section disappears and the virtual source keeps mirroring the monitor with no control for it while brightness control stays off. please run this cleanup whenever the brightness service stops, whichever path stopped it

Sources/Vorssaint/Services/Display/DisplayResolutionService.swift
```diff
@@ -0,0 +1,530 @@
...
+            .sink { [weak self] _ in
+                self?.scheduleScreenRefresh()
+            }
+            .store(in: &cancellables)
+    }
+
+    /// Display mode notifications can arrive in bursts while EDR ramps. Keep
+    /// one refresh scheduled from the first event so those bursts do not run
+    /// repeated CoreGraphics/SkyLight mode enumeration on the main thread.
+    private func scheduleScreenRefresh() {
```
vorssaint: this window brings the storm down to one pass every quarter second, but each pass still lists every mode of every display and assigns all three dictionaries even when nothing changed, so publishing only on a change, the second half of the earlier request, is still open. this observer also never stops once the brightness section first appears in the panel, so it keeps listing modes after brightness control is turned off or uninstalled, which was already in the earlier pushes and should have come with the first requests. please publish only what changed, and stop or skip this refresh while brightness control is off and no virtual mirror is left

Sources/Vorssaint/Services/Display/DisplayResolutionService.swift
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    private func _performRefresh() {
+        // The screen-change notification is also our lifecycle signal for a
+        // physical target disappearing. Only mutate virtual state after a
+        // successful online-list query; a transient query failure must never
+        // be mistaken for every monitor being unplugged.
+        let onlineIDs = Self.onlineDisplayIDs()
+        if let onlineIDs {
+            let removedTargets = VirtualDisplayService.shared
```
vorssaint: this only catches a target that left the online list. when the mirror is undone while the monitor stays connected, for example by choosing extended display for it in system settings, the association stays, the row keeps reading virtual hidpi and the virtual source becomes an invisible extended desktop that windows and the pointer can move into, the same failure the unplug fix was for. please also release the virtual source here when its target no longer mirrors it, checking cgdisplaymirrorsdisplay for each associated target

Sources/Vorssaint/Services/Display/BrightnessService.swift
```diff
@@ -1507,25 +1512,25 @@ final class BrightnessService: ObservableObject {
...
-            // reading it replaces.
-            if (info?["kCGDisplayIsVirtualDevice"] as? Bool ?? false) { virtualIDs.insert(id) }
-            // A mirroring display follows its source; the source's slider is
-            // the real control.
-            guard CGDisplayMirrorsDisplay(id) == 0 else { continue }
+            // Ordinary mirror targets follow their source and do not need a
+            // separate row. A physical target managed by our virtual HiDPI
+            // mirror is different: that row is the only stable in-app handle
+            // for brightness, resolution and turning virtual HiDPI back off.
+            let virtualMirrorTarget = VirtualDisplayService.shared.isVirtualMirrorTarget(for: id)
```
vorssaint: this row is now the handle for the mirrored monitor, but brightness still picks the display to act on from the screen under the pointer, and on that monitor the screen carries the virtual source's id, since the virtual source is the display apps draw to and the physical target is only its mirror. cgmaindisplayid gives the same id when that mirror set is the main one. the virtual source is skipped a few lines below, so with virtual hidpi on, the brightness keys that follow the pointer find no route and fall through to the system, which moves its own target instead, the display brightness shortcuts in stepdisplaybrightness find nothing eligible and do nothing, and the command bar brightness action that just landed on main beeps there instead of dimming it. please map a virtual source id back to its physical target wherever brightness resolves the display to act on, and the reverse where the overlay looks for the screen of the physical id. this gap was also in the earlier pushes and belonged with the first round of requests

vorssaint:
the description still lists refs #1575 and refs #802. with the overdrive gone this pr no longer touches extra brightness, and the reporter of #802 narrowed that issue to extra dimming, so after a merge the fix status workflow would label both reports as fixed and later ask their reporters to confirm a close. please keep only refs #331, or say exactly which part of the other two this covers. the branch also conflicts with main in build.sh, where main now compiles the app in a single batched swiftc call, so the virtual display include flag has to move into that call when rebasing

TomTang197:
Addressed the review feedback in commit `3fa60dd1`:

1. **PR Description & Issue Links**: Retained only `Refs #331`.
2. **Rebase & Build**: Rebased onto latest `upstream/main` and placed `${VIRTUAL_DISPLAY_FLAGS[@]}` into the batched `swiftc` compilation command in `build.sh`.
3. **Lifecycle Cleanup**: Invoked `DisplayRecoveryManager.shared.cleanupForBrightnessFeatureRemoval()` from `BrightnessService.syncWithPreferences()` and `stop()`, releasing virtual dummies and restoring pending transactions whenever Brightness is disabled or removed. Triggered `DisplayResolutionService.shared.refresh()` when started.
4. **Change-Only Publishing & Gating**: Added `isResolutionManagementActive` to skip background mode queries and clear published state in `DisplayResolutionService` when brightness control is disabled without active dummies. Gated `@Published` updates on value inequality (`!=`) to avoid redundant SwiftUI invalidations.
5. **Disassociation Teardown**: Replaced missing-only checks with `reconcileVirtualMirrors` and `disassociatedTargetIDs`, tearing down virtual dummies when the physical target goes offline or when the user un-mirrors the display in macOS System Settings (`CGDisplayMirrorsDisplay(targetID) != virtualID`).
6. **Bidirectional Display ID Routing**: Added `resolvePhysicalTarget(for:)` across brightness hotkeys, pointer-following keys (`functionKeyKeyTap`, `mediaKeyTap`), and Command Bar actions; mapped physical target IDs back to `virtualDisplayID(for:)` in `BrightnessOSD.show` to position the overlay on the active `NSScreen`.
7. **Regression Tests**: Added `disassociatedTargetIDs` test and source-shape contracts in `Tests/DisplayRestorationTests.swift`. Verified release build completes with 0 warnings and runtime self-test outputs `SELFTEST OK`.

PR #1789, opened 2026-09-20
refactor(popover): update popover behavior and animation handling

archieamas11:
## Summary

Replace the menu panel’s custom fade animations with the native macOS `NSPopover` transition and transient behavior.

This makes the menu panel feel more natural and interactive while reducing custom animation and lifecycle code.

## Preview / Recording
https://github.com/user-attachments/assets/651dff65-ab00-4948-85d8-a3b8727eca19

## Why This Change Was Made

The menu panel previously used a simple fade animation when opening and closing. This change uses the standard macOS popover animation instead. The native transition gives the panel a more lively and familiar interaction while allowing AppKit to handle the behavior consistently.

It also avoids working against the platform by replacing native popover behavior with a custom animation. Users who prefer simpler animations can enable **Reduce motion** in macOS Accessibility settings.

## What Changed

- Uses the native `NSPopover` transition for opening and closing.
- Sets the popover behavior to `.transient`, allowing it to close when the user interacts outside the panel.
- Removes the custom fade animation code and related timing constants.
- Temporarily disables popover animations during internal repositioning so the panel moves instantly without visual bouncing.
- Detects the beginning of the close process through `popoverWillClose`.
- Uses `popover.performClose(nil)` to let AppKit handle the closing transition.

## What This Improves

- **More native interaction:** The panel follows the standard macOS popover behavior and animation. [`NSPopover` supports configurable behavior and animation through its `behavior` and `animates` properties.](https://developer.apple.com/documentation/appkit/nspopover)
- **More responsive feel:** The transition provides clearer visual feedback than a simple alpha fade.
- **Less custom logic:** Removes approximately 40 lines of manual opacity, timing, and animation coordination code.
- **Easier maintenance:** AppKit now manages the popover transition instead of the application tracking animation state manually.
- **Better platform consistency:** The menu panel behaves more like other native macOS popovers.

## Verification

Tested locally on macOS 27 m1 pro:

```sh
./build.sh
./build.sh --test
```

Results:

- Build completed successfully without warnings.
- All 58,660 checks passed across all test suites.


Sources/Vorssaint/App/AppDelegate.swift
```diff
@@ -1153,6 +1120,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
         popoverIsClosing || !PanelInteractionState.shared.preventsPopoverDismissal
     }
 
+    func popoverWillClose(_ notification: Notification) {
+        popoverIsClosing = true
```
vorssaint, line 1133: setting this flag for every close makes anchorAfterForeignClose treat system closes as app-requested, disabling recovery after a fresh click inside the panel. please track app-requested closes separately and cover willClose followed by didClose in the recovery tests

Sources/Vorssaint/App/AppDelegate.swift
```diff
@@ -404,13 +402,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate, NSW
...
 
     private func setUpPopover() {
-        // Application-defined (not .transient) so the panel stays open while the
-        // user works in our own Settings window and sees changes live. Click
-        // monitors below dismiss it when it would block that same Settings window.
-        popover.behavior = .applicationDefined
-        // We animate the underlying popover window ourselves so applicationDefined
-        // dismissal, right-click menus and live Settings previews stay predictable.
-        popover.animates = false
+        popover.behavior = .transient
```
vorssaint: transient dismissal bypasses the existing policy that keeps the panel open beside settings for live previews. when no dismissal protection is active, interacting with settings can close the panel even without overlap. please preserve this workflow while changing the animation and cover both overlapping and side-by-side windows

archieamas11: Thanks for catching this! I didn't realize transient mode was closing the panel when clicking into Settings. I've switched the behavior back to `applicationDefined` so the panel stays open beside Settings for live previews as intended, while keeping the smoother popover animation intact.

vorssaint:
LGTM

PR #1616, opened 2026-09-13
feat(window-layout): add vertical quarters and thirds

samueljim:
## Summary
- Adds full-width **vertical quarters** so four windows can stack on a tall/portrait monitor: Top 1/4, Upper center 1/4, Lower center 1/4, Bottom 1/4
- Adds **vertical thirds** (and top/bottom 2/3) alongside the existing horizontal thirds
- Surfaces them in the panel grid, Settings shortcuts, Command Bar, and Radial Menu bindings (shortcuts start unassigned, same as other newer layouts)

## Test plan
- [ ] Open Window layout in the menu panel and confirm Quarters + vertical Thirds appear
- [ ] On a portrait (or tall) display, place four windows into the four quarter strips
- [ ] Place three windows into Top / Middle / Bottom 1/3
- [ ] Assign shortcuts in Settings → Window layout and verify they apply
- [ ] Confirm existing Top/Bottom halves and horizontal thirds are unchanged
- [x] `./build.sh --test-suite=core`
- [x] `./build.sh --test-suite=localization`

PR #1792, opened 2026-09-20
feat(window-layout): add quarter rows and quarter columns

DocEmmetBrown:
Rebase and continuation of #1616 by @samueljim, onto current `main`, with the
four review comments addressed and the quarters extended to both axes.

## What this adds

Window Layout can now cut the screen into quarters two ways.

- **Quarter columns** — four full-height bands side by side, so four windows sit
  across an ultra-wide monitor. These work like the existing left/center/right
  thirds, one column narrower.
- **Quarter rows** — four full-width bands stacked top to bottom, for a tall or
  portrait display. This is #1616's original contribution.

Vertical thirds (top / middle / bottom) and top / bottom two-thirds come along
too, from the original PR.

Every new placement starts with no shortcut, so turning Window Layout on never
claims extra system-wide combinations unasked.

## Review comments from #1616

1. **Shared icons.** Every new row and column now has its own symbol. The
   full-width rows use portrait rectangles, so a stacked band never looks like a
   placement that splits the width; `top 1/3`, `top 2/3` and `top 1/4` are now
   distinct, and the middle third no longer matches the centre third. Top and
   bottom half also picked up the half-height icons they should always have had.
   All 13 new placements have unique symbols, each checked to resolve at runtime.
2. **"Second top 1/4" / "Third top 1/4".** Now *upper middle 1/4* and *lower
   middle 1/4*, so the label says where the strip lands instead of counting from
   an edge. Applied in every language.
3. **"Quarters" colliding with "Corners".** There is no "Quarters" section any
   more: the placements are grouped as **Quarter rows** and **Quarter columns**,
   which says what each one is and keeps clear of the corner placements.
4. **Fraction word order.** Portuguese, Spanish, French and Italian lead with the
   fraction for the new strips (`1/3 topo`, `1/4 esquerda`), matching the
   left/right thirds those languages already ship.

## Notes for the reviewer

`centerTwoThirds` landed on `main` while #1616 was open and had taken shortcut id
56, which the original branch also used for `topQuarter`. Since 56 is already
registered for released builds, `topQuarter` takes the next free id instead of
moving an existing shortcut.

The Window Layout settings form also moved its placement rows behind a small
`actionSection` helper: with both quarter families present the form otherwise
exceeded SwiftUI's ten-view `ViewBuilder` limit.

Tests follow the split from `cf1e7736`, so the window layout coverage lives in
`Tests/WindowLayoutFeatureTests.swift` rather than the old monolithic file.

## Testing

- `./build.sh` — clean
- `./build.sh --test` — `window-layout` 208 checks, `localization` 5729 checks,
  all suites green
- Tried by hand from `./build.sh --dev --install` on a wide display: the four
  columns line up left to right, and the rows still stack as before.

The commits keep @samueljim as the author of the original feature.

Closes #1616


vorssaint:
LGTM

PR #1795, opened 2026-09-20
fix(switcher): press past the window instead of posting no location

integral-llc:
## Summary

On macOS 27, picking a window in the App Switcher resizes it. About 0.2 s after every Cmd+Tab the window's top-left corner slides to the top-left of the display while the bottom-right corner stays put, so a window you placed by hand ends up stretched into the screen corner. It happens on every switch, to any app.

Cause is the press/release pair `SpaceWindowBridge.frontWindow` posts to make the window key. The press is aimed at (-1, -1) so it can't click content, but one point outside the frame is the top-left corner's resize area, and apps here answer it as a corner grab. Nothing else in that path writes size or position, and the motion matches a corner drag exactly: origin goes to the display corner, the opposite corner is pinned, a window that can't get wider (System Settings) only grows upward.

The press now lands far past the bottom-right corner, at (300000, 300000), and the release is gone. This is what AltTab ended up with for the same bug (lwouis/alt-tab-macos@ec30bb1): near the frame is a grab region, no location gets read by some apps as (0, 0) and clicks their top-left control, and a point that gets clamped from the positive side can only fall on the bottom-right. Without a release nothing can be activated wherever an app decides the point is. They measured that the press alone makes the window key. Here the picked window still takes the keyboard with only the press, though the AX focus pass runs right after it so I can't split the two.

My first push used the no-location form. That was wrong, thanks for catching it.

`frontWindow` has three callers, the switcher's normal activation, `SpaceHop` and `focusPlacedWindow` for Dock Preview drops. The change is in the shared function so all three get it.

## Verification

MacBook Pro M2 Max (Mac14,6), lid closed, macOS 27.0 (26A428). Three external displays side by side at 3008x1692 (Pro Display XDR as main, two LG 4K), separate Spaces on. My own build, stable signing identity, app in English.

I recorded window frames with a small CGWindowList poller while switching between Safari, Chrome and System Settings on the left display. Before, every switch was followed by this, no modifier and no mouse button down:

```
45.75 FRONT  Google Chrome -> Safari
45.95 RESIZE Safari (-2226,237 1644x1411) -> (-2298,218 1716x1430)
...
46.14 RESIZE Safari (-2969,40 2387x1608) -> (-3008,30 2426x1618)
```

With the current push, tested by hand on that setup:

- ordinary Cmd+Tab between windows I'd placed myself: nothing resizes, the picked window comes forward and takes the keyboard
- picking a window that sits on another Space: it travels there and the window is focused

Not tested: fullscreen targets, and the Dock Preview drop. Fullscreen is the one I'd look at first since that's where AltTab saw the stray clicks, of the apps they named I only have Telegram here, not tried against this yet. Nothing older than macOS 27 either, and no notch display in the mix.

```sh
./build.sh --test              # TESTS OK (58725 checks)
```

`SwitcherActivationTests` now expects one press at that point and no release. It pins what gets sent, it can't show that an app doesn't act on it, that part is only the manual checks above.

## Anything else

Couldn't find an issue for this, searched switcher + resize. Probably because it only shows on macOS 27.


Sources/Vorssaint/Services/Switcher/SpaceWindowBridge.swift
```diff
@@ -259,12 +261,11 @@ enum SpaceWindowBridge {
         let userGenerated: UInt32 = 0x200
         guard setFrontProcess(&psn, windowID, userGenerated) == .success else { return false }
         var targetID = windowID
-        var clickPoint = CGPoint(x: -1, y: -1)
         var record = [UInt8](repeating: 0, count: 0x100)
         record[0x04] = 0xf8 // declared record length
         record[0x3a] = 0x10
         withUnsafeBytes(of: &targetID) { record.replaceSubrange(0x3c..<0x3c + $0.count, with: $0) }
-        withUnsafeBytes(of: &clickPoint) { record.replaceSubrange(0x20..<0x20 + $0.count, with: $0) }
+        record.replaceSubrange(0x20..<0x30, with: repeatElement(0xff, count: 0x10))
```
vorssaint: alttab [removed this sequence](https://github.com/lwouis/alt-tab-macos/commit/ec30bb13084e68cb3cde32ec415fdba1fd92876e) after fullscreen apps treated nan as a top-left click. the new assertion only checks the bytes sent, not whether content gets clicked. please address that regression and verify focus without clicks for fullscreen windows, space changes and dock preview drops

integral-llc: you're right on this one. I remembered alttab using the 0xff form and I didn't check that they had dropped it since.

pushed 5e75ff3, it's basically what they landed on in ec30bb1. The press goes to (300000, 300000), which is clear of every grab region, and if an app clamps the point it's gonna fall on the bottom-right and not on (0,0). There is no release anymore, so even when an app pulls the point back onto content nothing can actually be activated. I have updated the description too, the alttab line in it was wrong.

On the test, agreed, it only pins what we send. I don't see a way to assert from the fixture that no app acted on it, so that part stays manual.

What I have run by hand, macOS 27.0 (26A428) with three externals and separate Spaces: normal cmd+tab between windows I placed myself, no resize and the right window gets the keyboard. And then a window sitting on another Space, it travels there and it's focused.

Not run yet: fullscreen targets and the dock preview drop. Of the apps alttab named I only have telegram here, no figma or wezterm, and I haven't run it against this yet. I can try telegram plus fullscreen chrome/vscode/cursor, or if you have figma handy that would be the faster check. Happy to leave this as draft until fullscreen is covered.

integral-llc: saw #1844 go in and rebased on top of it. one thing though. ec30bb1 is the commit where alttab dropped the 0xff form, and the comment in it says why: some apps sanitize NaN back to (0, 0) and press whatever sits top-left. figma's home button, telegram's sidebar, wezterm's first pane, their #5381. #1844 shipped that exact form.

this is now just the follow-up then. far point at (300000, 300000), no release. kept your byte check in SwitcherActivationTests, it asserts the far point and a lone press, and it fails if I put the 0xff loop back. 59740 checks ok.

honestly I only have telegram here of the apps they list, haven't tried it against this yet, so the fullscreen case is still on their word not mine.

Sources/Vorssaint/Services/Switcher/SpaceWindowBridge.swift
```diff
@@ -256,11 +256,14 @@ enum SpaceWindowBridge {
...
-    /// resize border instead, and the repeated focus pass then finished a
-    /// resize that dragged the window's top-left corner to the screen's own.
+    /// The follow-up press makes the window key without clicking any of its
+    /// content, and both halves of it matter. It lands far past the
+    /// bottom-right corner: a point just outside the frame is the resize
+    /// border, which macOS 27 answers by dragging that corner to the screen's,
+    /// while no location at all (NaN) is sanitized by some apps back to (0, 0),
+    /// their top-left control (Figma's Home button, Telegram's sidebar; AltTab
+    /// ec30bb1). It is never released: a press alone makes the window key, and
+    /// without a release nothing can be clicked wherever an app puts the point.
```
vorssaint: the project keeps third party app names out of code, and this comment now names two apps plus another switcher and its commit. please go back to generic wording like the previous push, where some apps read no location as their top left corner and a control usually sits there. the last line also says more than the press can promise, since controls that act on the press itself can still react without a release

Sources/Vorssaint/Services/Switcher/SpaceWindowBridge.swift
```diff
@@ -275,14 +278,10 @@ enum SpaceWindowBridge {
...
-        // click cannot hit content, a title bar or the resize border.
-        for offset in 0x20..<0x30 { record[offset] = 0xff }
-        record[0x08] = 0x01 // left mouse down…
-        let down = postEventRecord(&psn, &record)
-        record[0x08] = 0x02 // …then up: the pair makes the window key
-        let up = postEventRecord(&psn, &record)
-        return down == .success && up == .success
+        var pressPoint = CGPoint(x: 300_000, y: 300_000)
+        withUnsafeBytes(of: &pressPoint) { record.replaceSubrange(0x20..<0x20 + $0.count, with: $0) }
+        record[0x08] = 0x01 // left mouse down, never released
```
vorssaint: the press is never released now, so apps that track the button from events may treat it as still held. for the pending fullscreen and dock preview drop checks, telegram in fullscreen with a chat open and a drop from the dock preview are enough, since the drop sends two or three of these presses in a row and often leaves the pointer right on the window's top left corner. in each case, does the window take the keyboard with a focused text field still focused, with nothing at its corners reacting, and does moving the pointer afterwards stay a plain move instead of a drag or selection?

Sources/Vorssaint/Services/Switcher/SpaceWindowBridge.swift
integral-llc: fair, I went back to the generic wording in 2baf5ae, so no app names and no other switcher in there anymore. and honestly the last line was overselling it. now it just says the press alone makes the window key and a release would finish a click wherever the app puts the point, it doesn't claim anything about the press itself being harmless. cleaned up the commit message the same way. also rebased on beta.4, so the changelog line moved under Unreleased

Issue #1793, opened 2026-09-20
App Switcher: select apps by scrolling while holding ⌘


veged:
## Problem

Scrolling does not change the selected app in the open Vorssaint App Switcher. This prevents the familiar macOS interaction: open ⌘Tab, keep holding ⌘, scroll to the desired app, then release ⌘ to activate it.

The comparison with the native switcher is based on the reporter's observation. The missing scroll-navigation handler was also confirmed by inspecting current `main` at `b84919d43e7fe39b882a6fde2bc5680bfffbed75`. No independent hardware gesture test has been performed for this report yet.

## Steps to reproduce

1. Enable Vorssaint's App Switcher and open several apps.
2. Press ⌘Tab and keep holding ⌘ so the switcher stays open.
3. Try to select another app by scrolling with a mouse wheel or trackpad.
4. Release ⌘.

**Expected:** scrolling moves the selection through the apps, and releasing ⌘ activates the selected app.

**Actual:** there is no handler that turns scrolling into a selection change. An overflowing content area may scroll, but that does not implement navigation between apps.

## Code evidence

- The [switcher's event mask](https://github.com/vorssaint/vorssaint-utils/blob/b84919d43e7fe39b882a6fde2bc5680bfffbed75/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift#L393-L398) includes keyboard and mouse-button events, but not `CGEventType.scrollWheel`.
- The [event handler](https://github.com/vorssaint/vorssaint-utils/blob/b84919d43e7fe39b882a6fde2bc5680bfffbed75/Sources/Vorssaint/Services/Switcher/AppSwitcher.swift#L684-L740) also has no scroll-wheel handling.
- The [card grid's scroll view](https://github.com/vorssaint/vorssaint-utils/blob/b84919d43e7fe39b882a6fde2bc5680bfffbed75/Sources/Vorssaint/UI/Switcher/SwitcherView.swift#L151-L189) reveals the already selected card when `selectedIndex` changes. It does not turn scrolling into a selection change.

## Proposed solution

Allow scrolling to change the selection within an already open switcher, reusing the existing app/window navigation and modifier-release confirmation. Cover both mouse wheels and continuous scrolling from a trackpad or Magic Mouse, so momentum or smooth scrolling does not race through the entire list.

[...]

PR #1799, opened 2026-09-20
fix(switcher): select apps with the scroll wheel or trackpad

veged:
## Summary

Scrolling while the ⌘Tab switcher is open now changes the selection. Releasing the modifier activates the selected app through the existing confirmation path. Navigation reuses the same app grouping and window-selection logic as keyboard navigation, and stops at the ends of the list.

Mouse-wheel events advance one item regardless of acceleration. Continuous trackpad scrolling accumulates to a movement threshold. Both vertical and horizontal scrolling are supported. Momentum events and remaining frames from the built-in smooth scroller do not change the selection. Accumulated movement resets between gestures and switcher sessions. Moving cards beneath a stationary pointer should not override scroll navigation.

While the switcher is open, Smooth scrolling, Scroll direction, and side-wheel shortcut mappings yield the original events to it. This adds no settings, permissions, dependencies, or user-facing strings. Existing SwiftUI scroll views continue to reveal the selected item.

The expected interaction comes from the reporter's experience with the native macOS switcher. Exact direction and sensitivity parity has not yet been verified.

## Verification

MacBook Pro (Mac15,7), Apple M3 Pro, macOS 26.6.2 (25G83), Swift 6.3.3, macOS 26 SDK. Local build based on `main` at `b84919d43e7fe39b882a6fde2bc5680bfffbed75`.

- `./build.sh` passed with no compiler warnings. `actool` could not compile the adaptive icon catalog, so packaging used the existing `AppIcon.icns` fallback. The app was signed ad hoc.
- All 2,308 checks in `pointer-input`, `switcher-model`, and `switcher` passed during the full `./build.sh --test` run. The full run executed 58,749 checks, with three failures outside the changed area.
- The two Homebrew popularity-formatting failures at `RepositoryFeatureTests.swift:456–457` also occur on unchanged `main`: the current locale produces `1,2K` and `1,2M`, while the tests expect `1.2K` and `1.2M`. The baseline full run executed 58,731 checks and failed only these two.
- The history-virtualization check at `NotchCompactTests.swift:168` failed in the first full run. A separate rerun of `notch` on the final patch passed all 22,633 checks. It also passed on unchanged `main`. The first full run overlapped app compilation, so load may have contributed, but the cause of the intermittent failure has not been established.
- `git diff --check` passed. Independent review identified the side-wheel mapping conflict, which was fixed and passed a focused recheck.
- After the full build and test run, only comments and test diagnostic messages were translated to English. All 2,308 checks in the three affected suites passed again on the final files.

For the regression check, the actual navigation helper and its tests were compiled separately. Disabling scroll-to-selection conversion makes seven checks fail.

Not yet verified on hardware: mouse and trackpad interaction, direction and sensitivity with natural scrolling, Smooth scrolling and Scroll direction combinations, assigned side-wheel shortcuts, edge hover in an overflowing row, multiple displays and Spaces, and releasing ⌘ after scrolling. This is intended as a draft PR until those checks are complete.

### Review follow-up (2026-09-22)

The navigation helper and event-producing tests now use `CGScrollPhase`. Core Graphics reports `ended` as 4 and `cancelled` as 8, unlike `NSEvent.Phase`. Regression tests cover both terminal phases with zero and nonzero deltas, verify that they do not move the selection, and confirm that accumulated movement is cleared before subsequent scrolling.

The new regression tests failed three checks on the previous helper. After the fix, all 2,319 checks passed with `./build.sh --test-suite=switcher-model --test-suite=pointer-input --test-suite=switcher`. The production helper is compiled by that test build. The full app build and full test suite were not rerun for this focused follow-up.

Refs #1793


Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -5,6 +5,43 @@ import AppKit
...
+struct SwitcherScrollNavigation {
+    static let gestureStep: Double = 30
+    private var accumulated: Double = 0
+    private var lastTimestamp: CGEventTimestamp?
+
+    mutating func selectionDelta(for event: CGEvent) -> Int {
+        guard event.getIntegerValueField(.eventSourceUserData) != ScrollWheelSupport.syntheticTag else {
+            return 0
+        }
+        let phase = NSEvent.Phase(rawValue: UInt(event.getIntegerValueField(.scrollWheelEventScrollPhase)))
```
vorssaint: core graphics uses 4 for the ended phase, while the appkit value checked here is 8. a synthetic event with the core graphics ended value advances selection instead of clearing the remainder. please use core graphics phase values in both this helper and its tests

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -5,6 +5,43 @@ import AppKit
...
+        let horizontal = event.getDoubleValueField(continuous ? .scrollWheelEventPointDeltaAxis2 : .scrollWheelEventDeltaAxis2)
+        let delta = abs(horizontal) > abs(vertical) ? horizontal : vertical
+        guard delta.isFinite, delta != 0 else { return 0 }
+        if !continuous {
+            accumulated = 0
+            return delta < 0 ? 1 : -1
+        }
+        if accumulated * delta < 0 { accumulated = 0 }
+        accumulated += delta
+        guard abs(accumulated) >= Self.gestureStep else { return 0 }
```
vorssaint: this and the scroll inverter note were already in the earlier push and should have come with the phase request. for a discrete event this reads only the whole line field, but high resolution wheels report fractions of a line in the fixed point field while the whole line field reads zero, as the smooth scroll ticks helper notes. on those mice a slow tick returns zero here while the switcher still swallows it, so slow scrolling does nothing while the panel is open. please fall back to the fixed point delta when the whole line field is zero and add up those fractions until they reach one line before moving one app. mice that report the wheel as continuous events without a phase also go through the 30 point trackpad step, and a 250 ms pause clears the remainder, so does one slow notch on such a mouse still move one app?

Sources/Vorssaint/Services/ScrollInverter.swift
```diff
@@ -153,6 +153,9 @@ final class ScrollInverter: ObservableObject {
             return Unmanaged.passUnretained(event)
         }
         guard type == .scrollWheel else { return Unmanaged.passUnretained(event) }
+        if AppSwitcher.shared.scrollNavigationActive {
+            return Unmanaged.passUnretained(event)
+        }
```
vorssaint: with scroll direction reversing the mouse wheel, this hands the switcher the unreversed event, and smooth scrolling now yields the raw tick before its own flip as well. for anyone who reversed only the wheel, the wheel then moves the selection the opposite way from how it scrolls content everywhere else, unlike the trackpad. please let the scroll direction flip reach the switcher, for example by dropping this early return, since smooth scrolling already hands this tap the raw tick during the session

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
veged: Fixed in b3251c2a. Both the navigation helper and the event-producing tests now use CGScrollPhase. Regression tests cover ended (4) and cancelled (8), with zero and nonzero deltas, and verify that terminal events neither move the selection nor leave an accumulated remainder. These checks also pass on the current head, e3fd50b4.

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
veged: Fixed in e3fd50b4. Wheel events now fall back to the fixed-point delta when the whole-line field is zero, accumulate to one line, and retain the fractional remainder across pauses. Phaseless continuous mouse events use the shared wheel classification and the same one-line threshold, with a point-delta fallback for point-only events. A total of one line therefore advances one app even when samples are more than 250 ms apart. Trackpad events retain their separate threshold and gesture reset behavior. Regression tests cover both axes, discrete and continuous wheels, slow accumulation, remainder carry, direction reversal, and inversion. All 2,701 checks in the four affected suites pass. Physical-device behavior remains unverified.

Sources/Vorssaint/Services/ScrollInverter.swift
veged: Fixed in e3fd50b4 by removing the early return from ScrollInverter. Smooth scrolling yields the raw event, then Scroll direction applies the existing inversion settings and app exceptions before the switcher consumes it. Synthetic glide frames remain excluded. The direction helper also now preserves point and fixed-point deltas when the whole-line field is zero, so inversion does not erase high-resolution wheel input. Tests cover both axes and signs, plus fractional navigation with inversion enabled and disabled. All 2,701 checks in the four affected suites pass.

Issue #1791, opened 2026-09-20
Add visible Back and Forward buttons to Settings


samanyudas:
### Problem or Motivation

When moving between Settings pages, I sometimes lose track of the pages I previously visited. I noticed this while exploring Vorssaint and testing for bugs.

I want to revisit those pages without finding each one again in the sidebar.

### Proposed Solution

Add visible Back and Forward buttons that navigate through previously visited Settings pages.

Disable each button when there is no page available in that direction.

### Feature Area

Other

### Does another app already do this?

macOS System Settings

### Mockups / References

<img width="332" height="46" alt="Image" src="https://github.com/user-attachments/assets/5577854f-c4f9-4b78-851b-f9c283371b0b" />

Something like this.

PR #1815, opened 2026-09-21
feat(settings): add back and forward toolbar buttons

samanyudas:
Settings already supports navigation history through the Go menu, keyboard shortcuts, and mouse side buttons, but offers no visible Back and Forward controls.
Add native toolbar buttons that reuse the existing router and localized labels, disable when no visible destination is available, and use the same feature visibility check as the current page fallback.

Validation:
- Developer build passed.
- Settings and features suites passed: 1,003 checks.
- Packaged Developer app: SELFTEST OK.
- Live UI on macOS 27.0: both buttons initially disabled; visiting About enabled Back; Back returned to General and enabled Forward; Forward restored About and disabled itself.
- Visually inspected the toolbar and ran git diff --check.

Older macOS versions and other languages were not exercised live.

Refs #1791


vorssaint:
LGTM

Issue #1441, opened 2026-09-06
Global Direct Triggers for Command Bar Shortcuts, UI Width Customization, and Currency Conversion


SukeshSiva:
### Problem or Motivation

1. Direct Global Hotkeys for App/Folder Shortcuts:

- Current Behavior: The Command Bar allows assigning shortcuts (e.g., Super + F for Finder), but using them requires a multi-step workflow: you have to manually open the Command Bar interface first (via Cmd + Space) and then press the shortcut.
- Pain Point: This defeats the purpose of global shortcuts, which should launch items instantly from anywhere in macOS without requiring an extra UI layer.
- Proposed Solution: Add a setting or toggle allowing these shortcuts to be triggered globally and immediately.

2. Command Bar Width Customization
-Proposed Solution: Add a UI/appearance setting to resize or adjust the width of the Command Bar search window to better fit longer paths, app names, or personal layout preferences.

3. Bonus / Nice-to-Have: Inline Currency Conversion
Proposed Solution: Integrate quick inline currency conversion directly into the search bar (e.g., typing 50 USD to EUR) to make it an even more powerful day-to-day utility launcher.

### Feature Area

Tools

### Does another app already do this?

Raycast

### Mockups / References

Global Shortcuts vs. Aliases: Raycast executes global shortcuts by default for instant system-wide triggering, while also providing an alias option that behaves more similarly to the current Vosaint Command Bar integration.

UI Width: Raycast doesn't include a width adjustment setting, but its default width is already wide enough for most typical use cases.

References: The attached a Screenshot  that demonstrates the proposed inline currency conversion layout alongside a reference for ideal window width, modeled after Raycast's proportions.

Thankyou!

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/fab22220-954b-4cd1-9934-e592e3fa1813" />

Issue #977, opened 2026-08-25
Command Bar Coverage: 38 open issues the bar already answers, or would with one row


PathGao:
I went through all 203 open issues asking one question of each: **can the command bar answer this?** Baseline is `main` @ `ecc1c1c7`, the `[Unreleased]` block above `3.3.3-beta.2`. Every line anchor below was read on that commit.

Checked 2026-08-25T14:40Z. Companion writeups: #836 Bug Summary, #837 PR Summary, #838 Enhancement Summary.

38 issues land in one of three buckets. The other 165 are unrelated to the bar.

---

## The one thing worth knowing first

**Every catalog row can be bound to its own global shortcut**, up to 20 of them — `CommandBarRowShortcuts.swift:16` (`limit = 20`), resolved per row at `CommandBarService.swift:426`. Three of the five "already possible" issues below are only requests for a dedicated hotkey on something the bar already runs. That is a documentation gap, not a feature gap, and it will keep generating issues until the setting is discoverable.

The second thing: **quick toggles are not mirrored into the bar automatically.** Each one is hand-written in the catalog (`CommandBarCatalog.swift:546-594`). Any new toggle needs its catalog row written by hand or it exists on one surface only.

---

## A — already possible today (5)

These need a reply, not code. The reporter did not know the row exists.

| # | Request | What already ships |
|---|---|---|
| **613** | Master volume slider | `action.volume` takes a typed 0-100 and calls `AppVolumeMixer.setSystemOutputVolume` — `CommandBarCatalog.swift:497` (`numericRange: 0...100` at `:501`), mute at `:516`. The premise of the issue, that only per-app volume exists, is wrong. |
| **754** | Emoji picker triggered by a colon | The emoji source searches by name and types the character at the caret: rows at `CommandBarCatalog.swift:1158`, insertion via `typeAtCursor` at `:1171`. Per-app blacklisting is moot — the bar is opened deliberately. |
| **708** | `Copy text from screen` shares a key with screenshot | They are already two rows — `:289` screenshot, `:332` screen OCR — and either can take its own global key through row shortcuts. |
| **788** | Separate shortcuts for the capture tools | Same mechanism. Four rows: `:289`, `:309`, `:332`, `:342`. |
[...]

Issue #576, opened 2026-08-10
Support direct execution of Shell scripts (.sh / .command) instead of opening them in text editors


zhizhi2213:
### Problem or Motivation

I am currently using Vorssaint as my daily launcher. When I search for a shell script (e.g., test.sh or test.command) and press Enter, Vorssaint relies on the macOS default action, which
  opens the file in a text editor (like VS Code or TextEdit). To actually execute a script via Vorssaint without popping up an editor, I am forced to wrap my scripts into macOS .app bundles
  using osacompile.

### Proposed Solution

Like other productivity launchers (e.g., Raycast, Alfred), it would be great if Vorssaint could natively recognize executable scripts and execute them.

  Specific features requested:

  1. Direct Execution: When pressing Enter on an executable file (.sh, .command, or Unix executables), execute it directly instead of opening it.
  2. Action Modifiers: Provide contextual actions (e.g., Cmd + Enter) to choose between:
      • "Run in Terminal" (Visible execution)
      • "Run in Background" (Silent execution)
      • "Open with Editor" (The current behavior)
  3. (Optional but awesome) Metadata parsing: Support reading standard comment blocks in scripts (like Raycast script commands) to automatically map arguments, dropdowns, and UI elements.

### Feature Area

Windows and Dock

nawazish2:
I'd like to take a narrow slice of this, not the full request.

What I would change:
- When Command Bar Return hits an executable file row (`.sh`, `.command`, or a Unix executable), run it instead of opening it in an editor.
- Keep "Open with Editor" as an extra action on that row (the existing ⌘K panel, next to Show in Finder). ⌘Return stays Show in Finder.
- Reuse `Shell.run` (the same process helper saved script links already use) rather than adding a second runner.

Out of scope for this PR:
- Raycast-style comment-block metadata, argument UI, dropdowns, or a script-command folder (#1178)
- A plugin host
- Changing how saved Command Bar shortcuts of kind `script` already work (live output while typing)
- Remapping ⌘Return, or a "Run in Terminal" action that picks a terminal by bundle id

I can test this on a real Apple Silicon Mac (chmod +x `.sh` and `.command`, plus a non-executable `.sh` that should still open). Happy to wait if someone is already on it, or if this direction should stay closed until a later release.

youtextme:
Raycast Script Commands are a solid Alfred-workflow stand-in for one-off automations like this — free tier is enough; Pro isn’t required to replace a single Alfred script.

[...]

PR #1822, opened 2026-09-21
feat(command-bar): run a script directly from its global shortcut

frixaco:
_Thank you for your work on Vorssaint! Been daily driving it for a month - it is just missing this one thing for me._

## Before / After

Script rows always kept the Command Bar open, because running a script from the bar means "run it, show the output, let me copy it". The side effect: a global shortcut bound to a script row could never actually run the script - it just opened the bar (#1441).

This adds a per-script toggle: **"Run from its global shortcut without opening the bar"**. With it on, the shortcut runs the script immediately - no argument, no UI. If the file is missing or exits non-zero, you get a beep, same as an app shortcut that fails to open. Nothing about the bar changes.

https://github.com/user-attachments/assets/f5fcf093-053b-4b61-a8e2-0ef120cd7826

## Implementation

- `CommandBarLink.runsDirectly` - new bool, decoded with a default of false, so existing saved links are unaffected.
- `CommandBarLinks.directRunScript(forStableKey:in:)` - pure lookup from a row key (`link.<uuid>`) to a script marked for direct run. Anything else returns nil and the shortcut keeps opening the bar as before.
- `CommandBarService.runRow` checks the lookup after the usual entry checks, so hidden rows and a disabled Links source don't run anything. On a hit, `CommandBarCatalog.runScriptDirectly` runs the file with the same bounded `Shell.run` the bar's script runner uses (5s timeout, 64 KB output cap), on a background queue.
- One new toggle in the script editor, localized in all 13 languages.

## Why this shape

#977 : in the bar, a script row is "print an answer", and I tried to not change that. So running silently is opt-in per script - nothing runs without the person who saved it asking for it. Scripts like `open -a App --args …` are the target here.

## Issues

Refs #1441
Refs #977
Refs #1393

- #1441 - covers only the script-row case: a global shortcut that runs the script directly.
- #977 - kept the "scripts print answers" semantics; direct run is opt-in per script.
- #1393 - deep links go through the same row-execution path, so a script row hit via `vorssaint://run/<id>` would still just open the bar. `runsDirectly` is the piece that PR would need.

Not covered here: running scripts from Return in the bar (#576), radial menus (#530).

I couldn't a PR that implements/addresses this. Closest seems #1393, which is a different trigger and still opens the bar for scripts.

## Tested

Apple Silicon Mac, macOS 27.0.

1. Saved an executable script (`#!/bin/zsh` + `open -a Helium --args --remote-debugging-port=9222`) as a Script link with the toggle on.
2. Bound ⌃⌘B to the row (⌘K in the bar → "Give it a shortcut").
3. Pressed ⌃⌘B with any app frontmost and Shortcuts closed - the script ran, no bar. A failing script beeps.
4. `./build.sh --test` - all suites pass (59,376 checks), including new tests for the flag decode, the key→script lookup, and loading links saved before this field existed.
5. `./build.sh --dev --install` - builds, signs, launches, hotkey works.

Untested: scripts that run longer than the 5-second timeout (they get killed and beep), and the release-workflow build path.


Sources/Vorssaint/Services/CommandBar/CommandBarService.swift
```diff
@@ -667,6 +667,15 @@ final class CommandBarService: ObservableObject {
             NSSound.beep()
             return
         }
+        // A script marked to run directly does its work at once, with no
+        // argument and nothing on screen. The lookup rides behind the entry,
+        // so a hidden row or a switched-off Links source keeps its silence
+        // here too.
+        if let link = CommandBarLinks.directRunScript(forStableKey: key, in: CommandBarLinks.decode(
```
vorssaint, line 673: the entry lookup includes hidden rows and disabled sources, so the shortcut still launches this script after its row is hidden or links is switched off. please check those saved preferences before the direct-run branch

frixaco: Thanks for reviewing. Should be fixed - the direct-run branch now checks the row’s hidden state and the Your shortcuts source before execution.
Settings changes also refresh those caches immediately (so disable, re-enable, and unhide actions apply without reopening Command Bar).

Tested both Hiding and "Your shortcuts" toggle cases.

Issue #1796, opened 2026-09-20
Add a configurable keyboard shortcut to upload the last taken screenshot with a default link expiry


samanyudas:
### Problem or Motivation

Uploading the latest screenshot currently requires opening the chain icon menu and choosing an expiry each time.

A keyboard shortcut would make repeated uploads faster.

### Proposed Solution

Add a configurable keyboard shortcut for uploading the most recent screenshot.

In Settings, let the user choose the expiry used by that shortcut:

* 1 hour
* 6 hours
* 24 hours

When the shortcut is pressed, Vorssaint should:

1. Upload the most recent screenshot using the selected expiry.
2. Copy the generated link to the clipboard.

This should behave the same as manually choosing that expiry from the chain icon menu.

### Feature Area

Tools

### Mockups / References

<img width="496" height="313" alt="Image" src="https://github.com/user-attachments/assets/5e864a76-7f40-4879-8f10-6abf23a72d0f" />

The existing chain icon menu already has the required **1 hour**, **6 hours**, and **24 hours** options. The request is to expose one of those as a default action that can be triggered with a configurable keyboard shortcut.

PR #1829, opened 2026-09-21
feat(screenshot): upload latest capture with a default link expiry

samanyudas:
## Summary

Add an opt-in, configurable shortcut to upload the latest Vorssaint screenshot and copy its link, using a saved expiry of 1, 6, or 24 hours.
The preview's main link button uses the same default, while its dropdown lets users choose a different duration for that upload.
Successful link copying dismisses the floating or Dynamic Island preview, including when triggered by the shortcut.

The default shortcut is Control+Option+Command+U with a 1-hour expiry.
Settings participate in backup, all 13 locales are updated, and the latest capture remains available independently of the editor shortcut.

Undelivered links are revoked when their preview closes before completion.
A failed clipboard copy reports that the link was created, and the shortcut retries copying the same active link instead of uploading again.

## Verification

- `./build.sh --test`: passed all 59,609 checks, including preference cleanup.
- `./build.sh --dev`: passed without compiler warnings.
- `./build/VorssaintDeveloper --selftest`: `SELFTEST OK`.
- `git diff --check`: passed.
- Added tests for expiry selection and fallback, upload request duration, enablement and retention rules, settings backup, and successful versus failed copy dismissal.
- Reproduced the original menu-only control and verified the rebuilt split button, duration dropdown, and default-expiry label on Mac15,12 running macOS 27.0 (26A428), in English.
- Contributor reports the manual checks pass. Live upload and Dynamic Island dismissal were not independently exercised during automated validation, but passed manual checks. 

- Added production-callback regression tests for closing before and immediately after completion, releasing the preview, and retaining copy controls after clipboard failure.
- The close-after-callback regression fails against the previous handler and passes with the fix.
- Added retry tests for new captures, expiry, and deletion.
- Follow-up failure scenarios used controlled dependencies; live clipboard-failure injection and remote revocation were not exercised.

## Related issues

Refs #1796


Sources/Vorssaint/Services/QuickTools/ScreenshotQuickPreviewController.swift
```diff
@@ -647,10 +655,14 @@ private struct ScreenshotQuickPreviewView: View {
                     Image(systemName: "link")
                 }
             }
+            .frame(width: 22, height: 18)
+        } primaryAction: {
+            share(.saved())
             .frame(width: embedded ? 28 : 22, height: embedded ? 28 : 18)
```
vorssaint: the primary action returns void, so attaching frame to it prevents both ci configurations from compiling. please move the frame modifier back onto the menu label, keeping the embedded sizing on the view rather than on the upload call

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -478,6 +496,49 @@ final class ScreenshotService: ObservableObject {
...
+        }
+        let uploadingPreview = preview
+        let captureID = latestCaptureID
+        uploadingLatestCapture = true
+        QuickToolHUD.show(icon: "link", message: strings.sharingHUD)
+        shareDirect(capture, duration: .saved()) { [weak self, weak uploadingPreview] record in
+            guard let self else { return }
+            self.uploadingLatestCapture = false
+            guard let record else { return }
+            self.copyUploadedLink(record, captureID: captureID, preview: uploadingPreview)
```
vorssaint, line 528: please use the same preview-lifetime handling as the link button when the shortcut starts with that preview open. this callback copies the link even after the preview closes or a new capture replaces it; a nil weak preview only makes dismissal a no-op. uploading capture a and then taking b can overwrite the newer clipboard contents with a's link, while the button path revokes the undelivered link. keep the standalone shortcut usable without a preview, but revoke abandoned preview uploads and avoid the stale copy

Sources/Vorssaint/Services/QuickTools/ScreenshotQuickPreviewController.swift
samanyudas: Fixed. The frame modifier is back on the menu label, with the embedded size applied there. The primary action now only calls share(.saved()).

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
samanyudas: Fixed. If a shortcut upload starts with a preview open, it only copies the link while that preview is still active. Closing or replacing the preview revokes the uploaded link and leaves the clipboard unchanged. Uploads started without a preview still work. I added regression tests for these cases.

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -478,6 +496,56 @@ final class ScreenshotService: ObservableObject {
...
+            self?.uploadingLatestCapture = false
+            guard let record else { return }
+            guard let self,
+                  !hadPreview || (uploadingPreview != nil && self.preview === uploadingPreview) else {
+                Task { @MainActor in
+                    try? await ScreenshotShareService.shared.delete(record)
+                }
+                return
+            }
+            self.copyUploadedLink(record, captureID: captureID, preview: uploadingPreview)
```
vorssaint: only the link button puts the preview in the sharing state that pauses its auto dismiss, so a shortcut upload started with the preview open leaves the timer running. if the upload outlasts the time the preview has left with the pointer away from it, three seconds after a default action and twelve otherwise, the timer closes the preview and this guard deletes the new link, so the press ends with nothing copied and no message. please keep the preview open while an upload tied to it runs, as the link button does, so only a new capture or an explicit close revokes it. running the open preview's own link action from the shortcut would also cover a preview reopened from history, which now shows an older capture while the shortcut uploads the latest one and closes that preview with link copied. an upload started with no preview still copies after a newer capture, so taking b while a uploads puts a's link over b's automatic copy and shows link copied while b's preview is open. please copy only while no newer capture was taken since the press and revoke the link otherwise

samanyudas: Fixed in 0f8b734b. The shortcut now uses the open preview's link action, so it pauses dismissal, uploads the displayed capture (including history), and reuses its copy retry. Without a preview, completion checks the capture ID and revokes the link if a newer capture has arrived, without copying or closing a later preview.

The new regressions fail six assertions against the old handler and pass with the fix. All 60,079 checks passed, including 736 screenshot checks; the developer build and selftest also passed.

I also verified live uploads through the floating preview and Dynamic Island: each copied a working link and dismissed its capture preview. For cancellation, a temporary QA-only 30-second delay after the real server response let me close the floating preview before delivery. The app removed the record, and both the hosted page and image returned 404 after previously serving the link. The delay was reverted and the test links were revoked. The shortcut-specific timing, history, and newer-capture races were verified with production-handler regression tests, not physical hotkey end-to-end testing.

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -481,6 +499,57 @@ final class ScreenshotService: ObservableObject {
...
+            preview.shareLink()
+            return
+        }
+        guard !uploadingLatestCapture else { return }
+        if let record = linkCopyRetry.record(for: latestCaptureID,
+                                             availableRecords: ScreenshotShareService.shared.records) {
+            copyUploadedLink(record, captureID: latestCaptureID)
+            return
+        }
+        guard let capture = ScreenshotLastCaptureStore.load() else {
```
vorssaint: with an editor open on the latest capture there is no preview, so the shortcut reaches this load and uploads the untouched capture that route saved, not what the editor shows. anything pixelated, redacted or cropped out there still goes into the public link, annotations drawn there are left out, and the message says link copied. this is the usual path when captures open straight in the editor, and it also follows edit in the preview or the edit latest screenshot shortcut. this was already in the earlier push and should have come with the last round. please upload what that editor renders, as its own link button does, or keep the shortcut from sending the original while that editor is open

samanyudas: Fixed in `3547f5fe`.

The root cause was that the shortcut only knew about the preview and the stored untouched capture. It now tracks every open editor associated with the latest capture and stops before retrying or loading that stored capture while any of those editors remains open. Default Edit, Edit from the current preview, and Edit Latest Screenshot register that ownership; history and clipboard editors do not. A newer capture clears the old association.

I kept a visible preview ahead of this guard, so a reopened history preview still runs its own Link action and uploads the pixels it shows.

I used the allowed suppression path because calling `ScreenshotEditorController.share` directly would bypass the view's pending text commit and could still omit visible edits. The screenshot suite passes 758 checks, the developer app builds, and a production-service upload/revocation check returned 201, 200 before deletion, 204 on deletion, and 404 afterward.

Sources/Vorssaint/Services/QuickTools/ScreenshotService.swift
```diff
@@ -397,31 +417,35 @@ final class ScreenshotService: ObservableObject {
...
+        presentPreview(capture, defaultAction: ScreenshotDefaultAction.current,
+                       ownsLatestCapture: true)
     }
 
     /// A history item returns to the same floating preview without repeating
     /// automatic copy or save actions that already ran when it was captured.
     func restorePreview(_ capture: ScreenshotSelectionController.Capture) {
         preview?.close()
-        presentPreview(capture, defaultAction: .none)
+        presentPreview(capture, defaultAction: .none, ownsLatestCapture: false)
```
vorssaint: a capture reopened from recent captures comes through here with ownslatestcapture false, so choosing edit on that preview opens an editor the new guard never counts. restoring the newest screenshot there is a common way back to a capture whose preview has already gone, and the shortcut then loads the stored original, publishes whatever was pixelated, redacted or cropped out in that editor and says link copied. editing the clipboard image, from its shortcut or from clipboard history, ends up the same way after the latest capture was copied. the earlier request should have named these routes too. please count these editors as well while they show the latest capture, or hold the stored upload while any screenshot editor is open, and have a blocked press beep or say why instead of doing nothing

Issue #1503, opened 2026-09-08
App switcher brings all windows to the front


galpressman:
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When multiple windows of the same app are open, switching to one of them brings all windows to the front (not just the one I selected).

My expectation is that only the one I selected would be brought to the front.

### Steps to Reproduce

1. Open multiple windows of the same app.
2. Open other windows the will cover the screen.
3. Alt-tab to one of the windows from step 1.
4. All windows from step 1 will be brought to the front.

### Feature Area

Windows and Dock

### Vorssaint Version

3.3.5

### macOS Version

macOS 26 (Tahoe)

### Language

English

PR #1830, opened 2026-09-21
fix(switcher): stop stale window focus retries

DiogoDuart3:
## Summary

- stop redundant ordinary App Switcher retries when the selected window is already focused
- preserve the ordinary source-to-target handoff retry while the target settles
- retain the frontmost app as the activation source for Dock Preview and Command Bar, so ordinary windows still get their settling retry without reclaiming focus after a later unrelated activation
- preserve initial minimized-window restoration with and without a source app
- keep fullscreen and SpaceHop delayed-retry behavior intact
- add regression coverage for focused, stale, handoff, fullscreen, minimized-window, and dock/command-bar source retention cases

This addresses a delayed App Switcher focus pass that could re-focus the selected Cursor window after Safari was already active, changing the Cursor window frame.

## Verification

- `./build.sh --test-suite=switcher-model` — TESTS OK (1342 checks)
- `./build.sh --test-suite=metrics` — TESTS OK (213 checks)
- Manual macOS 27 validation of the focused-window fix in `587b99bb` after reinstalling and restoring Accessibility plus Screen Recording permissions: repeated Safari/Cursor Command-Tab switches no longer expanded the Cursor window. This follow-up preserves that behavior while restoring the reviewed handoff, fullscreen, minimized-window, and dock/command-bar paths.


Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -1364,13 +1373,19 @@ enum SwitcherSupport {
...
+        let waitingForInitialMinimizedRestore = targetStartedMinimized
+            && targetIsMinimized
+            && initialFrontmostPID == sourcePID
         if !ignoresForeground,
-           let sourcePID, let initialFrontmostPID,
-           initialFrontmostPID != targetPID && initialFrontmostPID != sourcePID && initialFrontmostPID != ownPID {
+           let initialFrontmostPID,
+           initialFrontmostPID != targetPID,
+           initialFrontmostPID != ownPID,
+           !waitingForInitialMinimizedRestore {
```
vorssaint, line 1406: this exception only holds when the frontmost app is the source. dock preview and the command bar reach activate with no source, so for them a window that started minimized now loses its restore pass whenever the previous app is still frontmost at the 120 ms mark, where a nil source used to keep every pass. is that intended, or should a nil source keep the pass while the window is still minimized

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -54,9 +54,18 @@ final class SwitcherWindowFocusRetryState {
...
+                        targetWindowIsFocused: @autoclosure () -> Bool = false,
                         ignoresForeground: Bool = false,
                         ownPID: pid_t = ProcessInfo.processInfo.processIdentifier) -> Bool {
         guard isActive else { return false }
+        let observedFrontmostPID = frontmostPID()
+        if !ignoresForeground,
+           !targetStartedMinimized,
+           observedFrontmostPID == targetPID,
+           targetWindowIsFocused() {
+            isActive = false
```
vorssaint, line 68: once a pass finds the target frontmost with the selected window focused this ends the whole state, so in the fullscreen chain the later passes at 0.38 s and 0.68 s never run, and the ordinary path now also stops while the source is still frontmost, which was the one retry for an app that settles late. did the frame change in cursor still happen with only this focused window check, without the foreground change from 2e198900? if this check alone covers it, keeping the handoff retry would preserve the late settle case

vorssaint:
refs #1503 points at the report about every window coming forward, which #1545 already fixed, so the bot would ask that reporter to confirm this change too. if there is no report for the frame change, describing the symptom in the body without that reference keeps the two apart

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
DiogoDuart3: Good catch. That was not intended. The retry now remains active while the target is still minimized when sourcePID is nil, and the regression suite covers dock/command-bar activation without a source app.

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
DiogoDuart3: Correct: the focused-window check is sufficient for the Cursor frame regression. I scoped it to the ordinary retry path, restored the source handoff exception, and kept the fullscreen chain eligible for its later passes. Regression tests now cover all three cases.

Sources/Vorssaint/Services/Switcher/SwitcherSupport.swift
```diff
@@ -1364,13 +1375,20 @@ enum SwitcherSupport {
...
+        let waitingForInitialMinimizedRestore = targetStartedMinimized
+            && targetIsMinimized
+            && (sourcePID == nil || initialFrontmostPID == sourcePID)
         if !ignoresForeground,
-           let sourcePID, let initialFrontmostPID,
-           initialFrontmostPID != targetPID && initialFrontmostPID != sourcePID && initialFrontmostPID != ownPID {
+           let initialFrontmostPID,
+           initialFrontmostPID != targetPID,
+           initialFrontmostPID != ownPID,
+           initialFrontmostPID != sourcePID,
```
vorssaint, line 1405: the earlier request missed ordinary windows in this handoff. dock preview and the command bar omit the source for these too, so if the previous app is still frontmost at the 120 ms pass, this drops the only settling retry. can those calls retain the initial source so the handoff survives without reclaiming focus after a later unrelated activation?

DiogoDuart3: Good catch. Dock Preview and Command Bar now retain the frontmost app at activation start (Command Bar before its beat), and `WindowActivator` resolves the same retained source when callers omit one. Ordinary handoff retries keep settling while that source is still frontmost, and still stand down after an unrelated activation. Regression coverage includes the helper, the retry cases, and the call sites.

Sources/Vorssaint/Services/Switcher/WindowActivator.swift
```diff
@@ -35,6 +35,15 @@ enum WindowActivator {
...
         guard let app = NSRunningApplication(processIdentifier: item.pid) else { return }
+        // Dock Preview and the Command Bar often omit a session source. Keep
+        // the frontmost app from activation start so the ordinary handoff
+        // retry still settles, without reclaiming focus after a later
+        // unrelated activation.
+        let sourcePID = SwitcherSupport.retainedActivationSourcePID(
+            explicit: explicitSourcePID,
+            targetPID: item.pid,
+            frontmostPID: NSWorkspace.shared.frontmostApplication?.processIdentifier
+        )
```
vorssaint: this fallback makes the app in front at activation start the source for every caller that passes none, and that value also arms the minimize restore in watch target minimize if needed, which runs before the retry check and even with retry off. after a dock preview click, a command bar window pick or a focus follows mouse hover into another app, minimizing that window later would now bring back the app that was in front before, which none of these paths did. focus follows mouse activates with retry off, so there the fallback only adds this restore, plus an accessibility observer set up on the main thread for every such hover. the source that dock preview and the command bar now pass explicitly takes the same route. can the retained app reach only the focus retry guard, for those two callers as well, so the minimize restore stays limited to the switcher session source?

DiogoDuart3: Agreed, that fallback reached too far. In dfd311a3, activation no longer adopts the frontmost app on its own. Dock Preview and the Command Bar pass it as a separate `handoffSourcePID`, which only reaches the delayed focus guards (the ordinary and fullscreen retry checks plus the app-only fullscreen retries). The minimize restore, source staging and Space hops keep using the App Switcher session source alone. So after a Dock Preview click or a Command Bar window pick, minimizing that window later no longer brings back the previous app. Focus follows mouse passes no source, so it arms neither the restore nor its Accessibility observer. Regression coverage includes the source resolution helper and the activator wiring (only the session source reaches `watchTargetMinimizeIfNeeded` and SpaceHop), and checks that the Dock and Command Bar call sites pass only the handoff source.

Sources/Vorssaint/Services/Switcher/WindowActivator.swift
```diff
@@ -35,6 +36,14 @@ enum WindowActivator {
...
 
         guard let app = NSRunningApplication(processIdentifier: item.pid) else { return }
+        // Only the delayed focus guards may treat a caller's handoff app as
+        // the source. The minimize restore, source staging and Space hops stay
+        // limited to the App Switcher session source.
+        let retrySourcePID = SwitcherSupport.focusRetrySourcePID(
+            sessionSourcePID: sourcePID,
+            handoffSourcePID: handoffSourcePID,
+            targetPID: item.pid
+        )
```
vorssaint: dropping the fallback also takes the settling retry away from switcher sessions that open without a source, which happens whenever `sessionsourceitem` returns nil, for example when the app in front has no window left and no windowless entry. `commitsession` then passes a nil `sourcepid` and no handoff, so if that app is still frontmost at the 120 ms pass the foreground check, now applied to a nil source as well, ends the retry that main kept and that the previous push kept through the fallback. can `commitsession` pass the app in front as `handoffsourcepid` too? a session source still wins in `focusretrysourcepid`, so only sourceless sessions would change and the minimize restore, staging and hops stay on the session source

DiogoDuart3: Yes. In a720e5cb, `commitSession` captures the app in front before `endSession()` and passes it as `handoffSourcePID`. When `sessionSourceItem` returns nil, that app keeps the 120 ms settling retry, as on main. A session source still wins in `focusRetrySourcePID`, so sessions with a source item behave the same, and the minimize restore, staging and hops still use only the session source. This process and the target are filtered out as before. A regression check covers the capture order and both arguments at the call site.

PR #1831, opened 2026-09-21
fix(issue#1707): search settings no longer throws focus out

### LINKED ISSUE #1707: Search settings throws you out after typing 1 letter
### Before you file

- [x] If the feature does nothing at all, I checked that Vorssaint is listed and switched on for the permission it needs in System Settings, Privacy and Security, and toggled it off and back on.
- [x] It still happens on the latest version.

### Description

When typing in the search bar of the Vorssaint settings, only 1 letter is being typed. Then it throws the focus out of the search bar and mac plays a "bonk" sound for every key I press.

Expected behavior:
All letters that I type end up in the search bar, not just the first one.

### Steps to Reproduce

1. Open Vorssaint settings
2. Click in search bar on top left
3. Start typing e.g. "window"
4. Only "w" will end up in the search bar
5. Rest of the letters only trigger mac to play a "bonk" sound

### Feature Area

Other

### Screenshots / Recordings

https://github.com/user-attachments/assets/530a0926-068f-4d16-ae31-b55efbf667d8

### Vorssaint Version

3.3.5

### macOS Version

macOS 15 (Sequoia)

### Language

English (US)

### Mac and displays

MacBook Pro M1 (2021), built in

### Installation Method

Homebrew

overstock718:
## Summary

Only the first typed character used to register the rest cause macos to beep, because the search field lost focus as soon as results appeared. Now the whole query types normally into the field, with focus held

## Verification

Tested in the built app by typing in Settings Search. It did not throw me out. Additionally here automated tests:
```
./build/VorssaintDeveloper --selftest

SELFTEST OK
./build.sh --test-suite=preferences
▸ Building & running tests against MacOSX26.sdk…
preferences: OK (87 checks, 0.11s)
TESTS OK (87 checks)
 ./build.sh --test-suite=settings   
▸ Building & running tests against MacOSX26.sdk…
settings: OK (97 checks, 0.11s)
TESTS OK (97 checks)
 ./build.sh --test-suite=keyboard
▸ Building & running tests against MacOSX26.sdk…
keyboard: OK (20586 checks, 0.16s)
TESTS OK (20586 checks)
```

## Related issues
Fixes #1707


Sources/Vorssaint/UI/Settings/SettingsView.swift
```diff
@@ -195,6 +156,43 @@ struct SettingsView: View {
...
+                        .tag(item.page)
+                    }
+                }
+            }
+        }
+    }
+
+    @ViewBuilder
+    private func searchResultRows(_ searchResults: SearchResultsSnapshot) -> some View {
+        ForEach(searchResults.groups) { group in
```
vorssaint: the group id here is a SettingsPage, the same type the list selection is bound to, so swiftui tags every row of a group with that page. while a query is active, the group of the page you are on comes up with the sidebar selection highlight on its page row and on each of its suggestion rows, beside the accent highlight that activeSearchIndex draws, and the arrow keys then move one highlight while the other stays put. a small list probe mirroring this structure returns those rows as selected in the table. keeping one list is the right direction, the search rows just need an id the selection cannot match, for example ForEach(searchResults.groups, id: \.parentSuggestion.id), or the list can pass selection nil while hasSearchQuery is true

vorssaint:
the report behind #1707 is from macOS 15, where the sidebar field comes from the searchable modifier, and on macOS 26 the field is a separate view above the list. which macOS did you reproduce the focus loss on before this change, and did that same system stop losing focus with it? i could not exercise macOS 15 here

Sources/Vorssaint/UI/Settings/SettingsView.swift
overstock718: thank you and good catch. I have addressed your comment. I reproduced the selection issue before the change and verified that the extra highlight is gone afterward

overstock718:
I reproduced the issue on macOS 15, and that's also what I used to develop and test this fix. Before the change I  consistently got the focus loss, with this change applied, the search field no longer loses focus

Issue #1737, opened 2026-09-18
Wallpaper panel: pick and apply stills from the menu bar


GabrielDazzi:
### Problem or Motivation

Swapping a desktop still means opening System Settings → Wallpaper and digging through the library. For a folder of my own photos or a known Apple still, that is a long trip for something the Mac already does.

The menu panel already has other thin tools that wrap a one-shot system job. Wallpaper fits that shape: pick and apply, not a wallpaper app inside Vorssaint.

### Proposed Solution

Opt-in Tools feature. New **Wallpaper** section in the menu panel:

- Apple stills from `/System/Library/Desktop Pictures` (stills only; no Aerials / Dynamic)
- Your pictures: images and folders you add, kept as security-scoped bookmarks in Vorssaint (not added to the System Settings wallpaper library)
- Filters: All / Your pictures / Apple
- Click a thumb to apply
- Toggle **Show on all Spaces** (same idea as System Settings):
  - On: every Space and display gets that still
  - Off: current Space only (switch Space, apply again, different still per Space)
- Add image, add folder, open System Settings → Wallpaper
- Long lists page; we do not scrape or write the Settings library UI

**Apply**

- Current Space: `NSWorkspace.setDesktopImageURL`
- Show on all Spaces: patch Desktop slots in `~/Library/Application Support/com.apple.wallpaper/Store/Index.plist` (WallpaperAgent’s store), backup first, then restart WallpaperAgent. If that layout breaks on a later OS, the click still applies via AppKit on the current Space; gallery and bookmarks stay usable.

Adding in Vorssaint only grows Vorssaint’s list. No `UserFolderPaths` / “add to Wallpaper library”.

**Non-goals**

- Aerials, Dynamic Desktop, shuffle playlists, screen savers
- Replacing or syncing the System Settings wallpaper library

### Feature Area

Tools

### Does another app already do this?

macOS System Settings → Wallpaper, including Show on all Spaces. Same job, from the menu bar.

### Mockups / References

Implementation is on `feat/wallpaper-panel`, ready for a PR once this direction is ok.

PR #1836, opened 2026-09-21
feat(wallpaper): add menu panel still picker

GabrielDazzi:
## Summary

Opt-in Wallpaper section in the menu panel: Apple stills, bookmarked own images/folders, click to apply, and Show on all Spaces (WallpaperAgent store). Same job as System Settings → Wallpaper, from the menu bar.

Follow-up for review: restore the panel overlay scroller, publish the catalog before thumbnail decode with nearby-page prefetch and cancel on uninstall, refresh own-picture roots when reopening the section, require a successful Index.plist backup and a supported store shape (including the Linked layout), and apply stills without a blocking sleep.

Refs #1737

## Verification

```sh
./build.sh
./build/Vorssaint --selftest
./build.sh --test
```

All three passed locally (M3 Pro, macOS 26, single built-in display, multiple Spaces). Dev build used for manual apply / all-Spaces checks; release build for the commands above. Review follow-ups also checked with `./build.sh --dev` and `./build.sh --test-suite=wallpaper`.

## Test plan

- [ ] Install Wallpaper from Features hub (opt-in)
- [ ] Panel gallery: All / Your pictures / Apple; paging
- [ ] Click apply with Show on all Spaces off → only current Space
- [ ] Toggle on → apply → other Spaces updated
- [ ] Add image / add folder; reopen section after folder changes → gallery refreshes
- [ ] Long panel still shows a draggable overlay scroller
- [ ] Re-apply the current still without freezing the panel
- [ ] Open System Settings link


Sources/Vorssaint/UI/MenuPanel/MenuPanelView.swift
```diff
@@ -2260,7 +2260,7 @@ private struct OverlayScrollView<Content: View>: NSViewRepresentable {
 
     func makeNSView(context: Context) -> NSScrollView {
         let scroll = NSScrollView()
-        scroll.hasVerticalScroller = true
+        scroll.hasVerticalScroller = false
```
vorssaint: this scroll view also hosts the existing menu sections, metric details and the embedded panel, even when wallpaper is uninstalled. removing its vertical scroller takes away the draggable scroll control from all of them when content exceeds the available height. please retain the existing overlay scroller here so the new feature does not change navigation in the rest of the panel

Sources/Vorssaint/Services/Wallpaper/WallpaperService.swift
```diff
@@ -0,0 +1,362 @@
...
+                if isDir.boolValue {
+                    imageURLs.append(contentsOf: WallpaperSupport.images(inFolder: root))
+                } else if WallpaperSupport.isStillImageURL(root) {
+                    imageURLs.append(root)
+                }
+            }
+            let own = WallpaperSupport.ownEntries(from: imageURLs)
+            let merged = WallpaperSupport.merge(apple: apple, own: own)
+            // HEIC decode here, not on first filter flip
+            WallpaperThumbnailCache.prefetchSync(apple.map(\.previewURL))
```
vorssaint: every refresh decodes every picture before publishing the catalog, although only 24 entries are displayed and the cache holds 400 images. a large added folder therefore delays the whole gallery and repeats this work at launch. uninstalling only invalidates the final publish token, so the scan and decoding continue and can refill the cleared cache. please publish the catalog without waiting for all thumbnails, bound prefetch to the visible or nearby pages, and cancel pending scanning and decoding when the feature is removed

Sources/Vorssaint/UI/MenuPanel/WallpaperSection.swift
```diff
@@ -0,0 +1,260 @@
...
+                pager
+                if let error = service.lastError {
+                    Text(error)
+                        .font(.caption)
+                        .foregroundStyle(.orange)
+                }
+            }
+            .onAppear {
+                // skip rescan if already warm
+                if service.entries.isEmpty {
```
vorssaint: once the first scan has any entries, reopening the gallery never reads the bookmarked folders again. adding or deleting pictures in one of those folders leaves the gallery unchanged until a restart or another add operation, and there is no refresh action. please refresh the own-picture roots when reopening the section or provide a refresh action while retaining the apple catalog cache

Sources/Vorssaint/Services/Wallpaper/WallpaperStore.swift
```diff
@@ -0,0 +1,60 @@
...
+        guard var root = try? PropertyListSerialization.propertyList(
+            from: data, options: [.mutableContainers], format: &format
+        ) as? [String: Any]
+        else { return false }
+
+        guard WallpaperSupport.patchStoreRoot(&root, imageURL: url) else { return false }
+
+        let backup = index.deletingLastPathComponent()
+            .appendingPathComponent("Index.plist.vorssaint-bak", isDirectory: false)
+        try? FileManager.default.removeItem(at: backup)
```
vorssaint: the store replacement proceeds even if creating the backup fails, and the patch helper accepts any dictionary by creating missing containers and always returning success. that means an unsupported store layout is still rewritten instead of taking the current-space fallback, without a guaranteed recovery copy. please validate the supported store shape and require a successful backup before replacing the original file

Sources/Vorssaint/Services/Wallpaper/WallpaperService.swift
```diff
@@ -0,0 +1,362 @@
...
+                do {
+                    try NSWorkspace.shared.setDesktopImageURL(
+                        URL(fileURLWithPath: ""), for: screen, options: [:])
+                    needsPause = true
+                } catch {
+                    // still try the set below
+                }
+            }
+        }
+        if needsPause {
```
vorssaint, line 197: this runs synchronously from the thumbnail button, so reapplying the current image blocks the main thread for 400 ms after the clear succeeds, freezing the panel and the rest of the app. please replace the blocking sleep with a nonblocking continuation and invalidate that continuation if the feature is removed or a newer image is selected

Sources/Vorssaint/Services/Wallpaper/WallpaperSupport.swift
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        return true
+    }
+
+    // current macOS uses Type=linked + Linked; older trees use Desktop
+    @discardableResult
+    static func patchWallpaperSlot(_ container: inout [String: Any],
+                                   slot: [String: Any]) -> Bool {
+        if container["Linked"] is [String: Any] {
```
vorssaint: the comment above and the new fixture say current macos keeps type linked with a linked slot in all spaces and displays and in system default. this branch writes the image choice into that slot and keeps type linked, and because it looks for the linked key rather than the type, a container of type individual that still carries a linked key would lose its desktop slot. if a linked slot is the one choice shared by the desktop and the screen saver, after the wallpaper agent restarts the screen saver may switch to the still or lose the choice the person made. what does the screen saver show after an apply from that state, and what does system settings write when a still is picked there, system default included? please write that same shape, or keep the current space fallback for linked containers until it is known, and add it as a fixture

Sources/Vorssaint/Services/Wallpaper/WallpaperService.swift
```diff
@@ -0,0 +1,407 @@
...
+
+            var imageURLs: [URL] = []
+            for root in roots {
+                guard self?.refreshToken == token else { return }
+                var isDir: ObjCBool = false
+                guard FileManager.default.fileExists(atPath: root.path, isDirectory: &isDir) else {
+                    continue
+                }
+                if isDir.boolValue {
+                    imageURLs.append(contentsOf: WallpaperSupport.images(inFolder: root))
```
vorssaint: images in folder walks the whole tree of an added folder, including the contents of packages such as photo libraries, and the refresh token is only checked outside that walk. with one large folder, removing the feature does not stop the walk, and because every appearance of the section now starts a new refresh, returning to the tab a few times leaves several full walks of the same folder running at once. please check the token inside the folder enumeration so an older walk stops as soon as a newer refresh or the removal happens, and skip package contents

Sources/Vorssaint/Services/Wallpaper/WallpaperService.swift
```diff
@@ -0,0 +1,407 @@
...
+            return NSScreen.screens
+        }
+        if let screen = NSScreen.withMouse {
+            return [screen]
+        }
+        return NSScreen.screens.first.map { [$0] } ?? []
+    }
+
+    // fill crop; clear same-URL first (macOS skips refresh otherwise)
+    private func setDesktopImage(_ url: URL, on screens: [NSScreen], token: UUID) async -> Bool {
```
vorssaint: set desktop image is an async method on a class that has no actor, so it runs on a background executor even though apply awaits it from a main actor task. the desktop image calls and the reads of apply token and is available then happen off the main thread, while apply and sync with preferences write apply token on the main thread. please isolate this method to the main actor, the sleep already keeps the main thread free

Sources/Vorssaint/Services/Wallpaper/WallpaperService.swift
```diff
@@ -0,0 +1,407 @@
...
+
+    // bookmark roots only; folder walk is off-main
+    private func ownRoots() -> [URL] {
+        ownBookmarks.compactMap(resolveBookmark)
+    }
+
+    private func resolveBookmark(_ data: Data) -> URL? {
+        var stale = false
+        return try? URL(resolvingBookmarkData: data,
+                        options: [.withSecurityScope, .withoutUI],
```
vorssaint: without the option that skips mounting, resolving a bookmark can try to mount the volume of the added folder, and refresh resolves every bookmark twice on the main thread, once in start accessing and again in own roots. refresh runs at launch and on every appearance of the section, so a folder on a network share that is not mounted can stall the launch and the panel each time the tab shows. please add that option, as the other bookmark readers in the app do, and resolve each bookmark once per refresh

Sources/Vorssaint/UI/MenuPanel/WallpaperSection.swift
```diff
@@ -0,0 +1,257 @@
...
+                image = cached
+            }
+        }
+        .task(id: previewPath) {
+            if let cached = WallpaperThumbnailCache.image(for: entry.previewURL) {
+                image = cached
+                return
+            }
+            let url = entry.previewURL
+            let loaded = await Task.detached(priority: .utility) {
```
vorssaint: the earlier thumbnail request is only partly done. each cell still decodes in a detached task that neither the view nor the generation can cancel, and a page change queues a new prefetch under the same generation, so paging quickly or closing the panel keeps every requested decode running, and a cell and the prefetch can decode the same image twice. for own images and the stills at the top of the system folder the preview is the full size file. please load cells through the generation, start a new one on page changes, and cancel the cell load when it goes away

Sources/Vorssaint/UI/MenuPanel/WallpaperSection.swift
```diff
@@ -0,0 +1,257 @@
...
+                set: { service.applyAllDisplays = $0 }
+            ))
+            .toggleStyle(.switch)
+            .controlSize(.small)
+
+            HStack(spacing: 8) {
+                Button(text.addImage) { service.addImages() }
+                    .buttonStyle(.bordered)
+                    .controlSize(.small)
+                Button(text.addFolder) { service.addFolder() }
```
vorssaint: this was already in the earlier push and should have come with the first requests. added images and folders can never be removed, since neither the tab nor settings offers a way to do it, and removing the feature keeps them for the next install. a folder added by mistake stays in the list and is walked again at every launch and every time the tab shows. please add a way to remove an added image or folder

Sources/Vorssaint/Services/Wallpaper/WallpaperStore.swift
```diff
@@ -0,0 +1,67 @@
...
+            from: data, options: [.mutableContainers], format: &format
+        ) as? [String: Any]
+        else { return false }
+
+        guard WallpaperSupport.patchStoreRoot(&root, imageURL: url) else { return false }
+
+        let backup = index.deletingLastPathComponent()
+            .appendingPathComponent("Index.plist.vorssaint-bak", isDirectory: false)
+        do {
+            if FileManager.default.fileExists(atPath: backup.path) {
```
vorssaint: the backup is replaced before every write, so from the second apply on it holds what the previous apply wrote and no longer the store the person had before the feature touched it, including any wallpaper set per space. nothing reads it, and it stays in the system store folder after the feature is removed. please keep the copy taken before the first write, and say how it is restored and when it is removed

Sources/Vorssaint/UI/MenuPanel/WallpaperSection.swift
GabrielDazzi: Sorry about that. I completely forgot a way to remove an added image or folder.
Day to day I never needed to take one back out, so it never came up while I was using it. Just didn't notice it was missing.

Sources/Vorssaint/Services/Wallpaper/WallpaperSupport.swift
vorssaint: the individual branch is fixed and covered. the linked branch is unchanged, and the new comment says system settings writes a still into linked and that this store holds no screen saver choice, without a captured store to show either. the linked fixture is built by hand with an empty choice list and its checks already hold before the patch runs, so the test passes whatever is written there. this branch also replaces any linked choice, and a store where no still was ever picked, such as one on the default wallpaper, may already be linked, so linked alone does not show the still shape. please attach the index, or its key structure, that system settings leaves after picking a still in a store this feature never wrote, and say what the screen saver shows after an apply here, or keep the current space fallback for linked containers without reporting it as a failed apply

Sources/Vorssaint/Services/Wallpaper/WallpaperStore.swift
vorssaint: the copy is now kept from the first write. the new comment keeps it for manual recovery, but restore original index and remove backup have no caller and nothing in the app points to the file, so it is never restored and stays in the system wallpaper store after uninstall vorssaint completely and the uninstall script, which only clear the app's own folders. please keep the copy in the app's own support folder, which both already clear, or delete it during those uninstalls, and either give the restore a reachable caller or drop it

Sources/Vorssaint/Services/Wallpaper/WallpaperService.swift
```diff
@@ -0,0 +1,625 @@
...
+    var isAvailable: Bool { AppFeature.wallpaper.isAvailable }
+
+    var ownFolderSources: [OwnSource] {
+        ownSources.filter(\.isFolder)
+    }
+
+    // folders + unreachable (no gallery cell) — chips in remove mode
+    var removableChipSources: [OwnSource] {
+        ownSources.filter { $0.isFolder || !$0.isReachable }
+    }
```
vorssaint: the type gap was already in the earlier push and should have come with the first requests. add image accepts any image type, while the gallery lists only the extensions in image extensions, so a picked heif, avif, camera raw or psd file is saved as a bookmark that never gets a cell. remove mode offers an x only on gallery cells and chips only for folders and unreachable sources, so that bookmark can be neither seen nor removed, and when it is the only added source the remove button opens an empty remove mode. please limit the open panel to the listed extensions, or show a chip for any source without its own gallery cell

Sources/Vorssaint/Services/Wallpaper/WallpaperSupport.swift
```diff
@@ -0,0 +1,360 @@
...
+            .appendingPathComponent("\(stem).heic")
+        let plistThumbnail = madesktopThumbnailPath(at: url).map { URL(fileURLWithPath: $0) }
+
+        let imageURL: URL?
+        if fileManager.fileExists(atPath: wallpaperAsset.path) {
+            imageURL = wallpaperAsset
+        } else if let plistThumbnail, fileManager.fileExists(atPath: plistThumbnail.path) {
+            imageURL = plistThumbnail
+        } else if fileManager.fileExists(atPath: thumbnail.path) {
+            imageURL = thumbnail
```
vorssaint: this was already in the earlier push and should have come with the first requests. when a madesktop entry has no still under .wallpapers, it applies the file from its thumbnail path or from .thumbnails, and those files are previews a few hundred pixels wide, so picking such an entry leaves a blurred wallpaper. please list only entries that have a full size still, and keep the thumbnail as the preview only

Sources/Vorssaint/Services/Wallpaper/WallpaperSupport.swift
GabrielDazzi: Checked on a store System Settings wrote alone, no Vorssaint bak in the mix
After picking a still there, AllSpacesAndDisplays and SystemDefault are Type: linked with a Linked slot. Provider com.apple.wallpaper.choice.image, imageFile config. Index.plist has no separate screen saver key. The still choice itself sits in Linked
We write that same shape now, keyed off Type, so an individual container with a leftover Linked key keeps its Desktop slot
I applied from Vorssaint with a screen saver already in play. After apply, the screen saver showed the same still I had just set as wallpaper. Same as picking that still in System Settings.
If a linked layout cannot be patched, current-space AppKit apply still runs and we do not treat that as a failed apply

PR #1841, opened 2026-09-21
fix(switcher): stop the window-fronting click from zooming the window

sarat03:
Since 3.4.0-beta.3, switching to a window with the App Switcher zooms it: the window animates to its zoomed frame at the top-left of the work area and resizes. Every app, keyboard only, and a later switch toggles it back out. 3.4.0-beta.2.1 is unaffected.

Fronting the selected window (added in 0a9c36cf, #1545) posts a synthetic mouse down/up pair into the target app so the window becomes key. The location was written as a `CGPoint`, which is two 64-bit doubles. Those sixteen bytes run past the coordinates and into the fields packed beside them, the click count among them, so an app could read the pair as a double click in its title bar. A title-bar double click zooms, which is exactly the observed animation, at each app's own zoomed size.

This fills the location and its neighbouring fields with `0xff`, so no usable click can be read out of the record and only the key-window side effect remains.

How it was confirmed: a sampler recorded every window frame change alongside the frontmost app and the screen work area. With Vorssaint running, six keyboard-only switches produced six frame changes; with it quit, the same gesture produced zero. The work area never changed, which ruled out a Dock or menu bar effect, and turning Dock Preview off changed nothing (10 / 10 / 11 across on, off, on). System Settings zoomed to full height at its fixed 740pt width, which is the app's own zoom rather than any frame this code computes. With this commit, the same measurement reports zero frame changes.

Tests: `./build.sh --test` passes, 59745 checks.


vorssaint:
the fix in #1844 already puts the same no-location bytes into the window-fronting record on main and adds coverage for the press and release. this patch has no remaining behavior or test to carry over, so closing it as superseded

PR #1842, opened 2026-09-21
feat(finder): copy selected item paths

ajitkshukla:
## Summary

Describe what changes and why. For a fix, explain the behavior before and after.
Mention partial fixes or dependencies on another PR.

## Verification

List checks actually run and their results, plus anything you could not test.
Include Mac, macOS, display or language details when relevant to the behavior.
See the [contribution guide](https://github.com/vorssaint/vorssaint-utils/blob/main/CONTRIBUTING.md#validation) for available checks.


Sources/Vorssaint/Services/Finder/FinderCutPaste.swift
```diff
@@ -337,6 +376,29 @@ final class FinderCutPaste: ObservableObject {
...
+    private func copySelectedPathsAsync() {
+        copyPathInProgress = true
+        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+            guard let text = FinderCopyPathSupport.text(for: FinderBridge.selectionURLs()) else {
+                DispatchQueue.main.async { self?.copyPathInProgress = false }
+                return
+            }
+            GeneralPasteboardAccess.shared.async {
+                let pasteboard = NSPasteboard.general
+                pasteboard.clearContents()
```
vorssaint, line 392: if finder takes time to reply, this clears anything copied in the meantime and can still write after copy path is switched off. please discard stale requests and preserve newer clipboard contents before writing the paths

Issue #1743, opened 2026-09-18
Scratchpad: add a formatting toolbar so the Markdown support is findable


sarat03:
### Problem or Motivation

Splitting this out of #395 rather than adding to that thread, since that
issue is a direction discussion covering seven separate things and this is
one concrete change.

Markdown landed in 3.3.2 and it works. Headings, lists, quotes, code, bold,
italic and strikethrough all render correctly.

The problem is that nothing in the pad tells you so. The eye button's tooltip
says "preview formatting", but it never says what formatting you can write,
and the placeholder in an empty pad does not mention Markdown either. So if
you already know the syntax you get a good editor, and if you do not, you get
a plain text box and no way to discover otherwise. Plenty of people who would
use headings and bullets have never typed a # or a *.

The capability is built and paid for. It is just invisible.

### Proposed Solution

A small formatting toolbar: bold, italic, strikethrough, heading, bullet.
Five buttons that wrap or prefix the current selection with the exact
characters you would have typed by hand.

What this is not:

* Not a rich text editor. The buttons insert Markdown characters into the
  same plain text buffer. What lands on disk is unchanged.
* Not new Markdown capability. Every mark a button would insert is already
  parsed and rendered today. No parser work, no new block types, no file
  format change, no migration.
* Inert in preview mode, so it shows only while you are editing.

This fits "support only as little as necessary", because it adds nothing to
support. It makes the existing support findable.

Open question on placement: the footer row already carries the preview
toggle, copy, clear and export, and the pad is narrow. It could be a second
row visible only while editing, or a row that appears only when there is a
selection. Happy to build whichever you prefer.

If the toolbar is unwanted, the cheap version of this is one line in the
empty pad's placeholder saying Markdown works. That solves most of the
discovery problem for almost no code.

I am willing to implement either.

### Feature Area

Other

PR #1847, opened 2026-09-22
feat(scratchpad): add a formatting bar and find text integration

sarat03:
## Summary

Markdown has worked in the Scratchpad since 3.3.2, but nothing in the pad said so. The preview toggle is labelled "Show formatting" and never says what formatting you can write, and the empty pad's placeholder does not mention Markdown either. If you already knew the syntax you had a good editor. If you did not, you had a plain text box and no way to find out.

**A Formatting chip** opens the marks: in the floating pad's footer, and in the island's toolbar where clicking it swaps the tab strip for the row and clicking it again puts the tabs back. The island has one row to spend, so the marks take that row rather than squeezing in beside the tabs. The chip is tinted while the row is up and disabled while previewing, and it stays live on an empty pad, since that is exactly where someone reaches for a first heading.
image

| Floating pad | Dynamic Island |
| --- | --- |
| image | image |

Nine marks: bold, italic, strikethrough, heading, bullet, numbered list, quote, inline code, link. Each writes the characters a typist would have written by hand, so the stored text stays plain and the preview parser is untouched. This adds nothing new to support; it makes the existing support findable.

Behaviour worth knowing:

* Every mark toggles. A second click takes it back off, including from the selection the first click left behind, which for bold is the words and for a link is the address.
* Heading steps down its levels before coming off, so H1, H2 and H3 are all reachable from the one button. It is drawn as "H", because every symbol for a heading reads as font size, which is a different thing plain text cannot say.
* A line mark applied to a line already carrying one replaces it instead of stacking, so bullets convert to numbers, numbers to headings, and nothing becomes "# - text".
* Ordered items are numbered as they are written, so the list reads correctly as plain text and not only once the preview has renumbered it.
* Italic nests inside bold rather than tearing the pair apart, since its marker is also half of bold's.

The row is not a preference. Both pads open with it closed and one click brings it back, so nothing formatting related appears uninvited, and it closes when you switch to preview. Both pads share one definition of the row, with the appearance as a parameter, so a mark added later shows up in both.

**Find** is Command-F, and it turns on AppKit's own find bar rather than a new one. That brings the system match counter, highlighting, Command-G and Command-Shift-G, and the Find and Replace modes an editable text view carries, none of which was worth reimplementing. The key joins the table both pads already consult for Command-T and Command-W; that table stopped being tabs-only, so `ScratchpadFocusedTabShortcut` is now `ScratchpadFocusedShortcut`. The island routes it through the serial its Command-W already uses, since the service does not hold the island's text view.

**Text size** stays a pad-wide preference, beside the background opacity slider. Plain text cannot carry a point size, so one selection cannot differ from another, and the heading levels are the format's own way of making words bigger. They now step up from whatever the preference is set to rather than from fixed points, so the document keeps its shape at any size.

Every edit goes through the text view's own `shouldChangeText` and `didChangeText`, so one Command-Z takes a whole mark back, and selections are handled in UTF-16 throughout, which is what the text view reports.

## Verification

* `./build.sh --test` passes, 59423 checks. 39 of them are new, in `Tests/ScratchpadMarkTests.swift`, covering every mark, both directions of each toggle, carets with no selection, blank lines inside a selection, cross-family line mark replacement, an offset past an emoji, a stale out of range selection that would otherwise raise in NSString, and the size clamp.
* `./build.sh --dev` builds clean, and the marks, the chip, the find bar and the size preference were exercised by hand on macOS 27 in both the floating pad and the Dynamic Island.

## Related issues

Refs #1743


Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
```diff
@@ -427,17 +462,29 @@ final class ScratchpadService: NSObject, ObservableObject, NSWindowDelegate {
...
         }
     }
 
+    /// The text view runs the find itself; it only has to be told which of the
+    /// finder's actions was asked for, and that arrives as a sender's tag.
+    func showFindBar(in editor: NSTextView? = nil) {
+        guard let textView = editor ?? textView.flatMap({ $0.window === panel ? $0 : nil }) else { return }
+        let sender = NSMenuItem()
+        sender.tag = NSTextFinder.Action.showFindInterface.rawValue
+        textView.window?.makeFirstResponder(textView)
```
vorssaint, line 484: please make find target a visible surface. both scratchpad hosts keep the editable text view at zero opacity while previewing, but this path still makes that hidden view first responder and opens its find bar. command-f from show formatting therefore searches the hidden source with no visible find interface. leave preview before showing find, or route it to the visible preview, without focusing an invisible editable view

Sources/Vorssaint/Services/QuickTools/ScratchpadSupport.swift
```diff
@@ -340,3 +358,275 @@ enum ScratchpadSupport {
...
+
+    /// Italic's marker is also half of bold's. Without this, asking for italic
+    /// inside `**bold**` would tear the bold pair apart instead of nesting.
+    private static func isHalfOfBoldPair(wrap: String,
+                                         in ns: NSString,
+                                         before: NSRange,
+                                         after: NSRange) -> Bool {
+        guard wrap == "*" else { return false }
+        let star = "*"
+        let priorIsStar = before.location > 0
```
vorssaint, line 551: please distinguish a bold pair from italic already nested inside bold. starting with a bold word selected, the first italic click produces three stars on each side and leaves the word selected; the second click still sees neighboring stars and adds another pair, producing four stars on each side instead of returning to bold alone. the bold guard needs to preserve the existing bold while allowing the italic toggle to come back off

Sources/Vorssaint/UI/Scratchpad/ScratchpadView.swift
```diff
@@ -572,31 +595,35 @@ struct MarkdownPreview: NSViewRepresentable {
...
         return result
     }
 
-    private static func rendered(_ block: ScratchpadMarkdownBlock) -> NSAttributedString {
+    private static func rendered(_ block: ScratchpadMarkdownBlock,
+                                 base: CGFloat) -> NSAttributedString {
         let paragraph = NSMutableParagraphStyle()
         paragraph.lineSpacing = 2
-        var font = NSFont.systemFont(ofSize: 13)
+        var font = NSFont.systemFont(ofSize: base)
```
vorssaint, line 616: please carry the text-size preference through code blocks too. this base size reaches paragraphs and headings, but the code branch below still replaces it with a fixed 12-point font. setting the pad to 22 makes a pasted fenced code block shrink back to 12 in the preview, so that content stays hard to read even after the user enlarges the pad

Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
sarat03: Confirmed, fixed in aade7d30. The pad leaves preview before the find bar opens, so the view that takes focus is the one on screen.

Following it through turned up a second problem in the island. Leaving preview asks its view for the caret back a turn later, and that refocus landed after the bar had opened, taking the keyboard off the search field and dropping the caret at the end of the note, so the first thing typed went into the text instead of the search. The refocus now stands down while the find bar is up, in ab94c3e6.

Sources/Vorssaint/Services/QuickTools/ScratchpadSupport.swift
sarat03: Confirmed, and it behaved exactly as you describe: four stars on each side, growing by a pair on every click.

One neighbouring character cannot tell italic's star from half of bold's. What settles it is the parity of the run, since an odd run has an italic star sitting inside however many bold pairs are outside it. Italic now has its own path on that rule and toggles back off leaving the bold alone, in aade7d30.

Worth saying why this got through: my tests only ever checked one click, and the bug needs the selection that the first click leaves behind. There is now a round trip check that clicks each mark twice from that selection. Against the old code it fails with `****note****` and then `******note******`.

Sources/Vorssaint/UI/Scratchpad/ScratchpadView.swift
sarat03: Confirmed, fixed in aade7d30. The code branch now takes the same base size as paragraphs and headings, so a pad at 22 keeps fenced code at 22.

Inline code had the same bug in a different shape, a hard `max(12, ...)` floor, which meant it grew instead of shrinking once the pad went below 12. It now steps down from its own base with a floor of 9.

Sources/Vorssaint/UI/Notch/NotchScratchpadView.swift
```diff
@@ -243,6 +273,11 @@ struct NotchScratchpadView: View {
     /// puts it after a tab change.
     private func focusEditor() {
         guard !pad.isPreviewing, let view = editor.view, let window = view.window, window.isKeyWindow else { return }
+        // Leaving preview asks for the caret back, and Command-F from preview
+        // leaves preview on its way to the find bar. This runs a turn later, so
+        // without this it would take the keyboard off the search field the user
+        // just opened and drop the caret at the end of the note.
+        guard view.enclosingScrollView?.isFindBarVisible != true else { return }
```
vorssaint: this check sits in focus editor, so it also skips putting the caret back after a tab change and after a plain return from preview whenever the find bar is still up, not only after command-f from preview. with the bar left open, entering preview resigns the editor and clicking edit now stops here, so the island ends with no text focus and typing does nothing until a click. command-t or a tab change while the search field has the keyboard keeps it in the search instead of putting the caret in the pad, and if the search field has the keyboard when preview starts nothing resigns it, so typing during preview goes into a search nobody can see. please let only the command-f path keep the search field, for example by closing the find bar when the island enters preview and checking it only on the refocus that follows leaving preview

Sources/Vorssaint/Services/QuickTools/ScratchpadSupport.swift
```diff
@@ -189,6 +194,19 @@ enum ScratchpadSupport {
     /// frosted pad, while one fully covers what is behind the window.
     static let backgroundOpacityRange: ClosedRange<Double> = 0...1
 
+    /// The size the whole pad draws at, for anyone who finds 12 small to live
+    /// in. It is a preference rather than a mark: plain text cannot carry a
+    /// point size, so one selection cannot differ from another, and the heading
+    /// levels are the format's own way of making words bigger. They step up
+    /// from whatever this is set to.
+    static let defaultTextSize: Double = 12
```
vorssaint: the pad draws at 13 today, the editor's own default, but this preference defaults to 12, so everyone who never touches the new slider gets a smaller pad in the floating window and in the island, with body text and placeholder at 12, headings a point smaller and inline code at 11. please default it to 13 so an untouched pad keeps today's sizes, and draw fenced code a point under the body as inline code already is, which keeps blocks at today's 12 while still following the slider

Sources/Vorssaint/UI/PlainTextEditor.swift
```diff
@@ -16,9 +16,13 @@ struct PlainTextEditor: NSViewRepresentable {
     /// `lineFragmentPadding` sits inside the inset and is set on the text
     /// container below rather than assumed, so this stays the source of
     /// the number instead of a copy of AppKit's default.
-    static let fontSize: CGFloat = 13
+    static let defaultFontSize: CGFloat = 13
```
vorssaint: main has since added a still of the scratchpad page to the island editor in settings, and it reads the static font size of plain text editor. with font size now an instance property that line stops compiling once main is merged in. please merge main and have the still draw at the pad's text size, and it could show the new formatting button with the others too

Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
```diff
@@ -427,15 +462,33 @@ final class ScratchpadService: NSObject, ObservableObject, NSWindowDelegate {
...
         case .hidePad:
             hide()
+        case .find:
+            showFindBar()
+        }
+    }
+
+    /// The text view runs the find itself; it only has to be told which of the
+    /// finder's actions was asked for, and that arrives as a sender's tag.
+    func showFindBar(in editor: NSTextView? = nil) {
```
vorssaint: esc never reaches the find bar's search field. the island's key monitor collapses the island on it and the floating pad's monitor hides the pad, so the usual way to put find away closes the whole note, and since that monitor only checks the editor for a live composition, esc while composing a search term hides the pad too. please let the search field take esc first, as the mixer's percent field already does in the island, and leave the next esc to close the pad

Sources/Vorssaint/UI/Notch/NotchScratchpadView.swift
sarat03: Confirmed, the guard was in the wrong place. It is out of focusEditor now, so a tab change and Command-T always put the caret back in the pad. Entering preview closes the find bar and resigns whichever of the editor or the search field had the keyboard, so nothing can be typed into a search nobody can see. The floating pad does the same now, through one shared method. Since preview always closes the bar, a bar that is up when preview ends can only have been opened by Command-F, and that refocus is the one place that still checks it. In 65ecfedb.

Sources/Vorssaint/Services/QuickTools/ScratchpadSupport.swift
sarat03: Confirmed, and that was my mistake: I read the default off the wrong number. It is 13 now, so an untouched pad draws body, headings and inline code at exactly the sizes it did before. Fenced code draws a point under the body like inline code, which keeps it at today's 12 and still follows the slider. That also corrects what I said on the earlier thread: code at 22 is now 21, not 22. There is a test pinning the default to 13. In 65ecfedb.

Sources/Vorssaint/UI/PlainTextEditor.swift
sarat03: Merged main in a53829a0. The still draws its note at the pad's text size and shows the formatting chip before the others, in the same order as the island's toolbar, in 65ecfedb.

Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
sarat03: Confirmed in both pads. Both monitors now let Esc through while the find bar's search field has the keyboard, which also covers composing a search term. AppKit's bar closes itself on that Esc and gives the keyboard back to the text, so the next Esc closes the pad or collapses the island as before. I checked that handoff against a bare NSTextView with the find bar on before relying on it. In 65ecfedb.

sarat03:
Before asking for another look I went back over the whole branch and fixed a few things nobody had raised yet, in d38dabd0:

* Taking a line mark off a selection that starts at the line's start (Select All over "### a", then heading) returned a selection longer than the text it covered, so it ran into the line below and the next mark wrapped the wrong span. Both ends of the selection are now worked out separately.
* Line marks keep indentation, so a nested bullet toggles off and back on at its own depth instead of becoming "-   - b".
* Inline marks leave the space a drag picks up outside the markers. "**hello **world" never renders as bold, since Markdown will not close a mark against a space.
* The text size slider reads through a clamping binding like the opacity slider, so a bad value from an imported backup cannot put the thumb off the track.
* Dropped a `glyph` parameter on NotchIconButton that nothing used any more.

Each of the mark fixes has a check in `ScratchpadMarkTests` that fails against the old code.

Sources/Vorssaint/Services/QuickTools/ScratchpadSupport.swift
```diff
@@ -178,6 +181,8 @@ enum ScratchpadFocusedTabShortcut {
             return canCreatePad ? .createPad : nil
         case "w":
             return canClosePad ? .closeSelectedPad : .hidePad
+        case "f":
+            return .find
```
vorssaint: the earlier requests missed this, it was already in the first push. command-g and shift-command-g do nothing in either pad after a search, although the plain text editor comment promises command-g with the find bar. the pad key tables only take t, w and f, neither panel handles key equivalents, the app menu has no find items and nothing in the find bar takes command-g, so both keys fall through to the text view, which has no binding for them. please route g and shift-g to the text finder's next and previous match actions in both pads the way f is routed, keeping in mind that the table's command only guard turns shift-g away as it stands

sarat03: Routed command-g and shift-command-g to the text finder's next and previous match in both pads. The key table now takes shift as its own flag and only lets it through on g, so shift-command-t, w and f still fall through to the text. Stepping through matches leaves the keyboard where it is, so command-g from the search field keeps you in the field, and from preview it comes back to the text first like command-f does. Checks are in ScratchpadMarkTests.

Sources/Vorssaint/Services/QuickTools/ScratchpadService.swift
```diff
@@ -450,17 +486,52 @@ final class ScratchpadService: NSObject, ObservableObject, NSWindowDelegate {
...
+        if isPreviewing {
+            isPreviewing = false
+        }
+        let sender = NSMenuItem()
+        sender.tag = action.rawValue
+        // Stepping through matches leaves the keyboard where it is, so
+        // Command-G from the search field keeps typing in the field.
+        if action == .showFindInterface {
+            textView.window?.makeFirstResponder(textView)
+        }
```
vorssaint: command-g and shift-command-g pressed in preview do not end with the match selected in a focused editor. entering preview took the keyboard from the editor, and a step leaves preview here without making the text view first responder again, so the floating pad selects the match in a view that is not taking keys and typing goes nowhere until a click. in the island the refocus that follows leaving preview still runs, because preview closed the bar and a step does not reopen it, so focus editor moves the caret to the end of the note and the match is gone. please give the text the keyboard when a step ends preview and keep the island's refocus from moving the caret after it, or leave these keys alone while previewing

PR #1856, opened 2026-09-22
fix(mixer): volume slider exponential scaling

ahmedabdullah-o:
## Summary

Now the audio mixer uses exponential gain scaling instead of linear scaling.
Linear scaling is not intuitive because humans hear changes in volume in
a logarithmic way.

## Verification

Device: MacBook Air M1 8GB 256GB

I ran the routine verifications:
```sh
./build/Vorssaint --selftest
./build.sh --test
```
No warnings, no failed tests.
This approach resolves all issues in previously rejected issue:
- [x] Max volume capped to 200%.
- [x] No UX disruption, the previously set values stay untouched.
- [x] Numbers next to sliders show real values, not slider position.

Ref: #1588

### DIFF
--- a/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
+++ b/Sources/Vorssaint/Services/Audio/AppVolumeMixer.swift
@@ -1902,7 +1902,7 @@ private final class TapGainEngine: GainEngine {
             // here on leaves the output written.
             let frames = MixerRender.render(source: inputBuffers[tapIndex],
                                             into: outputBuffers,
-                                            gain: gain)
+                                            gain: pow(gain, 2))
             // Keep the tiny delay filled for every live engine. Crossing from
             // attenuation into boost then changes level without inserting a
             // fresh block of silence into audio that is already playing.
--- a/Sources/Vorssaint/Services/Audio/MixerRender.swift
+++ b/Sources/Vorssaint/Services/Audio/MixerRender.swift
@@ -101,9 +101,7 @@ enum MixerRender {
                                              channels: buffer.mNumberChannels))
         }
         guard frames > 0, outputChannels > 0, hasWritableOutput else { return 0 }
-        // Converting from linear to exponential gain for perceptually smooth gain adjustments.
-        // This is due to the non-linear perception of human hearing.
-        var gain = pow(gain, 2)
+        var gain = gain
 
         // The usual shape: one interleaved buffer wanting exactly what the
         // tap produced. One pass, nothing to map.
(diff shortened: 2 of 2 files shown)


Sources/Vorssaint/UI/MenuPanel/MixerSection.swift
```diff
@@ -834,8 +834,8 @@ private struct MixerRow: View {
 
     private var volumeBinding: Binding<Double> {
         Binding(
-            get: { app.volume },
-            set: { mixer.setVolume($0, for: app) }
+            get: { app.volume < 1 ? sqrt(app.volume) : app.volume },
+            set: { mixer.setVolume($0 < 1 ? pow($0, 2) : $0, for: app) }
```
vorssaint: only the slider position bends here, the saved gain and the percent stay linear and 100% and 200% stay where they were, which is the shape left open in #1588. two places still disagree with the curve. the island mixer that landed on main after this branch drives the same app volume through its own fader in notch mixer view with a plain binding, so with the island turned on a saved 50% sits at a different spot in the panel and in the island and each one drags on a different curve. mixer volume slider also takes its voiceover value from this binding, so a row showing 25% is announced as 50%. could the curve live in two small helpers in mixer routing support, gain to slider position and back, used here and by the island fader, with mixer volume slider taking the percent to announce the way notch level slider takes its value label, and a check in mixer feature tests that 0, 100% and 200% map to themselves and that a position round trips?

ahmedabdullah-o: alright, what if we add a binding function to the MixerRoutingSupport enum that takes app and return a binding for it? it's gonna be exponential and then we'll be able to use it with every slider in the app UI. I think it's gonna be much cleaner. It’s gonna be a single source of truth for all sliders in the app. if you're gonna go back to linear, change the exponential or take a whole different approach to calculating slider position it's gonna be a one-line change.
tell me what you think of this approach.

vorssaint: a single place for the curve is the right goal, but mixer routing support is compiled for the test build without the mixer or its app type, so a function there that takes the app and returns a binding would break that build. two plain conversions there, gain to slider position and back, give the same single place and let mixer feature tests check them, and the panel row and the island fader can each build their binding from those two. keep the curve on the app volume sliders only, the output and input sliders set the system volume itself and sit where the system's own sliders show it, so a curve there would pull them apart. the voiceover part stays either way, mixer volume slider announces the value it is bound to, so it still needs the gain percent passed in the way the island fader passes its value label

